Issue #13204, opened 2018-07-24
In UI tests automatically check if all outgoing links have noopener noreferrer


diosmosis:
After https://github.com/matomo-org/matomo/pull/12780 is merged most links will have `noopener noreferrer`. To make sure all links now & in the future have this link, we should add a check to the UI test system:

- After a screenshot test passes, scrape the page in phantomjs and look for links that are to other domains.
- For each of these links, if it is missing `noopener noreferrer`, report an error and fail the test.

tsteur:
Instead there could be also a JS that runs onDomReady, onDomLoad and when components are updated through angular or so to check for any outgoing link in the modified dom and set it dynamically.

tsteur:
Since browsers now set `noopener` by default we could close this issue I suppose. It be mostly something for IE11.

see https://www.chromestatus.com/feature/6140064063029248
https://bugzilla.mozilla.org/show_bug.cgi?id=1522083
https://webkit.org/blog/8475/release-notes-for-safari-technology-preview-68/
https://chromium-review.googlesource.com/c/chromium/src/+/1630010

tsteur:
Also we could be sending a `Cross-Origin-Opener-Policy` header. However, probably most of the browsers that support this header also apply noopener by default.

tsteur:
For noopener we don't need to do any changes. Generally though, if we add `noreferrer` anyway, then we could also add `noopener` as well.

@Findus23 do we still need noreferrer in links considering we have https://github.com/matomo-org/matomo/pull/17382 ?

tsteur:
@Findus23 closing this one as the referrer policy header should take already care of this. Let me know if that's not the case

PR #12780, opened 2018-04-28
Replace proxy redirect with rel=noreferrer

c960657:
[All modern browsers](https://caniuse.com/#search=noreferrer) support `rel=noreferrer`. This seems like a better and easier way to link to external sites.

When browsers eventually get support for [Referer-Policy](https://www.w3.org/TR/referrer-policy/), all these `rel=noreferrer` can be replaced with a single HTTP header.

PR #13479, opened 2018-09-25
don't send referrer to plugin authors website

Findus23:
Followup to https://github.com/matomo-org/matomo/pull/12780 and related to #13204

I noticed seeing foreign Matomo domains in my reports which seems to be caused by this missing `noreferrer`.


sgiehl (approved):
good catch :+1:

plugins/CorePluginsAdmin/templates/macros.twig
```diff
@@ -162,7 +162,7 @@
...
+                                                <a class="donation-link paypal" target="_blank" rel="noreferrer noopener" href="https://www.paypal.com/cgi-bin/webscr?cmd=_donations&item_name=Matomo%20Plugin%20{{ name|escape('url') }}&bn=PP%2dDonationsBF%3abtn_donateCC_LG%2egif%3aNonHosted&business={{ plugin.info.donate.paypal|escape('url') }}"><img src="plugins/CorePluginsAdmin/images/paypal_donate.png" height="30"/></a>
                                             {% endif %}
                                             {% if plugin.info.donate.flattr is defined and plugin.info.donate.flattr %}
                                                 <a class="donation-link flattr" target="_blank" href="{{ plugin.info.donate.flattr }}"><img class="alignnone" title="Flattr" alt="" src="plugins/CorePluginsAdmin/images/flattr.png" height="29" /></a>
```
sgiehl: should we add that to the other donation links as well?

PR #13481, opened 2018-09-25
Proper encoding of segment values for Actions reports.

### LINKED ISSUE #13460: #13113 not fixxed - segmented visitors empty if url contains whitespaces
@sgiehl I just took a look at this with your fix applied and it is not fixxed for me.

matomo2

diosmosis:
Changes:
* Encode segmentValue for Actions reports. This extra encoding is required because of how many times a segment is urldecoded:
  * The first decode is on the whole segment: https://github.com/matomo-org/matomo/blob/3.x-dev/core/Segment.php#L104 . The corresponding encode was added in 3.6.0 here: https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/Live/javascripts/SegmentedVisitorLog.js#L123
  * the second decode is on individual conditions (ie, `pageUrl==...`): https://github.com/matomo-org/matomo/blob/3.x-dev/core/Segment/SegmentExpression.php#L91
  * the third decode is on individual values: https://github.com/matomo-org/matomo/blob/3.x-dev/core/Segment/SegmentExpression.php#L112

  This means the value needs to be triple encoded for values like plus signs to be used properly in the segment.

* Change the urldecode in Actions filter to a urlencode, since SafeDecodeLabel will already run a urldecode later. This fixes the display for URLs w/, eg, '+' signs in them (eg, `http://example.org/with+plus/` will display like that, instead of `http://example.org/with plus/`).

I am not sure if this fix will have repercussions so I haven't modified the tests. Will do so after someone else gives their opinion.

Fixes #13460


tsteur:
If it fixes your issue and tests pass probably all good to go. It's almost impossible to review it since the segments are just a mess and we never know where they are urlencoded or urldecoded etc.

diosmosis:
It will change a lot of the test output I think (another encoding in the segment property in API output). Think this won't be a BC break? at least not a potentially huge one.

tsteur:
> Think this won't be a BC break? 

I really don't know. As long as we can still handle the segments urlencoded and urldecoded it should be fine.

fdellwing:
Applied the patch locally and it resolves the issue for me.

core/DataTable/Filter/SafeDecodeLabel.php
```diff
@@ -10,13 +10,21 @@
...
  *
  */
 class SafeDecodeLabel extends BaseFilter
 {
+    /**
+     * On this day we changed the default behavior of this filter. Before it called `urldecode()` on the label
+     * which would turn '+' signs, for instance, into spaces. Now we don't do this, but there may be very old
+     * archived data that has URL encoded labels. So for reports archived before this date, we decode the URL.
+     */
+    const DATE_OF_URLDECODE_CHANGE = '2018-09-25';
```
tsteur: this would imply that the user updated Matomo to that version on that day?

tsteur:
https://github.com/matomo-org/matomo/issues/13460#issuecomment-423856472
>  Don't have a fix yet, but found the cause of the error. Actions are stored urlencoded in the log_action table, but the value in the segment when opening the segmented visitor log/transitions modal/etc. is not. 

I'm not sure this is the case? That might have been the case quite a while ago but shouldn't be the case anymore? at least for me it was stored normally
image

or you mean the URL is encoded after tracking?

plugins/Contents/tests/System/expected/test_Contents__Actions.getPageUrls_month.xml
```diff
@@ -26,6 +26,6 @@
 		<exit_rate>100%</exit_rate>
 		<avg_time_generation>0.333</avg_time_generation>
 		<url>http://www.example.org/page</url>
-		<segment>pageUrl==http%3A%2F%2Fwww.example.org%2Fpage</segment>
+		<segment>pageUrl==http%253A%252F%252Fwww.example.org%252Fpage</segment>
```
tsteur: is this double encoded on purpose?

core/DataTable/Filter/SafeDecodeLabel.php
```diff
@@ -28,18 +36,37 @@ public function __construct($table)
...
      */
-    public static function decodeLabelSafe($value)
+    public static function decodeLabelSafe($value, $shouldUrlDecode = true)
     {
         if (empty($value)) {
             return $value;
         }
-        $raw = urldecode($value);
+
+        $raw = $shouldUrlDecode ? urldecode($shouldUrlDecode) : $value;
```
tsteur: could this be a potential security issue?

tsteur:
I've looked at the PR for quite a while but the PR is quite complex and not trivial to understand especially re side effects etc. Left some comments though. Do you think there would be otherwise a solution where we don't need to change anything except for maybe the segment expression?

plugins/Contents/tests/System/expected/test_Contents__Actions.getPageUrls_month.xml
diosmosis: yes, because one decode is for the condition, and one for the value

core/DataTable/Filter/SafeDecodeLabel.php
diosmosis: no longer relevant

core/DataTable/Filter/SafeDecodeLabel.php
diosmosis: no longer relevant

diosmosis:
> I'm not sure this is the case? That might have been the case quite a while ago but shouldn't be the case anymore?

It's not the case, it looked like it was, but I think I tracked something wrong.

plugins/Contents/tests/System/expected/test_Contents__Actions.getPageUrls_month.xml
mattab: I don't think it's correct, would expect to have here either:
`pageUrl==http%253A%252F%252Fwww.example.org%252Fpage` (current behavior)
or
`pageUrl%3D%3Dhttp%253A%252F%252Fwww.example.org%252Fpage` (current behavior URL encoded)

but `pageUrl==http%253A%252F%252Fwww.example.org%252Fpage` wouldn't be a correct value since Matomo would then segment based on Page URL `http%3A%2F%2Fwww.example.org%2Fpage` which wouldn't exist (didn't check but it's possible it currently works anyway since we try the url decoded version in https://github.com/matomo-org/matomo/blob/2020b122615789c36eaa0917e10b483f12793b3c/core/Segment.php#L99-L108 but still: we wouldn't want to show this double encoded URL in the API output)

diosmosis: It is required to have the value encoded 3 times because it is urldecoded 3 times, see the PR description for more info.

mattab:
@diosmosis Feedback:
- Would it be possible to create a failing UI tests in 3.x-dev branch (in a separate PR) which shows the bug(s) currently in Matomo
- Then could you update this PR and solve conflicts and show that the bug(s) are now fixed.

Looking forward to merging this :+1:

diosmosis:
This change turned out to be rather complex, I doubt I'll be able to finish it soon, and it's not as high value as other tickets. Would vote to put it in the priority backlog.

mattab:
@diosmosis :ok_hand: But be good to finish this later (before Matomo 4 ideally...). Moving it to 3.12.0 milestone for now.

diosmosis:
@mattab  / @tsteur I added a system test for checking the segment values and found a couple more segment related bugs (which are fixed).

diosmosis:
Nevermind, now all the other tests fail.

tsteur:
@diosmosis let me know when it's ready for a review

tsteur:
image

works 👍  code looks good too 👍

PR #13483, opened 2018-09-26
New feature: allow setting goal revenue to event value for event matching goals.

### LINKED ISSUE #13476: When a Goal is created as "Matching event", allow to set Goal value as the event value
When a Goal is created as "Matching event", allow to set Goal value as the event value.

This is useful because in some cases, users have a lot of tracking code in complicated places. The goals to track are hard to know in advance, so they use a lot of Goals matching events. (since they track everything with events already). It works great and can even be used to track multiple [funnels](https://matomo.org/docs/funnels/) for one goal (by duplicating the goals).

The problem is that it's not possible to set the Goal value automatically from the Event value. For example the use case is that the event tracks a currency amount of a payment that was attempted, but failed. So the event value is the "lost revenue" or so, and this value would fit perfectly in the  "Goal revenue".

The goal is to add a new option when creating a goal, and when the goal is set to match an event, where the new option would appear and read eg . "When enabled, and if the event triggering the conversion has a value defined, set Goal's revenue to the event value."

diosmosis:
Fixes #13476

### DIFF
--- a/core/Tracker/GoalManager.php
+++ b/core/Tracker/GoalManager.php
@@ -678,13 +678,7 @@ class GoalManager
             $conversionDimensions = ConversionDimension::getAllDimensions();
             $conversion = $this->triggerHookOnDimensions($request, $conversionDimensions, 'onGoalConversion', $visitor, $action, $conversion);
 
-            if ($this->isEventMatchingGoal($convertedGoal)
-                && !empty($convertedGoal['event_value_as_revenue'])
-            ) {
-                $conversion['revenue'] = ActionEvent::getEventValue($request);
-            }
-
-            $this->insertNewConversion($conversion, $visitProperties->getProperties(), $request, $action);
+            $this->insertNewConversion($conversion, $visitProperties->getProperties(), $request, $action, $convertedGoal);
         }
     }
 
@@ -697,7 +691,7 @@ class GoalManager
      * @param Action|null $action
      * @return bool
      */
-    protected function insertNewConversion($conversion, $visitInformation, Request $request, $action)
+    protected function insertNewConversion($conversion, $visitInformation, Request $request, $action, $convertedGoal = null)
     {
         /**
          * Triggered before persisting a new [conversion entity](/guides/persistence-and-the-mysql-backend#conversions).
@@ -717,6 +711,16 @@ class GoalManager
          * @ignore
          */
         Piwik::postEvent('Tracker.newConversionInformation', array(&$conversion, $visitInformation, $request, $action));
+
+        if (!empty($convertedGoal)
+            && $this->isEventMatchingGoal($convertedGoal)
+            && !empty($convertedGoal['event_value_as_revenue'])
+        ) {
+            $eventValue = ActionEvent::getEventValue($request);
+            if ($eventValue != '') {
+                $conversion['revenue'] = $eventValue;
+            }
+        }
 
         $newGoalDebug = $conversion;
         $newGoalDebug['idvisitor'] = bin2hex($newGoalDebug['idvisitor']);
--- a/core/Tracker/GoalManager.php
+++ b/core/Tracker/GoalManager.php
@@ -158,7 +158,7 @@ class GoalManager
           || ($attribute == 'file' && $actionType != Action::TYPE_DOWNLOAD)
           || ($attribute == 'external_website' && $actionType != Action::TYPE_OUTLINK)
           || ($attribute == 'manually')
-          || $this->isEventMatchingGoal($goal) && $actionType != Action::TYPE_EVENT
+          || self::isEventMatchingGoal($goal) && $actionType != Action::TYPE_EVENT
         ) {
             return null;
         }
@@ -879,7 +879,7 @@ class GoalManager
         return '/' . $pattern . '/';
     }
 
-    private function isEventMatchingGoal($goal)
+    public static function isEventMatchingGoal($goal)
     {
         return in_array($goal['match_attribute'], array('event_action', 'event_name', 'event_category'));
     }
--- a/core/Tracker/GoalManager.php
+++ b/core/Tracker/GoalManager.php
@@ -16,6 +16,7 @@ use Piwik\Piwik;
 use Piwik\Plugin\Dimension\ConversionDimension;
 use Piwik\Plugin\Dimension\VisitDimension;
 use Piwik\Plugins\CustomVariables\CustomVariables;
+use Piwik\Plugins\Events\Actions\ActionEvent;
 use Piwik\Tracker;
 use Piwik\Tracker\Visit\VisitProperties;
 
@@ -157,7 +158,7 @@ class GoalManager
           || ($attribute == 'file' && $actionType != Action::TYPE_DOWNLOAD)
           || ($attribute == 'external_website' && $actionType != Action::TYPE_OUTLINK)
           || ($attribute == 'manually')
-          || in_array($attribute, array('event_action', 'event_name', 'event_category')) && $actionType != Action::TYPE_EVENT
+          || $this->isEventMatchingGoal($goal) && $actionType != Action::TYPE_EVENT
         ) {
             return null;
         }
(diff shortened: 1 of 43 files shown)


core/Tracker/GoalManager.php
```diff
@@ -677,6 +678,12 @@ protected function recordStandardGoals(VisitProperties $visitProperties, Request
             $conversionDimensions = ConversionDimension::getAllDimensions();
             $conversion = $this->triggerHookOnDimensions($request, $conversionDimensions, 'onGoalConversion', $visitor, $action, $conversion);
 
+            if ($this->isEventMatchingGoal($convertedGoal)
+                && !empty($convertedGoal['event_value_as_revenue'])
+            ) {
+                $conversion['revenue'] = ActionEvent::getEventValue($request);
```
tsteur: Don't know if it's any important or makes a difference... usually for the revenue column we do a 
```php
$revenue = Common::getRequestVar('revenue', $defaultGoalRevenue, 'float', $this->params);
$revenue = BaseConversion::roundRevenueIfNeeded($revenue)
```

I presume this logic cannot be in a dimension itself and be defined through the "onGoalConversion"? I can see various problems like "revenue  might be overwritten by other dimension", "it might not know if the goal is an event matching a goal etc", "it doesn't know that flag "event_value_as_revenue" is enabled, ... so all good if it is not possible.

I wonder though if it still worked if the conversion revenue is defined before line 679 before we trigger the hook, so dimension can still change it if wanted/needed. But this may not be something we really want and we may want to enforce that revenue in this case so all good probably.

plugins/Goals/API.php
```diff
@@ -158,6 +159,7 @@ public function addGoal($idSite, $name, $matchAttribute, $pattern, $patternType,
             'allow_multiple'  => (int)$allowMultipleConversionsPerVisit,
             'revenue'         => $revenue,
             'deleted'         => 0,
+            'event_value_as_revenue' => (int) $useEventValueAsRevenue,
```
tsteur: do we need to add a check that setting can be only true when using a event match attribute?

plugins/Goals/angularjs/manage-goals/manage-goals.controller.js
```diff
@@ -165,14 +167,15 @@
...
-            initGoalForm('Goals.addGoal', _pk_translate('Goals_AddGoal'), '', '', 'url', '', 'contains', /*caseSensitive = */false, /*allowMultiple = */'0', '0');
+            initGoalForm('Goals.addGoal', _pk_translate('Goals_AddGoal'), '', '', 'url', '', 'contains', /*caseSensitive = */false, /*allowMultiple = */'0', /*useEventValueAsRevenue = */ false, '0');
             scrollToTop();
         }
 
         this.editGoal = function (goalId) {
             this.showAddEditForm();
             var goal = piwik.goals[goalId];
-            initGoalForm("Goals.updateGoal", _pk_translate('Goals_UpdateGoal'), goal.name, goal.description, goal.match_attribute, goal.pattern, goal.pattern_type, (goal.case_sensitive != '0'), goal.revenue, goal.allow_multiple, goalId);
+            console.log(goal);
```
tsteur: console.log is maybe there by accident?

plugins/Goals/templates/_formAddGoal.twig
```diff
@@ -120,6 +120,13 @@
             </div>
         </div>
 
+        <div piwik-field uicontrol="checkbox" name="use_event_value"
```
tsteur: Can we move this after case sensitive? That's because case sensitive is much more valuable option and more often used maybe than the event option and be good for consistency to always have case sensitive directly below. All good though to keep it also.

tsteur: be good to add some inline help to explain what it does actually

core/Tracker/GoalManager.php
diosmosis: I thought about putting it in a dimension, but was unsure as to which dimension to put it in. Eg, does it go in EventValue and access goal data there? Or in a goal dimension?

I could maybe put it in an event handler in Goals if we make sure it overrides other changes. Or maybe a RequestProcessor.

tests/PHPUnit/Fixtures/ThreeVisitsWithCustomEvents.php
```diff
@@ -91,7 +92,7 @@ protected function trackEventWithoutUrl(PiwikTracker $vis)
     {
         $url = $vis->pageUrl;
         $vis->setUrl('');
-        self::checkResponse($vis->doTrackEvent('CategoryTriggersGoal here', 'This is an event without a URL'));
+        self::checkResponse($vis->doTrackEvent('CategoryTriggersGoal here', 'This is an event without a URL', $name = false, $value = 23));
```
tsteur: be good to also add a system test for this to see that it was attributed in Goals report?

plugins/Goals/API.php
diosmosis: I can add one, but given it shouldn't be used unless it's an event match attribute, I didn't think it would be necessary.

tsteur: it's only to avoid problems if people use the API to create events... they might not know it can only be used for events or so and prevents wrong expectations I guess

plugins/Goals/templates/_formAddGoal.twig
diosmosis: Was hoping the translated label would be clear enough, is there something confusing about it or something that needs explanation?

tests/PHPUnit/Fixtures/ThreeVisitsWithCustomEvents.php
diosmosis: There should be revenue in the existing Goals.get report, is that what you were thinking of or a new test entirely?

plugins/Goals/API.php
diosmosis: I see, that makes sense. A clear error message would be useful.

core/Tracker/GoalManager.php
tsteur: Yeah not sure either. Would need to be a dimension I suppose but not sure if the dimension would all the information it needs? I think it's all good like this just in case we don't know which order things are executed or so. Was just a random thought I wanted to mention :)

plugins/Goals/templates/_formAddGoal.twig
tsteur: image

It would be good to say something like:

"Enable this setting if you are sending an event value along with your event tracking request. Any number you send as event value will be also used as the revenue instead of the configured Goal revenue. It may be useful to enable this option if the event value you are tracking also matches the revenue it is worth to you. If the revenue is always the same and does not vary per conversion, we recommend configuring the goal default revenue below."

There might be more we want to add. I'm sure @mattab has some thoughts on this. I think there are only very few cases when you want to enable this setting as it can mess with your revenue quite a bit.

Can we maybe also disable the setting "Goal default revenue is (optional)" setting when this setting is enabled? Or maybe we show the setting there? By the looks in the code it would always overwrite the revenue with the event revenue and never use the goal revenue.

tests/PHPUnit/Fixtures/ThreeVisitsWithCustomEvents.php
tsteur: Sorry I missed that expected change. Thought I looked at them all but obviously didn't notice that one. all good.

plugins/Goals/templates/_formAddGoal.twig
tsteur: could work. would need to see it though I guess. unless @mattab has any specific thoughts

mattab: Sounds good to put them together and add some inline help as suggested. the label is clear but not so much if you are not familiar with the concepts, so the descriptive inline help text makes things more clear and removes any possible confusion👍

diosmosis:
@tsteur moved the option to next to the default revenue option. Here's what it looks like:

image

tsteur:
👍  I would maybe clarify whether the event value overwrites any default value, or what happens when no event value is set

diosmosis:
Updated.

core/Tracker/GoalManager.php
tsteur: @diosmosis just checking: wouldn't this always overwrite the default revenue? Even if eventValue is `0` for example or none set?

diosmosis: Good point, will fix w/ test.

tsteur:
There are some failing tests, otherwise LGTM

PR #13487, opened 2018-09-27
Fix incorrect sort order when scheduled report uses custom report

diosmosis:
Make sure request parameters are loaded in superglobals when DataTablePostProcessor gets the Report instance.

DataTablePostProcessor will call `ReportsProvider::factory()`, but it will call it outside of `Piwik\Api\Proxy`, which means the request parameters won't be in `$_GET/$_POST`. So if the Report instance tries to access these in order to configure itself, it may not be able to find them.

### DIFF
--- a/core/API/Request.php
+++ b/core/API/Request.php
@@ -12,6 +12,7 @@ use Exception;
 use Piwik\Access;
 use Piwik\Cache;
 use Piwik\Common;
+use Piwik\Context;
 use Piwik\DataTable;
 use Piwik\Exception\PluginDeactivatedException;
 use Piwik\IP;
@@ -261,7 +262,12 @@ class Request
             // call the method
             $returnedValue = Proxy::getInstance()->call($apiClassName, $method, $this->request);
 
-            $toReturn = $response->getResponse($returnedValue, $module, $method);
+            // get the response with the request query parameters loaded, since DataTablePost processor will use the Report
+            // class instance, which may inspect the query parameters. (eg, it may look for the idCustomReport parameters
+            // which may only exist in $this->request, if the request was called programatically)
+            $toReturn = Context::executeWithQueryParameters($this->request, function () use ($response, $returnedValue, $module, $method) {
+                return $response->getResponse($returnedValue, $module, $method);
+            });
         } catch (Exception $e) {
             Log::debug($e);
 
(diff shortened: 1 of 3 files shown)


core/Context.php
```diff
@@ -13,6 +13,23 @@
...
 class Context
 {
+    public static function executeWithQueryParameters(array $parametersRequest, callable $callback)
+    {
+        // Temporarily sets the Request array to this API call context
+        $saveGET = $_GET;
+        $saveQUERY_STRING = @$_SERVER['QUERY_STRING'];
+        foreach ($parametersRequest as $param => $value) {
+            $_GET[$param] = $value;
+        }
```
tsteur: wouldn't it also need to set $_POST or unset it temporarily etc?

core/API/Request.php
```diff
@@ -261,7 +262,12 @@ public function process()
             // call the method
             $returnedValue = Proxy::getInstance()->call($apiClassName, $method, $this->request);
 
-            $toReturn = $response->getResponse($returnedValue, $module, $method);
+            // get the response with the request query parameters loaded, since DataTablePost processor will use the Report
```
tsteur: @diosmosis does it make sense to move `$returnedValue = Proxy::getInstance()->call($apiClassName, $method, $this->request);` into this `Context::executeWithQueryParameters` below and remove it from the `call` method? The call method isn't API so we can change it.

core/Context.php
diosmosis: Possibly, this code was moved from API\Proxy.php, I didn't change it. We can make it override `$_POST` too, just didn't want to accidentally break something.

core/API/Request.php
diosmosis: Some code does call `Proxy::getInstance()->call()` directly, eg, `DataTableManipulator`s so I think it could break things.

diosmosis:
@tsteur Updated to save $_POST parameters as well.

core/Context.php
tsteur: sorry... only just noticing... I wonder if it breaks functionality if we also did `$_GET = array();$_POST=array()` before the foreach? but I think it actually could break functionality. not needed for now I guess as it probably has always been that way.

tsteur: actually pretty sure it would break behaviour... this is why we currently use `$default = []` and needs to be changed there I suppose.

tsteur (approved):
some ui tests are failing in https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/30298/ but doesn't look to be related to this change 👍

PR #13493, opened 2018-09-28
 Report tracking into wrong Site ID and missing token auth

### LINKED ISSUE #12922: Report tracking into wrong Site ID
We often see users tracking traffic into no longer existing siteIds or siteIds that never existed.

In https://github.com/matomo-org/matomo/issues/7550 we already have an issue to report back tracking errors. As this is quite a big project, I think it would be valuable to start with some more simple use case. 

When a site is being tracked into a website that doesn't exist anymore  (we could have the highest siteId as option and check if `$trackedSiteId < $highestSiteId`). If so, we could report this in some way to the super user (as the site no longer exist it cannot be reported to any admin or view user).
FYI: I wouldn't show it for `$trackedSiteId > $highestSiteId` as you could send lots of random requests to users every day annoying them by getting notifications about wrong requests and bloating tracker cache etc.

The question is where to store this info efficiently and where to show it.

Regarding storing: We could store it in a simple table or option table or something. To minimize DB requests during tracking we would have in cache `highestSiteId` and all currently wrong tracked siteIds. This way we hit the DB only once per site. 
We would maybe reset the list of "wrong tracked siteIds" daily or weekly. It would be also good to store along this info at what time we received the first wrong tracking request for that site and on which URL.

Regarding showing the notice: Not sure. Maybe weekly we could send a summary mail if there is one or multiple siteIds still in use? In the admin we could maybe have a screen for this under diagnostics and maybe a notification if data is tracked into a wrong siteId?

tsteur:
fix #12922

* When a tracking request is invalid for any of the 2 reasons, we log the request as a failure
* Any failed request older than 2 days ago is automatically deleted again
* Weekly we check for tracking failures and email super users only (doing this for admins is bit tricky)
* Admins and SuperUsers see a new widget on Admin Home screen plus it can be used in the dashboard:
image
* New diagnostics page that shows tracking failures in depth
image

I was also thinking about showing a box in the reporting toolbar, eg on the right... but then thought it may not be needed for now and might be rather annoying. May be good to first see how many false reports maybe come in etc.

I reckon it is a good working solution v1 for this tool and see from there how/where it goes. Heaps could be done there.

Todo: We should write FAQ articles for those 2 particular errors and adjust the links in the code


mattab:
@tsteur just thought of another use case which occurs from time to time, if Matomo is in an unstable state, there may be columns missing, this happened to me regularly, such errors:
```sql

DEBUG: Exception: Error query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'last_idlink_va' in 'field list' In query: SELECT visit_last_action_time, visit_first_action_time, idvisitor, idvisit, user_id, visit_exit_idaction_url, visit_exit_idaction_name, visitor_returning, visitor_days_since_first, visitor_days_since_order, visitor_count_visits, visit_goal_buyer, location_country, location_region, location_city, location_latitude, location_longitude, referer_name, referer_keyword, referer_type, idsite, visit_entry_idaction_url, visit_total_actions, visit_total_interactions, visit_total_searches, config_device_brand, config_device_model, config_device_type, visit_total_events, visit_total_time, location_ip, location_browser_lang, campaign_content, campaign_id, campaign_keyword, campaign_medium, campaign_name, campaign_source, last_idlink_va, custom_var_k1, custom_var_v1, custom_var_k2, custom_var_v2, custom_var_k3, custom_var_v3, custom_var_k4, custom_var_v4, custom_var_k5, custom_var_v5   FROM log_visit WHERE visit_last_action_time >= ? AND visit_last_action_time <= ? AND idsite = ? AND config_id = ?
DEBUG:                 ORDER BY visit_last_action_time DESC
DEBUG:                 LIMIT 1 Parameters: array (
DEBUG:   0 => '2018-10-03 05:21:30',
DEBUG:   1 => '2018-10-03 06:21:30',
DEBUG:   2 => 1,
DEBUG:   3 => '/ƫK���',
DEBUG: )
```

So it would be really great if we could detect these problems and also report them in the page :+1:

core/Tracker/Failures.php
```diff
@@ -0,0 +1,178 @@
...
+        }
+        return Date::now();
+    }
+
+    public function logFailure($idFailure, Request $request)
+    {
+        $visitExcluded = new VisitExcluded($request);
+        if ($visitExcluded->isExcluded()) {
+            return;
+        }
```
diosmosis: `VisitExcluded::isExcluded()` is called during tracking correct? If it were possible to set the result in `VisitProperties`, we could use that value here instead of checking if the request should be excluded again. Might be a bit faster?

diosmosis: `VisitExcluded::isExcluded()` is called during tracking correct? If it were possible to set the result in `VisitProperties`, we could use that value here instead of checking if the request should be excluded again. Might be a bit faster?

core/Tracker/Failures.php
```diff
@@ -0,0 +1,178 @@
...
+            if (isset($params[$key])) {
+                $params[$key] = '__ANONYMIZED__';
+            }
+        }
+        foreach ($params as $key => $value) {
+            if (!empty($token) && $value === $token) {
+                $params[$key] = '__ANONYMIZED__'; // user accidentally posted the token in a wrong field
+            } elseif (!empty($value) && is_string($value) && Common::mb_strlen($value) >= 29 && Common::mb_strlen($value) <= 36 && ctype_xdigit($value)) {
+                $params[$key] = '__ANONYMIZED__'; // user maybe posted a token in a different field... it looks like it might be a token
+            }
```
diosmosis: Would it be simpler to do `str_replace($token, '__ANONYMIZED__', $value)` on each param?

Also I think __ANONYMIZED__ might be confusing to some people, since requests are also anonymized by changing IP addresses. Maybe `__TOKEN_AUTH__` might be clearer?

core/Tracker/Failures.php
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    private function enrichFailures($failures)
+    {
+        foreach ($failures as &$failure) {
+            try {
+                $failure['site_name'] = Site::getNameFor($failure['idsite']);
+            } catch (UnexpectedWebsiteFoundException $e) {
+                $failure['site_name'] = Piwik::translate('General_Unknown');
+            }
```
diosmosis: If a site is not known would it be more useful to have the idSite displayed in the name? Not sure if the idSite is also displayed in the UI.

core/Tracker/Failures.php
```diff
@@ -0,0 +1,178 @@
...
+            switch ($failure['idfailure']) {
+                case self::FAILURE_ID_INVALID_SITE:
+                    $failure['problem'] = Piwik::translate('CoreAdminHome_TrackingFailureInvalidSiteProblem');
+                    $failure['solution'] = Piwik::translate('CoreAdminHome_TrackingFailureInvalidSiteSolution');
+                    $failure['solution_url'] = 'https://matomo.org/faq/todo...';
+                    break;
+                case self::FAILURE_ID_NOT_AUTHENTICATED:
+                    $failure['problem'] = Piwik::translate('CoreAdminHome_TrackingFailureAuthenticationProblem');
+                    $failure['solution'] = Piwik::translate('CoreAdminHome_TrackingFailureAuthenticationSolution');
+                    $failure['solution_url'] = 'https://matomo.org/faq/todo...';
```
diosmosis: Guess these todo links need to be filled out before merging.

core/Tracker/Failures.php
```diff
@@ -0,0 +1,178 @@
...
+                    break;
+            }
+        }
+
+        Piwik::postEvent('Tracking.makeFailuresHumanReadable', array(&$failures));
```
diosmosis: Should this have an @ignore or docs?

core/Tracker/Request.php
```diff
@@ -522,8 +524,11 @@ public function getIdSite()
          */
         Piwik::postEvent('Tracker.Request.getIdSite', array(&$idSite, $this->params));
 
-        if ($idSite <= 0) {
+        if ($idSite < 0) {
             throw new UnexpectedWebsiteFoundException('Invalid idSite: \'' . $idSite . '\'');
+            // we allow 0... the request will fail anyway as the site won't exist... allowing 0 will help us
+            // reporting this tracking problem as it is a common issue. Otherwise we would not be able to report
+            // this problem in tracking failures
```
diosmosis: Does this mean the exception cannot be caught later? If so, could log the failure there.

plugins/CoreAdminHome/Tasks.php
```diff
@@ -133,6 +143,36 @@ private function rememberTrackingCodeReminderRan($idSite)
...
+     * To test execute the following command:
+     * `./console core:run-scheduled-tasks "Piwik\Plugins\CoreAdminHome\Tasks.cleanupTrackingFailures"`
+     *
+     * @throws \Exception
+     */
+    public function cleanupTrackingFailures()
+    {
+        // we remove possibly outdated/fixed tracking failures that have not occurred again recently
+        $this->trackingFailures->removeFailuresOlderThanDays(Failures::CLEANUP_OLD_FAILURES_DAYS);
+    }
```
diosmosis: If this deletes tracking failures that are older than 2 days, and the notify task is run weekly, then that means the email will never report failures or not report all failures (since they can be removed before being sent off). Is that correct or am I missing something?

plugins/CoreAdminHome/lang/en.json
```diff
@@ -108,6 +109,24 @@
...
+        "Problem": "Problem",
+        "Solution": "Solution",
+        "TrackingURL": "Tracking URL",
+        "LearnMore": "Learn more",
+        "DeleteAllFailures": "Delete all failures",
+        "NTrackingFailures": "%s tracking failures",
+        "ViewAllTrackingFailures": "View all tracking failures",
+        "TrackingFailureInvalidSiteProblem": "The site does not exist.",
+        "TrackingFailureInvalidSiteSolution": "Update the configured idSite in the tracker.",
+        "TrackingFailureAuthenticationProblem": "Request was not authenticated but should have.",
```
diosmosis: Would suggest a change to `Request was not authenticated but authentication was required.` (if I am understanding the error correctly)

plugins/CoreAdminHome/lang/en.json
```diff
@@ -108,6 +109,24 @@
...
+        "TrackingFailureAuthenticationSolution": "Set or correct a \"auth_token\" in your tracking request.",
```
diosmosis: auth_token or token_auth?

plugins/CoreAdminHome/lang/en.json
```diff
@@ -108,6 +109,24 @@
...
+        "ConfirmDeleteAllTrackingFailures": "Are you sure you want to delete all tracking failures?",
+        "ConfirmDeleteThisTrackingFailure": "Are you sure you want to delete this tracking failure?",
+        "TrackingFailuresEmailSubject": "Tracking failures in your Matomo Analytics",
+        "TrackingFailuresEmail1": "This is just to let you know that %s different kind of tracking failures have occurred in the last days.",
```
diosmosis: kind => kinds

core/Tracker/Failures.php
diosmosis: Nevermind, see the screenshot below.

diosmosis:
Left some comments & need to resolve some conflicts. Haven't tested locally yet, will do so next review.

core/Tracker/Failures.php
tsteur: good idea 👍  adding it

core/Tracker/Failures.php
tsteur: Will change it to `__TOKEN_AUTH__ `. didn't use only `str_replace($token, '__ANONYMIZED__', $value)` in case there were any typos or just one character missing or one too much.

core/Tracker/Failures.php
tsteur: Created the FAQs and will update 👍

core/Tracker/Failures.php
tsteur: yes 👍  will add an ignore for now, and add proper docs as soon as we use it in another plugin. don't want to make it an API just yet before not having actually verified it works well like this.

core/Tracker/Request.php
tsteur: Yes it would otherwise not be caught later. and logging wouldn't work as people wouldn't see the logged error. Like nobody would really look there and eg on cloud people wouldn't even be able to see the logs. We have heaps of logs like `idsite=0` already and want to make users aware of this problem with this PR. It's really the most common tracking error.

plugins/CoreAdminHome/Tasks.php
tsteur: It would report failures that happened in the last 2 days before the email was sent. Eg if email is sent on Sunday at 8am, it would report failures that happened between Friday 8am and Sunday 8am. But not any problems before that... the assumption is that those problems have been fixed basically, or occurred maybe only once but are not a permanent problem.

plugins/CoreAdminHome/angularjs/trackingfailures/trackingfailures.controller.js
```diff
@@ -0,0 +1,59 @@
...
+                piwikApi.fetch({method: 'CoreAdminHome.deleteAllTrackingFailures'}).then(function () {
+                    self.fetchAll();
+                });
+            }});
+        };
+
+        this.deleteFailure = function (idSite, idFailure) {
+            piwik.helper.modalConfirm('#confirmDeleteThisTrackingFailure', {yes: function () {
+                self.failures = [];
+                piwikApi.fetch({method: 'CoreAdminHome.deleteTrackingFailure', idSite: idSite, idFailure: idFailure}).then(function () {
```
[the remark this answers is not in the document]
tsteur: I don't think it's needed? Not sure what the difference would be... should be all fine to delete it through GET and the token will be posted anyway.

plugins/CoreAdminHome/lang/en.json
tsteur: Will update it 👍

plugins/CoreAdminHome/lang/en.json
tsteur: token_auth 👍  will update

plugins/CoreAdminHome/lang/en.json
tsteur: cheers 👍  will update

core/Tracker/Failures.php
diosmosis: I see, 👍

core/Tracker/Request.php
diosmosis: I meant log through `Failures::logFailure()`, like catch in `Tracker::main()` and log that failure. But if it can't be caught then that's not an option I guess.

plugins/CoreAdminHome/Tasks.php
diosmosis: I guess for tracking code issues it's probably not a big deal, people are like invested in making sure they work at first, and then if they break later, it will be a consistent problem. So makes sense to me.

plugins/CoreAdminHome/angularjs/trackingfailures/trackingfailures.controller.js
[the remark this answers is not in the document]
diosmosis: A GET will be cached in the browser?

diosmosis:
Left one further comment, but will probably merge this today. if my comment is important can be done in a new PR

tsteur:
@diosmosis just realising a feature to log not auth request regressed due to another PR... because `isVisitExcluded` seems to no longer work basically when CIP is set... be good to not merge.

tsteur:
@diosmosis just pushing a fix, and also fixing the failed system and ui tests with this commit

diosmosis:
Looks like there are some other test failures now? the novisit tests are failing (and possibly a unit test). Maybe some others too.

tsteur:
this was tricky... hopefully this time it fixes most or all of the tests :)

mattab:
Will be very useful :+1: 

> Not really so much possible. Cause it was said it is not needed to support anything else and therefore it is not built to include metadata etc. If something like this is possible, it would be only possible to say a column was missing, but not which column or table etc.

Would it be possible to catch these sql error cases, and report them as failures in this new tool's UI?

tsteur:
No, at least not as part of 3.8 release.

diosmosis:
Almost working, looks like BulkTracking & QueuedTracking have some failing integration tests

tsteur:
@diosmosis fixed the bulk and queued tracking tests. Looks like `MarketingCampaignsReporting ` isn't directly failing because of this change. Hopefully can be finally merged :)

PR #13496, opened 2018-09-28
Flattened "entry page" and "exit page" take a long time to be generated

### LINKED ISSUE #11645: flattened "entry page" and "exit page" take a long time to be generated
In an environment, having around 90k pageviews per month, the flattened entry page report and exit page report take some time to be generated:

selecting top 10, takes around 10 seconds,
selecting top 50, takes around 19 seconds,
selecting the default top 100, takes around 22 second

If one would compare this to the pages report top 100 plattened report, it loads in around 4 seconds. In the past the flattened pages report were slow but the speed of those report was improved at some point (I don't recall what commit caused the improvements).

Could the Entry page and Exit page report be a candidate for the same loadtime improvements?

tsteur:
fix https://github.com/matomo-org/matomo/issues/11645

Debugged this on demo.matomo.org.... 

Wasn't as guessed https://github.com/matomo-org/matomo/blame/3.5.1/plugins/Actions/DataTable/Filter/Actions.php#L65-L67  but the flat parameter. After applying it the performance improved from minutes to less than a second.

### DIFF
--- a/plugins/Actions/API.php
+++ b/plugins/Actions/API.php
@@ -167,11 +167,11 @@ class API extends \Piwik\Plugin\API
      * Returns a DataTable with analytics information for every unique entry page URL, for
      * the specified site, period & segment.
      */
-    public function getEntryPageUrls($idSite, $period, $date, $segment = false, $expanded = false, $idSubtable = false)
+    public function getEntryPageUrls($idSite, $period, $date, $segment = false, $expanded = false, $idSubtable = false, $flat = false)
     {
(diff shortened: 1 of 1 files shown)


sgiehl (approved):
Good catch. Didn't have a closer look with debugger, but seems to be logical, as that way the fattening is already handled by the archive creator already and not in the end by DataTablePostProcessor

tassoman:
Thank you team! :medal_sports:

PR #13497, opened 2018-09-28
Add max_allowed_packet diagnostic check

### LINKED ISSUE #1781: New diagnostic checking that max_allowed_packet is at least 64M or issue warning, and recommend 128M
From http://dev.mysql.com/doc/refman/5.1/en/packet-too-large.html

```
When a MySQL client or the mysqld server receives a packet bigger than max_allowed_packet bytes, it issues a Packet too large error and closes the connection. With some clients, you may also get a Lost connection to MySQL server during query error if the communication packet is too large.
```

Default server configuration is 1MB.  For larger sites, it needs to be increased to MEDIUMBLOB (16MB).

tsteur:
fix #1781 

thought about linking to https://matomo.org/faq/troubleshooting/faq_183/ but the FAQ is actually not very helpful.

### DIFF
--- a/plugins/Diagnostics/Diagnostic/DbMaxPacket.php
+++ b/plugins/Diagnostics/Diagnostic/DbMaxPacket.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Piwik\Plugins\Diagnostics\Diagnostic;
+
+use Piwik\Db;
+use Piwik\MetricsFormatter;
+use Piwik\Piwik;
+use Piwik\SettingsPiwik;
+use Piwik\Translation\Translator;
+
+/**
+ * Check if Piwik is connected with database through ssl.
+ */
+class DbMaxPacket implements Diagnostic
+{
+    /**
+     * @var Translator
+     */
+    private $translator;
+
+    const MIN_VALUE_MAX_PACKET_MB = 32;
+
+    public function __construct(Translator $translator)
+    {
+        $this->translator = $translator;
+    }
+
+    public function execute()
+    {
+        if (!SettingsPiwik::isPiwikInstalled()) {
+            return array(); // only possible to perform check once we have DB connection
+        }
+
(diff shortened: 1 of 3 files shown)


plugins/Diagnostics/Diagnostic/DbMaxPacket.php
```diff
@@ -0,0 +1,50 @@
...
+        $status = DiagnosticResult::STATUS_OK;
+        $label = $this->translator->translate('Diagnostics_MysqlMaxPacketSize');
+        $comment = '';
+
+        $minSize = self::MIN_VALUE_MAX_PACKET_MB * 1000 * 1000; // not using 1024 just in case... this amount be good enough
+        if (!empty($maxPacketBytes['Value']) && $maxPacketBytes['Value'] < $minSize) {
+            $status = DiagnosticResult::STATUS_WARNING;
+            $pretty = MetricsFormatter::getPrettySizeFromBytes($maxPacketBytes['Value'], 'M');
+            $configured = str_replace(array(' M', '&nbsp;M'), 'MB', $pretty);
+            $comment = Piwik::translate('Diagnostics_MysqlMaxPacketSizeWarning', array('32MB', $configured));
```
fdellwing: This should not be hardcoded I think.

sgiehl: The referenced issue suggest to set it to at least 64MB, but guess 32MB might be fine as well?

sgiehl (approved):
Left a small comment. Otherwise looks good and worked locally.

plugins/Diagnostics/Diagnostic/DbMaxPacket.php
tsteur: I don't remember exactly but we talked about it in chat or so. Somewhere 32 was said. Can change it to 64.

PR #13500, opened 2018-09-29
Tweaks to escaping strategy in a couple places

diosmosis:


plugins/Annotations/templates/getEvolutionIcons.twig
```diff
@@ -5,7 +5,7 @@
         <span data-date="{{ date }}" data-count="{{ counts.count }}" data-starred="{{ counts.starred }}"
               {% if counts.count == 0 %}title="{{ 'Annotations_AddAnnotationsFor'|translate(date) }}"
               {% elseif counts.count == 1 %}title="{{ 'Annotations_AnnotationOnDate'|translate(date,
-              counts.note)|raw }}
+              counts.note)|e('html_attr') }}
```
sgiehl: Theoretically it would be enough to only escape `counts.note`. That would allow the translation to contain html tags (which it currently doesn't have)

sgiehl (commented):
Locally the encoding in site selector looks fine when creating a site with special chars in the name. But the UI tests shows the values encoded. See https://builds-artifacts.matomo.org/matomo-org/matomo/email-escaping/30265/SiteSelector_loaded.png

diosmosis:
@sgiehl Updated.

sgiehl (approved):
Test failures seem to be unrelated to this changes so should be good to merge

PR #13501, opened 2018-09-29
Ability to change periods over which evolution graphs display in scheduled reports

diosmosis:
Changes:
* Add new INI config param `[General] graphs_default_evolution_graph_last_days_amount` to match other params.
* Add new option to scheduled reports to allow specifying per report, what periods are displayed in evolution graphs.

Tests don't pass just yet, need to add a new file.


diosmosis:
It's required for a cloud report, so if moved to 3.7.0, then we'd have to wait until 3.7.0 before it can be used.

CC @mattab

plugins/ImageGraph/ImageGraph.php
```diff
@@ -174,4 +174,10 @@ public function getReportMetadata(&$reports, $info)
             }
         }
     }
+
+    public static function getDefaultGraphEvolutionLastPeriods()
+    {
+        $lastPeriods = Config::getInstance()->General['graphs_default_evolution_graph_last_days_amount'];
```
tsteur: should we maybe fail if someone configures a value <= 0? so the user can see it was not correctly changed? not needed though...

plugins/ScheduledReports/API.php
```diff
@@ -86,18 +87,22 @@ public function __construct(LoggerInterface $logger)
...
      * @param array $parameters array of parameters
      * @param bool|int $idSegment Segment Identifier
+     * @param bool $evolutionPeriodFor If true, evolution graphs cover each day within the period. If false, evolution graphs
+     *                                 cover the previous N periods.
+     * @param int|null $evolutionPeriodN The previous N periods to query in evolution graphs if $isEvolutionGraphEachDay is true.
      *
      * @return int idReport generated
      */
-    public function addReport($idSite, $description, $period, $hour, $reportType, $reportFormat, $reports, $parameters, $idSegment = false)
+    public function addReport($idSite, $description, $period, $hour, $reportType, $reportFormat, $reports, $parameters, $idSegment = false, $evolutionPeriodFor = 'prev',
```
tsteur: `$evolutionPeriodFor = 'prev'` doesn't match the php docs where it says a boolean?

plugins/ScheduledReports/API.php
```diff
@@ -86,18 +87,22 @@ public function __construct(LoggerInterface $logger)
      * @param array $reports array of reports
      * @param array $parameters array of parameters
      * @param bool|int $idSegment Segment Identifier
+     * @param bool $evolutionPeriodFor If true, evolution graphs cover each day within the period. If false, evolution graphs
```
tsteur: I'm actually not quite understanding the difference between the two but reckon will get more clear once having a look at it.

plugins/ScheduledReports/API.php
```diff
@@ -345,88 +359,98 @@ public function generateReport($idReport, $date, $language = false, $outputType
...
-                $mustRestoreGET = $_GET;
-                $_GET['enhanced'] = true;
-
-                if ($apiAction == 'getAll') {
-                    $_GET['filter_truncate'] = false;
-                    $_GET['filter_limit'] = -1; // show all websites in all websites report
+        $originalShowEvolutionWithinSelectedPeriod = Config::getInstance()->General['graphs_show_evolution_within_selected_period'];
+        $originalDefaultEvolutionGraphLastPeriodsAmount = Config::getInstance()->General['graphs_default_evolution_graph_last_days_amount'];
+        try {
+            Config::getInstance()->General['graphs_show_evolution_within_selected_period'] = (bool)$report['evolution_graph_within_period'];
```
tsteur: I haven't tested it yet but are you sure this works? I think in the past usually it was often needed to do

`$general = $config->General;$general['...']=(bool)....; $config->General = $general;`. Not sure if this is still a thing...

Also I'm not sure this is actually applied... without testing or looking further I would have expected this needs to be set before calling `\Piwik\Plugins\API\API::getInstance()->getReportMetadata($idSite);`

diosmosis: I remember that was for specific versions of PHP. I can do that to be safe.

plugins/ScheduledReports/API.php
diosmosis: `$evolutionPeriodFor` determines if looking within range or using lastN value. I originally only had one parameter `$evolutionPeriodLastN` where if it was <= 0, it would look at days within the range, but that's not really clear w/ just one parameter, so I opted for being explicit.

plugins/ScheduledReports/API.php
diosmosis: Note: it's used in `API.getProcessedReport`, but can set it before all the API calls.

plugins/ScheduledReports/templates/_addReport.twig
```diff
@@ -92,6 +92,25 @@
                  ng-show="manageScheduledReport.report.displayFormat == '2' || manageScheduledReport.report.displayFormat == '3'"
                  title="{{ 'ScheduledReports_EvolutionGraph'|translate(5)|e('html_attr') }}">
             </div>
+
+            <div
+                class="row evolution-graph-period"
+                ng-show="manageScheduledReport.report.displayFormat == '2' || manageScheduledReport.report.displayFormat == '3'"
```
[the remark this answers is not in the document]
diosmosis: It has an effect, so I added it.

plugins/ScheduledReports/lang/en.json
```diff
@@ -53,6 +53,8 @@
         "SuccessfullyUnsubscribed": "You have been successfully unsubscribed from the report %1$s.",
         "UnsubscribeFooter": "To unsubscribe from this report please follow this link: %1$s",
         "NoTokenProvided": "No token was provided in the URL",
-        "NoSubscriptionFound": "No subscription found. Maybe the report was already unsubscribed or removed."
+        "NoSubscriptionFound": "No subscription found. Maybe the report was already unsubscribed or removed.",
+        "EvolutionGraphsShowForEachInPeriod": "Evolution graphs show the evolution for %1$seach day%2$s in the last %3$s",
```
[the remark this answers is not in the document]
diosmosis: should be 'show' since 'evolution graphs' is plural (in this case the option is for all evolution graphs in the report, which is why it's plural in the text)

diosmosis:
@tsteur Updated the PR and fixed the issue you noticed.

plugins/ScheduledReports/API.php
tsteur: > I remember that was for specific versions of PHP. I can do that to be safe.

might be good. I don't really remember why it was needed or whether it's still the case. Haven't tested to write it the normal way in a while as gotten used to it :)

tsteur:
@diosmosis looks good. Be maybe great to change the config access... I reckon it be good to have a method `Config::setSetting($category, $name, $value)` which sets it in a save way that works cross PHP versions? Seeing myself repeat that code quite a few times...

Otherwise: I've updated to Mac OS Mojave and GD+freetype isn't working anymore so cannot test the graphs but otherwise looks good 👍

diosmosis:
Ok, will merge after Config::setSetting change.

tsteur:
LGTM

PR #13504, opened 2018-09-29
Avoid creating any archive tables for future dates

sgiehl:
Under certain circumstances Archiving might create empty `archive_numeric` tables for future dates:
- `enable_browser_archiving_triggering` is set to `0`
- requesting a date range in the future. e.g. something like `&period=range&date=2018-01-01,2019-09-29`

### DIFF
--- a/core/DataAccess/ArchiveSelector.php
+++ b/core/DataAccess/ArchiveSelector.php
@@ -166,6 +166,9 @@ class ArchiveSelector
         $monthToPeriods = array();
         foreach ($periods as $period) {
             /** @var Period $period */
+            if ($period->getDateStart()->isLater(Date::now()->addDay(2))) {
+                continue; // avoid creating any archive tables in the future
+            }
             $table = ArchiveTableCreator::getNumericTable($period->getDateStart());
             $monthToPeriods[$table][] = $period;
         }
--- a/tests/PHPUnit/Integration/ArchiveTest.php
+++ b/tests/PHPUnit/Integration/ArchiveTest.php
@@ -8,6 +8,7 @@
 namespace Piwik\Tests\Integration;
 
 use Piwik\API\Proxy;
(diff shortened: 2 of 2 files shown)


core/DataAccess/ArchiveSelector.php
```diff
@@ -166,6 +166,9 @@ public static function getArchiveIds($siteIds, $periods, $segment, $plugins)
         $monthToPeriods = array();
         foreach ($periods as $period) {
             /** @var Period $period */
+            if ($period->getDateStart()->isLater(Date::now()->addDay(2))) {
+                continue; // avoid creating any archive tables in the future
+            }
```
diosmosis: Should we add this code to ArchiveTableCreator? I can't think of a reason for code to create a table in the future (except possibly tests, though probably not). If some other code uses ArchiveTableCreator directly, this safeguard would be applied there too.

sgiehl: Not sure if it might make sense for some custom plugins to create archives in the future. e.g. for some kind of prediction reports of whatever. Nevertheless the ArchiveSelector shouldn't trigger the creation

diosmosis: Makes sense.

diosmosis:
Is it possible to add a test for this? Otherwise looks good to me.

sgiehl:
@diosmosis  what shall the test check? That no archive table is created when ArchiveSelector::getNumeicTable is called with a future date?

diosmosis:
That could work. Or a system/integration test that uses a date in the future w/ VisitsSummary.get and has a check that the table wasn't created.

mattab:
Test sounds good :+1: then good to merge

sgiehl:
I've added two simple tests. (Note: The first of the tests would currently fail on 3.x-dev branch, as an archive table is created)

Issue #13507, opened 2018-09-29
Error message in various places when invalid parameters are used


mattab:
as seen on production many times:

## first error message (more below)
Argument 1 passed to Piwik\\Plugins\\UsersManager\\Model::getUsers() must be of the type array, null given, called in /plugins\/UsersManager\/API.php on line 351

URL: index.php?module=API&method=UsersManager.getUsers&format=Tsv&token_auth=XYZANONYMIZED&translateColumnNames=1&userLogins[]=&userLogins[]=

mattab:
### And this second one

Error: {"message":"Argument 1 passed to Piwik\\Period::__construct() must be an instance of Piwik\\Date, array given, called in \/core\/Period\/Factory.php on line 91","file":"\///core\/Period.php","line":61}


URL: index.php?date[$acunetix]=1&format=rss&idSite=1&method=Transitions.getTransitionsForPageUrl&module=API&pageUrl=http://forum.piwik.org/&period=day&token_auth=XYZANONYMIZED&translateColumnNames=1

mattab:
## And this third one

Error: {"message":"Argument 1 passed to Piwik\\Period::__construct() must be an instance of Piwik\\Date, array given, called in \/core\/Period\/Factory.php on line 91","file":"\/core\/Period.php","line":61}


URL: index.php?date[]=yesterday&format=Tsv&idSite=1&method=Overlay.getFollowingPages&module=API&period=day&token_auth=XYZANONYMIZED&translateColumnNames=1&url=http://forum.piwik.org/

mattab:
## And this fourth one:


Error: {"message":"Call to undefined method Piwik\\DataTable\\Map::getRowFromLabel()","file":"\/homewww\/*.innocraft.cloud\/plugins\/UserCountry\/API.php","line":45}


URL: index.php?module=API&method=API.getRowEvolution&idSite=1%2c%201%2c%202%2c%203)%3bdeclare%20@q%20varchar(99)%3bset%20@q%3d'%5c%.irtree%20@q%3b--%20&period=day&date=last10&apiModule=UserCountry&apiAction=getCountry&legendAppendMetric=1&labelUseAbsoluteUrl=1&format=rss&token_auth=XYZANONYMIZED&translateColumnNames=1

mattab:
## And this fifth one

[...]

PR #13511, opened 2018-09-30
Fix various notices caused by invalid parameter values

sgiehl:
refs #13507


plugins/API/API.php
```diff
@@ -449,6 +449,10 @@ public function get($idSite, $period, $date, $segment = false, $columns = false)
      */
     public function getRowEvolution($idSite, $period, $date, $apiModule, $apiAction, $label = false, $segment = false, $column = false, $language = false, $idGoal = false, $legendAppendMetric = true, $labelUseAbsoluteUrl = true, $idDimension = false)
     {
+        // check if site exists
+        $idSite = (int) $idSite;
+        $site = new Site($idSite);
```
diosmosis: I'm not sure if this is **really** an issue, but someone could check if a site exists by querying this method.

PR #13512, opened 2018-09-30
Ensure segment param is used from url if no hash available

sgiehl:
When viewing widgets in widgetize mode, all parameters are directly contained in the query string instead of the hash. When appending a segment to the query string it wasn't automatically added to ongoing ajax request.

### DIFF
--- a/plugins/Morpheus/javascripts/ajaxHelper.js
+++ b/plugins/Morpheus/javascripts/ajaxHelper.js
@@ -529,10 +529,16 @@ function ajaxHelper() {
      */
     this._mixinDefaultGetParams = function (params) {
 
+        if (window.location.hash) {
+            var segment = broadcast.getValueFromHash('segment', window.location.href.split('#')[1]);
+        } else {
+            var segment = broadcast.getValueFromUrl('segment');
+        }
+
         var defaultParams = {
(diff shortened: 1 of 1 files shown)


plugins/Morpheus/javascripts/ajaxHelper.js
```diff
@@ -529,10 +529,16 @@ function ajaxHelper() {
      */
     this._mixinDefaultGetParams = function (params) {
 
+        if (window.location.hash) {
```
diosmosis: If the hash isn't a valid query string (or has no segment), should we default to the segment in the URL?

sgiehl: thought about it. but wasn't sure about any side effects. e.g. opening the dashboard with having a segment in the query string and in the hash. If it would be removed from the hash it would still load the one in query string which might not be correct/wanted

diosmosis: Hmm, not sure about that. Thought this was how it worked for other query params (like date/period), but maybe not... cc @mattab

sgiehl: Don't think the segment of the query string should be used if the one from the hash gets removed (be choosing no segment in UI)

PR #13515, opened 2018-10-01
Improve parameter handling in UsersManager.getUsers

### LINKED ISSUE #13507: Error message in various places when invalid parameters are used

sgiehl:
aims to fix https://github.com/matomo-org/matomo/issues/13507#issue-365156020

### DIFF
--- a/plugins/UsersManager/API.php
+++ b/plugins/UsersManager/API.php
@@ -343,7 +343,12 @@ class API extends \Piwik\Plugin\API
     {
         Piwik::checkUserHasSomeAdminAccess();
 
+        if (!is_string($userLogins)) {
+            throw new \Exception('Parameter userLogins needs to be a string containing a comma separated list of users');
+        }
+
         $logins = array();
+
         if (!empty($userLogins)) {
             $logins = explode(',', $userLogins);
         }


diosmosis:
Can a test be added to UsersManager/tests/Integration/APITest.php?

plugins/UsersManager/API.php
```diff
@@ -336,15 +336,18 @@ public function getUsersPlusRole($idSite, $limit = null, $offset = 0, $filter_se
...
      * @return array the list of all the users
      */
     public function getUsers($userLogins = '')
     {
         Piwik::checkUserHasSomeAdminAccess();
 
         $logins = array();
-        if (!empty($userLogins)) {
+
+        if (is_array($userLogins)) {
```
tsteur: or should we just fail if userLogins is not a string? In theory you may be able to send now some weird array structures to maybe cause an error? (haven't tested, just a thought)

sgiehl: don't have a strong opinion on that. failing if no string is given works as well for sure

tsteur: lets do this to keep things simple 👍

tsteur (approved):

PR #13517, opened 2018-10-01
Show unique visitors on visitor map as default if available

### LINKED ISSUE #13456: Locations users country map should use Unique visitors by default
image

All the other reports show unique visitors by default, but the map shows visits. Then the numbers don't end match up. Ideally it would show unique visitors by default when available. 

Also when switching to unique visitors in the map, in the top left corner it still shows "visits"

sgiehl:
fixes #13456


plugins/UserCountryMap/javascripts/visitor-map.js
```diff
@@ -464,31 +464,33 @@
...
-                    $('.map-stats').html(
-                        _.nb_visits.replace('%s', '<strong>' + formatNumber(totalVisits, metric) + '</strong>') + (id != 'world' ? ' (' +
-                            formatPercentage(totalVisits / worldTotalVisits) + ')' : '')
-                    );
-                }
+                var data = {};
+                data[metric] = totalMetricValue;
+                $('.map-stats').html(
+                    '<strong>' + formatValueForTooltips(data, metric, false) + '</strong>' +
+                    (id != 'world' ? (' (' + formatPercentage(totalMetricValue / self.config.visitsSummary[metric]) + ')') : '')
```
diosmosis: should the percentage calculation be the other way around? ie, `self.config.visitsSummary[metric] / totalMetricValue`? And it's expected that for the first and last case in the if statement (`id.length == 3` and `else`), that it will just be 100%, right?

sgiehl: No. Should be correct that way. Maybe the names of the variables are a bit misleading. `totalMetricValue` holds the metric value for the selected/shown region (e.g. world,continent,country), and that is compared to the overall total. So `self.config.visitsSummary[metric]` is 100%. For the worldmap that would be always 100%, so it's now shown at all...

diosmosis: Makes sense, 👍

PR #13414, opened 2018-09-12
Add new category profiles plus new visitors report

tsteur:
We could also add a Real-Time section as the current real time widgets are not exposed in the UI...

image

Real-Time Map etc may be still better under "Visitors".

Renamed "Visitor Log" to "Visits Log" as this is more accurate, but could also keep the original name.

The "Visitors" report is queried live and not archived as we would otherwise need to archive thousands or millions of rows to make this usable properly.

PR #13518, opened 2018-10-01
Updated menu structure and naming fixes

tsteur:
Naming fixes:
* Visitors in Real Time => Visits in real time
* Visitor Log => Visits log

Menu additions:
* visits in Real time is now in the visitors menu

Menu changes:
* Engagement => moved from visitors to behaviour
* Slightly reordered reports within Visitors to list "Overview, Visits Log, Real time reports, Location, ... User ID, Custom Variables, Custom Dimensions".

Replaces https://github.com/matomo-org/matomo/pull/13414 which we keep open for now as it contains the "Visitors" report which we won't add for now due to performance reasons.


tsteur:
There are a few more tests to fix but I reckon it is best to fix them after review and closer to the merge re conflicts.

PR #13526, opened 2018-10-03
Improve usage of requested site in plugin Controllers

sgiehl:


plugins/CoreHome/Controller.php
```diff
@@ -177,19 +176,15 @@ protected function setDateTodayIfWebsiteCreatedToday()
...
-        $websiteId = Common::getRequestVar('idSite', false, 'int');
-
-        if ($websiteId) {
-
-            $website = new Site($websiteId);
-            $datetimeCreationDate      = $website->getCreationDate()->getDatetime();
-            $creationDateLocalTimezone = Date::factory($datetimeCreationDate, $website->getTimezone())->toString('Y-m-d');
-            $todayLocalTimezone        = Date::factory('now', $website->getTimezone())->toString('Y-m-d');
+        if ($this->idSite) {
+            $datetimeCreationDate      = $this->site->getCreationDate()->getDatetime();
```
diosmosis: I notice in Plugin/Controller `$this->site` is set to `new Site()`, which can throw if the site doesn't exist. So it might be possible for the site to be null while `$this->idSite` is set. I think we should check for `$this->site` in the if statement to avoid that.

sgiehl: makes sense

PR #13534, opened 2018-10-04
don't post travis results to IRC

### LINKED ISSUE #13531: don't post travis build status to IRC
As no one is left in the #piwik-dev channel, I guess we can also stop posting travis results there.

https://github.com/matomo-org/matomo/blob/9544e8af2a970944470d6153f0ff09d5bc30d4cd/.travis.yml#L152-L160

Does this part get generated by the travis-scripts (I couldn't find it there) or can it just be deleted directly?

Findus23:
fixes #13531

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -149,12 +149,3 @@ notifications:
     on_success: :change
     on_failure: :change
 
-  irc:
-    skip_join: true
-    on_success: change
-    on_failure: change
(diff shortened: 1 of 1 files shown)


sgiehl:
Guess we could remove the slack part as well, as the room it should post the messages imho doesn't exist anymore.

PR #13187, opened 2018-07-20
Changed wording of YouAreOptedIn

electric-paper:
fixes #12954

PR #13538, opened 2018-10-04
Changed wording of YouAreOptedIn

### LINKED ISSUE #12954: Please change wording of "You are currently opted in."
When we added the opt-out iframe in the past we received a complaint that we were asserting consent where no consent had been given. This was due to the message in the opt out frame defaulting to "You are currently opted in." the first time it is displayed (before the user has had a chance to opt out).

While the wording doesn't have any impact in how we actually regard the data collection, it is now more important to communicate that we do not regard a lack of response as positive consent. (e.g. [GDPR section 32](https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX:32016R0679) states "Silence, pre-ticked boxes or inactivity should not therefore constitute consent.")

Therefore I feel that is it safer to express the status (in english) as "not opted out" rather than "opted in".

```json
      'YouAreOptedIn' => 'You are not currently opted out.',
```

Related #12810

tsteur:
fix https://github.com/matomo-org/matomo/issues/12954

replaces https://github.com/matomo-org/matomo/pull/13187
kudos to @electric-paper


plugins/CoreAdminHome/lang/en.json
```diff
@@ -98,8 +98,8 @@
         "UseCustomLogo": "Use a custom logo",
         "ValidPiwikHostname": "Valid Matomo Hostname",
         "WithOptionalRevenue": "with optional revenue",
-        "YouAreOptedIn": "You are currently opted in.",
         "YouAreOptedOut": "You are currently opted out.",
+        "YouAreNotOptedOut": "You are currently not opted out.",
```
diosmosis: I think we can just say "You are not opted out."?

tsteur: good point, will update

diosmosis (approved):
screenshots will probably need an update.

PR #13540, opened 2018-10-04
Use db sessions by default, deprecate file session handler

### LINKED ISSUE #12170: Enable by default to store all session data in the database + remove feature of file-based sessions in tmp/sessions/*
In order to simplify life (for example when we refactor/improve security in our sesions in #12164), I'd like to propose that we remove the File Sessions Handler in Piwik, and default everyone to use the Database session handler.

*  there is no advantage or reason to support the file sessions handler AFAIK.
* Instead the database session handler works well in all cases, especially in the case where multiple piwik frontends are used and sessions must be shared (load balanced Piwik), or when NFS is used to store Piwik files (in which case file sessions are very slow)
* We also support the Redis session handler, [see redis session faq](https://piwik.org/faq/how-to/faq_20521/)

Notes:

* we could then delete this database session FAQ: https://piwik.org/faq/how-to-install/faq_133/
* and update the [load balancer FAQ](https://piwik.org/faq/new-to-piwik/faq_134/) to remove the instructions about DB sessions
* Update the redis session FAQ https://piwik.org/faq/how-to/faq_20521/ and remove `By default Piwik uses the filesystem as a session handler. `

tsteur:
fix https://github.com/matomo-org/matomo/issues/12170

### DIFF
--- a/core/Session/SaveHandler/DbTable.php
+++ b/core/Session/SaveHandler/DbTable.php
@@ -9,7 +9,11 @@
 
 namespace Piwik\Session\SaveHandler;
 
+use Piwik\Common;
 use Piwik\Db;
+use Piwik\DbHelper;
+use Exception;
+use Piwik\Updater\Migration;
 use Zend_Session;
 use Zend_Session_SaveHandler_Interface;
 
@@ -21,6 +25,8 @@ class DbTable implements Zend_Session_SaveHandler_Interface
 {
     protected $config;
     protected $maxLifetime;
+
+    const TABLE_NAME = 'session';
 
     /**
      * @param array $config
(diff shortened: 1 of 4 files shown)


core/Session/SaveHandler/DbTable.php
```diff
@@ -138,8 +175,25 @@ public function gc($maxlifetime)
...
 
-        Db::get()->query($sql, array(time()));
+        $this->query($sql, array(time()));
 
         return true;
     }
+
+    private function migrateToDbSessionTable()
+    {
+        // happens when updating from Piwik 1.4 or earlier to Matomo 3.7+
```
tsteur: Noticed this because of failing upgrade test. Tried to fix it in different ways but couldn't. Seems also quite good fix as will only be executed if table does not exist.

PR #13541, opened 2018-10-04
Only show default number of rows for device type report

### LINKED ISSUE #12266: Device Type report: hide device type without visits
For consistency and clarity, let's remove rows that don't have "visits" in the Device Type report. The report currently looks like this: 
device type

The rows Camera/Car browser/Feature phone/Smart display should be hidden.

tsteur:
fix #12266

fyi @mattab couldn't make it work to show `0` instead of `-`. 
Now all the reports have the same length by default on the devices page. If someone wants to see more, they can increase the limit. Or if many are `-` then they can choose only `5` in the limit selection and it will apply this setting next time.

I reckon this is a good compromise. It lets users also still perform row evolution plus discover all the things/devices we can detect.

image

### DIFF
--- a/plugins/DevicesDetection/Reports/GetType.php
+++ b/plugins/DevicesDetection/Reports/GetType.php
@@ -8,10 +8,10 @@
  */
 namespace Piwik\Plugins\DevicesDetection\Reports;
 
+use Piwik\Common;
 use Piwik\Piwik;
 use Piwik\Plugin\ViewDataTable;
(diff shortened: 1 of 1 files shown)


sgiehl:
@tsteur the `-` can be found here: https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/CoreHome/templates/_dataTableCell.twig#L50
But that's global for all datatables

sgiehl (approved):
LGTM

PR #13543, opened 2018-10-04
Remove alias from user settings.

### LINKED ISSUE #10165: in Settings > Users, "Alias" term may be confusing?
A user has reported that the term "Alias" in Settings > Users, may be confusing, or unclear.

> Quick Feedback: The column "Alias" is a misnomer and should be changed to "Description" oder something else.

Could we maybe rename this term in the UI to `Description` or something else... what do you think?

diosmosis:
* [ ] create issue to remove alias from DB in Matomo 4.0

Fixes #10165


tsteur:
Haven't tested it but looks straight forward 👍 

FYI: You could also already create a PR for Matomo 4 instead of creating an issue if it's easy to do.

PR #13544, opened 2018-10-05
Use different icon for segmented visitor log

### LINKED ISSUE #13419: Same icon is used for "segmented visitor log" and "show visitor profile"
image

Need to replace the "segmented visitor log" icon I would say.

tsteur:
fixes https://github.com/matomo-org/matomo/issues/13419

as discussed using this icon 
image

image


sgiehl (approved):
LGTM. UI Tests will need to be updated before merge.

PR #13546, opened 2018-10-05
Add campaign url builder widget

### LINKED ISSUE #12795: Integrate the URL Builder directly into the app
One of the most useful tools on the website is our URL Builder: https://matomo.org/docs/tracking-campaigns-url-builder/ - This tool should be used generously, as it provides an efficient way to [measure marketing campaigns](https://matomo.org/docs/tracking-campaigns/) and provide insights into conversion rates per channel.

### To make the URL builder as easy and as fast as possible to use, we need to integrate it into Matomo directly. 

-> The question remains: what would be the best way to integrate it? 

Thoughts:

* it needs to be directly accessible from the Campaigns report. 
* Maybe it could be opened in a popover on a click?

tsteur:
fix https://github.com/matomo-org/matomo/issues/12795

image


plugins/Referrers/Widgets/GetCampaignUrlBuilder.php
```diff
@@ -0,0 +1,49 @@
...
+use Piwik\SettingsPiwik;
+
+class GetCampaignUrlBuilder extends \Piwik\Widget\Widget
+{
+    public static function configure(WidgetConfig $config)
+    {
+        $config->setCategoryId('Referrers_Referrers');
+        $config->setSubcategoryId('Referrers_URLCampaignBuilder');
+        $config->setName('Referrers_URLCampaignBuilder');
+        $config->setIsEnabled(SettingsPiwik::isSegmentationEnabled());
```
sgiehl: Any reason for hiding the widget if segmentation is disabled?

plugins/Referrers/angularjs/campaign-builder/campaign-builder.controller.js
```diff
@@ -0,0 +1,76 @@
+/*!
+ * Piwik - free/libre analytics platform
```
sgiehl: outdated header. but guess we may need to update all files at once one day...

sgiehl (changes requested):
Works as expected. Left some minor comments.
Btw. wouldn't that widget be a good place to suggest to install the `MarketingCampaignsReporting` plugin, if it's not already installed? Guess we already do the same below the campaigns report?

plugins/Referrers/Widgets/GetCampaignUrlBuilder.php
tsteur: Nope, I'll update it.

plugins/Referrers/angularjs/campaign-builder/campaign-builder.controller.js
tsteur: yeah at some point we need to  update them all... will update it here anyway 👍

tsteur:
Pushed the update. We chose to not advertise the plugin there as we don't have installed on all of the accounts on the plugin for now and to keep things simple as we wouldn't want to show such a notice there etc.

sgiehl (approved):

PR #13549, opened 2018-10-05
Fix possible error in visitor profile when no visit found

tsteur:
Not tested but fix this fatal error
> Error: {"message":"Call to a member function getColumn() on boolean","file":"plugins\/Live\/VisitorProfile.php","line":86}

Happened eg on 
> module=API&method=Live.getVisitorProfile&idSite=1&format=JSON&token_auth=XYZANONYMIZED&visitorId=b06bc6d6274c1244&limitVisits=-1


sgiehl (approved):
Wasn't able to reproduce the error locally. But code looks good :+1:

PR #13551, opened 2018-10-07
Manage users: when clicking "edit" button, scroll back to top

### LINKED ISSUE #13379: Manage users: when clicking "edit" button, scroll back to top
in "Manage Users" there might be 20 or more users. When clicking "Edit" button, let's scroll back to the top of the screen otherwise we see juts a grey screen and takes a while to notice to scroll up. 
grey

oswdr:
Fixes #13379


tsteur (approved):
Cheers 👍  works

PR #13552, opened 2018-10-07
Added Fallback Method for Alexa in SEO Plugin

### LINKED ISSUE #13427: SEO plugin Alexa broken
related to https://github.com/matomo-org/matomo/issues/12248

While checking how the SEO plugin works, I noticed that it parses this url:
https://github.com/matomo-org/matomo/blob/b8e4d27f2151103df826e5dd643b433cdde05aaf/plugins/SEO/Metric/Alexa.php#L20

But it seems like this API doesn't exist anymore (besides the fact that sending the urls via HTTP to other servers isn't a good idea)

http://data.alexa.com/data?cli=10&url=matomo.org
only returns `Okay` for me and it seems like I am not the only one:
https://stackoverflow.com/questions/3676376/fetching-alexa-data#comment87619041_6224304

ozdemirburak:
fixes #13427

Greetings, 

I've added an alternative method for fetching the global ranking from Alexa.

On the other hand, calling `http://data.alexa.com/data?cli=10&url=DOMAIN` is working currently in my case, however I had this issue in the previous month, so I do not know whether they are blacklisting or limiting IP ranges or not.

Finally, if one unexpectedly calls this new method frequently, probably will end up with being blacklisted.


Findus23:
Hi,

Many thanks for your contribution. Can you explain what the fallback does?

I still only get `Okay` when I request http://data.alexa.com/data?cli=10&url=example.com

ozdemirburak:
Hello,

First, if an exception occurs, like in your case, then it sends a HTTP request to public Alexa ranking page of the website, for instance https://www.alexa.com/siteinfo/example.com, then matches the global ranking and local ranking rows.

Since that HTML is kinda dirty, it [replaces comments and multiple spaces with a single space](https://github.com/matomo-org/matomo/pull/13552/commits/5e5818559f6448b39be84a8cc45dc354cdfa2e63#diff-00a42427dfbda7c954fd6f2126263332R57) first, then what we get is something like below.

```html
<strong class="metrics-data align-vmiddle">19,460</strong>
<strong class="metrics-data align-vmiddle">6,517</strong>
```



BTW, this is my output right now, queried from Turkish IP address.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!-- Need more Alexa data?  Find our APIs here: https://aws.amazon.com/alexa/ -->
<ALEXA VER="0.9" URL="example.com/" HOME="0" AID="=" IDN="example.com/">
<SD><POPULARITY URL="example.com/" TEXT="19460" SOURCE="panel"/><REACH RANK="16581"/><RANK DELTA="-3629"/><COUNTRY CODE="IN" NAME="India" RANK="6517"/></SD></ALEXA>
```

Findus23:
Ah, I misunderstood your code. Now it makes sense and at least according to https://stackoverflow.com/questions/50279057/alexa-site-rank-api your solution is the only way that works anymore (you could maybe try https://www.alexa.com/minisiteinfo/stackoverflow.com as it is probably simmer)

I have now tried from multiple networks including an university network and iPv6 and in never worked for me.

ozdemirburak:
Updated the URL, and now using DOMXPath if it is OK to filter out the node value since it will need a complex and not good looking regex to filter out that value.

PR #13554, opened 2018-10-08
Update session cookie timeout on each request if session was remembered.

### LINKED ISSUE #13327: “Remember me” doesn’t, very much
I expect “Remember Me” to keep my session alive at the very *least* until I have not used it for thirty days.

Instead, it looks like it produces a cookie which dies after two weeks? That’s *super* annoying.

Please let me stay logged in either indefinitely or at least until I haven’t used it for a month.

diosmosis:
Fixes #13327


sgiehl (approved):
Code looks good to me

Issue #5267, opened 2014-05-30
Display a new  row  "Total" showing the sum of values for all DataTable columns


mattab:
Since #57 we process in Metadata Reports, a new `<reportTotal>`

For example in [this metadata api output](https://github.com/piwik/piwik/blob/master/tests/PHPUnit/Integration/expected/test_ecommerceOrderWithItems__API.getProcessedReport_day.xml#L89-97)

```
    <reportTotal>
        <nb_visits>3</nb_visits>
        <nb_uniq_visitors>2</nb_uniq_visitors>
        <nb_actions>13</nb_actions>
        <nb_conversions>3</nb_conversions>
        <bounce_count>0</bounce_count>
        <revenue>3121</revenue>
    </reportTotal>
```

For Actions.getPageUrl here is the total ([source code](https://github.com/piwik/piwik/blob/master/tests/PHPUnit/Integration/expected/test_OneVisitorTwoVisits_showColumns___API.getProcessedReport_day.xml#L56-60)):

```
    <reportTotal>
        <nb_hits>4</nb_hits>
        <entry_nb_visits>2</entry_nb_visits>
    </reportTotal>
```

## The goal of this issue is to:
- Check/ensure that the &lt;reportTotal> includes the total for all columns that can be summed
- Add a new Cog icon option to show the Report Total in a new row in the report
- By default we don't enable it currently, but we may enable it in the future

Feel free to add your feedback as a comment here!

tsteur:
fyi: looking into report totals calculator this currently only works for a few defined metrics that can be easily summed. It doesn't work for avg, min, max, or any processed metric etc. This won't be too easy to develop.

You'd think using `sumRow` might help like this (as it supports min, max, and custom aggregators): https://github.com/matomo-org/matomo/compare/5267?expand=1

but even this won't work in all cases. Eg many filters and processors wouldn't be applied, depending on the report you may even need to apply the queued filters first or later after summing all rows etc. Also it wouldn't process any processed metrics. for many reports it would likely work just like that but not for all and not in all cases depending on which features are used.

[...]

PR #13555, opened 2018-10-08
Some work on report totals

tsteur:
@mattab I would need someone to check all reports and various features to check if the totals make sense and when it breaks and when it works etc.

A bonus is that it shows on hover percentage values for some `%` rate columns. The shown percentage here is basically compared to the average (would need to somehow adjust hover text). Doesn't work for time metrics though because then other percentages break. 

While this implementation could work in some cases, I have no idea if that always works or shows correct data. 

If this implementation doesn't work as expected, we may need to schedule this for 3.8 or 3.9 if it is important as it would be a lot of work.

refs https://github.com/matomo-org/matomo/issues/5267


core/API/DataTableManipulator/ReportTotalsCalculator.php
```diff
@@ -83,30 +84,54 @@ protected function manipulateDataTable($dataTable)
...
+        $request['disable_generic_filters'] = 1;
+
+        $processor = new DataTablePostProcessor($this->report->getModule(), $this->report->getAction(), $request);
+        $clone = $processor->process($clone);
+
+        foreach ($clone->getRows() as $row) {
+            if ($row->getColumn('label') === 'Total12345') {
+
+                $this->totals = $row->getColumns();
+                $dataTable->setMetadata('totals', $this->totals);
```
tsteur: We could likely also set here something like `$dataTable->setTotalsRow()` so plugins can at least have the chance to hook onto it... 

It's not that easy though. Tried eg this:

```diff
diff --git a/core/API/DataTableManipulator/ReportTotalsCalculator.php b/core/API/DataTableManipulator/ReportTotalsCalculator.php
index 4baac2f939..b1439f61ae 100644
--- a/core/API/DataTableManipulator/ReportTotalsCalculator.php
+++ b/core/API/DataTableManipulator/ReportTotalsCalculator.php
@@ -127,6 +127,8 @@ class ReportTotalsCalculator extends DataTableManipulator
             if ($row->getColumn('label') === 'Total12345') {
 
                 $this->totals = $row->getColumns();
+                $row->setColumn('label', 'Totals');
+                $dataTable->setTotalsRow($totalRow);
                 $dataTable->setMetadata('totals', $this->totals);
                 break;
             }
diff --git a/core/DataTable.php b/core/DataTable.php
index 22b09b97ee..9189170dac 100644
--- a/core/DataTable.php
+++ b/core/DataTable.php
@@ -301,6 +301,11 @@ class DataTable implements DataTableInterface, \IteratorAggregate, \ArrayAccess
      */
     protected $summaryRow = null;
 
+    /**
+     * @var \Piwik\DataTable\Row
+     */
+    protected $totalsRow = null;
+
     /**
      * Table metadata. Read [this](#class-desc-the-basics) to learn more.
      *
@@ -402,6 +407,16 @@ class DataTable implements DataTableInterface, \IteratorAggregate, \ArrayAccess
         }
     }
 
+    public function setTotalsRow(Row $totals)
+    {
+        $this->totalsRow = $totals;
+    }
+
+    public function getTotalsRow()
+    {
+        return $this->totalsRow;
+    }
+
     /**
      * Returns the name of the column this table was sorted by (if any).
      *
diff --git a/core/Plugin/Visualization.php b/core/Plugin/Visualization.php
index 614aefcffa..82129b8f8b 100644
--- a/core/Plugin/Visualization.php
+++ b/core/Plugin/Visualization.php
@@ -209,9 +209,8 @@ class Visualization extends ViewDataTable
         }
 
         if ($this->dataTable && $this->dataTable instanceof DataTable) {
-            $totals = $this->dataTable->getMetadata('totals');
+            $totals = $this->dataTable->getTotalsRow();
             if (!empty($totals)) {
-                $totals['label'] = 'Totals';
                 $this->dataTable->addRow(new DataTable\Row(array(DataTable\Row::COLUMNS => $totals)));
             }
         }
```
but breaks eg totals feature when flattening as it seems to create an empty clone currently and not copy totals row etc (to be implemented)

tsteur:
Should work

core/API/DataTableManipulator/ReportTotalsCalculator.php
```diff
@@ -83,30 +80,76 @@ protected function manipulateDataTable($dataTable)
...
+            $totals = $row->getColumns();
+            unset($totals['label']);
+            $dataTable->setMetadata('totals', $totals);
+
+            if (1 === Common::getRequestVar('keep_totals_row', 0, 'integer', $this->request)) {
+                $row->deleteMetadata(false);
+                $row->setColumn('label', Piwik::translate('General_Totals'));
+                $dataTable->setTotalsRow($row);
+            }
+        }
```
diosmosis: I guess this code is what you are referring to in the ticket description? It definitely seems complicated... If the main issue for not doing it the simple way is BC (ie, it might break some reports), maybe we can whitelist which reports support it?

Eg, in the Report class there could be a `supportsTotalsRow()` method that defaults to false, then we could make sure core reports support it even if we have to tweak the API code (eg, changing MultiSites to ignore the totals row). What do you think?

tsteur: Yes, I think now it works quite well though, and might actually even work better than before but not sure if I am missing something.

diosmosis: I was thinking about having to clone the datatable and make accomadations for specific filters/processed metrics/etc. What do you think about the idea w/ `supportsTotalsRow()`?

tsteur: Sure could add it, but I think it's not really needed. Can't hurt though maybe to add it and we just set it to `true` by default so a report can opt out. Before it was basically supported on pretty much all reports so be great to still have it on those reports.
It should support pretty much all reports as it is now using 

```
$tableMeta = $firstLevelTable->getMetadata(DataTable::COLUMN_AGGREGATION_OPS_METADATA_NAME);
$totalRow->sumRow($row, $copyMetadata = false, $tableMeta)
``` 
instead of doing the sum in the code directly where I noticed even some problems but forgot which ones. There's some "hack" for some reports that don't add the processed metrics in the report directly `$clone->filter('AddColumnsProcessedMetrics', array($deleteRowsWithNoVisit = false));`  but otherwise it should be fine.

diosmosis:
Would it help to make the totals row bold to differentiate it from other rows (since this row will stay even while paginating)? Eg:

image

diosmosis:
I suppose the percent doesn't need to be shown here:

image

diosmosis:
Doesn't seem to work w/ goals columns:

image

diosmosis:
don't think it's a big deal, but I noticed the total row option will display, even if there's no data in the report:

image

I'm sure will understand, but it seems a bit odd.

diosmosis:
Still need support for goal columns if possible. Unless it's hard to do I guess?

tsteur:
@diosmosis I just remember a report can already disable totals row through viewDataTable config `show_totals_row` and can also set a request parameter `totals=0` so I'm thinking it is actually better to remove the `supportsTotalsRow` method again which only applies to one visualisation kind of anyway (and the API but there it shouldn't cause a problem, it's more visualisation specific). Keeps the report class more simple.

PR #13558, opened 2018-10-08
don't consider .swf files safe

Findus23:
While writing a better nginx config I noticed that swf doesn't fit in the list, as I wouldn't consider accessing random swf files safe (https://github.com/matomo-org/github-issues-mirror/issues/5).

This shouldn't break anything as there is no reason for swf files in Matomo.


tsteur:
should be fine if tests pass IMO

tsteur (approved):

PR #12695, opened 2018-04-04
add woff2 fonts

Findus23:
fixes #12662 

Woff2 is supported by [every modern browser](https://caniuse.com/#feat=woff2) and [the rest](https://caniuse.com/#feat=woff) falls back to woff or worst case ttf.
Woff2 uses brotli for compression and is therefore far smaller than woff:
```text
-rw-r--r-- 1 lukas lukas  73K Mär 23 10:07 matomo.svg
-rw-r--r-- 1 lukas lukas  26K Mär 23 10:07 matomo.ttf
-rw-r--r-- 1 lukas lukas  26K Mär 23 10:07 matomo.woff
-rw-r--r-- 1 lukas lukas  11K Apr  4 10:31 matomo.woff2
-rw-r--r-- 1 lukas lukas  71K Feb 10 18:17 piwik.svg
-rw-r--r-- 1 lukas lukas  25K Feb 10 18:17 piwik.ttf
-rw-r--r-- 1 lukas lukas  25K Feb 10 18:17 piwik.woff
-rw-r--r-- 1 lukas lukas 9,9K Apr  4 10:31 piwik.woff2
```
I have also changed the order because every browser uses the first font it supports and therefor woff wasn't used before.

The file has been generated with the official google tool: https://github.com/google/woff2

I guess we'll also need to update the server config as in #11091.

PR #13559, opened 2018-10-08
Woff2 test

### LINKED ISSUE #13210: Create automated test to make sure woff2 files are up to date
Refs #12695 

One way might be to just run woff2_compress in a test and see if the result is the same as what's in git. A test like this would probably be added to ReleaseCheckListTest. Would need to add woff2 to travis.

woff2 is not available as a package on travis, but it seems it can be built from source (and then added to a travis cache directory), see:

https://github.com/ManageIQ/font-fabulous/blob/master/.travis.yml

Findus23:
related to #12695 and fixes #13210

While @diosmosis' idea of running `woff2_compress` on travis is probably the best solution, I think checking if both files have been modified at the same time should also work and is much simpler.

The test seems to work locally, but someone should take a look (and afterwards all files need to be updated to fix the test)


tsteur:
Can you maybe test if it fails when updating the woff files but not woff2 as a test? if then the build fails it should work 👍

Findus23:
When you commit woff, but not woff2 the time difference between the two commits should be more than 24 hours and the test should fail.
But you probably won't get that far as without updating the woff2 the browser won't show the changes during development.

Findus23:
I'm not entirely sure why the test didn't fail:
https://travis-ci.org/matomo-org/matomo/jobs/438743756

tsteur:
fyi in case it helps what I'm getting:

```
$ git log -1 --format='%ad' plugins/Morpheus/fonts/matomo.woff
Wed May 2 07:59:51 2018 +1200
$ git log -1 --format='%ad' plugins/Morpheus/fonts/matomo.woff2
Wed Jul 25 21:54:48 2018 +0200
```

tsteur:
I don't know if it is supposed to fail right now?
Could echo some information on travis... eg `var_dump($woff_last_change);var_dump($woff2_last_change);`

diosmosis:
Started debugging this on the travis-debug branch, for some reason the date for matomo.woff is the same as the date for matomo.woff2, but only on travis.

diosmosis:
I already gave up on debugging it :)

Findus23:
It is supposed to fail now as they weren't committed at the same time and my plan was then to recommit them afterwards to fix it.
Thanks for confirming that it's not me going crazy :slightly_smiling_face:

tsteur:
Maybe https://github.com/travis-ci/travis-ci/issues/7254#issuecomment-387689631 helps?

Findus23:
@tsteur That makes sense. Travis probably does a `git clone --max-depth=0` and therefore doesn't know what happened before. But I guess a full clone would be far slower, so I guess this test isn't possible.

tsteur:
👍  shall we close this PR then?

diosmosis:
Was hoping getting the log from the origin remote would get around the lack of history, but it didn't...

Findus23:
Unfortunatly

diosmosis:
Thought of a way to do the test this way, could use the github API to get the last modified time (using whatever travis environment variable gets the commit hash). Probably wouldn't work on PRs, but doesn't really have to. @Findus23 would you have time for this? Found an example request here: https://stackoverflow.com/questions/50194241/get-when-the-file-was-last-updated-from-a-github-repository

diosmosis:
Made the change to use the github API and ran into another problem, rate limiting due to travis-ci usage. Need to create a bare oauth token and use as a secure travis env variable if want to continue down this path.

cc @tsteur / @mattab

mattab:
> Need to create a bare oauth token and use as a secure travis env variable if want to continue down this path.

@diosmosis Is this the only task left to do? if so I can do this. Can you send the link to the github guide to follow?

diosmosis:
@mattab should follow https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line and give the token *no* permissions. Then need to add it as a secure environment variable to travis-ci: https://docs.travis-ci.com/user/environment-variables/#encrypting-environment-variables

tsteur:
@mattab @diosmosis what's the status here? @mattab I suppose you need to do something here?

diosmosis:
@tsteur two options:

- Add a github oauth token to travis-ci so we can make API requests to the github API for this one check.
- Or somehow get `woff2_compress` to be installed on travis-ci.

I don't know if either are high value enough to spend time on.

tsteur:
@diosmosis now that we use chrome for tests, maybe chrome loads the woff2 anyway and this test is not needed anymore?

diosmosis:
@tsteur this test is to make sure the woff2 file is up to date, since it has to be generated manually after every change to the icon font file. So we need to run woff2_compress in travis and compare w/ the file or some other way to make sure we know it is up to date. I guess it's not super important if we don't have this test, but I suspect someone will forget w/o the check.

diosmosis:
Or do you mean because chrome will try to use woff2 first? If so, I guess we don't need this then...

tsteur:
> Or do you mean because chrome will try to use woff2 first? If so, I guess we don't need this then...

That's what I meant. We're mentioning it also first in the font so likely this should be the case.

diosmosis:
Ok, I'll close this then. No issue for me if someone wants to re-open and finish this, but it doesn't seem necessary anymore.

PR #13563, opened 2018-10-09
limit Sparkline size

Findus23:


core/Visualization/Sparkline.php
```diff
@@ -92,7 +94,7 @@ public function getWidth() {
      * @param int $width
      */
     public function setWidth($width) {
-        if (!is_numeric($width) || $width <= 0) {
+        if (!is_numeric($width) || $width <= 0 || $width > self::MAX_WIDTH) {
```
sgiehl: would it maybe make sense to set width/height to maximum values if something greater is given?

Findus23: Makes sense as one also doesn't have to test around to find out the maximum that way.

Findus23: Done :heavy_check_mark:

sgiehl (approved):

PR #13570, opened 2018-10-10
When showing license, check plugin name is valid

mattab:


tsteur:
https://github.com/matomo-org/matomo/pull/13569 :)

PR #13573, opened 2018-10-10
Truncate referrer name & keyword in Base class so the value used there matches what is in the DB.

diosmosis:
Prevents creating new visits when referrer is the same but is cut off in the DB because it is too long.

refs https://github.com/matomo-org/matomo/pull/13469


tsteur:
LGTM if tests pass

PR #13574, opened 2018-10-10
fix domain age from archive.org

Findus23:
https://wayback.archive.org doesn't seem to exist anymore and the new website loads the content dynamically via react. But there is no need to parse the archive.org website, because they provide an API, that's perfect for this use case as it returns the closest match to a timestamp.

https://archive.org/help/wayback_api.php

### DIFF
--- a/plugins/SEO/Metric/DomainAge.php
+++ b/plugins/SEO/Metric/DomainAge.php
@@ -74,16 +74,12 @@ class DomainAge implements MetricsProvider
      */
     private function getAgeArchiveOrg($domain)
     {
-        $data = $this->getUrl('https://wayback.archive.org/web/*/' . urlencode($domain));
-        preg_match('#<a href=\"([^>]*)' . preg_quote($domain) . '/\">([^<]*)<\/a>#', $data, $p);
-        if (!empty($p[2])) {
-            $value = strtotime($p[2]);
-            if ($value === false) {
-                return 0;
-            }
-            return $value;
+        $response = $this->getUrl('https://archive.org/wayback/available?timestamp=19900101&url=' . urlencode($domain));
+        $data = json_decode($response, true);
+        if (empty($data["archived_snapshots"])) {
+            return 0;
         }
-        return 0;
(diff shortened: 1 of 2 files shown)


tsteur:
Do you know if there is a test for this feature? For some reason I'm not seeing an expected system test file for `SEO.getRank` or any other test.

Findus23:
@tsteur This is the only test of the SEO plugin:
https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/SEO/tests/Integration/SEOTest.php#L47

And only tests the rank and therefore couldn't detect this bug.

tsteur:
Be good to add a test 👍

Findus23:
@tsteur Seems like I was wrong and the test tests all four values (as they are all called `rank`)
But because of 0e727311408cef75c7ba7bcb11b4fd989d8227e3 it skipped when one of them fails and as the alexa rank always fails (https://github.com/matomo-org/matomo/issues/13427), it didn't actually test anything.

plugins/SEO/Metric/DomainAge.php
```diff
@@ -74,16 +74,12 @@ public function getMetrics($domain)
...
-        preg_match('#<a href=\"([^>]*)' . preg_quote($domain) . '/\">([^<]*)<\/a>#', $data, $p);
-        if (!empty($p[2])) {
-            $value = strtotime($p[2]);
-            if ($value === false) {
-                return 0;
-            }
-            return $value;
+        $response = $this->getUrl('https://archive.org/wayback/available?timestamp=19900101&url=' . urlencode($domain));
+        $data = json_decode($response, true);
+        if (empty($data["archived_snapshots"])) {
```
tsteur: to be safe... should this be `if (empty($data["archived_snapshots"]["closest"]["timestamp"]))` ?

Findus23: Archive.org is currently down for maintenance, so I can't test it.
I thought that would throw an error if `$data["archived_snapshots"]["closest"]` doesn't exist, but it seems like the PHP empty function takes care of that.

PR #13578, opened 2018-10-10
Detect destroyed session use

diosmosis:
& make sure session is destroyed after a day if remember me is not used.

### DIFF
--- a/core/Session.php
+++ b/core/Session.php
@@ -11,7 +11,9 @@ namespace Piwik;
 use Exception;
 use Piwik\Container\StaticContainer;
 use Piwik\Exception\MissingFilePermissionException;
+use Piwik\Plugins\Login\Login;
 use Piwik\Session\SaveHandler\DbTable;
+use Piwik\Session\SessionNamespace;
 use Zend_Session;
 
 /**
@@ -20,6 +22,7 @@ use Zend_Session;
 class Session extends Zend_Session
 {
     const SESSION_NAME = 'PIWIK_SESSID';
+    const NON_REMEMBERME_GC_SESSION = 86400;
 
     public static $sessionName = self::SESSION_NAME;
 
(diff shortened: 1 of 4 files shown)


tsteur:
Lets move this maybe into 3.7.0?

diosmosis:
@tsteur moved

diosmosis:
Verified it works.

Note: in order to test the `session.gc_maxlifetime` change, you'll have to add `@ini_set('session.gc_divisor', 1);` to make sure session GC is triggered. (I also set the lifetime to 60 so I wouldn't have to wait a day.)

core/Session.php
```diff
@@ -78,8 +81,8 @@ public static function start($options = false)
         @ini_set('session.referer_check', '');
 
         // to preserve previous behavior piwik_auth provided when it contained a token_auth, we ensure
-        // the session data won't be deleted until the cookie expires.
-        @ini_set('session.gc_maxlifetime', $config->General['login_cookie_expire']);
+        // the session data won't be deleted until the cookie expires (if form rememberme is used).
+        @ini_set('session.gc_maxlifetime', Login::isRememberMeLogin() ? $config->General['login_cookie_expire'] : self::NON_REMEMBERME_GC_SESSION);
```
tsteur: I'm not sure about this one @diosmosis I  would have expected to always have it set to login cookie expire. 

Reading in http://php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime I'm not sure I understand it. Most of the time, `Login::isRememberMeLogin()` isn't true so it would maybe destroy the session too early? Also does this setting affect other sessions  or only the current session?

Is this change really needed?

diosmosis: I was trying to make sure that the session data is removed after a reasonable time when session is not remembered, but I'm not sure there is a good way to do this. I'll just remove this code.

diosmosis:
Actually this PR doesn't even work... not sure why I thought it did. Closing.

PR #13584, opened 2018-10-11
Add more tests for format=original.

diosmosis:


sgiehl (commented):
@diosmosis any reason for adding those tests? Imho it's very hard to validate if the content of each test file is correct. If it's only for proving the format `original` works, maybe there's a better way to check without comparing any file contents?

diosmosis:
The original format is going to be used for another plugin that will unserialize after getting serialized data back over a network. Which means it can be easy to add new data to the result of an API method that cannot be unserialized.

I suppose an alternative here is to just serialize & then unserialize it. Will look into that.

diosmosis:
Updated to just unserialize the response and check it unserialized.

sgiehl (approved):
Looks much better now :ok_hand:

Issue #13413, opened 2018-09-11
When exporting data and "Show export URL" is clicked, don't reveal the full token_auth 


mattab:
The new feature "Show export URL" is very valuable in giving everyone quick access to the API and seeing how the URL is constructed, making it easy to share, etc.

However for security reasons we would not want to reveal the full token_auth on screen. 
Similarly in the Personal settings page where the `token_auth` is displayed to the user, it requires an extra click to reveal the full token. 

So the goal if this issue is to slightly change the behavior, proposal:
* When "Show export URL" is clicked, show the textarea but in the string, only show the first few characters and write `...`.
* When user clicks the field to copy/paste it, then reveal the full token_auth and full URL

follows up #11958 #12987

tsteur:
Is this really needed considering you already have to click to see it?

mattab:
Just tested and noticed that currently you need to click, and then double click to select the string. Ideally on the first click to reveal the full URL then the full URL would be selected and then user can copy. it would be more usable.

But an even more secure / usable solution could be maybe:

* The full token_auth is never displayed on screen, and instead it shows the full URL and replaces `token_auth=full_token_here` by `token_auth=start_tok***********`
* on click on the textarea, the value is copied into the clipboard
* and a feedback "Copied to your clipboard!" is displayed

Not showing the token_auth on screen would be a security improvement as the token can be easily seen/recorded by someone viewing the screen.

tsteur:
I won't be working on this, so unassigning myself. I don't agree with any of the flows. Even the original in the issue.

mattab:
is there maybe a better way not to show the token_auth on screen? anyway it's not urgent for now so removing from milestone

tsteur:
Also when you paste a URL in the address bar, it will show the end of the URL which is usally the token. Next, a more likely risk than screen recording is that users send the exported URL to other people. Ideally you neither show them the URL nor let them copy the URL. Then when we POST to export the URL, the token won't be in the URL at all and will neither appear in access logs.

sgiehl:
We meanwhile show `ENTER_YOUR_TOKEN_AUTH_HERE` instead of the real token in the generated export url. This prevents any possible token leakage. So I guess this issue can be considered as won't fix, as we don't want to show the token there at all anymore.

PR #13585, opened 2018-10-11
When exporting data and "Show export URL" is clicked, don't reveal the full token_auth until field is clicked

tsteur:
refs https://github.com/matomo-org/matomo/issues/13413

Seeing there are no UI tests for this feature so far but also don't think it needs any. Logic is simple and otherwise end up with thousands of UI tests :)


sgiehl (approved):
Still unsure if that "improvement" is good regarding usability but works as expected

PR #13594, opened 2018-10-11
Keep a copy of triggered event names for better debugging

tsteur:
Sometimes you are in a method that can be triggered through many different events and you want to know which event triggered the method or which events happened before this as the event might be triggered by another event etc. Not keeping any params re possible memory issues.

Might be even useful to log like the last X occurred events in some places on top of a backtrace or so... (be something maybe for another PR). Refs DEV-1456

### DIFF
--- a/core/EventDispatcher.php
+++ b/core/EventDispatcher.php
@@ -30,6 +30,27 @@ class EventDispatcher
     const EVENT_CALLBACK_GROUP_SECOND = 1;
     const EVENT_CALLBACK_GROUP_THIRD = 2;
 
+    private static $lastEventNames = array();
+
+    /**
+     * A list of triggered event names, useful to debug problems and to better understand how something happens
+     * @param  int $limit if a value is set, will return only the last X triggered events
+     * @return string[]
+     */
+    public static function getTriggeredEventNames($limit = 0)
+    {
+        if (empty($limit)) {
+            return self::$lastEventNames;
+        } else {
+            return array_slice(self::$lastEventNames, -1 * $limit, $limit);
+        }
+    }
+
+    private static function logLastEvent($eventName)
+    {
+        self::$lastEventNames[] = $eventName;
+    }
+
     /**
      * Array of observers (callbacks attached to events) that are not methods
      * of plugin classes.
@@ -90,6 +111,8 @@ class EventDispatcher
         if ($pending) {
             $this->pendingEvents[] = array($eventName, $params);
         }
+
+        self::logLastEvent($eventName);
 
         $manager = $this->pluginManager;
 


diosmosis:
Is this needed because we don't have access to the event name in a shared event handler? Like in this example:

```
public function getListHooksRegistered()
{
    return [
        'event1' => 'myEventHandler',
        'event2' => 'myEventHandler',
        'event3' => 'myEventHandler',
        ...
    ];
}

public function myEventHandler()
{
    $logger->info("myEventHandler() triggered by ???");
}
```
? Or are there other reasons (or maybe this isn't a reason)?

tsteur:
yes for example this.

sometimes there might be also APIs that call other APIs that call other APIs and you want to get a trace of how we ended up there to understand what happened. Eg in a `API.TagManager.addContainer` which may be triggered by a `SitesManager.addSite.end` which may be triggered by ... etc

diosmosis:
For the first case, would it be simpler to just have separate event handlers? For the other use case, would a backtrace work too? Eg `debug_backtrace(IGNORE_ARGS_OR_WHATEVER_ITS_CALLED)`.

Think this could be useful for fatal error emails, but if we save our logs (to eg, cloudwatch or on a server), you could simply look up the logs for the request in either case. Eg, we could log in EventDispatcher and the event handlers (or maybe just in EventDispatcher like `invoking event handler for event MyEvent: MyPlugin.myEventHandler`). Or maybe I'm missing another use case?

diosmosis:
Couldn't you just have a log in EventDispatcher like:

```
$logger->info("Dispatching event $eventName to handler $pluginName.$methodName");
```

? You wouldn't need to save and later process the events in memory in this case, you would just need to look at the logs. And you would be able to see what events are triggered (ie `CustomPiwikJs.trackerChanged` and what event handlers are triggered (`Heatmap.handleTrackerChanged` or whatever it's called).

tsteur:
In `debug` level that might be at some point useful too. For my case it is way too much noise and would log way too much. Also I can't log to file without also logging it to the user which would be a problem as I need it always on. 

In my case, when something specific happens, I need to log this info to a specific place, eg to `error_log`. I'm actually only interested in like the last 5 events. So if memory is a problem, we could always keep max 50 events in there or so? Looking at logs doesn't work for me, as I need this info with the actual log message in a different log file. It happens like once a week or few times a week that it gets logged and then looking through many GB of other regular logs wouldn't work for me to find the matching info.

Eg it allows me to 
`cat my.log | grep "my error message" | cut -d' ' -f1 | sort | uniq -c | sort -n`

and I can see the top reasons why some specific error or event occurred, including the last 3 or 5 events that occurred before that.

like 
```
134343 my error message, events: Foo.Bar, X.Baz, Y.Baz
43434 my error message, events: Bar.Baz, A.Baz, B.Baz
2323 my error message, events: Hello.Bar, X.Baz, B.Baz
```

diosmosis:
> For my case it is way too much noise and would log way too much. Also I can't log to file without also logging it to the user which would be a problem as I need it always on.

Could always log as info, default the logger to NullLogger, then supply a special logger through DI that sends it to error_log(). Or something along those lines.

> It happens like once a week or few times a week that it gets logged and then looking through many GB of other regular logs wouldn't work for me to find the matching info.

Perhaps it might be a good idea to look into something like cloudwatch/sumologic/etc.

To me memory usage isn't really the issue, and I can see that it could be useful, but you can get the same thing from just logging (if you can get the request ID for the request you are looking for more info for, then grep for those logs then grep for the event log message, you'd get the same info). This just seems like unneeded code.

diosmosis:
Could also wrap EventDispatcher through DI so this code isn't in core.

tsteur:
I'll just close the PR. Getting bit too complicated. I'll log the messages in my plugin without any extra info.

PR #13596, opened 2018-10-11
Make matomo.js and matomo.php the default Tracking API endpoints used

### LINKED ISSUE #12785: Make matomo.js and matomo.php the default Tracking API endpoints used
As part of our ongoing to https://github.com/matomo-org/matomo/issues/12420 renaming of everything [from Piwik to Matomo](https://matomo.org/blog/2018/01/piwik-is-now-matomo/), the next most important step is that our JS Tracking code should reference `matomo.js` and `matomo.php` instead of `piwik.js|php`. This remaining use of "Piwik" seems to be the last highly visible mention left in the product.

Suggested steps:

* The tracking code in Administration > Tracking code should reference matomo.js|php
* The API output for SitesManager.getJavascriptTag and others like SitesManager.getImageTrackingCode  should also reference matomo.js|php 
* it's very important that we keep backward compatibility, so the file "piwik.js" should still be there and be a copy of matomo.js file (we will need to update instructions on how to build the compiled matomo.js in https://github.com/matomo-org/matomo/blob/master/js/README.md#deployment
* And add a ReleaseChecklist test that the matomo.js is the same as piwik.js
* Search and check/replace all occurences of `piwik.js` in the codebase (there are many)
* Ideally we also rename the `CustomPiwikJs` plugin to `CustomMatomoJs` or `CustomTrackerJs`
* Possibly other steps I missed

tsteur:
fix https://github.com/matomo-org/matomo/issues/12785

* Only for new installations we use matomo.js|matomo.php through the API as default. For older installations we use `piwik.js|piwik.php`. This way we won't break anything when tracking code is embedded automatically (eg the case on 70k+ wordpress installations).
* The tracking code generator and "site without data" screens show the matomo endpoint though. 
* Custom Piwik JS plugin has now an alias command `./console custom-matomo-js:update`(won't update the plugin as it may break things, is complicated, etc. lets do this in Matomo 4.0 https://github.com/matomo-org/matomo/issues/13604)
* The Custom JS plugin now updates `/matomo.js` and `/piwik.js` when generating the tracker file
* The Custom JS plugin diagnostic check may check whether `matomo.js` and `piwik.js` is writable depending on when Matomo was installed
* The Matomo endpoint can be enforced in the tracking API through a parameter (not really worth mentioning in the dev changelog as it is pretty much only for internal purposes). 
* Plugins that hook to `Piwik.getJavascriptCode` could now in theory rename the tracker endpoints `matomo.js|php` to anything else but the plugin would need to ensure those files exist etc.

We now also record the version number that was used to install Matomo so we can easier make changes in the future without breaking things for existing users.


tsteur:
fyi created https://github.com/matomo-org/matomo-log-analytics/issues/224

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ This is the Developer Changelog for Matomo platform developers. All changes in o
 
 The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)** lets you see more details about any Matomo release, such as the list of new guides and FAQs, security fixes, and links to all closed issues. 
 
+## Matomo 3.7.0
```
diosmosis: Needs to be updated.

core/Db/Schema/Mysql.php
```diff
@@ -472,6 +477,28 @@ public function createAnonymousUser()
...
+    public function recordInstallVersion()
+    {
+        if (!self::getInstallVersion()) {
+            Option::set(self::OPTION_NAME_MATOMO_INSTALL_VERSION, Version::VERSION);
+        }
+    }
+
+    /**
+     * Creates an entry in the User table for the "anonymous" user.
+     */
```
diosmosis: Think this & the above docblock needs to be changed or removed.

core/SettingsPiwik.php
```diff
@@ -213,6 +213,15 @@ public static function getPiwikUrl()
...
     }
 
+    /**
+     * @see SettingsPiwik::isPiwikInstalled
+     * @return bool
+     */
+    public static function isMatomoInstalled()
+    {
+        return self::isPiwikInstalled();
+    }
```
diosmosis: Should isPiwikInstalled() be deprecated?

tests/PHPUnit/proxy/misc
```diff
@@ -0,0 +1 @@
+/home/vagrant/www/piwik/misc
```
diosmosis: This file probably shouldn't be added, just a symlink (probably should be in .gitignore).

diosmosis:
VisitorGenerator could probably use an update.

Otherwise looks good, conflicts need to be resolved though.

core/Db/Schema/Mysql.php
tsteur: 👍 will update

core/SettingsPiwik.php
tsteur: good point 👍  will do

tests/PHPUnit/proxy/misc
tsteur: yes definitely 👍  no idea how it ended up there.

tsteur: also adding it to gitignore

tsteur:
> VisitorGenerator could probably use an update.

@diosmosis  you mean to send to `matomo.php`? I'll create an issue on the repository to do this at the latest for Matomo 4.0 This way we don't need to increase the required version for now etc. and ensures the plugin continues to work even if they block matomo.php . 

@diosmosis can you maybe look into https://github.com/matomo-org/matomo-log-analytics/issues/224 ?

diosmosis:
Looking now

tsteur:
fyi created https://github.com/matomo-org/plugin-VisitorGenerator/issues/35 and will also prepare a PR for visitor generator shortly (renaming of the log files takes quite a while...)

diosmosis:
@tsteur log importer PR: https://github.com/matomo-org/matomo-log-analytics/pull/235 and submodule update PR: https://github.com/matomo-org/matomo/pull/13776

The submodule update's tests won't pass until this PR is merged.

tsteur:
UI tests should be fixed now. Merge conflicts resolved. Will probably merge soon if ok @diosmosis to avoid further conflicts

diosmosis:
no issue for me if you merge

PR #13605, opened 2018-10-14
allow admin user to copy dashboards

### LINKED ISSUE #13567: admins can't share dashboard with users
Matomo 3.6.0 
only super-users added can share dashboards with other users, the "copy dashboard to user" isn't visible in the dropdown for admins. 
According to https://matomo.org/faq/how-to/faq_25986/ admin's should be able to share dashboards also

tsteur:
fix #13567

### DIFF
--- a/plugins/Dashboard/API.php
+++ b/plugins/Dashboard/API.php
@@ -7,6 +7,7 @@
  */
 namespace Piwik\Plugins\Dashboard;
 
+use Piwik\API\Request;
 use Piwik\Piwik;
 
 /**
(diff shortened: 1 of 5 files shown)


plugins/Dashboard/API.php
```diff
@@ -112,7 +113,21 @@ public function removeDashboard($idDashboard, $login='')
...
+        $userFound = false;
+        foreach ($users as $user) {
+            if ($user['login'] === $copyToUser) {
+                $userFound = true;
+                break;
+            }
+        }
+
+        if (!$userFound) {
+            throw new \Exception(Piwik::translate('Cannot copy dashboard to user %s, user not found or user does not have access to same site.', $copyToUser));
```
sgiehl: Guess this should be moved to a translation key, or the `Piwik::translate` can be removed

tsteur: didn't create translation key similar to the end of the API method where it is also not translated as when triggered through the UI this cannot happen... will remove the piwik translate

diosmosis:
Testing locally I see the "Copy dashboard to user" modal will show every user in the system, even for an admin user. This is a security risk, no?

diosmosis:
Testing locally it seems that even for an admin user it shows the entire list of users in the "Copy dashboard to user" modal. Think this is a security risk.

tsteur:
I tested locally here and it didn't. Did you test if the users you see are in the same sites as the admin user?

diosmosis:
I had a single site every user had access to, w/o that site it works (though some users still result in the "Cannot copy dashboard to user" error; they probably don't have access to the site).

tsteur:
When every user has access to this site, and the user is an admin user, then the user can see all the other users. That's expected. I just tested here again and it works as expected. As admin users you can see all other users that have access to the same site.

say you have 
* site 1 admin access
* site 2 write access
* site 3 view access
* site 4 admin access
* site 5 admin access
* site 6 view access

Then you are allowed to see a list of all users that have access to site 1, 4 and 5.

The method `UsersManager.getUsers` is used to show the list of available users and is also used to validate in the backend.

diosmosis:
Some test failures, otherwise looks good to merge & works locally

tsteur:
Fixed the test 👍

PR #13606, opened 2018-10-14
When token_auth is re-generated, remove wording that user will be logged out

### LINKED ISSUE #13597: When token_auth is re-generated, remove wording that user will be logged out
When token_auth is re-generated, we have wording warning users that they will be logged out:

remove

But we have changed this behavior recently and so we should remove these wordings that user will be logged out.

tsteur:
fix #13597


plugins/UsersManager/lang/en.json
```diff
@@ -79,9 +79,7 @@
         "TheLoginScreen": "The login screen",
         "ThereAreCurrentlyNRegisteredUsers": "There are currently %s registered users.",
         "TokenAuth": "API Authentication Token",
-        "TokenRegenerateConfirm": "Changing the API authentication token will invalidate the user's current token. If the user is currently logged in, they may be required to log in again. Do you really want to continue?",
-        "TokenRegenerateConfirmSelf": "Changing the API authentication token will invalidate your own token. You may be required to log in again if you proceed. Do you really want to change your authentication token?",
-        "TokenRegenerateLogoutWarning": "When you re-generate your token you will be logged out and need to log in again.",
+        "TokenRegenerateConfirmSelf": "Changing the API authentication token will invalidate your own token. If the current token is in use, you need to update all API clients with the newly generated token. Do you really want to change your authentication token?",
```
diosmosis: Not sure if this changed, but do we need to change the key here since the translation content is different?

tsteur: AFAIK it is not needed anymore @sgiehl

sgiehl: Not needed. Tranisfex will automatically clear all translations of this key if the content changes,

diosmosis (approved):

PR #13607, opened 2018-10-14
When creating HTML reports in tmp/assets/ include a random string in the filename

### LINKED ISSUE #13586: When creating HTML reports in tmp/assets/ include a random string in the filename
For users using Nginx or when .htaccess is disabled, it may be possible to guess the filenames and access HTML/PDF report data. 

To prevent this issue we could for example introduce a random string in the filename, which would be removed when the file is being downloaded via the API.

This should not impact most users as they would use apache with .htaccess support, which would prevent direct file access.

tsteur:
fix #13586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ This is the Developer Changelog for Matomo platform developers. All changes in o
 
 The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)** lets you see more details about any Matomo release, such as the list of new guides and FAQs, security fixes, and links to all closed issues. 
 
+## Matomo 3.7.0
+
+### Breaking Changes
+* The output type "save on disk" in the API method `ScheduledReport.generateReport` has been replaced by the download output type.
+
 ## Matomo 3.6.1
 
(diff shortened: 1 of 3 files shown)


diosmosis:
Should this also remove the suffix from the filename? Could be possible w/ a `Content-Disposition` header I think.

tsteur:
Not sure what you mean?

diosmosis:
The issues says "which would be removed when the file is being downloaded via the API". I assumed that there needed to be some code to remove the extra string, maybe that's not needed? I haven't tested locally, will do that.

mattab:
I can't remember why we need to store reports on disk? but if we can get rid of that, it would be best :+1:

diosmosis:
> The file wouldn't be downloaded, it would be persisted on the file system and a user can access the report through the returned output path at any time...

I see so there's no API method like `downloadReport()`.

tsteur:
Done.

> I see so there's no API method like downloadReport().

Nope there isn't, there is only a direct download option for which the option "save on disk" is not needed.

tsteur:
Fixed the test 👍

PR #13608, opened 2018-10-15
Dashboard selector listing all widgets is hidden once the scroll bar is used

### LINKED ISSUE #13489: Dashboard selector listing all widgets is hidden once the scroll bar is used
To reproduce:
* Click on the dashboard selector
* Scroll down using browser scroll bar
* Got: the dashboard selector closed
* Expected: the dashboard selector stays opened

(It's expected the selector closes when one clicks outside of it, but this should exclude the scrollbar area)

tsteur:
fix https://github.com/matomo-org/matomo/issues/13489

### DIFF
--- a/plugins/CoreHome/angularjs/common/directives/focus-anywhere-but-here.js
+++ b/plugins/CoreHome/angularjs/common/directives/focus-anywhere-but-here.js
@@ -22,12 +22,16 @@
             restrict: 'A',
             link: function(scope, element, attr, ctrl) {
 
-                var ignoreOutsideElement = false;
-                var scrollTimeout = false;
+                var isMouseDown = false;
+                var hasScrolled = false;
 
                 function onClickOutsideElement (event) {
-                    if (ignoreOutsideElement) {
-                        return; // used scroll bar just before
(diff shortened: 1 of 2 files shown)


diosmosis:
Tested on windows and it looks like it solves the problem in some cases but not all. Here's an example of it not working:

https://ufile.io/ores8

Didn't look into why exactly. Can probably merged though since it's not that big a bug.

tsteur:
might just need a hire timeout maybe?

diosmosis:
Maybe we could keep the scroll timeout going if the mouse is still pressed?

diosmosis:
Think you detect if mouse is down, then in the timeout if it's down, schedule another 500ms timeout. When the mouse goes back up, the timeout gets cleared so even if the user pauses for a while while scrolling, it won't close the dropdown.

tsteur:
done

diosmosis:
Thanks, I realize it's probably not that important, but figured it was a simple change to apply.

sgiehl (approved):
Tested it in Chrome, Firefox and Opera. Seems to work as expected :+1:

PR #13616, opened 2018-10-16
Add possibility to queue tracking requests so they are sent in bulk

tsteur:


js/piwik.js
```diff
@@ -5584,6 +5590,58 @@ if (typeof window.Piwik !== 'object') {
...
+                        return;
+                    }
+
+                    this.requests.push(requestUrl);
+
+                    if (this.timeout) {
+                        clearTimeout(this.timeout);
+                        this.timeout = null;
+                    }
+                    // we always extend by another 1.75 seconds after receiving a tracking request
```
tsteur: couldn't decide between 1.5 and 2 seconds :)

tsteur: fyi https://github.com/matomo-org/matomo/pull/13444 will be useful to have to ensure the bulk request won't be too large in case tracking requests keep getting queued

js/piwik.js
```diff
@@ -5584,6 +5590,58 @@ if (typeof window.Piwik !== 'object') {
...
+                requests: [],
+                timeout: null,
+                sendRequests: function () {
+                    var requestsToTrack = this.requests;
+                    this.requests = [];
+                    if (requestsToTrack.length === 1) {
+                        sendRequest(requestsToTrack[0]);
+                    } else {
+                        sendBulkRequest(requestsToTrack);
+                    }
```
diosmosis: Does sending a single request vs bulk request w/ single entry make a big difference?

js/piwik.js
diosmosis: Could make the time configurable. At least that way it won't require a commit to core to experiment with it.

diosmosis:
Left some comments, otherwise looks good.

js/piwik.js
tsteur: shouldn't make too much of a big difference. Although... for bulk requests we currently don't fallback to GET requests and always do a POST request compared to usually GET request, and it will be wrapped with `{"requests":["?' + requests.join('","?') + '"]}` so not sure if it can be replayed easily using log analytics etc.

I would say it is better when possible to use `sendRequest` when there is only one request and keep differentiating.

js/piwik.js
tsteur: I'd prefer to not make it configurable for now to keep it simple but could add it later if needed. For now only premium features will use it anyway I think. Although could maybe also start using it for content impressions eventually. The defined interval shouldn't make too much of a difference.

js/piwik.js
diosmosis: I see 👍

tsteur:
@diosmosis I'll merge for now so I can merge it into https://github.com/matomo-org/matomo/pull/13596 and resolve merge conflicts etc. Let me know if I should change anything further.

PR #13620, opened 2018-10-16
Added site setting: "userId linked to visitorId"

MichaelRoosz:
Added a new setting "userId linked to visitorId" (default: enabled), that allows to turn off the default "userId overrides visitorId" behaviour as a per-site setting.

This is useful for example when using the third party cookie, and thus all Matomo sites use the same "global" visitorId for the same device, and some Matomo sites set a userid. In this case, once a site sets a userid, the vistiorId will change and make it impossible to follow a visitor when traveling through different matomo sites.

With this setting, this behavior can be adjusted. userId and visitiorId will no longer influence each other when the new setting is disabled.
However, different devices with the same userId will also no longer share the same visit and visitorId.


MichaelHeerklotz:
Apparently "isUserIdLinkedToVisitorId()" in the Request class will cause a database operation which is undesired when using the QueuedTracking plugin. I guess the settings needs to be made global. I will update my PR tommorrow.

MichaelHeerklotz:
Depending on when this gets pulled, the "core/Updates/3.6.1-b4.php" file needs to be adjusted to the correct version.

MichaelHeerklotz:
In matomo-php-tracker this line https://github.com/matomo-org/matomo-php-tracker/blob/master/PiwikTracker.php#L1243 will return a wrong visitorId if this new setting is deactivated, however since the JS tracker only returns the cookie value https://github.com/matomo-org/matomo/blob/3.x-dev/js/piwik.js#L5693 maybe it would be good anyway, to sync the php tracker with the js tracker and do the same (just return the cookie value)

mattab:
Thanks for the PR :+1: 
Closing in favor of https://github.com/matomo-org/matomo/pull/14360

PR #13624, opened 2018-10-17
add getGlobalVisitorId() to tracker js

MichaelRoosz:
this pr adds the function getGlobalVisitorId() to the tracker js, which returns the visitorId stored in the 3rd party cookie (_pk_uid)


MichaelHeerklotz:
It this gets accepted, I can add getGlobalVisitorId() also to matomo-php-tracker, if you are intersted.

tsteur:
@MichaelHeerklotz sorry for the late reply. Are you still needing this? As it's quite edge case with third party cookies and only works when Matomo is also installed on the same domain I'm thinking it might be better to use `tracker.getCookie('_pk_uid')` and do a custom implementation in your code?

tsteur:
@MichaelHeerklotz I'll close this PR for now as mentioned ideally you would use `tracker.getCookie('_pk_uid')`. Let me know if that won't work and I'll be happy to reopen and see what we can do. Generally `globalVisitorId` might be bit confusing and too edge case. We have now also documented how to build a custom JS tracker see https://developer.matomo.org/guides/enrich-js-tracker . Meaning this could be even solved in a plugin.

MichaelHeerklotz:
Hello @tsteur, if we want to add it, I think we should get the 3rd party / global visitor id from an api call to the Matomo installation. If this is something you would accept, I can create a pr.

tsteur:
@MichaelHeerklotz we'd suggest creating a custom plugin for this to not "bloat" the default tracker. This should be possible using https://developer.matomo.org/guides/enrich-js-tracker

eg you could do something like this in a `tracker.js` file:

```js
(function () {

    function init() {
        function enrichTracker(tracker) {
            tracker.getThirdPartyId = function (callbackToBeCalledWhenDataFetched) {
            ...
            }
        }
        if (windowAlias.Piwik.initialized) {
            var asyncTrackers = Piwik.getAsyncTrackers();
            asyncTrackers.forEach(enrichTracker);
            Piwik.on('TrackerSetup', enrichTracker);

        } else {
            Piwik.on('TrackerSetup', enrichTracker);
        }
    }

    if ('object' === typeof window.Matomo) {
        init();
    } else {
        // tracker might not be loaded yet
        if ('object' !== typeof window.matomoPluginAsyncInit) {
            window.matomoPluginAsyncInit = [];
        }

        window.matomoPluginAsyncInit.push(init);
    }

})();
```

and then it works to do eg `tracker.getThirdPartyId(callback)`

PR #13025, opened 2018-05-31
Make join table sort stable

c960657:
As discussed in #12778, the table sort in `Piwik\DataAccess\LogQueryBuilder\JoinGenerator` should be stable for more reliable testing.

In case the order of two joins is irrelevant, they are now sorted by the table alias which should be unique. This makes the sorting identical in PHP 5 and PHP 7.

PR #13634, opened 2018-10-21
Fix build and make join table sort stable by using DFS

diosmosis:
Changes:

* Fix original format tests that fail on mysqli due to it returning ints by casting Site properties to int.
* Make sure table sort in join tables is stable by using DFS instead of weighting tables. Join dependencies are parsed/assumed and dependent table/table-aliases get ordered first.
* Couple other minor test fixes.


diosmosis:
Refs #13025

sgiehl (commented):
Code changes look good to me and tests seem to pass. Would maybe be good if @tsteur would have another look

core/DataAccess/LogQueryBuilder/JoinTables.php
```diff
@@ -137,5 +132,147 @@ private function checkTableCanBeUsedForSegmentation($tableName)
...
+    private function parseDependencies(array $tables)
+    {
+        // NOTE: joins can be specified explicitly as arrays w/ 'joinOn' keys or implicitly as table names. when
+        // table names are used, the joins dependencies are assumed based on how we want to order those joins.
+        // the below table list the possible dependencies of each table, and is specifically designed to enforce
+        // the following order:
+        // log_link_visit_action, log_action, log_visit, log_conversion, log_conversion_item
+        // which means if an array is supplied where log_visit comes before log_link_visitAction, it will
+        // be moved to after it.
+        static $implicitTableDependencies = [
```
tsteur: is that needed to be static?

tsteur:
It's a complex PR/topic and don't understand all the code / hard to review so focused on looking at tests. Ran the system tests for most premium features and they seem to pass. Also checked custom reports generates similar queries that still work (some SQLs changed but are still valid). As nothing seems to break I reckon it's good to merge  (in 3.8).

core/DataAccess/LogQueryBuilder/JoinTables.php
diosmosis: Not really, it's just a constant. I can remove the static modifier.

core/Site.php
```diff
@@ -82,6 +82,13 @@ public function __construct($idsite)
         self::setSiteFromArray($this->id, $site);
 
         $this->site = $site;
+
+        // for serialized format to be predictable across php/mysql/pdo/mysqli versions, make sure the int props stay ints
+        foreach ($this->site as $key => $value) {
+            if (is_numeric($value)) {
```
tsteur: what about floats etc? It'd be actually better to really specify the properties that should be converted (like we do in other plugins) and not something generic

tsteur:
fyi we will need to test this with all premium features again, and especially custom reports to make sure the generated queries are still correct.

tsteur:
ran the custom reports tests  @diosmosis and looks all good 👍

core/Site.php
diosmosis: changed

Issue #13418, opened 2018-09-12
Make Matomo PHP 7.3 compatible


Findus23:
In 3 months PHP 7.3 will be released, so let's start making Matomo compatible!

I tried setting up a new Matomo instance using PHP 7.3 RC2 (so all bugs found now will probably also be in the final release) and most things are working fine. These errors occurred:

-------
```
ErrorException: strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior
#17 vendor/tedivm/jshrink/src/JShrink/Minifier.php(196): handleError
```
[Stack Trace](https://sentry.lw1.at/share/issue/d2252951764d49d5a009409bdd400202/)
from https://github.com/tedious/JShrink
 - [x] fixed
----------
```
2018/09/12 20:46:17 [error] 8919#8919: *137700 FastCGI sent in stderr: 
"PHP message: PHP Warning:  
"continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?
in /var/www/matomo-beta/vendor/tedivm/jshrink/src/JShrink/Minifier.php on line 243"
```
also from https://github.com/tedious/JShrink
 - [x] fixed
--------
```
WARNING [2018-09-12 19:04:26] 20837  /var/www/matomo-beta/plugins/LanguagesManager/Commands/CreatePull.php(217): Warning - "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? - Matomo 3.6.0 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)
```
https://github.com/matomo-org/matomo/blob/b8e4d27f2151103df826e5dd643b433cdde05aaf/plugins/LanguagesManager/Commands/CreatePull.php#L214-L217
 - [x] fixed in #13635
-------

```
ErrorException: compact(): Undefined variable: idGoal
#15 plugins/API/API.php(401): handleError
```
[Stack Trace](https://sentry.lw1.at/share/issue/021ddebfb19649d984b9e288f73f39bc/)
 - [x] fixed by #13593
--------

```
ErrorException: Undefined index: never
#15 core/View/HtmlReportEmailHeaderView.php(41): handleError
#14 core/View/HtmlReportEmailHeaderView.php(41): __construct
#13 core/ReportRenderer/Html.php(84): renderFrontPage
```
[Stack Trace](https://sentry.lw1.at/share/issue/92d331022a934ad688be6f05af5eb364/)
[...]

PR #13635, opened 2018-10-21
fix switch continue in CreatePull

Findus23:
Removes warning in PHP 7.3 (#13418)


sgiehl (approved):
don't have a 7.3 running, but changes look good

PR #13639, opened 2018-10-22
update JShrink

Findus23:
should hopefully make Matomo compatible with 7.3 and maybe even fix other issues without creating new ones.
see #13418


tsteur:
Be good to run the Jshrink tests with PHP 5.5 maybe. Looks like they run tests only for PHP 7+. Could obviously also easily regress later. Not sure if UI tests use the minified version or not (I think they do)

diosmosis:
The UI tests should use minified assets, but they won't test premium features, so maybe we want to hold off on #13418 until 3.9.0? I moved this in to 3.8.1 since #13418 was scheduled for it, but it might be better to put this in a beta or test on staging.

diosmosis:
BTW @Findus23 can you rebase/merge w/ 3.x-dev? I can't seem to get composer to create the composer.lock file correctly, it always includes two JSShrinks.

Findus23:
See #14011

PR #13646, opened 2018-10-26
Added API endpoint to return the php version info

Toflar:
Hey everybody,

I'm using the `getMatomoVersion` endpoint to see if my Matomo setups are all up-to-date. I would like to also know if the PHP version they're all running on are maintained. Unfortunately there's no endpoint yet for that.

I'm very new to the code base of Matomo but actually very experienced in php and testing. Also followed your setup guide for the tests but when trying to run `./console tests:run unit` it tried to connect to the db (which is very strange to me for unit tests) so I didn't explore more on it.

Feel free to guide me to the things I need to do to have this tested, I'll happily update the PR if you like this addition 😄

### DIFF
--- a/plugins/API/API.php
+++ b/plugins/API/API.php
@@ -90,6 +90,23 @@ class API extends \Piwik\Plugin\API
     }
 
     /**
+     * Get PHP version
+     * @return array
+     */
+    public function getPhpVersion()
+    {
+        Piwik::checkUserHasSomeViewAccess();
+        return array(
+            'version' => PHP_VERSION,
+            'major' => PHP_MAJOR_VERSION,
+            'minor' => PHP_MINOR_VERSION,
+            'release' => PHP_RELEASE_VERSION,
+            'versionId' => PHP_VERSION_ID,
+            'extra' => PHP_EXTRA_VERSION,
+        );
+    }
+
+    /**
      * Get Matomo version
      * @return string
      * @deprecated
--- a/plugins/API/API.php
+++ b/plugins/API/API.php
@@ -95,7 +95,7 @@ class API extends \Piwik\Plugin\API
      */
     public function getPhpVersion()
     {
-        Piwik::checkUserHasSomeViewAccess();
+        Piwik::checkUserHasSuperUserAccess();
         return array(
             'version' => PHP_VERSION,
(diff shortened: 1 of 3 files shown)


fdellwing:
Hey @Toflar 

please see https://travis-ci.org/matomo-org/matomo/builds/446616865 for your tests results (not all fails are your fault, just take a look if anything looks like it could be your fault ;))

Toflar:
Oh nice, UI tests using screenshots, that's cool!
Anyway, I would've liked to add unit tests but the result is dynamic (depending on the running php version) so I couldn't just add the XML, I would've had to add quite some logic to it and so I tried to see how `getMatomoVersion()` is tested and it's not at all so :-(

tsteur:
FYI: I quickly checked... the DB connection should be only needed to bootstrap the test system but the unit tests itself shouldn't connect to the DB.

plugins/API/API.php
```diff
@@ -89,6 +89,23 @@ public function getMatomoVersion()
         return Version::VERSION;
     }
 
+    /**
+     * Get PHP version
+     * @return array
+     */
+    public function getPhpVersion()
+    {
+        Piwik::checkUserHasSomeViewAccess();
```
tsteur: This would need to be `Piwik::checkUserHasSuperUserAccess();`.

Toflar: Can you elaborate why? I mean, why do I need super user access to get the php version but not to get the matomo version?

tsteur: We're currently only exposing this information to super users. Partially for security reasons. If you needed it for View users, you could develop a simple plugin that defines this API using eg 

```
./console generate:plugin --name="MyPlugin"
./console generate:api
```

You could even put the plugin on the Marketplace https://developer.matomo.org/guides/distributing-your-plugin

Findus23: I agree with @tsteur.
There are many ways a user with view access could find out the Matomo version without the api, but there should be non to find out the php version.

tsteur:
Looks good for me. Any thoughts @mattab ?

mattab:
LGTM :+1:

tsteur:
Cheers @Toflar

PR #13658, opened 2018-10-31
Let plugins change content of JS Tracking Code Missing Email

tsteur:
Through DI...


diosmosis:
👍 makes sense. other option is to make `Mail.send` event pass `$mail` by reference.

PR #13660, opened 2018-11-01
Show the CORS settings in the UI when enable_general_settings_admin=0

mattab:
The INI setting description does not mean to include CORS:
```ini

; By setting this option to 0, the following settings will be hidden and disabled from being set in the UI:
; - "Archiving Settings"
; - "Update settings"
; - "Email server settings"
enable_general_settings_admin = 1
```


sgiehl (approved):
Code looks fine. Just to be sure, this will also activate the possibility to change the trusted hostnames

tsteur (changes requested):
trusted hostnames should not be writable in this case

PR #13664, opened 2018-11-02
Autodiscover tables for segments w/ complex segment expressions

diosmosis:
Currently if a segment has a custom `$sqlSegment` that includes functions/operations (like `UNIX_TIMESTAMP(log_visit.visit_first_action_time) - log_visit.days_since_first_visit * 86400`), the joins for segment expressions won't be discovered. So if we try to use this segment w/ log_conversion, it will skip the log_visit join and fail.

This PR fixes that by parsing out table/column names from the segment expression.

Also includes a new public method to return the ArchiveWriter instance in ArchiveProcessor so it's easier to create new ArchiveProcessor instances from within Archivers.


tsteur:
Can you add some tests for this @diosmosis ?

diosmosis:
👍 will do

diosmosis:
Added tests.

core/Segment/SegmentExpression.php
```diff
@@ -301,11 +301,24 @@ protected function getSqlMatchFromDefinition($def, &$availableTables)
...
         return array($sqlExpression, $value);
     }
 
+    /**
+     * @param string $field
+     * @return string[]
+     */
+    private function parseColumnsFromSqlExpr($field)
+    {
+        preg_match_all('/\b([a-zA-Z0-9_`]+\.[a-zA-Z0-9_`]+)\b/', $field, $matches);
```
tsteur: @diosmosis not sure if on purpose... when field is eg "`log_action`.`type`" then

`preg_match_all('/\b([a-zA-Z0-9_`]+\.[a-zA-Z0-9_`]+)\b/', '`log_action`.`type`', $matches);`

returns this:
image
also see https://3v4l.org/0v32G 
not sure if the field can be like this.

diosmosis: Updated to better handle backticks (it's not super intelligent, but should work given this is for plugin devs).

tsteur:
Tested and worked. Only thing I would have done be to maybe write some tests for `parseColumnsFromSqlExpr($field)` to make sure it works in various cases. I think the other tests don't quite cover this as extensively? not sure... maybe they do indirectly (I think so).

I used xdebug to see what the inputs are, how they are generated etc. to see about possible inputs etc... seeing maybe just a few examples in tests might help understand what is being down there but all good if not adding more tests and if it is indirectly covered by other tests.

diosmosis:
>  Only thing I would have done be to maybe write some tests for parseColumnsFromSqlExpr($field) to make sure it works in various cases.

All it should do is look for `table.column` expressions in strings. Ie, it doesn't parse the SQL expression into pieces at all, so the only real variants are having one vs. many & backticks vs. no backticks. Ie, the presence of operators, function calls shouldn't have an effect.

tsteur:
oki 👍

diosmosis:
Added a test as you suggested. Will merge if the build passes.

Issue #13662, opened 2018-11-01
Add ability to access XHR object in the callback of trackPageView


adityasharma7:
First of all, Kudos to Matomo community for the amazing analytics tools.

I am using Matomo for tracking page view. I have a requirement where on completion of XHR request of track page view I need to perform some operation which involves accessing xhr ( `XmlHttpRequest` ) object. Basically, I need some of the data extracted by Piwik to be sent to another server too which is not a Matomo server. I figured out that with the `trackPageView()` method we can pass `customTitle, customData, callback`. I was able to pass a callback function as a parameter but I am not able to access xhr in the callback.
I did some exploration towards the flow of Piwik.js.
In `sendXmlHttpRequest()` function,
`if(this.readyState===4&&(typeof callback ==='function')){callback()}`
can be changed to
`if(this.readyState===4&&(typeof callback ==='function')){callback(xhr)}`
When the library is providing a callback for xhr, it should also provide access to the xhr object.

tsteur:
That might be possible. In the future this behaviour may change though as we maybe group couple requests into one bulk request for example. I guess it depends why and what you need from the XHR object?

adityasharma7:
@tsteur Thanks for your response. I have added a comment for the same on the pr #13665 
Thanks!

Findus23:
This was fixed a long time ago with #13855

PR #13451, opened 2018-09-19
Add option to opt in to use send beacon

tsteur:
fixes https://github.com/matomo-org/matomo/issues/6641

At some point, maybe with Matomo 4.0 we might even want to make this the default? I can't think of too many downsides of using it except for when you want to make use of the `callback` feature in some tracking methods.

It also doesn't fallback to GET if the POST doesn't work but in most cases the POST should work just fine.

PR #13665, opened 2018-11-02
Add ability to access XHR object in the callback of trackPageView

adityasharma7:
implements  #13662


adityasharma7:
@tsteur Thanks for your quick response.
As stated on the issue we have another non Matomo system and when we track any page with Matomo some of the data extracted by Piwik.js will also be sent to a different system that may be able to give a quick glimpse about the tracking.
I explored about the getImage() method, it overrides onload so xhr object is there in that case. Though, thanks for pointing that.

tsteur:
Cheers, so it is mainly the request url you are interested in? The PR looks good and can be merged, it's just to understand the use case so I can check if there's maybe a better way to do it that ensures better backwards compatibility in the future etc.

tsteur:
Could we maybe pass `request` then? This way it would also work in `getImage()` method for example.

adityasharma7:
Indeed, `request` seems a much better solution. 

Though I found another way which I am uncertain of a right solution. There is a method `getRequest()` available with `piwikTracker` which can be called in the callback. I can get all the parameters except the page title. I digged further, as it is a generic method the request string is initialized with specific parameters for various applications. 

Is it the right way? If no I can use the above method till the changes you suggested are available with the release. if yes, we won't need a change.

Looking forward to Matomo 3.8 with https://github.com/matomo-org/matomo/pull/13451/files#diff-1279d666063b65e6d6777f902d11574fR3766 :)

Thanks for sharing that.

One more thing came to my mind what if we make another tracker and use it to send the request to a non matomo server, is it a valid solution?

tsteur:
Calling getRequest would work to get all the parameters (it wouldn't include the `idgoal=X` though when you track a goal). I just realize it seems there is no `getDocumentTitle` available which be useful to add. Also realising it would be actually great to pass the `tracker` instance in the callback as first argument... then optionally we could also pass `request` variable.

adityasharma7:
> I just realize it seems there is no getDocumentTitle available which be useful to add. 
+1. I will see if I can contribute.

> Also realising it would be actually great to pass the tracker instance in the callback as first argument... then optionally we could also pass request variable.
Sounds promising. I will update the code soon

tsteur:
@adityasharma7 I needed the feature now myself so added it here: https://github.com/matomo-org/matomo/pull/13855 

Cheers for your PR and the suggestion 👍  and also for letting me better understand why people use the callback 👍

adityasharma7:
@tsteur That's great!!!!
Thanks for putting your time into it!
Cheers!

PR #13667, opened 2018-11-04
Support tideways xhprof profiler

tsteur:
XHProf isn't working so well anymore for PHP7.X so be good to support tideways profiler https://github.com/tideways/php-xhprof-extension

### DIFF
--- a/core/Profiler.php
+++ b/core/Profiler.php
@@ -209,12 +209,19 @@ class Profiler
             return;
         }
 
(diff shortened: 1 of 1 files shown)


diosmosis:
Merged, any thought about integrating some kind of viewer? doesn't seem like tideways has one (found https://github.com/docteurklein/xhprof-analyze but haven't used it).

tsteur:
I viewed it using xhprof that we have already embedded. Worked nicely.

PR #13472, opened 2018-09-24
Lock down accounts by IP after N failed attemps at logging in

tsteur:
As discussed in #2888 a quite simple implementation that should do the job. Heaps more could be done as mentioned in the comments but might not be needed.

It took me couple hours to understand all the authentication/login flows as there are lots of different components now such as Auth, SessionAuth, SessionInitializer, Access, etc. and it wasn't clear where to put things to make it work universally while still being clear about when it happens etc.

Also had to add couple of events eg for failed logins. The most important part is really to register a failed login attempt when there is one (we often authenticate eg API requests etc even though the user didn't, or we always authenticate first against session then regular auth adapter etc which makes things complicated). And it is important to check whether IP is blocked when trying to attempt to log in. All those parts should be covered.

When an IP is blocked, I did not want to block any already logged in users in the UI. This could be annoying when a company shares the same IP and it would otherwise even block the super user.

The superuser has a diagnostic page to unblock all users and a console command. Can also document in a FAQ how to do it through the SQL database directly.

For system settings and diagnostics page screens see screenshot tests.

Functionality wise you specify two values: X amount of login retries allowed in Y amount of minutes. If you have more than X amount of logins within Y amount of minutes, you are blocked and cannot log in. Even when successful token provided for example. You can only log in again, when some time has passed and you had less than X amount of login retries within the last Y amount of minutes. There is no specific lock out time that you can configure so if people lock themselves out they can use UI again after eg 30 min or 60min. By default we allow 20 logins within 60 minutes. When sharing an IP within an office those 20 logins could be reached within an hour but should be maybe fine. Hard to find the right balance here.

You can also whitelist and blacklist specific IPs.

fixes #2888 

I'll try to write some more tests for the different login failures once the PR is reviewed.

PR #13670, opened 2018-11-05
Implement Two Factor Authentication

tsteur:
fixes #13325 in a new plugin TwoFactorAuth

Also provides a new class `PasswordVerifier` which requires a user to first confirm their password in a new screen before continuing with a specific action (not available in a dialog currently).


tsteur:
Todo: We need to update either this PR or https://github.com/matomo-org/matomo/pull/13472 depending on which one gets merged first to apply brute force / lock down changes to #13670

plugins/Login/Controller.php
```diff
@@ -148,6 +159,51 @@ protected function configureView($view)
...
+            $nonce = Common::getRequestVar('nonce', null, 'string', $_POST);
+            if (!Nonce::verifyNonce($nonceKey, $nonce)) {
+                $messageNoAccess = $this->getMessageExceptionNoAccess();
+            } elseif ($this->verifyPasswordCorrect()) {
+                $this->passwordVerify->setPasswordVerifiedCorrectly();
+                return;
+            } else {
+                $messageNoAccess = Piwik::translate('Login_WrongPasswordEntered');
+            }
+        }
```
diosmosis: If this method is for verifying the two factor auth code, I think it might be confusing to use the word password for both the user's password and the two factor code. Some services use "two-factor passcode", maybe we could use that?

plugins/TwoFactorAuth/Commands/Disable2FAForUser.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->checkAllRequiredOptionsAreNotEmpty($input);
+        $login = $input->getOption('login');
+
+        StaticContainer::get(TwoFactorAuthentication::class)->disable2FAforUser($login);
+        $message = sprintf('<info>Disabled two-factor authentication for user: %s</info>', $login);
+        $output->writeln($message);
```
diosmosis: Don't think it's super important, but should this go through `Request::processRequest()`?

plugins/TwoFactorAuth/Validator.php
```diff
@@ -0,0 +1,84 @@
...
+            return false;
+        }
+
+        return !Piwik::isUserIsAnonymous();
+    }
+
+    public function checkCanUseTwoFa()
+    {
+        Piwik::checkUserIsNotAnonymous();
+    }
```
diosmosis: Should this also have the `SettingsPiwik::isPiwikInstalled()` check?

plugins/TwoFactorAuth/templates/userSettings.twig
```diff
@@ -0,0 +1,41 @@
...
+            {{ 'TwoFactorAuth_TwoFactorAuthenticationRequired'|translate }}
+            <br />
+            <br />
+            <a class="btn btn-link enable2FaLink" href="{{ linkTo({'module': 'TwoFactorAuth', 'action': 'setupTwoFactorAuth'}) }}">{{ 'TwoFactorAuth_ConfigureDifferentDevice'|translate }}</a>
+        {% else %}
+            <a class="btn btn-link enable2FaLink" href="{{ linkTo({'module': 'TwoFactorAuth', 'action': 'setupTwoFactorAuth'}) }}">{{ 'TwoFactorAuth_ConfigureDifferentDevice'|translate }}</a>
+            <a href="{{ linkTo({'module': 'TwoFactorAuth', 'action': 'disableTwoFactorAuth', 'disableNonce': disableNonce}) }}" style="display:none;" id="disable2fa">disable2fa</a>
+            <input type="button"
+                   class="btn btn-link disable2FaLink"
+                   onclick="piwikHelper.modalConfirm('#confirmDisable2FA', {yes: function () { broadcast.propagateNewPage('module=TwoFactorAuth&action=disableTwoFactorAuth&disableNonce={{ disableNonce|e('url') }}'); }})"
```
diosmosis: Should code this be in a JS file instead of the onclick attribute?

plugins/UsersManager/lang/en.json
```diff
@@ -1,5 +1,10 @@
 {
     "UsersManager": {
+        "2FA": "2FA",
+        "UsesTwoFactorAuthentication": "Uses two-factor authentication",
+        "TwoFactorAuthentication": "Two-factor authentication",
+        "ResetTwoFactorAuthentication": "Reset two-factor authentication",
+        "ResetTwoFactorAuthenticationInfo": "If the user cannot log in anymore because of lost recovery codes and the authentication device you can reset two-factor authentication so the user can log in again.",
```
diosmosis: Translation tweak: `"If the user can no longer log in due to lost recovery codes or a lost authentication device, you can reset two-factor authentication for the user so they can log in again."`

diosmosis:
From testing locally:
* in the setupTwoFactorAuth page, if I copy text manually, it doesn't enable the 'next' button, which confused me a bit.
* not sure how important it is, but the 2 factor auth code input seems to save values:

image

seems a bit odd to see past values here.
* the mailto for the "Ask super user to reset your authentication code " link isn't set properly for non-superusers. likely due to the change in UsersManager::enrichUser (the email is removed if the user isn't the current user)

diosmosis:
Looked through the code, seems ok but there's a lot to review, so could have missed something. @sgiehl might want to review as well

plugins/TwoFactorAuth/Commands/Disable2FAForUser.php
tsteur: will change it. 👍

plugins/TwoFactorAuth/Validator.php
tsteur: yep, makes sense 👍  will update.

plugins/TwoFactorAuth/templates/userSettings.twig
tsteur: I'll move the code within `onclick` into a JS file but will likely leave the onclick itself re simplicity...

plugins/UsersManager/lang/en.json
tsteur: 👍  will update

tsteur:
@diosmosis fyi applying all the changes you suggested.

> the mailto for the "Ask super user to reset your authentication code " link isn't set properly for non-superusers. likely due to the change in UsersManager::enrichUser (the email is removed if the user isn't the current user)

this is a bug and regressed... will also look into how to fix it...

> in the setupTwoFactorAuth page, if I copy text manually, it doesn't enable the 'next' button, which confused me a bit.

not sure if we can detect whether all the codes have been copied manually? sounds bit tricky maybe but will check if possible

tsteur:
> Could just allow users to click next w/o copying/etc. Would only be important if lots of people tried to copy manually.

platforms like github force you to press one of the buttons basically to increase likelihood that they have created a backup of the recovery codes. I just checked and it does show a notification

> Please backup your recovery codes using one of the above methods before continuing the two-factor authentication setup.

So I reckon it should be good for now... also fixed couple screenshot tests but a few more need to be updated after next run. Also added support for activity log...

@sgiehl let me know whether you want to have a look or whether it can be merged.

plugins/TwoFactorAuth/templates/userSettings.twig
diosmosis: 👍

sgiehl:
I will have a look now...

core/Db/Schema/Mysql.php
```diff
@@ -52,6 +53,14 @@ public function getTablesCreateSql()
                           ) ENGINE=$engine DEFAULT CHARSET=utf8
             ",
 
+            'twofactor_recovery_code'    => "CREATE TABLE {$prefixTables}twofactor_recovery_code (
+                          idrecoverycode BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,
+                          login VARCHAR(100) NOT NULL,
+                          recovery_code VARCHAR(40) NOT NULL,
```
sgiehl: That one differs to the update script, it only creates a `VARCHAR(20)`

sgiehl (changes requested):
Some Feedback:
- Left some comments

- When 2FA is already setup the UI imho does not make it clear enough that `Setup a different device` will also remove the currently configured 2FA and also the recovery codes. Github for example displays this warning when changing the 2FA device:
image

- Migration from GoogleAuthenticator plugin seems to work fine. Shall I release a new version that marks the plugin not compatible with Matomo 3.8?

plugins/UsersManager/lang/en.json
```diff
@@ -1,5 +1,10 @@
 {
     "UsersManager": {
+        "2FA": "2FA",
```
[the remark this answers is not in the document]
tsteur: I left it there separately in case TwoFactorAuth plugin is disabled. I don't think we load the translation then? We're not checking in the UI whether 2FA plugin is disabled in UsersManager as it's edge case and bit tricky. This way it still shows correct translation there.

core/Db/Schema/Mysql.php
tsteur: 👍  will fix

tsteur:
@sgiehl now showing a warning re changing the device. Also fixed the update schema 

Yes sounds good to mark Google authenticator as invalid with 3.8 👍

PR #13675, opened 2018-11-06
Always fail during tracking when authentication is required but invalid token is used

### LINKED ISSUE #13471: Tracking API: when an invalid token_auth is specified, should the request be entirely dropped, or tracked anyway?
A token_auth can be invalid when, for example:
* The user has re-generated a new token_auth in the personal settings, and forgot to update the token in the SDK
* The wrong token is used
* The user whose token was used has been deleted
* An attacker is trying to guess valid tokens

When a token_auth is invalid, some API features (which are usually essential for SDK users who can't use the JS tracker) won't work well:

* Setting custom IP  address
* Setting custom date and time
* Overriding the default geolocation

We have two choices when it comes with dealing with these requests that have an invalid token_auth:
1. track them as if there was no token specified (possibly tracking a wrong IP address, or a wrong custom date & time) - **this is current behavior**
2. drop them entirely

-> What do you think?

Personally I'm not sure what is the best solution. If we decide to go with 2) we should make sure that, the requests are *not* dropped, when the token_auth was invalid AND there was no parameter in the request that need token_auth, ie. no`cip`, `cdt`, `country`, `city`, `region`.... (these requests with and without a valid token, would have the exact same behavior).

tsteur:
fix https://github.com/matomo-org/matomo/issues/13471

Don't think a  test is needed for the location ones.

### DIFF
--- a/core/Tracker/Request.php
+++ b/core/Tracker/Request.php
@@ -862,7 +862,7 @@ class Request
 
         if (!$this->isAuthenticated()) {
             Common::printDebug("WARN: Tracker API 'cip' was used with invalid token_auth");
-            return IP::getIpFromHeader();
+            throw new InvalidRequestParameterException("Tracker API 'cip' was used, requires valid token_auth");
         }
 
         return $cip;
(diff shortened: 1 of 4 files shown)


sgiehl (approved):
LGTM. rebased the branch to resolve the conflicts. Can be merged if tests passes

Issue #10914, opened 2016-11-26
Read Environment variables into Matomo config 


pierreozoux:
Modern web application read their config from env variables.

https://12factor.net/config

it would be a really nice to have :)

tsteur:
This would be useful for passwords etc. For other config there is some "danger" that config is not the same on all servers which causes Piwik to not work properly etc but for passwords it would be useful for sure

SonOfBytes:
Yes 12 factor support (environment var injection) for all the variables that happen at config would be really useful.  That would allow us  to run dev/staging piwik containers on the same hosts pointing to separate DBs etc.

webwurst:
For configuration `env` variables are great, but it is discouraged to use them for secrets: https://github.com/docker/docker/pull/9176#issuecomment-99542089

SonOfBytes:
Secrets are not a clear definition, because the domain of trust needs to be known.  Often secrets contained to the environment is good enough so trying to protect processes from each other is overkill.  Other times not leaking to the underlying system is critical.  The solution and the controls it offers needs to be assessed with the risk and against relevant attack trees.

So for example using something like credstash in the container runtime to inject sensitive vars is good enough for many use cases as it is separating the management of the app from the management of the environment it is running in.

butzopower:
This would specifically be useful for deploying Piwik on cloud-based platforms with ephemeral filesystems that don't persist file changes across deploys / restarts, such as Cloud Foundry or Heroku.

vfbsilva:
Any advance regarding this?

mattab:
No

mattab:
Hi @pierreozoux @vfbsilva @SonOfBytes 
We are considering implementing this feature in a plugin. 
Could you maybe confirm if reading ENV variables was for database information (host, name, password,...) only if if you had other use cases of ENV variables for Matomo?

bbodenmiller:
I think it could be useful for pretty much all settings. Take Grafana for instnace... all settings can be set by env variable: http://docs.grafana.org/installation/configuration/#using-environment-variables

tsteur:
FYI: I've implemented to read database config from environment variables optionally during installation see https://github.com/matomo-org/matomo/pull/13676

To read any other variables during runtime basically I suggest someone could work on a plugin. I might quickly have a look if that plugin can be easily developed.

pierreozoux:
Wow :) amazing ;)

Https://12factor.net ftw!

mattab:
[...]

PR #13676, opened 2018-11-07
During installation allow to prefill DB form from environment variables

tsteur:
refs https://github.com/matomo-org/matomo/issues/10914

@mattab as discussed pre-filling DB values from environment optionally. Supported are:
* MATOMO_DATABASE_HOST
* MATOMO_DATABASE_ADAPTER
* MATOMO_DATABASE_TABLES_PREFIX
* MATOMO_DATABASE_USERNAME
* MATOMO_DATABASE_PASSWORD
* MATOMO_DATABASE_DBNAME

and as fallback if that is not defined (we could also remove that fallback)
* DATABASE_HOST
* DATABASE_ADAPTER
* DATABASE_TABLES_PREFIX
* DATABASE_USERNAME
* DATABASE_PASSWORD
* DATABASE_DBNAME

The structure being taken from the global ini as in `MATOMO_$CATEGORY_$SETTINGNAME`.

I'm not going to implement that it overrides the config and that Matomo could be basically even used without config etc. Some settings will need to be writable I suppose and won't be possible to be set through environment variables. I wouldn't want it in core actually, and to overwrite / configure any environment setting I suggest this will be implemented in a separate plugin which can be put on the marketplace. For performance etc I wouldn't want to check environment variables for heaps of settings etc.

### DIFF
--- a/plugins/Installation/FormDatabaseSetup.php
+++ b/plugins/Installation/FormDatabaseSetup.php
@@ -80,12 +80,26 @@ class FormDatabaseSetup extends QuickForm2
         $defaultDatabaseType = Config::getInstance()->database['type'];
         $this->addElement( 'hidden', 'type')->setLabel('Database engine');
 
+
+        $defaults = array(
+            'host'          => '127.0.0.1',
+            'type'          => $defaultDatabaseType,
+            'tables_prefix' => 'matomo_',
+        );
+
+        $defaultsEnvironment = array('host', 'adapter', 'tables_prefix', 'username', 'password', 'dbname');
+        foreach ($defaultsEnvironment as $name) {
+            $envName = 'DATABASE_' . strtoupper($name); // fyi getenv is case insensitive
+            $envNameMatomo = 'MATOMO_' . $envName;
+            if (getenv($envNameMatomo)) {
+                $defaults[$name] = getenv($envNameMatomo);
+            } elseif (getenv($envName)) {
+                $defaults[$name] = getenv($envName);
+            }
+        }
+
         // default values
-        $this->addDataSource(new HTML_QuickForm2_DataSource_Array(array(
-                                                                       'host'          => '127.0.0.1',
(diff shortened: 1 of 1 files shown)


tsteur:
As mentioned in the issue developed https://github.com/matomo-org/plugin-EnvironmentVariables but haven't tested it. If someone can confirm this works, we can publish it on the marketplace.

baywet:
I can test it, but not immediately, what do I need to do to test it? install matomo from this specific branch? Could you provide me a documentation on how to get the equivalent of the matomo.zip (release) from this branch please?

tsteur:
Easiest be to download the regular matomo.zip, install, and then replace the content of the file `plugins/Installation/FormDatabaseSetup.php` with this content: https://raw.githubusercontent.com/matomo-org/matomo/04381b4bec7a9afa7f923bd3dc1769e4e4768aa0/plugins/Installation/FormDatabaseSetup.php

baywet:
thanks! I'll try that as soon as I have a minute

baywet:
Sorry for the long delay on this one @tsteur 
So this works, but in my understanding the only thing it does is reading the environment variables at setup time to pre-fill the form and then stores it into the config file.
It'd be better if matomo could read the configuration values from environment variables at runtime all the time. This is standard practice for all templated services approach (like arm templates, docker composed services….)

tsteur:
see https://github.com/matomo-org/matomo/pull/13676#issuecomment-436464925 @baywet

baywet:
Thanks for the fast reply! And sorry about that, I missed that comment today…
I works like a charm. Now as you might know I'm working on the Azure offering. Part of how this works is why a "deployment" repository which is used for the file copy of what needs to be hosted.
Right now I'm maintaining that repo manually but I'll have to automate that part now.
Is the EnvironmentVairables plugin going to be part of the default plugins (the matomo-latest.zip) ? and enabled by default (global.ini.php) ? or do I need to script that part as well?

PR #13159, opened 2018-07-11
Fix missing 3rd party cookie when using POST

MichaelRoosz:
Added "xhr.withCredentials = true;" so that the third party cookie is included in POST requests.

This fixes visitor duplication when POST gets used.

PR #13677, opened 2018-11-07
Improved CORSHandler

MichaelRoosz:
When using the 3rd party cookie, and tracking to piwik with AJAX, the request will (must) include the 3rd party cookie (see https://github.com/matomo-org/matomo/pull/13159 ).

It is already possible to set "cors_domains[] = *" in config.ini.php to allow this.
However "The value of * is special in that it does not allow requests to supply credentials, meaning it does not allow HTTP authentication, client-side SSL certificates, or cookies to be sent in the cross-domain request." (see https://en.wikipedia.org/wiki/Cross-origin_resource_sharing ).

Thus this change avoids the "*" value and also adds the 'Vary: Origin' header.

### DIFF
--- a/core/API/CORSHandler.php
+++ b/core/API/CORSHandler.php
@@ -25,8 +25,22 @@ class CORSHandler
 
     public function handle()
     {
+        if (empty($this->domains)) {
+            return;
+        }
+        
+        Common::sendHeader('Vary: Origin');
+        
         // allow Piwik to serve data to all domains
         if (in_array("*", $this->domains)) {
+            
+            Common::sendHeader('Access-Control-Allow-Credentials: true');
+            
+            if (!empty($_SERVER['HTTP_ORIGIN'])) {
+                Common::sendHeader('Access-Control-Allow-Origin: ' . $_SERVER['HTTP_ORIGIN']);
+                return;
+            }
+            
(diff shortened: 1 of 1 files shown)


core/API/CORSHandler.php
```diff
@@ -25,8 +25,22 @@ public function __construct()
 
     public function handle()
     {
+        if (empty($this->domains)) {
+            return;
+        }
+        
+        Common::sendHeader('Vary: Origin');
```
diosmosis: Is there a reason for this new header? From what I understand about the header, it exists to tell http caches how to identify a resource when deciding which cached object to use. It doesn't seem related to CORS.

MichaelHeerklotz: > 6.4 Implementation Considerations
This section is non-normative.

Resources that wish to enable themselves to be shared with multiple Origins but do not respond uniformly with "*" must in practice generate the Access-Control-Allow-Origin header dynamically in response to every request they wish to allow. As a consequence, authors of such resources should send a Vary: Origin HTTP header or provide other appropriate control directives to prevent caching of such responses, which may be inaccurate if re-used across-origins.

https://www.w3.org/TR/cors/#resource-implementation

diosmosis: I see, 👍

Issue #2888, opened 2012-01-26
Lock down accounts by IP after N failed attemps at logging


mattab:
Our security policy aims to make security a principal design behind Piwik. One aspect that bugs me currently is that good old brute force attacks could be vector of penetration in Piwik (if eg. attacker knows the login).

We should provide a core mechanism that would lock out, for  30min for example, a user after N failed attemps. Settings could be changed by the Super User and feature would be enabled by default, lock 30 min out after 5 failed attempts.

Implementation proposal:
- Record, using Piwik_SetOption, count of lockdown for each IP that fails to enter valid login / pwd combination
- After N failures, lock IP down and refuse authentication (even if the combination is actually valid!).
- Document as FAQ, linked from UI, the sql to delete all locked out IPs in case the SU was actually locked out and can't wait.

jkraemer:
Actually just adding the logging (including timestamp and remote ip) might solve the whole issue for a lot of people. Fail2ban is made exactly for this purpose, there's no need to re-invent the wheel here imho.

tsteur:
I think this is quite an important issue for security of Piwik. It is not too difficult to brute force installations otherwise.

> Lockdown IP after X attempts. Some kind of blacklist management will be need.

After eg 100 attempts within 12 hours I would lock down IP (we'd need config for this if all users come from same IP or reuse `trust_cookies` setting which is used in Intranets and depending on this disable it etc). 

> On bad credential wait X seconds before showing login screen (minimize web brute force).

After say 5 wrong login attempts for same user, I would make login slower (also on API level but won't be trivial) each time. Eg in the beginning wait 1 second, next try wait 3 seconds, next try wait 6 seconds ... This needs to be implemented wisely since one could "shut down" a Piwik under circumstances by doing wrong login requests on purpose etc. Won't be trivial to implement I reckon but I'm sure for such things there are good solutions available on the internet

mattab:
@tsteur moved to Short term and added `Major` tag - how much effort do you think it would take to seriously mitigate this security risk?

gaumondp:
In TYPO3 we got Backend access with a sleep(5) on bad login for 15 years. There was a discussion about brute force that could be interesting to read : 

https://forum.typo3.org/index.php/t/196184/

tsteur:
From https://forum.typo3.org/index.php/t/196184/ :

> You will get a warning email (if you set up an address in the install tool) after four unsuccessful attempts anyway.

[...]

PR #13683, opened 2018-11-08
When changing password or email address, require to type old password 

### LINKED ISSUE #2932: When changing password or email address, require to type old password
If you leave Piwik open and logged in, anyone accessing the computer could change the email address or the password. Changing email address would allow to "reset" the password.

Therefore, as an extra security measure, we should require the old password to change the password or the email address. 

When changing other settings inputting the password wouldn't be necessary.

tsteur:
When changing email or password, the password will be required for the current user.

refs https://github.com/matomo-org/plugin-LoginLdap/pull/174
fix https://github.com/matomo-org/matomo/issues/2932

Todo: We need to update either this PR or #13472 depending on which one gets merged first to apply brute force / lock down changes


core/Piwik.php
```diff
@@ -321,7 +321,7 @@ public static function isUserIsAnonymous()
     {
         $currentUserLogin = Piwik::getCurrentUserLogin();
         $isSuperUser = self::hasUserSuperUserAccess();
-        return !$isSuperUser && $currentUserLogin == 'anonymous';
+        return !$isSuperUser && $currentUserLogin && strtolower($currentUserLogin) == 'anonymous';
```
tsteur: not really needed but might avoid some random case where anonymous user was not detected correctly in case someone passes eg `anOnymous` to an API

plugins/UsersManager/API.php
```diff
@@ -795,26 +798,35 @@ public function regenerateTokenAuth($userLogin)
...
+                               $_isPasswordHashed = false, $passwordConfirmation = false)
     {
+        $requirePasswordConfirmation = self::$UPDATE_USER_REQUIRE_PASSWORD_CONFIRMATION;
+        self::$UPDATE_USER_REQUIRE_PASSWORD_CONFIRMATION = true;
+
         Piwik::checkUserHasSuperUserAccessOrIsTheUser($userLogin);
         $this->checkUserIsNotAnonymous($userLogin);
         $this->checkUserExists($userLogin);
 
+        if (!Url::isValidHost()) {
```
tsteur: Moved this from UsersManager controller into here. Might eventually make sense to apply this to more login/user related API methods or even simply in general?

plugins/UsersManager/API.php
```diff
@@ -795,26 +798,35 @@ public function regenerateTokenAuth($userLogin)
...
+     * If password or email changes, it is required to also specify the password of the current user needs to be specified
+     * to confirm this change.
      *
      * @see addUser() for all the parameters
      */
     public function updateUser($userLogin, $password = false, $email = false, $alias = false,
-                               $_isPasswordHashed = false)
+                               $_isPasswordHashed = false, $passwordConfirmation = false)
     {
+        $requirePasswordConfirmation = self::$UPDATE_USER_REQUIRE_PASSWORD_CONFIRMATION;
```
tsteur: Couldn't find a better way to make `updateUser` verify the password but sometimes not. Eg passwordResetter wouldn't know the current password of the user as it is stored hashed... 

things like only checking it when `Request::isRootRequestApiRequest()` would not work (cause some plugins call it through API) and neither `$requirePasswordConfirmation = Request::isCurrentApiRequestTheRootApiRequest()` cause eg because people could avoid it by using bulk requests etc.

plugins/UsersManager/API.php
```diff
@@ -860,6 +884,18 @@ public function updateUser($userLogin, $password = false, $email = false, $alias
         Piwik::postEvent('UsersManager.updateUser.end', array($userLogin, $passwordHasBeenUpdated, $email, $password, $alias));
     }
 
+    private function verifyPasswordCorrect($userLogin, $password)
+    {
+        /** @var \Piwik\Auth $authAdapter */
+        $authAdapter = StaticContainer::get('Piwik\Auth');
```
tsteur: I'm hoping this works with other auth plugins

plugins/UsersManager/UserUpdater.php
```diff
@@ -0,0 +1,44 @@
...
+     * @param bool $password
+     * @param bool $email
+     * @param bool $alias
+     * @param bool $_isPasswordHashed
+     * @throws \Exception
+     */
+    public function updateUserWithoutCurrentPassword($userLogin, $password = false, $email = false, $alias = false,
+                                                     $_isPasswordHashed = false)
+    {
+        API::$UPDATE_USER_REQUIRE_PASSWORD_CONFIRMATION = false;
```
tsteur: Added this class so each time this is needed it doesn't need to be re-developed and no mistakes can be made. (eg forgetting to reset `API::$UPDATE_USER_REQUIRE_PASSWORD_CONFIRMATION = true;`)

core/Piwik.php
diosmosis: 👍

plugins/UsersManager/API.php
diosmosis: Could you just use the `UserUpdater` that you introduced? For PasswordResetter I assume it would be ok, since it doesn't use `Request::processRequest()`.

plugins/UsersManager/API.php
diosmosis: Will the Host header always be sent for API requests? If someone curls this API method they would have to send the Host header now, right?

plugins/UsersManager/API.php
diosmosis: I haven't looked at your other PR, but will this be an issue for the new two factor auth plugin?

plugins/UsersManager/API.php
diosmosis: Oh I see below that class just issues an API request. Could that class instead just use the model & set the password directly? I assume all it would need to do is hash the password and call Model::updateUser().

plugins/UsersManager/angularjs/user-edit-form/user-edit-form.component.js
```diff
@@ -142,15 +143,17 @@
...
                 vm.isSavingUserInfo = false;
+                vm.passwordConfirmation = false;
                 throw e;
             }).then(function () {
                 vm.isSavingUserInfo = false;
-                vm.isPasswordChanged = false;
+                vm.passwordConfirmation = false;
                 vm.isUserModified = true;
 
                 showUserSavedNotification();
```
diosmosis: Haven't tested yet so not sure if this is an issue or not, but should superusers have to confirm the user's password when editing a user? I guess this would only be a problem for them since they're the only ones who can edit other users' info.

core/Piwik.php
diosmosis: Found one more place that might need to use Piwik::isUserIsAnonymous():
* LastSeenTimeLogger::logCurrentUserLastSeenTime()

There some other places that use `'anonymous'` directly but don't do a strtolower, not sure if any of them need to change to ensure consistency.

diosmosis:
Noticed one thing while testing locally: whenever the password confirm comes up, I would immediately enter my password (which would work since it is focused), then press enter. But that doesn't submit the modal, so I'd have to click Yes explicitly. Might be good to add this behavior, was annoying to me at least.

core/Piwik.php
tsteur: will update 👍  I think looked over all/most of them and updated where needed.

plugins/UsersManager/API.php
tsteur: I thought about it as well and then figured it may be beneficial to still use the API just to have all the permission checks running again, the code is covered by tests, deletes tracker cache after updating user etc. Otherwise might be too easy to regress something

plugins/UsersManager/API.php
tsteur: just double checked, when no host header is set, it will be considered valid. see https://github.com/matomo-org/matomo/blob/3.8.0-b2/core/Url.php#L210-L215

tsteur: I'll move it back to the UI controller again though I think. seems not needed or much useful here. (not even sure why it is in the controller there)

plugins/UsersManager/API.php
tsteur: I don't think it is. It's only needed for login initially and the password is only a double check here. We will need to update the brute force prevention PR though (lock down IP #2888 ).

tsteur:
@diosmosis I'm resolving merge conflicts and fixing tests... is this otherwise good to merge?

tsteur:
> Noticed one thing while testing locally: whenever the password confirm comes up, I would immediately enter my password (which would work since it is focused), then press enter. But that doesn't submit the modal, so I'd have to click Yes explicitly. Might be good to add this behavior, was annoying to me at least.

I'll quickly look into it but have a feeling it might not be so easy

tsteur:
added the "enter" functionality @diosmosis

diosmosis:
@tsteur 👍 looks good

PR #13686, opened 2018-11-08
remove special handling of browsers with DNT enabled by default

### LINKED ISSUE #3531: New DoNotTrack setting to enable 100% DNT support including for IE browsers
This is a privacy concern and a step backwards for both Piwik and DoNotTrack in general. 

Instead, allow the user to choose at install or configuration time whether or not to honor the setting, with a clear statement of why each way, such as the following example text:
Respect DoNotTrack Preference?
It is encouraged to respect user privacy choices by respecting the DoNotTrack browser privacy setting. Please be aware, though, that IE10 has it enabled by default and some traffic may not be recorded.
How do you want to handle this?

[X] Respect DoNotTrack in all situations (recommended)
[ ] Respect DoNotTrack for all browsers except IE10 (for more accurate stats)
[ ] Do not respect DoNotTrack at all (not recommended)

Also, DoNotTrack should be able to be set per-site, rather than globally. 
Keywords: DoNotTrack

### LINKED ISSUE #9987: Do not ignore the "do not track"-option
The "do not track” option of the browser should not be ignored regardless of the browser used, its version or its default settings. This means especially IE10 and IE11.

It is more important to protect users against unwanted/uncontrollable tracking than to enable the collectors to get user data.

Data collection and its analysis must be based on an explicit agreement between all affected parties which means, any behavior of the sytems must be clearly declared and all available functionality must work as expected.

Kind regards,
Jonas Lanz

Findus23:
fixes #9987
also fixes #3531
Now that we advertise that Matomo respects DNT, we should do this as precisely as possible.

Of course this makes Matomo less accurate (as now IE11 users that never enabled DNT, but had it enabled by default are not tracked anymore), but I think those browsers are becoming less relevant in the future and removing the special handling is easier than explaining the technical details to all Matomo users.

A move verbose compromise solution would be having an option or plugin that restores this behavior.


sgiehl (approved):
LGTM

sgiehl:
We maybe should mention that in a changelog

mattab:
Suggested steps:

* Mention in developer changelog the change
* When DNT is activated, in the browser report footer, show a message explaining that `Did you know? Internet Explorer  11 and newer are mostly not tracked because they have DoNotTrack activated by default and Matomo <a href=privacy-settings>is configured to respect DoNotTrack</a>`
* Update the DNT inline help in the Privacy page and explain that some browsers will mostly not be tracked as they enable DoNotTrack by default and list them eg. "IE11 and newer"

tsteur:
FYI It shouldn't be mentioned in the developer changelog I would say. It's not related to the platform.

tsteur:
Also IE usage is nowadays at around 3% and maybe doesn't need to be mentioned in the browser report? And maybe only in the help text instead of footer to keep UI nice and clean and not clutter it with too unimportant things? Probably way more users are excluded with ad blockers etc

sgiehl:
> Also IE usage is nowadays at around 3%

That strongly depends on the website. Company intranets for example might have almost 100% IE in some cases. And especially those customers might need to change the Matomo config, as before the DNT setting didn't have any effect

fdellwing:
> That strongly depends on the website. Company intranets for example might have almost 100% IE in some cases. And especially those customers might need to change the Matomo config, as before the DNT setting didn't have any effect

We provide business software for hospitals and have >66% IE11 users.

tsteur:
I suppose best be to disable DNT then. For sure we shouldn't show the footer message to all users, only when IE 11 used to make a significant amount of traffic. Maybe someone could also develop a simple plugin to keep ignoring DNT for IE11 but not others

Findus23:
@tsteur I think adding a plugin is the best way to handle the cases where the feature is needed (this way it is expected and users know the implications)

Would it work to post an event like 
```
Piwik::postEvent('PrivacyManager.handleDNT', array($???));
```
here where plugins would get the user agent and detected DNT and then could return a boolean if the user should be tracked or not. (not sure if that is a bit too specific and if there is a more general way to make a plugin possible)
https://github.com/matomo-org/matomo/pull/13686/files#diff-698d1d7fe9c92e8a27eda0e8c4a88dedR75

tsteur:
Yes such an event would work. Eg `postEvent('PrivacyManager.shouldIgnoreDnt', array(&$shouldIgnore = false));`

sgiehl:
refs #3531

Findus23:
I'm afraid, we might have a problem:
https://developer.apple.com/documentation/safari_release_notes/safari_12_1_release_notes
> Removed support for the expired Do Not Track standard to prevent potential use as a fingerprinting variable.

Can someone on iOS 12.2 or the next Mac OS beta please test, if this means that Apple stops sending the header at all (which would make their reason useless as they are now more identifiable than before) or if they just always send DNT:1 without a setting to change it.

The latter would mean quite some issues as we couldn't merge this PR.

tsteur:
Quickly checked browser stack but there iOS 12.1 is running

btw here a bit more on the background: https://www.gizmodo.com.au/2019/02/apple-is-removing-do-not-track-from-safari/

Findus23:
I have now installed the 12.2 beta and indeed the option is gone.
But unlike I feared they are not sending the header by default, but just removed the header and the possibility to change it.
So this doesn't really have an influence on this PR.

Findus23:
In theory this PR is ready, but I totally fail to get the plugin event to work.

No matter what I change the `handleDNTHeader()` function is never executed. It seems like `getPluginsLoadedAndActivated()` doesn't include the plugin even though it is enabled.

Is there anything special a plugin has to do to work during tracking?

```php
<?php
/**
 * Piwik - free/libre analytics platform
 *
 * @link http://piwik.org
 * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
 */

namespace Piwik\Plugins\IgnoreDNTByDefault;

use Piwik\Common;
use Piwik\Tracker\Request;

class IgnoreDNTByDefault extends \Piwik\Plugin
{
    public function registerEvents() {
        return array(
            'PrivacyManager.shouldIgnoreDnt' => 'handleDNTHeader'
        );
    }

    public function handleDNTHeader(&$shouldIgnore) {
        Common::printDebug($shouldIgnore);
        $shouldIgnore = $this->isUserAgentWithDoNotTrackAlwaysEnabled();
    }

    public function isUserAgentWithDoNotTrackAlwaysEnabled() {
        $request = new Request($_REQUEST);
        $userAgent = $request->getUserAgent();
        $browsersWithDnt = $this->getBrowsersWithDNTAlwaysEnabled();
        foreach ($browsersWithDnt as $userAgentBrowserFragment) {
            if (stripos($userAgent, $userAgentBrowserFragment) !== false) {
                return true;
            }
        }
        return false;
    }

    /**
     * Some browsers have DNT enabled by default. For those we will ignore DNT and always track those users.
     *
     * @return array
     */
    protected function getBrowsersWithDNTAlwaysEnabled() {
        return array(
            // IE
            'MSIE',
            'Trident',
            // Maxthon
            'Maxthon',

            // Epiphany - https://github.com/matomo-org/matomo/issues/8682
            'Epiphany',
        );
    }
}
```

tsteur:
You need to add a method 

```
public function isTrackerPlugin(){
return true;
}
```

It can't detect it is a tracker plugin cause the event it is listening to doesn't start with `Tracker` or `Tracking`. During tracking for performance and security reasons etc we only load needed plugins.

Findus23:
@tsteur Many thanks for the info, it works perfectly now.

Findus23:
Totally missed this was merged.

Did someone have time to test the plugin (https://github.com/Findus23/plugin-IgnoreDNTEnabledByDefault)? @fdellwing ?

If so, I'd like to publish it before the release.

PR #13687, opened 2018-11-08
Auto zoom real time and visitor map

### LINKED ISSUE #4301: Visitor Profile - Zoom map as near as possible
Could be useful to use the map zoom to cover all locations as near as possible to see as many details as possible.

e.g. show only Germany if all visits are from this country.

Maybe it could be even possible to make the country names clickable and on click on the country name it displays/zooms the specific country?

tsteur:
Tested it locally and worked. 

Real-time map: Zooming into a country when all visits come from same country. Zooming into a continent when all visits come from same continent.

Visitor map: Zooming into a country when all visits come from same iso code... 

Not sure actually how often this happens in real life.

fix #4301

### DIFF
--- a/plugins/UserCountryMap/javascripts/realtime-map.js
+++ b/plugins/UserCountryMap/javascripts/realtime-map.js
@@ -96,6 +96,7 @@
                 colorMode = 'default',
                 currentMap = 'world',
                 yesterday = false,
+                userHasZoomed = false,
                 colorManager = piwik.ColorManager,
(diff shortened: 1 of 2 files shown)


fdellwing:
> Not sure actually how often this happens in real life.

We provide industry software, it happens every day for us ;)

diosmosis:
tested locally, works well

fdellwing:
Well I would have not merged it until #13691 is resolved, but ok. I guess there will be only a few questions why people are seeing an empty map.

tsteur:
We could otherwise disable temporarily the zoom into country and only zoom into continent if that helps until the other issue is fixed...

PR #13688, opened 2018-11-08
Do not specify a category delimiter by default

### LINKED ISSUE #2956: Actions>Page Titles believes there is a subpage if page <title> includes a forwards slash
Currently, if the html page &lt;title> is "Test / Hello", Piwik will report in Actions > Page Titles the directory "Test", clickable to open the sub page "Hello".

This is confusing for users, because often, slash characters in the page title, does not imply a notion of hierarchy.

However, [as we document it on the JS tracking doc](http://piwik.org/docs/javascript-tracking/#toc-customize-the-page-name-displayed-in-piwik), it is possible to set the document title to a given value and use this technique to classify page titles in a given hierarchy, eg. "Article / Sport / XYZ".

Because by default Page titles do not contain hierarchy, I propose that:
- when setDocumentTitle is called, we do split the page title based on the slash character
- In the default Javascript tracking code, when setDocumentTitle is not used, we do NOT split the page titles and never create the sub category.
  - Implementation wise, we could eg. have a different named argument when setdocumentTitle is called ("title") VS when the default &lt;title> value is used (parameter currently named "action_name").

This was [reported in this post](http://forum.piwik.org/read.php?3,85953), #1532, #3364, and I often got frustrated myself with this problem.

tsteur:
Doesn't require a note in the developer changelog but we should ideally mention this in the changelog as it is changing the behaviour for people that actually want to use it.

Tested it locally and worked.

fix #2956


sgiehl:
Guess it makes sense to set it empty by default. Regarding mentioning it in a changelog: Might it make sense to trigger a notification within Matomo after updating to explain the possible change to the super user once?

tsteur:
I reckon it's not needed as long as it is mentioned clearly in the regular changelog. I suppose not too many people are too keen on how this feature worked previously and might maybe rather confuse people?

sgiehl:
Guess you're right. The page title report might not be one of the frequently used reports at all

sgiehl (approved):

diosmosis:
Looks like this change causes issues in displaying the "Page Name not defined" text, not sure why on first glance.

tsteur:
pushed a fix that hopefully resolves the issue. I'll have a look at the failing tests and update them later.

tsteur:
Things should work now. There were various bugs around the delimiter in the code and the code is quite hard to understand / messy. Also added to some tests the delimiter with `/` to make sure this doesn't regress.

plugins/Actions/ArchivingHelper.php
```diff
@@ -443,12 +443,20 @@ public static function getActionExplodedNames($name, $type, $urlPrefix = null)
 
         $name = str_replace("\n", "", $name);
 
+        if ($type == Action::TYPE_PAGE_TITLE && self::$actionTitleCategoryDelimiter === '') {
+            if (empty($name)) {
+                $name = self::getUnknownActionName($type);
+            }
+            return array(trim($name));
```
diosmosis: Should we do the trim() before checking for an `empty($name)`?

tests/PHPUnit/System/LabelFilterTest.php
```diff
@@ -29,6 +29,11 @@ public function testApi($api, $params)
         $this->runApiTests($api, $params);
     }
 
+    public static function setUpBeforeClass()
+    {
+        parent::setUpBeforeClass(); // TODO: Change the autogenerated stub
+    }
```
diosmosis: Not super important to me, but don't need this method.

tests/PHPUnit/System/OneVisitorTwoVisitsWithCookieSupportTest.php
```diff
@@ -46,6 +46,18 @@ public static function getOutputPrefix()
...
+    {
+        return array(
+            'Piwik\Config' => \DI\decorate(function ($previous) {
+                $general = $previous->General;
+                $general['action_title_category_delimiter'] = "/";
+                $previous->General = $general;
+                return $previous;
+            }),
+        );
+    }
```
diosmosis: Should we just change the tests w/ the new config or is there value in keeping the old behavior in all tests?

diosmosis: Actually, it looks like only a couple of the tests have this, so I guess most use the new default?

tsteur: Most use default empty, but wanted to make sure some tests still use this behaviour to test it as well.

tests/PHPUnit/System/LabelFilterTest.php
tsteur: will remove it 👍

plugins/Actions/ArchivingHelper.php
tsteur: done, also realised `empty` but not be best if the name of a title is `0` (which might not happen but who knows)

tests/PHPUnit/System/OneVisitorTwoVisitsWithCookieSupportTest.php
diosmosis: 👍

Issue #955, opened 2009-08-26
Evolution graphs: buttons to switch the display of historical data by day, week, month or year  


mattab:
Currently there is a widget that shows visits per month if you select month, or per week if Week is selected as a date range. 

Allow users to display new visits evolution graph:
- visits by hour (when period is day) (eg. requested in #7498)
- visits by day in the week (when period is week) 
-  visits by day in the month (when period is month)
- visit by day in the year (when period is year) (does that fit in the small widget?)
- when the graph data is exported, it should show the same data as displayed in the graph (whether showing daily or monthly metrics) 

UI: not sure, how this would be best done. But a great looking solution seen in other tools is to show little buttons [ day | week | month | year ] eg. on the top right of the graph.

mattab:
In this [forum post](http://forum.piwik.org/read.php?2,126167) I think for the first time someone requested to have this similar feature but for Email reports Graphs.

dandv:
Looks like a significant number of issues have been marked as duplicates of this one.

Maybe this suggests that the UI of the timeframe selection isn't the clearest possible?

mattab:
@dandv indeed, we should and will fix it, not sure when it will be possible as our resources are limited.

mattab:
Here is example of this feature in a Google Product:
![pasted_image_at_2015_10_05_12_53_pm_1024](https://cloud.githubusercontent.com/assets/466765/10296561/2b438736-6c26-11e5-8fc6-205b9cbb6588.png)

hpvd:
there is a small but important difference: google talks about "daily" and piwik talks about a "day"+ we select a date - so the user has completely different expectations...

for piwik:
selecting a day and showing data for only this day should be really consistent in large graph, in small graph previews and in all tables

doing it any other way is a **huge problem** (as also reported in #8917 and in several other ticket):
the user select something and got the results for something else

on other places e.g. in visitor log the user gets what he has selected (the place I'm often working with)
So how should one notice and understand this inconsistency? (sorry for these hard words)

tsteur:
> visits by hour (when period is day) (eg. requested in #7498)

I don't think this is possible. We don't have hourly data for all graphs.

PR #13689, opened 2018-11-09
Add possibility to switch period in evolution graph

tsteur:
Fix https://github.com/matomo-org/matomo/issues/955

re-used the calendar icon. same period will be applied in export dialog. won't work for range dates just like the limit selector.


core/ViewDataTable/Config.php
```diff
@@ -515,6 +527,12 @@ public function __construct()
             Metrics::getDefaultProcessedMetrics()
         );
 
+        $periodValidator = new PeriodValidator();
+        $this->selectable_periods = $periodValidator->getPeriodsAllowedForUI();
+        $this->selectable_periods = array_diff($this->selectable_periods, array('range'));
```
diosmosis: Should this filtering also be done before rendering? For example, if someone sets the periods to `['day', 'week', 'garbage', 'range']`, should we remove the values that aren't allowed?

tsteur: I would have said it's not needed maybe cause a plugin might support it but not actually implement it fully in Matomo but that shouldn't be an issue I suppose. I think the feature usually won't be used so wouldn't do too much about it.

PR #13692, opened 2018-11-10
Do not remove viewdatatable params in test env during tracking or event...

diosmosis:
… handlers for Tracker.getDatabaseConfig will not run.

@tsteur FYI, will merge if tests pass.


tsteur:
👍

PR #13693, opened 2018-11-10
Fix wrong Site class is used which may not be available in build

tsteur:
See https://github.com/matomo-org/matomo/commit/2d2abcc9576079a8496ad8bd4a038ba07cf1fb2b#r31248004


sgiehl (approved):

Jon-IB:
Thanks, all.

PR #13694, opened 2018-11-11
Fix build

diosmosis:
Notes:
* Made sure `CorePluginsAdmin.getSystemSettings` is not tested when `'all'` API test is run.
* Verified API changes are correct.


diosmosis:
@tsteur fyi, merging this one to fix the build

PR #13696, opened 2018-11-11
domain name with hyphen was separated wrong in title

### LINKED ISSUE #13649: Bad position of the date in the title
If the title of the website contains a `-` (hyphen), the date is positioned after it.

For example: `my-domain` is change in `my- 2018-10-26 -domain`

selection_013

mrothauer:
fixed #13649

### DIFF
--- a/plugins/CoreHome/angularjs/common/services/piwik.js
+++ b/plugins/CoreHome/angularjs/common/services/piwik.js
@@ -68,10 +68,10 @@
 
             // Cache server-rendered page title
             originalTitle = originalTitle || document.title;
-            var titleParts = originalTitle.split('-');
-            var dateString = ' ' + piwikPeriods.parse(period, date).getPrettyString() + ' ';
+            var titleParts = originalTitle.split('- ');
+            var dateString = piwikPeriods.parse(period, date).getPrettyString() + ' ';
             titleParts.splice(1, 0, dateString);
-            document.title = titleParts.join('-');
+            document.title = titleParts.join('- ');
         }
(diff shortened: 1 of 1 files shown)


plugins/CoreHome/angularjs/common/services/piwik.js
```diff
@@ -68,10 +68,10 @@
 
             // Cache server-rendered page title
             originalTitle = originalTitle || document.title;
-            var titleParts = originalTitle.split('-');
-            var dateString = ' ' + piwikPeriods.parse(period, date).getPrettyString() + ' ';
+            var titleParts = originalTitle.split('- ');
```
sgiehl: This won't work for page titles like `Brand - Shop`. Not much into that topic atm, but can't we use the globally available `piwik.siteName` to ensure the site name isn't splitted up?

mrothauer: I'm completely new to matomo and just tried to fix the wrong title for my own domain rothauer-it.com (which was separated into) "rothauer- Von 2018-11-11 -it.com - Webanalytik-Berichte - Matomo"
My commit fixes that issue.
I have no idea what values the originalTitle can hold besides domain names, so please go ahead and change my PR.

sgiehl:
@mrothauer  Thanks for your approach. Had a quick look and opened a new PR to fix it. Would you maybe test if it solves your problem as well?
See #13697

mrothauer:
yes, it solves my problem as well. thanks!

PR #12222, opened 2017-10-24
Correctly set date in page title when period changed via period selector

spacenate:
Fixes #12162

Hello! In this change:
- Added a new protected method `getLongPrettyDate` to `Piwik\Plugin\Controller`.
- Added a new public method `getDateString` to `Piwik\Plugins\CoreHome\Controller`.
- Added a new XHR request to the calendar.js module in CoreHome, to get a new "long pretty date string" when the user has updated the date or range.

<strike>In calendar.js, the logic to fetch new string/update page title has been appended to `propagateNewUrlParams`, but in retrospect this would be better in its own function, perhaps still invoked from `propagateNewUrlParams`?</strike> (done)

In `Piwik\Plugin\Controller`, the logic inside `getLongPrettyDate` is currently is copy/pasted from `setGeneralVariablesView`. I tried to tease apart the duplicated logic, but it started to become a time sink so I wanted to seek feedback on this approach before spending further time de-duplicating this logic.

Your thoughts?

PR #13697, opened 2018-11-11
Update date in page title correctly

### LINKED ISSUE #13649: Bad position of the date in the title

sgiehl:
replaces #13696 
fixed #13649

### DIFF
--- a/plugins/CoreHome/angularjs/common/services/piwik.js
+++ b/plugins/CoreHome/angularjs/common/services/piwik.js
@@ -68,10 +68,11 @@
 
             // Cache server-rendered page title
             originalTitle = originalTitle || document.title;
-            var titleParts = originalTitle.split('-');
-            var dateString = ' ' + piwikPeriods.parse(period, date).getPrettyString() + ' ';
-            titleParts.splice(1, 0, dateString);
-            document.title = titleParts.join('-');
+
+            if (0 === originalTitle.indexOf(piwik.siteName)) {
+                var dateString = ' - ' + piwikPeriods.parse(period, date).getPrettyString() + ' ';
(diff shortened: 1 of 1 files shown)


plugins/CoreHome/angularjs/common/services/piwik.js
```diff
@@ -68,10 +68,11 @@
...
             originalTitle = originalTitle || document.title;
-            var titleParts = originalTitle.split('-');
-            var dateString = ' ' + piwikPeriods.parse(period, date).getPrettyString() + ' ';
-            titleParts.splice(1, 0, dateString);
-            document.title = titleParts.join('-');
+
+            if (0 === originalTitle.indexOf(piwik.siteName)) {
+                var dateString = ' - ' + piwikPeriods.parse(period, date).getPrettyString() + ' ';
+                document.title = piwik.siteName + dateString + originalTitle.substr(piwik.siteName.length);
+            }
```
diosmosis: Are there any cases where the date should be added and the document title does not include the site name? If there are, then this won't add the dates to those titles.

sgiehl: Wasn't 100% sure if that might be the case somewhere. That's the reason for the if.
But the target was to put the date after the website name again. See https://github.com/matomo-org/matomo/pull/12222/files#diff-1eb3ce37602c6f5a47792ccdf0bce2ff

mrothauer: your solution solves my problem, thank you. see #13696

PR #13714, opened 2018-11-14
Use angular to sanitize plugin description.

diosmosis:


plugins/CorePluginsAdmin/templates/macros.twig
```diff
@@ -141,7 +141,7 @@
                         </div>
                         <div class="plugin-desc-text">
 
-                            {{ plugin.info.description|raw|nl2br }}
+                            <div ng-bind-html="{{ plugin.info.description|raw|nl2br|json_encode|e('html_attr') }}"></div>
```
tsteur: It seems the only plugin that uses html is custom variables? I'd be tempted to simply not allow HTML in the plugin description.

diosmosis: Don't have an opinion personally.

tsteur: I'll remove the HTML from custom variables and no longer allow HTML there.

plugins/CorePluginsAdmin/templates/macros.twig
```diff
@@ -150,7 +150,7 @@
                                 'https://matomo.org', 'https://www.matomo.org', 'https://matomo.org/', 'https://www.matomo.org/'
                             ] %}
                                 <span class="plugin-homepage">
-                            <a target="_blank" rel="noreferrer noopener" href="{{ plugin.info.homepage }}">({{ 'CorePluginsAdmin_PluginHomepage'|translate|replace({' ': '&nbsp;'})|raw }})</a>
+                            <a target="_blank" rel="noreferrer noopener" href="{{ plugin.info.homepage|e('url') }}">({{ 'CorePluginsAdmin_PluginHomepage'|translate|replace({' ': '&nbsp;'})|raw }})</a>
```
tsteur: `e('url')` cannot be used on a full link, but only on a component. It breaks the URL...

I'll replace it with `html_attr`

tsteur:
@diosmosis see last commit, I fixed the homepage URL link and no longer allow HTML in the description.

PR #13715, opened 2018-11-14
Fix lbraces in widget names before setting widget titles.

diosmosis:


sgiehl (approved):
Should work as expected, but actually I'm not sure if that might cause any double encoding in some cases.

diosmosis:
Since I changed it from `.text(...)` to `.html(escape(...))`, it should be ok?

sgiehl:
Guess should be fine

PR #13727, opened 2018-11-19
It is not possible to record more than one conversion per second for the same visit

### LINKED ISSUE #9916: It is not possible to record more than one conversion per second for the same visit
This happens in v2.15.

If a page sends two conversion calls that are processed in the same second, only the first one is saved to the database.  This prevents from recording several conversions that happen in the same page at the same time and you don't want to (or you can't) record those as several items for the same conversion.

Currently this is due to the primary key in the conversions table.

tsteur:
fix https://github.com/matomo-org/matomo/issues/9916

we randomize the first 2 numbers increasing the chances to record multiple goal conversions per second per visit.

### DIFF
--- a/core/Tracker/GoalManager.php
+++ b/core/Tracker/GoalManager.php
@@ -691,14 +691,9 @@ class GoalManager
     
     private function makeRandomMySqlUnsignedInt($length)
     {
-        // mysql int unsgined max value is 4294967295 so we cannot start with a 5 or higher
-        $randomInt = Common::getRandomString(1, '1234');
-        if ($length >= 2) {
-            $randomInt .= Common::getRandomString(1, '012');
-        }
-        if ($length > 2) {
-            $randomInt .= Common::getRandomString($length - 2, '0123456789');
-        }
+        // mysql int unsgined max value is 4294967295 so we want to allow max 39999...
+        $randomInt = Common::getRandomString(1, '123');
+        $randomInt .= Common::getRandomString($length - 1, '0123456789');
         return $randomInt;
     }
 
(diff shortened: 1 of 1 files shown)


sgiehl (changes requested):

core/Tracker/GoalManager.php
```diff
@@ -682,6 +689,12 @@ protected function recordStandardGoals(VisitProperties $visitProperties, Request
         }
     }
 
+    private function makeRandomMySqlUnsignedInt($length)
+    {
+        // mysql int unsgined max value is 4294967295 so we cannot start with a 5 or higher
+        return Common::getRandomString(1, '1234') . Common::getRandomString($length - 1, '0123456789');
```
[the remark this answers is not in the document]
tsteur: definitely, updated 👍

core/Tracker/GoalManager.php
```diff
@@ -671,16 +671,36 @@ protected function recordStandardGoals(VisitProperties $visitProperties, Request
...
+    private function makeRandomMySqlUnsignedInt($length)
+    {
+        // mysql int unsgined max value is 4294967295 so we cannot start with a 5 or higher
+        $randomInt = Common::getRandomString(1, '1234');
+        if ($length >= 2) {
+            $randomInt .= Common::getRandomString(1, '012');
+        }
+        if ($length > 2) {
+            $randomInt .= Common::getRandomString($length - 2, '0123456789');
+        }
```
sgiehl: Actually even that could generate a too high number, even though the range is now much smaller between 4294967295 and 4299999999. Maybe it would be simpler to restrict the number to be between 1000000000 and 3999999999, e.g. disallow the 4 as first char?

tsteur: sure, makes totally sense.

tsteur: changed it.

PR #13728, opened 2018-11-19
Fix two issues causing new/returning visitor conversion rate to not appear

diosmosis:
Changes:
* when getting archive IDs must not ignore idarchives w/ different done flags.
  conversion rate requires number of conversions and number of visits, which will appear in two different archives, one for Goals, one for VisitsSummary. W/o grouping by `name`, only one was selected, the one for Goals. Which meant nb_visits was queried as 0, causing a conversion rate of 0.
* and aggregate dependent plugins for aggregate periods too.


diosmosis:
Some tests will need updating otherwise, ready for review.

plugins/Goals/Archiver.php
```diff
@@ -424,5 +424,8 @@ public function aggregateMultipleReports()
                 $columnsAggregationOperation,
                 $columnsToRenameAfterAggregation = null,
                 $countRowsRecursive = array());
+
+        $this->getProcessor()->processDependentArchive('Goals', API::NEW_VISIT_SEGMENT);
```
tsteur: can you quickly help me out where the stats for those two segments are shown?

diosmosis: here's the controller action: https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/Goals/Controller.php#L89

tsteur:
I meant some/many of the Goal metrics/archives are maybe not needed to be archived. All good though.

PR #13729, opened 2018-11-19
In LogAggregator, allow the use of a complex dimension w/ an already defined select as.

diosmosis:
This change allows the use of complex expressions as dimensions. Currently if someone uses, eg, a complex `CASE WHEN ... END` statement as a dimension, it will be duplicated in the select and in the group by of the query. This will allow users to add their own alias like `CASE WHEN ... END AS blahblah`, which will appear in the select as the entire `CASE WHEN ...` but as `blahblah` in the group by.

Also includes a `hasTable($table)` method addition to DataTable\Map so we can check if a table exists w/o triggering an error w/ getTable.

### DIFF
--- a/core/DataAccess/LogAggregator.php
+++ b/core/DataAccess/LogAggregator.php
@@ -423,6 +423,7 @@ class LogAggregator
      * @param $dimensions
      * @param $tableName
      * @param bool $appendSelectAs
+     * @param bool $parseSelectAs
      * @return mixed
      */
     protected function getSelectDimensions($dimensions, $tableName, $appendSelectAs = true)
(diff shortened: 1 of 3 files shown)


core/DataAccess/LogAggregator.php
```diff
@@ -432,11 +433,15 @@ protected function getSelectDimensions($dimensions, $tableName, $appendSelectAs
 
             if (!is_numeric($selectAs)) {
                 $selectAsString = $selectAs;
-            } else {
-                // if function, do not alias or prefix
-                if ($this->isFieldFunctionOrComplexExpression($field)) {
-                    $selectAsString = $appendSelectAs = false;
+            } else if ($this->isFieldFunctionOrComplexExpression($field)) {
```
tsteur: can you add a test for this?

diosmosis: Will add a unit/integration test. the system tests should show a difference (I haven't copied the files over)

diosmosis: Added a test

tsteur: just reading https://dev.mysql.com/doc/refman/8.0/en/select.html ... the `AS` is optional. I suppose we don't bother about this for now until this is a use case? Be best I think... 

does the regex need to check it case insensitive?

also just double checked there cannot be two AS statements in there cause we have an entry for each dimension... just fyi ... was wondering about that.

diosmosis: This PR isn't about requiring the `AS`, it's about detecting when it's there and just using it instead of assigning a new one. It's a workaround for when a complex sql expression is used as a dimension, which currently doesn't work since it just gets duplicated in both the select & group by parts.

Since it has to be used explicitly by a plugin developer, I think we don't have to check for uniqueness of the AS, since there will be an error and it will be programmer error.

Should be a case insensitive check, will look into that.

tsteur: > This PR isn't about requiring the AS, 

I meant if someone used `foo bar` we wouldn't detect the `bar` alias. All good though and not needed to be detected right now.

diosmosis:
FYI @tsteur assuming this is ok to merge, let me know if it needs another change, will make it in another PR

PR #13731, opened 2018-11-19
Improve check for dangerous links...

sgiehl:
and some other UI test related stuff

### DIFF
--- a/plugins/CorePluginsAdmin/tests/UI/TagManagerTeaser_spec.js
+++ b/plugins/CorePluginsAdmin/tests/UI/TagManagerTeaser_spec.js
@@ -15,7 +15,7 @@ describe("TagManagerTeaser", function () {
 
     function setPluginsToLoad(plugins)
     {
-        testEnvironment.pluginsToLoad = plugins
+        testEnvironment.pluginsToLoad = plugins;
         testEnvironment.save();
     }
 
@@ -82,11 +82,11 @@ describe("TagManagerTeaser", function () {
     });
 
     it('should be possible to disable page and redirect to home', function (done) {
-        capturePage(done, 'admin_page_disable', function (page) {
+        expect.page("").contains("#dashboard", function (page) {
             unloadTagManager();
             setAdminUser();
             page.click('.activateTagManager .dontShowAgainBtn');
-        }, '.pageWrap');
+        }, done);
     });
 
 });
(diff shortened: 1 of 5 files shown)


tests/lib/screenshot-testing/support/chai-extras.js
```diff
@@ -501,14 +501,37 @@ function performAutomaticPageChecks(done) {
...
+            var linkElements = document.getElementsByTagName('a');
+
+            elements:
+            for (var i = 0; i !== linkElements.length; ++i) {
+                var element = linkElements.item(i);
+
+                var href = element.getAttribute('href');
+                if (/^(javascript|vbscript|data):;*[^;]+/.test(href)) {
+                    for (var index in safeJsLinks) {
+                        if (safeJsLinks[index].test(href)) {
```
fdellwing: Is there any reason why you keep using regex here? Wouldn't it be easier to just test via `Array.prototype.indexOf()`? You are using `^` and `$` so, if you do not plan to use wildcards in safe links any time soon, it would be performanter to use index of.

fdellwing:
It might be better to run `Array.prototype.indexOf()` instead of `RegExp.prototype.test()`: https://jsperf.com/matomo-org-matomo-13731/1

But I have to admit, the difference is very minor on my system. And the performance gain will cannibalize itself, as soon as you use more wildcards.

plugins/CorePluginsAdmin/tests/UI/TagManagerTeaser_spec.js
```diff
@@ -82,11 +82,11 @@ describe("TagManagerTeaser", function () {
     });
 
     it('should be possible to disable page and redirect to home', function (done) {
-        capturePage(done, 'admin_page_disable', function (page) {
+        expect.page("").contains("#dashboard", function (page) {
```
tsteur: Is that change on purpose?

tsteur: there is no screenshot name etc... and shouldn't it capture?

sgiehl: yes. it now only checks if the loaded page has the element `#dashboard`. Imho there is no need to compare the actual content as a screenshot

sgiehl:
closing in favor of #13741

PR #13743, opened 2018-11-21
update cacert

Findus23:
Inspired by https://forum.matomo.org/t/certificate-issues-during-update/30238/6?u=lukas

There don't seem to be any major changes


Findus23:
Alternatively we could use https://github.com/composer/ca-bundle which would update automatically.

diosmosis:
@Findus23 is using https://github.com/composer/ca-bundle something you'd like to implement? I can merge this now otherwise.

PR #13748, opened 2018-11-22
don't allow chrome to translate Matomo pages

Findus23:
avoid unintentionally sending private data to foreign servers by Matomo users who have selected "always translate $language" in Chrome/Chromium

Currently this story of a webmail user who noticed that all his messages where sent to Google translate is going arround:
https://old.reddit.com/r/ProtonMail/comments/9yl94k/never_connect_to_protonmail_using_chrome/

While I don't believe that this happened without clicking on translate at least once, I think it shouldn't even be possible to accidentally send all Matomo do other servers.

Bonus fix: The safemode template was missing `<meta charset="utf-8">`. I am not sure if this was intentionally, but I think it is required for a proper HTML5 utf-8 page.


rugk (approved):
Nice :+1: 

Ref https://github.com/threema-ch/threema-web/pull/681

tsteur:
Should be good to merge if tests pass. What does it exactly do though? Can people still translate eg an admin page if they want to? It may be helpful for some people to translate some admin pages on occasion. Not everyone knows they can change language etc. Not a problem though if not possible anymore I suppose, just be good if it is still possible.

Findus23:
@tsteur 
https://support.google.com/webmasters/answer/79812?hl=en says the following regarding `notranslate`.
> When we recognize that the contents of a page are not in the language that the user is likely to want to read, we often provide a link to a translation in the search results. In general, this gives you the chance to provide your unique and compelling content to a much larger group of users. However, there may be situations where this is not desired. This meta tag tells Google that you don't want us to provide a translation for this page.

A quick test shows that I can still translate the page like before via right click, but setting Chrome to always translate German to English leaves the page alone.

tsteur:
👍

Issue #13466, opened 2018-09-21
possible error while exporting API data


mattab:
Error triggered: `Error: {"message":"Argument 1 passed to Piwik\\Plugins\\SegmentEditor\\UnprocessedSegmentException::__construct() must be an instance of Piwik\\Segment, string given, called in \/core\/ArchiveProcessor\/PluginsArchiver.php on line 185","file":"plugins\/SegmentEditor\/UnprocessedSegmentException.php","line":36}`

URL triggering issue: `/index.php?dataSource=page_title&date=2018-09-14,2018-09-20&expanded=1&filter_limit=-1&format=JSON2&idSite=8&limitActionsPerStep=6&method=UsersFlow.getUsersFlow&module=API&period=range&segment=dimension4%3D%xxxx%2520(x)`

mattab:
as we cannot reproduce this so far, and it only happened a few times, we will wait some more before working on this

mattab:
We again received this error many times eg.

`index.php?module=API&method=Actions.getPageTitles&idSite=1&period=range&date=2018-09-01,2018-09-30&segment=countryCode==BR&format=JSON&token_auth=XYZANONYMIZED&callback=jQuery111104736784728773332_1538375609097&_=1538375609098`

and

`index.php?module=API&method=UserCountry.getCity&idSite=1&period=range&date=2018-09-01,2018-09-30&segment=countryCode==CO&format=JSON&token_auth=XYZANONYMIZED&callback=jQuery111104736784728773332_1538375609111&_=1538375609112`

mattab:
Another request that triggered this error was `/index.php?dataSource=page_url&date=2018-11-16,2018-11-17&expanded=1&filter_limit=-1&format=JSON2&idSite=1&limitActionsPerStep=5&method=UsersFlow.getUsersFlow&module=API&period=range&segment=pageUrl%3D%5Ehttps%253A%252F%252FXXXXXXXXXXXXXXX.com%252Fjoin;pageUrl!@https%253A%252F%252FXXXXXXXXXXXX.com%252Fjoinnow`

PR #13758, opened 2018-11-24
Use new exception in PluginsArchiver & add previous exceptions to backtrace in fatal error report

diosmosis:
Refs #13466 

Changes:
* When wrapping thrown exception, use custom exception class since thrown exception may not have compatible constructor.
* Add previous exceptions to fatal error info's backtrace property.
* Recongize previous exceptions in API responses as well (so they will appear in core:archive output).

### DIFF
--- a/core/ArchiveProcessor/PluginsArchiver.php
+++ b/core/ArchiveProcessor/PluginsArchiver.php
@@ -177,7 +177,7 @@ class PluginsArchiver
                         $this->params->getSegment() ? sprintf("(for segment = '%s')", $this->params->getSegment()->getString()) : ''
                     );
                 } catch (Exception $e) {
-                    throw new PluginsArchiverException($e->getMessage() . " - caused by plugin $pluginName", $e->getCode(), $e);
+                    throw new PluginsArchiverException($e->getMessage() . " - in plugin $pluginName", $e->getCode(), $e);
                 }
             } else {
                 Log::debug("PluginsArchiver::%s: Not archiving reports for plugin '%s'.", __FUNCTION__, $pluginName);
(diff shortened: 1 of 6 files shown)


core/ArchiveProcessor/PluginsArchiver.php
```diff
@@ -177,15 +177,7 @@ public function callAggregateAllPlugins($visits, $visitsConverted, $forceArchivi
...
-                    $className = get_class($e);
-
-                    if ($className === 'PHPUnit_Framework_Exception' || (class_exists('PHPUnit_Framework_Exception', false) &&  is_subclass_of($className, 'PHPUnit_Framework_Exception'))) {
-                        $exception = new $className($e->getMessage() . " - caused by plugin $pluginName", $e->getCode(), $e->getFile(), $e->getLine(), $e);
-                    } else {
-                        $exception = new $className($e->getMessage() . " - caused by plugin $pluginName", $e->getCode(), $e);
-                    }
-
-                    throw $exception;
+                    throw new PluginsArchiverException($e->getMessage() . " - in plugin $pluginName", $e->getCode(), $e);
```
tsteur: I suppose that should work. Although wondering if anything may trigger like an `NoAccessException` which should maybe show login page etc. I don't think it's an actual issue though but maybe could be. Not sure if we handle some other exceptions in any specific way.

diosmosis: we could use reflection to change the `$message` property of the exception. Not sure how future proof that is, but I suppose the tests will tell us quickly if it breaks.

tsteur: Could work as well I suppose. I reckon it'd be just like https://stackoverflow.com/a/29542873 

Although I reckon the current approach is fine as well, just not sure if any specific exceptions would be treated differently.

diosmosis: @tsteur Does that mean this PR is ok to merge or do you want to think about it some more?

tsteur: I reckon it should be fine 👍

PR #13762, opened 2018-11-25
Make sure Period::toString() handles case when child period returns array and add a unit test.

### LINKED ISSUE #13709: Notice - Array to string conversion - Matomo 3.7.0-rc1 - core/Period.php(288)
Hello,

enabling debug file logs in Matomo I got this Warning : 
```
core/Period.php(288): Notice - Array to string conversion - Matomo 3.7.0-rc1
```

To reproduce : got to Behaviour/Outlinks/ then click on "Open segmented visitor log".

When Segmented link return no result, this message is thrown (uri with ampersands - https://github.com/matomo-org/matomo/issues/11806).

When results are returned this notice is not thrown.

PHP 5.5.9 

Thanks,

Eric

diosmosis:
Fixes #13709


core/Period/Day.php
```diff
@@ -84,21 +84,11 @@ public function addSubperiod($date)
...
-
-    /**
-     * Returns the current period as a string
-     *
-     * @return string
-     */
-    public function __toString()
-    {
-        return $this->toString();
+        return [$this->date->toString($format)];
```
tsteur: I'm thinking this would eg break Search Engine Keywords...

diosmosis: fixed

tsteur:
For some reason I can't reproduce the notice... could you? I assume it is fine to merge. Not sure where/if we use `$period . ''` or something similar in any meaningful context besides debug logs.

tsteur:
some failing tests but otherwise good to merge 👍

tsteur:
I don't think it is a BC concern but hard to say. But thinking about it again... a `toString` method should really not return arrays. Not sure how it is used... but ideally it would actually return a string and not arrays.

diosmosis:
I think currently __toString() returns a string, and toString() returns an array of strings. Could try and change it to return a string, will have to see how much code uses it.

tsteur:
All good, don't worry about it right now 👍

tsteur:
Feel free to merge if it fixes the problem. Only usage I could find was SearchEngine feature that used the day period and it doesn't seem to regress there.

PR #13763, opened 2018-11-25
Allow form select to show specific text for empty option that is not placeholder option…

### LINKED ISSUE #13698: Filter in Edit user permission view is confusing
This is a pure UX issue.

If you create a new user and want to give them read access to all websites, you will end up in this menu.

But while it seems clear that ticking the top tick selects all website, it is not obvious how you can get all websites to show up here.

The right selection has no label and the "show all" selection is the blank one. 
In addition the textbox to the left explains what it does, but in an nearly unreadable font color.

More information: https://forum.matomo.org/t/add-all-sites-to-user-permission/30407

diosmosis:
… & add missing translation.

image

Changes:

* In select form-fields', if there is already an `''` empty option, use that in the dropdown. This still allows showing the placeholder if the value is set to, eg, null or undefined).
* Add a 'Show all' entry to the access level filter dropdowns.
* Add a missing translation.

Fixes #13698

@Findus23 what do you think? Would this make selecting all sites clearer to you?

The placeholder font color is theme related so won't address in this issue. Though if it's a problem, perhaps a separate issue is called for.


tsteur:
fyi it worked for me and code looks 👍 . 

From a UX point of view it might be worth thinking about having some content above this saying something like "Give access to all websites [permission selector] [confirm button]". Then we could also add a note that user will get only access to currently existing websites, not to websites created in the future. But we could always add this later in a later release as it doesn't seem too important maybe.

diosmosis:
@tsteur  added the addition you suggested to the component

tsteur:
Works :+1: and nice to use. feel free to merge once tests pass

Findus23:
@diosmosis Thanks for the fix.

PR #13767, opened 2018-11-26
Remove old link

Findus23:
This URL hasn't existed in a long time and only redirects to the page that is already linked to in the same sentence.

BTW: Is there any reason for `Piwik::translate('<br/>')`?


sgiehl:
> BTW: Is there any reason for `Piwik::translate('<br/>')`?

That doesn't make sense. Feel free to replace it with a simple `<br/>`

sgiehl (approved):

PR #13768, opened 2018-11-26
add tagmanager containers to .gitignore

Findus23:
another really minor pull request, but otherwise my git status gets even more messy than it already is.


sgiehl (approved):
makes sense :+1:

PR #13769, opened 2018-11-26
Enable row evolution for flattened reports

### LINKED ISSUE #7414: When a report is flattened, show the Row Evolution icon
The goal of this issue is to display the "Row Evolution" icon on all reports, when this report is flattened.

Currently we do not display the Row Evolution icon because the Flatten reports were too slow to generate. Therefore exposing this feature would have been bad user experience in many cases. Now that we have made Flatten much faster in https://github.com/piwik/piwik/issues/6763#issuecomment-77797219 then we could again re-display this icon. I believe that speed was the only reason we didn't display the icon (rather than having to write new code etc.).

The value for users is huge: for example seeing Evolution of metrics for Flattened page urls is priceless.

sgiehl:
should be tested maybe on demo or on another instance with some more data to check if performance is fine.
Also not sure if flattened custom reports are fast enough for this. ping @tsteur 

fixes #7414

### DIFF
--- a/plugins/CoreHome/javascripts/dataTable.js
+++ b/plugins/CoreHome/javascripts/dataTable.js
@@ -630,11 +630,17 @@ $.extend(DataTable.prototype, UIControl.prototype, {
         var tableRowLimits = piwik.config.datatable_row_limits,
         evolutionLimits =
         {
-            day: [8, 30, 60, 90, 180, 365, 500],
-            week: [4, 12, 26, 52, 104, 500],
+            day: [8, 30, 60, 90, 180],
+            week: [4, 12, 26, 52, 104],
             month: [3, 6, 12, 24, 36, 120],
             year: [3, 5, 10]
         };
+
+        // only allow big evolution limits for non flattened reports
+        if (!parseInt(this.param.flat)) {
+            evolutionLimits.day.push(365, 500);
+            evolutionLimits.week.push(500);
+        }
 
         var self = this;
         if (typeof self.parentId != "undefined" && self.parentId != '') {
--- a/plugins/CoreHome/javascripts/dataTable_rowactions.js
+++ b/plugins/CoreHome/javascripts/dataTable_rowactions.js
@@ -97,9 +97,6 @@ DataTable_RowActions_Registry.register({
         return (
             typeof dataTableParams.disable_row_evolution == 'undefined'
                 || dataTableParams.disable_row_evolution == "0"
-            ) && (
-            typeof dataTableParams.flat == 'undefined'
-                || dataTableParams.flat == "0"
             );
     },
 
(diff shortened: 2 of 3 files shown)


tsteur:
I would suggest to first test it on various production systems as well. I wouldn't be surprised if this takes ages and not even sure this should be merged but to be seen re performance.

tsteur:
@sgiehl I've tried this patch on https://demo.matomo.org/index.php?module=CoreHome&action=index&idSite=62&period=day&date=yesterday#?idSite=62&period=day&date=yesterday&category=General_Actions&subcategory=General_Pages&popover= but it says 

`Error: Expecting only one row per DataTable`

Query URL in this case was https://demo.matomo.org/index.php?date=yesterday&apiMethod=Actions.getPageUrls&label=%40%252Fcart%252F&disableLink=1&module=CoreHome&action=getRowEvolutionPopover&colors=%7B%22backgroundColor%22%3A%22%23ffffff%22%2C%22lineColor%22%3A%22%23162c4a%22%2C%22minPointColor%22%3A%22%23ff7f7f%22%2C%22maxPointColor%22%3A%22%2375bf7c%22%2C%22lastPointColor%22%3A%22%2355aaff%22%2C%22fillColor%22%3A%22%23ffffff%22%7D&flat=1&idSite=62&period=day 

it also included the flat=1

sgiehl:
hm... it worked for me locally for various reports. Seems to come from the sparklines. Need to check if I'm able to reproduce it locally somehow

sgiehl:
It works for example for https://demo.matomo.org/index.php?module=CoreHome&action=index&idSite=3&period=day&date=yesterday&updated=1#?idSite=3&period=day&date=yesterday&category=General_Actions&subcategory=General_Pages&popover=RowAction$3ARowEvolution$3AActions.getPageUrls$3A$257B$2522flat$2522$253A$25221$2522$257D$3A$40$252Fphotos.php$252Fvirtual-drums-laval-virtual-007$252F

sgiehl:
@tsteur tried to reproduce that locally until I recognized that it has nothing to do with the changes here. The row evolution doesn't even work when clicking it for an unflattened page url report for that page on demo. Is that page maybe proxied or has anything special? Don't have access to that page with my account on demo, so can't look by myself...

tsteur:
Good find... works for eg idSite=1. Must be an issue with the siteproxy (fyi @diosmosis ).

It takes about 8 seconds to load for a flattened URL when 24 data points are shown, took more than 3 minutes when choosing 120 data points in the selector (was expected I guess). Don't wanna know how long 500 data points take :) Maybe we should remove those options for flattened reports :) (but probably not easily doable).

sgiehl:
@tsteur I've disabled the higher limits for flattened reports now (and also disabled the flatten selector as it doesn't make sense to unflatten an evolution chart)

tsteur:
tested & worked. LGTM if tests pass 👍

PR #13773, opened 2018-11-28
Update custom dimension submodule

tsteur:


tsteur:
There are some failing tests but shouldn't be related to this change.

PR #13787, opened 2018-12-02
Add timezone to current date hours instead of directly setting to timezone offset.

### LINKED ISSUE #13786: Report for "today" is not the same as report for any date explicitly selected
I have my Matomo dashboard set to default to "today", and the string "today" appears in a query field of the URL when I go to the dashboard.  When I go to the dashboard, I get the current day's stats under "Visits overview" but the previous day's graph under "Visits over time," and the previous day's date is shown in the drop-down date selector widget.  But if I manually select the current date, I get the current date shown in the URL and the date selector, and the graph shows today's statistics.  Choosing the string value "today" as a default has a different effect from explicitly selecting the current date; some parts of the dashboard use the previous day's date when "today" is the default.

See screenshot - Matomo is configured to use "today," as seen in the URL; and as seen in the live visitor log, today is December 1; but the date-selector widget has defaulted to November 30.

screenshot_2018-12-01_17-15-44

I don't believe that this is caused by a time zone issue; all my time zones match, and all aspects of it appear to roll over at midnight local time every day.  It seems to be new behaviour in 3.7.0; the previous version correctly showed today's stats in all parts of the dashboard when so configured.

diosmosis:
Fixes #13786

### DIFF
--- a/plugins/CoreHome/angularjs/common/services/periods.js
+++ b/plugins/CoreHome/angularjs/common/services/periods.js
@@ -346,7 +346,7 @@
         date.setTime(date.getTime() + date.getTimezoneOffset() * 60 * 1000);
 
         // apply piwik site timezone (if it exists)
-        date.setHours((piwik.timezoneOffset || 0) / 3600);
+        date.setHours(date.getHours() + ((piwik.timezoneOffset || 0) / 3600));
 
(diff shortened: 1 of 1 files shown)


mskala:
I haven't tested it extensively for side effects, but it seems to work.  Thanks!

PR #13788, opened 2018-12-02
Mention the --yes flag in core:update description

axilleas:
The --yes flag is essential if you want to trigger unattended Matomo
upgrades. This is only mentioned in the Debian package post-install
script:

https://github.com/matomo-org/matomo-package/blob/204f845ddbe3527104994788b6beac1ea6190ece/debian/matomo.postinst#L96

This change adds the --yes flag to the core:update description.

---

I had to try different things like `--no-interaction`, which I thought would work with `core:update`, but it didn't. I had to search and find this issue https://github.com/matomo-org/matomo-package/issues/23, which gave me the idea to search the post-install script of the Debian package.

PS. Are translations handled automatically?


tsteur:
cheers 👍

PR #13789, opened 2018-12-02
Fix header color style should not be applied to menu

tsteur:
Otherwise, when setting for example a white background for the header, the menu items won't be visible. refs DEV-1504


sgiehl (approved):
Works

PR #13798, opened 2018-12-04
Introduce whitelist test for link protocols.

diosmosis:


core/UrlHelper.php
```diff
@@ -17,7 +17,14 @@
  */
 class UrlHelper
 {
-    
+    private static $validLinkProtocols = [
+        'http',
+        'https',
+        'tel',
```
fdellwing: Should we also asume `callto` as safe, as it is still used by some old system for initiating Skype calls.

Findus23: When in doubt, I'm against it. `callto` is no official standard, so I can't promise that there is no way to exploit people clicking on the link accidentally and maybe automatically call a paid number or something like this, because some old badly written software doesn't want to make a confirm screen.

fdellwing: This can also happen with `tel` if you (or some malware) changed the default program.

plugins/Actions/templates/_profileSummary.twig
```diff
@@ -11,7 +11,7 @@
                             url|trim|lower starts with 'data:' %}
                                 {{ url }}
                             {% else %}
-                                <a href="{{ url }}" rel="noreferrer noopener" target="_blank"
+                                <a href="{{ url|e('url') }}" rel="noreferrer noopener" target="_blank"
```
tsteur: haven't tested but think the URL escape should be usually used only on components, not on the whole URL? maybe double check those links still work when they include protocol etc

tsteur: from the docs

> url: escapes a string for the URI or parameter contexts. This should not be used to escape an entire URI; only a subcomponent being inserted.

tsteur: same for the other usages of this

tsteur: @diosmosis can you change the escaping?

diosmosis: Updated the PR, added a `whole_url` filter that can be used to whitelist the url. Has to be used in conjunction w/ html_attr or `|raw` otherwise twig will automatically escape w/ `html`.

core/Twig.php
```diff
@@ -65,7 +65,9 @@ function piwik_fix_lbrace($string)
 
 function piwik_escape_filter(Twig_Environment $env, $string, $strategy = 'html', $charset = null, $autoescape = false) {
 
-    $string = twig_escape_filter($env, $string, $strategy, $charset, $autoescape);
+    if ($strategy != 'whole_url') {
+        $string = twig_escape_filter($env, $string, $strategy, $charset, $autoescape);
```
tsteur: any reason this is not done for whole_url? wouldn't it be the same as simple adding like a filter `whole_url` that does below logic?

core/Twig.php
```diff
@@ -74,6 +76,11 @@ function piwik_escape_filter(Twig_Environment $env, $string, $strategy = 'html',
         case 'url':
             $encoded = rawurlencode('{');
             return str_replace('{{', $encoded . $encoded, $string);
+        case 'whole_url':
+            if (!UrlHelper::isLookLikeSafeUrl($string)) { // if this could be a dangerous link, replace it w/ an empty url
+                return 'javascript:;';
```
tsteur: would this need to be `javascript:void(0)`?

core/Twig.php
diosmosis: When twig sees an escape like `|e('html_attr')` it doesn't apply the default `|e('html')` escape to a variable. But when twig sees `|e('whole_url')` it applies the default escape. So if we call `twig_escape_filter` above, it double encodes the URL. Only way I could come up w/ to not do that, is to make sure whole_url doesn't escape anything.

core/Twig.php
diosmosis: I guess it could be? I'm not sure if it would change anything (assuming it's for `preventDefault()`?).

core/Twig.php
tsteur: My message was bit confusing, not sure if understood what I meant. Just to be sure... what I meant was we could add `whole_url` as a regular twig filter like this maybe:

```
    private function addFilter_wholeUrl()
    {
        $filter = new \Twig_SimpleFilter('wholeUrl', function ($value) {
            if (!UrlHelper::isLookLikeSafeUrl($string)) { ... } return $value
         });
        $this->twig->addFilter($filter);
    }
```

and then use it like `{{ link|wholeUrl|e('html_attr')}}`.
Then we wouldn't need the above `if`... as `whole_url` is not really escaping but more like filtering/changing the value? And using `|e('whole_url')` could give a developer the impression the value will be escaped there (and think another escape html_attr is not needed) which in reality it doesn't...

diosmosis: Oh I see, yes, makes sense.

diosmosis:
@tsteur looks like there was already a safelink filter, modified & used that.

core/UrlHelper.php
```diff
@@ -135,10 +143,20 @@ public static function isLookLikeUrl($url)
...
         return preg_match('~^(([[:alpha:]][[:alnum:]+.-]*)?:)?//(.*)$~D', $url, $matches) !== 0
             && strlen($matches[3]) > 0
-            && !in_array($matches[1], array('javascript:', 'vbscript:', 'data:'))
+            && !preg_match('/^(javascript:|vbscript:|data:)/i', $matches[1])
             ;
     }
 
+    public static function isLookLikeSafeUrl($url)
+    {
+        if (strpos($url, ':') === false) {
```
tsteur: Looks good to merge but maybe we should here also execute `self:: isLookLikeUrl ($url)` and only return true if it actually is a url?

core/UrlHelper.php
```diff
@@ -135,10 +143,20 @@ public static function isLookLikeUrl($url)
...
+            return true;
+        }
+
+        $protocol = explode(':', $url, 2)[0];
+        return preg_match('/^' . implode('|', self::$validLinkProtocols) . '/i', $protocol);
```
tsteur: small tweak maybe... should the protocol also check for ends with this? as in `'/^' . implode('|', $validLinkProtocols) . '$/i'` otherwise `smstest` might be considered a valid protocol?

core/UrlHelper.php
diosmosis: `tel:555...` and `sms:...` links do not pass isLookLikeUrl, since they don't have `//`. I didn't want to regress anything since isLookLikeUrl is used in tracking and probably everywhere else, so didn't modify it.

diosmosis:
Tweaked the regex.

core/UrlHelper.php
diosmosis: Note: I did try and make the `//` optional, but then everything passes `isLookLikeUrl`

tsteur:
feel free to merge when tests pass

PR #13823, opened 2018-12-11
Fix possible json encoding error

tsteur:
some API's in tests don't return a response cause they fail to json_encode because of `Malformed UTF-8 characters, possibly incorrectly encoded`. Tested this fix and it worked for me. Would maybe also need to create a general JSON class for this that always takes care of this in case we have same problem somewhere else? Not sure how good a fix it is as the input might not be utf8?

In test case there was eg `/page/index.htm?q=non unicode keyword ��������";s:7:"segment";s:103:"pageUrl==http%3A%2F%2Fexample.org%2Fpage%2Findex.htm%3Fq%3Dnon+unicode+keyword+%EC%E5%F8%EA%EE%E2%FB%E5";}i` and it was converted to `??????`

### DIFF
--- a/core/DataTable/Renderer/Json.php
+++ b/core/DataTable/Renderer/Json.php
@@ -74,7 +74,7 @@ class Json extends Renderer
         $str = @json_encode($array);
 
         if ($str === false && json_last_error() === JSON_ERROR_UTF8) {
-            $str = @json_encode(self::makeArrayUtf8($array));
+            $str = json_encode(self::makeArrayUtf8($array));
         }
 
         return $str;
--- a/core/DataTable/Renderer/Json.php
+++ b/core/DataTable/Renderer/Json.php
@@ -73,7 +73,28 @@ class Json extends Renderer
         // silence "Warning: json_encode(): Invalid UTF-8 sequence in argument"
         $str = @json_encode($array);
 
+        if ($str === false && json_last_error() === JSON_ERROR_UTF8) {
+            $str = @json_encode(self::makeArrayUtf8($array));
+        }
+
         return $str;
+    }
+
+    private static function makeArrayUtf8($array)
+    {
+        if (!function_exists('mb_convert_encoding')) {
+            return $array;
+        }
+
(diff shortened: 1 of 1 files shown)


core/DataTable/Renderer/Json.php
```diff
@@ -73,9 +73,30 @@ protected function renderTable($table)
...
+        if (!function_exists('mb_convert_encoding')) {
+            return $array;
+        }
+
+        if (is_array($array)) {
+            foreach ($array as $key => $value) {
+                $array[$key] = self::makeArrayUtf8($value);
+            }
+        } elseif (is_string($array)) {
+            return mb_convert_encoding($array, 'UTF-8', 'UTF-8');
```
tsteur: or maybe `return mb_convert_encoding($array, 'UTF-8', 'auto');` is better?

core/DataTable/Renderer/Json.php
```diff
@@ -73,9 +73,30 @@ protected function renderTable($table)
         // silence "Warning: json_encode(): Invalid UTF-8 sequence in argument"
         $str = @json_encode($array);
 
+        if ($str === false && json_last_error() === JSON_ERROR_UTF8) {
+            $str = @json_encode(self::makeArrayUtf8($array));
```
diosmosis: Maybe we shouldn't silence this second function call? Just so any other issues won't be silenced.

tsteur: makes sense 👍

tsteur:
fyi we use slightly different version in https://github.com/matomo-org/matomo/blame/3.x-dev/core/DataTable/Renderer/Csv.php#L477

but same version for https://github.com/matomo-org/matomo/blame/3.x-dev/core/DataTable/Renderer.php#L188-L193

not sure if any is better... the one in data table renderer should be fine though I suppose

diosmosis:
Tests are passing, feel free to merge (not sure if you want to change to use the DataTable\Renderer code).

PR #13830, opened 2018-12-12
Increase the time before an archive is considered outdated

tsteur:
Browser archiving is already slow enough so we shouldn't do it every 2.5minutes IMO. Personally, I would even set it to 30 or 60 minutes, not 15 minutes like in this PR.
This improves the performance at least a little when users don't set up cli archiving.
@mattab


diosmosis:
FYI @mattab merging this. cc-ing just in case you want to say something

PR #13832, opened 2018-12-12
Allow period used to generate scheduled report to be different from email schedule.

### LINKED ISSUE #13819: Email Report - allow selection of data time range independently from schedule
Hello,

I would have a feature request which I believe many users might share.

The use case I have is that I would like to send periodically (say "weekly") an email report, and the report should include data over (say) the past 12 months.

I do not seem to find a way to automate this: I can only pick the email schedule of the report (say "weekly") and this will send me, every week, data related to the last week.
I would instead like to be able to define what **relative** date range the report should cover (e.g. last N weeks/months).

Manually I can do this already: if in the top date picker I select the past year, then I can manually download / send the PDF report too.
Any way this can be added to the report configuration?

Thank you,
Michele

diosmosis:
Simple solution that does not support range periods.

Refs #13819


tsteur:
Can we list the schedule first and then report period? and add some inline help for report period? also I would maybe automatically set report period to the same value when changing the email schedule but not the other way around. By default they should always match unless someone changes report period. If they changed report period, and then change the schedule, I would again also set report period to the same as email schedule.(no need to remember whether it was different I think, but could maybe... not sure which is better).

diosmosis:
👍 will do

diosmosis:
Updated the PR. @tsteur let me know if this can go in 3.8

sgiehl:
@diosmosis this one needs a rebase to update the version and the update script name

diosmosis:
Updated.

tsteur:
Looks good to merge, we only need to update version number and update script. Feel free to merge if tests pass 👍

mattab:
Seeing this late and commenting late, but is this feature any useful? I understand the use case in https://github.com/matomo-org/matomo/issues/13819 "I would like to send periodically (say "weekly") an email report, and the report should include data over (say) the past 12 months.". But in its current form the feature does not let users do this. I can't think of any use case for it. Should we maybe revert the PR to not make the scheduled reports harder to use? or is there some use case i'm missing?

mattab:
Actually i can think of a use case, where you can now use this feature to receive "Daily" the "Current month" report (or "Current week") report. Which seems valuable :+1: 
Would this use case work? (haven't tested)

diosmosis:
That would work, was mostly for getting weekly email of daily report (for the system report in cloud since we don't want to sum each day up for that; only the daily report is accurate).

PR #13838, opened 2018-12-12
Make sure tag manager can only be deactivated, not uninstalled

### LINKED ISSUE #13835: Matomo Tag Manager can be uninstalled
MTM is a core plugin and shouldn't be uninstallable (only can be disabled)

Nevertheless Matomo shows a working uninstall button
grafik
See https://forum.matomo.org/t/fehler-nach-desinstallation-von-tag-manager/30904/8 for more details.

I think this is caused by
https://github.com/matomo-org/matomo/blob/e571b67d8f6b2764a2bdbe3d017f0f9ddf147508/core/Plugin/Manager.php#L256-L259

https://github.com/matomo-org/matomo/blob/e571b67d8f6b2764a2bdbe3d017f0f9ddf147508/core/Plugin/Manager.php#L675-L680

But TagManager is not enabled by default, but also doesn't appear in 
https://github.com/matomo-org/matomo/blob/e571b67d8f6b2764a2bdbe3d017f0f9ddf147508/core/Application/Kernel/PluginList.php#L34-#L45

So I guess it should either be added to the list (which may have side effects) or maybe a special case for TagManager should be added to `isPluginBundledWithCore`

tsteur:
fix https://github.com/matomo-org/matomo/issues/13835

### DIFF
--- a/core/Application/Kernel/PluginList.php
+++ b/core/Application/Kernel/PluginList.php
@@ -41,7 +41,8 @@ class PluginList
         'ExampleTracker',
         'ExampleReport',
         'MobileAppMeasurable',
-        'Provider'
+        'Provider',
+        'TagManager'
     );
 
     // Themes bundled with core package, disabled by default


Findus23 (approved):
Thanks for the fix.

This solves the issue for me.

PR #13842, opened 2018-12-13
Add capability editing to users manager

diosmosis:
Fixes #13842 

Capability editing:

image

Capability tooltip:

image

Capability help in form:

image

filter by capability:

image

filter by capability in paged users list:

image


tsteur:
image
Seeing only "write" there we may need to prefix the category? But then might take too much space... or could we group them by category name? Or maybe it's just renaming the name of the "Write" category now to "Tag Manager Write". I guess also the other capabilities may not be clear what they refer to unless you hover the tooltip. At the same time I like the compact style.

Do we also need ability to set capabilities when giving access to all websites? Maybe not needed for now?

tsteur:
Not sure what it would look like... would it be possibly by default show kind of as regular text `no capabilities` if none, or the assigned capabilities as words if some are granted `Tag Manager: Write, Publish Live Container`. So `Category: capabilities, Category: Capabilities`. It would be only one row because Tag Manager is the only plugin defining it so far.

This only viewing what the current state is. To assign them, we could have behind each page an edit /expand button? I know it could be quite painful when wanting to grant many capabilities to many sites, but for this we also have the bulk actions... and I reckon those capabilities are very rarely used. So personally I'd be fine with that. 

Or maybe we could somehow use [chips](http://archives.materializecss.com/0.100.2/chips.html) which are a bit more efficient?

diosmosis:
Chips might work if we combine it w/ a dropdown (ie, instead of the user typing in a new chip, the blank chip is a select w/ the list of available capabilities).

If a capability is included in a role, we could show it w/ the remove icon disabled. Will try it.

tsteur:
image

I just remember we have this UI field as well in Custom Reports called `expandable-select`.

```
<div piwik-field uicontrol="expandable-select" name="..."
                             full-width="true"
                             ng-change="editReport...."
                             title="..."
                             ng-model="editReport.report...."
                             options="editReport....">
                        </div>
```

not sure if that maybe is useful?

diosmosis:
Could help, thanks!

diosmosis:
@tsteur changed to use chips + added capabilities to add access to all sites + to bulk actions. screenshot:

image

tsteur:
Looks good 👍

PR #13845, opened 2018-12-13
Reject urls w/ control characters.

diosmosis:


tsteur:
Looks good if tests pass

PR #13848, opened 2018-12-13
add Themes to System Menu

### LINKED ISSUE #13211: Make possibility for themes more prominent
I have a feeling that many Matomo users aren't aware of the fact that if they dislike a part of the design of Matomo, they can write a very basic theme that overwrites the css/less and with little effort can publish it on the Marketplace for everyone to use.

At the moment, themes are quite hidden. Especially the management of themes is hidden behind a small link in the header of the plugin management.

screenshot_20180725_220450

I am not sure where to link instead, but maybe even a separate menu item for themes could be added.

Findus23:
fixes #13211

I created a PR to get the discussion about #13211 rolling. 
Personally I think most Matomo users are not aware of the fact that you can really easily change the style of Matomo just by switching out a handful of color.
My Matomo plugins that are only useful to a really small amount of people have more downloads than [a theme](https://themes.matomo.org/ClassicTheme) that could be interesting to many Matomo users.

This would also allow creating more Matomo Themes as they would then be seen by more people.
(I think a high contrast accessibility and a dark/night theme would be quite useful)

Regarding the Change: Updates would still be shown in the Plugins menu, but I am not sure if showing them twice is easy to do.

### DIFF
--- a/plugins/CorePluginsAdmin/Menu.php
+++ b/plugins/CorePluginsAdmin/Menu.php
@@ -67,7 +67,10 @@ class Menu extends \Piwik\Plugin\Menu
         if ($hasSuperUserAcess) {
             $menu->addSystemItem(Piwik::translate('General_Plugins') . $pluginsUpdateMessage,
                 $this->urlForAction('plugins', array('activated' => '')),
-                $order = 20);
+                $order = 18);
+            $menu->addSystemItem(Piwik::translate('CorePluginsAdmin_Themes'),
+                $this->urlForAction('themes'),
+                $order = 19);
         }
     }
 
--- a/plugins/Marketplace/Plugins.php
+++ b/plugins/Marketplace/Plugins.php
@@ -172,7 +172,6 @@ class Plugins
     }
 
     /**
-     * @param bool $themesOnly
      * @return array
      */
     public function getPluginsHavingUpdate()


tsteur:
Wouldn't users maybe discover them more through the marketplace than the themes page? I think the amount of people that would actually start developing a theme and take the time to customise the look can be likely counted on two hands or so :) And the ones that are interested in doing this, might still find it. We removed Themes from the main menu on purpose as themes are not even much used even when in main menu. Maybe instead rather add a notice in "Branding settings" that it is possible to customize themes (it would need to be possible to disable this message or it should be disabled if plugins admin is disabled)?

Findus23:
> Wouldn't users maybe discover them more through the marketplace than the themes page?

Maybe you are right. But the marketplace page is also a bit hidden and looks broken at the moment as two of the six themes, don't have a screenshot and therefore just show `src="?w=250&h=150"`. Maybe we could add a requirement on submitting a theme that there has to be a screenshot.
grafik

> I think the amount of people that would actually start developing a theme and take the time to customise the look can be likely counted on two hands or so :)

Well in that case that would mean two times more Matomo themes 😉
We don't need as many themes as plugins, but a well maintained set of a handful different themes (especially a dark and an accessibility theme) would attract many users in my opinion.

I'm also not totally happy with this PR, but we need some other way to link to the themes page as the current link is completely invisible. Maybe someone else has an idea!

tsteur:
> We don't need as many themes as plugins, but a well maintained set of a handful different themes (especially a dark and an accessibility theme) would attract many users in my opinion.

To attract more users for this we would possibly need to mention it a bit more on the website. Otherwise users would only notice it when already installed or so. We might actually mention this a bit more as part of the white label page etc. 

I'll create an issue for requiring screenshots but it's a bit of work (cause we would need to change quite some logic and validate whether it has screenshots before unzipping and creating any plugin internally in the DB etc). I'll create an internal issue for this for now 👍 Maybe the developer docs could also be tweaked to highly encourage people to upload screenshots?

Findus23:
> I'll create an issue for requiring screenshots but it's a bit of work

In that case just leave it as it is and I'll continue to contact the developers and encourage them to add screenshots (as I have done now)

tsteur:
👍  awesome

sgiehl:
@mattab code looks fine. Shall we merge this one then? Would update the tests if so...

diosmosis:
@mattab can you reply to this thread?

mattab:
Reckon we don't need the new menu entry, but if needed we could maybe show Themes more clearly on the Plugins page itself.

PR #13855, opened 2018-12-13
Make sure to always trigger a callback in JS trigger when specified

tsteur:


js/piwik.js
```diff
@@ -3561,6 +3568,10 @@ if (typeof window.Piwik !== 'object') {
                     return false;
                 }
 
+                if (success && typeof callback === 'function') {
+                    callback({request: request, trackerUrl: configTrackerUrl, success: true});
```
diosmosis: Since this assumes success, would it help to add, eg, `isSendBeacon: true`? Just in case someone might want to tell. I'm not sure of a use case for that though.

tsteur: doesn't hurt 👍  will add it

tsteur: done

PR #13866, opened 2018-12-16
Write correct output when a process is skipped

tsteur:
Prevents errors like ` Error unserializing the following response from ?module=API&method=API.get&idSite=1&period=year&date=last3&format=php&trigger=archivephp: Skipped`

When I tested locally it also avoided the message `WARNING [2018-12-16 22:40:18] 5677  piwik/core/Common.php(271): Notice - unserialize(): Error at offset 0 of 7 bytes `

refs DEV-1521

It is unserialized eg here: https://github.com/matomo-org/matomo/blob/3.x-dev/core/CronArchive.php#L867 and then evaluated here: https://github.com/matomo-org/matomo/blob/3.x-dev/core/CronArchive.php#L1433-L1463 (the other one expects an array of array but still works cause of `empty` check)

### DIFF
--- a/core/CliMulti.php
+++ b/core/CliMulti.php
@@ -142,7 +142,7 @@ class CliMulti
             if ($shouldStart === Request::ABORT) {
                 // output is needed to ensure same order of url to response
                 $output = new Output($cmdId);
-                $output->write(serialize(array('skipped' => '1', 'nb_visits' => 0)));
+                $output->write(serialize(array('aborted' => '1')));
                 $this->outputs[] = $output;
             } else {
                 $this->executeUrlCommand($cmdId, $url);
--- a/core/CliMulti.php
+++ b/core/CliMulti.php
@@ -142,7 +142,7 @@ class CliMulti
             if ($shouldStart === Request::ABORT) {
                 // output is needed to ensure same order of url to response
                 $output = new Output($cmdId);
-                $output->write('Skipped');
+                $output->write(serialize(array('skipped' => '1', 'nb_visits' => 0)));
                 $this->outputs[] = $output;
             } else {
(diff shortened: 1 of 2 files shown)


core/CliMulti.php
```diff
@@ -142,7 +142,7 @@ private function start($piwikUrls)
             if ($shouldStart === Request::ABORT) {
                 // output is needed to ensure same order of url to response
                 $output = new Output($cmdId);
-                $output->write('Skipped');
+                $output->write(serialize(array('skipped' => '1', 'nb_visits' => 0)));
```
tsteur: the output doesn't really matter as long as it is a non-empty serialized array as we check with `empty($metric['nb_visits'])` etc.

tsteur: In theory the response would need to depend on the caller but maybe in this case it isn't too important? checked it is used as well in `ScheduledTaskRunner` but it doesn't use the "abort" feature and the response format doesn't matter there

tsteur: It's not really an error I think. It is just not needed to execute it anymore basically (eg because the same job is already running). Setting `null` could be difficult maybe as it would be hard to differentiate between the aborted case and eg no response received etc. 

So far we call CLI Multi only in Task Runner and Cron Jobs and there it is all fine. The `abort` feature is not really any documented feature for plugins so shouldn't be a problem. Could set it to `$output->write(serialize(array('aborted' => '1')))` though and explain the output in the description of the constant... maybe that makes it more clear. A caller has to specifically implement some support to `ABORT` a command so I don't think it's a problem.

tsteur: Just changed the output... the `ABORT` feature in there is currently actually specifically in there for the archiver anyway.

Issue #13852, opened 2018-12-13
Upgrade to 3.8.0b4 triggered failure to login without 'session_save_handler = '


GreenReaper:
332f42c8798685d1a7df425ebfa42b35937b9b5f (#13540) deprecates the files session hander, setting dbtable as default.

This, or this in combination with some other change, appears to have broken my setup - after upgrading using automatic update from 3.8.0b2, I received continuous 'form security failed' errors when attempting to login with the superuser (the only user, other than anonymous), after submitting to apply database changes (having also run into #13836).

I was only able to login correctly after changing to the file save handler by setting in config.ini.php:
> [General]
> session_save_handler =

Looking at the table `piwik_session` there are seemingly valid Base64 serialized strings decoding as e.g.:
> a:2:{s:11:"Login.login";a:1:{s:5:"nonce";s:32:"1d94bbb57080ef82062935becf00da7c";}s:4:"__ZF";a:1:{s:11:"Login.login";a:1:{s:4:"ENVT";a:1:{s:5:"nonce";i:1544720442;}}}}

I'm guessing this is the partially-logged-in-session as the ID matches the cookie.

But there are also strange, seemingly corrupted entries that don't even have the "data?" at the front of the data column and presumably aren't going to decode right either, like this:
> | id                         | modified   | lifetime | data     
> | ISbcHaC6nB1ACClXZ059KUsqkN | 1544725496 |  1209600 |    datYjowOw== 

I also saw errors like this pop up in the debug log:
> WARNING Piwik\ErrorHandler[2018-12-13 19:09:25 UTC] [b9532] /[censored]/piwik/libs/Zend/Session.php(495): Notice - unserialize(): Error at offset 43 of 47 bytes - Matomo 3.8.0-b4
... 
> WARNING Piwik\ErrorHandler[2018-12-13 19:32:30 UTC] [bbf15] /[censored]/piwik/libs/Zend/Session.php(495): Notice - unserialize(): Error at offset 1195 of 1199 bytes - Matomo 3.8.0-b4

I think #13823 was intended to fix this, but maybe it did not work, or caused more problems than it solved.

In case it matters
* I'm using nginx 1.15.7, accessing `/piwik/index.php`, the site tracked being at `/` (but for some reason it seems to be setting a cookie on the '/' path when logging in, and also on '/piwik' once it gets in).
* The server is using PHP 7.3.0; session.save_handler = files, and session.serialize_handler = igbinary.
[...]

PR #13869, opened 2018-12-17
Send all session cookie params when updating session expire time.

diosmosis:
Refs #13852

### DIFF
--- a/core/Session.php
+++ b/core/Session.php
@@ -81,6 +81,8 @@ class Session extends Zend_Session
         // the session data won't be deleted until the cookie expires.
         @ini_set('session.gc_maxlifetime', $config->General['login_cookie_expire']);
 
+        @ini_set('session.cookie_path', empty($config->General['login_cookie_path']) ? '/' : $config->General['login_cookie_path']);
+
         $currentSaveHandler = ini_get('session.save_handler');
 
         if (!SettingsPiwik::isPiwikInstalled()) {
--- a/core/Session/SessionAuth.php
+++ b/core/Session/SessionAuth.php
@@ -180,7 +180,10 @@ class SessionAuth implements Auth
 
     private function updateSessionExpireTime()
     {
+        $sessionParams = session_get_cookie_params();
+
         $sessionCookieLifetime = Config::getInstance()->General['login_cookie_expire'];
-        setcookie(session_name(), session_id(), time() + $sessionCookieLifetime);
+        setcookie(session_name(), session_id(), time() + $sessionCookieLifetime, $sessionParams['path'],
+            $sessionParams['domain'], $sessionParams['secure'], $sessionParams['httponly']);
     }
 }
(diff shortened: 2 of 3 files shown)


core/Session/SessionAuth.php
```diff
@@ -180,7 +180,10 @@ public function getTokenAuth()
 
     private function updateSessionExpireTime()
     {
+        $sessionParams = session_get_cookie_params();
+
         $sessionCookieLifetime = Config::getInstance()->General['login_cookie_expire'];
-        setcookie(session_name(), session_id(), time() + $sessionCookieLifetime);
+        setcookie(session_name(), session_id(), time() + $sessionCookieLifetime, $sessionParams['path'],
```
tsteur: interesting, wouldn't have thought this makes a difference. reading in docs eg for path

> path= The default value is the current directory that the cookie is being set in.

but maybe the other defaults from session cookie params were not applied. Or does this work better in case user has defined a different session path when the cookie is created the first time? 

I presume using the `Cookie` class here doesn't help much?

I'm seeing there is also `@Config::getInstance()->Tracker['cookie_path']` which maybe needs to be respected?

diosmosis: I think in the user's case the session cookie path was different from the default cookie path, but not completely sure.

> I presume using the Cookie class here doesn't help much?

I don't think so. This resets the session cookie, and the session cookie is originally set by PHP, not through Matomo. So using Cookie would be using different logic altogether and might have some strange side effects?

> I'm seeing there is also @Config::getInstance()->Tracker['cookie_path'] which maybe needs to be respected?

This is for the session and the tracker doesn't use the session? Ie, the tracker cookie_path is for piwik_ignore cookies right?

tsteur: > This is for the session and the tracker doesn't use the session? Ie, the tracker cookie_path is for piwik_ignore cookies right?

Yes, sorry wasn't looking right. So maybe it came from deprecating `login_cookie_path`? Does it make sense to apply it again? Or I see in `SessionInitalizer` it is still used. Getting bit confused re the paths and different usages etc.

diosmosis: it's used in the deprecated SessionInitializer that creates an auth cookie, not in the core one.

We could default the session cookie path to the value of login_cookie_path in `Piwik\Session`. That's probably a good idea, since multiple apps could use the same PHP INI config?

tsteur: 👍

core/Session.php
```diff
@@ -81,6 +81,8 @@ public static function start($options = false)
         // the session data won't be deleted until the cookie expires.
         @ini_set('session.gc_maxlifetime', $config->General['login_cookie_expire']);
 
+        @ini_set('session.cookie_path', empty($config->General['login_cookie_path']) ? '/' : $config->General['login_cookie_path']);
```
tsteur: 👍

PR #13874, opened 2018-12-18
Add generated unique request ID used in logs as custom HTTP header.

diosmosis:


tsteur:
haven't tested but looks good.

diosmosis:
@tsteur FYI, added the request ID to error reports as well (ie, via `$lastError`).

PR #13875, opened 2018-12-18
Use Request::processRequest in UserId::isUsedInSite.

diosmosis:


plugins/CoreHome/Columns/UserId.php
```diff
@@ -98,7 +98,12 @@ private function isUsedInSiteCached($idSite, $period, $date)
 
     private function isUsedInSite($idSite, $period, $date)
     {
-        $result = VisitsSummaryApi::getInstance()->get($idSite, $period, $date, false, 'nb_users');
+        $result = \Piwik\API\Request::processRequest('VisitsSummary.get', [
+            'columns' => 'nb_users',
+            'idSite' => $idSite,
+            'period' => $period,
+            'date' => $date,
```
tsteur: does segment need to be set to `false` or this automatically happens because of `$defaultRequest = []`? Never really sure what happens if the system maybe picks it up somehow... Maybe doesn't hurt to mention `segment => false`? Eg the segment string might be read from `$_SERVER['QUERY_STRING]` here https://github.com/matomo-org/matomo/blob/3.8.0-b5/core/API/Request.php#L512-L518 because of https://github.com/matomo-org/matomo/blob/3.8.0-b5/core/API/Request.php#L592-L594 etc

Otherwise looks good to merge 👍

diosmosis: Segments are handled very strangely in matomo.

tsteur: Yes!!! 100% strangely handled.

Issue #13882, opened 2018-12-19
Referrer Websites report should group same label (HTTP vs. HTTPS)


GreenReaper:
While reviewing visits from Tumblr, I expanded referrals and clicked "open row evolution" on the first sub-row:

![tumblrevolution](https://user-images.githubusercontent.com/383043/50219979-d8d84b00-0388-11e9-9fab-429bd32d8366.png)

To my surprise it showed a very sparse graph with only a few views each day. I found that whichever of the two rows I clicked evolution on, it was giving me the data for the second row. Hovering over the 'index' links to the left, I found that the first row was for referrals from the HTTPS url, and the second for HTTP.

The parameters encoded within the graph's request URL are:

> date=today
> apiMethod=Referrers.getWebsites
> label=t.umblr.com+>+@index
> disableLink=1
> module=CoreHome
> action=getRowEvolutionPopover
> colors={"backgroundColor":"#ffffff","lineColor":"#162c4a","minPointColor":"#ff7f7f","maxPointColor":"#75bf7c","lastPointColor":"#55aaff","fillColor":"#ffffff"}
> flat=0
> idSite=1
> period=day

These requests are the same if I click on either row evolution icon - which makes sense, the path just says "index". But when processing the request, it picked the less-used option - _not_ the option I selected. If it's going to separate them by scheme/URL, duplicating the label, it should use scheme/URL for the evolution.

"Metrics for Website" on the popup itself showed the website domain name as well as the path, but _not_ the access scheme. Perhaps if it had shown "http:" or "https:" as well, this issue would have been more visible.

There seems to be plenty of information within the row's <tr> to make this distinction, e.g.:
`<tr data-segment-filter="referrerUrl==https%3A%2F%2Ft.umblr.com%2F" data-url-label="https://t.umblr.com/" data-row-metadata="{&quot;segment&quot;:&quot;referrerUrl==https%3A%2F%2Ft.umblr.com%2F&quot;,&quot;url&quot;:&quot;https:\/\/t.umblr.com\/&quot;}" class=" ">`

In this _particular_ case it doesn't matter because I can just use the top-level graph which does work as it combines views from all paths (and here, there is only one path). However, in the general case it is a problem:

![multipleurl](https://user-images.githubusercontent.com/383043/50221426-fdcebd00-038c-11e9-8e4e-5352bd84451d.PNG) - HTTP or HTTPS? Can't tell at a glance. Row evolution for both picks the second.

[...]

PR #13884, opened 2018-12-19
Make sure to group the label in Referrer Websites report

tsteur:
I just debugged and can confirm behaviour described in #13882 
The label includes a URL like `http://demo.matomo.org/foo/bar` and the label is then removed in the `ColumnCallbackReplace` filter meaning there could be several entries with the same label and a group by is needed.


tsteur:
User confirmed it works, might cause the report to load a little less slow since the filters needed to be executed directly and the GroupBy is not the fastest either.

plugins/Referrers/DataTable/Filter/UrlsFromWebsiteId.php
```diff
@@ -31,10 +31,11 @@ public function filter($table)
     {
         // the htmlspecialchars_decode call is for BC for before 1.1
         // as the Referrer URL was previously encoded in the log tables, but is now recorded raw
-        $table->queueFilter('ColumnCallbackAddMetadata', array('label', 'url', function ($label) {
+        $table->filter('ColumnCallbackAddMetadata', array('label', 'url', function ($label) {
             return htmlspecialchars_decode($label);
         }));
-        $table->queueFilter('ColumnCallbackReplace', array('label', 'Piwik\Plugins\Referrers\getPathFromUrl'));
+        $table->filter('ColumnCallbackReplace', array('label', 'Piwik\Plugins\Referrers\getPathFromUrl'));
+        $table->filter('GroupBy', array('label'));
```
diosmosis: Just wondering, if the GroupBy is done before ColumnCallbackAddMetadata (w/ a reduce function specified as `Piwik\Plugins\Referrers\getPathFromUrl`), then maybe `ColumnCallbackAddMetadata` could b queueFilter and we could get rid of the extra `ColumnCallbackReplace`. Does that sound right to you?

diosmosis:
There's an interesting test failure w/ the Others row. Also probably good to have this bug covered by a system test.

PR #13886, opened 2018-12-19
Fix possible notice in safe mode handler

### LINKED ISSUE #13885: Notice undefined index type
> WARNING: www/piwik/core/FrontController.php(255): Notice - Undefined index: type - Matomo 3.8.0-b5 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)

Just updated to latest 3.x-dev and seeing this notice.

`$lastError` looks like this:
image

fyi @diosmosis

tsteur:
fyi @diosmosis  fix #13885

### DIFF
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -248,7 +248,7 @@ class FrontController extends Singleton
     {
         $lastError = error_get_last();
 
-        if (isset(self::$requestId)) {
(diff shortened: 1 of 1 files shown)


diosmosis:
Thanks for fixing

PR #13889, opened 2018-12-20
QR code library adds several new files to vendor dir that is breaking build 

### LINKED ISSUE #13794: QR code library adds several new files to vendor dir that is breaking build
See https://travis-ci.org/matomo-org/matomo/jobs/462646517

Many files like /vendor/endroid/qrcode/assets/data/qrv36_2.dat are added. Not sure if they are needed for qr code generation or not.

CC @mattab / @tsteur

tsteur:
Worked for me. Likely only need to adjust ui test.

fix https://github.com/matomo-org/matomo/issues/13794


tsteur:
Fixed UI tests

sgiehl (approved):
Works nicely. Tested on Chrome and IE11

PR #13891, opened 2018-12-20
Do not fetch totals for insights

tsteur:
Got this error warning

> WARNING: Trying to add two strings in DataTable\Row::sumRowArray: 35.8% + 48.6% for row # ['label' =&gt; -2, 'nb_visits' =&gt; 379, 'growth_percent' =&gt; '35.8%', 'growth_percent_numeric' =&gt; '35.8', 'grown' =&gt; 1, 'value_old' =&gt; 162, 'value_new' =&gt; 220, 'difference' =&gt; 58, 'importance' =&gt; 58, 'isDisappeared' =&gt; , 'isNew' =&gt; , 'isMover' =&gt; 1] [] [idsubtable = ]&lt;br /&gt;

The totals row should be only needed for HTML table visualisations. Bar and Pie, etc should not need this totals row and therefore disabled it by default so other visualisations don't need to disable it. Makes viewing reports faster...

### DIFF
--- a/core/ViewDataTable/Request.php
+++ b/core/ViewDataTable/Request.php
@@ -71,6 +71,7 @@ class Request
             'filter_column',
             'filter_pattern',
             'flat',
+            'totals',
             'expanded',
             'pivotBy',
             'pivotByColumn',
--- a/core/ViewDataTable/RequestConfig.php
+++ b/core/ViewDataTable/RequestConfig.php
@@ -156,7 +156,7 @@ class RequestConfig
      *
      * Default value: false
      */
-    public $totals = false;
+    public $totals = 0;
 
     /**
      * If set to true, the returned data will contain the first level results, as well as all sub-tables.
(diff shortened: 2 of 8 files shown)


tsteur:
Checked on our instance and the notice don't seem to come up anymore.

tsteur:
cheers @sgiehl  fixed it 👍  There was also an issue that when switching visualisation from HtmlTable to Insights it would still send `totals=1` in the URL.

And the `$totals = false` didn't work cause here it would add the parameter only when the default is not `false` : https://github.com/matomo-org/matomo/blob/3.8.0-b5/core/ViewDataTable/Request.php#L82-L83  Set it to `0` instead.

diosmosis:
Code seems to work, looks ok and build passes, will merge. @sgiehl if you want to review more, feel free to do so, will create a new issue or just apply a fix if everyone's on vacation.

PR #13894, opened 2018-12-21
Improve performance of delete query and make sure it doesn't lock existing tracking requests

### LINKED ISSUE #13800: deleteLogDataForDeletedSites locks database for a long period, bogus query?
The changes made in #13227 for deleteLogDataForDeletedSites seem to have caused a major performance issue. They result in a query that locks a large Piwik database for a long time, potentially several minutes. As far as I can see the queries will also never accomplish anything, since the where clause in the queries contains "site.idsite IS NULL" (created here: https://github.com/matomo-org/matomo/blob/70505b1ebb5584a8a859bcb9a54e0cb50d66677e/plugins/PrivacyManager/Model/DataSubjects.php#L408), and idsite in the site table or log_visit table cannot be null. The resulting queries look like this:

    DELETE log_conversion FROM piwik_log_conversion log_conversion LEFT JOIN piwik_log_visit log_visit ON log_conversion.idvisit = log_visit.idvisit  LEFT JOIN piwik_site site ON site.idsite = log_visit.idsite WHERE site.idsite IS NULL AND log_visit.idsite <= '3953'
    DELETE log_conversion_item FROM piwik_log_conversion_item log_conversion_item LEFT JOIN piwik_log_visit log_visit ON log_conversion_item.idvisit = log_visit.idvisit  LEFT JOIN piwik_site site ON site.idsite = log_visit.idsite WHERE site.idsite IS NULL AND log_visit.idsite <= '3953'
    DELETE log_link_visit_action FROM piwik_log_link_visit_action log_link_visit_action LEFT JOIN piwik_log_visit log_visit ON log_link_visit_action.idvisit = log_visit.idvisit  LEFT JOIN piwik_site site ON site.idsite = log_visit.idsite WHERE site.idsite IS NULL AND log_visit.idsite <= '3953'
    DELETE log_visit FROM piwik_log_visit log_visit LEFT JOIN piwik_site site ON site.idsite = log_visit.idsite WHERE site.idsite IS NULL AND log_visit.idsite <= '3953'

It seems that whatever these queries should be doing is not happening. Instead they're causing tracking to be blocked for a significant time. The query plan for the above queries shows how bad it is since it goes through the whole log_visit table:

    +------+-------------+-----------+--------+---------------------------------------------------------------------------+---------+---------+------------------------+----------+-------------------------+
    | id   | select_type | table     | type   | possible_keys                                                             | key     | key_len | ref                    | rows     | Extra                   |
    +------+-------------+-----------+--------+---------------------------------------------------------------------------+---------+---------+------------------------+----------+-------------------------+
    |    1 | SIMPLE      | log_visit | ALL    | index_idsite_config_datetime,index_idsite_datetime,index_idsite_idvisitor | NULL    | NULL    | NULL                   | 48810128 | Using where             |
    |    1 | SIMPLE      | site      | eq_ref | PRIMARY                                                                   | PRIMARY | 4       | piwik.log_visit.idsite |        1 | Using where; Not exists |
    +------+-------------+-----------+--------+---------------------------------------------------------------------------+---------+---------+------------------------+----------+-------------------------+

tsteur:
fix #13800 

I'm thinking this should improve the performance of the delete queries:

* Before it would run all the delete queries mentioned in #13800 every week even if no site was deleted
* Before it would potentially do a full table scan but now the `where idsite in(...)` should only lock rows belonging to that site. New tracking requests for those rows / idsite shouldn't come in cause the site no longer exists (would be discarded by tracker) so it shouldn't be a problem.

Initially I also implemented a check for each table that had `idsite` column to check if tracking requests for that site exists and only delete if the log table has tracking requests for that site, but it shouldn't be needed.

### DIFF
--- a/plugins/PrivacyManager/Model/DataSubjects.php
+++ b/plugins/PrivacyManager/Model/DataSubjects.php
@@ -31,22 +31,48 @@ class DataSubjects
         $this->logTablesProvider = $logTablesProvider;
     }
 
+    private function getDistinctIdSitesInTable($tableName, $maxIdSite)
+    {
+        $tableName = Common::prefixTable($tableName);
+        $idSitesLogTable = Db::fetchAll('SELECT DISTINCT idsite FROM ' . $tableName);
+        $idSitesLogTable = array_column($idSitesLogTable, 'idsite');
+        $idSitesLogTable = array_map('intval', $idSitesLogTable);
+        $idSitesLogTable = array_filter($idSitesLogTable, function ($idSite) use ($maxIdSite) {
+            return !empty($idSite) && $idSite = $maxIdSite;
+        });
+        return $idSitesLogTable;
+    }
+
     public function deleteDataSubjectsForDeletedSites($allExistingIdSites)
     {
         if (empty($allExistingIdSites)) {
             return array();
         }
 
+        $allExistingIdSites = array_map('intval', $allExistingIdSites);
+        $maxIdSite = max($allExistingIdSites);
+
         $results = [];
 
-        $logTables = $this->getLogTablesToDeleteFrom();
-        $deleteCounts = $this->deleteLogDataFrom($logTables, function ($tableToSelectFrom) use ($allExistingIdSites) {
-            return $this->getWhereToChooseVisitsForOtherSites($tableToSelectFrom, $allExistingIdSites);
-        }, function ($tableToSelectFrom) {
-            return 'LEFT JOIN ' . Common::prefixTable('site') . ' site ON site.idsite = ' . $tableToSelectFrom . '.idsite';
-        });
+        $idSitesLogVisit = $this->getDistinctIdSitesInTable('log_visit', $maxIdSite);
+        $idSitesLogVisitAction = $this->getDistinctIdSitesInTable('log_link_visit_action', $maxIdSite);
+        $idSitesLogConversion = $this->getDistinctIdSitesInTable('log_conversion', $maxIdSite);
+        $idSitesUsed = array_unique(array_merge($idSitesLogVisit, $idSitesLogVisitAction, $idSitesLogConversion));
 
-        $results = array_merge($results, $deleteCounts);
+        $idSitesNoLongerExisting = array_diff($idSitesUsed, $allExistingIdSites);
+
+        if (empty($idSitesNoLongerExisting)) {
+            // nothing to be deleted... if there is no entry for that table in log_visit or log_link_visit_action
+            // then there shouldn't be anything to be deleted in other tables either
+            return array();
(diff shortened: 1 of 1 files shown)


plugins/PrivacyManager/Model/DataSubjects.php
```diff
@@ -31,22 +31,47 @@ public function __construct(LogTablesProvider $logTablesProvider)
...
+        $allExistingIdSites = array_map('intval', $allExistingIdSites);
+        $maxIdSite = max($allExistingIdSites);
         $results = [];
 
+        $idSitesLogVisit = $this->getDistinctIdSitesInTable('log_visit', $maxIdSite);
+        $idSitesLogVisitAction = $this->getDistinctIdSitesInTable('log_link_visit_action', $maxIdSite);
+        $idSitesLogConversion = $this->getDistinctIdSitesInTable('log_conversion', $maxIdSite);
+        $idSitesUsed = array_unique(array_merge($idSitesLogVisit, $idSitesLogVisitAction, $idSitesLogConversion));
+
+        $idSitesNoLongerExisting = array_diff($idSitesUsed, $allExistingIdSites);
```
diosmosis: Possible edge case that I haven't thought through too much, but what if a site hasn't tracked any visits yet, and this runs. It will be treated as a deleted site, correct? If tracking is installed while this process is under way, will it be a problem?

tsteur: It shouldn't cause it would only return sites here that were tracked into and not exist in sites?

diosmosis: If sites has ids, [1,2,3,4,5] and sites that are tracked has ids [1,2,3], then the array_diff for $idSitesNoLongerExisting would be [4,5], correct?

diosmosis: Let me try and explain again (I might be seeing it wrong too). There are 5 sites total:

[1,2,3,4,5]

No site is deleted. Only three of the sites have tracked data, [1,2,3]. So when looking through log tables for idsites, we only get `[1,2,3]`.

In this situation, we will try to delete log data for [4,5], which would be ok since there isn't any log data, except possibly if tracking begins during the deletion process. Again, I'm not sure if this is a problem ultimately, just mentioning it.

tsteur: @diosmosis it would return `[]` not `[4,5]` see eg https://3v4l.org/Q4uci

array_diff returns all entries from the first array, that are not in the second array.

diosmosis: I see, no issue then. 👍

PR #13901, opened 2018-12-21
fix donate widget

Findus23:
- fixes the recurring donation widget (`on0` has to be `Piwik Supporter`)
- doesn't allow to modify the redirected URL
- removes `bb2_screener_` parameter. I am pretty sure it doesn't do anything.
- fix image breaking into the next line when selecting `$120/yr.`

### DIFF
--- a/plugins/CoreHome/Controller.php
+++ b/plugins/CoreHome/Controller.php
@@ -277,8 +277,21 @@ class Controller extends \Piwik\Plugin\Controller
                 unset($parameters[$name]);
             }
         }
+        $paypalParameters = [
+            "cmd" => "_s-xclick"
+        ];
+        if (empty($parameters["onetime"]) || $parameters["onetime"] != "true") {
+            $paypalParameters["hosted_button_id"] = "DVKLY73RS7JTE";
+            $paypalParameters["currency_code"] = "USD";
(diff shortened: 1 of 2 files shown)


plugins/CoreHome/Controller.php
```diff
@@ -277,8 +277,21 @@ public function redirectToPaypal()
                 unset($parameters[$name]);
             }
         }
+        $paypalParameters = [
+            "cmd" => "_s-xclick"
+        ];
+        if (empty($parameters["onetime"]) || $parameters["onetime"] != "true") {
+            $paypalParameters["hosted_button_id"] = "DVKLY73RS7JTE";
+            $paypalParameters["currency_code"] = "USD";
+            $paypalParameters["on0"] = "Piwik Supporter";
```
tsteur: do we need to change something in paypal setting to make it work for Matomo Supporter?

Findus23: Probably. Using `Matomo Supporter` just shows a Paypal error. I guess this is the name of the paypal "product".

tsteur: @mattab could you add this product to PayPal or configure it somehow?

Findus23:
As this has been merged, I guess we should move it from 3.9.0 to 3.8.0

PR #13904, opened 2018-12-21
improve RSS fetching

### LINKED ISSUE #5214: Feedburner behind the "wall"
By default all Piwik users can view the Piwik blog on their dashboard. This is a great feature that lets all users stay in touch. 

Feedburner.com is a web service that is practical, but is now owned by Google. It would be nice if we moved away from hosting our Piwik blog feed on Feedburner.

Suggested tasks
- Change blog feed URL to piwik.org/blog/feed/
- introduce server side caching of the feed in all piwik servers to minimise server load

Findus23:
Fixes #5214

While we are on the topic of improving old widgets:
- The Matomo RSS feed now includes a `.screen-reader-text` summary directly after `Read More`, which needs to be hidden
- Always use HTTPS
    - it's better to fail than show possibly mitm-ed HTML
- Don't use feedburner
    - this has the huge advantage of not sending data to Google
    - but this means many requests will reach wordpress, and the endpoint would probably need to be cached
    - speaking of caching: I think Matomo should also cache this data locally so that not every reload fetches the data freshly

Semi-related: https://matomo.org/faq/how-to/faq_99/ should also be removed, archived or updated.

BTW: Feedburner adds tracking-pixels to the feed, but only to the content, so Matomo isn't affected

### DIFF
--- a/plugins/RssWidget/RssRenderer.php
+++ b/plugins/RssWidget/RssRenderer.php
@@ -8,6 +8,7 @@
  */
 
 namespace Piwik\Plugins\RssWidget;
+use Piwik\Cache;
 use Piwik\Http;
 
 /**
@@ -19,10 +20,16 @@ class RssRenderer
     protected $count = 3;
     protected $showDescription = false;
     protected $showContent = false;
+    /**
+     * @var Cache\Lazy
+     */
+    private $cache;
+
 
     public function __construct($url)
     {
         $this->url = $url;
+        $this->cache = Cache::getLazyCache();
     }
 
     public function showDescription($bool)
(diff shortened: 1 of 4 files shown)


tsteur:
@Findus23 any chance we could not show the last part of each article saying "... appeared first on ..."?

image

Findus23:
@tsteur That was the reason why I initially looked into the rss fetching.
The sentence is part of the RSS feed, but maybe it is enough to always remove all text after `Read More`.

tsteur:
```css
#widgetRssWidgetrssPiwik .read-more-container+p {
    display: none;
}
```

something like this should do 👍

tsteur:
@Findus23 " It is unfortunately not in a special class"

do you mean that the sentence doesn't have a class? Cause the selector should still work. At least when I tested it ... or is the problem in case it changes in the future?

Findus23:
@tsteur 
What I meant, is that the HTML is structured like this:
```html
<p>There’s no doubt about it, the free version of Google Analytics offers great value when it comes to making
    data-driven decisions for your business. But as your business starts to grow, so does the need for a more powerful
    web ... </p>
<p class="read-more-container">
    <a title="Why Matomo is a serious alternative to Google Analytics 360"
       class="read-more button"
       href="https://matomo.org/blog/2018/12/why-matomo-is-a-serious-alternative-to-google-analytics-360/#more-30938">
        Read More
        <span class="screen-reader-text">
            Why Matomo is a serious alternative to Google Analytics 360
        </span>
    </a>
</p>
<p>The post
    <a rel="nofollow"
       href="https://matomo.org/blog/2018/12/why-matomo-is-a-serious-alternative-to-google-analytics-360/">
        Why Matomo is a serious alternative to Google Analytics 360</a> appeared first on
    <a rel="nofollow" href="https://matomo.org">Analytics
        Platform - Matomo</a>.
</p>
```

So by hiding `.screen-reader-text`, I can get rid of the screenreader summary, but not the last paragraph. So removing it via PHP (as I am doing now) is probably easier.

tsteur:
Did you try this?

```css
#widgetRssWidgetrssPiwik .read-more-container+p {
    display: none;
}
```

it should select it or am I not understanding it?

Findus23:
@tsteur That should also work. Do you think this is less prone to break in the future?

tsteur:
Not 100% sure which is better. Do we know the feed is always retrieved in English? If the wording changes it might break vs CSS will still work. Both have advantages and disadvantages really.

tsteur:
changed it @Findus23

Findus23:
@tsteur Thanks I removed the PHP code and it seems to work fine.
Can you please check if the Cache is implemented correctly and if https://matomo.org/feed is cached enough to be able to handle the new requests.

tsteur:
Not sure about that, probably not. I reckon we would need to put this behind our CDN but https://static.matomo.org/feed/ currently redirects to matomo.org and would need to figure out how to disable this redirect in wordpress.

diosmosis:
Noticed links in rss content didn't have `_target=blank` or `rel="noreferrer noopener"`, so added that. Looks good and works for me locally.

PR #13908, opened 2018-12-26
Make sure all Matomo emails use correct branding.

### LINKED ISSUE #13251: Make all emails use the pretty HTML emails with branding
After #13250 theme variables will be defined in a class and the LESS variables will be generated in an event. Before these APIs are made public the following things need to be determined/done:

* Non-report emails need to be themed this way.
* Every part of report emails need to use the correct theme variables.
* If there are redundant properties in `EmailStyles` and `ThemeStyles` they should be resolved.
* If it should be allowed, PDF reports should be themed the same way as well.

At the moment the styling of reports/emails isn't entirely consistent so can't be made public API.

diosmosis:
Fixes #13251


core/View/HtmlReportEmailHeaderView.php
```diff
@@ -77,7 +78,13 @@ public static function assignCommonParameters(View $view)
         $view->isCustomLogo = $customLogo->isEnabled() && CustomLogo::hasUserLogo();
         $view->logoHeader = $customLogo->getHeaderLogoUrl($pathOnly = false);
 
-        $view->hasWhiteLabel = \Piwik\Plugin\Manager::getInstance()->isPluginLoaded('WhiteLabel');
+        $view->hasWhiteLabel = \Piwik\Plugin\Manager::getInstance()->isPluginLoaded('WhiteLabel')
+            && \Piwik\Plugin\Manager::getInstance()->isPluginActivated('WhiteLabel')
+            && \Piwik\Plugin\Manager::getInstance()->isPluginInFilesystem('WhiteLabel');
+
+        $view->idSite = Common::getRequestVar('idSite', false);
```
tsteur: Not sure about those 3 lines. Are they needed @diosmosis ? For what? Cause the current set `idSite` in the request might not be related to any sent email. Same for period or date. Sometimes  there eg hooks that listen to something like a plugin update, and then send an email about something totally else.

tsteur:
Not sure if it is due to this PR but the sender is "Matomo Reports"... even when eg getting update notification 
image

tsteur:
Likely unrelated to this... noticed the bullet points are smaller in the scheduled reports than the rest of the font... didn't look in the code if it was before already like this

image

tsteur:
I suppose Custom alerts will need to be updated as well? Possibly also the mail send in QueuedTracking https://github.com/matomo-org/plugin-QueuedTracking/blob/3.3.1/Tasks.php#L73-L84 but not too important

core/View/HtmlReportEmailHeaderView.php
diosmosis: These three params are used in the dashboard link in the email header. Some emails had it, some didn't. Thought I'd add it to all if the params were available.

tsteur: Wouldn't we want to link the the default report?

diosmosis: This is how the code worked before, but I can change it to link to what's configured in settings.

diosmosis: Nvm, it already does this (by linking to CoreHome.index).

diosmosis:
> Not sure if it is due to this PR but the sender is "Matomo Reports"... even when eg getting update notification

Looks like that's the default report name if `[General] noreply_email_name` isn't set. Wonder if it should just be from `Matomo` or something?

core/View/HtmlReportEmailHeaderView.php
tsteur: Wouldn't it need to link to `CoreHome.redirectToCoreHomeIndex` (requires no site, period, date, ... AFAIK)? I guess the only time we wouldn't want this is when a report is sent for a specific site. In all cases might make sense to load the default report and default time to load.

diosmosis: Yes, would need to link to `redirectToCoreHomeIndex` or I guess to omit the action. I can make this change, except when a specific site is set.

diosmosis:
@tsteur re the bullet points, does this look better?

image

diosmosis:
Forgot to create the PR for CustomAlerts, can find it here: https://github.com/matomo-org/plugin-CustomAlerts/pull/54

diosmosis:
Created a queued tracking PR as well: https://github.com/matomo-org/plugin-QueuedTracking/pull/93

tsteur:
@diosmosis it should be fine as long as it is using the same font size 👍

diosmosis:
It is now, think it was using the default chrome font size before.

tsteur:
@diosmosis had a look at the other PRs and looks all good. I reckon can be merged as soon as tests succeed 👍

PR #13909, opened 2018-12-26
Add UI test for complex segment + fix couple bugs to get test to pass.

### LINKED ISSUE #11384: Add a UI test for a complex segment AND/OR/ and strange encoding
the UI test would reproduce #11372  #11321 /  #10973

 later when we will rewrite the segment editor and this will ensure we don't regress :+1:

diosmosis:
Fixes #11384 

Bugs fixed:
- When segment selector redirects to new segment, must make sure value is triple encoded to match 3 urldecodes done server side.
- Quotes and ampersands must be replaced globally throughout whole segment string, not just for first occurrence.


diosmosis:
@tsteur I wrote about it in this PR description: https://github.com/matomo-org/matomo/pull/13481

tsteur:
i see guess that makes sense. not sure re possible side effects.

diosmosis:
These changes should only affect the segment editor (ie, the segment definition that gets saved & what we redirect to). I suppose it might not recognize definitions for old segments that were created improperly... not too sure. Will test a bit more.

mattab:
@diosmosis did you maybe have a chance to do more tests?

c960657:
This introduced a new test failure. Fixed in #14021.

PR #13911, opened 2018-12-27
Use Request::processRequest instead of API instance directly.

diosmosis:


plugins/Live/Controller.php
```diff
@@ -97,9 +97,19 @@ public function getLastVisitsStart()
     private function setCounters($view)
     {
         $segment = Request::getRawSegmentFromRequest();
-        $last30min = API::getInstance()->getCounters($this->idSite, $lastMinutes = 30, $segment, array('visits', 'actions'));
+        $last30min = Request::processRequest('Live.getCounters', [
+            'idSite' => $this->idSite,
+            'lastMinutes' => 30,
+            'segment' => $segment,
+            'showColumns' => 'visits,actions',
```
tsteur: PR looks good. Just out of curiosity... passing an array for showColumns should do the same right?

diosmosis: Checked locally, and yep seems to work.

PR #13912, opened 2018-12-27
Add config setting to disable sending emails.

diosmosis:
Tiny change to allow users to make sure no emails are sent out from an instance. Useful if your testing w/ a DB dump that has other people's email addresses, and you don't want to send them spam from the test instance.


tsteur (approved):
Looks good. Haven't tested it though.

mattab:
fyi documented in https://matomo.org/faq/how-to/how-do-i-disable-all-automated-emails-so-that-matomo-doesnt-send-any-email/

PR #13914, opened 2018-12-27
update icons submodule

Findus23:
After a quite long time [the tests](https://travis-ci.org/matomo-org/matomo-icons) are now back to green as I added icons to the new device-detector brands and fixed imagemagick (it turns out `-floodfill` does not floodfill as one would expect if the selected pixel doesn't already have the to-be-filled color, but the obscure `-draw "color 0,0 floodfill"` does)

@sgiehl What was the reason for https://github.com/matomo-org/matomo/commit/a8afbc8b5b14fbaa20062d82440dcb2b77b173b0?
(It may be useful to wait with fixing the UI-tests until they succeed on 3.x-dev as thanks to imagemagick-patch-updates some single pixels change and break the ui-tests ([example](https://builds-artifacts.matomo.org/matomo-org/matomo/new-icons-submodule/32018/Theme_home.png)))


Findus23:
@sgiehl I noticed that one of the submodules changed the git history and the commit wasn't available anymore. I have since fixed it, so this should be solved.

PR #13916, opened 2018-12-27
Show actions in visitorlog as a (unnumerated) timeline

### LINKED ISSUE #13332: Improve the style of the visitor log
Currently, the visitor doesn't look so much appealing visually, especially the right part which shows all the actions. Plus it shows numbers in front of each action which is not relevant or helping.

image

Instead we could show the actions more in a "timeline" way also making it more clear the visitor log shows a timeline of all actions. This could be a vertical line on the left part of the list of actions, separated by an icon for each action. To the left of the icon, or on hover we would show instead when the action was performed. This gives a user a much better idea of how much time has passed between each action and lets you better understand the behaviour of a visitor. The time we show could be either the server time, the local time, or the time since the visit started. Because it would be confusing between local vs server time it may be best to show the time passed since visit start.

It could look a little bit like our [History page](https://matomo.org/history/) but instead of the red dot we show an icon depending on the action.

image

This could also work nicely with the new grouping feature we would want to have within the visitor log (https://github.com/matomo-org/matomo/issues/13136) as it would basically show another vertical timeline indented and it would be collapsible.

sgiehl:
first approach to improve the action listing layout:

image

I've only done some CSS changes by now. But we should also replace all icons with new ones, so they have the same size and look.

fixes #13332 
refs #12770


sgiehl:
I've tweaked the UI a bit and added some new icons from the materialize icon set. Now looks like this:
image

Did not spent too much time to search for proper icons by now, but currently I've added those:

image

sgiehl:
Also updated the folder icon used in realtime widget, and reused it for normal actions in visitor log as well:

image

diosmosis:
Code looks good, tested locally and the icons look a bit dark. Think it might look better if the icons are a bit lighter, to differentiate with text on the right and the "N Actions" label?

diosmosis:
@sgiehl can you take a look at my last comment?

sgiehl:
@diosmosis  I've lightened the icons a bit:
image

Feel free to adjust the colors a bit more. But actually all icons are currently only a suggestions. Maybe someone is keen on looking for some better ones?

diosmosis:
> Maybe someone is keen on looking for some better ones?

They honestly look quite good to me. Any thoughts @tsteur / @mattab (when you're back)?

I'd like to merge this soon, we can always change the icons later.

tsteur:
image
image

To me it looks a bit dark, and sometimes the icons are not aligned with the text. It be awesome if the first line was always centred vertically with the icon.

Also I think the vertical line and the icons shouldn't have the same color similar to https://user-images.githubusercontent.com/273120/44558721-d6c6fc80-a799-11e8-82e3-fc33f1890568.png

I feel like we may also need to tweak the spacing a bit but this can be done afterwards.
image

Eg it may be worth it to try always having things over two lines when possible so it looks bit less messy. But of course that's not always possible etc. We could ask a designer for some tweaks if needed.

diosmosis:
@sgiehl will you be able to address @tsteur's comment?

sgiehl:
I've updated the branch and adjusted the colors and alignment a bit more.

Note: Icons from our premium plugins (media / form) need to be adjusted to have the same color/size afterwards...

diosmosis:
Code looks good, though I think the icons could be even lighter, eg:

image

vs:

image

What do you thin @sgiehl, @tsteur? Will probably merge this PR as is in order to avoid going back forth forever, but can make a change in another PR if you guys agree.

tsteur:
Sound good @diosmosis and yes would make them even lighter.

diosmosis:
Noticed this gets rid of the goal conversion counter in a visit, eg:

image

(see the flag next to the cart)

Is this expected?

sgiehl:
I removed that because it looked quite ugly with the new icons and the border. But without the border the number looked a bit weird between the icons... Not sure if the goal count is that relevant (it's still shown on hover). Actually I don't have a good idea how to design that in a way it looks good

diosmosis:
I think it's ok to show it on hover... what do you think @mattab / @tsteur ?

tsteur:
Personally, I'm thinking it would be quite useful to still have the info directly visible how many goals were converted maybe in front of the icon. but it's also not too crazy important and on hover could be fine. not sure how often it would be really needed.

diosmosis:
What about a number in a small green circle at the bottom right of the icon? Like the number of notifications in some apps: 
image

tsteur:
could work if the icon and number is identifiable when it's shown so small 👍

tsteur:
Looks good 👍

diosmosis:
The visitor log looks much better w/ this PR!

PR #13921, opened 2018-12-30
Use current segment also in Segmented Visitorlog

### LINKED ISSUE #13344: Segmented visitor log does not inherit existing segment
When you view the dashboard segmented (E.g. Visitor location = US), then you click on a row to "Open segmented visitor log", the resulting visitor log is not segmented by the dashboard-level segment.

image

In the screenshot above, I would expect the visitor log to only show the visits that resulted in the 21 impressions.  However I actually receive a list of all visitors that generated impressions for that content (a much long list)

sgiehl:
this simple change should be enough so append the current (global) segment to the segment of the segmented visitor log.

fixes #13344


tsteur:
Just tested it and works for me. By the looks Funnels and Form Analytics seems to use `.show()` and already set the segment by the looks so should be all fine 👍  Not sure if used anywhere else.

PR #13922, opened 2018-12-30
replaces some remaining occurrences of Piwik with Matomo

sgiehl:
(reported an Transifex)


tsteur:
looks good if tests pass 👍

PR #13923, opened 2018-12-30
Enable fingers crossed handler via INI config and show backtrace in logs/archive api output

### LINKED ISSUE #8616: When an API request fails during archiving (core:archive), output the backtrace in the error message
When an API request fails during archiving (core:archive),  we currently display the plugin name that caused an error in the message. This helps finding out which plugin caused the failure.

In general, maybe we could make the error output even more verbose. Instead of just displaying the plugin name, could we also the full backtrace?

(afaik it used to work in this way few months ago, and was helpful in troubleshooting archiving issues.)

### LINKED ISSUE #8698: Enable FingersCrossedHandler logging handler to easier analyze errors
From https://github.com/Seldaek/monolog/blob/master/doc/02-handlers-formatters-processors.md#wrappers--special-handlers:

> FingersCrossedHandler: A very interesting wrapper. It takes a logger as parameter and will accumulate log records of all levels until a record exceeds the defined severity level. At which point it delivers all records, including those of lower severity, to the handler it wraps. This means that until an error actually happens you will not see anything in your logs, but when it happens you will have the full information, including debug and info records. This provides you with all the information you need, but only when you need it.

This will be very helpful to actually debug issues and to understand what happened as often an error message in a log standalone doesn't help much. It should be easy to do by just adding it to the configuration. In the future we might also want to add more log entries with `debug` or `trace` level (actually not sure if they support trace level). 

refs #8616

diosmosis:
Fixes #8698
Fixes #8616

Tested in the following contexts:
- web request (w/ warning/error all logs written to file log, only warning/error shows up in screen)
- normal console command (fingers crossed handler not used)
- core:archive top level command (fingers crossed handler not used)
- core:archive climulti processes (w/ warning/error all logs written to file log
- when a fatal error occurs

Also made following changes:
- log more exceptions
- make backtrace in logs include previous exceptions' backtraces
- don't use screen writer if in CLI mode (so notifications don't end up on the screen when we don't want them to). could also add a check for trigger=archivephp.
- always print backtrace if in CLI mode & trigger=archivephp

CC @tsteur not sure which milestone to put this in


config/global.ini.php
```diff
@@ -104,6 +104,14 @@
 ; if configured to log in a file, log entries will be made to this file
 logger_file_path = tmp/logs/piwik.log
 
+; if set to 1, will enable monolog's FingersCrossedHandler, which buffers all logs and flushes them if a warning or error
```
tsteur: I wonder if it actually needed to have an option for it or always enable it?

tsteur:
I'll put it into 3.9 for now. Can other plugins get all the logged messages during runtime? To send them eg by email

config/global.ini.php
diosmosis: ATM I'm not sure what effect this will have in production, thought it could use some more testing before it's enabled by default. I suppose it wouldn't be useful for most users anyway? Unless we create some sort of bug summary feature that grabbed these logs so users could upload these w/ error reports.

config/global.ini.php
```diff
@@ -104,6 +104,14 @@
...
+; log is seen. so the logs will contain debug logs, but only for requests where an error occurred. does not support the screen log writer
+enable_fingers_crossed_handler = 0
+
+; if enable_fingers_crossed_handler is set to 1, and this is set to 1, then the FingersCrossedHandler will stop buffering
+; on the first warning/error. if another one occurs later in the request, there will be no extra debug logs for them.
+fingers_crossed_stop_buffering_on_activation = 0
```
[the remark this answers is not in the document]
diosmosis: FingersCrossedHandler has an option where it stops buffering after the first error. I assume it's to avoid buffering too many logs in case there are lots of errors? I'm not exactly sure, but figured someone might want to set it true (by default I've left it disabled so every error triggers debug logs to be flushed).

diosmosis:
@tsteur added some more changes just now, might want to have another look.

> Can other plugins get all the logged messages during runtime?

Not at the moment. We could create a custom FingersCrossedHandler that posts an event when triggered. Or the plugin could grep the logs using the current request ID (that's what I was thinking at first).

config/global.ini.php
tsteur: I reckon it would always be useful to have it enabled so we get detailed logs when we ask users to enable it etc. Could also leave the option, test it for now on production, and follow up to remove the setting in 2 or 3 months and have it always enabled.

config/global.ini.php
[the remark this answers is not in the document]
tsteur: Same here I reckon we could keep the setting and see how it behaves in production and then possibly remove the setting and always have it disabled but to be seen after some more testing

tsteur:
when eg a fatal error occurs, we're sending an email during that request and looking through logs is not really an option as the log files could be many GB big. It would be great if there was some static method or some way to get the logger instance with all the log messages without using events (re performance).

diosmosis:
I'm not sure there is a way to do that currently. I think FingersCrossedHandler will abandon the saved logs once they are flushed to the internal handler, so they're not all retained. I guess we can try to extend/modify the handler.

Will we ever look into something like loggly/cloudwatch/etc.? If the logs were more easily accessible to us, it might not be needed to have them attached to an email (just a request ID would suffice).

tsteur:
We will likely look into CloudWatch eventually but we wouldn't send any data to a 3rd party. 

I thought there would be maybe some logging handler or writer that could be put in between maybe?

tsteur:
Also the goal is really to have all information directly in the fatal error email to have immediately all information.

diosmosis:
@tsteur Added the ability to get all logs in a request using a new handler. Also removed the INI config from global.ini.php and just check for them in code, so it's easier to remove if we find we don't need to include them later.

core/API/Request.php
```diff
@@ -269,7 +271,9 @@ public function process()
                 return $response->getResponse($returnedValue, $module, $method);
             });
         } catch (Exception $e) {
-            Log::debug($e);
+            StaticContainer::get(LoggerInterface::class)->error('Uncaught exception in API: {exception}', [
```
tsteur: I wonder if we should keep `debug` level when it is fetched through UI to not expose the specific error here? just not sure if there could be any problem such as sensitive information or something. Likely not a problem though.

core/ExceptionHandler.php
```diff
@@ -41,6 +44,8 @@ public static function handleException($exception)
      */
     public static function dieWithCliError($exception)
     {
+        self::logException($exception);
```
tsteur: Likely not too important but just realising that `console` is doing eg `Piwik\ExceptionHandler::setUp();` (and also `Piwik\ErrorHandler::registerErrorHandler();`) before the environment is loaded and container is built. So there could be under circumstances a problem where a message is maybe logged but the container is not ready yet. Not sure what would happen and likely the chances are maybe low there is an error between registering error handler and loading environment...

tsteur: actually stupid of me... I'm just seeing you catch an exception there anyway. Forget the comment :)

core/ExceptionHandler.php
```diff
@@ -137,4 +144,15 @@ private static function getErrorResponse($ex)
...
     }
+
+    private static function logException($exception)
+    {
+        try {
+            StaticContainer::get(LoggerInterface::class)->error('Uncaught exception: {exception}', [
+                'exception' => $exception,
+            ]);
+        } catch (ContainerException $ex) {
+            // ignore (occurs if exception is thrown when resolving DI entries)
```
tsteur: lets maybe also catch `ContainerDoesNotExistException` in case the container is not set up yet?

core/FrontController.php
```diff
@@ -107,6 +108,10 @@ private static function generateSafeModeOutputFromError($lastError)
      */
     private static function generateSafeModeOutputFromException($e)
     {
+        StaticContainer::get(LoggerInterface::class)->error('Uncaught exception: {exception}', [
```
tsteur: wouldn't that make the error visible in the UI?

core/testMinimumPhpVersion.php
```diff
@@ -84,6 +84,12 @@
      */
     function Piwik_ShouldPrintBackTraceWithMessage()
     {
+        if (\Piwik\SettingsServer::isArchivePhpTriggered()
```
tsteur: Not sure if backtrace should be included when triggered through the web? Would maybe rather not do it.

tsteur:
>  Also removed the INI config from global.ini.php and just check for them in code, so it's easier to remove if we find we don't need to include them later.

should we maybe create a follow up issue for 3.9 or 3.10 to enable it by default if our tests go well?

core/API/Request.php
diosmosis: If it's not debug, it won't trigger the fingers crossed handler, BUT I made sure the `screen` handler doesn't get wrapped in the fingers crossed handler. So I think this will result in an exception displaying on the screen for UI requests, but it should just print the exception message, which is the current behavior right? I'll make sure to see exactly what happens.

core/ExceptionHandler.php
diosmosis: 👍

core/FrontController.php
diosmosis: Wouldn't it always be visible in the UI? If it's not logged, there won't be information about it in the logs (and it won't trigger fingers crossed).

diosmosis:
> should we maybe create a follow up issue for 3.9 or 3.10 to enable it by default if our tests go well?

Makes sense, will do so once this PR is approved.

core/API/Request.php
tsteur: Was just not sure whether it could have any side effects re the exception message re security or whether it could be any annoying or cause any other problems. Wonder if it would even show during an API request etc or what would happen then? I wonder if we need to add tests for all those cases where this error is triggered.

Findus23:
> Will we ever look into something like loggly/cloudwatch/etc.? 
It would be amazing if there was an easy way to get all Matomo errors/warnings into sentry. 

I originally wanted to integrate Sentry into the Matomo monolog logger, but failed, so at the moment I am just using the default Sentry PHP error logger, which kind of works, but is incredibly hacky and probably doesn't show Matomo errors, just PHP errors and uncaught exceptions.
https://plugins.matomo.org/SentryLogger

Still using (a self-hosted) Sentry to keep track of Matomo errors has been incredibly helpful as I can keep track of all errors I come accross life and during development, can see if they occur regularyily or just once, have them automatically grouped and see stack traces including all variables that were set during the exceptions.

<details>
<summary>Screenshot</summary>

grafik
</details>

So if there is any way (for a plugin) to get all errors and warnings including stack traces in Matomo to send it to Sentry, it would be really helpful for me (and guessing by the fact that nearly 500 people have downloaded the really specific plugin, probably others too).

Having better error backtraces sounds also really useful for debugging odd issues on the forum.

core/API/Request.php
tsteur: I meant mainly to make sure that the API output is still correct when an error occurs during an API request. Might be even good for one UI test to see what it shows there. Wouldn't need to check re what is logged to the file.

diosmosis:
> So if there is any way (for a plugin) to get all errors and warnings including stack traces in Matomo to send it to Sentry, it would be really helpful for me (and guessing by the fact that nearly 500 people have downloaded the really specific plugin, probably others too).

@Findus23 Just read this, you could try creating a custom log handler. Exceptions will be in the `'exception'` property in the context (see eg how `ExceptionToTextProcessor` uses the property). I think you could use `\DI\add` or `\DI\decorate` w/ `'log.handlers'` in your plugin to add it. You won't be able to catch exceptions that aren't logged, but we should probably be logging everything.

core/API/Request.php
diosmosis: > I meant mainly to make sure that the API output is still correct when an error occurs during an API request.

There are currently plenty of API tests that trigger errors, so think this one is already good.

> Might be even good for one UI test to see what it shows there. 

Will try to include a UI test. I guess the existing safemode one is good for the fatal error check.

diosmosis: Created a UI test and noticed that the error will also appear as a notification. Not sure what the best way to handle this is, since w/o the log statement, there ends up being no record of the error in the file.

My personal opinion is that there really shouldn't be a screen logger, and everything should go into a log that only admins/devs can see. But I'm not sure if that change can be made.

tsteur: That's what I meant with changing the log level from debug to error. The screen logger is definitely useful in that it helps us finding bugs/issues etc. where otherwise 98% of users would never look in the logs and not even know they exist etc. 

The only thing I can think of be to change the log level depending on whether we are CLI or not. Or also enable the fingercrossedhandler on cli based on some other flags maybe. Can we maybe log it as `debug` and somehow pass some option to also attach all previous logs when this happens? Defeats the idea of the fingercrossedhandler maybe a little but could work maybe?

tsteur: Sure it should be fine this way as well 👍

diosmosis: Made the change + added some more logs. Also modified ExceptionToTextProcessor to replace `{exception}` in the message instead of the whole message.

core/API/Proxy.php
```diff
@@ -24,15 +25,13 @@
...
- * @method static Proxy getInstance()
  */
-class Proxy extends Singleton
+class Proxy
 {
     // array of already registered plugins names
     protected $alreadyRegistered = array();
 
-    private $metadataArray = array();
+    protected $metadataArray = array();
```
tsteur: not 100% sure if this was meant to be in the PR? Seeing there were couple tests for this but not sure if it was meant for different PR maybe?

diosmosis: It's used by the new UI test, see `CustomApiProxy` in UITestFixture.php.

diosmosis:
@tsteur any issue w/ merging this PR? could help on cloud

tsteur:
Nope

mattab:
>>  should we maybe create a follow up issue for 3.9 or 3.10 to enable it by default if our tests go well?
> Makes sense, will do so once this PR is approved.

@diosmosis have you created issue or could you do it? Be great to enable it eventually, once we know it works well..

diosmosis:
@mattab created an issue in 4.0: https://github.com/matomo-org/matomo/issues/15220

Issue #5309, opened 2014-06-05
SegmentEditor: display readable values for some metrics (e.g. FF -> Firefox)


peterbo:
When the segmenteditor is used, some select values are very technical / non user-friendly. These Values should be mapped / labeled to make them more meaningful.

For example this would make sense for:
- Browser - expects values like (FF, IE, MF, CM, IW, etc.) - This should be more readable (Firefox, Internet Explorer,??,??,??)
- Visit Location (Continent, Country)
- Operating system
- possibly others where it makes sense

hpvd:
uuuh this is still open (for mor than 3years), major flag seems not to help at all...
It's strongly related to
- Improve geographical segmentations #11261

tsteur:
FYI: This will be easier to implement with https://github.com/piwik/piwik/pull/12040 since we will then have all the information on how to render a value eg for country "nz".

hpvd:
wow that looks pretty awesome.
Many thanks for working on this!

mattab:
is this issue now fixed maybe @sgiehl ?

sgiehl:
@mattab not sure if there are some segments missing where it would make sense to have readable names as well. But it's implemented for browsers, operating system and country segments. For those there is now the code segment and a new name segment.
Guess we can close the issue and create a new one if someone sees a segment where readable value would be usable?

tsteur:
👍

PR #13929, opened 2019-01-03
Readable segment values for browser, os and country segments

### LINKED ISSUE #6772: DeviceDetection segment editor auto suggested values should show full strings
The goal of this issue is to review the Device Detection segments such as `browserCode` and `operatingSystemCode`.
- Currently the valid segment values  for `browserCode` are `FF`, `CH`, `IE`, etc.
  - would be more understandable and guessage to use `firefox`, `chrome`, `internet explorer` etc.
  - the segment auto suggest values should contain automatically all the top browsers we know of, instead of using the `Live` auto suggested feature. 
  - for Backward compatibility we should still accept `FF`, `CH` etc. and define `firefox` or `Chrome` as aliases to those ones.
- Same applies to `operatingSystemCode` where segment values would be pretty string such as `windows xp`, `windows 7`, `GNU/Linux`
  - it's case insensitive, both `Windows XP` and `windows xp` will alias to `WXP`

Implementation: refs `suggestedValuesCallback` feature.

sgiehl:
Current segments for browser, os or country are only usable using the short code (e.g. FF, IE, WIN, IOS, de, fr, ...).

refs #5309
fixes #6772
refs #14053


plugins/DevicesDetection/Columns/BrowserName.php
```diff
@@ -17,12 +18,23 @@ class BrowserName extends Base
 {
     protected $columnName = 'config_browser_name';
     protected $columnType = 'VARCHAR(10) NULL';
-    protected $segmentName = 'browserCode';
+    protected $segmentName = 'browserName';
```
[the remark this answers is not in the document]
sgiehl: Sure. There will be two segments each `browserName` and `browserCode` and same for os.

sgiehl:
Expected test files may need an update before merging

sgiehl:
I've updated the branch and added an additional segment for the english country name

plugins/DevicesDetection/Columns/BrowserName.php
```diff
@@ -17,12 +19,42 @@ class BrowserName extends Base
...
     protected $namePlural = 'DevicesDetection_Browsers';
-    protected $acceptValues = 'FF, IE, CH, SF, OP, etc.';
+    protected $acceptValues = 'FireFox, Internet Explorer, Chrome, Safari, Opera etc.';
     protected $type = self::TYPE_TEXT;
 
+    public function __construct()
+    {
+        $this->sqlFilterValue = function ($val) {
+            $browsers = Browser::getAvailableBrowsers();
+            $result   = array_search($val, $browsers);
```
diosmosis: Would it be a good idea if we made the segment case insensitive?

sgiehl: When choosing a suggest value that shouldn't matter. But shouldn't hurt to compare them lower case. Will change them

sgiehl:
@diosmosis browser and os names should now be compared case insensitive

diosmosis:
I'm seeing this change in the tests:

```
-    <segment>countryCode==gb</segment>
+    <segment>countryName==gb</segment>
```

Is that correct? Reporting APIs may need to be changed possibly...

sgiehl:
@diosmosis no, I've updated the code, so that doesn't change. Tests should also be updated now.

PR #13931, opened 2019-01-04
Avoid browser tooltips for piwik-fields

### LINKED ISSUE #13783: title-text of JavaScript Tracking option help box shows HTML
image

This seems to be caused by 
https://github.com/matomo-org/matomo/blob/e11866e11ddad88dbf7aee5f56431c8958acdb02/plugins/CoreAdminHome/templates/trackingCodeGenerator.twig#L61

and 
https://github.com/matomo-org/matomo/blob/e11866e11ddad88dbf7aee5f56431c8958acdb02/plugins/CoreAdminHome/angularjs/trackingcode/jstrackingcode.controller.js#L156

which of course doesn't work.

sgiehl:
by using data-title attribute instead of title

refs #13783

### DIFF
--- a/plugins/CoreAdminHome/templates/generalSettings.twig
+++ b/plugins/CoreAdminHome/templates/generalSettings.twig
@@ -82,7 +82,7 @@
         <div piwik-form ng-controller="MailSmtpController as mailSettings">
             <div piwik-field uicontrol="checkbox" name="mailUseSmtp"
                  ng-model="mailSettings.enabled"
-                 title="{{ 'General_UseSMTPServerForEmail'|translate|e('html_attr') }}"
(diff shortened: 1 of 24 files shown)


tsteur:
Haven't tested but looks good if tests pass 👍

diosmosis:
Plugins will have to be modified as well, correct? Would it be possible to remove the title attribute after the value is used in the piwik-field directive? This would make it work automatically for plugins, though I'm not sure if such a solution is possible.

tsteur:
Cheers @sgiehl 👍 
I'll merge this for now and leave a comment in https://github.com/matomo-org/matomo/issues/13783 that it still needs to be done for plugins.

PR #13932, opened 2019-01-05
Fixes some problems in export overlay

### LINKED ISSUE #13059: Hierarchical page export: label wrongly encoded
Matomo version: 3.5.0, MySQL version: 10.1.21-MariaDB, PHP version: 7.1.1

Hello, we encoutered a new issue since we updated to Matomo 3.5.0 : when we try to generate a report of a selected page, we get an empty report.
After some investigation, it seems that the parameter “label” is not well decoded in the query.
By default, the query contains : “label=sg%2520%2526gt%253B%2520pages%2520%2526gt%253B%2520%2540%25252Findex.aspx” and it doesn’t work.
If I manually edit the query in the URL with :
“label=sg%20%26gt%3B%20pages%20%26gt%3B%20%40%252Findex.aspx” then it does work. However, I cannot tell all my customers to use this workaround.

It can be reproduced here https://demo.matomo.org/index.php?module=CoreHome&action=index&idSite=7&period=day&date=yesterday&updated=1#?idSite=7&period=day&date=yesterday&category=General_Actions&subcategory=General_Pages&popover=

Select a page under a hierarchy (ex: top > /all) and generate a XML report for it, the query will be:
https://demo.matomo.org/index.php?date=2018-05-14,2018-06-12&expanded=1&filter_limit=100&format=XML&format_metrics=1&idSite=7&label=top%20%26gt%3B%20%40%252Fall&method=Actions.getPageUrls&module=API&period=day&token_auth=anonymous
And it’s empty.

### LINKED ISSUE #13160: Hierarchical reports exports empty datasets on leafs
When selecting "Row evolution" of a hierarchical report (IE download report) The CSV export function works on groups but not on "leafs" of the hierarchy. Specifically it exports an empty dataset.

Also reported on the forum by another user with screenshots here:

https://forum.matomo.org/t/report-has-data-but-export-is-empty/28094

sgiehl:
fixes incorrect detection of flat param and possible double encoding of label param

fixes #13160 
fixes #13059

### DIFF
--- a/plugins/CoreHome/angularjs/report-export/reportexport.directive.js
+++ b/plugins/CoreHome/angularjs/report-export/reportexport.directive.js
@@ -159,7 +159,7 @@
                         if (label.length > 1) {
                             exportUrlParams.label = label;
                         } else {
-                            exportUrlParams.label = encodeURIComponent(label[0]);
+                            exportUrlParams.label = label[0];
                         }
                     }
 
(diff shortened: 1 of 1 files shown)


plugins/CoreHome/angularjs/report-export/reportexport.directive.js
```diff
@@ -185,10 +185,10 @@
                     scope.reportType          = 'default';
                     scope.reportLimit         = dataTable.param.filter_limit > 0 ? dataTable.param.filter_limit : 100;
                     scope.reportLimitAll      = dataTable.param.filter_limit == -1 ? 'yes' : 'no';
-                    scope.optionFlat          = dataTable.param.flat;
+                    scope.optionFlat          = dataTable.param.flat === true || dataTable.param.flat === 1 || dataTable.param.flat === "1";
```
sgiehl: in some cases `dataTable.param.flat` is set to (string) `0`, which validated to true before, causing a `flat=1` in the export url where it shouldn't be

tsteur:
haven't tested it but looks good

PR #13933, opened 2019-01-06
Hide mainurl in visitor log action list

### LINKED ISSUE #13359: In visitor log do not show the domain for pageviews
See 
image

The domain is in most cases not needed and visible on hover anyway. This de-clutters the visitor log and makes it more clear which pages a user viewed. Also when the domain is really long, parts of the URL are otherwise not visible.

In case this is a problem to remove it, could be a config or better a check if multiple URLs are configured for a site and show it depending on that for example.

refs https://github.com/matomo-org/matomo/issues/13332

sgiehl:
fixes #13359

### DIFF
--- a/plugins/Live/VisitorDetails.php
+++ b/plugins/Live/VisitorDetails.php
@@ -88,7 +88,11 @@ class VisitorDetails extends VisitorDetailsAbstract
             return;
         }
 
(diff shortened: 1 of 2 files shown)


sgiehl:
main url is now only removed if no additional url is defined. Full url should still be shown in the action tooltip

mattab:
LGTM

Issue #11071, opened 2016-12-22
security 'forget your password' prone to phishing attacks


RMastop:
The 'lost your password' is vulnerable to fishing attacks.

Usecase:
An attacker tries to reset the password of an account, the email gets send out. The attacker needs to check if they can login using this new password. (either by a user clicking the link in the email, or even worse, an automated email vulnerability scanner checking the URL in the email)

To prevent this from succeeding, the following can be done. 
The confirmResetPassword class (in Login/controller.php) would need to have an extra check; 
After the validation of the resetToken, the user needs to be asked to retype the previous chosen password (used in the initial change your password screen.)
Only after the successful confirmation of this password, the account password needs to be reset.

I'm not a PHP programmer, would this be easy to implement?

KaanErturk:
Asking the new password in the very first page that can be accessed and submitted by anyone is kind of scary. I think it should only ask an email address (not even a username as it can't be changed) and the resulting page should have a generic message without verifying whether a user for that email address exists or not.

mattab:
Thanks for the report! Indeed our "Reset password" functionality would be better by having a reset link sent to the email to let the user then reset the password. I think it used to work in this way but i can't remember why we changed it.

> or even worse, an automated email vulnerability scanner checking the URL in the email

@RMastop I had not heard that such things exist. Do you know any tool that does this? Clicking automatically in email link is a very broken thing in general and should not be done by any tool IMHO (but my opinion does not matter in this case, agreed)

Findus23:
I think this is quite an important issue as it makes taking over an account too easy if the Matomo user is not careful with clicking on emails (which everyone should be, but no one is all the time).

I agree with @mattab that such a tool would be quite insane and will cause tons of damage with badly written websites, but Matomo shouldn't be one of them :slightly_smiling_face: and fundamentally a GET request should never do an irrevocable action like granting an attacker full access to a Matomo instance. 
So at the very least there should be a confirmation page after clicking on the link (correct me if there is already one, I didn't double-check), but I think rewriting this feature to work as on every other website (link allows user to set a new password) will cause less confusion, more security and (hopefully) shouldn't be that much more work.

(feel free to move it to a later release if you disagree)

tsteur:
3.9.0 is already pretty full, earliest we would do this maybe is 3.10 but to be seen. Moving it into 3.10 for now.

mattab:
[...]

PR #13934, opened 2019-01-06
Implements a rate limit for password resets

### LINKED ISSUE #13813: rate limit password resets
See also https://github.com/matomo-org/matomo/pull/13472#issuecomment-443685080
followup to #13472 and #2888 

Currently everyone can request an unlimited number of password requests which causes an unlimited amount of password reset emails which causes a mess in the inbox, overloads mailservers and may make it possible to let an attacker trick the user in accepting this request (https://github.com/matomo-org/matomo/issues/11071)

sgiehl:
currently only prevents from requesting new password resets within one hour. Maybe we could even increase the time limit.

fixes #13813

### DIFF
--- a/plugins/Login/PasswordResetter.php
+++ b/plugins/Login/PasswordResetter.php
@@ -462,7 +462,7 @@ class PasswordResetter
      */
     private function savePasswordResetInfo($login, $newPassword, $keySuffix)
     {
-        $optionName = $this->getPasswordResetInfoOptionName($login);
+        $optionName = self::getPasswordResetInfoOptionName($login);
 
         $existingResetInfo = Option::get($optionName);
 
--- a/plugins/Login/PasswordResetter.php
+++ b/plugins/Login/PasswordResetter.php
@@ -466,18 +466,28 @@ class PasswordResetter
 
         $existingResetInfo = Option::get($optionName);
 
+        $time = time();
+        $count = 0;
+
         if ($existingResetInfo) {
             $existingResetInfo = json_decode($existingResetInfo, true);
 
             if (isset($existingResetInfo['timestamp']) && $existingResetInfo['timestamp'] > time()-3600) {
-                throw new Exception(Piwik::translate('Login_PasswordResetAlreadySent'));
+                $time = $existingResetInfo['timestamp'];
+                $count = !empty($existingResetInfo['requests']) ? $existingResetInfo['requests'] : $count;
+
+                if(isset($existingResetInfo['requests']) && $existingResetInfo['requests'] > 2) {
+                    throw new Exception(Piwik::translate('Login_PasswordResetAlreadySent'));
+                }
             }
         }
+
 
         $optionData = [
             'hash' => $this->passwordHelper->hash(UsersManager::getPasswordHash($newPassword)),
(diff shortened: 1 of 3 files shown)


fdellwing:
I would maybe change the limit to 2 per hour? There are valid reasons to try it a second time.

sgiehl:
Are there? If sending the email fails on Matomo side the request get's removed again. So a new request can be done until the email was sent.

sgiehl:
Then it would make sense to allow resending the last request once instead of allowing a complete new one...

fdellwing:
I currently don't see a difference, but that seems fine for me.

sgiehl:
Requesting a new reset means filling out the form again and generating a new token for the reset. That means the email content is slightly another as the link differs

fdellwing:
Ok, it's fine for me to send the exact same mail again.

tsteur:
I would have also allowed 2 or 3 per hour as it can be useful eg if mailbox is full etc but if it otherwise works within an hour again should be fine. Could have been also useful to block per IP but not needed (to avoid someone requesting forget password feature every hour for all users and basically blocking the feature, not really an issue though)

sgiehl:
@tsteur Wondering what would be the userfriendliest and securest implementation.
- Allowing to request a password reset only once an hour (e.g. an error will be shown when the last request was within the last 60 mins)
- Limit the number of requests possible per day (or maybe month), without any "wait time" between the requests
- Allowing to request a password reset only every 3-x hours, but show a message when another request is made within that time, that allows to trigger a resend of the last request (new token, but still the same password from the original request). Resending the request would also be limited to 2 or 3 times and afterwards an error would be shown to contact the admin

tsteur:
Ideally it would use the new brute force feature added in 3.8.0. We would add a new `type` column to the DB and methods to be able to take advantage of all the methods etc. 

But it's a bit more work and #13813 is not even scheduled yet and not really important so would be too much work for now. I would probably simply allow 3 calls per hour for now or whatever is easiest.

sgiehl:
@tsteur guess make sense to reuse that. This was just a quick approach to implement a simple rate limit. Maybe we should simply close this and implement it properly later?

tsteur:
No preference. Could also merge and just leave it like that. up to @Findus23 and @mattab

Findus23:
This PR fixes the most important issue (you can endlessly spam users), so I would say this is good for now. And I wouldn't create a too complex logic as I still think that one day the password reset should be replaced with the "normal" way (#11071) and then there wouldn't be a difference between resending a reset and sending a new reset.

mattab:
just a quick feedback: reckon it should be possible to request again the password reset in the hour. For example if the mailserver was buggy and email didn't get sent, or any other problem. 3 times per hour sounds good and would not be considered spammy?

tsteur:
@sgiehl I set it back to `WIP`. Lets allow 3 times per hour.

sgiehl:
Changed it so only 3 requests are allowed per hour.

tsteur:
Any chance to create an integration test for this @sgiehl ?

sgiehl:
I've added some tests

tsteur (approved):

PR #13935, opened 2019-01-06
Couple assorted changes

diosmosis:
Changes:
* Allow annotations API to accept multiple periods (ie, date = 2018-02-01,2018-02-24 / period = day) so it will work when evolution graphs use those kinds of params.
* Remove warning that is emitted when rows_to_display is left at false.
* Allow individual cells in an HTML table based visualization to be styled.
* Remove a TODO that isn't needed.


sgiehl (commented):
Guess the failing Integration test is unrelated to this PR. 
Rebased the branch on latest `3.x-dev`. Should be good to merge if tests pass now

diosmosis:
Actually shouldn't merge just yet, forgot to mention the latest change will require changes to premium plugins, since the `'rows'` query param appears to be handled only in plugins not in core.

tsteur:
> Actually shouldn't merge just yet, forgot to mention the latest change will require changes to premium plugins, since the 'rows' query param appears to be handled only in plugins not in core.

Haven't looked at code or anything... will it be possible to keep BC? Or at least update plugins to be compatible with all Matomo versions?

core/Tracker/Model.php
```diff
@@ -357,23 +357,26 @@ public function findVisitor($idSite, $configId, $idVisitor, $fieldsToRead, $shou
         // 		this page view to the wrong visitor, but this is better than creating artificial visits.
         // 2) there is a visitor ID and we trust it (config setting trust_visitors_cookies, OR it was set using &cid= in tracking API),
         //      and in these cases, we force to look up this visitor id
-        $whereCommon = "visit_last_action_time >= ? AND visit_last_action_time <= ? AND idsite = ?";
-        $bindSql = array(
+        $configIdWhere = "visit_last_action_time >= ? AND visit_last_action_time <= ? AND idsite = ?";
+        $configIdbindSql = array(
```
mattab: One of the reason we added the 30min lookback window WHERE expression was to reduce performance problems: IIRC large scale Matomo instances have performance problems with selecting visitors in the past without WHERE visit_last_action_time limitations... (even though in theory the index should make it fast, I don't know why it wasn't the case before.)  some of these SELECT queries were slow and ended up piling up and tracking didn't work anymore. 

-> Maybe this problem would now be fixed (eg. a Mysql performance improvement etc.), but we need need to test the new code on a very large instance (under high traffic load, eg. 100 tracking api requests per second, without QueuedTracking) to check whether it works fast.

(If we merged this without testing performance it would possibly result in tracker performance regressions which is risky and high impact on users)

plugins/Annotations/API.php
```diff
@@ -334,7 +334,9 @@ public static function getDateRangeForPeriod($date, $period, $lastN = false)
             $endDate = $oPeriod->getDateEnd();
         } else // if the range includes the last N periods
         {
-            list($date, $lastN) = EvolutionViz::getDateRangeAndLastN($period, $date, $lastN);
+            if (!Range::isMultiplePeriod($date, $period)) {
```
mattab: Maybe a test can be added for this new Annotations API feature?

plugins/CoreHome/Columns/VisitorDaysSinceFirst.php
```diff
@@ -13,6 +13,7 @@
 use Piwik\Tracker\Request;
 use Piwik\Tracker\Visitor;
 
+// TODO: it would be better to save the time duration since the first visit rather than just the number of days, since that cuts of information. make matomo 4 issue?
```
mattab: Sure, be good to create an issue, and we shall see if/when to schedule it later :+1:

plugins/CoreHome/Columns/VisitorDaysSinceFirst.php
```diff
@@ -29,6 +30,11 @@ class VisitorDaysSinceFirst extends VisitDimension
      */
     public function onNewVisit(Request $request, Visitor $visitor, $action)
     {
+        // if the visitor is known, force days since first to 0, to ignore any potential bad values for _idts
```
mattab: should be "if the visitor is **new**, force days since first to 0"

plugins/CoreVisualizations/templates/_dataTableViz_htmlTable.twig
```diff
@@ -38,7 +38,8 @@
                     class="{{ row.getMetadata('css_class') }} {% if rowHasSubtable %}subDataTable{% endif %}{% if shouldHighlightRow %} highlight{% endif %}{% if isSummaryRow %} summaryRow{% endif %}"
                     {% if rowHasSubtable %}title="{{ 'CoreHome_ClickRowToExpandOrContract'|translate }}"{% endif %}>
                     {% for column in properties.columns_to_display %}
-                        <td>
+                        {% set cellAttributes = visualization.getCellHtmlAttributes(row, column) %}
+                        <td {% if cellAttributes is not empty %}{% for name, value in cellAttributes %}{{ name }}="{{ value|e('html_attr') }}" {% endfor %}{% endif %}>
```
mattab: maybe `{{ name }}` should also be encoded like the value is?

mattab (changes requested):

core/Tracker/Model.php
diosmosis: @tsteur ran the queries on a large instance and seemed to be ok. Will let him follow up.

tsteur: MySQL shouldn't even use that index re time and I would suspect it makes it rather slower but might be wrong. 

We were wondering in chat whether there are any privacy problems with this eg around 13 months etc.

plugins/Annotations/API.php
diosmosis: Added

core/Tracker/Model.php
mattab: As far as I see there is no privacy implications. (as long as there is a matching Visitor ID in the past, then the visitor ID will be the same in both visits and this PR actually makes data more accurate)

Todo:

* we just need to update one sentence in this FAQ that mentions the look back window https://matomo.org/faq/general/faq_21418/
* and also the comment of INI setting `window_look_back_for_visitor` 

to explain that for "Visitor ID" matching we look back in the whole history while for the Fingerprint we only look back window_look_back_for_visitor ie. 30min

Now I see that there is likely no performance regression here so this change should be fine :+1:

plugins/CoreHome/Columns/VisitorDaysSinceFirst.php
diosmosis: Created issue https://github.com/matomo-org/matomo/issues/14094

core/Tracker/Model.php
diosmosis: Made the change to INI config. Suggested faq change:

```
In all cases, when “looking for visits” in the past, we look back window_look_back_for_visitor seconds. By default, Matomo (Piwik) will look back 30 minutes (default value for window_look_back_for_visitor INI setting). This setting can be changed: learn more.
```

to

```
When matching visits by visitor/user ID, Matomo looks through the entire history of visits that are currently in the database. When matching visits by device fingerprint (config_id), we only look back window_look_back_for_visitor seconds. By default, this value is set to 30 minutes. This setting can be changed: learn more.
```

mattab: Thanks. i've already updated FAQ so we don't forget it later: https://matomo.org/faq/general/faq_21418/

mattab:
> I'm not to sure about BC.

So is the best next step to release new versions of all premium plugins which use this feature (how many plugins are affected, both free/premium?) to be compatible with this latest Matomo/patch?

diosmosis:
@tsteur  finally got the tests to pass, I think this could use another review. One big change: turns out the code always copied the last log's properties, even if the last log was part of a different visit. Since we only looked back the visit standard length time, this wasn't an issue, but now since we can look further back, I had to modify the VisitRecognizer code more.

core/Tracker/GoalManager.php
```diff
@@ -742,6 +742,7 @@ protected function insertNewConversion($conversion, $visitInformation, Request $
         Common::printDebug($newGoalDebug);
 
         $wasInserted = $this->getModel()->createConversion($conversion);
+        // TODO: if idsite/idorder are taken, but idvisit is different, this will silently fail. can we issue an error if possible?
```
tsteur: Issuing an error sounds good and useful to me 👍 Otherwise you have no chance to realise it failed.

plugins/CoreConsole/tests/System/ArchiveCronTest.php
```diff
@@ -64,14 +64,14 @@ public function getApiForTesting()
...
-        foreach($segments as $segment) {
+        foreach($segments as $index => $segment) {
             // Test with a pre-processed segment
             $results[] = array(array('VisitsSummary.get', 'Live.getLastVisitsDetails', 'VisitFrequency.get'),
                                array('idSite'     => '1',
                                      'date'       => '2012-08-09',
                                      'periods'    => array('day', 'year'),
                                      'segment'    => $segment,
-                                     'testSuffix' => '_preArchivedSegment',
+                                     'testSuffix' => '_preArchivedSegment' . $index,
```
tsteur: good find :)

plugins/CoreHome/Tracker/VisitRequestProcessor.php
```diff
@@ -112,6 +112,10 @@ public function processRequestParams(VisitProperties $visitProperties, Request $
         $isNewVisit = $this->isVisitNew($visitProperties, $request);
         $request->setMetadata('CoreHome', 'isNewVisit', $isNewVisit);
 
+        if (!$isNewVisit) { // only copy over known visitor's information, if this is for an ongoing visit
+            $this->visitorRecognizer->updateVisitPropertiesFromLastVisitRow($visitProperties);
```
tsteur: @diosmosis just curious was there a specific problem re moving it out here? Also wondering if we know for sure whether `CoreHome/RequestProcessor` will be executed first or whether other request processors may expect it to be set already? Is it because we might now find more often a `visitRow` but need to create a new visit because it is older than 30minutes maybe?

diosmosis: We don't know it's a new visit until this point. Before this point we only know if it's a known visitor (the last action might be from a visit that is > 30 mins). Since we weren't looking further than that before, it wasn't an issue, but now it's possible to find a log from a different visit.

tsteur: Can we somehow make sure that `CoreHome` request processor will be executed first?

tsteur: Just to prevent any regressions

diosmosis: Can look into that, though this used to be done in `findKnownVisitor` which is executed a couple lines above, so it's not like this code was moved from core => the CoreHome plugin. Ie, this request processor is executed at the same time it was before this change.

tsteur: got it. didn't realise that. shouldn't be an issue then 👍

core/Tracker/GoalManager.php
diosmosis: Throws an exception now if not inserted, though the error message doesn't appear in the tracker output (I think that's as it is normally).

core/Tracker/GoalManager.php
```diff
@@ -742,6 +742,11 @@ protected function insertNewConversion($conversion, $visitInformation, Request $
         Common::printDebug($newGoalDebug);
 
         $wasInserted = $this->getModel()->createConversion($conversion);
+        if (!$wasInserted) {
+            $idSite = $request->getIdSite();
+            $idorder = $request->getParam('ec_id');
+            throw new \Exception("Invalid non-unique idsite/idorder combination ($idSite, $idorder), conversion was not inserted.");
```
tsteur: @diosmosis here we cannot be 100% sure it failed due to the idOrder? Maybe we trigger the exception only when `ec_id` was set? 

Not sure if we want to trigger an exception in the other cases as it may prevent the tracking request to be fully processed as the `recordLogs` of other processors may not be executed anymore? Eg when it is a pageview and then triggers a conversion... on the other side we would never notice the conversion isn't being tracked. Tricky...

Also I think it should be maybe a `Piwik\Exception\InvalidRequestParameterException`? This should make sure to send the correct status code here: https://github.com/matomo-org/matomo/blob/3.9.0-b3/core/Tracker/Handler.php#L89-L91

diosmosis: Makes sense, will update.

tsteur:
Looks good to merge for me if tests pass 👍

diosmosis:
@tsteur looks like the silent fail might be a feature: https://github.com/matomo-org/matomo/blob/3.x-dev/tests/PHPUnit/Fixtures/TwoSitesEcommerceOrderWithItems.php#L142-L147

I guess that means I should remove the new exception... not sure how else we could report the issue, maybe a tracker failure in the UI, but then those "receipt views" might result in a lot of tracker failures...

tsteur:
@diosmosis maybe we could update the test to expect an error instead of `self::checkResponse()`? In general it be great to fail there so they have a chance to notice there was possibly something wrong. We could of course also just keep to silently fail as it does nothing in the end.

diosmosis:
@tsteur I thought about that, but the specific case linked is refreshing a "receipt" page after the order is tracked, which seems like it could happen regularly. Ie, this might be a common occurrence and not just an error. Actually I guess it wouldn't be a big issue if we gave a 400... users would likely know that it's fine for them.

tsteur:
Have no preference really, I reckon either is fine. I was thinking if there could be edge cases where it's maybe a problem to send the same request twice but can't really think of anything right now.

diosmosis:
@tsteur will just update the tests, think it should be ok if in these cases there's an error response.

PR #13939, opened 2019-01-07
Adds segmented log feature to more reports

### LINKED ISSUE #7212: Display the "Segmented Visitor Log" icon in more reports
The goal of this issue is to display the icon "Segmented Visitor Log" in more reports that currently don't show this great little icon.
#### Icon should display in the following reports
- ~~Provider (segment `provider`)~~ #12299
-  Visits per visit duration  (segment `visitDuration`)
-  Visits by Visit Number (segment `visitCount`)
-  Visits per number of pages  (segment `actions`)
-  Visits by days since last visit (segment `daysSinceLastVisit`)
- Page title (segment `pageTitle`)  
- Site Search (segment `siteSearchKeyword`)
- ~~Referrer type (segment `referrerType`)~~
#### Out of scope for this issue

 The following reports don't have an icon as they don't have yet a segment. We need to create a segment for these:
-  ~~Device brand~~ https://github.com/piwik/piwik/issues/9478
-  ~~Device model~~ #12298 
- ~~Browser language~~
- ~~Custom Variables will require a new "generic" custom variable segment, covered in #6031 #8755~~
- ~~Outlinks and Downloads, requires #4103~~

Follows up #4633

sgiehl:
fixes #7212


plugins/Actions/DataTable/Filter/Actions.php
```diff
@@ -50,8 +50,11 @@ public function filter($table)
 
             foreach ($dataTable->getRows() as $row) {
                 $url = $row->getMetadata('url');
+                $label = $row->getColumn('label');
                 if ($url) {
                     $row->setMetadata('segmentValue', urldecode($url));
+                } else if ($this->isPageTitleType && $label != -1) {
+                    $row->setMetadata('segmentValue', trim(urldecode($label)));
```
diosmosis: Is the `-1` for `ID_SUMMARY_ROW`, maybe it should also check for the new totals row? Also not sure if it's really necessary, but should we check if `$label` is not false?

core/DataTable/Filter/AddSegmentByRangeLabel.php
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * @param DataTable $table
+     */
+    public function filter($table)
+    {
+        if (empty($this->segment)) {
+            $msg = 'AddSegmentByRangeLabel is called without having any segment defined';
+            Development::error($msg);
+            return;
+        }
```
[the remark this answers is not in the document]
diosmosis: Also is this something that could happen in production? If so maybe the log should appear all the time instead of just in development mode.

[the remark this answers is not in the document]
sgiehl: Didn't have too much thoughts on that. Actually it's the same as in https://github.com/matomo-org/matomo/blob/3.x-dev/core/DataTable/Filter/AddSegmentByLabel.php
The error should not happen in production, as it should be fixed while development if it ever occurs.

plugins/Actions/DataTable/Filter/Actions.php
sgiehl: The check is required. Otherwise the summary row will have a segmented log feature, which can't work. Will add that for totals row as well....

sgiehl: Ok. No need to check for totals row, as that row will be added afterwards. Summary row might exist before.

diosmosis:
This doesn't seem to work in the "Visits by Days Since Last Visit" report. Clicking on the segment for "0 days" since last visit selects all visits. There's also no segment icon for the "New Visits" row.

diosmosis:
"Page Titles Following a Site Search"/"Pages Following a Site Search"/"Search Categories" don't have segment icons. Not sure if they're supposed to be added in this PR.

sgiehl:
> There's also no segment icon for the "New Visits" row.

This should be fixed now.

> This doesn't seem to work in the "Visits by Days Since Last Visit" report. Clicking on the segment for "0 days" since last visit selects all visits.

The segment should be set correctly. Maybe the segment isn't working correctly?

> "Page Titles Following a Site Search"/"Pages Following a Site Search"/"Search Categories" don't have segment icons.

Are there any segments we could use for this?

diosmosis:
> Are there any segments we could use for this?

Doesn't look like it, guess they'd have to be new if adding them.

sgiehl:
wasn't mentioned in #7212, so wouldn't include it here

diosmosis:
Tested locally, works for me.

PR #13949, opened 2019-01-11
Add extra plugins to load to test env manipulators fake plugin list.

diosmosis:
This way plugins are added to the list when the DI container is created, so DI config for plugins will be loaded during tests when Fixture::$extraPluginsToLoad is set.


diosmosis:
@tsteur merging, just fyi

PR #13960, opened 2019-01-14
Open site in new tab/window if ctrl key is pressed

### LINKED ISSUE #13701: Website selector: Ctrl+Click should open the website in a new tab
Steps to reproduce:

* Click on website selector
* Do Ctrl+Click on a website name
* Got: the website clicked opens in the same window
* Expected instead: the website clicked opens in a new tab

sgiehl:
fixes #13701

### DIFF
--- a/plugins/CoreHome/angularjs/siteselector/siteselector.controller.js
+++ b/plugins/CoreHome/angularjs/siteselector/siteselector.controller.js
@@ -16,7 +16,12 @@
         $scope.autocompleteMinSites = AUTOCOMPLETE_MIN_SITES;
         $scope.activeSiteId = piwik.idSite;
 
-        $scope.switchSite = function (site) {
+        $scope.switchSite = function (site, $event) {
+            if ($event && $event.ctrlKey && $event.target && $event.target.href) {
+                window.open($event.target.href, '_blank');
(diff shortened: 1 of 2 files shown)


diosmosis:
This doesn't work for me on OS X, I presume because the 'cmd' key needs to be pressed in order to launch in a new tab ('ctrl' opens a popup menu).

sgiehl:
@diosmosis  could you check if `metaKey` instead of `ctrlKey` works on OS X? I don't have a Mac, so can't test it

tsteur:
Can we change the HTML structure slightly to use an actual link for each site and the all websites plus `ng-href` or so which updates the link when things change? Then it would work properly on all OS etc

sgiehl:
actually there are "real" links with hrefs for each site. But all clicks on those links are ignored because of `piwik-ignore-click`

diosmosis:
@sgiehl w/ metaKey it works for me

Dzhuneyt (approved):

sgiehl:
@diosmosis added a check to use `metaKey` for OS X and `ctrlKey` for other systems

Dzhuneyt (approved):

Issue #9200, opened 2015-11-12
When clicking "Next" on Visitor log with an Action Segment applied, the next page may return no data


uedatakeshi:
piwik 2.14.3

conditions;
period=range
date=2015-10-01,2015-10-09
segment=customVariablePageValue1==ABC123

Visitors overview shows 250 visits.
But Visitor Log show only 32 lines.
Click Next> shows no data.

If I added '&filter_limit=-1' to the end of URL and hit enter, then it show 250 lines.

hpvd:
Is it still same behaviour in piwik 2.15?

uedatakeshi:
Yes, Piwik 2.15.0 works same.

mattab:
Hi @uedatakeshi can you reproduce this bug on our demo @ https://demo.piwik.org/index.php?module=CoreHome&action=index&idSite=7&period=day&date=yesterday ? we have a couple custom variables there but when applying the segment it seems to work: [link with segment shows as many visits in Visitor log as in Visitors>Overview](https://demo.piwik.org/index.php?module=CoreHome&action=index&idSite=7&period=day&date=yesterday&segment=customVariableName3%3D%3DForum%2520status%3BcustomVariableValue3%3D%3DLoggedIn%2520user#?module=Live&action=indexVisitorLog&idSite=7&period=day&date=yesterday&segment=customVariableName3%3D%3DForum%2520status%3BcustomVariableValue3%3D%3DLoggedIn%2520user)

uedatakeshi:
Thank you Matt.
I have no account to demo.piwik.org.
So I show you the difference between existing segment 'Logged in users' and ours.

demo.piwik.org
Logged in users
=customVariableName3==Forum status;customVariableValue3==LoggedIn user

our site
School segment
segment=customVariablePageName1==School;customVariablePageValue1==ABC123

Would you please create custome variable with scope="page"?
And then add segment to test.

mattab:
@uedatakeshi I put piwikjapan.org statistics as public on demo.piwik.org, so you are welcome to add the custom variable of scope 'page' and try to reproduce the issue there. See the demo: http://demo.piwik.org/index.php?module=CoreHome&action=index&date=today&period=month&idSite=33

Let me know if this is an issue that I put piwikjapan.org stats as public, I could remove it. Thanks!

uedatakeshi:
Hi, Matt!
Thank you for your advice.
I'm sorry I was reminded I have an account to login demo.piwik.org.

[...]

PR #13961, opened 2019-01-14
Show next link in visitor log only if more records are available

### LINKED ISSUE #10747: Visitor log: show link to next page only if there is a next page
Visitor log: show link to next page only if there is a next page
2016-10-17_12h03_29

sgiehl:
With the changes there will always be one record more requested than needed.
The extra record will be removed again if available and the next will be shown in this case.

fixes #10747

### DIFF
--- a/plugins/Live/Visualizations/VisitorLog.php
+++ b/plugins/Live/Visualizations/VisitorLog.php
@@ -51,6 +51,7 @@ class VisitorLog extends Visualization
             $this->requestConfig->filter_limit = $defaultLimit;
         }
 
+        $this->requestConfig->request_parameters_to_modify['filter_limit'] = $this->requestConfig->filter_limit+1; // request one more record, to check if a next page is available
         $this->requestConfig->disable_generic_filters = true;
         $this->requestConfig->filter_sort_column      = false;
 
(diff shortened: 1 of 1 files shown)


diosmosis:
Works for me locally, but will it be an issue if something like https://github.com/matomo-org/matomo/issues/9200 happens (wasn't able to reproduce that issue locally)?

sgiehl:
maybe that might be fixed as well, as the link should only be displayed if there is at least one more record available on the next page.

diosmosis:
That issue is about data not being displayed for some pages, but not all. So one page shows 20 visits, the next page shows nothing, the next page shows 10 visits. W/ this PR the next button would be removed so the user can't get to the pages w/ data, if one shows nothing. Is that understanding correct? Might be an issue for that situation.

sgiehl:
ah ok. now I understand the other issue correct. Guess you are right.

sgiehl:
@diosmosis had another thought on that right now. Actually I don't think it would be a problem. If a page is completely empty, there isn't a next link at the moment as well. It only shows the "no data message". So that behavior won't change with the PR

diosmosis:
@sgiehl makes sense, will merge.

PR #13962, opened 2019-01-14
Apply AddSegmentBySegmentValue filter recursive

### LINKED ISSUE #11849: Segmented Visitor Log icon does not show in Actions > Pages report in sub-tables, when a search is being done
Pages within sub-directories/folders in Actions report, should have the Segmented Visitor Log icon

### how to reproduce

* [Go to Actions>Pages](http://demo.piwik.org/index.php?module=CoreHome&action=index&idSite=3&period=day&date=yesterday#?idSite=3&period=day&date=yesterday&category=General_Actions&subcategory=General_Pages)
* Search a page in a sub-directory, in this example `index`
* The search results display, with the links at the root level showing the segmented visitor log icon:

ok

* But issue is that pages don't show the segmented visitor log icon:
nok

Similar to https://github.com/piwik/piwik/issues/9229 - but different (this other ticket is about the folders, whereas this present ticket is about the pages within the folders)

sgiehl:
When a datatable is loaded expanded the `segmentValue` metadata won't be converted into `segment` metadata for all subtables. This is also the case when search in action subtables as described in #11849.

fixes #11849

### DIFF
--- a/core/DataTable/Filter/AddSegmentBySegmentValue.php
+++ b/core/DataTable/Filter/AddSegmentBySegmentValue.php
@@ -62,6 +62,8 @@ class AddSegmentBySegmentValue extends BaseFilter
             return;
         }
 
+        $this->enableRecursive(true);
+
         /** @var \Piwik\Plugin\Segment $segment */
         $segment     = reset($segments);
         $segmentName = $segment->getSegment();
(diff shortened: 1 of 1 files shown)


diosmosis (approved):
Tests need an update, otherwise looks good

PR #13963, opened 2019-01-14
Append token_auth to sparkline urls for embedded widgets

### LINKED ISSUE #13625: Sparklines broken on Widgetize widget when viewed by anonymous (even with token_auth)
How to reproduce:

1. Get a URL for the Widgetize widget in Settings > Widgets > Generic > Metric
2. Go to your account settings and take note ot your `token_auth`
3. Log out Matomo
4. Paste the widget URL into the URL bar
5. Append your `token_auth` (i.e. add `&token_auth=XXXXX` at the end of the URL)
6. Load the page

Outcome: metric widget loads, but sparkline image is broken

Expected: sparklines image not broken

NB: sparkline URL generated by the URL does **not** contain the `token_url`, even though widget has it. If you copy and load the sparkline URL as is, you get redirected to the login page. If you add the `token_url` parameter manually, then the image loads.

Matomo 3.6.0 / PHP 7.2

sgiehl:
fixes #13625

### DIFF
--- a/plugins/CoreHome/angularjs/sparkline/sparkline.component.js
+++ b/plugins/CoreHome/angularjs/sparkline/sparkline.component.js
@@ -47,8 +47,8 @@
             var urlParams = piwikApi.mixinDefaultGetParams($element.extend(defaultParams, vm.params));
 
             // Append the token_auth to the URL if it was set (eg. embed dashboard)
-            var token_auth = piwik.broadcast.getValueFromUrl('token_auth');
-            if (token_auth.length) {
+            var token_auth = piwik.broadcast.getValueFromUrl("token_auth");
+            if (token_auth.length && "Widgetize" === piwik.broadcast.getValueFromUrl("module")) {
                 urlParams.token_auth = token_auth;
             }
 
--- a/plugins/CoreHome/angularjs/sparkline/sparkline.component.js
+++ b/plugins/CoreHome/angularjs/sparkline/sparkline.component.js
@@ -45,6 +45,13 @@
             };
 
             var urlParams = piwikApi.mixinDefaultGetParams($element.extend(defaultParams, vm.params));
+
+            // Append the token_auth to the URL if it was set (eg. embed dashboard)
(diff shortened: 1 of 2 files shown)


plugins/CoreHome/angularjs/sparkline/sparkline.component.js
```diff
@@ -45,6 +45,13 @@
             };
 
             var urlParams = piwikApi.mixinDefaultGetParams($element.extend(defaultParams, vm.params));
+
+            // Append the token_auth to the URL if it was set (eg. embed dashboard)
+            var token_auth = piwik.broadcast.getValueFromUrl('token_auth');
+            if (token_auth.length) {
+                urlParams.token_auth = token_auth;
+            }
+
```
diosmosis: I guess this means the token will become stored in the browser cache? I guess it can't be avoided in this case, but would it be a bit better if we only added it if in the widgetized iframe? So if someone adds the token auth to a URL when not in a widgetized whatever, it won't get added to every sparkline URL.

sgiehl: Actually I copied that from here:
https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/CoreHome/javascripts/sparkline.js#L31-L35

But guess makes sense to replace both places and check if `module` parameter is `Widgetize`?

sgiehl: changed it

diosmosis: Tested locally and noticed the token_auth is always added to the sparklines when token_auth is in the URL due to https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/CoreVisualizations/Visualizations/Sparklines/Config.php#L339. Doesn't affect the single metric view widget since that is done client side. Not sure if it's worth changing this.

diosmosis: The JavaScript would add it back if it's required, right?

PR #13972, opened 2019-01-17
Add diagnostic to check last time archiving was run successfully and …

### LINKED ISSUE #13753: Warn Super Users when archiving hasn't successfully finished in some time
A very common forums support request is when users don't see any data in their report. We have extensive documentation about the many [causes for no data.](https://matomo.org/faq/troubleshooting/faq_58/). 
A subset of these users actually successfully track the data (and can see it in the Visitor log) but they see  "There is no data for this report" in all reports. And the likely cause of this problem is that the `core:archive` task is not working. (Refers to the 3rd bullet point in our FAQ `Archiving triggered from the browser must be enabled, if it is disabled the Cron script must be configured properly`)

So the idea is to help these users and put them on the path to resolving the issue.

Suggested steps:

### In the System check page

* When archiving last successfully executed 
 * more than eg. 24 hours ago, issue a warning in the System check 
 * more than 48 errors ago,  maybe up the level to Error?
*  the message would say eg. "The archiving last processed successfully on Date_time which is 43 hours ago. Please check that you have setup a crontab calling `core:archive` console command, and that you have configured a `MAILTO` to receive errors by email if the archiving fails. You can also try to run the console command to archive your reports manually: \n ./path/to/matomo/console core:archive --url=http://example.org/piwik/" ( ideally we replace the URL automatically by the URL and maybe the path)

### In the Reports themselves

When we write "There is no data for this report" and detect that archiving hasn't been successful in a while, then
* we could add a message saying eg. "The archiving of your reports hasn't been executed recently, <a>learn more about how to generate your reports</a>."  
* maybe only Admins/Super Users should see this (or maybe it's useful to also show it to View users so they can alert others?) 

### Recommend users more strongly to deactivate browser archiving

we could guide users to setup the cron archiving for an optimal user experience. this would ultimately make Matomo much faster and would reduce amount of users who experience a slow Matomo (many people say Matomo is slow because browser archiving runs and is slow).

For example, if users don't have yet disabled the "Browser archiving trigger" feature:
  * We could write a message/even maybe a warning in the system check "For optimal performance and a speedy Matomo, it is highly recommended to set up a crontab to automatically archive the reports, and to disable browser triggering in the Matomo settings. <a target="_blank" href='https://matomo.org/docs/setup-auto-archiving/'>Learn more</a>"

diosmosis:
…display notification if empty report and archiving has not been run recently.

The notification won't be displayed with the segment notification in order to avoid confusion.

Fixes #13753


sgiehl (approved):
Seems to work as expected.
There might be an edge case, if someone sets up an archiving job, which only archives a specific site. But not sure if it's worth to solve that.

@mattab @tsteur maybe someone should check the new messages if they are fine

mattab:
Screenshot from 2019-03-12 13-21-28

Could we move the new diagnostic just below the other archive related diagnostic?

> There might be an edge case, if someone sets up an archiving job, which only archives a specific site. But not sure if it's worth to solve that.

Not needed to deal with this case, we can for now assume the crontab is setup correctly :+1:

mattab:
* Also would it be maybe possible to have a UI test showing the footer message?

mattab:
> Also would it be maybe possible to have a UI test showing the report footer message?

Actually maybe adding UI tests for these kind of features is not a good idea as it would be bloated if we test everything in this way....
-> Could we maybe instead add an integration test that checks the output of the report contains our footer message?

diosmosis:
> Could we maybe instead add an integration test that checks the output of the report contains our footer message?

@mattab it's actually in a notification, should it be a footer message in the report? Note: this will be added to every report, so you'll see the footer message everywhere, as opposed to a single notification.

mattab:
Notification sounds better 👍

Issue #6559, opened 2014-10-30
Add support for app specific passwords


tsteur:
It would be great if a user could create app specific passwords and revoke them if needed. 

For instance you might want to create a new app specific password for each device when using the mobile app. If you ever lose a device (phone) you can simply revoke the password that you have used for this device but all other passwords would be still valid. 

You do not want to create a new user for the mobile app with a different password as you would have to sync/change the dashboards and ViewDataTable params for each user. Once we implement features like "Continuity" one would still benefit of those features. With "Continuity" I mean when opening the mobile app it could open the same website and report that you have accessed last on the desktop although different passwords are used etc. This would not be possible if someone had to create multiple users for each device.

You might also want to generate app specific passwords for different servers that talk to the Piwik HTTP API etc. 

It will be also useful if we ever generate 2-factor authentication.

App specific passwords would consist of an App name (eg PiwikMobileApp, or PythonClientWhatever), a device (eg PhoneXYZ or ServerXYZ) and a generated password. This password would be ideally quite long, eg 16 characters. Before someone can generate a new specific app password we should ask him for his actual password and only generate / or only let him manage his app passwords if the actual password is correct. 

Some systems that do support app specific passwords would not let you log in into the browser version (meaning regular Piwik login) with an app specific password but we would probably still do?

* [ ] Store tokens / passwords hashed
* [ ] Do no longer show them anywhere in the UI

mattab:
Currently users have some issues with the fact that the `token_auth` is not found in API responses (see #11159 ). We know that one could create a plugin that would add back into API responses the `token_auth`. But I'm wondering if we could provide another (more secure) solution, with this App-specific passwords discussed here. 

### Use case

* As a Super User, I want to give my customers (read only) access to their reports via a [report Embed iframe](https://piwik.org/docs/embed-piwik-report/). For this as a Super User I want to write a script which creates a "Read-only" access to reports for a website, for which I need a API `token_auth` with read-only access.

### Would this use case be met with App Specific passwords?
Could App specific passwords also help with this use case above? @sgiehl maybe you can confirm whether App-specific password would help here and if so, how would it be scripted by calling the API?

sgiehl:
If we implement the app specific passwords as kind of additional token_auths for a user the api would be called as usual, but as token_auth a app specific password would be given. So it would match your use case

mattab:
* you mean the `&token_auth=` will also accept app-specific passwords as value?
[...]

PR #13975, opened 2019-01-18
Require password confirmation before setting/removing superuser access.

### LINKED ISSUE #13710: Personal token auth should be only visible after entering the password
Currently, you only need to click to view the token auth. I reckon instead we should ask a user to enter the password again before showing it. We could potentially always show the first 5 characters though. from @tsteur

### LINKED ISSUE #13711: Require password confirmation when giving someone super user access
Similar to #2932 and https://github.com/matomo-org/matomo/issues/13710 it would be also good to eventually have a password confirmation when giving someone user access

diosmosis:
Adds the confirmation to listAllAPI.twig and to userSettings.twig. There is some code redundancy between the two which I couldn't find a good way of removing.

Fixes #13711


tsteur:
Not sure if it's needed in `listAllApi`? You could also see the token in export container, or just by typing `piwik.token_auth` or something in the console... also by monitoring network requests etc. I reckon we would only do it in the personal settings screen for now? Or maybe we even don't do it at all until we removed it everywhere and replaced token_auth with something better?

diosmosis:
cc @mattab

mattab:
> Or maybe we even don't do it at all until we removed it everywhere and replaced token_auth with something better?

Sounds good to me :+1: 

 better API authentication will be done later in https://github.com/matomo-org/matomo/issues/6559

diosmosis:
Closing

diosmosis:
Whoops forgot there was another change here

diosmosis:
@tsteur removed the code to require password before showing token auth. UI tests may need to be fixed.

plugins/UsersManager/API.php
```diff
@@ -698,12 +698,14 @@ public function addUser($userLogin, $password, $email, $alias = false, $_isPassw
      * @param string   $userLogin          the user login.
      * @param bool|int $hasSuperUserAccess true or '1' to grant Super User access, false or '0' to remove Super User
      *                                     access.
+     * @param string $passwordConfirmation the current user's password.
      * @throws \Exception
      */
-    public function setSuperUserAccess($userLogin, $hasSuperUserAccess)
+    public function setSuperUserAccess($userLogin, $hasSuperUserAccess, $passwordConfirmation)
```
sgiehl: I guess we should mention that *breaking* API change in the changelog

plugins/UsersManager/API.php
```diff
@@ -698,12 +698,14 @@ public function addUser($userLogin, $password, $email, $alias = false, $_isPassw
...
     {
         Piwik::checkUserHasSuperUserAccess();
         $this->checkUserIsNotAnonymous($userLogin);
+        $this->confirmCurrentUserPassword($passwordConfirmation);
```
sgiehl: That method actually doesn't exists (anymore?)

sgiehl (changes requested):
Due to the not existing method confirming the password currently returns a `500 - internal server error`. Wondering if we should handle such issues in any way. Currently the popover is closed and nothing happens. Maybe it would be good to at least display an error message?

plugins/UsersManager/API.php
diosmosis: Added it, maybe I forgot to commit...

plugins/UsersManager/API.php
diosmosis: Added to the changelog

diosmosis:
CC @tsteur to make sure the breaking API change is ok

tsteur:
This breaking change should be fine 👍

tsteur:
Looks good. 
I looked at dependencies where we use this method and I think it be good if we could maybe disable the passwordConfirmation check when we are in CLI mode? This would eg help on the cloud.

Aso I think `Installation\Controller::createSuperUser` needs to be adjusted and some tests need to be fixed.

PR #13977, opened 2019-01-18
Don't cache tag manager preview files

tsteur:
Not sure if this will work across apache versions? Tested on Apache 2.2

Prevents preview files from being cached in the browser and that always the newest version will be served. Otherwise the user will need to use ctrl+r or something to reload a page that includes a preview file to make sure the cache is ignored and many users would not know that.

refs https://github.com/matomo-org/tag-manager/pull/150
for Nginx see https://github.com/matomo-org/matomo-nginx/issues/48


tsteur:
can you check if that works?

diosmosis:
👍 works

PR #13978, opened 2019-01-20
 Improve site search detection in fragment query

### LINKED ISSUE #13971: Site search query parameter is ignored if it is in a fragment.
If a URL like `path/to/page#this-location?q=word` is used the `q` parameter is ignored. If a URL like `path/to/page#this-location&?q=word` is used then the query is added to the site search results. The expectation is that any URL with a valid search parameter will work.  

> It seems Matomo concatenates the querystring and the fragment with an `&` and then tries to parse the whole string as querystring

On some systems it is not possible to modify the URL to append an `&` so the site search report does not record these queries.

sgiehl:
fixes #13971

### DIFF
--- a/plugins/Actions/Actions/ActionSiteSearch.php
+++ b/plugins/Actions/Actions/ActionSiteSearch.php
@@ -132,8 +132,20 @@ class ActionSiteSearch extends Action
         $keywordParameters = isset($website['sitesearch_keyword_parameters'])
             ? $website['sitesearch_keyword_parameters']
             : array();
-        $queryString = (!empty($parsedUrl['query']) ? $parsedUrl['query'] : '') . (!empty($parsedUrl['fragment']) ? $separator . $parsedUrl['fragment'] : '');
-        $parametersRaw = UrlHelper::getArrayFromQueryString($queryString);
+        $queryString = !empty($parsedUrl['query']) ? $parsedUrl['query'] : '';
+        $fragment = !empty($parsedUrl['fragment']) ? $parsedUrl['fragment'] : '';
+
+        // check if fragment contains a separate query (beginning with ?) otherwise assume complete fragment as query
+        if ($fragment && strpos($fragment, '?') !== false &&
+            (strpos($fragment, '&') === false || strpos($fragment, '?') < strpos($fragment, '&'))) {
+            $fragmentBeforeQuery = substr($fragment, 0, strpos($fragment, '?'));
+            $fragmentQuery = substr($fragment, strpos($fragment, '?')+1);
+        } else {
+            $fragmentQuery = $fragment;
+            $fragmentBeforeQuery = '';
+        }
+
+        $parametersRaw = UrlHelper::getArrayFromQueryString($queryString.$separator.$fragmentQuery);
 
         // strtolower the parameter names for smooth site search detection
         $parameters = array();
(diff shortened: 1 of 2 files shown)


plugins/Actions/Actions/ActionSiteSearch.php
```diff
@@ -132,8 +132,20 @@ protected function detectSiteSearchFromUrl($website, $parsedUrl)
...
             ? $website['sitesearch_keyword_parameters']
             : array();
-        $queryString = (!empty($parsedUrl['query']) ? $parsedUrl['query'] : '') . (!empty($parsedUrl['fragment']) ? $separator . $parsedUrl['fragment'] : '');
-        $parametersRaw = UrlHelper::getArrayFromQueryString($queryString);
+        $queryString = !empty($parsedUrl['query']) ? $parsedUrl['query'] : '';
+        $fragment = !empty($parsedUrl['fragment']) ? $parsedUrl['fragment'] : '';
+
+        // check if fragment contains a separate query (beginning with ?) otherwise assume complete fragment as query
+        if ($fragment && strpos($fragment, '?') !== false &&
+            (strpos($fragment, '&') === false || strpos($fragment, '?') < strpos($fragment, '&'))) {
```
diosmosis: I think the second condition is to differentiate between a fragment that is a query string and a fragment w/ `&`/`?` that isn't a query string, correct? I haven't tried it, but would it be easier to call `parse_url()` on the fragment and check for the `'query'` property?

plugins/Actions/tests/Integration/ActionSiteSearchTest.php
```diff
@@ -0,0 +1,67 @@
...
+            [$defaultWebsite, 'http://example.org/index.htm#&q=keyword', ['http://example.org/index.htm', 'keyword', '', false]],
+            [$defaultWebsite, 'http://example.org/index.htm#?cat=test&q=keyword', ['http://example.org/index.htm', 'keyword', 'test', false]],
+            [$defaultWebsite, 'http://example.org/index.htm#?cat=test&q=keyword&otherparam=1', ['http://example.org/index.htm#otherparam=1', 'keyword', 'test', false]],
+            [$defaultWebsite, 'http://example.org/index.htm?cat=test&otherparam=1&q=keyword', ['http://example.org/index.htm?otherparam=1', 'keyword', 'test', false]],
+            [$defaultWebsite, 'http://example.org/index.htm#anchor?cat=test&otherparam=1&q=keyword', ['http://example.org/index.htm#anchor?otherparam=1', 'keyword', 'test', false]],
+            [$defaultWebsite, 'http://example.org/index.htm?cat=test&otherparam=1&q=kw#?q=keyword', ['http://example.org/index.htm?otherparam=1', 'keyword', 'test', false]],
+            [$defaultWebsite, 'http://example.org/index.htm?k=keyword&cc=cat', ['http://example.org/index.htm', 'keyword', 'cat', false]],
```
diosmosis: Some other test cases that could be added:

- `http://example.org/index.html?a=b#?&&&?q=keyword`
- `http://example.org/#&?q=keyword`
- `#?q=keyword`
- `http://example.org/#&?var=val?q=keyword`
- `http://example.org/#q=keyword?q=anotherkeyword`
- maybe some other weird combos

What do you think? They're not likely to occur in practice, but it might be good to define the behavior in these cases.

sgiehl: sure. why not

sgiehl:
@diosmosis updated the code to use `parse_url` and added the suggested tests

PR #13981, opened 2019-01-20
Exclude non core plugins from translation percentage calculation

sgiehl:
Note: This will also exclude our Marketplace plugins that are not bundled with core.


diosmosis (approved):
Code looks good to me

PR #13986, opened 2019-01-22
Serialize scalars and objects in original format.

diosmosis:
Will change the API for format=original&serialize=1 for scalar values & objects (not sure if that's an issue).


tsteur:
👍

PR #13997, opened 2019-01-22
Ignore .lfsconfig in file integrity system check

### LINKED ISSUE #13995: Explain better why sometimes files "appear" after updates
I have read many questions recently by people asking why there is suddenly a `.lfsconfig` file in their Matomo directory that should be deleted according to integrity checker.

Of course most people don't notice that the file has been a part of Matomo for a while and then has been deleted in the latest release (https://github.com/matomo-org/matomo-package/pull/86) which causes it to be useless after the update and therefore flagged by integrity checker. 

I think we should add an explanation text to the integrity checker that lists reasons why the file may appear there and why this isn't a bug.

mattab:
fixes https://github.com/matomo-org/matomo/issues/13995

### DIFF
--- a/config/global.php
+++ b/config/global.php
@@ -86,6 +86,7 @@ return array(
         'favicon.ico',
         'robots.txt',
         '.bowerrc',
+        '.lfsconfig',
         '.phpstorm.meta.php',
         'config/config.ini.php',
         'config/config.php',


Findus23:
That's of course a solution I didn't think of. :slightly_smiling_face:  

But I am not sure if this is the full solution as it would mean we need to make an exception for every file that will get deleted by a Matomo release (e.g. this time there is also `misc/internal-docs/content-tracking.md`)

sgiehl (approved):

diosmosis:
Tried to add a test for this, but couldn't find a way to generate the manifest file. Perhaps if the logic were moved into a matomo command or something it would be possible. For now will just merge.

PR #13999, opened 2019-01-22
Allow config files to be placed per hostname in misc/user directory

tsteur:
We already allow users to have a custom config per hostname by putting them into `config/$hostname.config.ini.php`. On the contrary, custom logos are uploaded in `misc/user/$hostname/*`. 

It would make sense to have all files per hostname in `misc/user/$hostname`, including the config. This PR lets you put a config file into eg `misc/user/apache.matomo/config.ini.php` and then the custom config will be used for this host.

Old way is still supported, we just look in the two places now.

fyi @mattab


tsteur:
FYI: We're running this already in production and works well

core/Config.php
```diff
@@ -215,28 +229,42 @@ public static function getHostname()
...
     {
-        $hostConfig = self::getLocalConfigInfoForHostname($hostname);
+        $hostConfigs = self::getLocalConfigInfoForHostname($hostname);
+        $fileNames = '';
+
+        foreach ($hostConfigs as $hostConfig) {
+            if (count($hostConfigs) > 1
+                && $preferredPath
+                && strpos($hostConfig['path'], $preferredPath) === false) {
+                //
```
diosmosis: Was there supposed to be a comment here?

tsteur: will remove it

diosmosis (approved):
Wondering if this would be good to cover in a test somehow, otherwise looks good to me.

PR #14001, opened 2019-01-23
Do not enable brute force detection during update process.

### LINKED ISSUE #13836: Mysqli prepare error: Table 'piwik.piwik_brute_force_log' doesn't exist"
Hello,

While upgrading through automatic update to Matomo 3.8.0-b4, from 3.8.0b3, I have got this message : 

""Error in Matomo: Mysqli prepare error: Table 'piwik.piwik_brute_force_log' doesn't exist" while reading response header from upstream, client: xxx.xxx.xxx.xxx, server: domain.tld, request: "POST /index.php?module=CoreUpdater&action=oneClickResults HTTP/2.0", upstream: "fastcgi://unix:/var/php-nginx/xxxxxxxx.sock/socket:", host: "domain.tld", referrer: "https://domain.tld/index.php?module=CoreUpdater&action=oneClickUpdate"

Then the database upgrade screen appear as usual. Process.

All seems to have been well completed so Matomo is fonctional.

- MySQL version: 5.5.62-0ubuntu0.14.04.1
- PHP version: 5.5.9-1ubuntu4.26

So, report it in case I missed something.

Thanks,

Eric

diosmosis:
Should fix #13836 but I wasn't able to reproduce the issue locally so not 100% sure.

### DIFF
--- a/plugins/Login/Security/BruteForceDetection.php
+++ b/plugins/Login/Security/BruteForceDetection.php
@@ -36,6 +36,11 @@ class BruteForceDetection {
 
     public function isEnabled()
     {
+        $module = Common::getRequestVar('module', false);
+        if ($module == 'CoreUpdater') {
+            return false; // do not enable the brute force check during update
+        }
+
         return $this->settings->enableBruteForceDetection->getValue();
     }
 
--- a/plugins/Login/Security/BruteForceDetection.php
+++ b/plugins/Login/Security/BruteForceDetection.php
@@ -45,7 +45,7 @@ class BruteForceDetection {
     public function isEnabled()
     {
         $dbSchemaVersion = $this->updater->getCurrentComponentVersion('core');
-        if (version_compare($dbSchemaVersion, '3.8.0') == -1) {
+        if ($dbSchemaVersion && version_compare($dbSchemaVersion, '3.8.0') == -1) {
             return false; // do not enable brute force detection before the tables exist
         }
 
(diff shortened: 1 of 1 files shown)


plugins/Login/Security/BruteForceDetection.php
```diff
@@ -36,6 +36,11 @@ public function __construct(SystemSettings $systemSettings)
 
     public function isEnabled()
     {
+        $module = Common::getRequestVar('module', false);
```
tsteur: @diosmosis not sure about that way... eg you could maybe add a `&module=CoreUpdater` while maybe somehow still pinging the API or UI or archiving? like requesting `http://.../misc/user/archive.php?module=CoreUpdater`. I wonder if there's a different way to find out if we are currently "updating" or whether the `coreupdater` is actually being rendered

diosmosis: good point

diosmosis: I guess we could check `Updater::isNewVersionAvailable()` but I don't think we'd want to do that on every request...

Or could check that the version_core option >= 3.8.0. If version_core is autoloaded, then that would be ok I think...

diosmosis: @tsteur went w/ disabling it if version_core is < 3.8.0. should be the fastest approach. after 3.8 the tables should be there, so wouldn't this problem to arise.

tsteur:
Didn't try to reproduce but looks good. Looked through the code making sure that the `core` option gets only updated after components have been updated. In this case the code should work. Hard to know when we cannot reproduce it.

PR #14002, opened 2019-01-23
Reset tracked impressions for searches and only send tracking request when interaction was defined

### LINKED ISSUE #13988: Content Tracking: Interactions count greater than impressions count
May potentially be related to #6270

- I have single page application that features a search page
- Search results are manually using internal search tracking via `trackSiteSearch` 
- As per Piwik/Matomo [recommendation](https://developer.matomo.org/guides/tracking-javascript-guide#internal-search-tracking), `trackPageView` is not called on the search page since `trackSiteSearch` is already being called
```
// We recommend not to call trackPageView() on the Site Search Result page
```
- I am also tracking content impressions automatically using data-attributes and initializing via a call to `trackAllContentImpressions`
- Interaction tracking is handled by Piwik automatically on click, for each content block

The steps to reproduce this issue
- Perform a search query
- - the content impression should now be tracked for the items in the search result
- - `trackSiteSearch` is called
- Click on the trackable content in search result
- - on page transition, the interaction **_is_** tracked
- Navigate back to the Search results page by clicking the back button in the browser
- - the content impression **_is not_** tracked for the items in the result
- - click the trackable content in search result
- - on page transition, the interaction **_is_** tracked

The above steps will result in more content interactions than content impressions for the content block that is interacted with.

~~I believe the issue emanates from the fact that on back navigation, all content impressions are not registered in Piwik even though a call is made via `trackAllContentImpressions`, because a call to `trackPageView` is not made to force Piwik to register the impressions?~~

~~As per [Content Tracking](https://developer.matomo.org/guides/tracking-javascript-guide) section:~~
```
We won't send an impression of the same content block twice if you call this method multiple times unless trackPageView() is called meanwhile. This is useful for single page applications.
```

EDIT:

After a bit of further investigation, it does not seem that failure to call `trackPageView` is the cause of this issue. Manually, adding a call to `trackPageView` still allowed reproduction of the issue

tsteur:
fix #13988

In #13988 noticed couple problems:

* The content impressions are not reset when calling `trackSiteSearch`, only when calling `trackPageview`
* The method `buildContentInteractionRequestNode ` may return `null|undefined` in which case we should not send the request. Not sure what Matomo would do but it would possibly track a pageview instead of an interaction


theseyi:
Awesome! Thanks @tsteur :+1:

theseyi:
@tsteur not to bug, a couple questions
- Is there a ETA or timeline for this getting merged and publicly released? 
- in the interim, would it suffice for me to manually pull in the updated js files in my app  to take advantage of this fix?

thanks

tsteur:
> Is there a ETA or timeline for this getting merged and publicly released?

I reckon within the next 6 weeks

> in the interim, would it suffice for me to manually pull in the updated js files in my app  to take advantage of this fix?

Yes, you can simple update the piwik.js: https://raw.githubusercontent.com/matomo-org/matomo/fe2eebbca7c2f05e6ff95ff389d5878a3bb7ba5c/piwik.js and matomo.js: https://raw.githubusercontent.com/matomo-org/matomo/fe2eebbca7c2f05e6ff95ff389d5878a3bb7ba5c/matomo.js
(both same content)

diosmosis:
@tsteur is there a test that can be added for this?

tsteur:
@diosmosis added a test for the trackSiteSearch reset, and merged 3.x-dev in so the PR can be merged.

PR #14004, opened 2019-01-23
Allow DataTables to disable generic filters gthrough a datatable metadata.

diosmosis:
This is specifically for ProxySite where reports will request a filter_limit/offset, and the target instance will apply it once, and we don't want to apply it a second time in the proxy instance.

The alternative is to always send filter_limit=-1 and apply the limit in the proxy instance, but that would mean a lot of data transfer. Especially for Live APIs, which just time out for me (of course they would).

### DIFF
--- a/core/API/DataTableGenericFilter.php
+++ b/core/API/DataTableGenericFilter.php
@@ -155,6 +155,7 @@ class DataTableGenericFilter
             return;
         }
 
+        $tableDisabledFilters = $datatable->getMetadata(DataTable::GENERIC_FILTERS_TO_DISABLE_METADATA_NAME) ?: [];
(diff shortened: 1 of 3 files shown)


tsteur:
Haven't tested it but looks fine. A test may be good if easy to do.

Just to understand correctly on the instance that has proxySite we will basically disable the limit filter as this would have been applied again otherwise?

diosmosis:
> Just to understand correctly on the instance that has proxySite we will basically disable the limit filter as this would have been applied again otherwise?

Yes, correct.

diosmosis:
Added unit test.

tsteur (approved):

PR #14005, opened 2019-01-23
Show url and page title on different lines in realtime visits widget & show tooltip more quickly

### LINKED ISSUE #13991: Widget > Visits in Real-Time :: New 3.8.0 Hover URLs are unclear, cluttered and way too long just as this issue's title! pun fully intended :)
Ok,I've made my point
Just see the pictures, it's all there :)

**Matomo 3.7.0: (before 3.8.0 update)**
piwik-matomo-3 7 0-old-hover-tooltips

**Matomo 3.8.0 update**
piwik-matomo-3 8 0-new-hover-tooltips-include-urls

**Proposition / Some common sense here**
piwik-matomo-3 8 x-fix-proposition

Thanks!

diosmosis:
Also truncates url to 120 chars.

The tooltip delay is reduced to 100ms. I couldn't reproduce the stuck tooltip issue yhat the delay fixed, so hoping it will be ok.

Fixes #13991

### DIFF
--- a/plugins/CoreHome/angularjs/widget/widget.directive.js
+++ b/plugins/CoreHome/angularjs/widget/widget.directive.js
@@ -109,7 +109,7 @@
                             var title = $(this).attr('title');
                             return piwikHelper.escape(title.replace(/\n/g, '<br />'));
                         },
-                        show: {delay: 700, duration: 200},
+                        show: {delay: 100, duration: 200},
                         hide: false
                     });
 
--- a/plugins/CoreHome/angularjs/widget/widget.directive.js
+++ b/plugins/CoreHome/angularjs/widget/widget.directive.js
@@ -109,7 +109,7 @@
                             var title = $(this).attr('title');
                             return piwikHelper.escape(title.replace(/\n/g, '<br />'));
                         },
-                        show: {delay: 100, duration: 200},
+                        show: {delay: 700, duration: 200},
                         hide: false
                     });
 
--- a/plugins/Live/templates/getLastVisitsStart.twig
+++ b/plugins/Live/templates/getLastVisitsStart.twig
@@ -85,7 +85,7 @@
 {{ action.serverTimePretty }}
 {% if action.timeSpentPretty is defined %}{{ 'General_TimeOnPage'|translate }}: {{ action.timeSpentPretty|raw }}{% endif %}
 {%- endset %}
-                                    <img src="plugins/Live/images/file{{ col }}.png" title="{{ action.url|e('html_attr') }}&mdash;{{- title -}}"/>
+                                    <img src="plugins/Live/images/file{{ col }}.png" title="{{ action.url|truncate(120)|e('html_attr') }}<br/>{{- title -}}"/>
                                 {% elseif action.type == 'outlink' or action.type == 'download' %}
                                     <img class='iconPadding' src="{{ action.icon }}"
                                          title="{% if action.url is defined %}{{ action.url }} - {% endif %}{{ action.serverTimePretty }}"/>
--- a/plugins/Live/templates/getLastVisitsStart.twig
+++ b/plugins/Live/templates/getLastVisitsStart.twig
@@ -85,7 +85,7 @@
 {{ action.serverTimePretty }}
 {% if action.timeSpentPretty is defined %}{{ 'General_TimeOnPage'|translate }}: {{ action.timeSpentPretty|raw }}{% endif %}
 {%- endset %}
-                                    <img src="plugins/Live/images/file{{ col }}.png" title="{{ action.url|e('html_attr') }}<br/>{{- title -}}"/>
+                                    <img src="plugins/Live/images/file{{ col }}.png" title="{% if action.url|trim is not empty %}{{ action.url|e('html_attr') }}<br/>{% endif %}{{- title -}}"/>
                                 {% elseif action.type == 'outlink' or action.type == 'download' %}
                                     <img class='iconPadding' src="{{ action.icon }}"
                                          title="{% if action.url is defined %}{{ action.url }} - {% endif %}{{ action.serverTimePretty }}"/>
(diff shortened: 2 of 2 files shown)


mzcarlos:
> Also truncates url to 120 chars.
> 
> The tooltip delay is reduced to 100ms. I couldn't reproduce the stuck tooltip issue the delay fixed, so hoping it will be ok.
> 
> Fixes #13991

I tried your commit but it didn't fix the delay. It must be anything else.

However, I noticed something that might help you to find the issue. In the live widget, the country, browser, operating system and device type tooltips still appear immediately. The visitor profile and the viewed pages do not.

Thank you!

diosmosis:
@mzcarlos can you try clearing your browser cache and running the `./console cache:clear` command?

mzcarlos:
@diosmosis the `./console cache:clear` command fixed it! I was clearing the browser cache but didn't know about the server cache. Now it works fine. The title in its own line looks nice too. Thank you very much.

plugins/CoreHome/angularjs/widget/widget.directive.js
```diff
@@ -109,7 +109,7 @@
                             var title = $(this).attr('title');
                             return piwikHelper.escape(title.replace(/\n/g, '<br />'));
                         },
-                        show: {delay: 700, duration: 200},
+                        show: {delay: 100, duration: 200},
```
tsteur: is that needed? We set it to 700 on purpose to simulate browser behaviour a bit more. We don't want them to show up so quick as it can be quite annoying.

plugins/Live/templates/getLastVisitsStart.twig
```diff
@@ -85,7 +85,7 @@
 {{ action.serverTimePretty }}
 {% if action.timeSpentPretty is defined %}{{ 'General_TimeOnPage'|translate }}: {{ action.timeSpentPretty|raw }}{% endif %}
 {%- endset %}
-                                    <img src="plugins/Live/images/file{{ col }}.png" title="{{ action.url|e('html_attr') }}&mdash;{{- title -}}"/>
+                                    <img src="plugins/Live/images/file{{ col }}.png" title="{{ action.url|truncate(120)|e('html_attr') }}<br/>{{- title -}}"/>
```
tsteur: Not sure about the truncate, usually we don't truncate data on hover I think and show full URL.  @mattab

plugins/CoreHome/angularjs/widget/widget.directive.js
diosmosis: Note: the tooltip doesn't come up immediately over hover, might be required to try it locally and see if it's ok or not...

plugins/Live/templates/getLastVisitsStart.twig
mattab: Indeed it is better not to truncate to allow people to see the full value.. it is ok if the value wraps over several lines and actually a good thing 👍

plugins/CoreHome/angularjs/widget/widget.directive.js
diosmosis: Did we get those complaints before? Still think it would be good to try and see if it's an issue. Don't need an answer immediately or anything, though.

plugins/Live/templates/getLastVisitsStart.twig
diosmosis: Removed.

plugins/CoreHome/angularjs/widget/widget.directive.js
mzcarlos: I don't know how this setting may relate to other widget's tooltips. The behaviour of the action tooltips up to the 3.7.0 version made it possible to browse the titles of the pages seen by the user at a glance. With a 700ms delay this becomes very frustrating. As an alternative I found myself opening the "visitor profile" but most of the time I rather prefer the previous tooltip behaviour.

sgiehl: @mzcarlos this is the global tooltip class. So it will have an effect on all tooltips that are not created somewhere else.

tsteur: It is somehow possible to change it only for real time widget? There it should make sense to not use a delay basically

diosmosis: @tsteur updated

plugins/Live/templates/getLastVisitsStart.twig
```diff
@@ -85,7 +85,7 @@
 {{ action.serverTimePretty }}
 {% if action.timeSpentPretty is defined %}{{ 'General_TimeOnPage'|translate }}: {{ action.timeSpentPretty|raw }}{% endif %}
 {%- endset %}
-                                    <img src="plugins/Live/images/file{{ col }}.png" title="{{ action.url|e('html_attr') }}&mdash;{{- title -}}"/>
+                                    <img src="plugins/Live/images/file{{ col }}.png" title="{{ action.url|e('html_attr') }}<br/>{{- title -}}"/>
```
tsteur: Works but when I tested it I didn't have a URL set and then an empty line appears. that should be fine though I suppose (but maybe easy to fix?). 
image

tsteur:
Left a comment, otherwise looks good

diosmosis:
Added if to detect empty actions.

fdellwing:
Why did the truncation get removed again in the progress of this PR?

diosmosis:
@fdellwing the comment thread is here: https://github.com/matomo-org/matomo/pull/14005#discussion_r250837759

dev-101:
Any chance we can customize the tooltip?
I mean, I don't (no, really don't) need URL address in the tooltip. It just makes things cramped. KISS rules.

tsteur:
@dev-101 @fdellwing every one uses it differently and we can't make it right for everyone. The only solution be probably to write a plugin that removes this part through JS or maybe PHP. May not be that easy.

dev-101:
That's why I'm asking ;)

I don't see any filter hooks in there, and no classes to customize it through CSS. Just some hackish solution if possible at all. I would also remove timestamps, again, if I need them > full log is available.

diosmosis:
I can add CSS classes to each line in the tooltip.

dev-101:
Please do :)
Thanks!

PR #14006, opened 2019-01-23
Sparklines do not show correct values if values are close to 1.0.

diosmosis:
The sparkline lib we use converts all values to integers (see the FormatTrait.php file) which means float
numbers that are close to 1.0 or 0.0 will get floored to 1/0. This can happen in the average page generation time report, and cause some values which are, eg, around ~.9 to appear as 0 in the sparkline. To workaround this I made a change to scale the values if a float is found.


core/Visualization/Sparkline.php
```diff
@@ -70,8 +70,26 @@ public function main() {
             }
             $values[] = $value;
         }
-        $sparkline->setData($values);
 
+        $hasFloat = false;
+        foreach ($values as $value) {
+            if (is_float($value)) {
```
tsteur: is there a chance that we are dealing with strings that are floats like `'5.51'`?

diosmosis: I suppose it's possible, will update.

diosmosis: Updated.

tsteur: I would maybe also add an `is_numeric($value)` maybe to avoid `Warning: A non-numeric value encountered in /in/00aDd on line 3`? https://3v4l.org/00aDd

Not sure what can happen... like

`if ($value && is_numeric($value){  $value = $value + 0; if () {is_float($value){...}  }`

Otherwise it seems to work re strings and float comparison

diosmosis: 👍

diosmosis: Updated.

tsteur:
Looks good if tests pass

diosmosis:
Test failing, moving to 3.9

sgiehl (commented):
updated the UI files. But actually I'm not sure if it wasn't on purpose that the sparkline didn't raise fully for quite low values?  ping @mattab / @Findus23 

Otherwise looks good to merge if all related UI tests are passing now

Findus23:
It definitely wasn't intentional by me and I think it looks better now (especially percentages)

Findus23:
Interestingly someone reported a similar sounding issue in 3.10:
https://forum.matomo.org/t/formattrait-php-193-notice-a-non-well-formed-numeric-value-encountered-matomo-3-10-0/33422
@diosmosis Do you have any idea why it could occur?

diosmosis:
@Findus23 from looking at the code I would guess the values have characters that aren't numeric. There's some code there that removes percents and other unit values, maybe it's not working here. I wonder if using `&format_metrics=0` in the sparkline visualization would prevent that...

PR #14011, opened 2019-01-23
update JShrink (followup)

Findus23:
follow-up to #13639 
see #13418

I destroyed the other branch by merging the composer.lock, so I tried new from 3.x-dev, applied the changes in composer.json and ran `composer require JShrink==1.3.1` and `composer remove tedivm/jshrink` which seems to have created a reasonable composer.lock


diosmosis:
Looks like this change causes some failures in TagManager, not sure if they are a problem though, cc @tsteur

PR #14015, opened 2019-01-23
Preserve case for CLDR month/day names

c960657:
Long date formatting for Danish (e.g. in e-mail reports) is wrong.

**Expected:**
Tirsdag den 22. januar 2019 

**Actual:**
Tirsdag den 22. Januar 2019 
(notice the uppercase J)

This bug occurs because month names are ucfirst'ed when imported from CLDR. This is generally not needed – there are two versions, and the stand-alone version already starts with an uppercase letter in the CLDR source data.

Do ucfirst() on formatted date instead.


core/Date.php
```diff
@@ -729,6 +731,10 @@ public function getLocalized($template)
             }
         }
 
+        if ($ucfirst) {
+          $out = Common::mb_strtoupper(mb_substr($out, 0, 1)) . mb_substr($out, 1);
```
sgiehl: Is that really required? If I understand that correctly it should**n't** be the default to uppercase the first letter at all. The formattings from CLDR should define if the standalone version is used or not. And if a month name has a capital first letter or not is part of the translations. Or am I'm missing something?

sgiehl (changes requested):

core/Date.php
sgiehl: Ok. So guess your approach should be good enough for now. To solve the correctly guess we would need to fetch the context transformation info from CLDR (e.g. https://github.com/unicode-cldr/cldr-misc-full/blob/master/main/de/contextTransforms.json)

See http://cldr.unicode.org/translation/capitalization

sgiehl (approved):

PR #14016, opened 2019-01-23
Do not render until form is initialized

c960657:
The email report list page throws a lot of JS errors:

image

The reason is that the add report form is rendered (though not visibly), but it is not initialised until it is opened.


sgiehl (approved):
Good catch :+1:

PR #14017, opened 2019-01-23
Use NumberFormatter to format metric values

### LINKED ISSUE #13994: Format currencies metrics (thousands separator)
Currently under Ecommerce reports such as Products, Sales, Multi-attribution, the currency metrics are not formatted and are for example missing the thousands separator. Expected to have all currency columns (such as "Product Revenue", "Average Price") to be formatted correctly. (Currently in Ecommerce > Overview, currency metrics are correctly formatted).

See screenshot below showing the non formatted currency metrics in Ecommerce and Goal reports.

### Ecommerce reports

currency ecommerce
 
### Goal reports and metrics

goals

sgiehl:
This will change all metric formatting's to use the local formats of NumberFormatter.
Not yet sure if that will have any side effects besides the change of all exports.

fixes #13994


Findus23:
In a quick test German continues to use `.` as a thousand seperator and English still shows `,`

diosmosis:
Money values are now formatted as `$50` instead of `$ 50`. Is this intentional? I can imagine in an edge case it could be seen as a break in backwards compatibility.

sgiehl:
kind of expected, as the formatter now uses the formats defined by CLDR. That means in some languages there will now be a space between the currency and the number in others not. Also the currency symbol is pre- or appended depending on the language.

sgiehl:
updated the branch and all test files. should be ready for a final review and merge...

diosmosis:
Looked through the test files and noticed one possible issue: https://github.com/matomo-org/matomo/pull/14017/files#diff-9f87ae8d49751f22224330816574b5a6L10888

It's printing `$233.70` as `$233.7`.

plugins/PrivacyManager/tests/System/expected/test_allSites__Live.getLastVisitsDetails_year.xml
```diff
@@ -11427,7 +11427,7 @@
 				<icon>plugins/Morpheus/images/ecommerceOrder.png</icon>
 				<iconSVG>plugins/Morpheus/images/ecommerceOrder.svg</iconSVG>
 				<title>Ordered (205)</title>
-				<subtitle>$ 233.70 revenue - 1 items: My product 205)</subtitle>
+				<subtitle>$233.7 revenue - 1 items: My product 205)</subtitle>
```
diosmosis: Is it correct that it's printing `$233.7` instead of `$233.70`?

diosmosis:
Saw one possible issue, everything else looks good.

sgiehl:
@diosmosis that was actually a general problem with the numberformatter. Should be fixed in https://github.com/matomo-org/matomo/pull/14017/commits/8a6c9502b6143dfea3cb635469962cefb654b17d

PR #14019, opened 2019-01-24
handle background color of opt out if font is almost white

### LINKED ISSUE #14008: Opt-out iframe builder broken
There are two problems with the iframe builder right now.

~~1. If you choose `em` instead of `px` and enter a float number, this number is not getting added to the iframe code. It only accepts integers.~~

2. If you choose `white` as font color the background color is still transparent (so the iframe seems empty), but the builder shows `black` in the color picker.

fdellwing:
If no background color is set, the background color is transparent. But the whole site background color is white, so if you choose white or a color that is nearly white as font color the iframe appears so be empty.

This commit implements a new function that tests if a hex color is nearly white and if yes the code will change the background color to a grey color.

Fixes #14008

### DIFF
--- a/plugins/PrivacyManager/angularjs/opt-out-customizer/opt-out-customizer.controller.js
+++ b/plugins/PrivacyManager/angularjs/opt-out-customizer/opt-out-customizer.controller.js
@@ -30,9 +30,9 @@
         vm.onUpdate = function () {
             if (vm.piwikurl) {
                 if (vm.backgroundColor === '' && vm.fontColor !== '' && vm.nearlyWhite(vm.fontColor.substr(1))) {
-                    vm.backgroundColor = '#4d4d4d';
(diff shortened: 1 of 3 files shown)


sgiehl:
Guess that makes it a bit better, but I'm not having a strong opinion on that one. 
@tsteur @mattab @Findus23  Does anyone have an opinion here? Would suggest to merge it, even though there is some more space left for improvements...

tsteur:
Not really any strong opinion

mattab:
Thanks for the PR @fdellwing - the idea and improvement looks good to me. Would it be maybe doable to have a UI test to test for this behaviour?

tsteur:
Maybe not needed a UI test for this?

mattab:
a UI test is not needed

PR #14023, opened 2019-01-25
Allow throwing exceptions for HTTP error codes to better check query param input.

### LINKED ISSUE #13752: fatal error triggered under some circumstances
The fatal error: `Neither the property "getRows" nor one of the methods "getRows()", "getgetRows()"/"isgetRows()" or "__call()" exist and have public access in class "Piwik\DataTable\Map".
in plugins/CoreVisualizations/templates/_dataTableViz_htmlTable.twig line 21`

Can be reproduced at https://demo.matomo.org/index.php?forceView=1&viewDataTable=..%2fWEB-INF%2fweb.xml%3bx%3d&module=MediaAnalytics&action=get&idSite=3&period=day&date=2018-10-16,2018-11-14&showtitle=1&random=84&columns=sum_total_time_watched&colors={%22backgroundColor%22%3A%22%23ffffff%22%2C%22lineColor%22%3A%22%23162c4a%22%2C%22minPointColor%22%3A%22%23ff7f7f%22%2C%22maxPointColor%22%3A%22%2375bf7c%22%2C%22lastPointColor%22%3A%22%2355aaff%22%2C%22fillColor%22%3A%22%23ffffff%22}

It is only triggered under edge case (ie. security scanner), but will be good to fix since we get this emailed to us quite a bit.

diosmosis:
Added two new exceptions `HttpCodeException` and `BadRequestException` that allows changing the HTTP code returned for an exception (in both controller & API).

This allows us to avoid certain failures caused by invalid input and correctly report the error, w/o triggering a 500 HTTP code (that could potentially also create an email alert?).

Used in ScheduledReports API to check for multiple periods.

Fixes #13752


plugins/CoreVisualizations/Visualizations/HtmlTable.php
```diff
@@ -10,6 +10,7 @@
 
 use Piwik\API\Request as ApiRequest;
 use Piwik\Common;
+use Piwik\Http\BadRequestException;
```
sgiehl: guess that could be removed again

sgiehl (changes requested):

sgiehl:
@diosmosis  this one needs a rebase / merge to solve the conflicts

core/API/ResponseBuilder.php
```diff
@@ -135,6 +136,10 @@ public function getResponseException($e)
         $e       = $this->decorateExceptionWithDebugTrace($e);
         $message = $this->formatExceptionMessage($e);
 
+        if ($this->sendHeader && $e instanceof HttpCodeException) {
```
tsteur: would you need to check if a code is set?

core/ExceptionHandler.php
```diff
@@ -70,6 +71,10 @@ public static function dieWithCliError($exception)
      */
     public static function dieWithHtmlErrorPage($exception)
     {
+        if ($exception instanceof HttpCodeException) {
```
tsteur: same here maybe we better check if a code is set?

tsteur:
Left 2 comments, otherwise good to merge when tests pass 👍

PR #14026, opened 2019-01-27
Change session cookie name to MATOMO_SESSID

diosmosis:
Only done in the Login controller so should not be done on every request.

Tested w/ a PHP INI config that set the cookie path to a subdirectory and was able to verify this fixed the issue.

h/t to @GreenReaper for identifying the issue (at least it seems likely this is the case).

### DIFF
--- a/core/Session.php
+++ b/core/Session.php
@@ -19,16 +19,11 @@ use Zend_Session;
  */
 class Session extends Zend_Session
 {
-    const SESSION_NAME = 'PIWIK_SESSID';
+    const SESSION_NAME = 'MATOMO_SESSID';
 
     public static $sessionName = self::SESSION_NAME;
 
     protected static $sessionStarted = false;
-
-    /**
-     * @var string
-     */
-    private static $originalCookiePath;
 
     /**
      * Are we using file-based session store?
@@ -86,7 +81,6 @@ class Session extends Zend_Session
         // the session data won't be deleted until the cookie expires.
         @ini_set('session.gc_maxlifetime', $config->General['login_cookie_expire']);
 
-        self::$originalCookiePath = ini_get('session.cookie_path');
         @ini_set('session.cookie_path', empty($config->General['login_cookie_path']) ? '/' : $config->General['login_cookie_path']);
 
         $currentSaveHandler = ini_get('session.save_handler');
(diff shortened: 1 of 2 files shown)


tsteur:
@diosmosis need to investigate but for some reason I cannot log in anymore. Can you still log in on this PR?

GreenReaper:
Are you using OAuth, like a LiveJournal logon or suchlike? GitHub deprecated it; you will need to do a reset to your email address, or contact them.
________________________________

tsteur:
No custom login plugin or anything. I always get this error:

image

tsteur:
Just double checked... I can log in with 3.x-dev but not this PR

tsteur:
image

and it goes into that method

tsteur:
maybe don't clear existing cookie if form was posted? not sure though...

diosmosis:
Can you check what the Set-Cookie is in the module=Login&action=login call? And if there is a PIWIK_SESSID on any path in the chrome dev tools?

diosmosis:
Oh maybe that's it...

tsteur:
image
 
image

diosmosis:
Updated the PR to check for non-POST requests.

diosmosis:
Oh nvm, this is a dumb mistake. Will push a fix now.

diosmosis:
@tsteur can you pull and try now?

tsteur:
Now I can log in @diosmosis 

btw the POST check didn't really help... The first time I submit the form it would show the error message again, and only the second time I submit the form again it would work

diosmosis:
> btw the POST check didn't really help... The first time I submit the form it would show the error message again, and only the second time I submit the form again it would work

The bug was because the final value was '/', but the config option would be empty, so it always removed the cookie. The POST check is definitely helpful since it might've lead to some weird bugs (though not sure exactly what)...

tsteur:
Just removed the POST check and still worked

just random thought... would have been an alternative been to rename the cookie name?

tsteur:
> The POST check is definitely helpful since it might've lead to some weird bugs (though not sure exactly what)...

👍

diosmosis:
> just random thought... would have been an alternative been to rename the cookie name?

You mean PIWIK_SESSID => MATOMO_SESSID or something? Think so... not tested though.

tsteur:
Might be good to change it to Matomo* anyway... can do it in a next release though. Would have expected this should fix any already set cookie

diosmosis:
> Would have expected this should fix any already set cookie

Just in the case of session.cookie_path != login_cookie_path, not every case. MATOMO_SESSID would be simpler.

tsteur:
I suppose it looks good to merge though. Hoping there are no side effects :)

PR #14034, opened 2019-01-28
When dispatch is disabled via a constant, it should not be dispatched

tsteur:
Eg when Matomo is not installed, it would still dispatch the request in https://github.com/matomo-org/matomo/blob/3.8.1-b1/plugins/Installation/Installation.php#L111 even when `PIWIK_ENABLE_DISPATCH` is disabled. 

Will set it to WIP for now as I'm not sure if we want to have this actually merged or not.


tsteur:
Not sure we can easily add a test for this? Basically the idea is to throw the exception when dispatch is disabled.

sgiehl (approved):
Code looks good. Not sure if we need to add a test for that, but can't we simply check if the response is empty when no config file exists?

tsteur:
Not sure it's so easily possible to test as we need to bootstrap Matomo and set `PIWIK_ENABLE_DISPATCH` to `false`. Could maybe embed a PHP file in the test directory that bootstraps it. It should actually throw an exception in this case. Might quickly have a look later.

tsteur:
added a test @sgiehl

PR #14037, opened 2019-01-28
Make more links clickable in New releases notification emails

AlainRnet:
Links in the (new?) notification email are no longer clickable.

### DIFF
--- a/plugins/CoreUpdater/templates/_updateCommunicationEmail.twig
+++ b/plugins/CoreUpdater/templates/_updateCommunicationEmail.twig
@@ -3,7 +3,7 @@
 <p>{{ 'CoreUpdater_ThereIsNewVersionAvailableForUpdate'|translate }}</p>
 
 <p>{{ 'CoreUpdater_YouCanUpgradeAutomaticallyOrDownloadPackage'|translate(latestVersion) }}<br/>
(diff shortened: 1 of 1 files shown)


Findus23:
Seems to be related to #13908 
@diosmosis Is there any reason (apart from the fact that most e-mail-clients make plain URLs clickable) for this?

diosmosis:
No reason, the HTML can be added, though perhaps w/ the appropriate rel= in case the email client doesn't add one itself.

plugins/CoreUpdater/templates/_updateCommunicationEmail.twig
```diff
@@ -3,16 +3,16 @@
 <p>{{ 'CoreUpdater_ThereIsNewVersionAvailableForUpdate'|translate }}</p>
 
 <p>{{ 'CoreUpdater_YouCanUpgradeAutomaticallyOrDownloadPackage'|translate(latestVersion) }}<br/>
-{{ host }}index.php?module=CoreUpdater&action=newVersionAvailable
+<a href="{{ host }}index.php?module=CoreUpdater&action=newVersionAvailable">{{ host }}index.php?module=CoreUpdater&action=newVersionAvailable</a>
```
tsteur: the host and the other variables need some escaping?

tsteur: maybe `|safelink` but not sure...

Issue #12988, opened 2018-05-24
Separate core plugins and "user" plugins


crazy-max:
Hi,

I'm here about an issue concerning plugins path. This comes from matomo-org/docker#89 and matomo-org/docker#97.

Today all plugins (including core ones) are written in `/var/www/plugins` folder. On a [load balanced environment](https://matomo.org/faq/new-to-piwik/faq_134/) you said to add `multi_server_environment=1` in the config and next you said :

> it will then not allow the installation of a plugin via the UI (or the plugin would be only installed on one server), or it will not allow config file changes from the UI (as the config file would be updated on one server only)

It's a major issue for me and i think this can be solved with a new feature. We should separate core plugins (included in piwik archive) and user plugins. Nextcloud uses this trick to be able to persist plugins folder on Docker. On Nextcloud you can have this config :

```
array (
    0 =>
    array (
      'path' => '/var/www/apps',
      'url' => '/apps',
      'writable' => false,
    ),
    1 =>
    array (
      'path' => '/data/userapps',
      'url' => '/userapps',
      'writable' => true,
    ),
  ),
```

This way user apps will be written in `/data/userapps` and plugins are listed from `/var/www/apps` (core) and `/data/userapps` (user).

And it's an issue on Docker too (specially on Swarm or k8s). We cannot persist the folder `/var/www/plugins` because on image update this folder will have core plugins from the previous version of the image.

tsteur:
Sorry I'm not quite getting how a different location would solve a problem? I doubt we will make this change as it would be quite a big refactoring going on for quite a while. How is plugins different to core, or third party plugins? Can it not persist writeable directories? FYI: Even certain paths like `misc/user`, `/js` and other paths need to be writeable will they be a problem as well? 

I don't know how that docker is being built, but can you start a docker image and deploy the code from a git repository or something? Or also persist the user plugins or load them on demand etc?

mattab:
We won't do this AFAIK

PR #14043, opened 2019-01-29
Don't hardcode plugins directory

tsteur:
Instead use the method that defines the plugin dir. No logic should change here. Doing this in preparation to potentially provide a way to define multiple plugin directories. But for this we first need to make sure that all code is using the same method when looking for plugin directory.

### DIFF
--- a/core/API/Proxy.php
+++ b/core/API/Proxy.php
@@ -13,6 +13,7 @@ use Exception;
 use Piwik\Common;
 use Piwik\Context;
 use Piwik\Piwik;
+use Piwik\Plugin\Manager;
 use Piwik\Singleton;
 use ReflectionClass;
 use ReflectionMethod;
@@ -437,7 +438,7 @@ class Proxy extends Singleton
     private function includeApiFile($fileName)
     {
         $module = self::getModuleNameFromClassName($fileName);
-        $path = PIWIK_INCLUDE_PATH . '/plugins/' . $module . '/API.php';
+        $path = Manager::getPluginsDirectory() . $module . '/API.php';
 
         if (is_readable($path)) {
             require_once $path; // prefixed by PIWIK_INCLUDE_PATH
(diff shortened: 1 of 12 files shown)


diosmosis:
Not sure if these need to change, but seeing the `/plugins/` in other places like:

* core/testMinimumPhpVersion.php
* core/ReportRenderer/Pdf.php
* core/Tracker/Response.php
* core/FrontController.php
* plugins/CorePluginsAdmin/PluginInstaller.php
* plugins/CorePluginsAdmin/Controller.php
* LanguagesManager files
* plus other places

tsteur:
cheers @diosmosis added it to more places where needed (didn't add it when they included/required specific files from a plugin)

Findus23:
I guess this will (someday in the future) help with https://github.com/matomo-org/matomo/issues/12988 which allows easier docker images and similar things.

tsteur:
Yes that can help with that. It may partially help us to at some point move the plugins directory somewhere else. May goal be likely to have it later possible to add more plugins directories while leaving the core plugins in the current place. After this exercise it seems likely too difficult though and I might fallback to symlinks instead. To be seen.

Findus23:
@tsteur I think this docker image does use symlink tricks to get around this issue:
https://github.com/crazy-max/docker-matomo
https://github.com/crazy-max/docker-matomo/blob/master/entrypoint.sh

diosmosis:
> didn't add it when they included/required specific files from a plugin

Will that be an issue if the plugins are located outside of PIWIK_INCLUDE_PATH . '/plugins'?

tsteur:
> Will that be an issue if the plugins are located outside of PIWIK_INCLUDE_PATH . '/plugins'?

If all plugins were moved outside then yes. For my use case I'm not really planning on the possibility to move plugins in another directory. Only to pick up non-core plugins from a different directory. The core plugins be always in that path.

diosmosis:
> If all plugins were moved outside then yes. For my use case I'm not really planning on the possibility to move plugins in another directory. Only to pick up non-core plugins from a different directory. The core plugins be always in that path.

Ah, I see, makes sense 👍

PR #14049, opened 2019-01-29
One Click Install UI test

diosmosis:
This PR can be reviewed but should not be merged until https://github.com/matomo-org/matomo-package/pull/93 is merged (and the git clone is changed in this PR).

Changes:
* Added new fixture LatestStableInstall that installs the latest stable version and packages the git checkout.
* Added a UI test for OneClickInstall.

h/t to @Findus23 for original idea

Todo:

* [ ] ensure core:archive command works after update


tests/UI/specs/OneClickUpdate_spec.js
```diff
@@ -0,0 +1,55 @@
+/*!
+ * Piwik - free/libre analytics platform
+ *
+ * Screenshot tests for main, top and admin menus.
```
sgiehl: The description is incorrect and maybe we should directly use a `Matomo` file header for new files...

sgiehl (changes requested):
Left some minor comments. Otherwise test seems to work :+1:

diosmosis:
Applied PR feedback.

sgiehl:
@diosmosis did you remove the `CoreUpdateDb` ui files on purpose? The spec file is still there and so the tests are failing

diosmosis:
Yes, thought I removed the spec file too, must have forgot. Will do so.

sgiehl (approved):

tests/resources/install-matomo.php
```diff
@@ -0,0 +1,226 @@
...
+use Piwik\Tracker\Cache;
+use Piwik\Config;
+use Piwik\Db;
+use Piwik\DbHelper;
+use Piwik\Option;
+use Piwik\Plugins\LanguagesManager\API as APILanguageManager;
+use Piwik\Updater;
+use Piwik\Plugins\CoreUpdater;
+
+$subdir = $argv[1];
```
tsteur: Maybe we could add a simple check to make sure it's not run through a web request? Accessing `$argv` should probably already make sure of that though. Not sure if there's any way this could be defined through web. Luckily Register Globals isn't a thing anymore :) 
Just thinking of people who deploy Matomo through git and where those files aren't removed.

Could add the check from `Common::isPhpCliMode` just to be safe

tsteur:
Looks good as well 👍  Didn't look in detail but as long as the tests work should be all fine.

PR #14051, opened 2019-01-30
Support multiple plugin paths

tsteur:
Adjusted most usages except for `LanguagesManager.getAvailableTranslationsInfo` (or similar).

refs https://github.com/matomo-org/matomo/issues/12988 

Basically, we already support a `bootstrap.php` in the root folder in which a user could place eg 

```
$GLOBALS['MATOMO_PLUGIN_DIRS'] = array(
    array(
        'pluginsPathAbsolute' => realpath(PIWIK_INCLUDE_PATH . 'tests/resources/custompluginsdir'),
        'webrootDirRelativeToMatomo' => '../../resources/custompluginsdir'
    ),
);
```

to pick up plugins from other directories. Each dir should contain plugins... simple test works so far but need to check further.

Out of scope: Same plugin in multiple folders. No support for themes.


tsteur:
Not sure how to best test this... I reckon we would need to place like a dummy plugin into a `tests/resources` folder and see if Matomo still works and picks it up? The problem is the functionality is used in many places... I assume when regular tests still work it is also a sign that everything seems to work nicely... to be seen.

core/Plugin/Manager.php
```diff
@@ -316,18 +318,71 @@ public function isPluginLoaded($name)
...
+     * @return array
+     */
+    public static function getPluginsDirectories()
+    {
+        $dirs = array(PIWIK_INCLUDE_PATH . '/plugins/');
+        if (!empty($GLOBALS['MATOMO_PLUGIN_DIRS'])) {
+            $GLOBALS['MATOMO_PLUGIN_DIRS'] = array_map(function ($dir) {
+                return rtrim($dir, '/') . '/';
+            }, $GLOBALS['MATOMO_PLUGIN_DIRS']);
+            $dirs = array_merge($dirs, $GLOBALS['MATOMO_PLUGIN_DIRS']);
```
diosmosis: Is it possible to initialize `$GLOBALS['MATOMO_PLUGIN_DIRS']` from an environment variable? Could be useful for some use cases.

diosmosis:
> Not sure how to best test this... I reckon we would need to place like a dummy plugin into a tests/resources folder and see if Matomo still works and picks it up?

I think this works, the dummy plugin could be generated in UITestFixture. I think it would be good to make it a tracker plugin too and generate it before loading the OmniFixture dump. Then if the UI tests pass, we have a reasonable confirmation it works.

diosmosis:
Any need to update the changelog for this?

tsteur:
> Any need to update the changelog for this?

I'll add one to let developers know to not hard code plugins directory 👍 

> I think this works, the dummy plugin could be generated in UITestFixture. I think it would be good to make it a tracker plugin too and generate it before loading the OmniFixture dump. Then if the UI tests pass, we have a reasonable confirmation it works.

Just wondering what plugins to add there or what functionality... guess will just pretty much copy the example plugins... was wondering if I should just move all example plugins in a custom directory... problem is there are some links pointing to those example directories and not sure where all those places are. also it is kind of nice to have the example plugins next to the other plugins.

note to myself: Need to check that classes with `Piwik\Plugins\...` can still be autoloaded correctly as some features rely on extracting the plugin name from the class name.

diosmosis:
I was just thinking of creating a simple plugin to the custom directory. I don't think it needs to be that complex, since you're just changing plugin paths here. I wouldn't it to expect it to affect anything other than what plugins are loaded/recognized. I guess we could add a simple report/API method? Maybe not something that gets shown in the dashboard, since those already have so many widgets...

tsteur:
fyi: configuration will need to be a bit more complex in some cases... eg we may require to define `webrootDirectory` etc. when plugins folder is outside matomo directory...

core/Application/Kernel/PluginList.php
```diff
@@ -42,7 +42,8 @@ class PluginList
         'ExampleReport',
         'MobileAppMeasurable',
         'Provider',
-        'TagManager'
+        'TagManager',
+        'CustomDirPlugin'
     );
```
diosmosis: @tsteur If you're trying to get this to load in the test, adding it to extraPluginsToLoad in Fixture should work.

tsteur: @diosmosis I'm trying to fix this test: https://travis-ci.org/matomo-org/matomo/jobs/489810408#L1048

as in UI test I did `testEnvironment.pluginsToLoad = ['CustomDirPlugin'];`... so hoping this way it will be maybe updated or work?

tsteur: @diosmosis didn't fix the error. do you have any idea how to make sure it updates the plugin? I wouldn't even really have expected an update.

diosmosis: That might do it, they should get added in TestingEnvironmentManipulator.php

diosmosis: The customdirplugin has a visit dimension, I guess that triggers the update. I think the problem is that UITestFixture uses the OmniFixture dump, but that of course wouldn't load CustomDirPlugin. UITestFixture.php calls Fixture::updateDatabase(), but I guess that isn't working here...

If you set UITestFixture::$extraPluginsToLoad to `['CustomDirPlugin']`, does that work?

diosmosis: Oh I think I see the problem now. The plugin has to be loaded before the fixture is setup, but setup-fixture command is made to be the first thing that is done in the test (see screenshot-testing/support/app.js).

tests/PHPUnit/Fixtures/UITestFixture.php
```diff
@@ -57,6 +57,8 @@ public function __construct()
 
     public function setUp()
     {
+        $this->extraPluginsToLoad[] = 'CustomDirPlugin';
```
diosmosis: Think you'll have to add it to top of performSetUp()

tsteur: thank you @diosmosis will try

tsteur: didn't help... https://travis-ci.org/matomo-org/matomo/jobs/489864454#L1203  will need to see tmrw and otherwise just remove all possible updates from that plugin

ddrjm:
Referencing #14307

PR #14057, opened 2019-01-30
Visitor Map tooltip shows %s for "No Visit"

### LINKED ISSUE #14054: Visitor Map tooltip shows %s
Just a quick note as I just noticed it (sorry if it has already been reported before)

The visitor map (e.g. on the dashboard of demo.matomo.org) shows the correct tooltip for countries with visitors but `%s No Visit` else.
grafik

tsteur:
fix #14054

fixed the issue for me locally.


Findus23:
:+1: Seems to work locally for me

PR #14058, opened 2019-01-30
New setting to disable Db version check, and do not show update screen when auto update disabled

tsteur:
See comments in code...

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -665,14 +665,6 @@ enable_internet_features = 1
 ; By setting this option to 0, it will disable the "Auto update" feature
 enable_auto_update = 1
 
-; By setting this option to 0, the check that the DB schema version matches the version of the source code will be no longer performed.
-; Thus it allows you to execute for example a newer version of Matomo with an older Matomo database version. Please note
-; disabling this setting is not recommended because often an older DB version is not compatible with newer source code.
-; If you disable this setting, make sure to execute the updates after updating the source code. The setting can be useful if
-; you want to update Matomo without any outage when you know the current source code update will still run fine for a short time
-; while in the background the database updates are running.
-enable_db_version_check = 1
-
 ; By setting this option to 0, no emails will be sent in case of an available core.
 ; If set to 0 it also disables the "sent plugin update emails" feature in general and the related setting in the UI.
 enable_update_communication = 1
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -665,6 +665,14 @@ enable_internet_features = 1
 ; By setting this option to 0, it will disable the "Auto update" feature
 enable_auto_update = 1
 
+; By setting this option to 0, the check that the DB schema version matches the version of the source code will be no longer performed.
+; Thus it allows you to execute for example a newer version of Matomo with an older Matomo database version. Please note
+; disabling this setting is not recommended because often an older DB version is not compatible with newer source code.
+; If you disable this setting, make sure to execute the updates after updating the source code. The setting can be useful if
+; you want to update Matomo without any outage when you know the current source code update will still run fine for a short time
+; while in the background the database updates are running.
+enable_db_version_check = 1
+
 ; By setting this option to 0, no emails will be sent in case of an available core.
 ; If set to 0 it also disables the "sent plugin update emails" feature in general and the related setting in the UI.
 enable_update_communication = 1
--- a/plugins/CoreUpdater/CoreUpdater.php
+++ b/plugins/CoreUpdater/CoreUpdater.php
@@ -14,6 +14,7 @@ use Piwik\Common;
 use Piwik\Filesystem;
 use Piwik\FrontController;
 use Piwik\Piwik;
+use Piwik\SettingsPiwik;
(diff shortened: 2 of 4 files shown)


config/global.ini.php
```diff
@@ -665,6 +665,14 @@
 ; By setting this option to 0, it will disable the "Auto update" feature
 enable_auto_update = 1
 
+; By setting this option to 0, the check that the DB schema version matches the version of the source code will be no longer performed.
+; Thus it allows you to execute for example a newer version of Matomo with an older Matomo database version. Please note
+; disabling this setting is not recommended because often an older DB version is not compatible with newer source code.
+; If you disable this setting, make sure to execute the updates after updating the source code. The setting can be useful if
+; you want to update Matomo without any outage when you know the current source code update will still run fine for a short time
+; while in the background the database updates are running.
+enable_db_version_check = 1
```
tsteur: When deploying a new version of Matomo, I'm currently always editing the `FrontController.php` and adding a `return` statement manually which I then remove after the deployment. As mentioned in the description disabling this version check helps me to update Matomo without any downtime. For most updates, even when there are DB updates, Matomo still continues to work nicely. It might sometimes only affect a few small features such as Goals or so that would be broken for a very short time while performing the DB updates...

Instead of editing the frontcontroller file it would be nice to have a config. 

it's not 100% needed though and we can just continue to edit the frontcontroller file instead.

plugins/CoreUpdater/CoreUpdater.php
```diff
@@ -52,6 +53,10 @@ public static function getComponentUpdates(PiwikCoreUpdater $updater)
 
     public function dispatch()
     {
+        if (!SettingsPiwik::isAutoUpdateEnabled()) {
```
tsteur: About this one I'm not 100% sure. I think it is excepted to not show the "Please update screen" when auto update is disabled since a user would perform the update through the console?

The problem is that when we update code eg on Cloud, we always have to add a `return;` statement here hard coded during deployment because otherwise for short time the update screen would be visible. We don't want this screen to be visible though since we're performing the DB update through the console...

config/global.ini.php
tsteur: It also allows me to update first the DB before updating the code... cause often new DB updates work fine with older versions of Matomo (eg only new tables added, or new columns that don't affect anything else).

So also allows me to update Matomo this way without downtime.

diosmosis: I wonder if we should instead just make sure new versions of matomo are compatible w/ older matomo's? Would make it easier to reach continuous deployment (in the distant future). We could perhaps find some way to run tests w/ the new matomo, using the old db schema (run updates up to before the new version?).

I think this sort of BC is doable, though it would be hard to say for sure whether this is possible in every case, so maybe the setting is needed. But it might be nice to be forced to make sure the schema changes aren't required for new code.

plugins/CoreUpdater/CoreUpdater.php
diosmosis: Makes sense to me.

config/global.ini.php
tsteur: Ideally Matomo would know whether it is compatible with the DB version or not but may be hard to know. Sometimes you can somewhat safely apply the DB updates afterwards, sometimes before, ... and then update source code before or afterwards... but not sure how we could manage that. Always having DB versions compatible may be quite hard and even harder to test.

diosmosis: It wouldn't be super simple, but we could have a build we run right before releasing that would install the previous version, install the new code w/o updating the database, then run all the tests. (Or maybe just UI/system tests.) Though having this type of compatibility is a new idea, so it's definitely hard to see if it's doable or not. I think it's something worth keeping in mind though. Trying to achieve it will eventually let us know how doable it is.

diosmosis: Well the idea would be, right before releasing 3.8.0 (or even a beta/rc), we would run the tests which would:

1) setup matomo 3.7.0 (or eg, 3.7.1 if there was one). this includes code + DB.
2) install matomo 3.x-dev code on top, without modifying the database (and disabling auto updates)
3) run the system/UI tests for 3.x-dev using the 3.7.0 DB.

If it's done every time, we'd know that, eg, 3.8.0 can be used w/o updating 3.7.1. Then we'd run it again when 3.8.1 is out, and we'd know that 3.8.1 can be used w/ 3.8.0 DB, and I think the transitive property would apply, so we'd know that the code would work w/ 3.7.1. Actually, I guess we could, say, add a check for a table in 3.8.0, then remove it in 3.8.1 and the tests would say it's fine going from 3.7.0 to 3.8.1 when it wouldn't be...

Maybe this is something that would mainly be useful for cloud? I guess we don't really need automated tests for that, though, just a focus on making sure DB changes are BC...

sgiehl:
Wondering if it might be better to solve that using `DI`, to avoid blowing up the config with more values that are typically unused/unchanged?

tsteur:
I changed it to `DI`. I think it might be interesting for quite a few users whenever they want to update without downtime basically. But it can be then also documented to configure it through DI or so.

PR #14063, opened 2019-02-01
Indent actions belonging to a pageview

### LINKED ISSUE #12894: More compact display of Contents impressions and Contents clicks in the Visitor Log
When using Content Tracking, by design there will often be 5-10-15 different elements that became visible. Currently they are displayed each on their own line such as below screenshot.

visitor log contents

Maybe we can make the visitor log display more compact?

For example maybe we could combine all "content impressions" (respectively "content clicks") that have also the same content name. For example the following 4 content impressions:

```
* preview-liveaboard - MV Ocean Quest
* preview-liveaboard - Coral Sea Dreaming
* preview-liveaboard - MV Scubapro III
* preview-liveaboard - MV Scubapro II
```

and change it to:
`* preview-liveaboard -  MV Ocean Quest, Coral Sea Dreaming, MV Scubapro III, MV Scubapro II`.
* the tooltip could be the same as the MV Ocean Quest's tooltip (ie. we pick the first). 
* on click, it would expand and show the content impressions on 4 lines.

thoughts?

### LINKED ISSUE #13136: Visitor log: Group all actions within a pageview together
In Matomo 3.0 we introduced a new `pageview id` so we can identify which events etc were performed within a specific page view. This is especially useful when users have multiple tabs open and events or tracking requests are sent from both pages at the same time.

Ideally, we would "group" all events that were done under a specific pageview together. 

For example, we could intent all events/form tracking requests/content tracking requests/ ... if we can identify it to a particular page view. A page view with all the related actions would then be collapsible so that it is easy to hide all actions that were part of a certain page view. This is especially useful when having many actions within one visit or page view. We could for example auto collapse all actions when there are > 10 actions for one page view. In this case we would show eg the first 3 actions, followed by "show all" actions.

image

(just as a rough example, we would need to show how to visualize and word things, how to show the plus/minus icon to collapse etc)

We would also need to make this work for our tracking plugins like form analytics. The idpageview  should be tracked there already. Unfortunately, I noticed we do not track it yet for media analytics and we would need an update for this (maybe as part of Matomo 4 release).

We might still need the possibility to not merge it? Not sure... as sometimes it may still be useful to see the actions one after another how they happened and not grouped by page view.

When someone doesn't track page views for example, or for old visits that don't have the page view ID we would just show it in the regular way.

diosmosis:
This will have to change when the visitor timeline is put in.

Fixes #13136 
Fixes #12894


diosmosis:
Merged w/ 3.x-dev, and here's what it looks like w/ a small visit:

image

and a large visit:

mamp_8888_index php_module corehome action index idsite 1 period day date 2019-02-06 updated 3

and for content actions:

image

image

sgiehl:
@diosmosis  I've merged in the latest 3.x-dev changes. But it seems it's currently broken. Always seeing the error `No entry or class found for 'Live.pageViewActionsToDisplayCollapsed'`. Guess that's caused here: https://github.com/matomo-org/matomo/pull/14063/files#diff-541de4bc6a9bb0d9cad5f733b4b853eeR122
Did you maybe forget to commit/add something?

diosmosis:
@sgiehl added the missing file.

sgiehl:
@diosmosis Seems the visitor profile is broken now. Getting this error when opening one:
```
Variable "actionGroups" does not exist.
in ./plugins/Live/templates/_actionsList.twig line 2            
```

diosmosis:
@sgiehl updated to work in visitor profile. not sure if the implementation is the best or not, let me know if you see room for improvement.

tsteur:
image
not sure if colours were adjusted in this or another PR but seeing there are icons from premium features in different colour. Is it better to fix this in core or in the premium features?

diosmosis:
@tsteur The css class used is determined by the plugin due to how the template is defined, so better in the plugin i'd say.

tsteur:
when I click on "show more actions" it doesn't show more actions. 

image

Also I have a visit where it sometimes shows the "show more link" after 7 actions within a pageview, and sometimes only after 20 shows the "show more link":
image

In general it looks a bit "bloated" I wonder if some additional spacing might help? We could maybe also ask our designers to tweak it maybe.

image

Also I see action images were changed to SVG? that's a problem cause it breaks the mobile app and possibly other consumers. They are supposed to return actual images. Maybe in Matomo we can use SVG but the API should still return proper images. I suppose this was maybe changed in another PR.

tsteur:
> @tsteur The css class used is determined by the plugin due to how the template is defined, so better in the plugin i'd say.

@diosmosis OK I will check to see how to maybe detect the Matomo version and then change the colour of the icon depending on the version cause for older installations we still want to show the black color

tsteur:
also just seeing for some reason it doesn't show form interactions within the page view. Is that possible?

image

diosmosis:
@tsteur 

> when I click on "show more actions" it doesn't show more actions.

I assume you did a hard reload? I've seen this issue, but should have been working... same for the show more link for 7 actions. Then again, maybe the selector is using the core `.action`... will look.

> also just seeing for some reason it doesn't show form interactions within the page view. Is that possible?

If there's no pageview ID set for the action (or the wrong pageview ID is set), then it won't be grouped. Are you looking at the demo data? pageview IDs aren't always set correctly there.

tsteur:
Another minor issue probably also from the other PR. The goal icon is not quite aligned on the left with the icons from the timeline

image

tsteur:
@diosmosis could we add `color: #999` to `.action-list-action-icon {}` then it would just render the color of the icons fine for all premium features as well and wouldn't need to define it individually in each premium feature.

tsteur:
> I assume you did a hard reload? I've seen this issue, but should have been working... same for the show more link for 7 actions. Then again, maybe the selector is using the core .action... will look.

I've done a hard reset but doesn't work. They don'T have the `.action` AFAIK but the `action-list-action-icon`.

> If there's no pageview ID set for the action (or the wrong pageview ID is set), then it won't be grouped. Are you looking at the demo data? pageview IDs aren't always set correctly there.

I checked and it was fetching the idpageview but not returning it. I will release a new version of form analytics 👍 

Overall love how much better it is with the grouping 👍 💯

diosmosis:
@tsteur Also I see action images were changed to SVG? that's a problem cause it breaks the mobile app and possibly other consumers. They are supposed to return actual images. Maybe in Matomo we can use SVG but the API should still return proper images. I suppose this was maybe changed in another PR.

This was done in a previous PR, the action images were turned into SVG, but we're still using `` links (though might be better to eventually embed svgs).

tsteur:
Code looks otherwise straight forward 👍

tsteur:
> This was done in a previous PR, the action images were turned into SVG, but we're still using  links (though might be better to eventually embed svgs).

Should I create an issue to undo this? The API would really need to return proper images to not be breaking. Not every system can render SVG

diosmosis:
@tsteur Made a couple fixes. Re it looking bloated, does this look better?

image

Upped the margin-bottom from 7px to 10px. Not sure if it really makes a difference, going up to 14px looked more or less the same to me.

> Should I create an issue to undo this? The API would really need to return proper images to not be breaking. Not every system can render SVG

Probably, not sure if the old images are still in the repo? Would want to cc @sgiehl , maybe we can add an iconSvg property to the API response.

tsteur:
`IconSVG` could work in the API response 👍 

Can we maybe remove the last vertical line under an icon for the last action within a pageview?

image

and for some reason sometimes it looks like there's not enough space between the last action of a pageview, and the next page view
image

while it seems better for some others

image

I wonder if we should even remove the vertical line under an icon when the action is on one line only as it is rather irritating maybe. Or maybe we could have a longer minheight for it? 

image


I reckon those changes could make a good difference. 

The "show more/less" works now 👍

diosmosis:
@tsteur Made some tweaks, can you check whether it looks better to you?

tsteur:
Yes looks good 👍

diosmosis:
Created iconSvg issue here: https://github.com/matomo-org/matomo/issues/14198

diosmosis:
@tsteur / @mattab updated to show refreshes and allow expanding, maybe you guys want to take a look before merging?

tsteur:
I suppose we could merge it now and make further changes if needed after using it for a while @mattab  @diosmosis ?

mattab:
> I suppose we could merge it now and make further changes if needed after using it for a while

 :+1:

PR #14069, opened 2019-02-04
Search for visitor IDs over whole table not just last 30 mins

diosmosis:
In the tracker when searching by visitor ID, search through entire log_visit table instead of just in the last 30 mins.

Should be fast thanks to already existing index (h/t @tsteur)

CC @mattab can you take a look at this re any possible privacy implications?

### DIFF
--- a/core/Tracker/Model.php
+++ b/core/Tracker/Model.php
@@ -357,23 +357,26 @@ class Model
         // 		this page view to the wrong visitor, but this is better than creating artificial visits.
         // 2) there is a visitor ID and we trust it (config setting trust_visitors_cookies, OR it was set using &cid= in tracking API),
         //      and in these cases, we force to look up this visitor id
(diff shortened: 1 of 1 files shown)


mattab:
Should work and in theory not introduce a performance regression :+1:

sgiehl:
> Should work and in theory not introduce a performance regression

@tsteur @mattab should we maybe test that on some bigger cloud instances before merging? Otherwise tests may need some updates, but would be good to merge then.

diosmosis:
FYI, I've been fixing tests for this in https://github.com/matomo-org/matomo/pull/13935 which contains this change + some fixes, this PR probably shouldn't be merged. I think it's good to have it open for any discussion though. I'll close it when I merge the other one.

diosmosis:
Included in #13935

PR #14076, opened 2019-02-06
Set segments for folder branches in page urls/page titles report

### LINKED ISSUE #9229: Add segmented visitor log to page url and page title folders
Once #8076 is resolved and once there is a "Starts with" segment operator we could provide the possibility to use the segmented visitorlog #4633 on folders in the Page URL and Page Title report I think.

Eg when structure is 
- blog
  - 2014
    - 01
      - index.html

I think we could segment `pageUrl=^example.com/blog/2014` to get all users that were on that folder

diosmosis:
Only way to get full path for subfolders is to set new metadata in Actions report. Which means this won't work for old data. I did add some code paths however that add the segment for top level folders. 

Fixes #9229


plugins/Actions/DataTable/Filter/Actions.php
```diff
@@ -35,6 +35,7 @@ public function __construct($table, $isPageTitleType)
     public function filter($table)
     {
         $table->filter(function (DataTable $dataTable) {
+            $urlPrefix = $dataTable->getMetadata('site')->getMainUrl();
```
sgiehl: When viewing an existing report **flattened** this throws a `Call to a member function getMainUrl() on boolean`

sgiehl (changes requested):

diosmosis:
Updated the PR.

sgiehl (approved):
Seems to work as expected now. Good to merge if tests pass

PR #14077, opened 2019-02-06
Hide View Detailed Visitor Log link when realtime visits widget is embedded.

### LINKED ISSUE #10079: When embedding full Piwik in another application via iframe, "View detailed visitor log" link does not work
## Reproduce
- Embed Piwik Dashboard via iframe
- In widget "Real time visitors" click on the footer link "View detailed visitor log" 
- Got: the page becomes white (see screenshot)
- Expected instead: the visitor log should load and replace the dashboard.

Screenshot

empty

Reproduce on demo: https://demo.piwik.org/index.php?module=Widgetize&action=iframe&moduleToWidgetize=Dashboard&actionToWidgetize=index&idSite=1&period=week&date=yesterday

Reported in forums: https://forum.piwik.org/t/piwik-and-iframe/18726/7

diosmosis:
Fixes #10079

### DIFF
--- a/plugins/Dashboard/javascripts/widgetMenu.js
+++ b/plugins/Dashboard/javascripts/widgetMenu.js
@@ -145,7 +145,9 @@ widgetsHelper.getWidgetNameFromUniqueId = function (uniqueId, callback) {
  */
 widgetsHelper.loadWidgetAjax = function (widgetUniqueId, widgetParameters, onWidgetLoadedCallback, onWidgetErrorCallback) {
     var disableLink = broadcast.getValueFromUrl('disableLink');
-    if (disableLink.length) {
+    if (disableLink.length
+        || $('body#standalone').length
+    ) {
         widgetParameters['disableLink'] = disableLink;
     }
 
--- a/plugins/Dashboard/javascripts/widgetMenu.js
+++ b/plugins/Dashboard/javascripts/widgetMenu.js
@@ -145,11 +145,7 @@ widgetsHelper.getWidgetNameFromUniqueId = function (uniqueId, callback) {
  */
 widgetsHelper.loadWidgetAjax = function (widgetUniqueId, widgetParameters, onWidgetLoadedCallback, onWidgetErrorCallback) {
     var disableLink = broadcast.getValueFromUrl('disableLink');
-    if (disableLink.length
-        || $('body#standalone').length
-    ) {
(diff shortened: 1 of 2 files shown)


sgiehl (changes requested):
When getting the embedded url for the realtime widget it includes a `&disableLink=1`, which removes the Link from HTML output. Wouldn't it be possible to always set `disableLink` to true when the widget is embedded instead of having an additional CSS solution

diosmosis:
I thought about doing that first, but I couldn't find a way to hide it that way since I couldn't find a way to tell if it was being loaded in a widgetized dashboard. The widgetized dashboard is loaded in a way we can tell it's embedded, but the individual widgets are loaded as normal, ie, there's nothing in the HTTP request for the embedded visitor log to show that it's widgetized. So the only way I could find to tell if the widget was in the Widgetize module was through CSS/JS.

sgiehl:
I see. Haven't thought that through yet, but couldn't we send the `disableLink=1` to all widgets on an embedded dashboard when loading them?

diosmosis:
@sgiehl updated the PR.

plugins/Dashboard/javascripts/widgetMenu.js
```diff
@@ -145,7 +145,9 @@ widgetsHelper.getWidgetNameFromUniqueId = function (uniqueId, callback) {
  */
 widgetsHelper.loadWidgetAjax = function (widgetUniqueId, widgetParameters, onWidgetLoadedCallback, onWidgetErrorCallback) {
     var disableLink = broadcast.getValueFromUrl('disableLink');
-    if (disableLink.length) {
+    if (disableLink.length
+        || $('body#standalone').length
+    ) {
         widgetParameters['disableLink'] = disableLink;
```
sgiehl: That doesn't work correctly. If `$('body#standalone').length` is `true`, but `disableLink` isn't set, it will append `disableLink`, which is empty. Guess that need to be something like `widgetParameters['disableLink'] = disableLink || $('body#standalone').length`

sgiehl (changes requested):

diosmosis:
Updated.

sgiehl (approved):

sgiehl:
Not sure if the submodule updates are on purpose. If so: good to merge :+1:

PR #13441, opened 2018-09-17
When you are logged out, the URL gets lost when you log in

tsteur:
It won't remember any hash as the hash won't be visible in the referrer etc but it would work for most other pages.

To make it work for hash it would get likely way more complicated like we would need to persist it through JS, temporarily store it somewhere and redirect accordingly. It fixes the case mentioned in the issue.

fix https://github.com/matomo-org/matomo/issues/13328

PR #14081, opened 2019-02-07
POST to login plugin in login form

### LINKED ISSUE #14075: Remember me - not working
Well, let's try this once again (previously in #13327). Old issue now closed is still very much actual in my case.

I haven't investigated this much, but it is totally annoying and unreliable. It doesn't work, really.

Steps to "reproduce" this issue:

1. run Chrome
2. access website/matomo/
3. enter log-in credentials, check 'remember me', submit login form
4. once logged-in, close Chrome
5. repeat from step (1) above - try to access website/matomo/ again, and ... log-in form is presented once again !?

I run Matomo on VPS and cPanel shared hosting plans, and it happens on both of them, and [once-in-a-while Matomo decides to finally remember me](https://github.com/matomo-org/matomo/issues/13327#issuecomment-427692656), so it works for a while until I clear App Storage in Chrome or history + cookies.

What is happening here? I know you switched away from cookies, but sessions in:

> /var/lib/php/sessions

files are all with 0 length.

Ok, so, Piwik/Matomo [maybe] stores sessions locally in its own folder (?):

> website/matomo/tmp/sessions/

Ok, there we go. Now, here's the funny thing - if I delete those temp sessions in matomo's local tmp folder, I am still logged-in. So, they aren't used for that, huh??

Also, how / when are they being created in the first place? When I log-in, I don't see new session file stored in there. Also, php's default sessions files are not related with matomo's log-in, that makes sense already (read above).

So, it is stored in browser (where else could it possible be?)? Then, why this happens? Why Matomo keeps 'forgetting' me?

*How can we fix this once and for all?*

(please do not close this issue too early)

Thanks!

diosmosis:
Fixes #14075

### DIFF
--- a/plugins/Login/Controller.php
+++ b/plugins/Login/Controller.php
@@ -303,7 +303,7 @@ class Controller extends \Piwik\Plugin\ControllerAdmin
         $this->passwordResetter->removePasswordResetInfo($login);
 
         if (empty($urlToRedirect)) {
-            $redirect = Common::getRequestVar('form_redirect', false);
+            $redirect = Common::unsanitizeInputValue(Common::getRequestVar('form_redirect', false));
             $redirectParams = UrlHelper::getArrayFromQueryString(UrlHelper::getQueryFromUrl($redirect));
             $module = Common::getRequestVar('module', '', 'string', $redirectParams);
             // when module is login, we redirect to home...
--- a/plugins/Login/Controller.php
+++ b/plugins/Login/Controller.php
@@ -307,10 +307,12 @@ class Controller extends \Piwik\Plugin\ControllerAdmin
             $redirectParams = UrlHelper::getArrayFromQueryString(UrlHelper::getQueryFromUrl($redirect));
             $module = Common::getRequestVar('module', '', 'string', $redirectParams);
             // when module is login, we redirect to home...
-            if ($module !== 'Login' && $module !== Piwik::getLoginPluginName() && $redirect) {
+            if (!empty($module) && $module !== 'Login' && $module !== Piwik::getLoginPluginName() && $redirect) {
                 $host = Url::getHostFromUrl($redirect);
                 // we only redirect to a trusted host
-                if ($host && Url::isValidHost($host)) {
+                if ($host == Url::getHost()
+                    && Url::isValidHost($host)
+                ) {
                     $urlToRedirect = $redirect;
                 }
             }
--- a/plugins/Login/Controller.php
+++ b/plugins/Login/Controller.php
@@ -313,8 +313,7 @@ class Controller extends \Piwik\Plugin\ControllerAdmin
                 $currentHost = explode(':', $currentHost, 2)[0];
 
                 // we only redirect to a trusted host
-                if ($host == $currentHost
-                    && Url::isValidHost($host)
+                if (!empty($host) && !empty($currentHost) && $host == $currentHost && Url::isValidHost($host)
                 ) {
                     $urlToRedirect = $redirect;
                 }
--- a/plugins/Login/Controller.php
+++ b/plugins/Login/Controller.php
@@ -21,6 +21,7 @@ use Piwik\Plugins\UsersManager\Model AS UsersModel;
 use Piwik\QuickForm2;
 use Piwik\Session;
 use Piwik\Url;
+use Piwik\UrlHelper;
 use Piwik\View;
 
 /**
@@ -126,7 +127,6 @@ class Controller extends \Piwik\Plugin\ControllerAdmin
     function login($messageNoAccess = null, $infoMessage = false)
     {
         $form = new FormLogin();
-        $form->removeAttribute('action'); // remove action attribute, otherwise hash part will be lost
         if ($form->validate()) {
             $nonce = $form->getSubmitValue('form_nonce');
             if (Nonce::verifyNonce('Login.login', $nonce)) {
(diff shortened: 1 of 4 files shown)


tsteur:
@diosmosis haven't tested but when doing the redirect, and then logging in, does the feature still work to load the same page as you loaded initially? Say you are logged out and open the users manager, then log in, is the users manager loaded after log in?

diosmosis:
@tsteur It doesn't, but this doesn't work on 3.x-dev for me either. (To be clear that's: login, load users manager, logout, log back in, see dashboard not users manager.)

tsteur:
What happens when you do `load users manager, login`? I think we added some feature to do that... but maybe it isn't merged yet or so

diosmosis:
Would there need to be anonymous access to a site to do that? I'm not sure how I'd get past the login page w/o logging in.

tsteur:
@diosmosis I might be understanding it wrong... it was implemented here: https://github.com/matomo-org/matomo/pull/13441

When you open eg `https://example.com/index.php?module=UsersManager&action=userSettings&idSite=1&period=day&date=yesterday` , then it should show the login form on that page... once you log in, it should show the same page again. Not sure if that URL would now redirect to login page?

fdellwing:
It might be naive, but why is `!$this->shouldHandleRememberMe()` even called? Why not just take a look at `$_POST['form_rememberme']` and if it is there set expires? Wouldn't that instantaniosly solve #14075 and not touch any redirect logic that is present in Matomo?

diosmosis:
@tsteur I see, it should remember the URL, was testing incorrectly.

@fdellwing it's only needed to perform that logic in a single controller action, and w/o that check, the logic could be triggered by a user by simply adding a query parameter. To protect against future bugs/exploits that could arise w/ new features written in completely different places by people who are not aware or have forgotten about this specific code path, we restrict the code's execution to exactly where it needs to be used.

diosmosis:
@tsteur interestingly enough, that feature still works, but remember me doesn't...

diosmosis:
@tsteur updated the code and tweaked the URL retaining feature. Now it posts to `module=Login` and gets the referrer URL from a POST parameter. Can you check if this works for you?

plugins/Login/FormLogin.php
```diff
@@ -19,6 +19,7 @@ class FormLogin extends QuickForm2
 {
     function __construct($id = 'login_form', $method = 'post', $attributes = null, $trackSubmit = false)
     {
+        $attributes = array_merge($attributes ?: [], [ 'action' => '?module=Login' ]);
```
tsteur: Would this need to be `getLoginPluginName()` in case a different login plugin is active? instead of hard coded `Login`

fdellwing:
Patch seems to work for me.

plugins/Login/FormLogin.php
diosmosis: 👍 makes sense

tsteur:
@diosmosis I got redirected to this:

image

What I did was open `https://apache.matomo/index.php?module=UsersManager&action=userSettings&idSite=1&period=day&date=yesterday` and then logged in.

It was supposed to render userSettings but it is rendering users manager because of the wrong URL `&amp;`

diosmosis:
@tsteur should be fixed.

plugins/Login/Controller.php
```diff
@@ -303,14 +303,15 @@ protected function authenticateAndRedirect($login, $password, $urlToRedirect = f
...
         if (empty($urlToRedirect)) {
-            $referrer = Url::getReferrer();
-            $module = Common::getRequestVar('module', '', 'string');
+            $redirect = Common::unsanitizeInputValue(Common::getRequestVar('form_redirect', false));
+            $redirectParams = UrlHelper::getArrayFromQueryString(UrlHelper::getQueryFromUrl($redirect));
+            $module = Common::getRequestVar('module', '', 'string', $redirectParams);
             // when module is login, we redirect to home...
-            if ($module !== 'Login' && $module !== Piwik::getLoginPluginName() && $referrer) {
-                $host = Url::getHostFromUrl($referrer);
+            if ($module !== 'Login' && $module !== Piwik::getLoginPluginName() && $redirect) {
```
tsteur: Be good to check that `$module` is not empty...

tsteur: may actually not be needed though because we do validate a host is set below just seeing

plugins/Login/Controller.php
```diff
@@ -303,14 +303,15 @@ protected function authenticateAndRedirect($login, $password, $urlToRedirect = f
...
+                $host = Url::getHostFromUrl($redirect);
                 // we only redirect to a trusted host
                 if ($host && Url::isValidHost($host)) {
```
tsteur: I just realised I should have maybe adjusted the description for the config setting

```
; Whether to enable trusted host checking. This can be disabled if you're running Matomo
; on several URLs and do not wish to constantly edit the trusted host list.
enable_trusted_host_check = 1
```

When disabled, you can redirect any user to any url... and now that it's even possible by GET parameter and not referrer it may be even more important. Basically, when the setting is disabled, you can link to `https://demo.matomo.org?...&redirect=https://demo.maomo.org` for example and after the succeeded login, the user would be redirected to that page. This could be some security issue if you ask the user on that similar domain to enter the password again. Wonder how we could mention it in the config? In general we should not recommend users to disable the trusted host check.

diosmosis: What if we also check that the host in the URL is the same as the Host header? Then the only way to force a redirect to a different domain, would be to send a curl request w/ the changed Host header.

tsteur: sounds good

plugins/Login/Controller.php
```diff
@@ -307,10 +307,12 @@ protected function authenticateAndRedirect($login, $password, $urlToRedirect = f
             $redirectParams = UrlHelper::getArrayFromQueryString(UrlHelper::getQueryFromUrl($redirect));
             $module = Common::getRequestVar('module', '', 'string', $redirectParams);
             // when module is login, we redirect to home...
-            if ($module !== 'Login' && $module !== Piwik::getLoginPluginName() && $redirect) {
+            if (!empty($module) && $module !== 'Login' && $module !== Piwik::getLoginPluginName() && $redirect) {
                 $host = Url::getHostFromUrl($redirect);
                 // we only redirect to a trusted host
-                if ($host && Url::isValidHost($host)) {
+                if ($host == Url::getHost()
```
tsteur: we probably still need to check `!empty($host)` as `getHost()` may return an empty host when the config setting is disabled for example. It would be able to trick matomo into redirecting maybe when providing some weird URL although I don't really think it's possible and it should be all fine without the !empty but better be safe maybe.

tsteur (approved):

PR #14083, opened 2019-02-10
Display all product categories if present in the visitor log

### LINKED ISSUE #14072: In the visitor log/ecommerce log, show all product categories not just the first
Currently in the visitor log, only the first product category is displayed (see https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/Ecommerce/VisitorDetails.php#L179-L180). But users can supply up to 5 different categories for a single product. We should display all of them.

diosmosis:
Fixes #14072


core/DataTable/Renderer/Xml.php
```diff
@@ -370,7 +370,11 @@ protected function renderDataTable($array, $prefixLine = "")
                 foreach ($row as $name => $value) {
                     // handle the recursive dataTable case by XML outputting the recursive table
                     if (is_array($value)) {
-                        $value = "\n" . $this->renderDataTable($value, $prefixLine . "\t\t");
+                        if (is_array(reset($value))) {
```
tsteur: can you add a (system) test for this maybe?

plugins/Ecommerce/VisitorDetails.php
```diff
@@ -165,19 +167,30 @@ protected function queryEcommerceConversionsForVisits($idVisits)
      */
     protected function queryEcommerceItemsForOrder($idVisit, $idOrder)
     {
+        $categorySelects = [];
+        $categoryJoins = [];
+        for ($i = 0; $i < self::CATEGORY_COUNT; ++$i) {
+            $suffix = $i === 0 ? '' : $i;
+            $column = $i === 0 ? 'idaction_category' : ('idaction_category' . ($i + 1));
+            $categorySelects[] = 'log_action_category' . $suffix . '.name as itemCategory' . $suffix;
+            $categoryJoins[] = 'LEFT JOIN ' . Common::prefixTable('log_action') . " AS log_action_category$suffix
```
tsteur: I reckon performance won't be a big problem because of adding so many more joins but may be still good to do (maybe we have many categories in the demo on staging?).

tsteur: If easily doable...

diosmosis: You mean performance test on staging?

tsteur: Yeah I wonder if we should deploy the change on the server and check if it still loads as fast when eg loading 500 or 1000 ecommerce log entries compared to before? not sure we have many entries with categories though

diosmosis: Could always add some more categories to the test data and load it on staging if not. Would take a couple days though.

core/DataTable/Renderer/Xml.php
diosmosis: Technically since it's used in the `<categories>` array, it's already tested in this PR. If possible I can add a unit test as well.

plugins/Ecommerce/VisitorDetails.php
tsteur: Nah not needed to add more categories. Not worth it I think

sgiehl:
UI and System test files need an update. 
@diosmosis do you think it may sense to extend the UI tests so we have multiple categories displayed somewhere? Otherwise should be good to merge I think

diosmosis:
I added an extra category to a system test and regenerated OmniFixture (not sure there was an existing fixture w/ multiple categories), hopefully it will show up in the ecommerce log UI test result.

tsteur:
Looks good to merge once tests pass. There are a few failing system tests and to fix the UI tests might need to be merge 3.x-dev into it.

Couldn't see multiple categories in a UI test but I think it's not needed.

PR #14084, opened 2019-02-10
Remove old unused files

Findus23:
while looking into #14082, I just found this old branch.

Can everyone please check if those files are really not needed?


tsteur:
I don't think they are needed

diosmosis:
👍

PR #14091, opened 2019-02-12
Some archiving optimizations

### LINKED ISSUE #13387: Archiver should skip creating a new "temporary" archive when there was no new visit/traffic since the last "temporary archive"
Challenge: make archiving faster when there are hundreds of websites

Solution:

* We only want to trigger the archiving for sites, that had at least one tracking request since the last archiving run. This will only help if one has many sites with 0 visits. It does not help for many sites with low traffic.
* Applies to any archive for today / this week / this month / this year, or a custom date range that includes today.

This should probably be implemented in two places. 

* in core archiver, while preparing to create the archive, whenever there is no visit since the last archive timestamp, we should realise the last processed temporary archive is still valid, and skip creating a new one
* in core:archive we would also be smart and automatically skip websites, and all segments, whenever there was no new visit. (we would need to skip websites and all segments only when we are sure that they were all successfully processed.)

from https://github.com/matomo-org/matomo/issues/5922#issuecomment-90352741

diosmosis:
WIP

Fixes #13387


tsteur:
@diosmosis in `getLatestArchiveTimestampForToday` we check for the `VisitsSummary` name condition and in the other method where we get the `ts_archived`. Does this return the time of the first blob/numeric value written within an archive? We we would need to make sure to check for the first created archive because it can take an hour to finish an archive or longer... In theory even before the first archive is written it may take a while to fetch the data for the first archive and write this. I suppose we can't log the time in the option table when the archiving was last launched (before executing any query) for a specific archive and compare against this?

Also I think the PR will need to a similar logic in `CronArchive` for segments? Which may not have been launched for quite a while?

diosmosis:
@tsteur made a comment in the original issue about the segment related change (the code already exists in CronArchive): https://github.com/matomo-org/matomo/issues/13387#issuecomment-468560620

> Does this return the time of the first blob/numeric value written within an archive? We we would need to make sure to check for the first created archive because it can take an hour to finish an archive or longer... In theory even before the first archive is written it may take a while to fetch the data for the first archive and write this. I suppose we can't log the time in the option table when the archiving was last launched (before executing any query) for a specific archive and compare against this?

Wouldn't the first created archive be the first archive of the day? That would defeat the purpose of making this change? This SQL was supposed to find the latest ts_archived value for an archive that can be used to query visits. Then the idea is, from that point on until 'now', if there are no visits found, then we don't need to initiate archiving and can just use the existing archive. The archiving code is extremely dense, though, so I am not entirely sure if this code has that effect... if you have any examples where it breaks, would like to see them.

tsteur:
It wouldn't need to use the first archive of the day I think (not 100% about everything though). Just the launch date of the last launch of the archiving for a given `idsite,date1,date2,period,segment`.

image

The thinking is that archiving may take several minutes or hours. If it uses the latest date, which may be `23:26:06` and there were visits between `2019-03-01 23:24:41` and `23:26:06` it may not be launched again as it thinks there was no new visit since `23:26:06`?

I don't think we have that check in the code already as https://github.com/matomo-org/matomo/blob/3.x-dev/core/CronArchive.php#L1261-L1292 is around midnight.

diosmosis:
> If it uses the latest date, which may be 23:26:06 and there were visits between 2019-03-01 23:24:41 and 23:26:06 it may not be launched again as it thinks there was no new visit since 23:26:06?

If the ts_archived time is 23:26:06, then wouldn't include visits before that time (so, any visits between 23:24:41  => 23:26:06)? Ie, ts_archived is for when the archiving query finishes (AFAIK), so when we aggregate for nb_visits and create that archive for today, the nb_visits would include visits from start of day => ts_archived, right?

Of course, this doesn't apply to archiving dates in the past, just when running archiving for today.

tsteur:
> If the ts_archived time is 23:26:06, then wouldn't include visits before that time

only for some archives. See above in the screenshot they are all for the same archive. But some may include those visits, some not... as it takes several minutes to hours to finish it.

diosmosis:
> only for some archives. See above in the screenshot they are all for the same archive. But some may include those visits, some not... as it takes several minutes to hours to finish it.

to finish all archiving or just the one archiving query? in this case we're only looking at archives for VisitsSummary / no segment. though I guess if it gets the all plugins that might take more time... actually, i'm not sure how ts_archived is even set... will take a look quickly.

diosmosis:
looks like the ts_archived value is set when inserting individual idarchives, so the `done` archives will be set to when archiving finishes, but the individual metrics will have the insert record for when they are actually inserted. So maybe we can just look for the ts_archived value of the nb_visits metric? We'd get the latest done idarchives, and check for the latest ts_archived nb_visits for those idarchives. Would that work?

tsteur:
it should work in most cases. There may be edge cases where it takes also a few ms or seconds or minutes to query and insert the `nb_visits` metric. Random thought: before querying any data, should we insert maybe also a `start` flag similar to the `done` flag? that may be consistent and precise

diosmosis:
So we add a `start` flag, then we look for the latest done/done-temp archive for the site/period/etc. and get the idarchive. Then we get the ts_archived for the `start` flag in that archive, and check if there were any visits between `start.ts_archived -> now`. Does that sound right to you?

tsteur:
That sounds right 👍

diosmosis:
@tsteur Updated

core/DataAccess/Model.php
```diff
@@ -333,6 +333,18 @@ public function getSitesWithInvalidatedArchive($numericTable)
...
 
+    /**
+     * Returns whether there is at least one visit that ended after the given timestamp.
+     *
+     * @param $tsArchived
+     * @return bool
+     */
+    public function hasEncounteredVisitsSinceLastArchive($tsArchived)
+    {
+        $rows = Db::fetchAll("SELECT idvisit FROM " . Common::prefixTable('log_visit') . " WHERE visit_last_action_time >= ? LIMIT 1", [$tsArchived]);
```
tsteur: I'm a bit worried this will do a full table scan. Can we include the idSite? This will be likely needed anyway I suppose to check if there were visits for that specific site

core/DataAccess/ArchiveSelector.php
```diff
@@ -67,15 +68,28 @@ public static function getArchiveIdAndVisits(ArchiveProcessor\Parameters $params
...
 
-        $results = self::getModel()->getArchiveIdAndVisits($numericTable, $idSite, $period, $dateStartIso, $dateEndIso, $minDatetimeIsoArchiveProcessedUTC, $doneFlags, $doneFlagValues);
-
+        $results = self::getModel()->getArchiveIdAndVisits($numericTable, $idSite, $period, $dateStartIso, $dateEndIso, null, $doneFlags, $doneFlagValues);
         if (empty($results)) {
             return false;
         }
 
-        $idArchive = self::getMostRecentIdArchiveFromResults($segment, $requestedPlugin, $results);
+        list($idArchive, $tsArchived) = self::getMostRecentIdArchiveFromResults($segment, $requestedPlugin, $results);
```
tsteur: Wouldn't you need to make sure to use the time of the `start` flag here?

tsteur: I mean just to be safe. Cause even archiving one plugin can take 1 hour.

tsteur:
Left some comments @diosmosis 

have you looked into triggering the archiving of segments only when needed in the `CronArchive`?

Also wondering have you tested what happens when receiving a tracking request for an older date and this older date needs to be rearchived maybe because the archive was invalidated or so?

So far looked only at the code so haven't tested it yet. Might be quite tricky.

diosmosis:
> have you looked into triggering the archiving of segments only when needed in the CronArchive?

That's what this commit should achieve: https://github.com/matomo-org/matomo/pull/14091/commits/23b5a1871f3c332b4ebc46235814a914abc3e7af

Don't think I've tested a lot, will have to do more...

core/DataAccess/ArchiveSelector.php
diosmosis: Think it could be easy to do, will change this.

tsteur:
> That's what this commit should achieve: 23b5a18

I mean we may have archived some segments at a later time than the regular archive and we wouldn't need to issue a request for a specific segment (but other segments) or maybe the segment was last executed a longer time ago and it needs to issue the segment request but not the regular archiving. 

It sounds all pretty tricky and not sure how easily things are doable.

diosmosis:
> I mean we may have archived some segments at a later time than the regular archive and we wouldn't need to issue a request for a specific segment (but other segments) or maybe the segment was last executed a longer time ago and it needs to issue the segment request but not the regular archiving.

This might require a larger refactoring of the cron archive process... maybe we can turn it into something a bit easier to manipulate in this way? I don't expect a change like this could be done by 3.9...

tsteur:
It won't be doable in 3.9

Maybe we could split it in two parts though: 
1) The core archiver / selector
2) Cron archiver not sending requests in the first place

This might make it also easier to review and I think most of the logic or all logic for `1)` is maybe already there?

diosmosis:
@tsteur they should both be there? The second one should be achieved by https://github.com/matomo-org/matomo/commit/23b5a1871f3c332b4ebc46235814a914abc3e7af (though there's some extra code that doesn't need to be there). Is there something off w/ it?

tsteur:
Sorry I might be confused. I see you added some code for `hadWebsiteTrafficSinceMidnightInTimezone` but this does only take into consideration the overall traffic. 

During one archive we may trigger heaps of requests for different periods and segments like
* Day
  * Regular request
  * Segment 1 
  * Segment 2
* Week
  * Regular request
  * Segment 1 
  * Segment 2
* Month
  * Regular request
  * Segment 1 
  * Segment 2

And if I see this correctly then we need to check for each request when that information was last archived before requesting the archive trigger URL. Because the `regular monthly archive` might have been archived way longer back than the daily archive and there could have been visits since the last archive time. The same for segments etc.

I would have basically kind of expected to see some new code before each `$this->request($url);` where we check when was this particular request last archived (based on `start flag`), and whether a visit was tracked since then.

Eg we might consider a `time_before_month_archive_considered_outdated` of `24 hours` to archive it only once a day. Meaning the last archive was maybe 23 hours ago, not just 1 hour for the regular daily archive.

diosmosis:
I see, so do the check I added before each request, for the specific archive in question, correct?

tsteur:
Yep

core/DataAccess/ArchiveSelector.php
```diff
@@ -338,6 +354,28 @@ private static function moveChunkRowToRows(&$rows, $row, Chunk $chunk, $loadAllS
         }
     }
 
+    public static function getLatestArchiveStartTimestampForToday($idSite)
+    {
+        $today = Date::today();
```
tsteur: does today need to be in timezone of the site?

core/DataAccess/ArchiveSelector.php
```diff
@@ -67,15 +68,28 @@ public static function getArchiveIdAndVisits(ArchiveProcessor\Parameters $params
         $doneFlags      = Rules::getDoneFlags($plugins, $segment);
         $doneFlagValues = Rules::getSelectableDoneFlagValues();
 
-        $results = self::getModel()->getArchiveIdAndVisits($numericTable, $idSite, $period, $dateStartIso, $dateEndIso, $minDatetimeIsoArchiveProcessedUTC, $doneFlags, $doneFlagValues);
-
+        $results = self::getModel()->getArchiveIdAndVisits($numericTable, $idSite, $period, $dateStartIso, $dateEndIso, null, $doneFlags, $doneFlagValues);
```
tsteur: not quite sure I understand why `minDatetimeIsoArchiveProcessedUTC ` is passed as `null` here? Wouldn't it make things faster to return immediately if no archive found that is newer than `minDatetime`? And in line `82` it might return even if the archive is older?

tsteur:
@diosmosis left a few comments. I reckon the main thing be to change as discussed `CronArchive` to check for new visits before each `archiving request` instead of just in the beginning. Eg because a segment has been processed more recently or a longer time ago which means while we can maybe skip a daily archive, we maybe cannot skip a segment archive.

core/DataAccess/ArchiveSelector.php
diosmosis: This might have been to fix a test, but I can't remember anymore. I'll see if the build passes.

diosmosis:
@tsteur made the requested changes & merged w/ 3.x-dev. Not 100% sure about whether the cronarchive changes will have side effects or not

core/DataAccess/ArchiveSelector.php
```diff
@@ -338,6 +359,30 @@ private static function moveChunkRowToRows(&$rows, $row, Chunk $chunk, $loadAllS
...
+                  FROM $table
+                 WHERE idsite = ? AND period = ? AND date1 = ? AND date2 = ? AND $nameCondition";
+        $bind = [$idSite, $periodObj->getId(), $periodObj->getDateStart()->getDatetime(), $periodObj->getDateEnd()->getDatetime()];
+
+        $idArchive = Db::fetchOne($sql, $bind);
+
+        $sql = "SELECT ts_archived
+                  FROM $table
+                 WHERE idarchive = ? AND name = ?";
+        $bind = [$idArchive, ArchiveWriter::ARCHIVE_START_RECORD_NAME];
```
tsteur: Initially I was wondering if we only should look at archives where the archiving finished  in case the previous archiving had an error and interrupted (same idArchive should also have the done flag with status OK or temporary). 

Then realised we can likely not do this because the current archiving might be in progress and that's the reason why the done flag is missing. => This should not be a problem though because the archiving would then be simply skipped as the logic should notice for that period/date there is already an archiver running right now.

So what if the last archive stopped in the middle because of an error, then we would maybe not start that archive again under circumstances? Say start flag is `00:15am`, then the archiving broke in the middle, ... it would maybe not start the archiving again? Maybe, to play it safe, we might need to always check to only look at `ts_archived` for archives that have the start and the done flag.

Not sure I explain it well.

core/CronArchive.php
```diff
@@ -863,6 +863,13 @@ protected function processArchiveDays($idSite, $lastTimestampWebsiteProcessedDay
             return false;
         }
 
+        if (!$this->hasThereBeenVisitsSinceLastArchiveTimestamp($idSite, 'day', $date)) {
```
tsteur: Do I see this correctly that if a user has no new visit recorded since last archive, but recorded some visits in the past, we wouldn't trigger the archiving until there's actually a new visitor with a more recent date? Do you know what I mean? Probably can't work around this I suppose.

tsteur (changes requested):
Did some basic testing and it worked. Although not sure re some edge cases, left some comments to discuss.

core/ArchiveProcessor/PluginsArchiver.php
```diff
@@ -62,6 +62,7 @@ public function __construct(Parameters $params, $isTemporaryArchive, ArchiveWrit
         $this->isTemporaryArchive = $isTemporaryArchive;
         $this->archiveWriter = $archiveWriter ?: new ArchiveWriter($this->params, $this->isTemporaryArchive);
         $this->archiveWriter->initNewArchive();
+        $this->archiveWriter->insertRecord(ArchiveWriter::ARCHIVE_START_RECORD_NAME, 1);
```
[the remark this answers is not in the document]
diosmosis: it will have the idArchive for the specific done flag, so as long as we find the idarchive first, it wouldn't need the suffix. Which makes the code a bit simpler since we don't have to compute it when querying.

core/DataAccess/ArchiveSelector.php
diosmosis: My understanding was that's what we do, at least that's what `$nameCondition` is for. Ie, it should be `(name IN ... AND value IN (DONE_OK, DONE_OK_TEMPORARY)`. So we're only looking at done archives.

Incidentally I think this is also why we shouldn't append the suffix to the `start` record, since we'd have to check the done flag value anyway.

core/DataAccess/ArchiveSelector.php
```diff
@@ -338,6 +359,30 @@ private static function moveChunkRowToRows(&$rows, $row, Chunk $chunk, $loadAllS
...
+    public static function getLatestArchiveStartTimestampForToday($idSite, $period, $date, Segment $segment)
+    {
+        $today = Date::factoryInTimezone('today', Site::getTimezoneFor($idSite));
+        $table = ArchiveTableCreator::getNumericTable($today);
+
+        $periodObj = Period\Factory::build($period, $date);
+
+        $nameCondition = self::getNameCondition(['VisitsSummary'], $segment);
+
+        $sql = "SELECT MAX(idarchive)
```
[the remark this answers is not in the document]
diosmosis: (see response in below comment)

core/DataAccess/ArchiveSelector.php
diosmosis: I know we've all said it before, but the archiving process confuses the heck out of me.

core/CronArchive.php
diosmosis: I see what you mean, I think we could take this into account by using idvisit somehow... Maybe find the minimum idvisit that is >= ts_archived, then check if there is an idvisit for that site > than the minimum idvisit? Does that sound right?

tsteur: something like this could work. Might be complicated though.  Wonder if https://github.com/matomo-org/matomo/pull/14091#issuecomment-490719464 would maybe make a lot of things easier. In theory the archiving logic should all be simple. After all it's not that complicated. We generate an archive and assume it's completed. Once a new tracking request comes in, we make sure to rearchive it. If no tracking request comes in again, we don't need to archive it again.

diosmosis: I'll see if I can add it w/ a test.

Your other comment makes sense to me, I'm actually not sure currently what temporary archives are used for currently. Removing and making the process simpler might be difficult given how complicated the system is currently, might be good to wait until matomo 4?

core/DataAccess/ArchiveSelector.php
tsteur: > I know we've all said it before, but the archiving process confuses the heck out of me.

me too :)

core/CronArchive.php
tsteur: I reckon it might not be all that complicated? And maybe a lot easier than this logic? Not sure though. @mattab any thoughts on this idea maybe?

diosmosis: I'm just not too sure if it will break anything, and we also have to deal w/ the fact that existing archives may have DONE_TEMPORARY. Seems like there could be room for bugs that would be hard to track down... Maybe we can schedule it for 3.11 at least, so when it's deployed on cloud you'll be around.

core/DataAccess/ArchiveSelector.php
tsteur: > My understanding was that's what we do, at least that's what $nameCondition is for. Ie, it should be (name IN ... AND value IN (DONE_OK, DONE_OK_TEMPORARY). So we're only looking at done archives.

Makes sense . Just need to double check that we also look at that in the `CronArchive`

diosmosis: CronArchive uses `ArchiveSelector::getLatestArchiveStartTimestampForToday()` which uses the name condition, so I think we're good here.

diosmosis:
I can create an issue to get rid of the done_temporary archives, otherwise I can move this to 3.11.

diosmosis:
I'll try to get rid of temporary tables in a new PR and get back to this one.

tsteur:
@diosmosis what's the status here?

diosmosis:
@tsteur closing this, we can fix this in a better way since the invalidated archive change was merged

PR #14092, opened 2019-02-12
In segmented visitor log, allow opening segment editor in new tab w/ segment used.

### LINKED ISSUE #7215: In the Segmented Visitor Log, let me open in a new tab the Piwik UI with the custom segment applied
The goal of this issue is to improve the great Segmented Visitor Log feature (from #4633), and let users apply the Segment to all Piwik reports (rather than just see the segment applied to the Visitor Log popover).

For example, maybe we could add an icon or a textual link "Apply this segment to the report [new tab]". When clicked, a new tab will open and this custom segment will be applied to all reports. 

Thoughts?

diosmosis:
Todo:

* ~add UI test~ tried to add one but realized since it opens in a new tab, it would be rather difficult w/ current UI tests...

Fixes #7215


plugins/CoreHome/templates/_dataTableActions.twig
```diff
@@ -103,6 +103,20 @@
             </a>
         {% endif %}
 
+        {% if properties.datatable_actions|default is not empty %}
+        {% for action in properties.datatable_actions %}
+            <a class='dataTableAction {{ action.id|e('html_attr') }}'
+               href='javascript:;' title="{{ action.title|e('html_attr') }}"
+            >
+                {% if action.icon starts with 'icon-' %}
+                    <span title="" class="{{ action.icon }}"></span>
```
sgiehl: any reason for not adding the title here? That prevents the tooltip from being displayed when hovering the icon directly.

sgiehl (changes requested):
Works as expected. Left a minor comment

plugins/CoreHome/templates/_dataTableActions.twig
diosmosis: No, probably a typo

diosmosis:
Updated.

sgiehl:
@diosmosis  actually guess it would make more sense to add `title="{{ action.title }}"` so the tooltip is visible correctly

diosmosis:
@sgiehl It looks ok to me? Or do you mean visible in the code?

sgiehl:
Hm. Works on Chrome on Linux as well. Need to check again on Windows, as it didn't work there last time...

sgiehl:
Tested on windows again. Seems to work now in Chrome, FF and IE. Maybe it has been some browser caching issue.

sgiehl (approved):

PR #14095, opened 2019-02-13
Display help icons for categories/subcategories

### LINKED ISSUE #13716: Show a message in each most valuable pages in Matomo to explain why it's important
To remind users why particular features/pages in Matomo are important, it would be valuable to show users a notification message, which reminds them of:
* What is this 
* Why is it important
* Where to learn more
* Could even show a video showcasing/explaining the feature

For example it is needed to explain to users why "Goals" are so important. So when people click on a Goals page, the first time, they would see a notification message explaining this, with links and video.

It will be also valuable for people browsing our [Matomo demo](https://demo.matomo.org/) so they can learn more about value of the product while browsing the demo.

Notes:
* A user could close the notification so it does not display again.
* Ideally we would do this for all major pages, but we can add them slowly. 
* maybe this could be merged with/rethink of the **(?) Help feature** within each report

diosmosis:
Fixes #13716


diosmosis:
@mattab could you try this PR locally and see if the UX works for you: https://github.com/matomo-org/matomo/pull/14095 ?

mattab:
Like where this is going! 
Feedback:
* add a useful tooltip displayed on hover on the icon (after the standard delay) like `
* only show the icon when the mouse is within the submenu's row, or when the help notification is visible. Otherwise hide the icon to not clutter when browsing reports.
* when notification is displayed and then the icon is clicked, hide the help notification
* when a custom dashboard is loaded, show the notification at the top
* if we have other icons with a grey (not black) tone, this could work but maybe it's fine in black...

To make this feature happen we also need to draft a description text for each page, a short text that introduces the page and makes it shine. For example consistently mention what info is there, why it is important and valuable (what problems does it solve, benefits does it bring), where to learn more (link to the user guide, the video), this in maybe 1-3 lines/paragraph.  (Maybe you can try draft a few of these in the next round but not required)

tsteur:
Personally, I'm finding the icon a bit disturbing. Maybe it's better in a light grey though. Maybe, we could spin this a bit further how we could give a lot more insights with even more content than just 2-3 lines but actually train people and give them a bit more insights like on the pages Marketing is currently working on. Just a random thought.

I know it'll be hard though to find possibly a better solution that isn't always shown. Maybe the icon could also be only shown when hovering that main menu item. 

Be maybe also good to check if we find maybe some solutions how other tools did it? (if there are some)

diosmosis:
@tsteur I saw this sort of intro tutorial a while ago:

image (2)

image (3)

I thought something like this could be used on demo.matomo.org to sell matomo & premium features (by, eg, pointing to data and reports in the demo and adding some context, plus interactively showing off certain features). Perhaps this sort of technique could also be used to show people around matomo the first time (so instead of some help text, we have a toggle-able tour?). It might not be as interesting w/o any data, however.

diosmosis:
If we need data, perhaps we can create an open source version of the demo data feature, by maybe creating a smaller demo data set and using a dump of the database. We'd have to securely download it, which might be a bit of work, and make sure we can insert it even after someone starts tracking.

diosmosis:
We could always link to the pages in the help text as well.

tsteur:
@diosmosis personally I don't like these tours but that's just me. Whenever I see one of those, I usually skip them and don't read them. I just wanna get started and use the tool etc. The problem is also that it shows you tips/tour for things I don't even know yet at this stage, and don't even want to use at that stage. And by the time I want to use that feature, I forgot what the text in the intro said about the feature etc. I believe those intro notifications only work for things when you want a user to take one or two specific actions after logging in for the first time. It's quite good for that. Like *click here to create a new user* and then *click here to create a new website* or *click here to upload xyz* (just a random example). 

I like the look of them though. They look great.

diosmosis:
@tsteur I usually skip those too. I also skim short pieces of text and skip longer text blocks. But we might not be the right users to think about, since we're pretty tech savvy, no?

The main idea I was thinking of was that some kind of set of text bubbles that pointed to features (displayed one at a time like a tour or all at once), would be better than a wall of text. Users could skim over and find what they need, and re-toggle the help when they need to get back to it. The background could be dimmed to make the bubbles more visible.

The simplest approach would be to have one or two sentence description and a link to the content page. We might also want to measure on cloud how many people actually click through.

mattab:
Quite like the direction of what this PR does (with the top notificaiton for the whole page), assuming with a lighter icon the visual side is acceptable? To be confirmed.

>  Maybe, we could spin this a bit further how we could give a lot more insights with even more content than just 2-3 lines but actually train people and give them a bit more insights like on the pages Marketing is currently working on. Just a random thought.

That would be great... maybe we could try (over the next few weeks) to create/draft this insightful content for a couple of pages and see what works (size, content) to help users leverage the full potential of the data on each page?

tsteur:
fyi @diosmosis be great to continue this work now as part of 4.1

diosmosis:
closing this pr since the commit history seems completely borked

PR #14112, opened 2019-02-18
 Add new config datatable_archiving_maximum_rows_site_search

### LINKED ISSUE #14087: New INI setting to specify the number of rows to archive in Behavior > Site Search
The [Site Search Keywords](https://matomo.org/docs/site-search/) reports are so valuable because they show what people are thinking about when browsing (and searching) for content on the website.

The reports are limited to 500 rows [by default.](https://matomo.org/faq/how-to/faq_54/)

The goal of this issue is to let users specify a higher limit (eg. 3000 or 10000) so that Matomo will keep the top 10,000 internal search engines keywords each day/week/month/year.

Currently the following INI settings are available:
```ini
; maximum number of rows for any of the Actions tables (pages, downloads, outlinks)
datatable_archiving_maximum_rows_actions = 500
; maximum number of rows for pages in categories (sub pages, when clicking on the + for a page category)
; note: should not exceed the display limit in Piwik\Actions\Controller::ACTIONS_REPORT_ROWS_DISPLAY
; because each subdirectory doesn't have paging at the bottom, so all data should be displayed if possible.
datatable_archiving_maximum_rows_subtable_actions = 100

; maximum number of rows for any of the Events tables (Categories, Actions, Names)
datatable_archiving_maximum_rows_events = 500
; maximum number of rows for sub-tables of the Events tables (eg. for the subtables Categories>Actions or Categories>Names).
datatable_archiving_maximum_rows_subtable_events = 500
```

So we have available this `datatable_archiving_maximum_rows_actions` but changing it will also affect other reports in Behavior menu (and use lots more disk space, CPU, etc.).

-> by adding a new `datatable_archiving_maximum_rows_site_search` then we could up the limit for the Site Search report only and save cost/performance.

katebutler:
Add new config datatable_archiving_maximum_rows_site_search to control number of rows archived for site search data table.

Fixes #14087


plugins/Actions/ArchivingHelper.php
```diff
@@ -319,6 +321,7 @@ public static function reloadConfig()
         self::$columnToSortByBeforeTruncation = PiwikMetrics::INDEX_NB_VISITS;
         self::$maximumRowsInDataTableLevelZero = Config::getInstance()->General['datatable_archiving_maximum_rows_actions'];
         self::$maximumRowsInSubDataTable = Config::getInstance()->General['datatable_archiving_maximum_rows_subtable_actions'];
+        self::$maximumRowsInDataTableSiteSearch = 50; //Config::getInstance()->General['datatable_archiving_maximum_rows_site_search'];
```
sgiehl: guess you forgot to uncomment this one

katebutler: Whoops, fixed

plugins/Actions/ArchivingHelper.php
```diff
@@ -228,7 +228,8 @@ public static function getRankingQueryLimit()
         $limit = $configLimit == 0 ? 0 : max(
             $configLimit,
             $configGeneral['datatable_archiving_maximum_rows_actions'],
-            $configGeneral['datatable_archiving_maximum_rows_subtable_actions']
+            $configGeneral['datatable_archiving_maximum_rows_subtable_actions'],
+            $configGeneral['datatable_archiving_maximum_rows_site_search']
```
tsteur: @katebutler I wonder if we could leave the general `rankingQueryLimit` untouched and instead only for 

`$this->archiveDayActions($rankingQueryLimit);` apply the possibly higher limit as in `$this->archiveDayActions(max($rankingQueryLimit, $config['datatable_archiving_maximum_rows_site_search'] );` in  https://github.com/matomo-org/matomo/blob/3.9.0-b1/plugins/Actions/Archiver.php#L60 ?

This way we would fetch only the needed rows for all other reports except for the one that generates site search. Eg all other "entry actions, exit actions, actions time, events, ... " would still use the correct limit directly and not fetch more than possibly needed.

diosmosis:
Can you add the new setting to BlobReportLimitingTest? Otherwise, this PR looks good.

katebutler:
@diosmosis have added to BlobReportLimitingTest

diosmosis:
👍 @katebutler merged

PR #14113, opened 2019-02-18
Add config option to use own cacert.pem file

### LINKED ISSUE #13742: Document the use of a matomo specific cacert.pem and allow to disable this behaviour
Follow up from https://forum.matomo.org/t/certificate-issues-during-update/30238/7

1. Please document this behaviour somewhere. I was not aware of this fact until @mattab pointed it out.
2. Add on option to disable this behaviour, it does not need to be accessible via UI but should be doable via config.php.

fdellwing:
Fixes #13742 

- [x] Add blog or FAQ article describing the problem and this new solution

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -690,7 +690,10 @@ piwik_professional_support_ads_enabled = 1
 num_days_before_tracking_code_reminder = 5
 
 ; The path to a custom cacert.pem file Matomo should use.
-; default is none.
+; By default Matomo uses a file extracted from the Firefox browser and provided here: https://curl.haxx.se/docs/caextract.html.
+; The file contains root CAs and is used to determine if the chain of a SSL certificate is valid and it is safe to connect.
+; Most users will not have to use a custom file here, but if you run your Matomo instance behind a proxy server/firewall that
+; breaks and reencrypts SSL connections you can set your custom file here. 
 custom_cacert_pem=
 
 [Tracker]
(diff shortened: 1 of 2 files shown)


tsteur:
Looks good in general @fdellwing 

I reckon it could be good to mention in the config comment when someone would want to adjust this and for what it is used.

fdellwing:
What do you think of? An explanation what file Matomo uses by default, where is comes from and what it does?

tsteur:
I reckon it be useful to know what the cert is used for, and what use case there is for someone to change it. Like we would someone ever want to change it? Could there be some simple example or so maybe?

fdellwing:
Please take a look now :)

tsteur (approved):
Awesome 👍 I'll merge next week and then fix ui test

tsteur:
@fdellwing if you did want to write an FAQ let me know and I'm happy to publish it. Otherwise they mind find the issue and pull request here by searching on the web

fdellwing:
@tsteur Is this ok for the FAQ? Should be in the troubleshooting section I guess.

## I get: "curl_exec: Peer’s Certificate issuer is not recognized… Hostname requested was: plugins.matomo.org"

This message will appear if you use a firewall or proxy server that breaks SSL connections and reencrypts them afterwards with its own certificate. Matomo uses a `cacert` file extracted from the Firefox browser and is therefore unable to connect to any SSL encrypted page with your specific setup.

To solve this issue you are able to provide your own `cacert` file that trusts your proxies certificate. To use your own file add `custom_cacert_pem = "/path/to/your/cacert.pem"` to the `[General]` section of your config. Make sure the file is readable by your webserver.

tsteur:
Awesome, thank you! Published https://matomo.org/faq/troubleshooting/faq_34226/

PR #14115, opened 2019-02-18
Realtime Log UI improvements/fixes

sgiehl:
- Initial tootips on icons didn't work
- positioning of icons was broken after #13916

### DIFF
--- a/plugins/Live/javascripts/live.js
+++ b/plugins/Live/javascripts/live.js
@@ -172,7 +172,7 @@
 
             var self = this;
 
-            this._initTooltips();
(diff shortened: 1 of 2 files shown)


diosmosis:
Looks like this causes a regression in the visitor log: https://builds-artifacts.matomo.org/matomo-org/matomo/realtimelog/32846/Live_visitor_profile.png I guess due to the inline-block

sgiehl:
I'm not able to reproduce that locally. Have retriggered the tests, let's see if it fails again...

sgiehl:
Seems this time the visitor profile tests passed.... Not sure why they failed before :man_shrugging:

PR #14121, opened 2019-02-21
Noindex, nofollow for login page and tracker default output

### LINKED ISSUE #6552: noindex, nofollow option
<meta name="robots" content="noindex, nofollow">

or an possibility to add own meta tags

tsteur:
Also increases security as less matomo instances will be found on search engines. 

fix https://github.com/matomo-org/matomo/issues/6552

### DIFF
--- a/core/Tracker/Response.php
+++ b/core/Tracker/Response.php
@@ -79,7 +79,7 @@ class Response
                 Common::sendResponseCode(400);
             }
             Common::printDebug("Empty request => Matomo page");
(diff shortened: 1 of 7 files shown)


sgiehl (approved):
Not sure what impact that will have regarding seo / page rank of matomo.org if all those possibly visible links are removed. Maybe @mattab has an opinion here. For me it's fine to merge

mattab:
Looks good to me (we're concerned the follow links may penalise us), just need to fix tests before merging

PR #14127, opened 2019-02-22
JS Tracker api: add ping() method

### LINKED ISSUE #14100: Add method to force heartbeat ping to Javascript API
There is no way to manually force a heartbeat ping request using the [Javascript Tracking Client API](https://developer.matomo.org/api-reference/tracking-javascript).  This is possible in the PHP API via [doPing()](https://developer.matomo.org/api-reference/PHP-Piwik-Tracker#doping) and in the HTTP API by setting the `ping=` parameter.

Forcing a ping can be desirable after calling a configuration setter, like `setCustomVariable` in order for it to be updated immediately.  Otherwise, the updated value is not sent to the server until the next tracking action.

pimlottc-gov:
Fixes #14100


js/piwik.js
```diff
@@ -7347,6 +7346,17 @@ if (typeof window.Piwik !== 'object') {
...
 
+            /**
+             * Sends a ping request.
+             *
+             * Ping requests do not track new actions. If they are sent within the standard visit length, they will
+             * extend the existing visit and the current last action for the visit. If after the standard visit
+             * length, ping requests will create a new visit using the last action in the last known visit.
+             */
+            this.doPing = function () {
+                trackRequest('ping=1', null, null, 'ping')
```
tsteur: This should be `this.trackRequest`

js/piwik.js
```diff
@@ -4526,8 +4526,7 @@ if (typeof window.Piwik !== 'object') {
             heartBeatPingIfActivityAlias = function heartBeatPingIfActivity() {
                 var now = new Date();
                 if (lastTrackerRequestTime + configHeartBeatDelay <= now.getTime()) {
-                    var requestPing = getRequest('ping=1', null, 'ping');
-                    sendRequest(requestPing, configTrackerPause);
+                    doPing();
```
tsteur: This should be `trackerInstance.doPing()`

js/piwik.js
tsteur: Also I wonder if we should rename it to maybe `trackPing()` as we usually name things like this `trackGoal, trackPageview, ...` or maybe `sendPing()` as it's not really tracking anything.

tsteur:
Cheers for this @pimlottc-gov  Left a few comments.

Ideally you also generate the up to date minified version see https://github.com/matomo-org/matomo/tree/3.x-dev/js#deployment and add an entry to the developer changelog but I suppose we could do this as well.

js/piwik.js
pimlottc-gov: I picked `doPing` to [match the PHP API](https://developer.matomo.org/api-reference/PHP-Piwik-Tracker#doping), but could easily change it.

tsteur: Good thinking. The JS tracker and PHP tracker do have bit different naming style though. Eg PHP tracker always adds the `do` which JS tracker doesn't. Better to maybe stay consistent within JS tracker and not use `do`.

pimlottc-gov: Okay, then how about just `ping`?

tsteur: sure 👍

pimlottc-gov:
> Ideally you also generate the up to date minified version see https://github.com/matomo-org/matomo/tree/3.x-dev/js#deployment and add an entry to the developer changelog but I suppose we could do this as well.

Thanks for the instructions, I went ahead and added a minified build.  I'm not sure, though, which changelog to update?

tsteur:
I've added the changelog entry and potentially fixed a test which doesn't liked a newline at the end. Will keep an eye on whether tests pass. Cheers for this 👍

tsteur:
Cheers @pimlottc-gov for this PR 👍

PR #14133, opened 2019-02-25
Propagate token auth in ajax requests for widgetized reports AND enable_framed_pages = 1.

diosmosis:


sgiehl:
@diosmosis  could you describe (or link) the issue that should be fixed by your changes here?

diosmosis:
@sgiehl There isn't one, it was from an innocraft support request. This makes embedding matomo in an iframe work w/ enable_framed_pages = 1 (so an iframe that links directly to `?module=CoreHome&action=index`,

sgiehl:
embedding the whole sites works with this branch. But I actually didn't see any obvious problems when embedding `index.php` in an iframe on 3.x-dev branch. Is there anything specific that didn't work before? Don't wont to waste time by clicking though the whole functions

diosmosis:
@sgiehl did you try logged out w/ a token_auth in the URL?

sgiehl:
Ah, guess I was still logged in. Will test that again tomorrow.

sgiehl (approved):
tested again. works as expected :+1:

PR #14136, opened 2019-02-26
Send email notification when user email changes.

### LINKED ISSUE #6125: Send email notifications when email address or passwords are changed
The goal of this ticket is to send an email when there is an email address or password change. Later this email could be extended to other security actions.

Your email address and password are how you safeguard your Piwik account, so monitoring it for any changes is critical. We could send emails when there are security related changes to your account, such as a password change. 

The information would give you more insight into when and where the account change took place, including 
* the date and time and
*  details on the device the device the changes were made on such as
* the browser it was running, 
* the Operating System (OS), 
* IP address, 
* and approximate physical location.

Example at linkedin:

password-change

### LINKED ISSUE #13533: Require email verification when changing email address.

diosmosis:
Fixes #6125

emailscreen

Not sure if the email text is ok, let me know if it should change.

### DIFF
--- a/plugins/UsersManager/API.php
+++ b/plugins/UsersManager/API.php
@@ -8,6 +8,7 @@
  */
 namespace Piwik\Plugins\UsersManager;
 
+use DeviceDetector\DeviceDetector;
 use Exception;
 use Piwik\Access;
 use Piwik\Access\CapabilitiesProvider;
@@ -17,6 +18,8 @@ use Piwik\Common;
 use Piwik\Config;
 use Piwik\Container\StaticContainer;
 use Piwik\Date;
+use Piwik\IP;
+use Piwik\Mail;
 use Piwik\Metrics\Formatter;
 use Piwik\NoAccessException;
 use Piwik\Option;
@@ -26,7 +29,7 @@ use Piwik\Plugins\Login\PasswordVerifier;
 use Piwik\SettingsPiwik;
 use Piwik\Site;
 use Piwik\Tracker\Cache;
-use Piwik\Url;
+use Piwik\View;
 
 /**
  * The UsersManager API lets you Manage Users and their permissions to access specific websites.
@@ -915,6 +918,10 @@ class API extends \Piwik\Plugin\API
 
         Cache::deleteTrackerCache();
 
+        if ($email != $userInfo['email']) {
+            $this->sendEmailChangedEmail($userInfo, $email);
+        }
+
         /**
          * Triggered after an existing user has been updated.
          * Event notify about password change.
@@ -1378,5 +1385,34 @@ class API extends \Piwik\Plugin\API
             }
         }
         return [$roles, $capabilities];
+    }
+
+    private function sendEmailChangedEmail($user, $newEmail)
+    {
+        $userAgent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
+
+        $uaParser = new DeviceDetector($userAgent);
+        $uaParser->parse();
+
+        $view = new View('@UsersManager/_emailChangedEmail.twig');
+        $view->accountName = Common::sanitizeInputValue($user['login']);
+        $view->newEmail = Common::sanitizeInputValue($newEmail);
+        $view->deviceName = $uaParser->getDeviceName();
+        $view->ipAddress = IP::getIpFromHeader();
+
+        $mail = new Mail();
+
+        // send the mail to both the old email and the new email
+        $mail->addTo($newEmail, $user['login']);
+        $mail->addTo($user['email'], $user['login']);
+        $mail->setSubject(Piwik::translate('UsersManager_EmailChangeNotificationSubject'));
+        $mail->setDefaultFromPiwik();
+        $mail->setWrappedHtmlBody($view);
+
+        $replytoEmailName = Config::getInstance()->General['login_password_recovery_replyto_email_name'];
+        $replytoEmailAddress = Config::getInstance()->General['login_password_recovery_replyto_email_address'];
+        $mail->setReplyTo($replytoEmailAddress, $replytoEmailName);
+
+        $mail->send();
     }
 }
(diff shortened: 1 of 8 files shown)


plugins/UsersManager/API.php
```diff
@@ -1379,4 +1386,33 @@ private function getRoleAndCapabilitiesFromAccess($access)
...
+    {
+        $userAgent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
+
+        $uaParser = new DeviceDetector($userAgent);
+        $uaParser->parse();
+
+        $view = new View('@UsersManager/_emailChangedEmail.twig');
+        $view->accountName = Common::sanitizeInputValue($user['login']);
+        $view->newEmail = Common::sanitizeInputValue($newEmail);
+        $view->deviceName = $uaParser->getDeviceName();
```
sgiehl: Might it maybe make sense to add some more details about the device if available? `getDeviceName` only give `desktop, smartphone, ...` Maybe we could if available also add `getBrand()` and `getModel()`. So we could print something like `smartphone (Sony Xperia Z5)` or only `dekstop` if not specified

sgiehl (approved):
Works nicely. Maybe @mattab or @tsteur should review the text as well.
Left a small suggestion, but it's not really necessary. Good to merge if text is approved.

Findus23:
:+1: Looking great. 

But I'd prefer to keep #13533 open as this doesn't solve the spam-possibility or the double-opt in feature. (even though I am not sure anymore how important it is)

Instead https://github.com/matomo-org/matomo/issues/6125 can be closed.

plugins/UsersManager/API.php
diosmosis: 👍 will do

diosmosis:
@Findus23 I can add a password changed email.

mattab:
Feedback:
* Looks good!
* To make email more friendly, let's add 
  * at the start "Hello,"
  * at the end:  "Thank you for using Matomo! <br>The Matomo team"  or similar?

> I can add a password changed email

Yes be important to notify on password change too :+1:

diosmosis:
Updated.

sgiehl (changes requested):
left some minor comments. Otherwise looks good to be merged.
Might that be something that needs to be considered for WhiteLabel plugin? ping @tsteur

plugins/UsersManager/templates/_emailChangedEmail.twig
```diff
@@ -0,0 +1,9 @@
+<p>{{ 'General_HelloUser'|translate('<strong>' ~ accountName ~ '</strong>')|raw }}</p>
+
+<p>{{ 'UsersManager_EmailChangedEmail1'|translate('<strong>' ~ newEmail ~ '</strong>')|raw }}.</p>
+
+{% set deviceDescription %}{{ deviceName }}{% if deviceBrand is not empty or deviceModel is not empty %} ({% if deviceBrand is not empty %}{{ deviceBrand }}{% endif %}{% if deviceModel is not empty %} {{ deviceModel }}{% endif %}){% endif %}{% endset %}
```
tsteur: is the device name always set? Looking at the code it may be empty but probably isn't really often the case. May not be too useful though to show whether it was updated from desktop or mobile if no device was detected. Not too important though and fine to leave it just like that.

plugins/UsersManager/API.php
```diff
@@ -1379,4 +1390,71 @@ private function getRoleAndCapabilitiesFromAccess($access)
         }
         return [$roles, $capabilities];
     }
+
+    private function sendEmailChangedEmail($user, $newEmail)
```
tsteur: could we maybe mostly reuse the same code for `sendEmailChangedEmail` and `sendPasswordChangedEmail ` and also it's templates and only have different mail subject and also different leading sentence in the email? not too important though.

plugins/UsersManager/templates/_emailChangedEmail.twig
tsteur: I reckon there can be various cases where user agent may not be set when details are updated through the API. Therefore device name may be empty on ocassion

plugins/UsersManager/API.php
diosmosis: Thought about it, but since there were just two emails, and the code isn't really complicated (just setting things on the view & Mail instance), it didn't seem that important. Maybe if there are more changes to a user that we should notify them about?

diosmosis:
@tsteur did you see @sgiehl's comment about WhiteLabel?

tsteur:
Yes, sorry forgot to reply. I haven't tested it but I think it should be all fine as it also works for all the other mails. Might need to double check at some point though whether we replace the brand eg in all `Mails.send` etc

diosmosis:
Updated.

tsteur:
minor last tweak... could we use `ucfirst` for device name? (desktop)
image

Also I thought it was sending out duplicate emails until I realised it was actually all fine but the source and target email was the same but `source=test+1@domain.com` and `target=test+2@domain.com`. No bug and nothing needs to change. Just in case someone tests it as well and runs in the same issue.

Looks good to merge if tests pass

mattab:
@diosmosis in my tests, when the email address is updated, it seems the email is sent twice, could you double check emails are not sent twice?

tsteur:
See the comments... you aren't change your email to your own email are you?

mattab:
> you aren't change your email to your own email are you?

oops I should have seen this one coming :-D  No more issue.

PR #8454, opened 2015-07-29
Added config to set email name in email FROM field

tsteur:
fixes #8252

PR #14137, opened 2019-02-26
Bugfix mail from must equal authorized user when sending email reports

### LINKED ISSUE #12426: Mail from must equal authorized user
Version:Piwik 3.2.1
When i send a mail report, it occurs the error: Mail from must equal authorized user

Is there something wrong in my mail configuration?

frankgx97:
fix #12426

Hi there,

For some email  providers (eg. qq.com, exmail.qq.com), sending emails via smtp without providing "from address" will get the "mail from must equal authorized user" error.

Though PR #8454 added "noreply_email_address" and "noreply_email_name" field to the config file, it placed these fields to the "General" section, instead of "mail" section, which is misleading. And it cannot be set from the General Settings page.

I moved the two options to the "mail" section of the config file, and added "SMTP from address" and "SMTP from name" to the General Settings page to enable users to configure these two fields from the Settings page.

Regards.


plugins/CoreAdminHome/templates/generalSettings.twig
```diff
@@ -129,6 +129,20 @@
                      autocomplete="off">
                 </div>
 
+                <div piwik-field uicontrol="text" name="mailFromAddress"
+                     ng-model="mailSettings.mailFromAddress"
+                     title="{{ 'General_SmtpFromAddress'|translate|e('html_attr') }}"
+                     value="{{ mail.noreply_email_address }}" inline-help="{{ 'General_UsuallyYourEmailAddress'|translate|e('html_attr') }}"
```
tsteur: I'm not 100% sure but it may need to be `{{ mail.noreply_email_address|e('html_attr') }}` to escape it re security

plugins/CoreAdminHome/templates/generalSettings.twig
```diff
@@ -129,6 +129,20 @@
...
+                     autocomplete="off">
+                </div>
+
+                <div piwik-field uicontrol="text" name="mailFromName"
+                     ng-model="mailSettings.mailFromName"
+                     title="{{ 'General_SmtpFromName'|translate|e('html_attr') }}"
+                     value="{{ mail.noreply_email_name }}" inline-help="{{ 'General_NameShownInTheSenderColumn'|translate|e('html_attr') }}"
```
tsteur: I'm not 100% sure but it may need to be `{{ mail.noreply_email_name |e('html_attr') }}` to escape it re security

plugins/CoreAdminHome/templates/generalSettings.twig
tsteur: Same probably for the other fields above that were already defined

core/Mail.php
```diff
@@ -42,15 +42,21 @@ public function setDefaultFromPiwik()
         /** @var Translator $translator */
         $translator = StaticContainer::get('Piwik\Translation\Translator');
 
-        $fromEmailName = Config::getInstance()->General['noreply_email_name'];
+        $fromEmailName = Config::getInstance()->mail['noreply_email_name'];
+        if(empty($fromEmailName)){
+            $fromEmailName = Config::getInstance()->General['noreply_email_name'];
```
tsteur: I reckon it might be better to create an update script using the command `./console generate:update --component=core` to migrate both config settings from `General` to `mail` within the `doUpdate()` method. This way it will be easier to change the value for users in the future (by directly seeing the default value `noreply@{DOMAIN}`) and it won't be defined in 2 places (less confusion).

You can see an example here: https://github.com/matomo-org/matomo/blob/3.x-dev/core/Updates/2.15.0-b12.php#L25-L41 

This means you can set the default value of `noreply@{DOMAIN}` for `[mail]noreply...`

tsteur:
Thanks for this @nyanim 👍  very appreciated. Left a comment in the PR

config/global.ini.php
```diff
@@ -869,6 +869,8 @@
 username = ; SMTP username
 password = ; SMTP password
 encryption = ; SMTP transport-layer encryption, either 'ssl', 'tls', or empty (i.e., none).
+noreply_email_address = ; standard email address displayed when sending emails
```
tsteur: We will also need to remove these settings from the `General` section

lang/zh-cn.json
```diff
@@ -292,6 +292,8 @@
         "OneVisit": "1 次访问",
         "OnlyEnterIfRequired": "在您的 SMTP 服务器需要用户名时才需要输入",
         "OnlyEnterIfRequiredPassword": "在您的 SMTP 服务器需要密码时才需要输入",
+        "UsuallyYourEmailAddress": "通常是您的Email地址",
```
tsteur: Unfortunately, we can only include english default translations. Other languages will need to be translated through our translation tool see https://matomo.org/translations

katebutler:
Duplicate of #14975

PR #14145, opened 2019-02-27
Report an error when the bzopen fails in system test sync command

katebutler:
The script was reporting a successful result, but was actually unable to extract the tar file because the PHP bz2 extension was missing.


diosmosis:
LGTM

PR #14146, opened 2019-02-27
fix queued requests in JS tracker and add possibility to disable it

### LINKED ISSUE #14143: JS tracker: Offer ability to disable queued tracking request
In Matomo 3.7 or 3.8 we added a feature for plugins to queue tracking requests so they can be sent together. It would be good if users had the ability to disable this feature in which case we send each queued request immediately.

When do you want to disable this feature? If you make for example use of the log import feature.

The method `tracker.queueRequest()` would then not push the request to the queue but call basically `trackRequest`

Be good to add an entry to the changelog and document this in the FAQ

tsteur:
fix https://github.com/matomo-org/matomo/issues/14143

I noticed while `queueRequest` was working in the tests and in DEV, it was broken in the release because `requestQueue` was placed within the `/*</DEBUG>*/` section and therefore removed in the production JS.


diosmosis:
@tsteur tracker JS tests are failing, think because the queue isn't re-enabled after disabling in tests.

tsteur:
@diosmosis they were succeeding locally for me cause I was only executing one test group. Was able to reproduce the error by running all JS tracker tests.

Fixed it locally but for some reason still fails on travis. Can you try to execute the tests on your local server to see if they succeed?

Had to move the `_beforeUnloadHandler` test to the bottom because it was actually setting `isPageUnloading` to `true` and all test suites after that were actually using `isPageUnloading=true` and therefore the queue wasn't actually adding a request and also it was even using `sendBeacon` instead of XHR in the tests.

diosmosis:
@tsteur I see one error when running all tests locally:

image

But it doesn't look like this the failure travis is having...

tsteur:
Tests should be fixed now. Was able to reproduce it once I also executed the Tag Manager plugin tests along the regular Matomo JS Tracker tests.

Basically, there is an `Unload Tag Manager Trigger` which tested its unload behaviour by triggering the `unload` event. This  also triggered the JS tracker unload event and therefore the JS tracker thought from the beginning of the tests that the page is unloading and wouldn't queue the request but track it directly. Now after each test unsetting the "page is unloading" variable

PR #14147, opened 2019-02-27
Add new parameter to Live.getLastVisitsDetails API to request more detailed information

tsteur:
I know it looks funny that parameter isn't used.

In Media Analytics we're adding a new feature around heatmaps and we do not want to expose this information by default through the `Live.getLastVisitsDetails` API because it is a wee bit slower and could return a lot of data. However, it can be useful to request it and we have users that ask specifically for that. I figured it be good to only return this information in the API when an additional flag is set. I named it `enhanced` as we use this already in `MultiSites` API and be good for consistency.

In order for this parameter to still appear in the documentation I've added it to the method here and eventually other plugins could do similar based on this. 

We don't have to merge it if it's a problem that the parameter isn't used in core. Be good to have it though maybe.

### DIFF
--- a/plugins/Live/API.php
+++ b/plugins/Live/API.php
@@ -157,9 +157,10 @@ class API extends \Piwik\Plugin\API
      * @param bool|int $minTimestamp (optional) Minimum timestamp to restrict the query to (useful when paginating or refreshing visits)
      * @param bool $flat
      * @param bool $doNotFetchActions
+     * @param bool $enhanced for plugins that want to expose additional information
      * @return DataTable
      */
-    public function getLastVisitsDetails($idSite, $period = false, $date = false, $segment = false, $countVisitorsToFetch = false, $minTimestamp = false, $flat = false, $doNotFetchActions = false)
+    public function getLastVisitsDetails($idSite, $period = false, $date = false, $segment = false, $countVisitorsToFetch = false, $minTimestamp = false, $flat = false, $doNotFetchActions = false, $enhanced = false)
     {
         Piwik::checkUserHasViewAccess($idSite);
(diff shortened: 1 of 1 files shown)


diosmosis:
Could `showColumns/hideColumns` be used instead, ie w/ an event or some mechanism to allow plugins to add to the default values? I guess it wouldn't work for information nested in the data structure unless we extend the parameters...

tsteur:
`show/hideColumns` would restrict what is returned and then you would only get that information but not any other. Besides it not working for these nested actions in the Live API.

diosmosis (approved):
CC @mattab for final approval of adding parameter in core (since I don't make final decisions about platform additions)

sgiehl:
wondering if we should pass through that parameter to the VisitorDetails classes, so they can set additional details or not there as well...

mattab:
LGTM

tsteur:
@sgiehl adding a parameter to visitorDetails would break the API wouldn't it? Like the PHP methods would have an additional parameter?

sgiehl:
we could add a new static method like `enableEnhancedMode` in the abstract class setting a static variable that is then accessible in all VisitorDetails classes or something like that? Not an ideal solution, but wouldn't break any API

tsteur:
sure that works... something like `isEnhancedModeEnabled` maybe and could be detected from that API parameter 👍   doesn't have to be static I suppose. Could be just a protected method or public method.

diosmosis:
@tsteur have time to add the property to VisitorDetails? if not, I'll merge & create a new issue.

tsteur:
Don't really have time :(

diosmosis:
np, will create issue (or maybe PR)

PR #14148, opened 2019-02-28
Run AllTests against PHP 7.3 on travis

### LINKED ISSUE #11936: Run automated tests against PHP 7.3 to ensure full compatibility
In PHP 7.2 some new methods will be deprecated and trigger notices see https://devzone.zend.com/7628/deprecations-php-7-2/

I have noticed eg `create_function()` is being used and `$php_errormsg`. Didn't check all deprecated functions. Goal would be to update libs and code needed to not get any notices.

sgiehl:
Let's see if anything is still failing on PHP > 7

fixes #11936


fdellwing:
//cc https://forum.matomo.org/t/empty-or-invalid-response/31848

Not sure if there is a testcase for `unserialize()` right now?

sgiehl:
waiting for #14158 as that should fix the remaining failing tests

PR #14157, opened 2019-03-02
Pagination localisation clean-up

c960657:
Various pagination issues:

* The MultiSites, ~ScheduledReports~ and SitesManager plugins use the `General_Pagination` translation key, but they still define their own copy that is no longer used. I have deleted those.
* A different variant, `CoreHome_PageOf` only takes two placeholders. The number range is always concatenated using a hyphen, but for some locales this is not the preferred formatting. I have removed this string in favour of `General_Pagination`.
* On the Sites Manager, the pagination text at the bottom says “0-10 of 14”. I changed this to “1-10 of 14”.
* Many locales separate the interval endpoints with spaces and hyphen, “1 - 10 of 14” instead of the more compact “1–10 of 14”. I suspect this is inspired by the English source string. However, according to CLDR (and other sources), the proper way to specify a numeric interval is with an em-dash and  without spaces. So using a script, I updated all locales to use the [CLDR number range pattern](https://github.com/unicode-cldr/cldr-numbers-full/blob/master/main/en/numbers.json#L146) together with the pattern specified for the previous key `CoreHome_PageOf`. Some locales used different translations for the “x of y” part of `General_Pagination` and `CoreHome_PageOf`. However, `PageOf` was used prominently in data tables, so I assume the translations are generally of high quality.


sgiehl:
@c960657 Thanks for the PR. Could you please revert all changes to translation files other than `*/en.json`?
All changes made to `en.json` files are automatically pulled from transifex. All changed translations are then unset and translators can translate them again. All changes to other languages would later be overwritten when pulling the translations from transifex.

If there is a way to generate the pagination translations using CLDR, it might make sense to include the needed stuff in the script to fetch the data and move them to the Intl plugin

c960657:
I have reversed the changes to non-English JSON files. Thanks for clarifying the process.

>If there is a way to generate the pagination translations using CLDR, it might make sense to include the needed stuff in the script to fetch the data and move them to the Intl plugin

CLDR does not have translations specifically for pagination (x-y of z), only for a number range (x-y). I'm not sure this pattern is always the most appropriate for pagination, so I think preserving the Transifex translation here is the safest choice.

diosmosis:
If you're willing to make another change, UsersManager has another duplicate pagination string (UsersManager_XtoYofN). Other than failing tests, LGTM (the tests can be fixed before merging, not a requirement for the PR).

c960657:
`UsersManager_XtoYofN` is not also replaced with `General_Pagination`.

diosmosis:
There appears to be a regression in the multisites dashboard: https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/33079/UIIntegrationTest_widgetize_allwebsites.png

It's not displaying the first site, perhaps due to the `+ 1` change.

c960657:
Fixed the off-by-one error.

diosmosis:
@c960657 I think the pagination numbers might be off, eg when all websites is limited to and display the 50 websites, the pagination label will say `1-49`, when there are 50 rows:

mamp_8888_index php_module multisites action index idsite 1 period day date yesterday

Could you double check the pagination label matches the number of rows displayed everywhere you made offset/limit changes?

c960657:
All pages double-checked.

diosmosis:
Thanks for another contribution @c960657 !

PR #14158, opened 2019-03-02
Improve total report values calculation

sgiehl:
While checking tests for PHP 7.3 in #14148 it turned out that in many tests a `A non well formed numeric value encountered` error occurs.
This happens as the total report calculation tries to sum up any values regardless of their type.
To prevent that I've added a check for `is_numeric`
Also I thought it might be good to deactivate the total calculation for minimum, maximum, rate or nested metrics, as it doesn't make sense to sum up those values at all.

### DIFF
--- a/plugins/API/ProcessedReport.php
+++ b/plugins/API/ProcessedReport.php
@@ -717,7 +717,7 @@ class ProcessedReport
         foreach ($simpleTotals as $metric => $value) {
             if (!array_key_exists($metric, $totals)) {
                 $totals[$metric] = $value;
-            } else {
+            } else if(is_numeric($value)) {
                 $totals[$metric] += $value;
             }
         }
--- a/plugins/API/ProcessedReport.php
+++ b/plugins/API/ProcessedReport.php
@@ -715,8 +715,8 @@ class ProcessedReport
         $simpleTotals = $this->hideShowMetrics($metadataTotals);
 
         foreach ($simpleTotals as $metric => $value) {
-            if (0 === strpos($metric, 'avg_') || '_rate' === substr($metric, -5)) {
-                continue; // skip average and rate metrics
+            if (0 === strpos($metric, 'avg_') || '_rate' === substr($metric, -5) || '_evolution' === substr($metric, -10)) {
+                continue; // skip average, rate and evolution metrics
             }
             if (is_array($value) ) {
                 continue; // skip totals for nested metrics
--- a/plugins/API/ProcessedReport.php
+++ b/plugins/API/ProcessedReport.php
@@ -718,8 +718,8 @@ class ProcessedReport
             if (0 === strpos($metric, 'avg_') || '_rate' === substr($metric, -5) || '_evolution' === substr($metric, -10)) {
                 continue; // skip average, rate and evolution metrics
             }
-            if (is_array($value) ) {
-                continue; // skip totals for nested metrics
+            if (is_array($value) || !is_numeric($value)) {
+                continue; // skip totals for nested or not numeric metrics
             }
 
             if (!array_key_exists($metric, $totals)) {
@@ -728,7 +728,7 @@ class ProcessedReport
                 $totals[$metric] = min($totals[$metric], $value);
             } else if(0 === strpos($metric, 'max_')) {
                 $totals[$metric] = max($totals[$metric], $value);
-            } else if(is_numeric($value)) {
+            } else {
                 $totals[$metric] += $value;
             }
         }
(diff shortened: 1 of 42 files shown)


plugins/API/ProcessedReport.php
```diff
@@ -715,9 +715,17 @@ private function aggregateReportTotalValues($simpleDataTable, $totals)
         $simpleTotals = $this->hideShowMetrics($metadataTotals);
 
         foreach ($simpleTotals as $metric => $value) {
+            if (0 === strpos($metric, 'avg_') || '_rate' === substr($metric, -5) ||
+                0 === strpos($metric, 'min_') || 0 === strpos($metric, 'max_')) {
```
tsteur: @sgiehl in general we do want min/max values to be considered here. `$row->sumRow` should notice whether it's min/max etc. and if not we could make the `sumRow` logic understands what to do. You actually want to have the max value of all of the rows if there is such a metric for example so you can apply processed metrics etc. if / when needed and allows for quite useful features maybe.

plugins/API/ProcessedReport.php
```diff
@@ -715,9 +715,17 @@ private function aggregateReportTotalValues($simpleDataTable, $totals)
...
+                continue; // skip average, minimum, maximum and rate metrics
+            }
+            if (is_array($value) ) {
+                continue; // skip totals for nested metrics
+            }
+
             if (!array_key_exists($metric, $totals)) {
                 $totals[$metric] = $value;
-            } else {
+            } else if(is_numeric($value)) {
```
tsteur: did you see why there are non numeric values? maybe we need to fix something? Just wanting to avoid we just "hide" the actual problem

sgiehl: the failures I've seen seems to be because of percentage values e.g. `15%` or nested data in arrays. So with the checks above it wouldn't be needed anymore. But as it could still for custom plugins we maybe should do the check nevertheless

plugins/API/ProcessedReport.php
sgiehl: ok. will fix the calculation for min/max values then instead. Wasn't sure what the feature is actually used for besides the totals row (that shouldn't be displayed for such metrics)

tests/PHPUnit/System/expected/test_ManyVisitorsOneWebsiteTest_sortByProcessedMetric_constantRowsCountShouldKeepEmptyRows__API.getProcessedReport_day.xml
```diff
@@ -402,23 +402,8 @@
...
-				<nb_conversions>8</nb_conversions>
-				<nb_visits_converted>8</nb_visits_converted>
-				<revenue>40</revenue>
-			</row>
-			<row idgoal="2">
-				<nb_conversions>4</nb_conversions>
-				<nb_visits_converted>4</nb_visits_converted>
-				<revenue>20</revenue>
-			</row>
-		</goals>
```
diosmosis: Is it expected we're removing total values for goal specific metrics? Seems like it might be useful to keep these.

sgiehl: I don't think those values were summed up before... It's a `+` operation on arrays, so guess they were combined and didn't represent a total value. If we want to keep them guess we would need to implement that in a recursive way?

diosmosis: I suppose so... @mattab / @tsteur worth keeping these totals (and calculating them correctly)?

tsteur: Ideally yes. Can we maybe use `$row->sumRow` there which may does that?

sgiehl: should now be processed recursively

sgiehl:
updated the implementation to also calculate nested values correctly. Ready for another review.

PR #14164, opened 2019-03-04
Add goal report metadata for overview.

### LINKED ISSUE #14162: Error in Goals
We are getting the error on demo.matomo.org
Error: requested report Goals.getVisitUntilConversion for websiteid = 62 not found in list of available reports.

Steps to reproduce.
1) Goals -> Overview -> Visits to Conversions -> Hover on any visit which has conversions -> Click on Open Row Evolution icon.

diosmosis:
The metadata entry for idGoal=0 was missing, causing row evolution to fail (since it requests a processed report w/ idGoal=0 in this case).

Fixes #14162


tsteur:
👍  looks good to merge, failing system tests will need to be updated though.

PR #14165, opened 2019-03-06
Do not calculate report totals if totals metadata is found.

diosmosis:
Otherwise, it will request the totals row when it doesn't need to.

(This was causing an issue w/ ProxySite since the target would no longer keep the totals row, causing the proxy to try to compute it.)

### DIFF
--- a/core/API/DataTableManipulator/ReportTotalsCalculator.php
+++ b/core/API/DataTableManipulator/ReportTotalsCalculator.php
@@ -87,7 +87,8 @@ class ReportTotalsCalculator extends DataTableManipulator
         $firstLevelTable = $this->makeSureToWorkOnFirstLevelDataTable($dataTable);
 
         if (!$firstLevelTable->getRowsCount()
-            || $firstLevelTable->getTotalsRow()
+            || $dataTable->getTotalsRow()
+            || $dataTable->getMetadata('totals')
         ) {
             return $dataTable;
         }


plugins/CoreHome/javascripts/dataTable.js
```diff
@@ -1306,7 +1306,14 @@ $.extend(DataTable.prototype, UIControl.prototype, {
             .each(function () {
                 setText(this, 'keep_totals_row', 'CoreHome_RemoveTotalsRowDataTable', 'CoreHome_AddTotalsRowDataTable');
             })
-            .click(generateClickCallback('keep_totals_row'));
+            .click(generateClickCallback('keep_totals_row', null, function () {
+                var newValue = self.param.keep_totals_row ? '0' : '1';
+                self.param.keep_totals_row = self.param.totals = newValue;
+                return {
+                    keep_totals_row: newValue,
+                    totals: newValue
```
tsteur: I haven't tested it, but wouldn't this disable the percent feature when you have a column for example in page report when you hover the pageviews column then it is supposed to show the percentages in grey?

diosmosis: Oh, hm, maybe (would be caught by one of the UI tests I believe). Originally I wanted to unset totals for the proxy request so we wouldn't calculate it again no matter what, but there wasn't a reliable way to do that.

Actually, maybe in the filter we can just check for the 'totals' metadata, not just the totals row.

diosmosis:
@tsteur updated to use better approach

tsteur (approved):

PR #14171, opened 2019-03-09
fix diagnostics template

Findus23:
I have been seeing this error quite often on my instance in Sentry
```
ErrorException: Array to string conversion
#33 vendor/twig/twig/lib/Twig/Extension/Core.php(740): handleError
```

It seems to be caused by my CustomiseTranslations plugin. Because when you are viewing the "Config File" Diagnostic page it tries to display the following data (as the plugin stores a list of dictionaries in the config)
```php
[{$$hashKey: object:457, translationKey: CorePluginsAdmin_PluginsManagement, translationText: This is text!!}, {$$hashKey: object:467, translationKey: , translationText: }]
```

My fix isn't great and I can't promise it works well, but maybe someone else can think of something more simple.

### DIFF
--- a/plugins/Diagnostics/stylesheets/configfile.less
+++ b/plugins/Diagnostics/stylesheets/configfile.less
@@ -7,6 +7,10 @@
     font-style: italic;
   }
 
+  .pre {
+    white-space: pre-wrap;
+  }
+
   td.name {
     max-width: 330px;
     word-wrap: break-word;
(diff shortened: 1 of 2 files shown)


diosmosis:
Would displaying the value as json be helpful?

tsteur:
Showing it json encoded sounds good 👍

tsteur:
@Findus23 should we maybe show it as json? or prefer current logic?

Findus23:
Sorry, I missed this PR.

I didn't think about JSON when creating the fix, this way it is easier and prettier.

PR #14175, opened 2019-03-11
Improve wording on anonymous user settings page

### LINKED ISSUE #13592: Make it clear, that giving anonymous users access to the login page does not remove existing view access for them
When the anonymous user has view access to one or more websites, you can access the following settings.
grafik
Setting this to `The login page` shows anonymous users only the login page, but they still have access to all websites they have view access (e.g. via the API).
The setting should make it absolutely clear that this does not revoke any kind of access, but just changes the homepage.

BTW: The German translation is even more confusing:
grafik

katebutler:
Fixes #13592


diosmosis:
FYI @Findus23 merged this, I think the wording is better, feel free to re-open the issue if you have some more thoughts

PR #14176, opened 2019-03-11
Add password strength indicator to user mgmt forms

### LINKED ISSUE #13070: Better password handling during installation
I think it'd be helpful for the admin to have the following dynamic (JS-driven) indicators, just like WordPress:

- Indicator for weak passwords
- Indicator for when passwords don't match

katebutler:
Uses the [jquery.pwstrength]https://github.com/matoilic/jquery.pwstrength) library.

pw-very-weak

pw-very-strong

Fixes #13070

### DIFF
--- a/LEGALNOTICE
+++ b/LEGALNOTICE
@@ -117,6 +117,10 @@ THIRD-PARTY COMPONENTS AND LIBRARIES
 	Link:  http://mths.be/placeholder
 	License:  Dual licensed: MIT (Expat) or GPL
 
+	Name:  jquery.pwstrength
+	Link:  http://matoilic.github.io/jquery.pwstrength/
+	License:  Dual licensed: MIT or GPL
+
 	Name:  qrcode.js
 	Link:  https://github.com/davidshimjs/qrcodejs
 	License:  MIT
--- a/libs/pwstrength/README.md
+++ b/libs/pwstrength/README.md
@@ -1,0 +1,26 @@
+# jQuery PWStrength #
+
+PWStrength is a jQuery plugin to indicate the strength of passwords.
+
+## Usage & Demo ##
+
+http://matoilic.github.com/jquery.pwstrength
+
+## Version History ##
+
+**0.1**
+
+* initial release
+
+**0.1.1**
+
+* updated plugin specification for the jquery plugin repository
+
+## Licence ##
+
+Copyright &copy; 2013 Mato Ilic <<info@matoilic.ch>>
+
+jquery.placeholder is dual licensed under the MIT and GPL licenses:
+
+* http://www.opensource.org/licenses/mit-license.php 
+* http://www.gnu.org/licenses/gpl.html
--- a/libs/pwstrength/jquery.pwstrength.js
+++ b/libs/pwstrength/jquery.pwstrength.js
@@ -1,0 +1,66 @@
+/**
+ * jquery.pwstrength http://matoilic.github.com/jquery.pwstrength
+ *
+ * @version v0.1.1
+ * @author Mato Ilic <info@matoilic.ch>
+ * @copyright 2013 Mato Ilic
+ *
+ * Dual licensed under the MIT and GPL licenses:
+ * http://www.opensource.org/licenses/mit-license.php
+ * http://www.gnu.org/licenses/gpl.html
+ */
+;(function($) {
+    $.pwstrength = function(password) {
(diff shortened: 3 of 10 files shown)


diosmosis:
I haven't looked into how it would have to be used, but would https://github.com/dropbox/zxcvbn be usable here? Given it was created by dropbox, it might do a better job than the jquery lib.

tsteur:
We had a look at that lib but I don't think an 800KB lib should be needed for a password check ;) We found the jquery lib is lightweight but does the most important job.

sgiehl:
@tsteur just out of curiosity, what is the plan regarding JS Frameworks? Wasn't it getting rid of jQuery and switching to AngluarJS? If so, we might better look for a small AngularJS lib for password strength.

Findus23:
I'm not sure we should add a dependency to 6 year old code just for 30 lines of JS.

I guess it should be possible to simply write the same feature in pure JS or Angular.

https://github.com/matoilic/jquery.pwstrength/blob/d6f07a13eba1479f481d0beceb612f439db406ae/jquery.pwstrength.js#L13-L44

But the main question is what excactly is a secure password (see https://xkcd.com/936/).
And when you take into account all possibilities (e.g. that `Password1!` is not a strong password), you either end up with a gigantic library [like the one by dropbxox](https://lowe.github.io/tryzxcvbn/) or start comparing it to [large databases of common existing passwords](https://plugins.matomo.org/PasswordVerifier) which also isn't perfect.

fdellwing:
I'm not a fan of such password strength indicators as they mostly give a false sense of security for awful passwords. So either use `zxcvbn` (or similar) or nothing at all.

diosmosis:
@tsteur looks like there's a php port (linked on the github repo desc). We could invoke it through ajax, what do you think?

Findus23:
@diosmosis It wouldn't be realtime, but I think that's not that bad as the library can even return a list of suggestions on how to improve the password.

The question is if this should be part of Matomo or a plugin.

diosmosis:
If it's meant to increase baseline security for users, I would think it should be part of Matomo, not an optional plugin. I was addressing the issue of it being an 800kb JS library. If the PHP port is used, then only minimal JS is added to the client.

tsteur:
> @tsteur just out of curiosity, what is the plan regarding JS Frameworks? Wasn't it getting rid of jQuery and switching to AngluarJS? If so, we might better look for a small AngularJS lib for password strength.

There's not really a plan, although for sure be great. The main plan is to eventually get rid of twig templates and PHP controllers. But that might be many years away :) If there's a simple lib that works and does the job based on jQuery I would say why not. Probably before that, we need to at some point see how to upgrade from Angular 1.X to a newer angular version but this will be big project in itself.

> I guess it should be possible to simply write the same feature in pure JS or Angular.

@Findus23 Can you maybe explain what the point is to rewrite it in pure JS or Angular? When there's already something that works etc?

@Findus23 @fdellwing I had a quick look on http://matoilic.github.io/jquery.pwstrength/ before we chose the lib and it was giving OK suggestions. You could of course also slightly adjust the code. A 1KB be obviously not as as good as a `zxcvbn` 800KB lib but it does the job and gives a better indication than it was before (which was none).

@diosmosis @Findus23 we could port it to PHP. The lib is also > 2MB though and more work for a feature that is currently not so relevant for us in terms of priority. I would say we either go with the jQuery lib here or we close the PR and move the issue back to the backlog. We only decided to implement and work on the issue cause it's a quick solution with the jQuery version that doesn't add much overhead. Happy either way.

diosmosis:
@tsteur it has already been ported to php: https://github.com/bjeavons/zxcvbn-php (the link is on the main zxcvbn github repo). Would just using this be an ok amount of work?

tsteur:
@diosmosis I know that's why I meant it's > 2MB. I meant the work to do the ajax, change wording in the UI based on result, colours, etc. The issue has just no priority currently.

diosmosis:
Ok, I'll leave it alone then.

fdellwing:
> I had a quick look on http://matoilic.github.io/jquery.pwstrength/ before we chose the lib and it was giving OK suggestions.

@tsteur I do see that differently. The lib does not look at entropie at all, but simply checks for "complex" chars. Ignoring both length and possible common char replacement in the password. I would therefor say: The lib is worse than not having an indicator at all, because

> give a false sense of security for awful passwords.

tsteur:
OK no worries. We can close the PR and move it out of the milestone for now.

PR #14184, opened 2019-03-11
New dimension for segmenting visitors by fingerprint

### LINKED ISSUE #14101: Visitor raw data API: expose the "fingerprint" value of each customer in API responses (log_visit.config_id)
In Matomo our goal is to make all the data available to users via the API.
One of the [piece of data](https://matomo.org/faq/general/faq_18254/) we collect is what we call the Fingerprint ([more details about how fingerprint is used](https://matomo.org/faq/general/faq_21418/)) (this is the value of the column `log_visit.config_id`)

There are 2 limitations currently:

* Fingerprint value is not exposed in Live.getLastVisitsDetails ([the API used to export RAW visitor data](https://matomo.org/faq/how-to/faq_24536/))
* Fingerprint is not available as a segment

The goal of this issue is to make these 2 changes so the fingerprint value is exposed and can be queried and used by users (and used as a segment).

katebutler:
Fixes #14101


plugins/CoreHome/Columns/VisitorFingerprint.php
```diff
@@ -0,0 +1,39 @@
...
+    protected $segmentName = 'fingerprint';
+    protected $acceptValues = '1eceaa833348b187 - any 16 Hexadecimal chars ID, which can be fetched from API.getLastVisitsDetails';
+    protected $allowAnonymous = false;
+    protected $sqlFilterValue = array('Piwik\Common', 'convertVisitorIdToBin');
+
+    public function configureMetrics(MetricsList $metricsList, DimensionMetricFactory $dimensionMetricFactory)
+    {
+        $metric = $dimensionMetricFactory->createMetric(ArchivedMetric::AGGREGATION_UNIQUE);
+        $metric->setTranslatedName(Piwik::translate('Visitor_Fingerprint'));
+        $metric->setName('nb_uniq_visitors_fingerprint');
```
diosmosis: Would we still need this call if we set `$metricId` to 'visitors_fingerprint'?

tsteur: When you set `visitors_fingerprint ` as segment name, then the autosuggest doesn't work automatically anymore.

diosmosis: I meant as `$metricId`, not `$segmentName`, unless they are related? Not sure I've ever used this API before... :)

diosmosis (approved):
Works for me, code looks good, needs a rebase/merge w/ 3.x-dev before it can be merged. 👍

plugins/CoreHome/Columns/VisitorFingerprint.php
tsteur: setting `$metricId` to `visitors_fingerprint` might be a good idea @katebutler  👍 . Shouldn't really change much though. In general the code in that method likely will still be needed to ensure only the `unique fingerprint` metrics is being created.

diosmosis: Hi @katebutler, could you please respond to this thread (either by applying the requested change or commenting on why you'd rather not)? Once the thread is resolved I can merge this.

PR #14187, opened 2019-03-12
Rename piwik to matomo in a few INI settings for consistency

mattab:
Just updated the website FAQs to mention matomo.log instead of piwik.log so thought we should do the same in Matomo


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)*
 ### Breaking Changes
 
 * `Referrers.getKeywordsForPageUrl` and `Referrers.getKeywordsForPageTitle` APIs have been deprecated and will be removed in Matomo 4.0.0
+* By default, Matomo [application logs](https://matomo.org/faq/troubleshooting/faq_115/) will now be logged in `tmp/logs/matomo.log` instead of `tmp/logs/piwik.log`. This log file path can be edited in your config/config.ini.php in the INI setting `logger_file_path`.
```
tsteur: this will definitely need to be announced big in the changelog as it can cause lots of trouble around log rotation, logs not being forwarded to Cloudwatch etc.

tsteur:
Screenshot tests will need to be updated otherwise looks fine.

PR #14188, opened 2019-03-12
Update Twig because of security

### LINKED ISSUE #14189: Twig has a security issue in sandbox
Created https://github.com/matomo-org/matomo/pull/14188

This is about:
https://symfony.com/blog/twig-sandbox-information-disclosure

mikkeschiren:
https://symfony.com/blog/twig-sandbox-information-disclosure
Fixes #14189

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,10 +30,7 @@
         "symfony/console": "~2.6.0",
         "mustangostang/spyc": "~0.5.0",
         "piwik/device-detector": "~3.0",
-        "piwik/decompress": "~1.0",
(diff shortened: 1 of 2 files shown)


sgiehl:
@mikkeschiren  thanks for PR. The piwik/matomo packages haven't been changed on purpose. The new packages are already in the `Matomo` namespace. To not break backward-compatibility for plugins,  that might use those components, we won't change that before Matomo 4.

I will close this PR and create a new one updating the other dependencies as well

PR #14193, opened 2019-03-12
Use tabular-nums in tables

c960657:
Numeric data in tables look “jumpy” when using a font where digits have different widths. Using the tabular-nums font variant gets around this problem.

**Without tabular-nums** (with some gridlines added)
image

**With tabular-nums**
image

The above screenshots are made on MacOS.

This change does not to affect the screenshots made by Phantomjs on Travis, i.e. it seems that the the digits in the font used by Phantomjs all have equal width even without using tabular-nums.


diosmosis (approved):
Change looks good, maybe @tsteur / @mattab have thoughts on how it looks?

tsteur:
Looks good 👍

mattab:
Nice improvement, thanks for the PR! @c960657 :+1:

PR #14196, opened 2019-03-13
Do not match by config ID if trust visitor cookies is used and a user ID is supplied.

### LINKED ISSUE #14018: UserId ignored when trust_visitors_cookies = 1
Noticed when testing locally that when `[Tracker] trust_visitors_cookies = 1` the user ID param is ignored in favor of the config ID. Which can group some visits together that shouldn't be grouped together.

diosmosis:
Fixes #14018

### DIFF
--- a/core/Tracker/VisitorRecognizer.php
+++ b/core/Tracker/VisitorRecognizer.php
@@ -161,15 +161,15 @@ class VisitorRecognizer
     {
         $isForcedUserIdMustMatch = (false !== $request->getForcedUserId());
 
+        // This setting would be enabled for Intranet websites, to ensure that visitors using all the same computer config, same IP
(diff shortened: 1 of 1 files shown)


tsteur:
Makes sense to me 👍  maybe @mattab can have a look as well

mattab:
Logic looks good to me too :+1: 

Btw: tests seem to fail with a new visit with an empty IP address, not sure if related to this PR:
```
-    <visitIp>175.41.192.9</visitIp>
+    <visitIp>0.0.0.0</visitIp>
```

diosmosis:
Test passes for me locally, appears to be an issue w/ test order... will look into it for a little, but will probably be extremely hard to debug.

PR #14201, opened 2019-03-13
Display a warning if using an EOL version of PHP

### LINKED ISSUE #14174: When using an old PHP version (PHP < 7.1) display permanent message inviting users to upgrade PHP
In Matomo 4, we will require at least PHP 7+ (in https://github.com/matomo-org/matomo/issues/13611 )

And PHP 5.6 and PHP 7.0 are already out of the PHP support cycle and therefore all users should upgrade PHP to keep their Matomo data safe.

To prepare users for this, we want to display a permanent notification which will be visible and invite users to ugprade PHP (and warn them that Matomo will soon not work for them).

A similar code was done previously in https://github.com/matomo-org/matomo/issues/8193 https://github.com/matomo-org/matomo/pull/9144  https://github.com/matomo-org/matomo/issues/8156

katebutler:
Warn the user if their version of PHP is < 7.1

image

Fixes #14174

### DIFF
--- a/core/Plugin/ControllerAdmin.php
+++ b/core/Plugin/ControllerAdmin.php
@@ -207,7 +207,7 @@ abstract class ControllerAdmin extends Controller
      */
     private static function getNextRequiredMinimumPHP()
     {
-        return '5.5.9';
+        return '7.1';
     }
 
     private static function isUsingPhpVersionCompatibleWithNextPiwik()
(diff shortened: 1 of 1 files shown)


diosmosis (approved):
UI tests will need to be updated before merge

core/Plugin/ControllerAdmin.php
```diff
@@ -207,7 +207,7 @@ private static function notifyIfEAcceleratorIsUsed()
      */
     private static function getNextRequiredMinimumPHP()
     {
-        return '5.5.9';
+        return '7.1';
```
simivar: Wouldn't it be better to move this to `const` and use it in line 371 as well?

core/Plugin/ControllerAdmin.php
```diff
@@ -365,8 +366,8 @@ protected static function getPiwikVersion()
         return "Matomo " . Version::VERSION;
     }
 
-    private static function isPhpVersionAtLeast55()
+    private static function isPhpVersionAtLeast71()
```
simivar: What do you think about renaming that method to `isMinimalSupportedPhpVersion()` or something along this line? There will be no need to rename it in the future.

simivar (changes requested):

tsteur (approved):

PR #14203, opened 2019-03-14
Ignore config files placed in misc folder in fileintegrity check

tsteur:
Follow up from https://github.com/matomo-org/matomo/pull/13999 where we allowed to place config files in misc/user directory. These need to be ignored in the file check


tsteur:
Tested them on production and works

PR #14204, opened 2019-03-14
Show device type and model in realtime map tooltip

sgiehl:


plugins/UserCountryMap/javascripts/realtime-map.js
```diff
@@ -141,7 +141,7 @@
                     filter_limit: maxVisits,
                     showColumns: ['latitude', 'longitude', 'actions', 'lastActionTimestamp',
                         'visitLocalTime', 'city', 'country', 'countryCode', 'referrerType', 'referrerName',
```
diosmosis: Should we also escape referrer & city? I think those two can also be specified through tracker query params.

sgiehl: doesn't hurt to do so, even if city can only be set when providing a token_auth

sgiehl: changed it.

Issue #14207, opened 2019-03-15
Move `action_name` query parameter off first position so requests do not get ad-blocked


kdekooter:
Despite being on the exact same domain as the website uBlock still blocks calls to matomo. It seems to block based on the phrase `.php?action_name=`

![image](https://user-images.githubusercontent.com/508983/54436922-cf664100-4733-11e9-9aa8-bc691a2b760f.png)

Findus23:
Correct, the EasyPrivacy list used by ublock and others is blocking the .php?action_name string.

So there is no bug in Matomo and everything is working as intended :slightly_smiling:

kdekooter:
Agreed that is nog a bug. But is being blocked by an adblocker intended?

kdekooter:
Maybe moving it to the second position in the query string would suffice.

tsteur:
I think there's not too much point cause ad blockers will eventually just find a different way to detect it. It would maybe work for a few weeks and then fail again.

fdellwing:
> It would maybe work for a few weeks and then fail again.

EasyPrivacy list still does not block matomo(js|php), so there is hope.

kdekooter:
Yes but it is quite sensitive to `.php` extensions.

I ended up working around this by letting the client send tracking calls to `/mtm` instead of `/matomo.php` and rewrite this back on the server side with nginx.

mattab:
Let's leave the ticket opened as it is a valid request (at least for some of our users, not all would agree) :ok_hand:

tsteur:
btw you can also just POST instead of using GET requests and the action_name will be gone. Something like `_paq.push(['setRequestMethod', 'POST']);`.

You won't be able to use log analytics to replay tracking requests with that, but most people don't do this anyway. I would say this issue is pretty much resolved and any further solutions would probably need to be developed in some plugin as we likely won't change it anyway.

kdekooter:
Thanks @tsteur I tried that but it turns out POST requests to this URL are blocked (by Brave) as well.

Shirkit:
If there's one thing I've learned is that blockers will block you. Their goal is to block analytics and tracking on client-side, so if you try to trick them, they'll just find a workaround.

The people who put out the lists do no want to be tracked online. It all started with annoying ads, then when remarketing showed up, now all analytics suffer as well.

tsteur:
[...]

Issue #7364, opened 2015-03-04
Piwik block by AdBlock even after renmaing piwik.php piwik.js


sambuev:
Hello, I have Piwik installed on my hosting with secret address: interes.site.com/sm/
and I have renamed piwik.php piwik.js to sm.php sm.js and I though it cannot by caught by AdBlock but unfortunately AdBlock on my Safari browser easily blocking Piwik and many other users with their Adblock (on Chrome, Firefox) out of my view :( 

![screen shot 2015-03-04 at 18 21 38](https://cloud.githubusercontent.com/assets/6326685/6495985/b28d4c06-c29c-11e4-924b-53789685e875.png)

What can I do to avoid blocking Piwik? 
Is it possible to avoid it?

MESWEB:
I got this same error but I think this is difficulty for dev piwik

kylekatarnls:
I use rewriting and replace /piwik.php with juste /p. Works like a charm.

MESWEB:
Can You give any solution? I trying this but without success - https://www.codelibrary.me/tag/ad-block/

kylekatarnls:
This is what I did (but with a proxy-pass in the same domain).

First remove "analytics" in the URL when you enter the rewrite rule and any word that can be seen as ads.

Then if it's not enough, use a proxypass (http://httpd.apache.org/docs/2.2/fr/mod/mod_proxy.html) on your own domain, so your URL to call Piwik will just look like "/p" and "/j"

```
(function() {
    _paq.push(['setTrackerUrl', '/p']);
    _paq.push(['setSiteId', 1]);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.type='text/javascript'; g.async=true; g.defer=true; g.src='/j'; s.parentNode.insertBefore(g,s);
  })();
```

Joey3000:
Guys, have you tried just replacing "piwik.php" and "piwik.js" with "js/", as described in https://github.com/piwik/piwik/blob/master/js/README.md? No .htaccess rewriting, file renaming, etc. needed.

See also:
http://forum.piwik.org/read.php?2,83940,125929
http://forum.piwik.org/read.php?4,14971

I've been using that for years. Just make sure that the "piwik" string doesn't appear anywhere in the URL (i.e. domain name, directories). After coming across this issue report, I've just tried it again with ABE on Firefox and it doesn't block it. I checked ABE lists again, and it only checks for the "piwik" string anywhere in the URL.

[...]

PR #14211, opened 2019-03-16
[POC] Obfuscate tracking parameters

sgiehl:
Even when renaming piwik.php (or matomo.php) some adblockers still seem to block some tracking requests based on the url parameters sent with the request.

This implementation will make it possible to obfuscate all request parameters by doing a base64 on the request string.

Not sure if it might make sense to even enable that by default if it works as expected.

refs #7364, #14207


tsteur:
wouldn't you just send it as POST maybe if someone really wanted to avoid it? Of course you can't replay maybe but would probably work otherwise. Not sure if it has high prio right now

sgiehl:
Correct, POST would be another way to avoid a parameter blocking, but as you mentioned it wouldn't work if replaying logs is required at some point. 

Actually this was just a simple idea that came up to my mind when reading some issues and forum posts. Thought it might be worth to create a simple draft for it so we can consider using it at some point if usefull...

fdellwing:
I kind of like this idea. Just to be sure, it will not be `.php?obfuscate=abc` but `.php?abc`?

kdekooter:
Try to get rid of the `.php` extension as well. It is a red flag for blockers.

tsteur:
@sgiehl I think we can maybe close the issue for now as we don't really want to work around ad blockers / privacy lists in such a way at this stage.

negreanucalin:
Maybe it would be a good idea to take this POC into account and maybe define some requirements, I have made the [the mentioned post above](https://github.com/matomo-org/matomo/issues/14938) with a temporary workaround.

Issue #14208, opened 2019-03-15
Changed period in row evolution breaks graph/metric association


peterbo:
Steps to reproduce: 
- Navigate to Visitors->Devices (maybe other reports are affected too) 
- open row evolution of a row in e.g. device type
- change period from (e.g.) week to day
- click through the metrics (association is broken)

Bug is reproducible on the demo.

![row-evolution-graph-selection](https://user-images.githubusercontent.com/1018973/54446345-ff6b0f80-4746-11e9-9ef9-71b5539958ad.JPG)

tsteur:
FYI: Tried to fix it but couldn't fix it within 4 hours. Pushed the changes here but there are a few challenges see comment: https://github.com/matomo-org/matomo/commit/7253a4fdbf2eda90130ba47e119117b07ac8679a

I reckon to fix this, we would actually need to reload the popover with some additional parameters eg which period to select, which metrics to preselect. I reckon fixing this will take another 8 hours as you need to put a lot of "hacks" into various visualisations to make this work. 

A problem is really that it not only remembers which metrics are selected, but it even shows metrics below the graph that aren't available for that period etc.

PR #14212, opened 2019-03-16
Ensure first sparkline is highlighted after switching periods in row evolution overlay

sgiehl:
When switching the periods in the evolution chart in the row evolution overlay, the selected metrics are unset. If another metric than the first one was selected at this point, no metric is highlighted in the sparkline list below afterwards.

Note: This is just a quick fix for the current behavior. Imho it would be a lot better to actually restore the selected metrics when switching the period

refs #14208

### DIFF
--- a/plugins/CoreVisualizations/javascripts/jqplot.js
+++ b/plugins/CoreVisualizations/javascripts/jqplot.js
@@ -761,6 +761,7 @@ RowEvolutionSeriesToggle.prototype.attachEvents = function () {
             // show the active series
             // if initiallyShowAll, all are active; otherwise only the first one
             self.activated.push(true);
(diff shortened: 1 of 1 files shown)


diosmosis:
As far as I can tell, this doesn't actually fix the bug, right? If I click on Unique Visitors, for example, it will still show the actions series:

image

Not sure what this PR is supposed to fix?

sgiehl:
that PR only fixes a highlight issue, When you choose to display not the first metric and switch the periods all metric sparklines will be not highlighted afterwards. With this PR the first sparkline should be highlighted again

diosmosis (approved):

diosmosis:
I see, looks good.

PR #14213, opened 2019-03-17
Strip HTML tags in API error messages

### LINKED ISSUE #12300: html tags are displayed in permission error message
related to  #10369

I tried to update a plugin but forgot to change the owner of the folder back to `ẁww-data` before, so Piwik couldn't create a file.
But the error message isn't displayed correctly:
grafik

```html
<div id="notificationContainer">
    <div style="display:none" data-role="notification" data-id="CorePluginsAdmin_InstallPlugin" data-type="persistent"
         data-title="" data-noclear="0" data-context="error" data-raw="">Error while creating/copying file to &lt;code&gt;/var/www/piwik/plugins//ActivityLog/Widgets/GetActivityLog.php&lt;/code&gt;.
        &lt;br /&gt;Please check that the web server has enough permission to write to these files/directories:&lt;br /&gt;For
        example, on a GNU/Linux server if your Apache httpd user is www-data, you can try to execute:&lt;br /&gt;
        &lt;code&gt;chown -R www-data:www-data /var/www/piwik&lt;/code&gt;&lt;br /&gt;&lt;code&gt;chmod -R 0755
        /var/www/piwik&lt;/code&gt;&lt;br /&gt;
    </div>
</div>
```

sgiehl:
fixes #12300

### DIFF
--- a/core/API/ResponseBuilder.php
+++ b/core/API/ResponseBuilder.php
@@ -168,6 +168,10 @@ class ResponseBuilder
     {
         $message = ExceptionToTextProcessor::getWholeBacktrace($exception, $this->shouldPrintBacktrace);
 
+        if ($exception instanceof \Piwik\Exception\Exception && $exception->isHtmlMessage()) {
+            $message = strip_tags(str_replace('<br />', PHP_EOL, $message));
+        }
+
         return Renderer::formatValueXml($message);
     }
 
--- a/core/API/ResponseBuilder.php
+++ b/core/API/ResponseBuilder.php
@@ -168,7 +168,7 @@ class ResponseBuilder
     {
         $message = ExceptionToTextProcessor::getWholeBacktrace($exception, $this->shouldPrintBacktrace);
 
-        if ($exception instanceof \Piwik\Exception\Exception && $exception->isHtmlMessage()) {
+        if ($exception instanceof \Piwik\Exception\Exception && $exception->isHtmlMessage() && Request::isRootRequestApiRequest()) {
             $message = strip_tags(str_replace('<br />', PHP_EOL, $message));
         }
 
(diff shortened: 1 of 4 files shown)


diosmosis:
Tested locally and this would decode messages for exceptions that also don't call `setIsHtmlMessage()`, which could introduce a possible vulnerability. Can we add an `is_html_message` property when the API returns exceptions? Not sure if it's the best idea to add to the API, but it's all I can think of.

CC @tsteur since it would be a change to the API

tsteur:
I don't think it's a good idea to decode all errors as it could definitely lead to various security issues. If there was such a `is_html_message` property, could we safely assume that it will work be safe to use these methods?

Took me a while to think about it... I would say in general, an API shouldn't return an HTML message as you don't know if the consumer can process HTML. Often it would not be possible to show HTML, eg in a mobile app etc. I would say we strip HTML tags server side for these messages, and / or have two properties: `message: 'message with stripped html tags', htmlMessage: 'formatted html message'`. The `htmlMessage` property could be eg only visible when `setIsHtmlMessage` is set (not sure if it is safe to put these htmlMessages into the DOM, would need to double check the `safeDecode` etc is all fine).

I think in the mentioned case in https://github.com/matomo-org/matomo/issues/12300 we could maybe just keep things simple and strip the tags and show it without the html tags?

mattab:
> I think in the mentioned case in #12300 we could maybe just keep things simple and strip the tags and show it without the html tags?

Sounds good to me :+1:

sgiehl:
stripping the tags should work as well. I'll adjust the PR

diosmosis (approved):
There are some test failures, but looks good

tsteur:
Random thought: Should this maybe only happen when the ROOT request is an API request? This way a lot of code can maybe still format a message correctly with HTML?

diosmosis:
Haven't tested (of course would be good to), this code might be skipped if the original format is used.

sgiehl:
I've added the check for root api requests...

tsteur:
looks good

PR #14216, opened 2019-03-17
Make sure to anonymize token only when needed

tsteur:
I was using the event `TwoFactorAuth.requiresTwoFactorAuthentication` to disable two factor authentication but then was wondering why all API calls fail. Then realised we are only looking at the event `onRequestDispatch` but not `onRequestDispatchEnd`.

### DIFF
--- a/plugins/TwoFactorAuth/TwoFactorAuth.php
+++ b/plugins/TwoFactorAuth/TwoFactorAuth.php
@@ -144,35 +144,10 @@ class TwoFactorAuth extends \Piwik\Plugin
         }
 
         if ($module === 'Proxy') {
-            return;
+            return false;
         }
 
-        if ($module === 'TwoFactorAuth' && $action === 'showQrCode') {
-            return;
-        }
-
-        if ($module === 'CoreUpdater') {
-            return;
-        }
-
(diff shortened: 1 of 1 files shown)


diosmosis (approved):
Looks good, if possible might be a good idea to cover this in an integration test.

tsteur:
@diosmosis tried for half hour to write a test but running out of available time. Might just not write one as it's indirectly covered by some UI tests (not the event though).
My problem is that  the token is always empty. No matter if I use `Access` or `FakeAccess` when calling `Piwik::getCurrentUserTokenAuth()`:
image

diosmosis:
@tsteur if it's an integration test, probably because there's no user created, can add one via:

```
    protected static function configureFixture($fixture)
    {
        parent::configureFixture($fixture);
        $fixture->createSuperUser     = true;
    }
```

then maybe call `Fixture::loginAsSuperUser()` in teh test instead of using FakeAccess. If that doesn't work, then it would probably take a while...

tsteur:
image

@diosmosis I'm giving up and will just merge for now. I've made it work by making it a migration test instead and not using `FakeAccess`. However above code will prevent the test from working easily. It would easily take 1-2 hours more but I don't think it's worth it right now. Will merge.

PR #14228, opened 2019-03-19
Adds missing / in directoy check

### LINKED ISSUE #14226: open_basedir restriction AND Support multiple plugin paths
I have an error since the last update (3.9.0)

> WARNING: /**INSTALL_DIR**/core/Plugin/Manager.php(451): Warning - is_dir(): open_basedir restriction in effect. File(/**INSTALL_DIR**plugins/Annotations) is not within the allowed path(s): (/**INSTALL_DIR**/) - Matomo 3.9.0 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)

Because there's a mistake in the path : **/INSTALL_DIRplugins/Annotations** (it's missing one /)

_$corePluginsDir = PIWIK_INCLUDE_PATH . 'plugins/' . $pluginName;_
should perhaps be replaced by:
_$corePluginsDir = PIWIK_INCLUDE_PATH . '/plugins/' . $pluginName;_

Thank you for your work / Merci beaucoup

Anthony

sgiehl:
fixes #14226

### DIFF
--- a/core/Plugin/Manager.php
+++ b/core/Plugin/Manager.php
@@ -447,7 +447,7 @@ class Manager
             return self::$pluginsToPathCache[$pluginName];
         }
 
-        $corePluginsDir = PIWIK_INCLUDE_PATH . 'plugins/' . $pluginName;
+        $corePluginsDir = PIWIK_INCLUDE_PATH . '/plugins/' . $pluginName;
         if (is_dir($corePluginsDir)) {
             // for faster performance
             self::$pluginsToPathCache[$pluginName] = self::getPluginRealPath($corePluginsDir);


Findus23:
Any reason why the PHP error/warning hasn't occurred in the tests?

sgiehl:
if no open_basedir restriction is active the `is_dir` simply returns `false`

core/Plugin/Manager.php
```diff
@@ -447,7 +447,7 @@ public static function getPluginDirectory($pluginName)
             return self::$pluginsToPathCache[$pluginName];
         }
 
-        $corePluginsDir = PIWIK_INCLUDE_PATH . 'plugins/' . $pluginName;
+        $corePluginsDir = PIWIK_INCLUDE_PATH . '/plugins/' . $pluginName;
```
tsteur: maybe we to a `rtrim(PIWIK_INCLUDE_PATH, '/')` just in case? I think tests add for example a slash at the end to the constant but not sure.

sgiehl: actually there is currently only one place, where we do a trim on `PIWIK_INCLUDE_PATH`:

https://github.com/matomo-org/matomo/blob/6395855aa7f3813cc7f413a18d31505ea26ba32a/core/Plugin/Manager.php#L477

But there are so many more uses so, I don't think it's actually required.

tsteur: :+1: looks good to merge

mickaelmonsieur:
and does not need to publish an immediate fix or remove the current version?

PR #14236, opened 2019-03-19
Delete cache after one click update to avoid un-updated in-memory classes being used.

### LINKED ISSUE #14227: One-click upgrade sometimes fails with errors such as "Call to undefined method Piwik\Plugin\Manager::getPluginDirectory() " (fails only once, works after a refresh)
Updating to 3.9.0 using the one click updater failed to me with the following message:

`Matomo encountered an error: Call to undefined method Piwik\Plugin\Manager::getPluginDirectory() (which lead to: Circular dependency detected while trying to resolve entry ‘Piwik\Twig’)`

At least one other users had this issue and reported it in the German forum: https://forum.matomo.org/t/update-fehler-matomo-3-9-0/32109

After that broken update, I'm now left with the issue described in #14226. Maybe there is a connection between the two, as both errors/warnings are about the plugin directory.

diosmosis:
Fixes #14227

### DIFF
--- a/plugins/CoreUpdater/Controller.php
+++ b/plugins/CoreUpdater/Controller.php
@@ -167,8 +167,6 @@ class Controller extends \Piwik\Plugin\Controller
             $messages = $e->getUpdateLogMessages();
         }
 
-        Filesystem::deleteAllCacheOnUpdate();
-
         $view->feedbackMessages = $messages;
         $this->addCustomLogoInfo($view);
         return $view->render();
@@ -176,6 +174,8 @@ class Controller extends \Piwik\Plugin\Controller
 
     public function oneClickResults()
     {
+        Filesystem::deleteAllCacheOnUpdate();
+
         $httpsFail = (bool) Common::getRequestVar('httpsFail', 0, 'int', $_POST);
         $error = Common::getRequestVar('error', '', 'string', $_POST);
 


diosmosis:
CC @tsteur

plugins/CoreUpdater/Controller.php
```diff
@@ -167,15 +167,15 @@ public function oneClickUpdate()
...
-        Filesystem::deleteAllCacheOnUpdate();
-
         $view->feedbackMessages = $messages;
         $this->addCustomLogoInfo($view);
         return $view->render();
     }
 
     public function oneClickResults()
     {
+        Filesystem::deleteAllCacheOnUpdate();
```
tsteur: @diosmosis I'm not too much into the code... should we maybe just update on both after one click update and one click results?

diosmosis: We can't do it in the one click update for the same reason we have this special view: https://github.com/matomo-org/matomo/blob/3.x-dev/core/View/OneClickDone.php#L15-L22

This method triggers a lot of code that could be using some out of date classes (in this case Twig was new, but Plugin\Manager was old).

tsteur: OK just to understand correctly: Was it an opcache that prevents this issue? If not using any PHP cache, this might still happen?

diosmosis: It did for me locally (I have opcache disabled). I believe the issue is as follows:

1) request begins, plugin manager loads (old Plugin\Manager class is in memory)
2) one click update finishes changes Twig.php file (Twig class is not loaded in memory yet)
3) Filesystem::deleteAllCacheOnUpdate() is called, we try to get Twig class (new Twig class is in memory)
4) new twig class tries to call new Plugin\Manager method, but old class is in memory so it doesn't work

tsteur: @diosmosis I'm bit confused now. So does it actually help to not invalidate cache anymore in `Filesystem::deleteAllCacheOnUpdate();` as it would still load the outdated Twig when not using opcache?

I'm pretty sure we had such issues a few times before. I wonder if we should clear the cache in `oneClickUpdate `, and then redirect to a new page where we render the view with the messages (would need to store these in session or so). This would maybe fix it once and for all?

diosmosis: The View isn't the issue, it doesn't use twig. Filesystem::deleteAllCacheOnUpdate() uses twig, and many other classes.

diosmosis: Perhaps there needs to be an intermediate method that deletes the cache and redirects again. Not sure.

diosmosis: Or maybe we can invoke a command? That way everything is done in a new process.

tsteur: You mean invoke command to clear the cache? I was thinking in general once the files have been updated, it is not really safe to use any other class afterwards that wasn't used before. As there can be always issues. That's why I thought redirecting may help and avoid any future issues. In theory there could be even issues with twig templates somehow not being compatible with controller and so on (probably not best example as likely not an issue).

diosmosis: > You mean invoke command to clear the cache?

Yes, via `shell_exec()`.

> I was thinking in general once the files have been updated, it is not really safe to use any other class afterwards that wasn't used before. As there can be always issues.

This is why OneClickDone is used instead of View/Twig.

> That's why I thought redirecting may help and avoid any future issues.

OneClickDone does a redirect, the problem occurs before this when clearing the cache.

tsteur: I see... had no idea this existed. should be fine then 👍  Wouldn't need to do a shell_exec as it might not work on all systems.

PR #14240, opened 2019-03-19
Do not send password changed email for automated use cases.

diosmosis:
Fixes https://github.com/matomo-org/plugin-LoginLdap/issues/180

Updated since previous solution wouldn't work.

### DIFF
--- a/plugins/UsersManager/API.php
+++ b/plugins/UsersManager/API.php
@@ -922,7 +922,9 @@ class API extends \Piwik\Plugin\API
             $this->sendEmailChangedEmail($userInfo, $email);
         }
 
-        if ($passwordHasBeenUpdated) {
+        if ($passwordHasBeenUpdated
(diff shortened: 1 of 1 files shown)


plugins/UsersManager/API.php
```diff
@@ -922,7 +922,9 @@ public function updateUser($userLogin, $password = false, $email = false, $alias
             $this->sendEmailChangedEmail($userInfo, $email);
         }
 
-        if ($passwordHasBeenUpdated) {
+        if ($passwordHasBeenUpdated
+            && $requirePasswordConfirmation
```
tsteur: not tested but looks good 👍

jgkirschbaum: Patch tested and no e-mail is sent after login with a ldap account.

sgiehl: @jgkirschbaum  thx for confirmation

PR #14253, opened 2019-03-22
Keep top level nb_visits_converted metric

### LINKED ISSUE #10109: On email reports the conversion rate is zero always
Hi
after receiving daily reports , on "all referrer " part , the last column is always zero 

3845fb3bb1b1dc333ff83500ba6579557fb31830_1_690x183

diosmosis:
Currently we delete this metric "because it could differ from the sum of each goal conversions", but it should, since nb_visits_converted is not the sum of each goal's conversions. Conversions can happen multiple times per visit, so this would not be accurate.

Also, it is not the sum of each goals' nb_visits_converted, since visits can share conversions for different goals. Keeping it puts the correct value of conversion_rate in scheduled reports.

Fixes #10109


plugins/CoreHome/Columns/Metrics/ConversionRate.php
```diff
@@ -43,8 +44,20 @@ public function format($value, Formatter $formatter)
...
     {
-        $nbVisitsConverted = $this->getMetric($row, 'nb_visits_converted');
         $nbVisits = $this->getMetric($row, 'nb_visits');
+        $nbVisitsConverted = $this->getMetric($row, 'nb_visits_converted');
+        if ($nbVisitsConverted === false) {
+            $goals = $this->getMetric($row, 'goals');
+
+            $nbVisitsConverted = 0;
+            foreach ($goals as $idGoal => $values) {
+                if ($idGoal === 0 || $idGoal === '0' || $idGoal == 'idgoal=0') {
```
tsteur: would this also need to be done for `goals = -1`?
@mattab maybe you can have a look at the PR some time I actually don't quite understand if the behaviour is expected.

diosmosis: yeah, that makes sense don't want to count "conversions" for abandoned carts here

plugins/CoreHome/Columns/Metrics/ConversionRate.php
```diff
@@ -43,8 +44,28 @@ public function format($value, Formatter $formatter)
...
         $nbVisits = $this->getMetric($row, 'nb_visits');
+        $nbVisitsConverted = $this->getMetric($row, 'nb_visits_converted');
+        if ($nbVisitsConverted === false) {
+            $goals = $this->getMetric($row, 'goals');
+
+            if (!empty($goals)) {
+                $nbVisitsConverted = 0;
+                foreach ($goals as $idGoal => $values) {
+                    if ($idGoal === 0
+                        || $idGoal === '0'
```
mattab: is there maybe a CONST we could reference instead of hard coding the value?

diosmosis: Can do

mattab:
Feedback
* in general i'm not sure if this is accurate. "Nb visits converted" should be number of visits with at least one converted goal. but this logic takes the "max" value of "nb visits converted" for any given goal. The "Real" accurate value of "Nb visits converted" will be equal or higher to the value processed here. For example we may have 5 visits converted to goal X and 10 visits converted to goal Y, this PR will say that 10 visits are converted, but it could be that anywhere between 10 and 15 visits are converted. So it wouldn't show accurate numbers. Is there maybe another way to have this accurate nb_visits_converted metric?
* A sub-module change is included in the PR which I suppose is not expected

diosmosis:
Moving to 3.11 since it's not a straightforward fix

diosmosis:
@mattab removed the code you were talking about (that was for showing the conversion rate for old reports, but of course that won't work).

From testing manually and looking at automated tests, the nb_visits_converted metric that's not in the `goals` array looks accurate. It's computed by `queryVisitsByDimension()` which aggregates over the log_visit table, so I'm not sure why it would be inaccurate. Don't think there's an issue in keeping it.

mattab:
LGTM 
(tests need updated before merge)

tsteur:
Some tests are still failing @diosmosis . Feel free to merge once the build is green

PR #14257, opened 2019-03-22
Only print warning if locking failed, fixes #14167

### LINKED ISSUE #14167: Scheduler fails with "LOCK TABLES privilege not granted"
Matomo version: 3.8.1
DB: MariaDB 10.1.38
When the cron job runs, I see this error message in the log:

> INFO [2019-03-07 00:50:30] 111882  Scheduler: executing task Piwik\Plugins\PrivacyManager\Tasks.deleteLogData...
> WARNING [2019-03-07 00:50:45] 111882  Piwik\Plugins\PrivacyManager\LogDataPurger: LOCK TABLES privilege not granted; skipping unused actions purge

But the user has all rights for the database:

> show grants for 'piwik'@'localhost';
> GRANT ALL PRIVILEGES ON `piwik`.* TO 'piwik'@'localhost'

Flakebi:
Hi,
I think I found the reason for #14167 (the PrivacyManager\LogDataPurger prints a warning which leads to a non-zero exit code of the console process).
I didn’t get any failure in the last 6 days (normally it happens every 5 days) so it seems to work.

### DIFF
--- a/plugins/PrivacyManager/LogDataPurger.php
+++ b/plugins/PrivacyManager/LogDataPurger.php
@@ -71,11 +71,13 @@ class LogDataPurger
         $logTables = self::getDeleteTableLogTables();
 
         // delete unused actions from the log_action table (but only if we can lock tables)
-        if ($deleteUnusedLogActions && Db::isLockPrivilegeGranted()) {
-            $this->rawLogDao->deleteUnusedLogActions();
-        } else {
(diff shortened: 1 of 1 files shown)


fdellwing:
Lol, makes sense to me.

tsteur:
:+1: good find

Issue #14258, opened 2019-03-23
Scheduler GeoIP2AutoUpdate gives warning


sju10544:
Hi all,

after having a look at the "piwik-archive.log" because the Maxminddb is not updating successfull I got this:

GeoIP2AutoUpdater: successfully downloaded 'https://geolite.maxmind.com/download/geoip/database/GeoLite2-City.tar.gz'
WARNING [2019-03-23 20:15:34] 23435  xxx/htdocs/piwik/plugins/Monolog/Processor/SprintfProcessor.php(24): Warning - vsprintf(): Too few arguments - Matomo 3.9.1 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)

The downloaded file is "GeoIP2-City.mmdb.tar.gz" and has only 6KB...
The URL for downloading the file is: https://geolite.maxmind.com/download/geoip/database/GeoLite2-City.tar.gz

Any ideas?

sgiehl:
> WARNING [2019-03-23 20:15:34] 23435 xxx/htdocs/piwik/plugins/Monolog/Processor/SprintfProcessor.php(24): Warning - vsprintf(): Too few arguments - Matomo 3.9.1 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)

Found the reason for this one. Will propose a PR for it.

sju10544:
Hmmm... sorry but after manually made the changes in the file and run
./console core:run-scheduled-tasks --force
I got the same error messages and behaviour.

mattab:
@sgiehl could you check @sju10544 comment, maybe the issue is not fixed?

SystemZ:
This problem is present in `3.12.0-b4`

```
INFO [2019-10-05 15:41:14] 14  ---------------------------
INFO [2019-10-05 15:41:14] 14  SCHEDULED TASKS
INFO [2019-10-05 15:41:14] 14  Starting Scheduled tasks... 
INFO [2019-10-05 15:41:14] 14  Scheduler: executing task Piwik\Plugins\GeoIp2\GeoIP2AutoUpdater.update...
INFO [2019-10-05 15:41:14] 14  GeoIP2AutoUpdater: successfully downloaded 'https://geolite.maxmind.com/download/geoip/database/GeoLite2-City.tar.gz'
[...]

PR #14259, opened 2019-03-24
Fix number of sprintf placeholders in GeoIP2AutoUpdater

sgiehl:
refs #14258

### DIFF
--- a/plugins/GeoIp2/GeoIP2AutoUpdater.php
+++ b/plugins/GeoIp2/GeoIP2AutoUpdater.php
@@ -52,14 +52,6 @@ class GeoIP2AutoUpdater extends Task
     );
 
     /**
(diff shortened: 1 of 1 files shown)


plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -527,7 +519,7 @@ protected function performRedundantDbChecks($logErrors = true)
             } catch (\Exception $e) {
                 if($logErrors) {
                     Log::error("GeoIP2AutoUpdater: Encountered exception when performing redundant tests on GeoIP2 "
-                        . "%s database: %s: %s", $type, $e->getMessage());
+                        . "%s database: %s", $type, $e->getMessage());
```
tsteur: makes sense 👍

PR #14265, opened 2019-03-24
misc/ cleanup

Findus23:
This is a WIP to delete old files in misc/ that don't work or are not needed anymore.
It also updates the GPL file to the latest version from https://www.gnu.org/licenses/gpl.txt

If someone needs those files or knows where they are still referenced or knows more files that can be deleted, comment here.

It would also maybe be useful to write an update script that deletes these files so that everyone doesn't need to delete them manually.


tsteur (approved):

tsteur:
Looks good to me. Not sure if @mattab has any thoughts. But better to have those docs rather in our guides etc.

> It would also maybe be useful to write an update script that deletes these files so that everyone doesn't need to delete them manually.

They should be automatically deleted or not? See `plugins/CoreUpdater/Model.php::removeGoneFiles`

mattab:
LGTM :+1:

Findus23:
@tsteur I just updated my instance to the beta2 and the files are still there and integrity checker is complaining about them as I expected.

Is it possible that this isn't working as it should?

tsteur:
Seems we don't apply this to the misc directory. Probably because there may be many user directories/ user files such as brand logos, config files, tracker files, ... https://github.com/matomo-org/matomo/blob/3.10.0-b3/plugins/CoreUpdater/Model.php#L28-L30

PR #14273, opened 2019-03-26
Overwrite direct entry referrer information if campaign referrer is found in later request.

diosmosis:
Noticed in one user's logs, the campaign information came in the second request. @mattab suggested that direct entry referrers should be overwritten by campaign information if supplied in a later request.


plugins/Referrers/Columns/Keyword.php
```diff
@@ -36,6 +36,18 @@ public function onNewVisit(Request $request, Visitor $visitor, $action)
         return $information['referer_keyword'];
     }
 
+    public function onExistingVisit(Request $request, Visitor $visitor, $action)
+    {
+        $information = $this->getReferrerInformationFromRequest($request, $visitor);
+        if ($this->isCurrentReferrerDirectEntry($visitor)
+            && $information['referer_type'] == Common::REFERRER_TYPE_CAMPAIGN
```
mattab: As well as Campaigns, expecting also Search engine / Website / Social to overwrite a "Direct entry" referrer (not just campaigns)

mattab (changes requested):

plugins/Referrers/Columns/Keyword.php
diosmosis: updated

mattab:
Feedback:

* When the Matomo admin didn't specify the website URLs alias for the Website (in Matomo), then Matomo will think that a visitor viewing Example.com/Page2 coming from Example.com/Page1 is coming from a Website Referrer Example.com/Page1. So it will mistakenly overwrite direct entry with the current website URL as referrer. So wondering if in: https://github.com/matomo-org/matomo/blob/9eb3be3378b0cf9ef8aadc2ed206c11188e446db/plugins/Referrers/Columns/Base.php#L108-L117
we would want to also look at the current Page URL and "Add to the Site URLs alias" the current page url hostname?

diosmosis:
@mattab Is `example.com` the main URL for the site? Or a different one?

mattab:
example.com would be the main URL (or even alias URL) but which wasn't specified in Website URLs field.

diosmosis:
@mattab I see there's no required main URL field anymore. So you mean to add `hostname(url)` to the list of website URLs, makes sense.

mattab:
Looks good to me (quick look). Looking forward to the tests  :+1:

diosmosis:
Added integration tests, system tests will need to be fixed, ready for a review.

plugins/Referrers/Columns/Campaign.php
```diff
@@ -50,6 +50,7 @@ public function shouldForceNewVisit(Request $request, Visitor $visitor, Action $
 
         if ($information['referer_type'] == Common::REFERRER_TYPE_CAMPAIGN
             && $this->isReferrerInformationNew($visitor, $information)
+            && !$this->isCurrentReferrerDirectEntry($visitor)
```
mattab: since it's quite tricky to undersatnd, could you add a comment above this line to explain that eg. "when a visit has referrer = direct entry, and that the current request's referrer = campaign, although you may expect it to create a new visit, actually the referrer will be updated in the current visit (rather than a new visit created.) This is to prevent issue where internal links with URL campaign parameters cause too many visits to be created. " or so

mattab (changes requested):
* be great to also update any other system tests impacted by this and check they are not regression

plugins/Referrers/Columns/Base.php
```diff
@@ -111,6 +111,10 @@ protected function getReferrerInformation($referrerUrl, $currentUrl, $idSite, Re
             $urlsByHost = $this->getCachedUrlsByHostAndIdSite();
 
             $directEntry = new SiteUrls();
+            if ($this->isSiteHasNoUrls($idSite)) {
```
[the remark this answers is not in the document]
diosmosis: Thought it might be better to avoid any possible attacks, since a user could specify any domain they want and maybe mess w/ the referrer URL reports. But after thinking about it I don't think it's possible.

plugins/Referrers/Columns/Campaign.php
diosmosis: done

Issue #14267, opened 2019-03-25
Do not send password changed email when triggered by API


futureweb:
When Password change is triggered by API "UsersManager.updateUser" Method (confirm PW = ADMIN PW) no "password changed" E-Mail should be sent.

Reason: We for our case got Matomo integrated into our SAAS CMS with automatic PW changes over the API. Since 3.9.1 thousands of our Customers getting double Mails ... one from our CMS and one from Matomo - resulting in countless Support Calls / Mails.

Related to: https://github.com/matomo-org/matomo/pull/14240 (Do not send password changed email for automated use cases)

Forum Reference: https://forum.matomo.org/t/3-9-1-password-changed-e-mail-on-automated-password-changes-by-api/32262

thx
Andreas Schnederle-Wagner

tsteur:
Makes sens 👍

tsteur:
Could be maybe an API parameter optionally that allows to send email but maybe not needed.

tsteur:
Actually, best would be to add an API parameter for whether to send an email or not and our UI that issues the API would force sending the email.

Findus23:
Makes sense, but this also raises the new issue that this kind of circumvents the original purpose (or at least one of them) of the feature (detect when an attacker tries to change your password).

If someone can simply change the password without sending the E-Mail then the E-Mail doesn't really carry that much information anymore.

Of course one could make this parameter only available to admin users but still then admin accounts wouldn't really be protected. 

Honestly I am not sure how other websites/webapps are implementing this.

tsteur:
I think it's fine to have the option to disable it through the API. Alternatively we could add a new config setting but this is hard for users that can't change the config option (eg on cloud). @mattab  any thoughts?

mattab:
* +1 for a new config file setting to disable the email notifications. (Config file settings can't be changed through the UI so an attacker couldn't disable it and it should keep the feature secure.)
* we can document the setting in a new FAQ "How do I disable the email notifications when password is changed and email address is changed?"

PR #14281, opened 2019-03-28
Add a config to disable e-mail notifications on password/email change through API

simivar:
Adds a new config option `api_update_users_email_notifications` that when set to 0 disables all emails from `UserManager::updateUser` API endpoint. #14267

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -453,6 +453,9 @@ noreply_email_name = ""
 ; set to 0 to disable sending of all emails. useful for testing.
 emails_enabled = 1
 
+; set to 0 to disable sending of emails in API UserManager::updateUser endpoint
+api_update_users_email_notifications = 1
+
 ; feedback email address;
 ; when testing, use your own email address or "nobody"
 feedback_email_address = "feedback@matomo.org"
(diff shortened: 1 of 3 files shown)


diosmosis:
@simivar can you add a test for this behavior in https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/UsersManager/tests/Integration/APITest.php ?

simivar:
@diosmosis done in 7a5ece1

simivar:
@diosmosis I was thinking... wouldn't it be better to move in .ini file `api_update_users_email_notifications` from `General` settings to it's own `[api]` and name it `update_users_email_notifications`?

diosmosis:
I don't think where the config is located in the INI file is all that important, as long as it's not outright inaccurate. Maybe @tsteur or @mattab has another opinion

tsteur:
@simivar the issue is already assigned to @katebutler and I think some work had already been done there. @katebutler did you push some code for #14267  already? @simivar be good to check if an issue is maybe already assigned to someone which means they are working on it already.

Re the name: It's not too important and actually the current implementation would also disable it for the UI I think.

tsteur:
Cheers for this @simivar

PR #14282, opened 2019-03-28
Load ThemeStyles in View and make theme-color themeable

### LINKED ISSUE #14230: Make theme-color themeable
Many people theme their Matomo instance to be more on brand, but when the main navbar color is changed, mobile phones still show the default blue color as it is hard-coded here:

https://github.com/matomo-org/matomo/blob/cb5453316711264b07b89a7077b51404a189aa98/plugins/Morpheus/templates/layout.twig#L25

It would be great if there was a way to modify this value without copying the whole template (which may cause issues after updates)

simivar:
Fixes #14230


diosmosis:
LGTM

PR #14283, opened 2019-03-28
Changing session name with const PIWIK_SESSION_NAME (#14254)

### LINKED ISSUE #14254: Changing session name with const PIWIK_SESSION_NAME not possible
According to https://matomo.org/docs/include-piwik-in-your-project/#bootstrap-php-execute-custom-code-before-matomo-runs the session name can be configured by setting the constant "PIWIK_SESSION_NAME".

This is not (longer?) possible. Tested with 3.9 and 3.9.1.

It seems that the session name is hardcoded in core/Session.php. I found no variable or constant to change it without fiddling around with the Session.php file and loosing my changes on every update.

simivar:
Fixes #14254

### DIFF
--- a/core/Session.php
+++ b/core/Session.php
@@ -54,6 +54,10 @@ class Session extends Zend_Session
             return;
         }
         self::$sessionStarted = true;
+    
+        if (defined('PIWIK_SESSION_NAME')) {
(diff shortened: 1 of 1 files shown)


diosmosis:
Thanks for the fix @simivar !

Issue #14219, opened 2019-03-18
Enabling SQL profiler on Response should be by setting enable_sql_profiler and not by debug


nicoaravena:
Hi there! I've been working on a fork of piwik (2.x) and I notice that when you enable the debug option, the database profiler it start to work and searching the origin of this I found in core/Tracker/Response.php this:

```php
class Response
{
    private $timer;

    private $content;

    public function init(Tracker $tracker)
    {
        ob_start(); // we use ob_start only because of Common::printDebug, we should actually not really use ob_start

        if ($tracker->isDebugModeEnabled()) {
            $this->timer = new Timer();

            TrackerDb::enableProfiling();
        }
    }
...
```

The problem is that the function isDebugModeEnabled() search for a global where it depends on the debug value in the config.php.ini and it should be validated by the setting enable_sql_profiler (as I mention in the title)

My propose is this:

```php

    public function init(Tracker $tracker)
    {
        // remove ob_start();
        if ((bool) TrackerConfig::getConfigValue('enable_sql_profiler')) {
            $this->timer = new Timer();

            TrackerDb::enableProfiling();
        }
    }
```

Also notice that `ob_start();` cause some problems with the printDebug function (specifically with the CLI), it is completely necessary? I think this could be removed also.

I forgot to say that I search in both branch version (2.x-dev and 3.x-dev) and this is untouched between those version changes.

tsteur:
The global is set depending on the `debug` tracker config setting. Not sure why it should be changed or what the benefit would be?

nicoaravena:
I think you misunderstood me, I was not saying to change the global debug, what I mean to say is that debugging is not always mandatory when you don't want to profile queries, because it increment process time when you are testing requests with a big records.

That is why I recommend to change the if condition.

Maybe this apply too `if ($tracker->isDebugModeEnabled() && (bool) TrackerConfig::getConfigValue('enable_sql_profiler')) `

tsteur:
Makes sense. The tracker doesn't have such a setting currently. Feel free to create a PR and add the setting `[Tracker] enable_sql_profiler = 1`.

Btw: we really don't recommend using 2.X anymore. It's not considered secure.

nicoaravena:
What about the existent `[Debug] enable_sql_profiler = 1` in `global.ini.php? This should be used for this purpose imo.

Also, as a team we are trying to upgrade as soon as we can.

tsteur:
[...]

PR #14288, opened 2019-03-29
New enable_sql_profiler INI setting for Tracker

simivar:
#14219

### DIFF
--- a/core/Tracker/Response.php
+++ b/core/Tracker/Response.php
@@ -25,7 +25,7 @@ class Response
     {
         ob_start(); // we use ob_start only because of Common::printDebug, we should actually not really use ob_start
 
-        if ($tracker->isDebugModeEnabled()) {
+        if ($tracker->isDebugModeEnabled() && TrackerConfig::getConfigValue('enable_sql_profiler')) {
             $this->timer = new Timer();
 
             TrackerDb::enableProfiling();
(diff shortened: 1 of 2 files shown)


Findus23:
I feel like this might get really confusing with the existing `enable_sql_profiler` setting:

https://github.com/matomo-org/matomo/blob/f6908294d00093234b2a27342f8f32b13a2b2232/config/global.ini.php#L142-L145

simivar:
That was @tsteur proposition from #14219. I feel like it can be confusing too but do you have any idea how it should be name? I can't think of anything myself.

tsteur:
@Findus23 using the sql profiler in tracking you don't want to use when you only want to use the SQL profiler for archiving or web.

core/Tracker/Response.php
```diff
@@ -25,7 +25,7 @@ public function init(Tracker $tracker)
     {
         ob_start(); // we use ob_start only because of Common::printDebug, we should actually not really use ob_start
 
-        if ($tracker->isDebugModeEnabled()) {
+        if ($tracker->isDebugModeEnabled() && TrackerConfig::getConfigValue('enable_sql_profiler')) {
```
tsteur: @simivar I'm wondering maybe it can be useful to enable `enable_sql_profiler` without enabling `$tracker->isDebugModeEnabled()`? Not sure... If we make it independent, we would also need to adjust the code in line `90` where we print the profiling results. I suppose it's fine though. Maybe let's just leave it like this for now.

tsteur:
Thanks for this @simivar Also created https://github.com/matomo-org/developer-documentation/pull/302/files to adjust the docs.

PR #14294, opened 2019-03-30
fix subtil typos

Findus23:
Hi,

Just wanted to test the code inspections in PhpStorm and came across two typos that I think cause the code to run incorrectly (but not wrong enough to be visible as a bug that has been reported)


diosmosis:
Looks good, will fix UI Tests on 3.x-dev

Issue #13666, opened 2018-11-03
Allow admin to create a password policy


Findus23:
similar to #13070 (show a strength meter in the browser)

At the moment the only limitations for Passwords in Matomo is that it needs to be from 6 to 200 characters long.
https://github.com/matomo-org/matomo/blob/b178978b979c5dcf20fe81a9b28034c0de4ce90e/plugins/UsersManager/UsersManager.php#L27-L28

But for organizations who have many employees it might be useful to disallow really weak passwords (123456).

Maybe for the beginning it would be enough to make the `PASSWORD_MIN_LENGTH` configurable (to avoid having overly complex password rules that force people to write down their passwords)

In addition it may be an idea for an plugin that checks all password hashes against https://haveibeenpwned.com/Passwords and disallows ones that are over a set threshold to avoid trivial passwords.

Findus23:
It turns out that thanks to the existing event, adding a haveibeenpwned-integration is really easy:

https://plugins.matomo.org/PasswordVerifier

simivar:
@Findus23 like discussed in #14295 I've created new plugin that is available on marketplace: https://github.com/simivar/matomo-password-policy-enforcer

Findus23:
I think with my plugin using haveibeenpwned and @simivar's plugin with common rules and the possibility to easily write a plugin that enforces any arbitrary rule, this should be solved.

mattab:
FYI we're thinking of implementing something simple for password policy in core, see https://github.com/matomo-org/matomo/issues/19961

PR #14295, opened 2019-03-31
Allow admin to create a password policy

simivar:
#13666 

Settings available in UI:
 2019-03-31_19h38_14


Findus23:
Hi @simivar,

Thanks for your contribution. In the future it might be a good idea to comment on an issue indicating that you want to work on the issue. Often I open issues to start a discussion on how something should be solved and not as a finished proposal.

In this case I honestly thought about a config file parameter as I think this UI would be only useful for a very small fraction of Matomo users and another UI makes the settings even more complex. So maybe this could be extracted as a plugin.
But maybe someone else knows more?

simivar:
@Findus23 sure, I'll remember that in the future. 

I was thinking about adding options to config file but decided for UI mainly because of two things:
- not all users read what options are available in config files 
- it should be encouraged to use Password Policy and to have all those options turned on, this way more people will see that and opt-in on that

If you strongly think that it should be moved to config I will gladly edit my PR.

fdellwing:
> it should be encouraged [...] to have all those options turned on

I highly disagree with this statement, see the discussion there: https://github.com/matomo-org/matomo/pull/14176

tsteur:
I think neither options nor config is needed and executing these checks for everyone without any setting or config should be fine. Yes you can have save passwords without a number, and having this or that doesn't make it safe... but it's still better maybe to just force different kind of characters (the UI doesn't need to say that makes it a strong password). Anyway, I would just not want any config or UI setting for this ideally. That's just me though.

Findus23:
If someone really needs this feature it should be easy to create a plugin similar to https://github.com/Findus23/plugin-PasswordVerifier/blob/master/PasswordVerifier.php that implements arbitrary rules.

simivar:
@mattab what's your take on that?

mattab:
Hi @simivar 
Thank you for the Pull Request! It's a valuable feature :+1:  But at this time we'd prefer to have this feature optional in a plugin rather than in core. 
-> Would it be possible to move your feature to a plugin and publish it on the marketplace?

simivar:
@mattab @Findus23 
I've created https://github.com/simivar/matomo-password-policy-plugin. But, there's a problem. I get an e-mail stating:

> -> The tag was created in the repository simivar/matomo-password-policy-plugin but we
expected the tag in PiwikPRO/plugin-PasswordPolicy. If the repository has moved, reply to
this email to prove ownership and indicate who the new owner should be. <-

From what I gather Matomo and PiwikPRO are two separate entities. I can't find "Password policy" plugin on Matomo Marketplace too. What should I do?

Findus23:
@simivar Every plugin ID has to be unique as that's what Matomo uses to identify plugins.

Even if the old plugin is removed from the marketplace (I think it is already or at least hidden on the public pages) having the same ID would maybe cause odd issues.

So I'd recommend you to just use another name for the plugin:
https://github.com/simivar/matomo-password-policy-plugin/blob/1839638a86ec612e0e1d340fb32e260fb4c9a46a/plugin.json#L2

simivar:
@Findus23 I think that the list of all, even hidden, plugins should be searchable somewhere so that would be clear that such plugin name is taken.

tsteur:
@simivar so far I think this is only the first time this happened. I think the error message is somewhat helpful in this case but happy to change / improve it. Like we could say "If this is not your repo, please rename the plugin and publish it again". I'm thinking even if we have such a list available somewhere, devs might not look at it first and changing the plugin name later should be in most cases somewhat easy (find/replace). BTW: The reason it isn't showing this plugin is because it is only available up to Piwik 2. So over time (in a few years) this might become more of an issue and might need to actually make such a list available at some point.

PR #14309, opened 2019-04-02
Let controllers specify what type a view to be rendered is

diosmosis:
Currently when normal non-admin views are rendered in admin controllers, it is assumed they are admin views. Settings for admin views are thus set, which can cause problems for normal views.

For example, say a matomo allows embedding pages, but not settings pages, then a user loads a site with no data. The `siteWithoutData()` action is invoked, and it is assumed to be an admin controller action, so `setBasicVariablesAdminView()` is called. This method assumes the view is a settings page so it will disable embedding pages since `enable_framed_settings` is not enabled.

This PR fixes this by allowing the view type to be specified to `renderTemplate()` and other methods. It's a bit odd looking, but it's hard to find a way to fix this properly given how many variables get set.


core/Plugin/Controller.php
```diff
@@ -576,11 +583,16 @@ protected function setMaxDateView(Date $maxDate, $view)
      * Will exit on error.
      *
      * @param View $view
+     * @param string|null $viewType 'basic' or 'admin'. If null, set based on the type of controller.
```
tsteur: if someone extends this controller, and passes `admin` it wouldn't work as expected right?

diosmosis: Are you suggesting if admin is passed, we should only allow it if `$this instanceof ControllerAdmin`? Or something else?

tsteur: I just meant might need to be updated in the docs ? 
The check with instanceof might be useful though and prevent false expectations if admin is passed. This way a developer would directly notice to extend different class.

diosmosis: It would sort of work as expected if `'admin'` is passed, since `setBasicVariablesView()` is overridden in ControllerAdmin. I think there would be many more variables set (and possibly an exception if there's no idsite or something like that).

We could simply keep it undocumented except for `renderTemplate()`, doesn't really need to be set anywhere else I think.

tsteur: Or should we just move `setBasicVariablesNoneAdminView` etc to `plugin\Controller` and make it work even when it extends only `Plugin\Controller` and not `ControllerAdmin`? just a random thought, no preference in any direction here

diosmosis: I'm not sure what that would solve?

tsteur: You have 

```php
class Foo extends Plugin\Controller {
   public function myAction(){
      return $this->renderTemplate('bar.twig', 'admin');
   }
}
``` 
would then maybe work instead of needing to change the parent controller to `class Foo extends Plugin\ControllerAdmin`? I haven't looked too much into the code though

diosmosis: The problem is with:

```
class SitesManagerController extends Plugin\AdminController {
    public function myNormalAction() {
        return $this->renderTemplate('bar.twig');
    }
}
```

Where `ControllerAdmin::setBasicVariablesView()` calls `ControllerAdmin::setBasicVariablesAdminView()`, when we don't want it to.

diosmosis: Or are you just talking about that specific case? I'm not sure if we should allow it, ie, should normal controllers be allowed to render admin pages?

tsteur: > should normal controllers be allowed to render admin pages?
No preference. It's just that `renderTemplate()` in `Plugin\Controller` has this parameter and it's documented to set `admin` optionally. So I would expect it to work even when `Plugin\Controller` is extended. We can also throw an exception in this case though.

diosmosis: Will throw an exception. We could change `$viewType` to `$isBasicView` or something, but then a dev could supply `false`.

This isn't an ideal solution, but I'm not too sure of another easy/quick one.

tsteur: > Will throw an exception. 

👍

tsteur:
@diosmosis it seems this is actually breaking API: https://travis-ci.org/matomo-org/matomo/jobs/515032508#L925

The method `renderTemplate` is marked as `@api` preventing us from adding any variables because it would be impossible for a plugin to be compatible with previous and newer versions (older versions would not want to see the parameter while newer ones should have). Assuming this isn't used much in the wild we could possibly be fine with the breaking change but would need to adjust tag manager and increase the required Matomo version number for tag manager.

I will look through the detailed code changes later again.

core/Plugin/Controller.php
```diff
@@ -269,27 +269,36 @@ protected function renderView(ViewInterface $view)
...
-        if ($this instanceof ControllerAdmin) {
-            $this->setBasicVariablesView($view);
-        } elseif (empty($this->site) || empty($this->idSite)) {
-            $this->setBasicVariablesView($view);
+        if ($this instanceof ControllerAdmin && $viewType == 'admin') {
+            $this->setBasicVariablesView($view, $viewType);
+        } elseif ((empty($this->site) || empty($this->idSite))
+            && $viewType == 'admin'
+        ) {
+            $this->setBasicVariablesView($view, $viewType);
```
tsteur: @diosmosis wouldn't we also call this when using `viewType=view` in some cases? basic variables does different things than general variables?

diosmosis: Yes seems so, I'm not sure why I made this change... I'll have to test again

PR #14316, opened 2019-04-05
Permit segmenting by order ID

### LINKED ISSUE #9981: New segment 'ecommerceOrderId' to select a particular Ecommerce Order ID
Goal: let user [segment](http://piwik.org/docs/segmentation/) Ecommerce Orders by specifying a segment `ecommerceOrderId` or `orderId`.

Requested in https://forum.piwik.org/t/using-the-reporting-api-to-query-an-ecommerce-order-by-orderid/18376

Similar to https://github.com/piwik/piwik/issues/4519 and https://github.com/piwik/piwik/issues/9175

katebutler:
Fixes #9981


plugins/Ecommerce/Columns/Order.php
```diff
@@ -19,13 +20,16 @@ class Order extends BaseConversion
...
-    protected $nameSingular = 'Ecommerce_Order';
+    protected $nameSingular = 'Ecommerce_OrderId';
     protected $namePlural = 'Ecommerce_Orders';
     protected $metricId = 'orders';
+    protected $segmentName = 'orderId';
 
     public function configureMetrics(MetricsList $metricsList, DimensionMetricFactory $dimensionMetricFactory)
     {
         $metric = $dimensionMetricFactory->createMetric(ArchivedMetric::AGGREGATION_UNIQUE);
+        $metric->setName("ecommerce_order_id");
```
tsteur: @katebutler do you remember if the name was needed? Changing the name of existing metrics may break existing custom reports that have the old metricId assigned.

tsteur:
Left a comment, otherwise worked for me.

plugins/Ecommerce/Columns/Order.php
katebutler: @tsteur Turns out it's not needed, have removed it

tsteur:
@katebutler this test seems to fail: https://travis-ci.org/matomo-org/matomo/jobs/517626504#L832

this one too which is maybe related to the order that we are now tracking and should be an expected change I reckon: https://travis-ci.org/matomo-org/matomo/jobs/517626504#L811

The UI tests stopped running in the middle of the test. [Restarted the tests](https://travis-ci.org/matomo-org/matomo/jobs/517626512) and hoping it completes. I suppose there might be a few screenshot tests to update. Can show you tomorrow how to do it. 

Seems otherwise good to merge except for the tests that need to be updated.

tsteur:
@katebutler just fyi here's a test that needs to be fixed: https://travis-ci.org/matomo-org/matomo/jobs/517626504#L811-L825

The test fails now correctly and we only need to update the values

PR #14317, opened 2019-04-05
Task to purge archives for deleted websites and segments

### LINKED ISSUE #13966: Check that when a website (or a segment) is deleted, we also delete all archive data (numeric, blob)
In order to space disk size and costs, when a website is deleted, or when a segment is deleted, we need to make sure that archived reports stored in the `archive_numeric*` and `archive_blob*` tables are also deleted.

katebutler:
Fixes #13966


core/Archive/ArchivePurger.php
```diff
@@ -154,6 +156,84 @@ public function purgeOutdatedArchives(Date $dateStart)
...
+
+            $this->logger->debug("[Deleted IDs: {deletedIds}]", array(
+                'deletedIds' => implode(',', $idArchivesToDelete)
+            ));
+        } else {
+            $this->logger->debug("No archives for deleted segments found in archive numeric table for {date}.", array('date' => $dateStart));
+        }
+
+        return $deletedRowCount;
+    }
```
diosmosis: There's some code redundancy in purgeDeletedSiteArchives & purgeDeletedSegmentArchives which could be removed.

core/Archive/ArchivePurger.php
```diff
@@ -154,6 +156,84 @@ public function purgeOutdatedArchives(Date $dateStart)
...
+        //Strip off the plugin name if there is one
+        $dotPlace = strpos($segmentHash, '.');
+        if ($dotPlace !== false) {
+            $segmentHash = substr($segmentHash, 0, $dotPlace);
+        }
+
+        //Check that it's in the list of segments, either for ALL sites or for THIS site
+        return  isset($validSegmentIds[0][$segmentHash]) ||
+                isset($validSegmentIds[$archiveRow['idsite']][$segmentHash]);
+    }
```
diosmosis: This method doesn't appear to be used, is it still needed?

core/Archive/ArchivePurger.php
```diff
@@ -154,6 +156,84 @@ public function purgeOutdatedArchives(Date $dateStart)
...
+                'deletedIds' => implode(',', $idArchivesToDelete)
+            ));
+        } else {
+            $this->logger->debug("No archives for deleted sites found in archive numeric table for {date}.", array('date' => $dateStart));
+        }
+
+        return $deletedRowCount;
+    }
+
+    public function purgeDeletedSegmentArchives(Date $dateStart, array $validSegmentHashes)
```
diosmosis: Not super important, but it might be a good idea to document the structure of `$validSegmentHashes` since I thought it was just a simple array at first. Or maybe use a more descriptive name.

diosmosis:
Left a couple minor comments, otherwise looks good to merge.

core/Archive/ArchivePurger.php
katebutler: Yup, have moved this logic into a purge() method

core/Archive/ArchivePurger.php
katebutler: Nope, removed

core/Archive/ArchivePurger.php
katebutler: Fair enough, it is now `$segmentHashesByIdSite`

plugins/CoreAdminHome/Tasks.php
```diff
@@ -258,6 +262,48 @@ public function updateSpammerBlacklist()
...
+            try {
+                $segment = new Segment($row['definition'], array());
+                $segmentHashes[(int)$row['enable_only_idsite']][] = $segment->getHash();
+            } catch (\Exception $ex) {
+                //Segment is invalid ... into the sea its archives shall go
+            }
+        }
+
+        return $segmentHashes;
+    }
```
diosmosis: Would it be possible to test this function somehow?

katebutler: Yup, have made the method public and added some tests

core/DataAccess/Model.php
```diff
@@ -333,6 +333,65 @@ public function getSitesWithInvalidatedArchive($numericTable)
...
+    /**
+     * Get a list of IDs of archives that don't have any matching rows in the site table. Excludes temporary archives
+     * that may still be in use, as specified by the $oldestToKeep passed in.
+     * @param string $archiveTableName
+     * @param string $oldestToKeep Datetime string
+     * @return array of IDs
+     */
+    public function getArchiveIdsForDeletedSites($archiveTableName, $oldestToKeep)
+    {
+        $sql = "SELECT DISTINCT idarchive FROM " . $archiveTableName . " a "
```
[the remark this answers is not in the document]
tsteur: If we went with this solution, would also need to make sure to execute this method for `blob` archive tables.

plugins/CoreAdminHome/Tasks.php
```diff
@@ -258,6 +262,48 @@ public function updateSpammerBlacklist()
...
+    public function getSegmentHashesByIdSite()
+    {
+        //Get a list of hashes of all segments that exist now
+        $sql = "SELECT DISTINCT definition, enable_only_idsite FROM " . Common::prefixTable('segment')
+            . " WHERE deleted = 0";
+        $rows = Db::fetchAll($sql);
+        $segmentHashes = array();
+        foreach ($rows as $row) {
+            try {
+                $segment = new Segment($row['definition'], array());
```
tsteur: I'm quite certain we will need to pass the `idSites` here. Otherwise it won't find segments that only exist for a specific site. Eg `dimension1` is a segment for custom dimensions and it will be only valid segment if the idSite is passed.  

You can probably reproduce this by following these steps (haven't tested it)

* Creating a custom dimension
* Creating a segment that uses the created custom dimension
* running this command using eg `./console core:run-scheduled-tasks "Piwik\Plugins\CoreAdminHome\Tasks.purgeOrphanedArchives" --force`.

If `idSite=0`, then it should be fine to not pass any site parameter AFAIK.

tsteur: Once you pass `idSite` it should work like `array($row['idSite'])` but you may need to update the `CustomDimensions` submodule and fetch the latest updates in `master` as I fixed an issue around this just the other day.

plugins/CoreAdminHome/Tasks.php
```diff
@@ -258,6 +262,48 @@ public function updateSpammerBlacklist()
...
+                $segmentHashes[(int)$row['enable_only_idsite']][] = $segment->getHash();
```
tsteur: Some servers may trigger notices when not setting `$segmentHashes[(int)$row['enable_only_idsite']]` to an empty array first but not sure. Be good just to double check like 

```php
if (!isset($segmentHashes[(int)$row['enable_only_idsite']])) {
$segmentHashes[(int)$row['enable_only_idsite']] = array();
}
```

plugins/CoreAdminHome/Tasks.php
```diff
@@ -258,6 +262,48 @@ public function updateSpammerBlacklist()
...
+            } catch (\Exception $ex) {
+                //Segment is invalid ... into the sea its archives shall go
```
tsteur: be good to log a notice in this case maybe? This will help us reproduce why potentially archives were deleted. 

Actually, I'm thinking it may be too risky if we delete them. There could be edge cases where we eg deploy code or there are temporary bugs where a segment can not be found. In this case we wouldn't want to delete the archives as it could cause quite a bit of trouble, especially on the cloud maybe? 

I wonder if we should in this case maybe try to get the segment hash manually by just doing `md5(urldecode($row['definition']));`. Maybe we could actually do this in general... then we also don't need to care about passing `idSites` to `new Segment()` etc? We would just need to create a static method for it that we use here like `Segment::getSegmentHash()` which could be also used in `$segment->getHash`?

core/DataAccess/Model.php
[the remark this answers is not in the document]
tsteur: I see there is already quite some logic for `ArchivePurger::deleteArchiveIds` there which is reused in this case so it's fine to keep the code this way. Just a bit scared of fetching too many archiveIds into memory etc but shouldn't be an issue I reckon.

core/DataAccess/Model.php
```diff
@@ -333,6 +333,65 @@ public function getSitesWithInvalidatedArchive($numericTable)
...
+        }
+
+        $isValidSegmentSql = implode(' OR ', $validSegmentClauses);
+
+        $sql = 'SELECT idarchive FROM ' . $archiveTableName
+            . ' WHERE name LIKE "done%" AND name != "done"'
+            . ' AND ts_archived < ?'
+            . ' AND NOT (' . $isValidSegmentSql . ')';
+
+        $rows = Db::fetchAll($sql, array($oldestToKeep));
```
tsteur: Just fyi in case someone wants to have a look at the query it generates... in my case it created a query like this (seems to look good):

```sql
SELECT idarchive FROM piwik_archive_numeric_2019_01 WHERE name LIKE "done%" AND name != "done" AND ts_archived < ? AND NOT 

((idsite = 1 AND (name LIKE "done11f27b46b44cf35a9ecc9bd3661022c3%" OR name LIKE "done3c6cd97084beb5acf44c03c8c2d90240%" OR name LIKE "done348b89787ca833d7b9c71a2583dabe2c%" OR name LIKE "done7c4db1e772ea091d315d3f86355f4644%" OR name LIKE "done1f151d88667d0493ef083fda0c75e399%" OR name LIKE "donebc7b3e3d298d188ed02841af5ede7e0e%" OR name LIKE "donec561c80c35c1196d4cec668c0730ce66%" OR name LIKE "done48b4361e2888dac452de11a8db8da931%" OR name LIKE "doned1c4f52a073c440a5111b50a1cf8f7fb%" OR name LIKE "done13328b096b251986da98c131fb703fba%" OR name LIKE "done5df82b31b7a978e326be3562b098e992%" OR name LIKE "done563819b40691e8fd5db123e2be3e7c07%" OR name LIKE "done598457137b9cbd5d5bb9f990d81e427f%" OR name LIKE "done4578744398d6a1ac38bb7532c75cc989%" OR name LIKE "donee9e48ac3f41fd5615137e0761fdc5c90%" OR name LIKE "done4d6ceb7591330b316782ef6bd1d2a080%" OR name LIKE "donecfb79a09244730090623e6c7227e8387%" OR name LIKE "doneddccaa229c48150e2920c7fe376c9f11%" OR name LIKE "done05c9661bc29fd4667b97e5e6e5a51a86%" OR name LIKE "done38ac1901cc3e433ad551f6544a1cc7a2%" OR name LIKE "done34207c7198689e9bdba534de32094f40%" OR name LIKE "done9e6275849e64cfa8a5376a7112b47a45%" OR name LIKE "done4e816a3d7169f9c3d20d2c5c7df1248d%" OR name LIKE "done12685caa4f74f8ecb62bcbeabde3b936%" OR name LIKE "done6fb490cfce17323e01f4d11d61ea9679%" OR name LIKE "donee42f13629073b12be80cc376a0040c14%" OR name LIKE "done642a27678c886639439fc1b922ac1af0%" OR name LIKE "done567bde329b39232dbf63690ed85394b8%" OR name LIKE "done86bcf168b0e43cfc46c5d1e68a6aaf62%" OR name LIKE "done43b771d91cfa18d12ccdf766ab1e3235%" OR name LIKE "donea8836c71226531af4e4298bae1dd5997%" OR name LIKE "done870df59630c7ed4e74b8584a0d72609b%" OR name LIKE "done5de5753680cae0cc8c26e6db6f2a8fc4%" OR name LIKE "donef3301aa5b40ba36bf5c7659294900f42%" OR name LIKE "done35ec2319cc96284b002c3b661add6e1f%")) OR (idsite = 14 AND (name LIKE "donefa4159eaae7296c3a8beea04da0642f7%")) 

OR (name LIKE "done1432586f23a19c5db5a3c03e06cbb55f%") OR (idsite = 99 AND (name LIKE "done56e6bad0310dad6bbbf1dad1db2cf096%")))
```

plugins/CoreAdminHome/Tasks.php
```diff
@@ -258,6 +262,48 @@ public function updateSpammerBlacklist()
...
+        $archiveTables = ArchiveTableCreator::getTablesArchivesInstalled('numeric');
+
+        $datesPurged = array();
+        foreach ($archiveTables as $table) {
+            $date = ArchiveTableCreator::getDateFromTableName($table);
+            list($year, $month) = explode('_', $date);
+
+            $dateObj = Date::factory("$year-$month-15");
+
+            $this->archivePurger->purgeDeletedSiteArchives($dateObj);
```
tsteur: just a random thought and not too important... from what I see you pass the date object to generate the archive table name in `getDeletedSiteArchiveIds` and `getDeletedSegmentArchiveIds`. If it makes things simpler, could probably remove a lot of code in line `272-275` and simply directly pass the `tableName`. But I suppose you do it because `ArchivePurger::purgeOutdatedArchives` does it too... makes sense. Probably best to keep it like it is for consistency. (ignore this comment :) )

image

tsteur (changes requested):
@katebutler left a few comments. Nothing major. Looks really good in general and be so good to have this feature 👍

core/DataAccess/Model.php
```diff
@@ -333,6 +333,65 @@ public function getSitesWithInvalidatedArchive($numericTable)
...
+    public function getArchiveIdsForDeletedSites($archiveTableName, $oldestToKeep)
+    {
+        $sql = "SELECT DISTINCT idarchive FROM " . $archiveTableName . " a "
+            . " LEFT JOIN " . Common::prefixTable('site') . " s USING (idsite)"
+            . " WHERE s.idsite IS NULL"
+            . " AND ts_archived < ?";
+
+        $rows = Db::fetchAll($sql, array($oldestToKeep));
+
+        return array_map(function($row) { return $row['idarchive']; }, $rows);
```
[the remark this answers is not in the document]
katebutler: Cool, changed

plugins/CoreAdminHome/Tasks.php
katebutler: OK have changed

plugins/CoreAdminHome/Tasks.php
katebutler: Have added a unit test for this case - now skipping segment validation anyway so this is no longer an issue

plugins/CoreAdminHome/Tasks.php
katebutler: Have extracted into a separate method as suggested, now skipping segment constructor/validation and just calculating the hash on the definition string

tsteur:
Well done @katebutler 👍

PR #14321, opened 2019-04-07
Use better random string generator in 2fa lib

tsteur:


diosmosis:
The docs for `Common::getRandomString` say `Do not use for security related purposes (the string is not truly random).`, maybe we should use `random_bytes()` if it's available?

fdellwing:
The best way imho is using `random_bytes()`, but this will only work for PHP7+.

So if this is not a problem, you should use this polyfill (https://github.com/paragonie/random_compat) and than use `bin2hex(random_bytes($secretLength));`.

If you do not want to implement another dependency, you can do your own PHP7+ check and otherwise use `bin2hex(openssl_random_pseudo_bytes($secretLength));`.

tsteur:
I doubt `openssl_random_pseudo_bytes ` will be available everywhere. @diosmosis I think we can remove the content as in most cases it will use `random_int` and the comment was probably never removed after updating the `getRandomInt` method.

diosmosis:
@tsteur makes sense to me

PR #14322, opened 2019-04-07
require TwoFA to be verified before updating code base

### LINKED ISSUE #14251: Update without proper two factor authentication
When visiting self-hosted Matomo installation (with two-factor authentication enabled):
1. Matomo prompted for login/pass, but did not require 2 factor authentication
2. Matomo prompted for upgrade
3. I initiated upgrade and Matomo completed it
4. An error page—referencing something like a loop. I didn't copy it, sorry.
5. Browsed back
6. Matomo showed upgrade successfully completed page
7. Matomo finally asks for 2 factor authentication

It seems to me 2 factor authentication should be successfully completed before prompting for upgrade and before permitting initiation of upgrade.

tsteur:
fix https://github.com/matomo-org/matomo/issues/14251

Not sure if there could be any side effect, can't really think of any though. @diosmosis do you maybe have any idea?

### DIFF
--- a/core/Common.php
+++ b/core/Common.php
@@ -635,7 +635,6 @@ class Common
 
     /**
      * Generate random string.
-     * Do not use for security related purposes (the string is not truly random).
      *
      * @param int $length string length
      * @param string $alphabet characters allowed in random string
--- a/plugins/TwoFactorAuth/TwoFactorAuth.php
+++ b/plugins/TwoFactorAuth/TwoFactorAuth.php
@@ -172,7 +172,7 @@ class TwoFactorAuth extends \Piwik\Plugin
             return false;
         }
 
-        if ($module === 'CoreUpdater') {
+        if ($module === 'CoreUpdater' && $action !== 'newVersionAvailable' && $action !== 'oneClickUpdate') {
             return false;
         }
 


diosmosis:
Maybe `oneClickResults` should require 2FA too?

Actually, I'm not so sure this solves the problem completely. Thinking about it, it doesn't really make sense that everyone is able to see the "complete the update" screen, maybe that can be the complete fix (for a later release, not necessarily in this PR).

tsteur:
> Actually, I'm not so sure this solves the problem completely. Thinking about it, it doesn't really make sense that everyone is able to see the "complete the update" screen, maybe that can be the complete fix (for a later release, not necessarily in this PR).

You mean the screen where the DB is being upgraded etc? I think it is because sometimes the login may not function anymore without the DB upgrade etc

diosmosis:
It might not, but it probably should work regardless. I think that's something we can test for before releasing (the one click update test might be a possible place).

tsteur:
It depends though eg from which version you up date.

diosmosis:
@tsteur do you think the `oneClickResults` controller action should also require 2FA?

tsteur:
@diosmosis I think it shouldn't. There may be various issues where you may not be logged in or have other problems etc. It should be fine to access this page even logged out.

PR #14332, opened 2019-04-09
Bar graph should only have Users and Unique Visitors options when they are valid

### LINKED ISSUE #14042: Dashboard screen is broken when using unique visitors for range
We have users experiencing this issue on the cloud. Example dashboard:

```
{"config":{"layout":"30-30-40"},"columns":[[{"uniqueId":"widgetVisitsSummarygetEvolutionGraphforceView1viewDataTablegraphEvolution","parameters":{"forceView":"1","viewDataTable":"graphEvolution","module":"VisitsSummary","action":"getEvolutionGraph","widget":1,"columns":["nb_uniq_visitors"],"columns_to_display":["nb_uniq_visitors"],"rows":[],"rows_to_display":[]},"isHidden":false},{"uniqueId":"widgetVisitsSummarygetforceView1viewDataTablesparklines","parameters":{"forceView":"1","viewDataTable":"sparklines","module":"VisitsSummary","action":"get","widget":1},"isHidden":false}],[{"uniqueId":"widgetCustomDimensionsgetCustomDimensionidDimension1","parameters":{"module":"CustomDimensions","action":"getCustomDimension","idDimension":"1","widget":1,"viewDataTable":"graphPie","columns":["nb_uniq_visitors"],"columns_to_display":["nb_uniq_visitors"],"rows":[],"rows_to_display":[]},"isHidden":false},{"uniqueId":"widgetCustomDimensionsgetCustomDimensionidDimension2","parameters":{"module":"CustomDimensions","action":"getCustomDimension","idDimension":"2","widget":1,"viewDataTable":"graphPie","columns":["nb_uniq_visitors"],"columns_to_display":["nb_uniq_visitors"],"rows":[],"rows_to_display":[]},"isHidden":false},{"uniqueId":"widgetCustomDimensionsgetCustomDimensionidDimension4","parameters":{"module":"CustomDimensions","action":"getCustomDimension","idDimension":"4","widget":1,"viewDataTable":"graphPie","columns":["nb_uniq_visitors"],"columns_to_display":["nb_uniq_visitors"],"rows":[],"rows_to_display":[]},"isHidden":false}],[{"uniqueId":"widgetCustomDimensionsgetCustomDimensionidDimension3","parameters":{"module":"CustomDimensions","action":"getCustomDimension","idDimension":"3","widget":1,"viewDataTable":"graphVerticalBar","columns":["nb_uniq_visitors"],"columns_to_display":["nb_uniq_visitors"],"rows":[],"rows_to_display":[]},"isHidden":false},{"uniqueId":"widgetCustomDimensionsgetCustomDimensionidDimension5","parameters":{"module":"CustomDimensions","action":"getCustomDimension","idDimension":"5","widget":1,"viewDataTable":"graphVerticalBar","columns":["nb_uniq_visitors"],"columns_to_display":["nb_uniq_visitors"],"rows":[],"rows_to_display":[]},"isHidden":false}]]}
```

As you can see it uses unique visitors which a user chose for example while selecting the day period. But then when switching to "Range" period for example, the widgets render an empty/broken widget and the whole UI is basically broken afterwards.

This is because there are rows for the data table, but there is no entry for the unique visitors (I think). Therefore there is no `this.data[0].length` in 
```
index.php?module=Proxy&action=getCoreJs&cb=52c065e5d0f32ad90504d57b7b3f5e72:formatted:57479 Uncaught TypeError: Cannot read property 'length' of undefined
    at exports.JqplotBarGraphDataTable._setJqplotParameters (VM1264 index.php:formatted:1)
    at exports.JqplotBarGraphDataTable.init (VM1264 index.php:formatted:1)
    at exports.JqplotBarGraphDataTable.DataTable (VM1264 index.php:formatted:1)
    at exports.JqplotBarGraphDataTable.exports.JqplotGraphDataTable (VM1264 index.php:formatted:1)
    at new exports.JqplotBarGraphDataTable (VM1264 index.php:formatted:1)
    at HTMLDivElement.<anonymous> (VM1264 index.php:formatted:1)
    at Function.each (VM1264 index.php:formatted:1)
    at n.fn.init.each (VM1264 index.php:formatted:1)
    at Function.DataTable.initNewDataTables (VM1264 index.php:formatted:1)
    at HTMLDocument.eval (eval at globalEval (VM1264 index.php:formatted:1), <anonymous>:2:39)
``` 

There must be some code like here: https://github.com/matomo-org/matomo/blob/3.8.1/plugins/CoreVisualizations/JqplotDataGenerator/Evolution.php#L54-L58 that only adds data to the graph series when there is data for the specific metric. 

When replacing all metrics with `nb_visits`, then the dashobard renders nicely. Used are pie and bar charts for example. Ideally we would show instead "no data" message or make the visualisations work with no data.

This is what the dashboard looks like:

image

katebutler:
These should only be displayed when the visualization's date range supports these metrics (i.e. for a single date, but not a date range). This prevents the user from making selections which will trigger JS errors and prevent the visualization from rendering anyway.

Fixes #14042


plugins/CoreVisualizations/Visualizations/Graph.php
```diff
@@ -135,26 +149,48 @@ public function afterAllFiltersAreApplied()
...
                 'displayed'   => in_array($column, $this->config->columns_to_display)
             );
         }
-
         $this->config->selectable_columns = $transformed;
     }
+
+    protected function generateSelectableColumns()
+    {
+        $selectableColumns = array('nb_visits', 'nb_actions');
```
tsteur: @katebutler not sure about the columns we add by default. Many reports won't have the columns `nb_actions`, `nb_conversions`, `revenue`, `nb_uniq_visitors ` and `nb_users`. Some reports might not even have the `nb_visits`. Did you copy this logic from somewhere?

Otherwise we could maybe simply set the first non-label column from the first row?

And if there is only `label` in the first row, we could maybe default to `nb_visits`?

plugins/CoreVisualizations/Visualizations/Graph.php
```diff
@@ -135,26 +149,48 @@ public function afterAllFiltersAreApplied()
...
+            $goalMetrics       = array('nb_conversions', 'revenue');
+            $selectableColumns = array_merge($selectableColumns, $goalMetrics);
+        }
+
+        $this->config->selectable_columns = $selectableColumns;
+    }
+
+    protected function ensureValidColumnsToDisplay()
+    {
+        $columnsToDisplay = $this->config->columns_to_display;
```
tsteur: I wonder if this logic here is needed? I wonder about possible side effects where reports declare `columns_to_display` but not `selectable_columns`.

In general this is a great addition that makes sense though. If something is not selectable, we shouldn't show it. Just scared about possible side effects or breaking something. Might not be an issue though but have a feeling there are some reports that might break as I can't find where the reports usually add this info.

tsteur: We can quickly chat about it tmrw might be easiest

plugins/CoreVisualizations/Visualizations/Graph.php
katebutler: The existing logic was to just use all of them and hope for the best :smile:  but I have reworked it to use all columns except `label` if none of these are present in the dataset

plugins/CoreVisualizations/Visualizations/Graph.php
katebutler: I've changed it to match against `$dataTable->getColumns()` rather than `selectable_columns` - this should hopefully avoid any side effects.

PR #14349, opened 2019-04-15
Remove diagnostics plugin from list of always-activated plugins

### LINKED ISSUE #14304: Allow to disable Diagnostics plugin
Currently the Diagnostics plugin cannot be disabled:

https://github.com/matomo-org/matomo/blob/2ec7485470701ef5f11521d6bbe62d583e25cddf/core/Plugin/Manager.php#L72-L77
 
For some reason we need to disable Diagnostics plugin on Matomo Cloud.

-> So we'd like to be able to disable the Diagnostics plugin in Matomo.

If this is not easily possible, then instead we can add a new config ini setting to disable a feature of the plugin in particular (not wanted on Cloud) which displays the UI notice:
image

Done here: 

https://github.com/matomo-org/matomo/blob/e26f5e7a359e990b4c4d726de85839e8923619ab/plugins/Diagnostics/Diagnostics.php#L38-L64

katebutler:
Fixes #14304

### DIFF
--- a/core/Plugin/Manager.php
+++ b/core/Plugin/Manager.php
@@ -71,7 +71,6 @@ class Manager
     // These are always activated and cannot be deactivated
     protected $pluginToAlwaysActivate = array(
         'CoreHome',
-        'Diagnostics',
         'CoreUpdater',
         'CoreAdminHome',
         'CoreConsole',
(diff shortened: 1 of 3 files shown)


tsteur:
Cheers @katebutler

PR #14351, opened 2019-04-16
Submit report download link as a post request

### LINKED ISSUE #12721: In links to HTML/PDF reports downloads, do not show token_auth
In Administration > Email reports, users can download an email report. The "Download" link includes the `token_auth`. This is problematic because `token_auth` are then leaked in server access logs and browser history.

-> We should change it so that the link doesn't include the token_auth, and instead the "download" should be a  POST request with the `token_auth` in the POST body.

katebutler:
Fixes #12721

### DIFF
--- a/plugins/ScheduledReports/angularjs/manage-scheduled-report/manage-scheduled-report.controller.js
+++ b/plugins/ScheduledReports/angularjs/manage-scheduled-report/manage-scheduled-report.controller.js
@@ -177,35 +177,8 @@
             resetParameters(this.report.type, this.report);
         };
 
-        this.displayReport = function (url) {
-            // Split the URL into its parts
-            var urlParts = url.split('?');
-            var basePath = urlParts[0];
-            var params = urlParts[1].split('&');
-
-            var formContent = '';
-            var separator = '?';
-            $.each(params, function(key, queryStr) {
-                var queryParts = queryStr.split('=');
-                if (queryParts[0] == 'token_auth') {
-                    // Hide the token_auth in an input element
-                    formContent = $('<input>')
-                        .attr('type', 'hidden')
-                        .attr('name', queryParts[0])
-                        .attr('value', queryParts[1]);
-                } else {
-                    // Everything else can be exposed to the user in the URL
-                    basePath += separator + queryStr;
-                    separator = '&';
-                }
-            });
-
-            // Build and submit the form
-            // The target="_blank" only works if the form's id attribute is also set
-            $('<form action="' + basePath + '" method="POST" target="_blank" id="downloadReportForm"></form>')
-                .html(formContent)
-                .appendTo($(document.body))
-                .submit();
+        this.displayReport = function (reportId) {
+            $('#downloadReportForm_' + reportId).submit();
         };
 
         // Email now
--- a/plugins/ScheduledReports/angularjs/manage-scheduled-report/manage-scheduled-report.controller.js
+++ b/plugins/ScheduledReports/angularjs/manage-scheduled-report/manage-scheduled-report.controller.js
@@ -177,6 +177,26 @@
             resetParameters(this.report.type, this.report);
         };
 
+        this.displayReport = function (url) {
+            // Split the URL into its parts
+            var urlParts = url.split('?');
+            var basePath = urlParts[0];
+            var params = urlParts[1].split('&');
+
+            // Generate an <input> for each query param
+            var form = '';
+            $.each(params, function(key, queryStr) {
+                var queryParts = queryStr.split('=');
+                form += '<input type="hidden" name="' + queryParts[0] + '" value="' + queryParts[1] + '">'
+            });
+
+            // Build and submit the form
+            // The target="_blank" only works if the form's id attribute is also set
+            $('<form action="' + basePath + '" method="POST" target="_blank" id="downloadReportForm">' + form + "</form>")
+                .appendTo($(document.body))
+                .submit();
+        };
+
         // Email now
         this.sendReportNow = function (idReport) {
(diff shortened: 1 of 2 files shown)


plugins/ScheduledReports/templates/_listReports.twig
```diff
@@ -75,12 +75,13 @@
             </td>
             <td>
                 {# download link #}
-                <a href="{{ linkTo({'module':'API', 'segment': null, 'token_auth':token_auth,
+                <a href="#"
```
tsteur: @katebutler random thought to prevent possible XSS etc... we could have for each report a hidden form like `<form action="index.php?module=API&..."><input name=token_auth type=hidden value="..." style="display:none"></form>` and then simply submit the hidden form when someone clicks on the link? see https://stackoverflow.com/questions/18406732/using-jquery-to-submit-form-from-link-outside-of-form

This way Twig can do all of the complicated escaping and we can still use `action="{{linkTo(...)}}`

plugins/ScheduledReports/angularjs/manage-scheduled-report/manage-scheduled-report.controller.js
```diff
@@ -177,6 +177,37 @@
...
+                } else {
+                    // Everything else can be exposed to the user in the URL
+                    basePath += separator + queryStr;
+                    separator = '&';
+                }
+            });
+
+            // Build and submit the form
+            // The target="_blank" only works if the form's id attribute is also set
+            $('<form action="' + basePath + '" method="POST" target="_blank" id="downloadReportForm"></form>')
```
tsteur: be good to set basePath also through `attr()` otherwise we may run into XSS again (not sure that this would fully prevent it though)

plugins/ScheduledReports/templates/_listReports.twig
```diff
@@ -75,12 +75,19 @@
...
-                            'format': (report.format in ['html', 'csv']) ? report.format : false
-                       }) }}"
-                   rel="noreferrer noopener" target="_blank" name="linkDownloadReport" id="{{ report.idreport }}" class="link_but withIcon">
+                <form action="{{ linkTo({ 'module':'API', 'segment': null,
+                    'method':'ScheduledReports.generateReport', 'idReport':report.idreport,
+                    'outputType':downloadOutputType, 'language':language,
+                    'format': (report.format in ['html', 'csv']) ? report.format : false }) }}"
+                      method="POST"
+                      target="_blank"
+                      id="downloadReportForm_{{ report.idreport }}"
```
tsteur: haven't tested it yet but code looks good overall. It is not really needed since we know the id will be a number but just to be safe in HTML attributes we escape the output of twig variables using `html_attr` like this:

```
{{ report.idreport|e('html_attr') }}
```

instead of the standard escaping functionality (the standard escaping isn't safe for attributes). Not really needed when we know it's a number, but good to have eg if someone else copy/pastes it or changes it from an ID to something else etc.

plugins/ScheduledReports/templates/_listReports.twig
```diff
@@ -75,12 +75,19 @@
...
+                >
+                    <input type="hidden" name="token_auth" value="{{ token_auth }}">
```
tsteur: same here, in theory it will always be an MD5 and safe. but just to be extra safe be good to use `{{ token_auth|e('html_attr') }}`

plugins/ScheduledReports/templates/_listReports.twig
tsteur: re the `href="#"`... I'm not sure if it's still an issue (not tested), but I think it may cause a scroll to the top or so under circumstances (maybe always, or when before a different anchor was active).

I think we usually use `href="javascript:void(0)"` for this reason.

katebutler: Changed

plugins/ScheduledReports/templates/_listReports.twig
katebutler: Have fixed both of these

tsteur:
I'm pretty confused right now :)

For some reason I'm getting the error

> The following error just broke Matomo (v3.9.1):

> Key "name" for array with keys "module, action, metrics, metricsDocumentation, processedMetrics, imageGraphUrl, imageGraphEvolutionUrl, uniqueId" does not exist.
in plugins/ScheduledReports/templates/_addReport.twig line 148

Where it expects the report to have a `name` property. But my DB doesn't have this column and it's neither defined in the model:

https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/ScheduledReports/Model.php#L74

So wondering how this ever worked? There's something I'm not seeing. Any idea @katebutler @diosmosis ? It's not related to this PR as the PR didn't change anything. Just my scheduled reports seem to play up?

diosmosis:
I looked into the code, but I have no idea why it would work. It does work on demo.matomo.org though, so it must've been a recent regression (though the name code has been around for a long time).

diosmosis:
@tsteur Oh, I see it's not actually the scheduled report's name, but the individual report in the list of reports to include in the scheduled report (like "Visits Over Time").

tsteur:
thanks very much @diosmosis that's it! I think it's actually cohorts that doesn't have a name defined, might be an old version not sure...

> {"module":"Cohorts","action":"getByFirstDayOfVisit","metrics":{"nb_uniq_visitors":"Unique 

disabled the plugin and works now

diosmosis:
Probably Cohorts as I was affected too, I'll add it.

tsteur:
Works 👍  Will merge now @katebutler

PR #14354, opened 2019-04-17
Docker compose development environment

dshiryaev-plesk:
Hello,
See CONTRIBUTING.md for documentation.

Benefits of Docker development environment for Matomo at Matomo source code:
- quick start for Matomo developers
    - no deployment and configuration of required versions of MySQL, PHP, web server
    - no configuration of Matomo, it's already configured
    - similar approach for all projects that use Docker Compose as a development environment: setup Docker, Docker Compose once for all projects that use Docker Compose, then run docker-compose up in directory of project source code 
- Matomo developers can work simultaneously with different projects that have different system requirements or with different Matomo versions, since the system environment is in Docker images and small dump of Matomo database for Docker development environment is in the source code of Matomo
- all Matomo developers can work with the same system environment in bounds of one version Matomo, since the system environment is in Docker images


CONTRIBUTING.md
```diff
@@ -1,16 +1,160 @@
...
 
 You can help improve translations in Matomo, please read [contribute to translations](https://github.com/matomo-org/piwik/blob/master/lang/README.md).
 
+<a name="how-to-submit-code-improvements-via-pull-requests"></a>
 ## How to submit code improvements via pull requests?
 
 You can help contribute to Matomo codebase via Pull Requests, see [Contributing to Matomo core](https://developer.matomo.org/guides/contributing-to-piwik-core)
+
+
+<a name="docker-based-development-environment"></a>
```
tsteur: Can we instead maybe simply link to our official docker? https://github.com/matomo-org/docker

Findus23:
I guess this should rather be a separate project or a part of https://github.com/matomo-org/docker than in the main repository.

tsteur:
@dshiryaev-plesk I noticed you're thumbs down. I suppose we would be wanting max a link to our official docker but for sure not want to maintain another docker. There's also the problem that we would need to potentially maintain the SQL fixture over time etc and having this in this repository doesn't quite work for us as we wouldn't even really be able to support and maintain it. Might be better to put it in a standalone repository?

mattab:
@tsteur what do you reckon we do here?
@dshiryaev-plesk any thoughts?

Findus23:
I still think that this shouldn't be in the main repo as it means that it needs to be maintained together with Matomo itself. (and having an SQL file in Matomo means people take it as official and will complain once it doesn't work)
I'm proposing:
- adding it to https://github.com/matomo-org/docker as the use case overlaps with it and maintaining two independent projects is a lot of useless work
- or adding it as a new https://github.com/matomo-org/docker-compose (or something like that) repo and having it maintained by someone
- or keeping it a seperate project just like https://github.com/bitnami/bitnami-docker-matomo and https://github.com/crazy-max/docker-matomo

@J0WI What's your opinion?

tsteur:
Definitely wouldn't merge in this repo.

J0WI:
I think it's the best to just use `FROM matomo` and add whatever you need for development. We should not add xdebug to https://github.com/matomo-org/docker and definitely not on DockerHub.
A minimal `docker-compose.yml` example could also be added to https://github.com/docker-library/docs/tree/master/matomo, but the example here is far more complex than what is required to play with this container.

tsteur:
Closing this issue for now as we agreed we don't really want to have this in core. @dshiryaev-plesk maybe look at the last comments re what could be done?

PR #14355, opened 2019-04-17
Adds new ViewDataTable.configure.end event

sgiehl:
While developing a plugin I tried to add a footer message to an existing report when the report is shown flattened. I tried that using the event `ViewDataTable.configure` with following code.

```php
if ('Plugin.method' === $view->requestConfig->apiMethodToRequestDataTable && $view->requestConfig->flat) {
    $view->config->show_footer_message = $message;
}
```

I wondered why that worked when switching to the flattened view, but not after reloading the page (having the report still flattened).
The reason is, that the saved viewdata parameters are injected after the event is posted. That means that  `$view->requestConfig->flat` holds the default value `false` until then.

My expectations would have been, that the event already gets the final parameters. If that's wrong maybe we could add a new event `ViewDataTable.configure.end` or similar that works with the final parameters?

ping @tsteur @diosmosis @mattab

### DIFF
--- a/core/Plugin/ViewDataTable.php
+++ b/core/Plugin/ViewDataTable.php
@@ -239,6 +239,9 @@ abstract class ViewDataTable implements ViewInterface
          * Triggered during {@link ViewDataTable} construction. Subscribers should customize
          * the view based on the report that is being displayed.
          *
+         * This event is triggered before view configuration properties are overwritten by saved settings or request
+         * parameters. Use this to define default values.
+         *
          * Plugins that define their own reports must subscribe to this event in order to
          * specify how the Piwik UI should display the report.
          *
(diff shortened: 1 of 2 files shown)


core/Plugin/ViewDataTable.php
```diff
@@ -235,6 +235,23 @@ public function __construct($controllerAction, $apiMethodToRequestDataTable, $ov
...
+        // to avoid issuing unnecessary requests when display properties are determined by metadata,
+        // we allow it to be a closure.
+        if (isset($this->requestConfig->filter_excludelowpop_value)
+            && $this->requestConfig->filter_excludelowpop_value instanceof \Closure
+        ) {
+            $function = $this->requestConfig->filter_excludelowpop_value;
+            $this->requestConfig->filter_excludelowpop_value = $function();
+        }
+
+        $this->overrideViewPropertiesWithParams($overrideParams);
```
tsteur: I'm pretty sure those two override methods should be executed after the event as otherwise the visualisation may overwrite settings that are supposed to be set by `overrideViewPropertiesWithParams ` etc.

I think the other ones should make sense to execute them before the event but haven't looked if there are any regressions 👍

sgiehl: But that way it won't be possible to receive the final parameters withing the event. Or shouldn't something like my example above be possible

tsteur: Do we need maybe a `configure.end` event instead? I reckon that makes more sense maybe

sgiehl: That would work as well. Will change the PR tmrrw

sgiehl:
@tsteur added a new event `ViewDataTable.configure.end` instead of moving the other one

PR #14358, opened 2019-04-18
Group referrer URLs with different protocols together

### LINKED ISSUE #13882: Referrer Websites report should group same label (HTTP vs. HTTPS)

katebutler:
Fixes #13882


plugins/Referrers/tests/System/ApiTest.php
```diff
@@ -209,6 +209,69 @@ public function test_forceNewVisit_shouldNotForceNewVisitWhenReferrerNameIsLonge
...
+
+        $visits = Request::processRequest(
+            'Referrers.getUrlsFromWebsiteId',
+            array('idSite' => $idSite, 'period' => 'day', 'date' => $dateTime, 'idSubtable' => $idSubtable)
+        );
+
+        $firstRow = $visits->getFirstRow();
+        $this->assertEquals('index', $firstRow->getColumn('label'));
+        $this->assertEquals(2, $visits->getFirstRow()->getColumn('nb_visits'));
+    }
```
diosmosis: 👍 for system tests like these, should probably do more like these

PR #14361, opened 2019-04-18
Fix typo in "Asking for Consent" JS code

mattab:
as reported by a user


tsteur (approved):
UI test will probably need to be updated. Otherwise 👍

PR #14363, opened 2019-04-22
Allow limit selection contents to be customized by specific visualization.

diosmosis:
For a plugin.


tsteur:
haven't tested it but looks fine 👍

PR #14368, opened 2019-04-23
Fix missing 3rd party cookie when using POST requests

MichaelRoosz:
Added "xhr.withCredentials = true;" so that the third party cookie is included in POST requests.
Currently cookies are missing when using POST (xhr) requests.
This fixes visitor duplication when POST gets used.

This is the same change as in https://github.com/matomo-org/matomo/pull/13159 , but this time I have included a new test.


tsteur:
Cheers @MichaelHeerklotz  noticed a test fails on travis: https://travis-ci.org/matomo-org/matomo/jobs/523527419#L776-L776

Just fyi: Not every older (IE) browser supports `withCredentials` but that shouldn't be a problem (source https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials). It won't break anything.

tsteur:
@diosmosis seems there are still some JSlint errors. If the tests pass, we could merge this in 3.10 I suppose.

tsteur:
Merging for now, will need to fix the failing tests afterwards (newline missing at the end)

PR #14369, opened 2019-04-23
Allow using image link in reporting menu if icon is not in font set.

diosmosis:


core/Category/Category.php
```diff
@@ -43,6 +43,12 @@ class Category
      */
     protected $icon = '';
 
+    /**
+     * A link to an image to use for this category, if the desired icon does not exist.
+     * @var string|null
+     */
+    protected $imageUrl = null;
```
tsteur: btw we could include icon in matomo core or load a separate font for that icon (i think we used to do that in a/b testing in the beginning)

diosmosis: This PR seemed a bit less time consuming, but if this is something we don't want to support I can go an alternate route.

tsteur: Be nice to always use icons as they have higher res, color can be changed easily etc. It might be even easier to just add another icon to core? or embed a separate font?

diosmosis: What's the process for adding another icon? A readme would be helpful (I think it involves a PR to the icons repo, but not sure how the font is built.)

tsteur: it's described here: https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/Morpheus/fonts/README.md basically you upload the `selection.json`, add new icons that you want, and export the font again. The updated font and selection.json is then put into the repo

PR #14372, opened 2019-04-25
Ensure segment definitions don't regress

### LINKED ISSUE #14364: Readable segment values for browser causes regression in custom reports
See https://github.com/matomo-org/matomo/pull/13929/files#diff-9b36d5d506002d73975c4f8486780ce5R23

The existing dimension should still use eg `browserCode` otherwise it will generates these segments that won't work and broke the API:
image

Adding `not-in-changelog` as the regression wasn't released. Ideally, `browserCode` is still defined in that dimension, and `browserName` in a new dimension.

fyi @sgiehl

sgiehl:
fixes #14364

### DIFF
--- a/plugins/DevicesDetection/API.php
+++ b/plugins/DevicesDetection/API.php
@@ -255,7 +255,7 @@ class API extends \Piwik\Plugin\API
     public function getBrowsers($idSite, $period, $date, $segment = false)
     {
         $dataTable = $this->getDataTable('DevicesDetection_browsers', $idSite, $period, $date, $segment);
-        $dataTable->filter('AddSegmentByLabel', ['browserCode']);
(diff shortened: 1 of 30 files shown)


sgiehl:
@tsteur could you check if the changes are fixing the regression in CustomReports?

tsteur:
Works @sgiehl 👍 tests now succeed in custom reports

tsteur:
Only tests need to be updated, eg https://travis-ci.org/matomo-org/matomo/jobs/524659395#L1539

sgiehl:
rebased the branch on latest changes to `3.x-dev`. Tests are passing now... Guess should be good to merge?

tsteur:
Yes 👍

PR #14380, opened 2019-04-29
Copy geoip archive into temporary folder instead of in the geoip folder

tsteur:
I noticed the geoip tar.gz archive is downloaded  and extracted into the `misc` directory or whatever is configured eg through ` 'path.geoip2' => DI\string('{path.root}/misc/geoip/'),`. However, only the `misc/GeoLite2-City.mmdb` might have write access but not any other file. This causes the updating of the geodb to fail. Instead, it should be extracted into the `tmp/latest` folder which we use for such cases (eg matomo update, plugin update, ...)

### DIFF
--- a/plugins/GeoIp2/GeoIP2AutoUpdater.php
+++ b/plugins/GeoIp2/GeoIP2AutoUpdater.php
@@ -129,7 +129,7 @@ class GeoIP2AutoUpdater extends Task
         // NOTE: using the first item in $dbNames[$dbType] makes sure GeoLiteCity will be renamed to GeoIPCity
         $zippedFilename = LocationProviderGeoIp2::$dbNames[$dbType][0] . '.' . $ext;
 
-        $zippedOutputPath = LocationProviderGeoIp2::getPathForGeoIpDatabase($zippedFilename);
+        $zippedOutputPath = self::getTemporaryFolder($zippedFilename);
 
         $url = self::removeDateFromUrl($url);
 
@@ -156,6 +156,11 @@ class GeoIP2AutoUpdater extends Task
         }
 
         Log::info("GeoIP2AutoUpdater: successfully updated GeoIP 2 database '%s'", $url);
+    }
+
+    protected static function getTemporaryFolder($file)
+    {
+        return \Piwik\Container\StaticContainer::get('path.tmp') . '/latest/' . $file;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -235,7 +240,7 @@ public static function unzipDownloadedFile($path, $dbType, $unlink = false)
                 'loc' => array(),
                 'isp' => array()
             );
-            $customDbNames[$dbType] = array($tempFilename);
+            $customDbNames[$dbType] = $outputPath;
```
tsteur: don't know how this worked before... had to remove the array as in the PHP provider it accesses that key directly.

plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -255,10 +260,10 @@ public static function unzipDownloadedFile($path, $dbType, $unlink = false)
             // delete the existing GeoIP database (if any) and rename the downloaded file
             $oldDbFile = LocationProviderGeoIp2::getPathForGeoIpDatabase($dbFilename);
             if (file_exists($oldDbFile)) {
-                unlink($oldDbFile);
+                @unlink($oldDbFile);
```
tsteur: we already ignored below `@rename //In case the $tempfile cannot be renamed, we copy the file.` but not here the unlink which also may fail but then the `copy` succeeds.

diosmosis (approved):
Code looks good, not sure if @sgiehl wants to take a look as well

plugins/GeoIp2/LocationProvider/GeoIp2/Php.php
```diff
@@ -320,7 +320,15 @@ public function getInfo()
     private function getGeoIpInstance($key)
     {
         if (empty($this->readerCache[$key])) {
-            $pathToDb = self::getPathToGeoIpDatabase($this->customDbNames[$key]);
+            if (is_string($this->customDbNames[$key])
+                && strpos( $this->customDbNames[$key], '/') !== false
+                && @file_exists($this->customDbNames[$key])) {
+                // for performance reason we only check for file exists if the string contains a slash and is not only a
+                // filename
+                $pathToDb = $this->customDbNames[$key];
```
sgiehl: Wondering if it might make sense to move that logic to `GeoIp2::getPathForGeoIpDatabase`. Not sure if there might be use cases that would need to set an array of custom database paths for one type at any time... Otherwise looks good.

tsteur: good point @sgiehl changed it. tested it and worked for me.

sgiehl (approved):

PR #14383, opened 2019-04-29
Issue: 2nd ?/& appended to cross domain link containing anchor

danlance:
### Description
When cross domain linking is being used, and tracked cross domain URL contains an anchor (#), an additional question mark (?) / ampersand (&) symbol is appended to the end of the destination URL.

This creates an invalid link, causing the anchor to fail, and can potentially cause the link to be processed incorrectly by the destination web server.

### Conditions:
- Cross Domain linking is enabled for tracking links within same Matomo site, but on a different domain
- Link href includes an anchor reference (#)
### Example Destination Links
1. https://examplesite/#anchor
1. https://examplesite/?existingQueryStringParam=value#anchor
1. https://examplesite/
1. https://examplesite/?existingQueryStringParam=value
### Expected Results:
1. https://examplesite/?pk_vid=ExampleVisitorID#anchor
1. https://examplesite/?existingQueryStringParam&pk_vid=ExampleVisitorID=value#anchor
1. https://examplesite/?pk_vid=ExampleVisitorID
1. https://examplesite/?existingQueryStringParam=value&pk_vid=ExampleVisitorID
### Actual Results:
1. https://examplesite/?pk_vid=ExampleVisitorID#anchor?
1. https://examplesite/?existingQueryStringParam&pk_vid=ExampleVisitorID=value#anchor&
1. https://examplesite/?pk_vid=ExampleVisitorID
1. https://examplesite/?existingQueryStringParam=value&pk_vid=ExampleVisitorID
### Cause:
- replaceHrefForCrossDomainLink within piwik.js appends an &/? to the whole linkURL, without taking into account potential of anchors, causing symbol to be anded to end of URL after any anchors
- addUrlParameter function within piwik.js which is called subsequently then correctly adds ?/& to the baseURL - and completely ignores that symbol has already been added to the full link, leaving the symbol within the urlHash which is subsequently appended to the baseURL
### Fix:
- Remove appending of ?/& from replaceHrefForCrossDomainLink function completely, as this functionality is already implemented within addUrlParameter function which is always called subsequently - and hence this functionality duplicated unnecessarily, as well as causing the described issue due to the differences in implementation.

This change is included in the pull request, as well as minifying the changed js using YUICompressor 2.4.8. as per the instructions contained within the ReadMe. (Please note, this was performed on Mac OSX not Linux)


js/piwik.js
```diff
@@ -5264,12 +5264,6 @@ if (typeof window.Piwik !== 'object') {
                 // and visitorId (eg userId might be set etc)
                 link = removeUrlParameter(link, configVisitorIdUrlParameter);
 
-                if (link.indexOf('?') > 0) {
```
tsteur: would you still need to do this logic when there is no hash in the URL?

danlance: No - the logic within the **addUrlParameter** function duplicates the removed logic, but applies to the **baseURL** as opposed to the full **url**:
```
            if (baseUrl.indexOf('?') === -1) {
                baseUrl += '?';
            } else if (!stringEndsWith(baseUrl, '?')) {
                baseUrl += '&';
            }
            // nothing to if ends with ?
```
The **baseURL** variable will be the same as the **url** variable in the case that there are no hashes in the url.

As the **addUrlParameter** function is always called from the **replaceHrefForCrossDomainLink** directly after the removed logic, then the inclusion of this logic is superfluous in all cases. 

I have tested the 4 use cases identified under **Example Destination Links** in the description above, to ensure that they are processed correctly to generate the **Expected Results**

tsteur: makes sense 👍

tsteur:
Looks good to merge but need to find time to merge as we need to update the piwik.js/matomo.js later as it currently fails due to some line break (https://travis-ci.org/matomo-org/matomo/jobs/526012825#L966) ... happens when generating the minified version eg on Mac compared to Linux. Bit annoying and need to tweak the test at some point to ignore this line break at the end.

tsteur (approved):

mattab:
Thank you for the report & the detailed PR @danlance - Appreciated! :+1:

PR #14386, opened 2019-04-29
Make sure evolution parameters are initialized in Config so Controllers can override them.

diosmosis:


tsteur:
Haven't tested but reckon it's fine to merge.

PR #14387, opened 2019-04-29
Require password confirmation before installing plugins

### LINKED ISSUE #13581: Before installing a new plugin (via Marketplace or direct upload), ask again Super User password
Before installing a new plugin (via Marketplace or direct upload), we should ask again the Super User to enter their password. Since a plugin can easily lead to RCE it is important to ensure that a Logged-in browser cannot be used to install custom plugins.

Similar to #2932

katebutler:
FIxes #13581


plugins/CorePluginsAdmin/Controller.php
```diff
@@ -87,6 +99,13 @@ public function __construct(Translator $translator, Plugin\SettingsProvider $set
...
     {
+        if (!$this->passwordVerify->isPasswordCorrect(
+            Piwik::getCurrentUserLogin(),
+            Common::getRequestVar('confirmPassword')
+        )) {
+            throw new \Exception($this->translator->translate('Login_LoginPasswordNotCorrect'));
+        }
+
         static::dieIfPluginsAdminIsDisabled();
         Piwik::checkUserHasSuperUserAccess();
```
tsteur: we need to make sure to first perform several checks before performing the password check. Meaning we need to make sure
* admin is not disabled
* user has super user access. 
* Plugin upload is enabled
* Nonce check is correct

Otherwise could be potentially misused for brute forcing password etc. (it would actually not really be possible) but it's a lot safer to do the password check only after a nonce etc. Could even check it only shortly before `$this->configureView('@CorePluginsAdmin/uploadPlugin');` so we make sure the basic things are correct.

Works otherwise, we only need to make sure to move the check further down

plugins/CorePluginsAdmin/Controller.php
```diff
@@ -87,6 +99,13 @@ public function __construct(Translator $translator, Plugin\SettingsProvider $set
 
     public function uploadPlugin()
     {
+        if (!$this->passwordVerify->isPasswordCorrect(
+            Piwik::getCurrentUserLogin(),
+            Common::getRequestVar('confirmPassword')
```
tsteur: probably doesn't make a difference but might be good to set `'confirmPassword', null, 'string'` to make sure we are getting a string and not something else (eg could be maybe an array not sure)

tsteur (changes requested):
@katebutler 👍  worked. Left 2 minor comments that be changed quickly, then we can merge :)

plugins/CorePluginsAdmin/Controller.php
katebutler: Fair enough, have moved it to just before validation of the uploaded file

plugins/CorePluginsAdmin/Controller.php
katebutler: OK, changed

PR #14399, opened 2019-05-02
Invalidate archives when visits removed via GDPR tool

### LINKED ISSUE #14195: When deleting visits via the GDPR Tools, invalidate the reports for each day of these deleted visits
This is a frequent bug report. When [deleting visits via the GDPR Tools](https://matomo.org/docs/gdpr/) then the deleted visits still show up in the reports. 

The goal of this issue is that, whenever some visits are deleted then we also should call the [Invalidate Reports](http://piwik.org/faq/how-to/faq_155/) code. As a result, next time `core:archive` will run all of the reports for the invalidated days (and the week, month, year) will be reprocessed, and the deleted data will be reflected in the reports (ie. not visible anymore). 

(this will also benefit Cloud customers since invalidating old reports is disabled on Cloud yet sometimes we really need some data to be deleted)

Also in the UI we can remove the text `In this case you may want to consider to invalidate reports after deleting these visits. ` since it will now be done automatically.

katebutler:
Fixes #14195


plugins/PrivacyManager/Model/DataSubjects.php
```diff
@@ -114,6 +119,51 @@ public function deleteDataSubjects($visits)
...
+            $idVisitsByIdSites[$idSite][] = (int)$visit['idvisit'];
+        }
+
+        $datesToInvalidate = array();
+        foreach ($idVisitsByIdSites as $idSite => $idVisits) {
+            $website = Cache::getCacheWebsiteAttributes($idSite);
+            $timezone = $website['timezone'];
+
+            $sql = 'SELECT visit_last_action_time FROM '
+                . Common::prefixTable('log_visit') . ' WHERE idsite = ' . $idSite
```
tsteur: @katebutler be good to cast `idSite` to `(int)` or use bind parameters just to be double safe. Can basically never be an issue here since we have checked above it is an idSite but better to be safe :)

tsteur (changes requested):
quickly mentioned a very very minor tweak. Haven't tested it yet but code looked good so far otherwise

plugins/PrivacyManager/Model/DataSubjects.php
tsteur: I'll delete this comment, it's already an integer :)

plugins/PrivacyManager/Model/DataSubjects.php
```diff
@@ -114,6 +119,51 @@ public function deleteDataSubjects($visits)
...
+            $idSite = (int)$visit['idsite'];
+            if (!isset($idVisitsByIdSites[$idSite])) {
+                $idVisitsByIdSites[$idSite] = array();
+            }
+            $idVisitsByIdSites[$idSite][] = (int)$visit['idvisit'];
+        }
+
+        $datesToInvalidate = array();
+        foreach ($idVisitsByIdSites as $idSite => $idVisits) {
+            $website = Cache::getCacheWebsiteAttributes($idSite);
```
tsteur: FYI here you can use `\Piwik\Site::getTimezoneFor($idSite)`.  We use `Cache::getCacheWebsiteAttributes` only in tracker mode (it's tracker cache).

katebutler: Thanks, changed.

tsteur:
Awesome, tested it and worked 👍 looks all good. Was just going to merge and then noticed we need to fix some tests:

* In https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/33885/ we need to update the expected screenshots that start with `PrivacyManager_gdpr_tools_*`
* https://travis-ci.org/matomo-org/matomo/jobs/527154664#L1578 is now failing for some reason. Seems some date is different: 
image

image

I can show you how to sync the expected test results from travis if I haven't done yet later.

tsteur:
Cheers for updating the tests @katebutler 👍  will merge now

PR #14400, opened 2019-05-02
Fix CSRF vulnerability in opt-out when setCookieInNewWindow=1

### LINKED ISSUE #12540: Opt-out does not unset session ID for CSRF protection
When using the form to opt-out of tracking (tested on https://matomo.org/privacy-policy/):

screen shot 2018-02-06 at 17 25 21

## Expected behaviour

The `PIWIK_SESSID` cookie is removed from my browser.

## Actual behaviour

The `PIWIK_SESSID` cookie remains with its previous value.

## Discussion

We have tested this on our own install and the opt out mechanism works, in the fact that this user's sessions do not appear in the dashboard. However, there are two primary concerns:
 
 1. If the cookie remains, there is no guarantee tracking will not continue - one must trust that the server side is honoring the `piwik_ignore` cookie. The Privacy by Design approach would be to unset the session cookie, then every user can be sure that the identifier is no longer present.
 2. From a compliance perspective (considering upcoming GDPR and ePrivacy regulations), if a user has explicitly opted-out of tracking, yet the identifier is still being sent by the client, I must be able to be 100% sure that this identifier is not being stored somewhere. On our installation I could see that a session ID for a user who was opted out from the start is still stored in the database. How can I verify that there are no other traces of this user in the database which could lead this to violate GDPR? Similarly any logs upstream on the webserver may need to be purged of this identifier. 

Therefore, it would be preferable if the server would remove/reset the `PIWIK_SESSID` cookie on opt-out, and enforce that this cookie does not persist. This would meet the goal on both sides: that the user can trust the opt-out mechanism, and that the administrator can prove compliance with user consent.

katebutler:
Fixes #12540

### DIFF
--- a/plugins/CoreAdminHome/OptOutManager.php
+++ b/plugins/CoreAdminHome/OptOutManager.php
@@ -175,6 +175,7 @@ class OptOutManager
             $reloadUrl = Url::getCurrentQueryStringWithParametersModified(array(
                 'showConfirmOnly' => 1,
                 'setCookieInNewWindow' => 0,
+                'nonce' => Common::getRequestVar('nonce')
             ));
         } else {
             $reloadUrl = false;
--- a/plugins/CoreAdminHome/OptOutManager.php
+++ b/plugins/CoreAdminHome/OptOutManager.php
@@ -180,8 +180,8 @@ class OptOutManager
         } else {
             $reloadUrl = false;
 
-            $nonce = Common::getRequestVar('nonce', false);
-            if ($nonce !== false && Nonce::verifyNonce('Piwik_OptOut', $nonce)) {
+            $requestNonce = Common::getRequestVar('nonce', false);
+            if ($requestNonce !== false && Nonce::verifyNonce('Piwik_OptOut', $requestNonce)) {
                 Nonce::discardNonce('Piwik_OptOut');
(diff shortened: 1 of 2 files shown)


Findus23:
Just FYI: As this only solves the CSRF-part of #12540 and not the unneeded-cookie-part I have created #14402 as a followup for that issue.

tsteur:
image

just tested it and actually didn't work for me. Instead seeing above error in chrome. 

In https://github.com/matomo-org/matomo/blob/3.10.0-b1/plugins/CoreAdminHome/javascripts/optOut.js#L6-L7 it actually does a GET request to the form action. You would think the form does a regular post but it opens the `<form action="">` URL in a new window totally ignoring the hidden `nonce` input field etc. 

That's why previously the `nonce` was before added in the redirect URL and caused the CSRF problem. 

To fix it, we need to add the nonce to the form `action url` in https://github.com/matomo-org/matomo/blob/3.10.0-b1/plugins/CoreAdminHome/OptOutManager.php#L195-L200 . It should be totally fine to send the nonce through GET parameter.

tsteur:
Tested using the URL: https://$DOMAIN/index.php?module=CoreAdminHome&action=optOut&idSite=1&period=day&date=yesterday

tsteur:
Works 👍

PR #14403, opened 2019-05-02
CronArchive: Catch UnexpectedWebsiteFoundException

### LINKED ISSUE #14393: Add option to core:archive that allows you to continue even if the site is not in matomo anymore
Hi, could a option be added that allows you to continue even if you get a error like:

  [Piwik\Exception\UnexpectedWebsiteFoundException]                               
  An unexpected website was found in the request: website id was set to '2897' .  

please?

This would be super useful.

Or is there a way to clear the deleted websites so they are not run?

paladox:
This is to allow the archive to continue if the site was deleted from matomo.

Fixes #14393

### DIFF
--- a/core/CronArchive.php
+++ b/core/CronArchive.php
@@ -474,7 +474,6 @@ class CronArchive
                         }
                     } catch (UnexpectedWebsiteFoundException $e) {
                         $this->logger->info("Skipped website id $idSite, got: UnexpectedWebsiteFoundException");
-                        $this->skipped++;
                         continue;
                     }
 
--- a/core/CronArchive.php
+++ b/core/CronArchive.php
@@ -459,17 +459,23 @@ class CronArchive
                 // to process the archives for this website (only if there were visits since midnight)
                 if (!$hasWebsiteDayFinishedSinceLastRun && !$isOldReportInvalidatedForWebsite) {
 
-                    if ($this->isWebsiteUsingTheTracker($idSite)) {
+                    try {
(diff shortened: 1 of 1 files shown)


core/CronArchive.php
```diff
@@ -716,7 +722,13 @@ private function archiveSingleSite($idSite, $requestsBefore)
         /**
          * Trigger archiving for non-day periods
          */
-        $success = $this->processArchiveForPeriods($idSite, $lastTimestampWebsiteProcessedPeriods);
+        try {
+            $success = $this->processArchiveForPeriods($idSite, $lastTimestampWebsiteProcessedPeriods);
+        } catch (UnexpectedWebsiteFoundException $e) {
+            // this website was deleted in the meantime
```
tsteur: do we need a `$this->skipped++;` here too?

tsteur (changes requested):
left a minor comment but otherwise looked and worked 👍

core/CronArchive.php
paladox: I doin't think so (after looking through), this is only used for the logger to tell how many sites were skipped.

tsteur: yeah just thought because we kind of skip the website here... at least the "log info" says "skipped" so assumed it might be needed but indeed may not be needed

paladox:
Thank you! (would this make it into matomo 3.10?):)

tsteur:
Yes 👍  Cheers for the PR

paladox:
Thanks!!!

PR #14405, opened 2019-05-02
Do not create request to validate license in tracker mode

### LINKED ISSUE #14401: DoS when using Premium Plugins caused by deleteTrackerCache() and getLicenseValidInfo()
This is a pretty nasty denial of service bug:

Some actions, for example:
- adding a new user
- updating a user or user rights
- deleting a user
- changing global site settings
- and more

will call core\Tracker\Cache->deleteTrackerCache(), which deletes the whole cache.
This causes core\Plugin\Manager->getLicenseValidInfo() to get called for every activated premium plugin on the next tracking request or admin panel request.
getLicenseValidInfo() will send a http request to the matomo marketplace server.

On big setups like mine (>8000 tracking requests per minute) there are 350+ php workers running at the same time, many of them will call getLicenseValidInfo() roughly at the same time after a deleteTrackerCache() cache.

This means: About every 4th time I edit a users permissions or add a new site, the API request doing this admin operation will take over 60 seconds to complete, if it completes at all.
And during and after this, all 350+ php workers will get stuck, all trying to send a http request to the matomo marketplace. Bascially thousands of http request at the same time.

Our load balancer will then report for about 60s with "502 Bad Gateway" because all web servers are stuck on sending thousands of http requests to the matomo marketplace.

We have about 12 premium plugins activated, and about 700 php workers, this means up to 700*12 = 8400 http requests at the same time caused by a single admin action. These will possibly add up if editing multiple users in a short time period.

During this 60+ seconds we lose thousands of tracking requests.

I think an easy fix for now would be to introduce a new "plugin cache",  that is independent from the normal "cache" and not affected by  deleteTrackerCache(),

Alternative fix: do not call getLicenseValidInfo() during tracking requests.

I can try to create a PR for this, but I would like to know how you want to proceed.

tsteur:
fix https://github.com/matomo-org/matomo/issues/14401

### DIFF
--- a/core/Plugin/Manager.php
+++ b/core/Plugin/Manager.php
@@ -30,6 +30,7 @@ use Piwik\Plugin\Dimension\ConversionDimension;
 use Piwik\Plugin\Dimension\VisitDimension;
 use Piwik\Settings\Storage as SettingsStorage;
 use Piwik\SettingsPiwik;
+use Piwik\SettingsServer;
 use Piwik\Theme;
 use Piwik\Translation\Translator;
 use Piwik\Updater;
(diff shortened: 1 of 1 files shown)


tsteur:
FYI: User confirmed the fix works.

tsteur:
Will merge as user confirms it fixes the issue.

PR #14410, opened 2019-05-04
Enable scrolling in Widgetize report iframes

### LINKED ISSUE #14056: Embed report iframe markup at scrolling=yes
see https://github.com/matomo-org/matomo/issues/10460#issuecomment-458926213

In Widgets listing  page when we generate the iframe HTML that users can copy maybe we could add this attribute to the iframe. To be tested

courtneywright:
Tested on Chromium and Firefox on Ubuntu and MS Edge.

Fixes #14056


tsteur:
Cheers 👍

PR #14412, opened 2019-05-05
Don't tell people to run chmod -R 755 on their install if update fails

### LINKED ISSUE #13628: Don't tell people to run chmod -R 755 on their install if update fails
Related to #13626 

Please don't tell people to run `chmod -R 755 /var/www/piwik`. This will make all the files executable by everyone.

Either use `chmod -R 754 /var/www/piwik` (the easy solution) or `find /var/www/piwik -type d -exec chmod 755 {} \;` (the good solution).

tsteur:
fix #13628

For not writable directory it generated eg this:
```
chown -R foo:localaccounts /var/www/matomo
find /var/www/matomo/tmp -type f -exec chmod 644 {} \;
find /var/www/matomo/tmp -type d -exec chmod 755 {} \;
```

for auto update it also adds execute permission for `console`

```
chmod 755 /var/www/matomo/console
```

Not sure if any other file needs permission to execute? Couldn't think of any right now. Of course the find commands above they don't change the permission of any symlinks or so. Not sure what users possibly have in there and if it could create some issue. Could extend the find command to something like `-type f or -type l`.

Also generally not sure if `find` is compatible like this on all *nix.

@Findus23 maybe any thoughts?

### DIFF
--- a/core/Filechecks.php
+++ b/core/Filechecks.php
@@ -111,8 +111,9 @@ class Filechecks
     {
         $realpath = Filesystem::realpath(PIWIK_INCLUDE_PATH . '/');
         $message = '';
-        $message .= "<code>" . self::getCommandToChangeOwnerOfPiwikFiles() . "</code><br />";
-        $message .= "<code>chmod -R 0755 " . $realpath . "</code><br />";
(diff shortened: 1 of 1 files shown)


fdellwing:
> Also generally not sure if find is compatible like this on all *nix.

I can't say anything about all *nix, but find will work on every Posix compliant (and compatibel) systems.

PR #14414, opened 2019-05-06
Allow Metric instances to be stored in Report::$metrics.

diosmosis:


tsteur:
I reckon should be fine if tests pass 👍 (didn't test, only looked at code)

PR #14415, opened 2019-05-06
Real-time report under Visitors does not refresh

### LINKED ISSUE #14374: Real-time report under Visitors does not refresh
The Visitors > Real-time report does not automatically refresh every few seconds (as we'd expect). When the page loads the following JS error is triggered early on:

> Error: cannot call methods on liveWidget prior to initialization; attempted to call method 'started'

The Real-time report is displayed correctly once but does not refresh then.
Also, when clicking on the Pause/Play button at the bottom of the page, it triggers slightly different JS errors:

> Error: cannot call methods on liveWidget prior to initialization; attempted to call method 'stop' index.php:4:1813
> Error: cannot call methods on liveWidget prior to initialization; attempted to call method 'start' index.php:4:1813

tsteur:
fixes #14374

### DIFF
--- a/plugins/Live/angularjs/live-widget-refresh/live-widget-refresh.directive.js
+++ b/plugins/Live/angularjs/live-widget-refresh/live-widget-refresh.directive.js
@@ -1,0 +1,64 @@
+/*!
+ * Piwik - free/libre analytics platform
+ *
+ * @link http://piwik.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ */
+
+/**
+ * Usage:
+ * <div piwik-live-widget-refresh>
+ */
+(function () {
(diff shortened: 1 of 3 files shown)


plugins/Live/angularjs/live-widget-refresh/live-widget-refresh.directive.js
```diff
@@ -0,0 +1,64 @@
+/*!
+ * Piwik - free/libre analytics platform
```
katebutler: Should this be updated to Matomo branding?

tsteur: we'll change this at some stage on all files in a new PR. so far we kept it for existing files.

PR #14416, opened 2019-05-06
Replace internal API requests with processed ones

sgiehl:
In order to only manipulate the subtables of e.g. `Referrers.getAll` report it's required to process them and not call the API directly.

refs DEV-1603


plugins/Referrers/API.php
```diff
@@ -143,7 +144,14 @@ public function getAll($idSite, $period, $date, $segment = false)
...
 
         $this->checkSingleSite($idSite, 'getAll');
-        $dataTable = $this->getReferrerType($idSite, $period, $date, $segment, $typeReferrer = false, $idSubtable = false, $expanded = true);
+        $dataTable = Request::processRequest('Referrers.getReferrerType', [
+            'idSite' => $idSite,
+            'period' => $period,
+            'date' => $date,
+            'segment' => $segment,
+            'expanded' => true,
+            'disable_queued_filters' => true
```
tsteur: does the generic filter also need to be disabled? or is no generic filter applied cause no filter is set? Looks otherwise good to merge if all the tests pass

sgiehl: actually there shouldn't be any applied. but shouldn't hurt to add the param nevertheless

tsteur:
Looks good if tests pass

PR #14421, opened 2019-05-06
Switch UI tests from phantomjs to chrome headless

tsteur:
Installing chrome headless:

```
$ cd /path/to/matomo/tests/lib/screenshot-testing
$ npm install
```

Then execute `tests:run-ui` as usual.

### Chromium Quirks

Generally to write new tests or convert existing tests, you can look at some of the already converted tests and copy the format. It's pretty straightforward and shouldn't take too long to figure out.

Note: the comparison threshold feature is still there, but it no longer uses resemblejs since that library can't be used synchronously (well, not easily). There's an example of the use of comparison threshold in UsersManager_spec.js.

Chromium does, however, have some quirks it helps to be aware of, including:

1) if inputs are focused when taking screenshots, tests will fail randomly since the cursor can appear in the screenshot
2) taking screenshots of modals is especially annoying as they will fail randomly in many cases. to reduce the randomness I've found it works to add wait times to make sure animations/css transitions complete (though we could potentially turn all transitions off during UI tests w/ `!important`), to wait for elements in the modal to become visible, and to experiment between `const e = await page.$('.modal'); await e.screenshot();` vs `await page.screenshotSelector('.modal');` (sometimes one will work, sometimes the other; usually the first format should be favored since it depends on puppeteer's code and not our own).
3) since chrome is pretty fast, you'll have to sometimes do multiple waits, especially on page loads/reloads. For example, when loading the main page w/ dashboard, we `await page.waitForNetworkIdle();` to make sure as many http requests are completed at first (so we don't take a screenshot before the page load occurs if there's a widget on the screen), then we `await page.waitFor('.widget', { visible: true });` to make sure a widget is visible (otherwise the screenshot could be taken before the widget HTTP requests are sent), then wait we `await page.waitForNetworkIdle();` because not all of the widgets may have completed loading. It takes a bit of experimentation for each test, but once figured out the tests seem to pass predictably.


.travis.yml
```diff
@@ -87,16 +53,19 @@ before_install:
 install:
   - git fetch -q
 
-  - export GENERATE_TRAVIS_YML_COMMAND="php ./tests/travis/generator/main.php generate:travis-yml --core --verbose"
-  - '[[ "$TRAVIS_JOB_NUMBER" != *.1 || "$TRAVIS_PULL_REQUEST" != "false" ]] || ./tests/travis/autoupdate_travis_yml.sh'
+  #- export GENERATE_TRAVIS_YML_COMMAND="php ./tests/travis/generator/main.php generate:travis-yml --core --verbose"
```
tsteur: we need to remember to uncomment later

plugins/API/templates/glossary.twig
```diff
@@ -43,7 +43,8 @@
                             {% endif %}
                                 <h3 style="color:#4183C4;font-weight: bold;">{{ entry.name }}</h3>
                                 {% if entry.subtitle|default is not empty %}<p style="color:#999;text-transform:uppercase;font-weight:normal;margin-top:-16px;">{{ entry.subtitle|translate }}</p>{% endif %}
-                                <p>{{ entry.documentation|raw }}
+                                <p>
+                                    <span ng-bind-html="{{ entry.documentation|json_encode|e('html_attr') }}"/>
```
tsteur: just need to double check things are still shown correctly

plugins/CoreAdminHome/API.php
```diff
@@ -225,6 +226,8 @@ public function getTrackingFailures()
     {
         if (Piwik::hasUserSuperUserAccess()) {
             $failures = $this->trackingFailures->getAllFailures();
+            file_put_contents(PIWIK_INCLUDE_PATH . '/test.out', "failures 1\n", FILE_APPEND);
```
tsteur: this debug code needs to be removed

plugins/Login/templates/login.twig
```diff
@@ -48,7 +48,7 @@
                             <input name="form_rememberme" type="checkbox" id="login_form_rememberme" value="1" tabindex="90"
                                    {% if form_data.form_rememberme.value %}checked="checked" {% endif %}/>
                             <label for="login_form_rememberme">{{ 'Login_RememberMe'|translate }}</label>
-                            <input class="submit btn" id='login_form_submit' type="submit" value="{{ 'Login_LogIn'|translate }}"
+                            <input class="submit btn" id="login_form_submit" type="submit" value="{{ 'Login_LogIn'|translate }}"
```
tsteur: no change needed just curious here as well... did the quote cause any issue? (just in case we need to pay attention to it in the future).

diosmosis: No issue, and I don't even remember changing this file... :)

plugins/API/templates/glossary.twig
diosmosis: Screenshot test seems to work, though let me know if there's an issue somewhere

mattab:
Nice work! @diosmosis @sgiehl  :rocket:

PR #14423, opened 2019-05-07
Changed period in row evolution breaks graph/metric association

### LINKED ISSUE #14208: Changed period in row evolution breaks graph/metric association

tsteur:
fix #14208

worked for me, also with multi row evolution. Only known issue I found is if requesting eg first yearly row evolution which may have the `users` metric hidden, then switching to `day` period, and then the `users` metric is maybe supposed to become available but didn't. That should be quite edge case though.


diosmosis:
Code looks good, but some of the UI tests are failing now (some failures seem to be legitimate).

PR #14424, opened 2019-05-07
Display error message after encountering ajax error in popover

### LINKED ISSUE #13951: When PHP or Matomo issues a fatal error (for any reason), and the page is in a popover, show error message to user instead of "forever loading"
It would be valuable for users that when PHP or Matomo fails and return fatal error, that we somehow manage to show this error message in the popovers as well... 

To reproduce:
* Open a report with row evolution feature
* Introduce a fatal error in the code (or lower memory setting to cause PHP to fail)
* Click on Row Evolution icon
* Got: the popover for Row Evolution shows "Loading Row Evolution..." forever, and likely a JavaScript error in the browser console such as `TypeError: self.datatable is null`
* Expected instead: The popover for Row Evolution should show the error message in red  like [this](https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/Marketplace/tests/UI/expected-screenshots/Marketplace_notification_plugincheck_noLicense.png) and ask to contact support/ask in forums (we have links to Forums/Support in the fatal error notifications already in Matomo so be great to reuse those)

katebutler:
Fixes #13951


plugins/Morpheus/javascripts/ajaxHelper.js
```diff
@@ -279,11 +279,24 @@ function ajaxHelper() {
...
+                $(document.body).html(deferred.responseText);
+            } else {
+                // Just display a generic error message
+                Piwik_Popover.setContent(loadingError.clone().html());
+            }
+        } else {
+            loadingError.show();
+            setTimeout( function(){
+                loadingError.fadeOut('slow');
+            }, 200000000);
```
diosmosis: If the timeout is going to be this big, is it even necessary to have it?

plugins/Morpheus/javascripts/ajaxHelper.js
```diff
@@ -279,11 +279,24 @@ function ajaxHelper() {
...
-    };
+
+        var loadingError = $('#loadingError');
+        if (Piwik_Popover.isOpen()) {
+            if (deferred && deferred.status === 500) {
+                $(document.body).html(deferred.responseText);
+            } else {
+                // Just display a generic error message
+                Piwik_Popover.setContent(loadingError.clone().html());
+            }
```
diosmosis: I'm not too sure if this suggestion will work (or is even necessary), but if we store the response text/HTML in a new iframe w/ `sandbox=""`, instead of directly setting the HTML will it help prevent possible XSS?

(cc @tsteur)

diosmosis: Actually, would be simpler to just use `piwikHelper.escape(...)`. And this is about the document.body bit above, not the Piwik_Popover.setContent call.

plugins/Morpheus/javascripts/ajaxHelper.js
tsteur: I reckon we could use same timeout as before?

plugins/Morpheus/javascripts/ajaxHelper.js
tsteur: @diosmosis if I read this right you are referring to the document.body above in https://github.com/matomo-org/matomo/pull/14424/files#diff-7d5f6e773fb7b595b1577f683ada8166R286

I reckon we could just show a generic error message about this like we do in some other places?

Haven't checkout out the PR otherwise to see what kind of message it returns here etc. It might be safe to overwrite the body though assuming all the content is properly escaped maybe.

plugins/Morpheus/javascripts/ajaxHelper.js
katebutler: Have removed the timeout since it doesn't really make sense for it to fade out anyway.

katebutler:
I'm actually unable to trigger an error to test this with now. The original issue used the example of loading the row evolution popup, but errors in this (and others that I tried, like visitor details popup etc) seem to be handled by `broadcast.js` now?

katebutler:
Exceptions are now handled elsewhere (can't find where!) so this PR  now only covers fatal errors.

plugins/Morpheus/javascripts/ajaxHelper.js
diosmosis: I tested w/ a fatal error, and while it works I don't think it's safe to just dump the HTML in the DOM. If an attacker can find a way to inject JS in the fatal error page, it would create an XSS vulnerability. One way around this might be to strip out the contents of the `<body>` element in the response, and pass it through `piwikHelper.escape(...)`, ie,

```
var body = // ...
$(document.body).html(piwikHelper.escape(body));
```

tsteur: that would be the case though for the regular safemode page as well or not? Like shouldn't any JS be escaped basically? Not sure... maybe it could be possible to escape the HTML and then run JS although would have expected any html entity etc to be escaped maybe.

When you escape the body then HTML tags won't be rendered ... I mean the would be shown?

diosmosis: Well, the `escape()` function might be named incorrectly, but it actually uses angular's sanitize functionality.

I think it would affect the regular safemode page, except I think the vulnerability would be a little less severe there, since the safemode page has no access to things like the token auth, whereas injecting html into a page w/ the modal would generally give an attacker's JS access to that info.

Regardless, I think the principle at play here isn't to avoid an established vulnerability (I don't think there's a way atm to make the safemode page unsafe), but to give attackers as few opportunities as possible to find/create an exploit.

katebutler: Now uses `piwikHelper.escape`.

diosmosis (approved):

PR #14425, opened 2019-05-07
Ignore variables that start w/ @ when auto-detecting tables in segment sql expressions.

diosmosis:


tsteur:
Can you add some tests for this @diosmosis ?

diosmosis:
will do

diosmosis:
@tsteur Updated

PR #14429, opened 2019-05-08
bower.json is aligned with librabies from libs/bower_components

mpashkovskiy:
I noticed that:
* versions of libraries in bower.json is out of sync with actual ones used from libs/bower_components
* some of the libraries are missing in bower.json


tsteur:
we're actually in the process of moving to NPM. Not sure if it is worth updating libs here in the meantime @diosmosis ? https://github.com/matomo-org/matomo/pull/14082

Would also need to check later if still all the ui tests work

diosmosis:
@tsteur probably not, I don't think it will confer any benefits to change bower.json now. But it would be good to double check we're using the right versions in the npm branch, so maybe don't close this PR.

tsteur:
@diosmosis closing this PR for now and as part of #14082 we can then make sure the right versions are used. Otherwise we just keep looking at the PR.

@mpashkovskiy cheers for the PR 👍  Sorry we can't merge it's simply due to the fact that we will be using npm in the future.

mpashkovskiy:
Frankly speaking I can't wait when Matomo migrate to npm!

PR #14436, opened 2019-05-09
Group different writings of Instagram in reports

### LINKED ISSUE #14433: Social networks report shows instagram twice
As a result of capitalising Instagram in https://github.com/matomo-org/searchengine-and-social-list/pull/58 we now sometimes have a mix of both entries especially when archiving months/years.

Screenshot from 2019-05-09 20-03-54

Would be good if we could merge these rows together in the API and keep the data clean.

sgiehl:
fixes #14433

### DIFF
--- a/plugins/Referrers/API.php
+++ b/plugins/Referrers/API.php
@@ -365,6 +365,35 @@ class API extends \Piwik\Plugin\API
 
         $dataTable = Archive::createDataTableFromArchive(Archiver::SOCIAL_NETWORKS_RECORD_NAME, $idSite, $period, $date, $segment, $expanded, $flat);
 
+        $groupIntagramLabel = function($dataTable) use (&$groupIntagramLabel) {
+
+            if ($dataTable instanceof DataTable\Map) {
+                $dataTables = $dataTable->getDataTables();
+            } else {
+                $dataTables = [$dataTable];
+            }
+
+            foreach ($dataTables as $table) {
+                if ($table instanceof DataTable\Map) {
+                    $groupIntagramLabel($table);
+                    continue;
+                }
+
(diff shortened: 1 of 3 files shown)


plugins/Referrers/API.php
```diff
@@ -365,6 +365,13 @@ public function getSocials($idSite, $period, $date, $segment = false, $expanded
 
         $dataTable = Archive::createDataTableFromArchive(Archiver::SOCIAL_NETWORKS_RECORD_NAME, $idSite, $period, $date, $segment, $expanded, $flat);
 
+        $dataTable->filter('GroupBy', ['label', function($value) {
```
tsteur: Could we do this only if one or the other label exists? Typically the GroupBy filter is quite slow, and checking if a label exists in the report is really fast usually. 

Not too important but could be worth thinking about it.

sgiehl: updated the branch

plugins/Referrers/API.php
```diff
@@ -365,6 +365,35 @@ public function getSocials($idSite, $period, $date, $segment = false, $expanded
 
         $dataTable = Archive::createDataTableFromArchive(Archiver::SOCIAL_NETWORKS_RECORD_NAME, $idSite, $period, $date, $segment, $expanded, $flat);
 
+        $groupIntagramLabel = function($dataTable) use (&$groupIntagramLabel) {
```
tsteur: @sgiehl might be better to create a filter for this? would remove some complexity and be better to read etc. like `$dataTable->filter('Piwik\Plugins\Referrers\DataTable\Filter\Group')` then you don't need to check if it's a map etc and in theory it would be even possible to write a unit test for the group filter but wouldn't really be needed.

PR #14449, opened 2019-05-15
Fix download issues via cURL with HTTP/2

### LINKED ISSUE #14339: HTTP header are not case insensitive
I am currently setting up a test-instance for my workshop and I try to be pedantic and not skip over potential bugs when doing so.

Currently I am really confused by the GeoIP downloader (I think I saw the same bug for the initial download on my real instance, but the updater works):

When I try to initially download the free geoip2 database, I just get the generic `General_DownloadFail_HttpRequestFail` error.

Adding a `print_r($expectedFileSizeResult);` shows the source of the issue:

https://github.com/matomo-org/matomo/blob/07adf0ac04dfb8f35be41d9add1855b6e646f8db/core/Http.php#L737-L745

```
Array
(
    [status] => 200
    [headers] => Array
        (
            [date] => Thu, 11 Apr 2019 16:29:34 GMT
            [content-type] => application/gzip
            [content-length] => 29444833
            [set-cookie] => __cfduid=d5e7c2e5b339a937a45385e60f1a69d7c1555000174; expires=Fri, 10-Apr-20 16:29:34 GMT; path=/; domain=.geolite.maxmind.com; HttpOnly
            [content-disposition] => attachment; filename=GeoLite2-City_20190409.tar.gz
            [last-modified] => Mon, 08 Apr 2019 12:07:23 GMT
            [cf-cache-status] => HIT
            [expires] => Thu, 11 Apr 2019 20:29:34 GMT
            [cache-control] => public, max-age=14400
            [accept-ranges] => bytes
            [expect-ct] => max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
            [server] => cloudflare
            [cf-ray] => 4c5e57111a0dbef8-FRA
        )

    [data] => 
)
```

Cloudflare only returns lowercase HTTP headers, but Matomo only checks for `'Content-Length'` and therefore the download fails.

Interestingly fixing this, the download still doesn't work as I just get an unhelpful `Failed sending HTTP request` here:
https://github.com/matomo-org/matomo/blob/07adf0ac04dfb8f35be41d9add1855b6e646f8db/core/Http.php#L581-L586

I'm also a bit clueless why I am the first one to come across this issue. Am I missing something?

katebutler:
Fix two issues observed when downloading the GeoIP database via cURL with HTTP/2:

- The response headers are returned in lowercase e.g. 'content-length' instead of 'Content-Length'
- Work around an apparent bug in libcurl where sending a `Range` header via `CURLOPT_HTTPHEADER` causes curl to error "Failed sending HTTP request". Sending it via `CURLOPT_RANGE` instead has been tested with HTTP/1.1 and HTTP/2 and works with both.

Fixes #14339

### DIFF
--- a/core/Http.php
+++ b/core/Http.php
@@ -182,9 +182,11 @@ class Http
             . ($userAgent ? " ($userAgent)" : '');
 
         // range header
+        $rangeBytes = '';
         $rangeHeader = '';
         if (!empty($byteRange)) {
-            $rangeHeader = 'Range: bytes=' . $byteRange[0] . '-' . $byteRange[1] . "\r\n";
+            $rangeBytes = $byteRange[0] . '-' . $byteRange[1];
+            $rangeHeader = 'Range: bytes=' . $rangeBytes . "\r\n";
         }
 
         list($proxyHost, $proxyPort, $proxyUser, $proxyPassword) = self::getProxyConfiguration($aUrl);
@@ -517,7 +519,6 @@ class Http
                 CURLOPT_HTTPHEADER     => array_merge(array(
                     $xff,
                     $via,
-                    $rangeHeader,
                     $acceptLanguage
                 ), $additionalHeaders),
                 // only get header info if not saving directly to file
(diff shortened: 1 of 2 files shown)


Findus23:
Perfect, this fixes it for me!
There might be more HTTP/2 bugs in Matomo, but at least this download now works fine.

diosmosis (approved):
LGTM

PR #14459, opened 2019-05-17
Segment menu behind overlay of transition

### LINKED ISSUE #14458: Segment menu behind overlay of transition
reported in https://forum.matomo.org/t/segment-menu-behind-overlay-of-transition/32915

> When I have opened the menu Behaviour > Transitions and select a new segment, the dropdown is shown behind the transition overlay. I think it is nicer to have the segment dropdown up front of the transition overlay. See the image.

grafik

I can reproduce this and it would be great if someone familiar with the z-index structure in Matomo could take a look at this.

fdellwing:
Fixes #14458

Someone not really knowing how `z-index` works seems to have done an oopsie here ;)

### DIFF
--- a/plugins/Transitions/stylesheets/transitions.less
+++ b/plugins/Transitions/stylesheets/transitions.less
@@ -202,7 +202,7 @@ body .ui-tooltip.Transitions_Tooltip_Small {
 
 #transitions_report {
     #Transitions_Container {
-        z-index: 980;
+        z-index: 0;
     }
     #Transitions_CenterBox {
         z-index: 987;


fdellwing:
Additional note: This was also broken if you added the report as widget to the dashboard. That is fixed with this PR as well.

fdellwing:
I think it would be good to have a test to make sure this does not regress, but I have no clue about writing these.

diosmosis (approved):
LGTM

No need for a new test, UI tests take a long time to run so we don't want to add new ones for small issues like this. Otherwise the run time can really add up.

PR #14461, opened 2019-05-17
make opt-out iframe reload cacheless

dazwiafl:
make opt-out iframe reload cacheless


diosmosis (approved):
LGTM

Thanks for the contribution @dazwiafl !

PR #14465, opened 2019-05-20
Use query params to add goal processed metrics instead of filter directly.

diosmosis:
Using query params here allows us to send query params over in ProxySite, so the processed metrics are computed properly and displayed properly.


core/API/DataTablePostProcessor.php
```diff
@@ -261,11 +261,21 @@ public function applyProcessedMetricsGenericFilters($dataTable)
             // ignore
         }
 
+        $goalsToProcess = null;
+        try {
+            $goalsToProcess = Common::getRequestVar('filter_show_goal_columns_process_goals', null, 'string', $this->request);
```
mattab: is `filter_show_goal_columns_process_goals`  a new parameter? is it possible to add some test for this feature?

mattab (changes requested):

core/API/DataTablePostProcessor.php
diosmosis: It's now used by the goals visualization so it should already be covered. It's not really an exposed API, it's just for ProxySite.

tsteur:
some tests are failing, eg this one: https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/34312/GoalsTable_goals_table_ecommerce.png

mattab:
@diosmosis Can you check the failing tests and adjust PR so we can hopefully merge? Or shall we maybe move it to 3.11.0 if you need more time?

mattab:
@diosmosis is this done?
@tsteur could you please review & merge if it looks good to you?

diosmosis:
Waiting for tests to finish before calling it done, but the issue mentioned is fixed

tsteur:
I didn't test it but looks good to me to merge if tests pass. Not quite sure how to reproduce it etc so hard to tell.

diosmosis:
Just to explain, bug here is that `AddColumnsProcessedMetricsGoal` uses processed metrics and is used directly in the goals table visualization. When proxying requests, the target instance computes processed metrics, so to avoid errors we don't do it again in the proxy instance. This means the processed metrics added by the filter are not computed. Using a query param here allows us to compute the metrics in the target instance.

PR #14466, opened 2019-05-20
Add newsletter signup button to user settings page

### LINKED ISSUE #14118: Let any user in Matomo signup to our Matomo newsletter
Currently we let Super Users signup to our newsletter during the installation of Matomo.
It looks like this:
https://github.com/matomo-org/matomo/raw/3.x-dev/plugins/Installation/tests/UI/expected-screenshots/Installation_superuser.png

(there are 2 different checkboxes)

In this issue we would let all users sign up to Matomo newsletter from within their profile page in Matomo. 

Also it may be worth thinking if we want to keep the 2 separate newsletters or simplify and only have one newsletter signup? cc @tsteur

katebutler:
Fixes #14118


katebutler:
Ready for review/merge

plugins/UsersManager/angularjs/personal-settings/personal-settings.controller.js
```diff
@@ -43,6 +43,40 @@
...
+            ajaxHandler.addParams(
+                {module: 'UsersManager', action: 'newsletterSignup'}, '' +
+                'GET'
+            );
+
+            var errorCallback = function() {
+                debugger;
+                $('#newsletterSignupMsg').hide();
+                $('#newsletterSignupFailure').show();
+                signupBtn.html('Try Again');
```
diosmosis: This + loading text should be internationalized, correct?

plugins/UsersManager/templates/userSettings.twig
```diff
@@ -120,6 +120,20 @@
...
+    <div id="newsletterSignupMsg">{{ 'UsersManager_NewsletterSignupMessage'|translate|raw }}<br/><br/></div>
+    <div id="newsletterSignupFailure" style="display:none">{{ 'UsersManager_NewsletterSignupFailureMessage'|translate|raw }}<br/><br/></div>
+    <div id="newsletterSignupSuccess" style="display:none">{{ 'UsersManager_NewsletterSignupSuccessMessage'|translate|raw }}<br/><br/></div>
+
+    <button class="btn btn-link" id="newsletterSignupBtn"
+            ng-controller="PersonalSettingsController as personalSettings"
+            ng-click="personalSettings.signupForNewsletter()"
+    >{{ 'General_Yes'|translate }}</button>
+</div>
+{% endif %}
```
diosmosis: @mattab Should there be any specific copy in this new section? The screenshot makes the section look a bit bare.

plugins/UsersManager/tests/UI/UserSettings_spec.js
```diff
@@ -0,0 +1,65 @@
...
+        expect(await pageWrap.screenshot()).to.matchImage('asks_confirmation');
+    });
+
+    it('should load error when wrong password specified', async function () {
+        await page.type('.modal.open #currentPassword', 'foobartest123');
+        btnNo = await page.jQuery('.modal.open .modal-action:not(.modal-no)');
+        await btnNo.click();
+        await page.waitForNetworkIdle();
+
+        pageWrap = await page.$('#notificationContainer');
```
diosmosis: FYI, variables in these tests don't have to be global, they can be scoped w/ `let`/`const`.

plugins/Installation/Controller.php
```diff
@@ -18,15 +18,16 @@
...
 use Piwik\Option;
 use Piwik\Piwik;
 use Piwik\Plugin\Manager;
 use Piwik\Plugins\Diagnostics\DiagnosticService;
 use Piwik\Plugins\LanguagesManager\LanguagesManager;
 use Piwik\Plugins\SitesManager\API as APISitesManager;
 use Piwik\Plugins\UsersManager\API as APIUsersManager;
-use Piwik\Plugins\UsersManager\UserUpdater;
+use Piwik\Plugins\UsersManager\NewsletterSignup;
+use Piwik\Plugins\UsersManager\UserUpdater;use Piwik\Plugins\UsersManager\UserUpdater;
```
diosmosis: Duplicate use statements here

plugins/UsersManager/Controller.php
```diff
@@ -405,6 +409,22 @@ public function recordUserSettings()
         return $toReturn;
     }
 
+    public function newsletterSignup()
+    {
+        Piwik::checkUserIsNotAnonymous();
```
diosmosis: I think we can add `$this->checkTokenInUrl();` here, though not having it wouldn't be too big of a deal.

plugins/UsersManager/Controller.php
```diff
@@ -405,6 +409,22 @@ public function recordUserSettings()
...
+        $userLogin = Piwik::getCurrentUserLogin();
+        $email = Piwik::getCurrentUserEmail();
+
+        $success = NewsletterSignup::signupForNewsletter($userLogin, $email, true);
+        if ($success) {
+            $result = array('success' => true);
+        } else {
+            $result = array('error' => true);
+        }
+        return json_encode($result);
```
diosmosis: You'll also need to call `Piwik\DataTable\Renderer\Json::sendHeaderJSON()` here since controllers will assume the output is `text/html`.

diosmosis (changes requested):
Left some comments on the code, otherwise seems to work when tested locally.

plugins/UsersManager/API.php
```diff
@@ -965,6 +965,7 @@ public function deleteUser($userLogin)
         }
 
         $this->model->deleteUserOnly($userLogin);
+        $this->model->deleteUserSettings($userLogin);
```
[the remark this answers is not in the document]
katebutler: Sure, have added one

plugins/UsersManager/angularjs/personal-settings/personal-settings.controller.js
```diff
@@ -43,6 +43,40 @@
...
+            };
+
+            ajaxHandler.setCallback(function (response) {
+                debugger;
```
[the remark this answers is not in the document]
katebutler: Err, nope, removed

plugins/UsersManager/angularjs/personal-settings/personal-settings.controller.js
katebutler: Changed

plugins/UsersManager/tests/UI/UserSettings_spec.js
katebutler: Cheers, changed

plugins/Installation/Controller.php
katebutler: Whoops, removed

plugins/UsersManager/Controller.php
katebutler: If I add this the ajax call from the user settings page breaks - I guess I would need to change something in the JS there to force it to pass the token?

plugins/UsersManager/Controller.php
katebutler: Done

plugins/UsersManager/Controller.php
diosmosis: For `ajaxHelper` I think `withTokenInUrl()` would do the trick.

katebutler: Thanks, working now.

tsteur:
In the personal settings page, should we mention this block maybe very first?
image

and I suppose we need to maybe show a privacy notice, similar to on our website https://matomo.org/newsletter : 
image

tsteur:
fyi @katebutler see previous comment

tsteur:
Actually, I reckon by the time we show the privacy notice it'll take too much space.

Also to be safe re GDPR it might be good to show an unchecked checkbox and the button be disabled initially (`disabled=true|false` attribute should do this in angular). Clicking on the checkbox would enable the Yes button. Not sure if checkbox is needed though @Findus23 any thoughts?

katebutler:
Have added a short privacy notice and a checkbox.

plugins/UsersManager/NewsletterSignup.php
```diff
@@ -0,0 +1,49 @@
...
+use Piwik\Http;
+use Piwik\Option;
+use Piwik\Url;
+
+class NewsletterSignup
+{
+    public static function signupForNewsletter($userLogin, $email, $matomoOrg = false, $professionalServices = false)
+    {
+        // Don't bother if they aren't signing up for at least one newsletter
+        if (! ($matomoOrg || $professionalServices)) {
```
tsteur: We will also need to return if `Piwik\SettingsPiwik::isInternetEnabled()` is disabled

plugins/UsersManager/Controller.php
```diff
@@ -183,6 +185,9 @@ public function userSettings()
         $view->userTokenAuth = Piwik::getCurrentUserTokenAuth();
         $view->ignoreSalt = $this->getIgnoreCookieSalt();
 
+        $newsletterSignupOptionKey = 'UsersManager.newsletterSignup.' . $userLogin;
```
tsteur: Be good to use a constant for this value `'UsersManager.newsletterSignup.'`

plugins/UsersManager/Controller.php
```diff
@@ -183,6 +185,9 @@ public function userSettings()
...
+        $view->showNewsletterSignup = Option::get($newsletterSignupOptionKey) === false;
```
tsteur: We could also not show the box if `Piwik\SettingsPiwik::isInternetEnabled()` is disabled

plugins/UsersManager/Controller.php
```diff
@@ -405,6 +410,24 @@ public function recordUserSettings()
         return $toReturn;
     }
 
+    public function newsletterSignup()
```
tsteur: @katebutler ideally we would put this into an API so we could implement for example the same action in the mobile app or somewhere else. (of course the mobile API would also need an API method whether it has been submitted already or not but that's not needed for now).

Once this is in the API, it'll be simply `return true/false` and it should simplify things. Only need to slightly adjust the JavaScript that calls it `like {module: 'API', method: 'UsersManager.signupNewsletter'}`

tsteur (changes requested):
@katebutler I left a few comments and also converted the PR to use angular instead of jquery, and our built-in form components and notification component see https://github.com/matomo-org/matomo/pull/14466/commits/8d83f66011385f8baa17604d7655c4e983dc3414 Can show you what it does tmrw. Now also when the checkbox is disabled, the confirm button will be disabled.

tsteur:
BTW there's also a merge conflict

tsteur:
@katebutler could you fix/update the expected screenshots for these 3 screenshots and then merge? thanks

* https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/35313/UIIntegrationTest_api_listing.png
* https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/35313/UserSettings_load.png
* This screenshot might show a bug as it shouldn't be visible when already signed up for ? https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/35313/UserSettings_already_signed_up.png

tsteur:
@katebutler there's still a failing test BTW: https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/35326/UserSettings_signup_success.png

The expected message is no longer showing as the success message is now shown in a success notification. We can usually take a screnshot of that using `#notificationcontainer` or something similar (don't remember the ID of the notifications container).

Issue #14455, opened 2019-05-17
Console command that sends a simple test email


mattab:
-> The goal of this issue is to create a simple console command that sends a test email. The email should have the default Matomo header and footer and just say something like `Hello, world! <br/> This is a test email sent from $MATOMO_URL` and subject `This is a test email sent from $MATOMO_URL` where MATOMO_URL is the matomo instance URL.

Why? To have a simple way to check if sending an email from CLI works. Because many people experience the problem that [Scheduled email reports](https://matomo.org/docs/email-reports/) are not sent from `core:archive` crontab, while they are sent correctly when one manually clicks on `Send report now` button. See this example bug report https://forum.matomo.org/t/auto-archiving-no-emails-sent/32643 (there are many other similar ones). In some (all?) cases, the problem is that the emails won't send from CLI. For example because PHP CLI config is different from the apache/nginx config and there may be something (eg. sendmail) mis-configured.

A common work-around is to use a SMTP server where usually this problem goes away.

Findus23:
Sounds a lot like #14302.
I'm closing this in favor of #14302 as I have cross-linked the related issues there.

PR #14467, opened 2019-05-20
New command to send test emails

### LINKED ISSUE #14302: New cli command to send test-mail to admin
See also #14455 for a description of this proposed feature.

Many bugs similar to #11534, #12634 (e-mail reports work when sent automatically, but not when sent regularily) have been reported in the last months and my current theory is that this is caused by servers where php mail is only set up correctly for the webserver, but broken in the php.ini read by php-cli.

One way to help debugging this (that would also be useful in other cases) would be a new `console` command that would send a testmail (or if it is easy a report selected by id) to the admin (or another address).

While I am not 100% sure if this will fix #11534 and co. it might be usefull for other people who want to make sure they have set up their Matomo instance correctly. (And getting e-mails to work is often one of the most complex parts as it totally depends on the server environment)

katebutler:
Fixes #14302

### DIFF
--- a/plugins/CoreConsole/Commands/TestEmail.php
+++ b/plugins/CoreConsole/Commands/TestEmail.php
@@ -12,6 +12,7 @@ namespace Piwik\Plugins\CoreConsole\Commands;
 use Piwik\Config;
 use Piwik\Mail;
 use Piwik\Plugin\ConsoleCommand;
+use Piwik\SettingsPiwik;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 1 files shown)


diosmosis:
The suggested message/subject described in https://github.com/matomo-org/matomo/issues/14455 (linked in #14302) is a bit different than what's in this PR, can that be changed? Also it mentions sending an HTML email so `Maill::setWrappedHtmlBody()` would need to be used.

katebutler:
Thanks @diosmosis, have updated with the content suggested on #14455

diosmosis (approved):
Looks good and works. Not sure if @mattab had any ideas regarding extra debug information, otherwise this can be merged.

Findus23 (approved):
Looking good.
I hope this helps debugging mail issues in the future.

mattab:
Would be great to have more debug information but I don't know how that's possible. Merging for now :+1:

mattab:
fyi Documented in: [ Emails from Matomo are not being sent, how do I troubleshoot and solve the issue?](https://matomo.org/faq/troubleshooting/faq_34856/)

PR #14469, opened 2019-05-21
If no pivotBy column is specified, make sure when getting first column in table

diosmosis:
… in table to find associated metric index value.


mattab:
@diosmosis looks good. Feedback: tests are failiing, can you update the expected files?

mattab:
actually test results look different... see eg. (is this expected or a regression?)

Screenshot from 2019-05-28 13-03-32

diosmosis:
Was a copy-paste failure, should be fixed.

diosmosis:
@mattab fyi, merged this

PR #14471, opened 2019-05-22
Do not do unprocessed segment check if within archiving process, sinc…

diosmosis:
…e it is not for the main HTTP request.

Just for safety, doesn't fix a known bug.


plugins/SegmentEditor/SegmentEditor.php
```diff
@@ -90,6 +91,11 @@ public function getKnownSegmentsToArchiveForSite(&$segments, $idSite)
 
     public function onNoArchiveData()
     {
+        // if we're within the archiving process, then means there's no data for a child archive, not the main one
```
mattab: Don't really understand the comment unfortunately... is there a way to explain it different maybe?

mattab (changes requested):

plugins/SegmentEditor/SegmentEditor.php
diosmosis: how about "if no archives are found while the archiving process is active, it means we are requesting archives to aggregate data for a period. we only want to do the segment check for the root archive query, so we ignore this" ?

tsteur: I'm not so much into it, aren't we also archiving when `callAggregateCoreMetrics ` is called etc? Do we maybe in general not want to go in here if `trigger=archivephp` is set? Which it might be when archiving? Of course doesn't work when browser archiving is enabled.

diosmosis: It's for browser archiving.

mattab: @diosmosis 

> if no archives are found while the archiving process is active, it means we are requesting archives to aggregate data for a period. 

I want to ask "Why it means we are requesting archives to aggregate data for a period. ", could you also explain this in the comment? (i'm not sure myself, didn't think too deep)

> we only want to do the segment check for the root archive query, so we ignore this

also here "Why we only want to do segment check for the root archive query" ? Be good to explain (i'm not sure myself, didn't think too deep)

diosmosis:
@mattab updated the comment

mattab:
@tsteur could you please review & merge if it looks good to you?

PR #14476, opened 2019-05-24
use PHP filter to validate E-Mails

### LINKED ISSUE #14475: Installer reports invalid email address when the email is valid
`admin@localhost` is a valid email address, however the installer will not let me continue with this as the address for the super user.
The installer did however accept `user@matomo.localhost`

This is a test account on a local install in a development environment, however since [TLDs](https://en.wikipedia.org/wiki/Top-level_domain) were opened up it is no-longer infeasible that someone may have e.g. `user@test` as their email address (i.e. no `.TLD` at the end e.g. `user@example.com`).

On internal (private) networks (e.g. large company intranets) an address without a `.` in the domain may still be valid.

Findus23:
fixes #14475

I think using PHP builtins instead of a 10 year old library avoids bugs and having to maintain a frequently changing list of all valid TLDs.

And the only email validation that is really correct is sending an actual email as suggested in https://github.com/matomo-org/matomo/issues/13533


diosmosis (approved):
LGTM

tsteur:
Looks good to merge. There are a few comments re this filter and that it maybe behave a bit more strict or so but the comments are old and I reckon it's fine 👍  https://www.php.net/manual/en/filter.filters.validate.php

Before merging though could you remove the code from here @Findus23 https://github.com/matomo-org/matomo/blob/3.10.0/config/global.php#L203-L209

and I reckon we could also remove the zend class itself? It doesn't seem to be used

Findus23:
@tsteur Should be done in https://github.com/matomo-org/matomo/pull/14476/commits/c9b59938461513bb67a0d3fa42aa30c89f7b8a2d

I guess one day we should be removing all of those zend classes and replace them with modern libraries.

PR #14482, opened 2019-05-25
send Referrer-Policy header

Findus23:
See also #13479, https://github.com/matomo-org/matomo/issues/13204 and https://github.com/matomo-org/matomo/pull/12780

`strict-origin` only sends a header to the same domain AND origin (so only to HTTPS if Matomo is HTTPS) and no Referer to all other origins.

This does not work in the email report as it seems to be setting headers somewhere else

### DIFF
--- a/core/View.php
+++ b/core/View.php
@@ -282,6 +282,8 @@ class View implements ViewInterface
         if(!empty($this->xFrameOptions)) {
             Common::sendHeader('X-Frame-Options: ' . (string)$this->xFrameOptions);
         }
+        // don't send Referer-Header for outgoing links
+        Common::sendHeader('Referrer-Policy: strict-origin');
 
         return $this->renderTwigTemplate();
     }
--- a/core/View.php
+++ b/core/View.php
@@ -283,7 +283,7 @@ class View implements ViewInterface
             Common::sendHeader('X-Frame-Options: ' . (string)$this->xFrameOptions);
         }
         // don't send Referer-Header for outgoing links
-        Common::sendHeader('Referrer-Policy: strict-origin');
+        Common::sendHeader('Referrer-Policy: same-origin');
 
         return $this->renderTwigTemplate();
     }


diosmosis:
Is there a reason to use `strict-origin` over `no-referrer`? (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#Directives)

Findus23:
I haven’t checked if there aren’t any parts of Matomo that require a referrer (maybe nonce check)

Lukas

diosmosis:
I see, looks like we do use it in some internal requests. Reading the docs more, I think what we want is `same-origin`, that should prevent it from going to any other domains, and from the examples it looks like the protocol must be the same too. Though the docs aren't really specific about that. I'm not sure if it's really a security risk though.

Only thing I might be worried about is the SAML plugin which uses quite a few redirects when logging a user in. Would need to test that still works and possibly allow some requests to omit the referrer policy.

Findus23:
Same-origin sounds like it has the potential to ne less secure than the default as it seems to enable referrers from https to http. That’s why I would go with strict-origin. 

Lukas

diosmosis:
But if it only sends the referrer to the same matomo instance, and not to any other domain, wouldn't it be more secure than strict-origin? Or am I missing something? Note: in the examples on MDN, the domain is sent on cross-origin requests w/ `strict-origin`.

Findus23:
I just checked again and compared with https://scotthelme.co.uk/a-new-security-header-referrer-policy/ and you are right. 
The `origin` in `strict-origin` refers to only sending the origin as referrer and not to being limited to the same origin.
`same-origin` is what we want, sending a referrer to the same origin (domain and protocoll) and no referrer otherwise.

diosmosis:
It took me a bit to realize as well, there's definitely a naming issue there.

tsteur:
image

tested it... it then sends only the domain as referrer instead of the full URL.
This breaks eg `redirectToReferrer` feature etc. Is there maybe a different way? Or maybe it's easier to go through most links and check there is noreferrer set as an attribute? Especially in the dataTable? The dataTable links might be enough?

diosmosis:
@Findus23 will you have time to update this PR?

Findus23:
Okay, after another huge confusion I have now updated the PR

PR #14488, opened 2019-05-27
fix URL to JavaScript tracking guide

Findus23:
The docs seem to be long gone and I'm not even sure which ones are meant.

I would recommend to also set up a redirect as this link is quite prominently on the "no data yet" page.


mattab:
Redirect is set on https://matomo.org/docs/javascript-tracking/

PR #14491, opened 2019-05-27
If new visit is forced, known visitors should still be recognized

### LINKED ISSUE #14452: Why the visitor_days_since_first is not gradually increasing
image
Look at this image, according to the concept of visitor_days_since_first, this value should have be six. 
There are many error values in my project like this, and it is happening yet.
My matomo version is 3.5.1

diosmosis:
Fixes #14452


sgiehl (approved):
Had a rough look through the changes. Looks all good to me.
@diosmosis conflicts with submodule needs to be resolved before merge...

PR #14497, opened 2019-05-29
Default to current ip in UserCountry.getLocationFromIp and add some tests.

### LINKED ISSUE #14496: API to return current user's geolocation "UserCountry.getLocationFromIP"
Goal of this issue is to make our existing API `UserCountry.getLocationFromIP` return the geolocation information for the requesting IP address by default.

Currently, the API requires an `ip` parameter. [Example API request](https://demo.matomo.org/?module=API&method=UserCountry.getLocationFromIP&ip=194.57.91.215&format=xml&token_auth=anonymous) 

But if we make it default to the current requester IP address then it will provide some great value and open new use cases. For example some users want to customise the website's content based on the user's geolocation. So they could simply call the API for example via: `https://analytics.example.comorg/?module=API&method=UserCountry.getLocationFromIP&format=json` in JavaScript.

diosmosis:
Fixes #14496

### DIFF
--- a/plugins/UserCountry/API.php
+++ b/plugins/UserCountry/API.php
@@ -13,6 +13,7 @@ use Piwik\Archive;
 use Piwik\Container\StaticContainer;
 use Piwik\DataTable;
 use Piwik\Date;
+use Piwik\IP;
 use Piwik\Option;
 use Piwik\Period;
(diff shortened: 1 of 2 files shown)


mattab:
Looks good @diosmosis :+1:

PR #14502, opened 2019-05-29
Detect expired session use

diosmosis:


config/global.ini.php
```diff
@@ -397,6 +397,9 @@
 ; Sets the session cookie path
 login_cookie_path =
 
+; the amount of time before a session that was created without the "remember me" option is considered expired
```
mattab: Is this from the time the session was created or rather from the last server side requets this user made?

mattab:
Feedback
* we could change the session idle timeout to 1 hour or maybe even 30min? 

according to https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Session_Management_Cheat_Sheet.md#session-expiration

> Both the idle and absolute timeout values are highly dependent on how critical the web application and its data are. Common idle timeouts ranges are 2-5 minutes for high-value applications and 15-30 minutes for low risk applications.
> When a session expires, the web application must take active actions to invalidate the session on both sides, client and server. The latter is the most relevant and mandatory from a security perspective.

config/global.ini.php
diosmosis: the time the last server side request the user made, will update the docs

diosmosis:
> we could change the session idle timeout to 1 hour or maybe even 30min?

Sure, I'll change the default config value.

diosmosis:
Updated. Feel free to change the default config value.

mattab:
@diosmosis Looking at https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Session_Management_Cheat_Sheet.md#session-expiration
I'm also wondering, do we already have `Absolute Timeout` implemented server side?

diosmosis:
I don't think we want both, otherwise some will complain that the session shouldn't just end while they're using it.

tsteur:
Maybe the absolute session timeout could be set to something like 3 months?

tsteur:
Would then happen max 4 times a year. Any shorter absolute value be annoying indeed.

diosmosis:
This PR implements an idle timeout, not an absolute one (the saved start time of the session updated on each request).

mattab:
@tsteur could you please review this and merge?

core/Session/SessionAuth.php
```diff
@@ -186,4 +192,24 @@ private function updateSessionExpireTime()
         setcookie(session_name(), session_id(), time() + $sessionCookieLifetime, $sessionParams['path'],
             $sessionParams['domain'], $sessionParams['secure'], $sessionParams['httponly']);
     }
+
+    private function isExpiredSession(SessionFingerprint $sessionFingerprint)
+    {
+        $nonRememberedSessionExpireTime = Config::getInstance()->General['login_session_not_remembered_idle_timeout'];
+        $sessionCookieLifetime = Config::getInstance()->General['login_cookie_expire'];
```
tsteur: is the check for `login_cookie_expire`  actually needed or already done somewhere else? Didn't even know we keep the session only active for 14 days.

tsteur: I see... we're extending the session for 14 days each time after it was used.

core/Session/SessionFingerprint.php
```diff
@@ -105,4 +105,9 @@ public function isRemembered()
         $userInfo = $this->getUserInfo();
         return !empty($userInfo['remembered']);
     }
+
+    public function updateSessionStartTime()
+    {
+        $_SESSION[self::SESSION_INFO_SESSION_VAR_NAME]['ts'] = Date::now()->getTimestampUTC();
```
tsteur: do I see this right that the timestamp was basically unused before that?

core/Date.php
```diff
@@ -1070,7 +1070,12 @@ private static function getInvalidDateFormatException($dateString)
...
 
-    private static function getNowTimestamp()
+    /**
+     * For tests.
+     * @return int|null
+     * @ignore
+     */
+    public static function getNowTimestamp()
     {
         return isset(self::$now) ? self::$now : time();
```
tsteur: In `updateStartTime` we use `Date::now()->getTimestampUTC();`. shouldn't we make sure to use the same logic? Likely it won't be an issue and in the end it is maybe the same logic though.

core/Session/SessionAuth.php
```diff
@@ -186,4 +192,24 @@ private function updateSessionExpireTime()
...
+        }
+
+        if ($sessionFingerprint->isRemembered()) {
+            $expireTime = $startTime + $sessionCookieLifetime;
+        } else {
+            $expireTime = $startTime + $nonRememberedSessionExpireTime;
+        }
+
+        $isExpired = Date::getNowTimestamp() > $expireTime;
+        return $isExpired;
```
tsteur: btw would we need to destroy the session when it is expired or not important?

core/Session/SessionAuth.php
```diff
@@ -186,4 +192,24 @@ private function updateSessionExpireTime()
...
+
+        $startTime = $sessionFingerprint->getSessionStartTime();
+        if ($startTime === null) {
+            return false;
+        }
+
+        if ($sessionFingerprint->isRemembered()) {
+            $expireTime = $startTime + $sessionCookieLifetime;
+        } else {
+            $expireTime = $startTime + $nonRememberedSessionExpireTime;
```
tsteur: Here I'm seeing this warning quite often:

> WARNING [2019-06-11 08:15:17 UTC] [5b1e8 ] > core/Session/SessionAuth.php(209): Warning - A non-numeric value encountered - Matomo 3.10.0-b3 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)

tsteur:
Not sure what happens, I set `login_session_not_remembered_idle_timeout  = 30` and I log in, then it appears like I was logged in successfully but am still anonymous and then get also these warnings:

> core/Session/SessionAuth.php(209): Warning - A non-numeric value encountered - Matomo 3.10.0-b3 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)

There is no login failure, I get forwarded to dashboard but am not logged in.

core/Session/SessionFingerprint.php
diosmosis: if I understand your question correctly, it's used via `getSessionStartTime()`, which is used in SessionAuth::isSessionStartedBeforePasswordChange(). Which I think is actually an issue... might need to explicitly store the expiration time instead of updating the start time.

tsteur: that's why I'm asking :)

core/Session/SessionAuth.php
diosmosis: I don't think we can in PHP, since it requires triggering session GC, and we set different session lifetimes based on whether remember me was checked or not.

tsteur:
Works for me 👍 should be good to merge.

PR #14507, opened 2019-06-02
Ensure incompatible visualization entries are still shown as footer icons when they are selected

### LINKED ISSUE #14071: switch back to simple table from insights visualization - how?
I was curious to test insights visualisation in behaviour > pages and when I was presented with couple of drop-down choices, I simply decided to cancel and go back. Except that I can't. Now, I always see this and just cannot switch to simple table view.

page-visualisation-insights

Hovering over bottom left corner only reveals export icon, no switch report visualization options.
Any help with this which does not involve deleting current session?

sgiehl:
When viewing for example a insights report for a unsupported date/period combination, the report currently won't show any visualization icons to switch to another one.

That is caused by the fact that the currently enabled visualization isn't available (`canDisplayViewDataTable` returns `false`). With the current visualization absent the view won't show any visualization icons (caused by https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/CoreHome/templates/_dataTableActions.twig#L9-L34 and https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/CoreHome/templates/_dataTableActions.twig#L54)

Imho it's makes most sense to add the current visualization to the icons even if wouldn't be present anymore. That way it will be shown as active, and it's possible again to switch to another one.

fixes #14071

### DIFF
--- a/core/ViewDataTable/Manager.php
+++ b/core/ViewDataTable/Manager.php
@@ -184,7 +184,7 @@ class Manager
         $nonCoreVisualizations = static::getNonCoreViewDataTables();
 
         foreach ($nonCoreVisualizations as $id => $klass) {
-            if ($klass::canDisplayViewDataTable($view)) {
+            if ($klass::canDisplayViewDataTable($view) || $view::ID == $id) {
                 $footerIcon = static::getFooterIconFor($id);
                 if (Insight::ID == $footerIcon['id']) {
                     $insightsViewIcons['buttons'][] = static::getFooterIconFor($id);


core/ViewDataTable/Manager.php
```diff
@@ -184,7 +184,7 @@ public static function configureFooterIcons(ViewDataTable $view)
         $nonCoreVisualizations = static::getNonCoreViewDataTables();
 
         foreach ($nonCoreVisualizations as $id => $klass) {
-            if ($klass::canDisplayViewDataTable($view)) {
+            if ($klass::canDisplayViewDataTable($view) || $view->isViewDataTableId($id)) {
```
tsteur: I haven't tested... does this somehow affect inherited tables as well? Like GoalTable is selected, and it would make regular htmltable or others maybe available accidentally?

tsteur: AFAIK `isViewDataTableId` also is true for any inherited.

sgiehl: changed the implementation to an direct compare

tsteur (approved):

PR #14508, opened 2019-06-03
New Goal type: Convert when user stays more than X minutes on the site

diosmosis:
This PR adds the following types of goals:

* goals based on visit duration


core/Tracker/GoalManager.php
```diff
@@ -195,6 +207,49 @@ public function detectGoalMatch($goal, Action $action)
...
 
+    private function detectNumericGoalMatch($goal, Action $action, VisitProperties $visitProperties)
+    {
+        switch ($goal['match_attribute']) {
+            case 'visit_duration':
+                $valueToMatchAgainst = ((int) $visitProperties->getProperty('visit_total_time')) / 60;
+                break;
+            case 'visit_total_actions':
+                $valueToMatchAgainst = $visitProperties->getProperty('visit_total_actions') ?: 0;
+                $valueToMatchAgainst = $valueToMatchAgainst + 1; // to account for current action
```
tsteur: not all type of actions account for this. So we can't just do `+1`

core/Tracker/GoalManager.php
```diff
@@ -195,6 +207,49 @@ public function detectGoalMatch($goal, Action $action)
...
+                break;
+            case 'visit_total_pageview':
+                $valueToMatchAgainst = $this->getPageviewCountInVisit($visitProperties->getProperty('idvisit'));
```
tsteur: Any chance to delay this to Matomo 4 where we might have a `visit_total_pageviews`? I would not want to have a query here.

tsteur:
Is there an issue for this?

core/Tracker/GoalManager.php
diosmosis: You mean not all `Action` types add to the `visit_total_actions` dimension?

diosmosis:
> Is there an issue for this?

Not that I could find, though matt thought there was one, I noticed the discrepancy when working on the GA importer. There are goal types we won't be able to import. It didn't seem like a complicated change.

GA provides 'visit duration' goals + 'total number of pages viewed'. They also allow matching on multiple event attributes, but that seems like a more complicated addition.

core/Tracker/GoalManager.php
tsteur: Yep see https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/Actions/Columns/VisitTotalActions.php#L55-L105

core/Tracker/GoalManager.php
diosmosis: I guess we can just ignore goals like this in the GA importer? I suppose it will be easy enough to add support later, some users will just have to wait.

In the meantime, what about visit duration goals? Is that ok for 3.11?

tsteur: visit duration should be fine? Not sure how accurate `visit_total_time` is though. Might be more accurate to do `$request->getCurrentTime() - visit_first_action_time` ?

diosmosis: Sounds safer.

diosmosis: Updated.

core/Tracker/GoalManager.php
```diff
@@ -146,13 +152,20 @@ public function detectGoalsMatchingUrl($idSite, $action)
...
+     * @param Request $request
+     * @return bool|null if a goal is matched, a string of the Action URL is returned, or if no goal was matched it returns null
      */
-    public function detectGoalMatch($goal, Action $action)
+    public function detectGoalMatch($goal, Action $action, VisitProperties $visitor, Request $request)
     {
         $actionType = $action->getActionType();
 
         $attribute = $goal['match_attribute'];
+print $attribute . "\n";
```
fdellwing: This line seems of?

diosmosis: good catch

plugins/Goals/API.php
```diff
@@ -260,22 +265,34 @@ private function checkDescription($description)
...
     {
         if (empty($patternType)) {
             return '';
         }
 
         $patternType = strtolower($patternType);
 
-        $validator = new WhitelistedValue(['exact', 'contains', 'regex']);
+        if (in_array($matchAttribute, GoalManager::$NUMERIC_MATCH_ATTRIBUTES)) {
+            $validValues = ['>'];
```
tsteur: not sure, should we maybe use `greater_than` or so?

tsteur: I think we do this in some other places.

tsteur:
image

is it possible to not show "case insensitive" for numeric goals? And maybe ideally pattern would show "in seconds" or so. Because the example says minutes but I reckon the number we enter is in seconds?

tsteur:
image

maybe the colon could go  "in triggered when" column and it could say "greater than" as not everyone would understand `>`?

tsteur:
Seems to work otherwise in general :+1: . Only now seeing we enter the visit duration in minutes, should it be seconds maybe? just to be a bit more flexible... although I don't think anyone would want to trigger a goal after a few seconds... so minutes should be fine.

plugins/Goals/API.php
diosmosis: There's a limit on the field size, I think 'greater_than' was a little too long. Didn't want to create a db upgrade for something so small, but I can if it's better.

mattab:
> Seems to work otherwise in general +1 . Only now seeing we enter the visit duration in minutes, should it be seconds maybe? just to be a bit more flexible... although I don't think anyone would want to trigger a goal after a few seconds... so minutes should be fine.

Minutes should be better I reckon, but maybe allow for less than 1 minutes, or in general decimal minutes (1.5, 5.5, etc) so it's flexible (didn't check if maybe this already works)

plugins/Goals/API.php
tsteur: it's fine to have an update for it as it's not a log table.

core/Tracker/GoalManager.php
diosmosis: Removed.

diosmosis:
@mattab Updated so it would allow floating point values for minutes (test should reflect this).

@tsteur Applied review fixes

mattab:
Feedback:

* Change `Pattern` to `Time in minutes`
* In inline help, explain the tip that one can specify time in seconds for example 30 seconds by writing 0.5 in the field.
* Does `Allow multiple conversions per visit` make sense for this new goal type? reckon it does not and could be removed. (This type of Goal would only be converted once per visit always?)
* can you update UI tests 
* The user guide https://matomo.org/docs/tracking-goals-web-analytics/ will need updated 


Looking forward to merging this so we can start beta testing our [Google Analytics importer](https://matomo.org/docs/google-analytics-importer/) :rocket: :rocket:

diosmosis:
> can we change it to Visit duration greater than 5 minutes

this is difficult since the code just prints out the pattern_type currently, w/o internationalization. I think this can be done in another PR.

diosmosis:
@mattab updated everything except screenshot test update

mattab:
Looks great & works! :+1: 

@diosmosis Be great if you could also edit the user guide at https://matomo.org/docs/tracking-goals-web-analytics/ and mention this new goal type and feature. 

(just added it now to: https://matomo.org/features)

diosmosis:
@mattab 
> Be great if you could also edit the user guide at https://matomo.org/docs/tracking-goals-web-analytics/ and mention this new goal type and feature.

Updated.

PR #14521, opened 2019-06-11
Fix incorrect video in share links on welcome widget

### LINKED ISSUE #13740: Replace Promo video with new Matomo videos
Now that there are new Matomo Marketing videos, we should replace the Welcome widget on the default dashboard (and also replace all mentions of Piwik with Matomo)

grafik

https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/CoreHome/templates/getPromoVideo.twig

katebutler:
Fixes #13740


plugins/CoreHome/tests/UI/WelcomeWidget_spec.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+describe('WelcomeWidget', function () {
+    var url = "";
+
+    it('should load correctly', async function () {
+        await page.goto(url);
+        await page.waitForNetworkIdle();
+        expect (await page.screenshotSelector('#widgetCoreHomegetPromoVideo')).to.matchImage('load');
+    });
```
diosmosis: Is this widget included in one of the dashboard tests? If so, we may not need to add a separate UI test for a widget like this. Given how heavy the UI tests are, it's better to avoid adding to them when possible IMO.

I'm not sure about the other test as well, if we're just testing the links we could do so in a PHP test. @tsteur what are your thoughts?

tsteur: Not needed in this case. If we can avoid a UI test that is always welcome :)

katebutler: OK then, tests have been removed.

PR #14526, opened 2019-06-14
fix no data message may not be shown when deleteLogs is enabled

tsteur:
* Fixes DEV-1671
* Also fixes `SitesManager::hasTrackedAnyTraffic($idSite)` was returning inverted result. Meaning it was returning true when no traffic was tracked, but the method suggested it should return false in this case.

### DIFF
--- a/plugins/SitesManager/SitesManager.php
+++ b/plugins/SitesManager/SitesManager.php
@@ -16,7 +16,6 @@ use Piwik\Exception\UnexpectedWebsiteFoundException;
 use Piwik\Option;
 use Piwik\Piwik;
 use Piwik\Plugins\CoreHome\SystemSummary;
-use Piwik\Plugins\PrivacyManager\PrivacyManager;
 use Piwik\Settings\Storage\Backend\MeasurableSettingsTable;
 use Piwik\Tracker\Cache;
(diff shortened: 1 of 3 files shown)


plugins/SitesManager/SitesManager.php
```diff
@@ -68,10 +69,30 @@ public function redirectDashboardToWelcomePage(&$module, &$action)
...
+                }
+
+                $module = 'SitesManager';
+                $action = 'siteWithoutData';
+            }
             return;
         }
 
-        if (self::hasTrackedAnyTraffic($siteId)) {
+        if (!self::hasTrackedAnyTraffic($siteId)) {
```
diosmosis: `delete_logs_enable` can be turned on/off, correct? Should we set the above option here too?

diosmosis: in an 'else' I mean

tsteur: :+1: pushing it now. That should make it a lot simpler

diosmosis:
One more comment: is there a way to make this change visible in tests?

diosmosis:
Going to add some tests myself then merge

tsteur:
Thank you very much @diosmosis

Issue #12630, opened 2018-03-20
Send tracking code instructions to a developer by email


tsteur:
In order to set up the JavaScript tracking for Matomo, it is usually required that a developer installs a tracking code. It would be a great if an admin or super user could enter an email address of a developer and directly send the tracking code instructions to the developer. This would work from the "No data screen in the reporting menu" but also from the "Tracking code page in the admin".

The email would show the JS Tracking code as currently shown in the UI and would also link to developer pages etc. Note: For WhiteLabeled emails we would maybe not include a link or we would need an event to be able to modify behaviour of what is sent out. 

I thought we have an issue for this already for quite a while but couldn't find it.

tsteur:
As we're now sending more people to the "No data has been recorded yet" page which you see when you haven't tracked any data yet, it would be very helpful to have this feature. 

The button should be visible on the "No data has been recorded page yet" and if easily possible also on the "Tracking code generator" page in admin and ideally also during Matomo installation where we show the tracking code as well.

When clicking the button, we would probably ask for an email address and send similar content as seen on the "No data has been recorded yet page" except for the first paragraph (`No analytics data has been tracked for this website yet. Please set up the tracking code below into your website or mobile app if you haven't done already. This message will disappear as soon as some data was tracked for this website. In the meantime, maybe ...`) which should maybe instead say some other message.

We need to see what else makes sense to include. We could let a user type an additional message but wondering if it might be used for spam then. 

Reply-to email should be set to the current user's email. We should also mention that this user has sent the email.

Maybe we need to limit how often this feature can be used? Like max 5 times per site re spam etc?

Any thoughts @mattab @Findus23

Findus23:
On one hand this sounds like a useful feature I have already seen in some other marketing targeting tools.
But on the other hand this means we are handling personal information by and sending E-Mails to a person that did not consent to either of those things. I don't know if such "direct messages" are an exception, but it might make things more complicated with anti-spam and privacy laws.

tsteur:
What if this was a mailto link actually?

Findus23:
I like that idea. It would be far simpler and also more natural (Who is sending automated messages to their colleagues?).

tsteur:
👍 like that much better too

PR #14527, opened 2019-06-14
Improve no data has been recorded page yet

tsteur:
This page has never been changed to the "new look".

Looks now like this:
nodatanew

compared to before:
nodatapage_current (1)

There were a few problems with the old page:

* Didn't look very inviting and bit unstructured
* The "ignore this message" button was shown first and probably often people would just click this button and ignore the rest but actually we want people to set up tracking and make sure they continue looking at the content. That's why I rather moved this button now towards the end of the page.
* The page said " no data..." but didn't really suggest a solution. The "Tracking code for XYZ" under the "ignore this message" button didn't really look too much related and it wasn't really clear the user needs to do any of these things
* HTTP API was missing

We also need to make sure to add a button to send all these instructions to a developer see https://github.com/matomo-org/matomo/issues/12630  I will schedule this issue as it's important people can easily forward this information.

Should we maybe swap "Integrations" and "JavaScript tracking code"? Not sure.

Could also change "No data has been recorded yet" to something like "No data has been recorded yet - You're only one step away from ..." eg "from getting a 360 degree of your visitors" or something (would more apply to cloud/premium features though, would need a different "from ..." message).

I suppose we don't need to mention the image tracking pixel?

In integrations section we could later directly link to some new user guides for wordpress, etc.

Be good if this could make it into 3.10 but all good if not. We could also apply the patch later.

ping @mattab


mattab:
Looks good! small tweaks that have impact :+1:

PR #14528, opened 2019-06-14
Fix GitHub links for 'About matomo'-Page (Footer)

obendev:
There are still old GitHub links in the footer of the 'About matomo'-Page
I changed them to the new ones, so that there is no unnecessary redirect

PR #14529, opened 2019-06-14
Update feedback icons

Findus23:
Thanks to @obendev for the idea in https://github.com/matomo-org/matomo/pull/14528

The newsletter icon is from https://material.io/tools/icons which I think is already mentioned in the legalnotice.

image


mattab:
it looks better now :+1:

PR #14537, opened 2019-06-18
Copy downloaded plugin to all plugins directories

### LINKED ISSUE #14534: Copy downloaded plugins from Marketplace to MATOMO_PLUGIN_DIRS folder too
Hi,
The new feature allowing different folders for plugins using the `MATOMO_PLUGIN_DIRS` env is great.

But is it possible to copy downloaded plugins from Marketplace UI also into declared directories in addition to Matomo `/plugins` folder.

Download logic copies extracted .zip file to '/plugins' folder and it's not using new `getPluginsDirectories` function

https://github.com/matomo-org/matomo/blob/6395855aa7f3813cc7f413a18d31505ea26ba32a/plugins/CorePluginsAdmin/PluginInstaller.php#L52

https://github.com/matomo-org/matomo/blob/6395855aa7f3813cc7f413a18d31505ea26ba32a/plugins/CorePluginsAdmin/PluginInstaller.php#L68

https://github.com/matomo-org/matomo/blob/6395855aa7f3813cc7f413a18d31505ea26ba32a/plugins/CorePluginsAdmin/PluginInstaller.php#L298

https://github.com/matomo-org/matomo/blob/6395855aa7f3813cc7f413a18d31505ea26ba32a/plugins/CorePluginsAdmin/PluginInstaller.php#L300

The goal is to easily "track" all the added/deleted plugins so creating new Matomo deployment becomes much easier.

This is useful with Kubernetes because i can define `MATOMO_PLUGIN_DIRS` folder to point to a Persistent Volume path shared between many PODs

nabiltntn:
When downloading a plugins from Marketplace UI, it would be useful to copy them also to folders defined in MATOMO_PLUGINS_DIRS . This is useful when Matomo is deployed in env with Persistent Volume like Kubernetes

The PR copies plugin to all directories and not to one of them because in practice `MATOMO_PLUGINS_DIRS` will have only one directory outside Matomo directory.

fixes #14534

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -705,9 +705,6 @@ num_days_before_tracking_code_reminder = 5
 ; breaks and reencrypts SSL connections you can set your custom file here. 
 custom_cacert_pem=
 
-; set to 1 to copy downloaded plugins to all plugins directories including those defined in MATOMO_PLUGIN_DIRS env variable 
-copy_to_plugins_dirs = 0
-
 [Tracker]
 
 ; Matomo uses "Privacy by default" model. When one of your users visit multiple of your websites tracked in this Matomo,
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -705,6 +705,9 @@ num_days_before_tracking_code_reminder = 5
 ; breaks and reencrypts SSL connections you can set your custom file here. 
 custom_cacert_pem=
 
+; set to 1 to copy downloaded plugins to all plugins directories including those defined in MATOMO_PLUGIN_DIRS env variable 
+copy_to_plugins_dirs = 0
+
 [Tracker]
 
 ; Matomo uses "Privacy by default" model. When one of your users visit multiple of your websites tracked in this Matomo,
--- a/plugins/CorePluginsAdmin/PluginInstaller.php
+++ b/plugins/CorePluginsAdmin/PluginInstaller.php
@@ -19,6 +19,7 @@ use Piwik\Plugin\Manager;
 use Piwik\Plugins\Marketplace\Marketplace;
 use Piwik\Unzip;
 use Piwik\Plugins\Marketplace\Api\Client;
+use Piwik\Config;
 
 /**
  *
@@ -297,12 +298,20 @@ class PluginInstaller
 
     private function copyPluginToDestination($tmpPluginFolder)
     {
-        $pluginsDir = Manager::getPluginsDirectory();
-        $pluginTargetPath = $pluginsDir . $this->pluginName;
-
-        $this->removeFolderIfExists($pluginTargetPath);
-
-        Filesystem::copyRecursive($tmpPluginFolder, $pluginsDir);
+        if (!empty(Config::getInstance()->General['copy_to_plugins_dirs'])) {
+            $pluginsDirs = Manager::getPluginsDirectories();
+            // Copy plugin to all plugins directories
+            foreach ($pluginsDirs as $dir) {
+                $pluginTargetPath = $dir . $this->pluginName;
+                $this->removeFolderIfExists($pluginTargetPath);
+                Filesystem::copyRecursive($tmpPluginFolder, $dir);
+            }
+        } else {
+            $pluginsDir = Manager::getPluginsDirectory();
+            $pluginTargetPath = $pluginsDir . $this->pluginName;
+            $this->removeFolderIfExists($pluginTargetPath);
+            Filesystem::copyRecursive($tmpPluginFolder, $pluginsDir);
+        }
     }
 
     /**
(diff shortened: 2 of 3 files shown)


config/global.ini.php
```diff
@@ -705,6 +705,9 @@
 ; breaks and reencrypts SSL connections you can set your custom file here. 
 custom_cacert_pem=
 
+; set to 1 to copy downloaded plugins to all plugins directories including those defined in MATOMO_PLUGIN_DIRS env variable 
+copy_to_plugins_dirs = 0
```
tsteur: @nabiltntn AFAIK `MATOMO_PLUGIN_DIRS ` is either defined as an environment variable or GLOBALS (through bootstrap.php). Should we have the same for this directory that it can be only configured as a GLOBAL and/or ENV variable? Because the feature should only work if the `copy to plugin directory` is also defined as a `plugin directory`.

plugins/CorePluginsAdmin/PluginInstaller.php
```diff
@@ -297,12 +298,20 @@ private function fixPluginFolderIfNeeded($tmpPluginFolder)
...
-        $pluginsDir = Manager::getPluginsDirectory();
-        $pluginTargetPath = $pluginsDir . $this->pluginName;
-
-        $this->removeFolderIfExists($pluginTargetPath);
-
-        Filesystem::copyRecursive($tmpPluginFolder, $pluginsDir);
+        if (!empty(Config::getInstance()->General['copy_to_plugins_dirs'])) {
+            $pluginsDirs = Manager::getPluginsDirectories();
+            // Copy plugin to all plugins directories
+            foreach ($pluginsDirs as $dir) {
```
tsteur: this copies it into all plugin directories, I reckon matomo may have trouble with this if there are multiple directories defined. It should copy the plugin only into one plugin directory. We have uses cases where we define multiple plugin directories (one directory for each plugin in fact)

config/global.ini.php
nabiltntn: I changed the implementation to use a new env var : `MATOMO_PLUGIN_COPY_DIR`

plugins/CorePluginsAdmin/PluginInstaller.php
nabiltntn: i changed to copy only to directory defined in `MATOMO_PLUGIN_COPY_DIR` if it's within `getPluginsDirectories ` array

plugins/CorePluginsAdmin/PluginInstaller.php
```diff
@@ -297,12 +297,19 @@ private function fixPluginFolderIfNeeded($tmpPluginFolder)
...
     private function copyPluginToDestination($tmpPluginFolder)
     {
-        $pluginsDir = Manager::getPluginsDirectory();
-        $pluginTargetPath = $pluginsDir . $this->pluginName;
+        // if MATOMO_PLUGIN_COPY_DIR is defined , copy plugin to this directory too
 
-        $this->removeFolderIfExists($pluginTargetPath);
+        $pluginsDirs = [Manager::getPluginsDirectory()];
+        
+        if (!empty($GLOBALS['MATOMO_PLUGIN_COPY_DIR'])) {
```
tsteur: Not sure why you are copying the plugin to two directories? Shouldn't it only copy the plugin to one directory?

nabiltntn: I thought it's important to copy plugin to Matomo's /plugins directory too.
I changed the code to use only one directory

plugins/CorePluginsAdmin/PluginInstaller.php
```diff
@@ -297,19 +297,16 @@ private function fixPluginFolderIfNeeded($tmpPluginFolder)
 
     private function copyPluginToDestination($tmpPluginFolder)
     {
-        // if MATOMO_PLUGIN_COPY_DIR is defined , copy plugin to this directory too
+        $pluginsDir = Manager::getPluginsDirectory();
 
-        $pluginsDirs = [Manager::getPluginsDirectory()];
-        
         if (!empty($GLOBALS['MATOMO_PLUGIN_COPY_DIR'])) {
```
tsteur: In case someone configures `$GLOBALS['MATOMO_PLUGIN_COPY_DIR']` directly, do we maybe also need to check that directory is actually one of the configured plugin directories?

nabiltntn: i made a change in `initPluginDirectories` to check `$GLOBALS['MATOMO_PLUGIN_COPY_DIR']` in both cases ( via env var or directly )

tsteur:
cheers @nabiltntn

PR #14538, opened 2019-06-19
Separate checking for empty site from checking whether we should do e…

diosmosis:
…mpty site check.

### DIFF
--- a/plugins/SitesManager/SitesManager.php
+++ b/plugins/SitesManager/SitesManager.php
@@ -65,7 +65,7 @@ class SitesManager extends \Piwik\Plugin
             return;
         }
 
-        $shouldPerformEmptySiteCheck = $this->shouldPerormEmptySiteCheck($siteId);
+        $shouldPerformEmptySiteCheck = self::shouldPerormEmptySiteCheck($siteId);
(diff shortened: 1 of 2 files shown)


plugins/SitesManager/SitesManager.php
```diff
@@ -65,6 +65,11 @@ public function redirectDashboardToWelcomePage(&$module, &$action)
             return;
         }
 
+        $shouldPerformEmptySiteCheck = $this->shouldPerormEmptySiteCheck($siteId);
```
tsteur: shouldn't that be `self::shouldPer..`?

plugins/SitesManager/SitesManager.php
```diff
@@ -88,6 +93,12 @@ public function redirectDashboardToWelcomePage(&$module, &$action)
     }
 
     public static function hasTrackedAnyTraffic($siteId)
+    {
```
tsteur: This method is also used in a task, does it need `shouldPerormEmptySiteCheck ()` there?

tsteur:
@diosmosis haven't tested it but if tests pass looks good to merge 👍

PR #14550, opened 2019-06-20
Add new translation keys for mobile app

tsteur:
See https://github.com/matomo-org/matomo-mobile-2/pull/5344#issuecomment-503764394 except for `Mobile_PossibleSslErrorExplanation2 ` which isn't used  currently

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -566,7 +566,13 @@
         "VerifyAccount": "Verifying Account",
         "ValidateSslCertificate": "Validate SSL Certificate",
         "VerifyLoginData": "Make sure your username and password combination is correct.",
-        "YouAreOffline": "Sorry, you are currently offline"
+        "YouAreOffline": "Sorry, you are currently offline",
+        "ExceptionNoViewAccess": "Please check your username and password and make sure you have %s access for at least one website.",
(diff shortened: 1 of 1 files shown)


lang/en.json
```diff
@@ -566,7 +566,13 @@
         "VerifyAccount": "Verifying Account",
         "ValidateSslCertificate": "Validate SSL Certificate",
         "VerifyLoginData": "Make sure your username and password combination is correct.",
-        "YouAreOffline": "Sorry, you are currently offline"
+        "YouAreOffline": "Sorry, you are currently offline",
+        "ExceptionNoViewAccess": "Please check your username and password and make sure you have %s access for at least one website.",
+        "Mobile_HowtoExitAndroid": "Please click BACK again to exit",
+        "PiwikMarketplace": "Matomo Marketplace",
```
sgiehl: Maybe we should directly change that key to `MatomoMarketplace` ?

tsteur: sure, will change it

sgiehl (approved):
Looks good now. Could also be merged for 3.10 I guess, as it doesn't affect core in any way...

PR #14551, opened 2019-06-21
Faster User ID reports processing: use ranking query

diosmosis:


plugins/UserId/Archiver.php
```diff
@@ -77,19 +78,31 @@ protected function aggregateUsers()
...
+            $rankingQuery
         );
 
         if ($query === false) {
             return;
         }
 
         $rowsCount = 0;
-        while ($row = $query->fetch()) {
+        foreach ($query as $row) {
```
tsteur: didn't know that worked :+1:

tsteur:
haven't tested it but looks good.
A test is still failing: https://travis-ci.org/matomo-org/matomo/jobs/548604098#L801

This test seems unrelated https://travis-ci.org/matomo-org/matomo/jobs/548604100#L925

mattab:
Be great if we can merge this soon and release rc2 👍

plugins/UserId/Archiver.php
diosmosis: Neither did I :)

plugins/UserId/Archiver.php
```diff
@@ -137,4 +150,15 @@ private function setVisitorIds(DataTable $dataTable)
         }
     }
 
+    private function getRankingQueryLimit()
+    {
+        $configGeneral = Config::getInstance()->General;
+        $configLimit = $configGeneral['archiving_ranking_query_row_limit'];
+        $limit = $configLimit == 0 ? 0 : max(
+            $configLimit,
+            $configGeneral['datatable_archiving_maximum_rows_standard']
```
mattab: * let's introduce a different setting eg `datatable_archiving_maximum_rows_userid` 
* set this to a higher value like 5000 rows (i expect users will want to archive more rows in this report.)

tests/PHPUnit/System/expected/test_reportLimiting__UserId.getUsers_day.xml
```diff
@@ -0,0 +1,183 @@
...
+		<nb_visits>5</nb_visits>
+		<nb_actions>10</nb_actions>
+		<nb_users>1</nb_users>
+		<max_actions>2</max_actions>
+		<sum_visit_length>5</sum_visit_length>
+		<bounce_count>0</bounce_count>
+		<nb_visits_converted>0</nb_visits_converted>
+		
+	</row>
+	<row>
```
mattab: Would be useful if the tests could set a lower limit like 10, and show the behavior works (show the row `Others`)

mattab (changes requested):

tests/PHPUnit/System/expected/test_reportLimiting__UserId.getUsers_day.xml
diosmosis: It does, that's the whole point of this test. See https://github.com/matomo-org/matomo/pull/14551/files#diff-5ca2964519f1aaf3fb5ca033dcb48da0

plugins/UserId/Archiver.php
diosmosis: @mattab changed & build fixed. Had to add another order by to make the results deterministic.

plugins/UserId/Archiver.php
```diff
@@ -118,6 +118,21 @@ protected function insertDayReports()
     {
         /** @var DataTable $dataTable */
         $dataTable = $this->arrays->asDataTable();
+
+        // deal w/ ranking query summary row
+        $rankingQuerySummaryRow = $dataTable->getRowFromLabel(DataTable::LABEL_SUMMARY_ROW);
+        if ($rankingQuerySummaryRow) {
+            $rankingQuerySummaryRowId = $dataTable->getRowIdFromLabel(DataTable::LABEL_SUMMARY_ROW);
```
mattab: can you explain why this code is needed, and do we do this kind of logic in other archivers too (which ones)? if not, why is it needed in this archiver?

diosmosis: Ranking query creates a row w/ label -1, this is added to the DataArray which is just an array, not a datatable so it has no concept of a summary row. When DataArray::asDataTable() is called, it adds the row as a regular row and not as a summary row.

mattab: Ok, wondering why is this logic not needed for other plugins?

diosmosis: It is required, Actions does it here: https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/Actions/ArchivingHelper.php#L383

It's not used in Events/Contents, but that's likely a bug that wasn't dealt with since it's not tested in BlobReportLimitingTest. I'll add it to my list to fix that.

mattab: @diosmosis do you mind creating an issue for this bug in case it later slips off the list? unless you will open the PR this week or so. Thanks!

PR #14560, opened 2019-06-23
Hide date selector when viewing live visits (not as widget)

### LINKED ISSUE #14517: In Visitors > Real-time report, hide the calendar
In Visitors > Real-time report, the report is real-time and it's not possible to change the date, so let's hide the calendar. This will be consistent with how it works on Visitors>Real-time map report.

Screenshot from 2019-06-10 15-53-29

Screenshot from 2019-06-10 15-53-22

sgiehl:
fixes #14517

### DIFF
--- a/plugins/CoreHome/angularjs/period-selector/period-selector.directive.js
+++ b/plugins/CoreHome/angularjs/period-selector/period-selector.directive.js
@@ -12,9 +12,9 @@
 (function () {
     angular.module('piwikApp').directive('piwikPeriodSelector', piwikPeriodSelector);
 
-    piwikPeriodSelector.$inject = ['piwik'];
+    piwikPeriodSelector.$inject = ['piwik', '$rootScope'];
 
-    function piwikPeriodSelector(piwik) {
+    function piwikPeriodSelector(piwik, $rootScope) {
         return {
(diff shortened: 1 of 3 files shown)


diosmosis:
Works for me. Only thing that worries me is if some error occurs and `_destroy()` isn't called. This would result in the period selector being hidden, which could be an annoying bug to track down.

sgiehl:
@diosmosis would it be better like it's done in the real time map:
https://github.com/matomo-org/matomo/blob/8f77ece35b3058251c4450b6d15f3d73df8bd825/plugins/UserCountryMap/javascripts/realtime-map.js#L63-L71

diosmosis:
Huh, I would expect that might cause problems w/ this change actually, since I think the event is never removed from `$rootScope`?

sgiehl:
I actually didn't have problems with that while testing. Guess the `piwikPageChange` event is triggered before the widgets are created or destroyed. 
I'll try to implement something else, that simplifies both implementations....

sgiehl:
@diosmosis tried to improve & simplify the code. Could you have another look?

diosmosis (approved):
haven't tested, but this looks great

PR #14566, opened 2019-06-24
JS Tracker: URL Custom Dimension values to prevent Error 400 issue on IE11 when dimension values include non ascii characters

mattab:
Issue found & Patch suggested by Customer


mattab:
Be great if you can review & check tests pass & merge this in 3.10.0 @diosmosis @tsteur as the bug is causing JS errors in IE11 when custom dimension values have non ascii characters.

diosmosis:
If a user is currently using this w/ encoded parameters can they encode the parameter values themselves to workaround this? And if so, would this change break that use?

tsteur:
@mattab JS tests are passing now

PR #14569, opened 2019-06-24
Send tracking code JS by email

### LINKED ISSUE #12630: Send tracking code instructions to a developer by email

katebutler:
Have implemented this for the "no data tracked yet" page and page 7 of the installation process. Now trying to figure out an elegant solution for the tracking codes page in the admin panel - the JS is generated dynamically on the client-side here so would need to update the mailto body dynamically too, also seems like some of the other content from the email would be unneeded in this case.

Fixes #12630


plugins/SitesManager/templates/_displayJavascriptCode.twig
```diff
@@ -31,6 +31,11 @@
     <p>{{ 'CoreAdminHome_HttpTrackingApiDescription'|translate('<a href="https://developer.matomo.org/api-reference/tracking-api" rel="noreferrer noopener" target="_blank">','</a>')|raw }}</p>
     <p></p>
 
-    {{ postEvent('Template.endTrackingHelpPage') }}
+    {% set emailSubject = 'Matomo' ~ 'General_JsTrackingTag'|translate %}
+    <a class="btn"
+       href="mailto:?subject={{ emailSubject|url_encode }}&body={{ emailBody|replace({">\n<": "><"})|url_encode }}">
```
tsteur: haven't really looked at the PR yet. Would it be possible to not use this replace here and instead directly remove the HTML in `_trackingCodeEmail.twig`?

katebutler: It's stripping out the newlines between tags - I could minify the template instead but I think that would make it very hard to read.

katebutler:
I've written a spec for the no data page (which provide coverage for the template) - but none for the tracking code generator yet. This seems to have pretty minimal coverage (just one spec in UIIntegration to check that the page loads) so maybe I should add some more while I'm here?

plugins/SitesManager/templates/_displayJavascriptCode.twig
tsteur: What I meant is it's not possible to add HTML tags in a mailto link anyway? And the line breaks can be taken over from the twig template?

katebutler:
Works in Thunderbird but some email clients (e.g. Outlook, Apple Mail) print the actual HTML tags in the message body rather than rendering it.  [This answer](https://stackoverflow.com/questions/5620324/mailto-link-with-html-body/46699855#46699855) suggests a solution that works for Outlook but will still not work intuitively in Apple. The RFC says that the body is intended to be plaintext only, so I don't think it is actually possible to use a mailto link to generate an email that will render as HTML in all clients.
Instead I think we are going to need a much simpler email (just the actual JavaScript and maybe one link to the tracker code generator). The `<script type="text/javascript">` will need to be removed off the JS block due to the inconsistency between clients - if they are escaped then Apple Mail prints the escaped version, but if they aren't then Thunderbird renders it like an actual `<script>` tag so that it doesn't appear in the message body.

tsteur:
Would it help to put a "<pre>" around it? We would then need to say something like "Without the <pre>" though... but at least we could have maybe the full tag in there?

tsteur:
@katebutler is this still WIP or can it be reviewed?

katebutler:
Have rewritten the email templates in plain text.

plugins/SitesManager/templates/siteWithoutData.twig
```diff
@@ -89,6 +89,11 @@
             <a href="{{ linkTo({module: 'SitesManager', action: 'ignoreNoDataMessage'}) }}"
                class="btn ignoreSitesWithoutData">{{ 'SitesManager_SiteWithoutDataIgnoreMessage'|translate }}</a>
 
+            <a class="btn" id="emailTrackingCodeBtn"
```
tsteur: This is what the mail looks for me here: 
image

There seem to be some escaped HTML elements, and some closing `</p>` elements visible. 

We should probably remove the link `You can customise this tracking code at https://...` because the developer might not have access to it? Could also leave it there though.

In general it be great if we could also mention the alternative tracking methods in all the different tracking code emails like:

```
....Here the text we already have in the mail....

------------------------------

** Further ways of tracking

*** Integrations
In most websites, blogs, CMS, etc. you can use a pre-made plugin to do the technical work for you:
https://matomo.org/integrate/

*** Log Analytics
If the JavaScript tracking method isn’t feasible, you can use server log analytics as an alternative method for tracking your website’s users.
https://matomo.org/log-analytics/

*** Mobile apps and SDKs
Not tracking a website? You can alternatively track a mobile app or any other type of application using one of the available SDKs.
https://matomo.org/integrate/#programming-language-platforms-and-frameworks

*** HTTP Tracking API
The HTTP Tracking API allows you to track anything. This may be useful if you are using a programming language for which no SDK exists yet. It may also be useful when you want to track devices or application in a special way.
https://developer.matomo.org/api-reference/tracking-api
```

To keep things simple let's maybe not use the Tag Manager for now since a developer would likely not have access to configure things anyway. We will instead need eventually a different email button for Tag Manager.

@katebutler  WhiteLabel has a setting to remove links named `removeLinksToMatomo`. We will need to test what happens if this is enabled. I suppose the links wouldn't be removed in this case since it is plain HTML. At the same time you can't know in core if this setting is enabled or not. We may need an event that WhiteLabel can listen to like:

```php
$showDifferentWaysOfTracking = true;
Piwik::postEvent('SitesManager.showDifferentWaysOfTrackingInTrackingCodeMail', array(&$showDifferentWaysOfTracking));
// then only include the different ways if `$showDifferentWaysOfTracking === true`
```

tsteur (changes requested):
Left a few comments @katebutler 

Wondering in general how we can make this button more visible? Maybe would somehow need to show it next the the headline on the right?

Eg 
image

vs 
image

and here maybe have the button next to the headline and in the bottom?

image

and here we show it only in the top? 
image


Not sure how easy it is to show this button next to the title.

core/Tracker/TrackerCodeGenerator.php
```diff
@@ -267,4 +267,19 @@ private function getJavascriptTagOptions($idSite, $mergeSubdomains, $mergeAliasU
...
+     * ensures consistent behaviour between mail clients that render the mailto body as plain text (as in the
+     * spec), and those which try to render it as HTML and therefore hide the tags.
+     * @param string $jsTrackingCode JS tracking code as returned from the generate() function.
+     * @return string
+     */
+    public static function stripTags($jsTrackingCode)
+    {
+        // Strip off open and close <script> tag and comments so that JS will be displayed in ALL mail clients
+        $rawJsTag = html_entity_decode($jsTrackingCode);
+        $rawJsTag = preg_replace('/<[^>]+>/', '', $rawJsTag);
```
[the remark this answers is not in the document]
tsteur: ping @katebutler  also maybe it's not needed anymore so much? I reckon the problem with strip tags is that it removes the content itself?

plugins/SitesManager/templates/siteWithoutData.twig
```diff
@@ -89,6 +89,11 @@
...
+               href="mailto:?subject={{ 'SitesManager_EmailInstructionsSubject'|translate|url_encode|e('html_attr') }}&body={{ emailBody|replace({">\n<": "><"})|url_encode|e('html_attr') }}">
```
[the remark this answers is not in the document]
tsteur: Is the replace still needed?

plugins/SitesManager/templates/_displayJavascriptCode.twig
```diff
@@ -31,6 +31,10 @@
     <p>{{ 'CoreAdminHome_HttpTrackingApiDescription'|translate('<a href="https://developer.matomo.org/api-reference/tracking-api" rel="noreferrer noopener" target="_blank">','</a>')|raw }}</p>
     <p></p>
 
-    {{ postEvent('Template.endTrackingHelpPage') }}
+    <a class="btn"
```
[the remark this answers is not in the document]
tsteur: @katebutler not sure if you checked out?

tsteur:
@katebutler not sure you had a look at this comment? https://github.com/matomo-org/matomo/pull/14569#discussion_r308049483 just tried it and it wasn't showing any more text than before

PR #14571, opened 2019-06-25
Fix error number is detected wrongly in tracker mode

tsteur:
Eg when the error is `Error query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry`, it would detect 2300 instead of 1062.
fix DEV-1683

### DIFF
--- a/core/Tracker/Db/Pdo/Mysql.php
+++ b/core/Tracker/Db/Pdo/Mysql.php
@@ -260,9 +260,10 @@ class Mysql extends Db
      */
     public function isErrNo($e, $errno)
     {
-        if (preg_match('/([0-9]{4})/', $e->getMessage(), $match)) {
+        if (preg_match('/(?:\[|\s)([0-9]{4})(?:\]|\s)/', $e->getMessage(), $match)) {
             return $match[1] == $errno;
         }
+
         return false;
     }
 
--- a/core/Tracker/Db/Pdo/Mysql.php
+++ b/core/Tracker/Db/Pdo/Mysql.php
@@ -260,11 +260,7 @@ class Mysql extends Db
      */
     public function isErrNo($e, $errno)
     {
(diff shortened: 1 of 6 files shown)


tsteur:
Now PDO and MySQLi uses same code in tracker and regular mode. Ideally would refactor it to use the same method in the background. Not sure if I find time.

tsteur:
Now reusing same code and added a basic test

core/Tracker/Db/Pdo/Mysql.php
```diff
@@ -260,10 +260,7 @@ public function lastInsertId()
      */
     public function isErrNo($e, $errno)
     {
-        if (preg_match('/([0-9]{4})/', $e->getMessage(), $match)) {
-            return $match[1] == $errno;
-        }
-        return false;
+        return \Piwik\Db\Adapter\Pdo\Mysql::isPdoErrorNumber($e, $errno);
```
diosmosis: Does this mean the `\Piwik\Db\Adapter\Pdo\...` classes need to be required in `/piwik.php` now?

mattab: will merge now but @tsteur if this needs an update please go ahead and merge it in a separate PR

tsteur: @diosmosis not sure what you mean? The files will be auto loaded if that's what you mean?

diosmosis: I thought autoloading wasn't used in the tracker side for performance (hence the `require_once`s in `PIWIK_INCLUDE_PATH/piwik.php`). I guess it is used?

tsteur: It is definitely used as it's needed for all the plugins etc. those requires could likely be removed.

PR #14578, opened 2019-06-27
Provide methods to force the update of tracker caches

tsteur:


tsteur:
Be great if this could make it to 3.10 but we can also move it to 3.11 and patch it @diosmosis @mattab @sgiehl

diosmosis (approved):
LGTM

PR #14581, opened 2019-06-27
Truncate very long useragents to prevent possible tracking slowdown

sgiehl:


tsteur:
Changed it to Common::mb_substr. Looks good to me if tests pass and could include it in 3.10 I suppose. Or in 3.11...

Issue #14535, opened 2019-06-17
log_visit.referer_url and log_action.name as TEXT instead of VARCHAR cause a lot of tmp table on disk


airblag:
Hi,

My matomo installation is "big" (log_link_visit_action is ~40GB big for 55 days retention, log_visit is ~4GB), and we see over 97% of temporary tables created on disk in our mysql monitoring.

After trying to tune mysql with increasing tmp_table_size without any sucess I noticed that the log_visit table has the referer_url field set as TEXT.

it blocks the use of the memory engine of mysql for tmp tables writing all of them to disk :

https://dev.mysql.com/doc/refman/5.7/en/blob.html

"[...]
 Instances of BLOB or TEXT columns in the result of a query that is processed using a temporary table causes the server to use a table on disk rather than in memory because the MEMORY storage engine does not support those data types (see Section 8.4.4, “Internal Temporary Table Use in MySQL”). Use of disk incurs a performance penalty, so include BLOB or TEXT columns in the query result only if they are really needed. For example, avoid using SELECT *, which selects all columns. 
[...]"

I'm not familiar with the code of matomo to look for the exact query producing the tmp_tables...

Is it thinkable to somehow convert it to varchar(65535) to avoid having all tmp_tables on disk ? Or would it have a lot of side effects ?

I tried to move the tmpdir of mysql to ramdisk, but since I try to optimize the DB once a week, and optimizing innodb needs a lot of place on /tmp it's always failing to optimize log_link_action_visit except I give more 40GB for my ramdisk...

tsteur:
@mattab that's interesting. Maybe would also help with `log_action.name`?

mattab:
Thanks for the report @airblag 

> Is it thinkable to somehow convert it to varchar(65535) to avoid having all tmp_tables on disk ? Or would it have a lot of side effects ?

Reckon that it should be fine for more than 99.9% of cases to have only 65535 character long URLs. 
@tsteur Maybe something worth investigating?

tsteur:
@mattab `text` also has only `65,535` characters

tsteur:
We will schedule this for Matomo 4. Thanks for this @airblag

tsteur:
@mattab I'm thinking we could actually already apply this change for new installs and only include the update in Matomo 4. This will save many thousands of migrations later.

airblag:
[...]

PR #14584, opened 2019-06-28
avoid using group by and select min for tracker actions query to avoid tmp tables refs #14535

tsteur:
The tests passed for me locally. Be good if someone can confirm this will end up being the same logic.

refs https://github.com/matomo-org/matomo/issues/14535


tsteur:
@diosmosis @mattab could you both have a look at this ideally?

diosmosis:
Looks ok to me

PR #14590, opened 2019-06-29
Activate Matomo Tour feature for all users

tsteur:


mattab:
Let's enable the Tour plugin for everyone in 3.11.0 after we got some initial feedback from new users :ok_hand: 
(Tour plugin will be enabled for new users).

PR #14591, opened 2019-06-29
Improve segments tests

sgiehl:
Some of the segments tests actually didn't test with a valid value, as the autosuggest returns a full list of possible values instead of a list of occurring values. I've changed the test to use some specific values instead for those segments.  The tests for`customVariablePageName4` and `eventValue` are still returning empty visitsummaries. Maybe we should check if that is correct, as they are using values that seems valid...


diosmosis (approved):
LGTM, good to have useful tests

Issue #12691, opened 2018-04-01
Some tests on travis are running too long


sgiehl:
In most cases the build is currently failing as the system tests for mysqli get terminated after 40min.
For testing purpose I added a small script to check the running times of tests > 5 sec:

https://travis-ci.org/matomo-org/matomo/jobs/360648037

Tests running very long are: 
- Test `Piwik\Tests\System\ArchiveCronTest::testArchivePhpCron` took 438.5 seconds
- ​​​​​​​​Test `Piwik\Tests\System\ArchiveWebTest::test_WebArchiveScriptCanBeRun_WithPhpCgi_AndWithoutTokenAuth` took 466.8 seconds

Setup of the fixture `MaySitesImportedLogs` only takes ~30 sec, but contains a lot of data. Maybe we should use a smaller fixture for those tests to run faster?

And some of the API tests `Piwik\Tests\System\AutoSuggestAPITest::testAnotherApi` take alomst 60sec

sgiehl:
ArchiveWebTest moved to Integration tests for now, so tests are passing again. Nevertheless we should check if we can improve the Archiving tests to run faster...

mattab:
As found by Phpunit speedtrap in https://github.com/matomo-org/matomo/pull/12763#issuecomment-383531452 the current slowest tests are:

```
You should really fix these slow tests (>500ms)...
 1. 215204ms to run Piwik\Tests\System\ArchiveCronTest:testArchivePhpCron
 2. 59605ms to run Piwik\Tests\System\TwoVisitorsTwoWebsitesDifferentDaysTest:testApi with data set #7
 3. 54062ms to run Piwik\Tests\System\AutoSuggestAPITest:testAnotherApi with data set #92
 4. 51416ms to run Piwik\Tests\System\AutoSuggestAPITest:testAnotherApi with data set #96
 5. 51372ms to run Piwik\Tests\System\AutoSuggestAPITest:testAnotherApi with data set #90
 6. 50215ms to run Piwik\Tests\System\AutoSuggestAPITest:testAnotherApi with data set #74
 7. 48881ms to run Piwik\Tests\System\AutoSuggestAPITest:testAnotherApi with data set #72
 8. 48671ms to run Piwik\Tests\System\AutoSuggestAPITest:testAnotherApi with data set #94
 9. 47777ms to run Piwik\Tests\System\AutoSuggestAPITest:testAnotherApi with data set #66
[...]

PR #14605, opened 2019-07-01
Improve speed of serving unreadable files

sgiehl:
While looking for long running Integration tests (#12691) I saw, that the test `Piwik\Tests\Integration\ServeStaticFileTest:test_nonReadableFile` takes around 65s on travis, as reading the file without permission seems to take a long time.
Adding a check for `is_readable` speeds up the test to less than 10 seconds (guess even less, but I used 10 seconds for reporting slow tests).

Not sure if that might decrease the speed for existing files as there is another file operation to be done...

### DIFF
--- a/core/ProxyHttp.php
+++ b/core/ProxyHttp.php
@@ -73,6 +73,11 @@ class ProxyHttp
             return;
         }
 
+        if (!is_readable($file)) {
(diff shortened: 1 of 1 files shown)


tsteur:
Feel free to merge if it increases the speed 👍

PR #14616, opened 2019-07-02
If URL starts with // in sites manager, only prepend http: instead of… http://

### LINKED ISSUE #14615: Bug: Website Measurable URLs unexpected behavior
# Summary

During the configuration of the website measurable, you have to specify one or several URLs.

## Current behavior

If I specify an URL like `https://mysite.com`, everything is ok.

But, if I use an URL like `//mysite.com` to avoid specify `https://` or `http://`:

* from `All Websites` page, **Visits** and **pageviews** keep working and increment values
* No Widgets from the Dashboard display new values (ex: real time count still at 0)
* When I come back to the configuration, the URL become `////mysite.com`

### First scenario

Display an error message if the URL is not valid (URL validator)

### Second scenario

Allow to use URLs that begin by `//` (without `https` or `http`) and support them into data incrementation.

tsteur:
fix https://github.com/matomo-org/matomo/issues/14615

alternatively we could throw an exception if URL starts with `//`. Configuring a URL with `//` is not supported currently.

### DIFF
--- a/plugins/WebsiteMeasurable/Settings/Urls.php
+++ b/plugins/WebsiteMeasurable/Settings/Urls.php
@@ -115,7 +115,11 @@ class Urls extends \Piwik\Settings\Measurable\MeasurableProperty
             if (empty($scheme)
                 && strpos($url, '://') === false
             ) {
-                $url = 'http://' . $url;
+                if (strpos($url, '//') === 0) {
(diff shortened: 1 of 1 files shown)


diosmosis (approved):
LGTM

Issue #14367, opened 2019-04-23
Config INI files are parsed on EVERY tracking request


soukicz:
Hi,
I am getting 100% CPU usage from tracking script on 3.9.1 with PHP 7.3. I have tried to disable tracking in config but CPU is still 100% (after webserver restart).
I have tried to run Blackfire on tracking script and it shows that 38% of time is spent on INI files parsing.
It seems pointless to parse the files on every request. Is it a bug or is it by design that there is no cache?

https://blackfire.io/profiles/b13eb852-e071-470f-8a96-d3690e5e5d9d/graph

tsteur:
There is no caching across requests on purpose. Ini parsing is fast compared to the whole tracking request. It looks like you weren't actually profiling a proper tracking request. At least I cannot find any queries to the DB etc and a lot of other calls that would be usually executed.

soukicz:
Yes, that percentage isn't relevant with normal request. But it is still pointless 6ms per request on i9-9900k CPU. That doesn't sound like a lot but for a file that is executed 500 times per second it sums up.

Do I undestand correctly that I shouldn't bother with pull request?

tsteur:
Right now we wouldn't really be interested. We could potentially use it optionally but people would need to remember to invalidate caches when changing the config file etc. and it gets more difficult when using multiple servers and you have maybe the config file in an EFS/shared drive and need to invalidate all local caches etc. A workaround could be to build the cache key based on the `md5_file($configFile)` or so. It's in general important to invalidate the cache as soon as the config changes especially when enabling maintenance mode / disable tracking as otherwise you risk running into big trouble.

Maybe check if you find there's a big performance improvement when caching eg content based on a key built with eg `md5_file()`. In general you really want to test the true performance of it though eg using the below code as profilers are always a bit of when some methods have a lot of calls in my experience (the ratio is more or less the same though).

```php
    public function readFile($filename)
    {
        $timer = new Timer();
        $ini = $this->getContentOfIniFile($filename);

        $test = $this->readString($ini);
        echo $timer->getTimeMs() . 'ms'; exit;
    }
```

On my slow local computer, with a rather slow HD, it takes on average 3ms for me and it makes only 1% of total request time. I think on our production server it's also only about 1%.

tsteur:
On our production server it takes 0.4ms.

fdellwing:
`microtime(true)` says it takes <0.1ms, so really not a long time at all.

---

[...]

PR #14617, opened 2019-07-02
Support to cache the config file

tsteur:
refs https://github.com/matomo-org/matomo/issues/14367

This supports caching the config ini file. It would be enabled through `$matomoDir/bootstrap.php` by putting this code in there (we cannot use DI or config file for this because they don't exist at the time of the config loading):

```php
<?php

$GLOBALS['ENABLE_CONFIG_PHP_CACHE'] = true;
$GLOBALS['MATOMO_PLUGIN_DIRS'] = array(); // this line is unrelated... just another setting we added recently by using bootstrap.php
```

It would pretty much only work for single server environments... all others would have the problem that the cache invalidation when the config changes wouldn't be synced. Also we do not want to have to document that users need to invalidate the cache when they make changes to it. It would pretty much only help us and I would probably not even document it. When loaded from a regular filesystem, it should be still very fast with the regular ini. If it works nicely, we could document it eventually.

By caching it, it would come from opcache instead of filesystem... and we would not need to do things like all these decodings etc https://github.com/matomo-org/component-ini/blob/master/src/IniReader.php#L149-L384

Any thoughts on this @mattab 

refs DEV-1692

### DIFF
--- a/core/Config/Cache.php
+++ b/core/Config/Cache.php
@@ -34,14 +34,11 @@ class Cache extends File
 
     private function makeCacheDir($host)
     {
-        return PIWIK_INCLUDE_PATH . '/tmp/' . $host . '/cache/tracker/';
+        return PIWIK_INCLUDE_PATH . '/tmp/' . $host . '/cache/tracker';
     }
 
     public function isValidHost($mergedConfigSettings)
     {
-        if (!$this->host) {
-            return true;
-        }
         if (!isset($mergedConfigSettings['General']['trusted_hosts']) || !is_array($mergedConfigSettings['General']['trusted_hosts'])) {
             return false;
         }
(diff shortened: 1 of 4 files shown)


tsteur:
As discussed we're wanting to give this a try.

core/Config/Cache.php
```diff
@@ -0,0 +1,89 @@
...
+        // because the config is not yet loaded we cannot identify the instanceId...
+        // need to use the hostname
+        $dir = $this->makeCacheDir($this->host);
+
+        parent::__construct($dir);
+    }
+
+    private function makeCacheDir($host)
+    {
+        return PIWIK_INCLUDE_PATH . '/tmp/' . $host . '/cache/tracker';
```
diosmosis: Should we check somewhere if this directory is writable?

tsteur: @diosmosis the directory might only be created when actually caching a file (the file backend does that). We already check if `PIWIK_INCLUDE_PATH . '/tmp/'` is writable as part of the system check not sure if it's enough?

diosmosis: That sounds like enough 👍

PR #14620, opened 2019-07-03
Remove "open source web analytics" tooltip in app logo when a custom logo is uploaded

tsteur:
refs DEV-1702
This way it won't be shown when using eg white label etc.


diosmosis (approved):
LGTM

Though we could also replace this in whitelabel correct? I assume the text isn't that important to keep, but thought I'd mention this possibility.

PR #14622, opened 2019-07-03
 Handle ranking query others row properly in contents & events archiver.

diosmosis:


core/DataArray.php
```diff
@@ -424,7 +424,7 @@ public static function isRowActions($row)
      *
      * @return \Piwik\DataTable
      */
-    public function asDataTable()
+    public function asDataTable($arrayHasSummaryRow = false)
```
tsteur: This will break things for our premium features which partially overwrite this method. Might be better to create a new method for this ?

core/DataTable.php
```diff
@@ -1573,7 +1576,12 @@ public static function makeFromIndexedArray($array, $subtablePerLabel = null)
             if (isset($subtablePerLabel[$label])) {
                 $cleanRow[Row::DATATABLE_ASSOCIATED] = $subtablePerLabel[$label];
             }
-            $table->addRow(new Row($cleanRow));
+
+            if ($arrayHasSummaryRow && $label == self::LABEL_SUMMARY_ROW) {
```
tsteur: Do we need `$arrayHasSummaryRow ` because `LABEL_SUMMARY_ROW` could be an actual label within the table?

diosmosis: Yes.

tsteur: Not sure I'm getting it fully. If there's a label `-1` then it would still aggregate it maybe wrongly? It be only helping if there are reports that never have a summary row but potentially a `-1` value? Which might be rarely the case? So maybe we could just execute it always? Not sure I'm fully understanding it

diosmosis: So DataTable identifies the summary row by ID, not by label (the ID is also -1). I assume this is because the label could be -1 and still be a valid dimension value (eg, for custom variable value/custom dimension). So we can't do this in every case, only when using ranking query and we know it's the right label. Alternatively, we could change the others label in RankingQuery to something like `__ranking_query_others__`, and always check for that.

tsteur: Not quite sure I understand as you mention the row by ID. Or maybe I wasn't clear what I meant. I reckon we could change 

`if ($arrayHasSummaryRow && $label == self::LABEL_SUMMARY_ROW)`
to 
`if ($label == self::LABEL_SUMMARY_ROW)`

and it wouldn't really make a difference?

diosmosis: It would make a difference because the label could legitimately be `-1` in certain cases, like custom variable value / custom dimension value. Not sure about other reports, but this would essentially be saying that `-1` could never be used as a dimension value for any dimension, because it is used to denote a summary row.

tsteur: Wouldn't CustomVariables or CustomDimensions also set `$arrayHasSummaryRow = true` because we also want to group these? Technically it could have a summary row as well?

diosmosis: Any report that uses RankingQuery would use `$arrayHasSummaryRow`, however, it would be a problem for reports where `-1` could be a valid label in the report.

tsteur: OK 👍  thought custom variables and custom dimensions would maybe also already use rankingQuery but they don't. Eventually all archivers should probably use the ranking query though (not going to happen any time soon though so can be ignored probably).

Custom Reports uses ranking query and possibly `-1` could be a valid label but I suppose we can't avoid this unless as you say we set it to something like `__ranking_query_others__ ` at some point. But even then I'm not sure it can be actually avoided so much, unless we also archived it using that label.

diosmosis: This code is executed before archiving so a special, very unlikely label like `__ranking_query_others__` should work (the summary row w/ ID = -1 is what is archived). I can look into that.

diosmosis:
@tsteur updated.

plugins/Actions/Archiver.php
```diff
@@ -200,7 +200,6 @@ protected function archiveDayActions($rankingQueryLimit, array $actionTypes, $in
         $rankingQuery = false;
         if ($rankingQueryLimit > 0) {
             $rankingQuery = new RankingQuery($rankingQueryLimit);
-            $rankingQuery->setOthersLabel(DataTable::LABEL_SUMMARY_ROW);
```
tsteur: I suppose we will need to do the same for Custom Reports where we currently use `-1`. We could simply change it depending on the Matomo version.

plugins/Ecommerce/tests/System/expected/test_ecommerceOrderWithItems_schedrep_html_tables_and_graph__ScheduledReports.generateReport_week.original.html
```diff
@@ -3408,7 +3408,7 @@ <h2 id="CustomVariables_getCustomVariables" style=" color:#0d0d0d;font-family:-a
                                                     
                                                                     <tr style=";">
                                                                 <td style="padding:17px 15px;;border-bottom:1px solid rgb(231,231,231);font-size: 15px;font-variant-numeric: tabular-nums;padding:17px 15px;color:#0d0d0d;font-family:-apple-system, BlinkMacSystemFont, &#039;Segoe UI&#039;, Roboto, Oxygen-Sans, Cantarell, &#039;Helvetica Neue&#039;, sans-serif;">
-                                                                                                                                                                        ValueIsZero - 0                                                                                                                        </td>
+                                                                                                                                                                        ValueIsZero - Others                                                                                                                        </td>
```
tsteur: @diosmosis this is bit unexpected maybe? I can see here we set the value to `0` https://github.com/matomo-org/matomo/blob/3.11.0-b1/tests/PHPUnit/Fixtures/TwoSitesEcommerceOrderWithItems.php#L68 so would have expected to see that value? Or is it because of https://github.com/matomo-org/matomo/pull/14622/files#diff-64a9dbfd70403b4ac01ee0d4cfd1894cR185 being set to 3/2 rows truncation and it's actually grouping values here? Actually all good... I can see here (https://github.com/matomo-org/matomo/pull/14622/files#diff-05a120dee88125d268f6cd302e02d6bdL161) there are already 2 other rows in the subtable so it is expected that it is grouped.

tsteur:
LGTM @diosmosis only need to fix a test in AnonymousUsageMeasurement afterwards see https://travis-ci.org/matomo-org/matomo/jobs/557605365#L809

tsteur:
There are now also some merge conflicts unfortunately.

PR #14624, opened 2019-07-04
various performance tweaks

tsteur:


core/Option.php
```diff
@@ -178,10 +178,6 @@ protected function getValue($name)
         $value = Db::fetchOne('SELECT option_value FROM `' . Common::prefixTable('option') . '` ' .
                               'WHERE option_name = ?', $name);
 
-        if ($value === false) {
```
tsteur: Initially I didn't get why there would be per request eg 140 calls to `Option::get()` from like 15 different places and 79 calls where actually hidding the DB. Until I realised for many of these values there is no value configured on `option` table and it would fetch the value from the DB again and again. Some option names it was fetching over 30 times.

tsteur: If a value is set later, it should go through ::set() anyway and be invalidated.

core/Plugin.php
```diff
@@ -438,6 +438,10 @@ public function getMissingDependencies($piwikVersion = null)
             return array();
         }
 
+        if (!SettingsPiwik::isSystemValidationEnabled()) {
```
tsteur: Small tweak to not having to execute this on very request as we know on production there are no missing dependencies. Happy to change the name of the setting. We might want to add something like this eventually to more places as we know our system is correctly configured. Saves eg 1% per tracking request.

core/Plugin/Manager.php
```diff
@@ -319,12 +319,14 @@ public function isPluginLoaded($name)
      */
     public function readPluginsDirectory()
     {
+        $isSystemValidationEnabled = SettingsPiwik::isSystemValidationEnabled();
```
tsteur: Same here as in https://github.com/matomo-org/matomo/pull/14624/files#r300292474

We know the plugin structures are valid and therefore we don't want to execute this on every request as it's a lot of stat calls.

core/SettingsPiwik.php
```diff
@@ -265,7 +265,18 @@ public static function isInternetEnabled()
...
+
+    /**
+     * If enabled, will skip some checks when bootstrapping Matomo for faster performance. Eg we won't check if
+     * plugin structures are valid.
+     *
+     * @return bool
+     */
+    public static function isSystemValidationEnabled()
+    {
+        return (bool) Config::getInstance()->General['enable_system_validation'];
```
tsteur: as mentioned happy to change name of the setting

core/View.php
```diff
@@ -266,8 +266,7 @@ public function render()
 
             $this->loginModule = Piwik::getLoginPluginName();
 
-            $user = APIUsersManager::getInstance()->getUser($this->userLogin);
-            $this->userAlias = $user['alias'];
+            $this->userAlias = $this->userLogin; // can be removed in Matomo 4.0
```
tsteur: We're not using the alias anymore, and we will even remove it in Matomo 4. Couldn't find any usage in our plugins of this...

Before we were requesting the `user` DB on every VIEW render. Especially when we render the visitor log we might have hundreds or thousands of VIEW renders saving hundreds or thousands of calls to the `user ` DB

core/Plugin/Manager.php
```diff
@@ -319,6 +319,10 @@ public function isPluginLoaded($name)
      */
     public function readPluginsDirectory()
     {
+        if (!empty($GLOBALS['MATOMO_PLUGIN_NAMES_AVAILABLE'])) {
```
tsteur: this along with the other related changes saves 1-2% of each tracking request. Eventually we could add more "caching" around this.

tsteur: In `$matomoDir/bootstrap.php` someone could basically configure which plugin names are available in an array.

PR #14627, opened 2019-07-05
Tweaks to Readme, updating links

mattab:


README.md
```diff
@@ -50,17 +50,19 @@ See https://matomo.org/docs/requirements/
 
 See https://matomo.org/docs/installation/
 
+(Alternatively you could also [deploy Matomo from the Git repository.](https://matomo.org/faq/how-to-install/faq_18271/))
```
tsteur: This we don't want to recommend anymore, only for development. We even mention it on the page itself.

tsteur:
Looks good besides the install from git repository

PR #14628, opened 2019-07-05
When installer is disabled and any module is requested, also show the not set up yet message

tsteur:
Expected (already working when requesting https://analytics.domain.com )

image

but doesn't work when eg requesting https://analytics.domain.com/index.php?module=API&action=home&idSite=1&period=day&date=yesterday

which was showing eg this

image

Adding another check for this case when a module is specified...  In the past we didn't want find a need to fix it but it's easy to do. A test for this special case shouldn't be needed.

### DIFF
--- a/core/Application/Kernel/EnvironmentValidator.php
+++ b/core/Application/Kernel/EnvironmentValidator.php
@@ -9,6 +9,7 @@
 namespace Piwik\Application\Kernel;
 
 use Piwik\Common;
+use Piwik\Config;
 use Piwik\Exception\NotYetInstalledException;
 use Piwik\Filechecks;
 use Piwik\Piwik;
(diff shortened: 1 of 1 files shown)


core/Application/Kernel/EnvironmentValidator.php
```diff
@@ -75,6 +76,12 @@ private function checkConfigFileExists($path, $startInstaller = false)
             return;
         }
 
+        $general = Config::getInstance()->General;
```
diosmosis: Does it work to access this here? This could be called before StaticContainer is initialized. Might be better to use `$this->settingsProvider`.

tsteur: @diosmosis done. It worked with the config as well but this way I suppose it is better 👍

PR #14631, opened 2019-07-05
Update GitHub links in tests/README.md

obendev:
Note:
Two of the links link to a 404. However I didn't remove them, because I didn't know what way you like to handle such a case.
1. https://github.com/matomo-org/matomo/blob/master/tests/PHPUnit/config.ini.travis.php#L23-27
(Line 156)
2. https://github.com/matomo-org/matomo/blob/master/tests/README.xhprof.md
(Line 179)

I still think that adjusting these links is right however because `/piwik/` will be a 404 someday anyway.


tsteur:
Thanks again 👍

PR #14635, opened 2019-07-05
Update the link tag for all php files

obendev:
I updated the link tag for all php files with the new matomo link.
Not sure if it is realistic that this will get merged but I thought that someday someone has to do this(?)

### DIFF
--- a/core/API/ApiRenderer.php
+++ b/core/API/ApiRenderer.php
@@ -2,7 +2,7 @@
 /**
  * Piwik - free/libre analytics platform
  *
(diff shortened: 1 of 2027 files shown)


tsteur:
It is very realistic this will get merged as I will do it right now :) 👍  thanks for this, very appreciated. We'll eventually also need to update `Piwik - free/libre analytics platform` => `Matomo - free/libre analytics platform`.
Btw if someone wants to see the diff easily: https://github.com/matomo-org/matomo/pull/14635.diff

PR #14637, opened 2019-07-05
Update GitHub links in LEGALNOTICE

obendev:
That should be all.
There are still some old links in the CHANGELOG.md, but I think we shouldn't touch them, because changelog is what it is, what do you think about it @tsteur ?

Note:
Do you still use jquery.truncate?
It is in the LEGALNOTICE, but I think it isn't used anymore (not sure).

### DIFF
--- a/LEGALNOTICE
+++ b/LEGALNOTICE
@@ -40,8 +40,8 @@ CREDITS
 
 	For detailed contribution history, refer to the source, tickets,
 	patches, and Git revision history, available at
-	https://github.com/matomo-org/piwik/issues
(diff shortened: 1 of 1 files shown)


tsteur:
The file for jquery truncate is still in the filesystem but I think it looks it's not used anymore.

tsteur:
CHANGELOG.md we could certainly fix links as well.

obendev:
Okay, I will change them later as well. 😊
Than there shouldn't be any old GitHub links anymore.

PR #14639, opened 2019-07-06
Remove temporary archive concept and just invalidate on tracking.

diosmosis:
CC @tsteur 

Seems like it works fine, tested manually. Automated tests shouldn't pass just yet, will work on that soon.

Refs #14091


plugins/SegmentEditor/SegmentEditor.php
```diff
@@ -61,8 +61,13 @@ public function addSystemSummaryItems(&$systemSummary)
 
     function getSegmentEditorHtml(&$out)
     {
-        $selector = new SegmentSelectorControl();
-        $out .= $selector->render();
+        try {
+            $selector = new SegmentSelectorControl();
+            $out .= $selector->render();
+        } catch (\Exception $ex) {
+            print $ex->getTraceAsString()."\n";
```
sgiehl: guess would be better to log the error instead of printing it...

diosmosis: sorry, that's was for debugging something else, accidentally committed it

diosmosis:
NOTE: this will require a change to CustomReports

tsteur:
@diosmosis sounds good to change custom reports as long as we can still support previous Matomo versions as well.

diosmosis:
@tsteur changes should be BC now. However, I noticed AbTesting does some stuff w/ ArchiveWriter that includes a temporary archive (in GetMetricsOverview.php). Will this be a problem? Not sure what the code does.

plugins/SegmentEditor/SegmentEditor.php
tsteur: btw this still needs to be removed right?

core/DataAccess/Model.php
```diff
@@ -205,20 +205,14 @@ public function deleteArchiveIds($numericTable, $blobTable, $idsToDelete)
...
-    public function getArchiveIdAndVisits($numericTable, $idSite, $period, $dateStartIso, $dateEndIso, $minDatetimeIsoArchiveProcessedUTC, $doneFlags, $doneFlagValues)
+    public function getArchiveIdAndVisits($numericTable, $idSite, $period, $dateStartIso, $dateEndIso, $doneFlags, $doneFlagValues)
     {
         $bindSQL = array($idSite,
             $dateStartIso,
             $dateEndIso,
             $period,
         );
 
-        $timeStampWhere = '';
```
tsteur: @diosmosis do we maybe need an update script to get rid of all temporary archives? Maybe some temporary archives would need to be converted to a completed archive? (not sure if that's an issue but wondering if it's possible that we were sometimes reading a temporary archive because for some reason no completed archive was created)

tsteur:
Looks promising so far @diosmosis . Should we maybe merge this with PR https://github.com/matomo-org/matomo/pull/14091 as we maybe still need the changes in `CronArchive`?

core/DataAccess/Model.php
diosmosis: This code should still look for temporary archives in addition to done archives (whichever is latest), so it shouldn't be necessary to do it now. I could add the update though.

diosmosis:
> Should we maybe merge this with PR #14091 as we maybe still need the changes in CronArchive?

@tsteur I'm not sure that PR as it is done is even needed anymore... If we only run archiving for invalidated/missing archives, then we don't even have to check for visits in log_visit. So instead of every other check, just see if there is an idarchive w/ DONE_OK for the site/period/segment/etc., and if not send the request. Otherwise, skip it. What do you think?

tsteur:
I meant eg in `CronArchiving`, before issuing an archive request in `$cronArchiver->request($url)` or so, we need to check if we actually need to archive any data. This will safe a lot of time / performance in the archiver. Eg if data has been invalidated recently, we will need to issue the request... if a completed archive is there, and no data has been invalidated, then we don't need to issue the archiving request.

diosmosis:
@tsteur Yes that's what I meant, but the previous PR checks if there are visits in log_visit. Here I'm suggesting just to check if there's an existing archive w/ `value = DONE_OK`. Which might be what you're suggesting as well? The plan was to do that in another PR.

tsteur:
Sounds good to do this in another plan, I'll review the PR again later this week then. 👍  Looks good so far though had a look already a few times.

core/DataAccess/ArchiveSelector.php
```diff
@@ -45,7 +45,7 @@ private static function getModel()
         return new Model();
     }
 
-    public static function getArchiveIdAndVisits(ArchiveProcessor\Parameters $params, $minDatetimeArchiveProcessedUTC)
+    public static function getArchiveIdAndVisits(ArchiveProcessor\Parameters $params)
```
tsteur: @diosmosis some usages still seem to pass two parameters here, eg Advertising, CustomReports, SearchEngineKeywordsPerformance, ... might be better to still keep the 2nd parameter but simply not use it. Otherwise it could break things and lead to notices etc. In Matomo 4 we got remove the 2nd parameter

plugins/SegmentEditor/SegmentEditor.php
tsteur: @diosmosis this still needs to be removed, be good to not forget.

tsteur (changes requested):

tsteur:
@diosmosis can you look at the 2 failing tests https://travis-ci.org/matomo-org/matomo/jobs/557621873#L829-L844 ? otherwise looks good to merge and work on the CronArchiver in another PR.

Also left a comment where we possibly break the API and need to put back a parameter. One print debug is also there, otherwise feel free to merge once these things are done .

core/DataAccess/ArchiveSelector.php
diosmosis: 👍

Issue #14549, opened 2019-06-20
update mocha


Findus23:
Since Mocha was added in 2016 (arround https://github.com/matomo-org/matomo/issues/8097), it hasn't been updated.

Since then three major releases have been published that might fix some bugs that occur during testing and might improve performance.
In addition, GitHub shows security issues for it (that shouldn't matter as it isn't included in the release).

Findus23:
I guess this solved itself with #14659

PR #14655, opened 2019-07-13
Bump lodash from 4.17.11 to 4.17.14 in /tests/lib/screenshot-testing

PatOnTheBack:
Automated fix by dependabot.

### DIFF
--- a/tests/lib/screenshot-testing/package-lock.json
+++ b/tests/lib/screenshot-testing/package-lock.json
@@ -1096,9 +1096,9 @@
       }
     },
     "lodash": {
-      "version": "4.17.11",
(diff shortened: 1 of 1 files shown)


Findus23:
This isn't that important as this directory isn't included in the released Matomo build, but just used for tests.
Instead all those dependencies should be updated sometime at once: #14549

PatOnTheBack:
@Findus23 Would it make sense to just merge this PR now and gradually update dependencies?

PatOnTheBack:
The CI test failed for PHP 5, not PHP 7.

sgiehl:
The library you updated is only used for screenshot testing, which is the build running on PHP 5 only. But that shouldn't be caused by this update. So guess it wouldn't harm to merge that :man_shrugging:

PR #14656, opened 2019-07-14
Migrate Tour plugin UI tests to headless Chrome...

sgiehl:
... and fix some other failing UI tests


sgiehl:
Currently there are three failing UI tests remaining. Wasn't able to find a way to fix them correctly... Might have a look again somewhen else...

PR #14657, opened 2019-07-14
Fix blocked XHR requests when widgetizing dashboard

### LINKED ISSUE #14446: 2FA Prevents dashboard from being embedded
When embedding as follows. 

https://test.innocraft.cloud/index.php?module=Widgetize&action=iframe&moduleToWidgetize=Dashboard&actionToWidgetize=index&idSite=8&period=week&date=yesterday&token_auth=xxxxxxxx

Widgets are not displayed in the iFrame, only website and segment selector. 

image

This issue occurs when a non-authenticated user attempts to view the embed. If the user is logged in the view behaves as expected.

katebutler:
Fixes #14446

### DIFF
--- a/plugins/TwoFactorAuth/TwoFactorAuth.php
+++ b/plugins/TwoFactorAuth/TwoFactorAuth.php
@@ -180,14 +180,11 @@ class TwoFactorAuth extends \Piwik\Plugin
             return false;
         }
 
-        if (Piwik::getModule() === 'Widgetize') {
-            // we cannot use $module as it would be different when dispatching other requests within the widgetized request
-            $auth = StaticContainer::get('Piwik\Auth');
-            if ($auth && !$auth->getLogin() && method_exists($auth, 'getTokenAuth') && $auth->getTokenAuth()) {
-                // when authenticated by token only, we do not require 2fa
-                // needed eg for rendering exported widgets authenticated by token
-                return false;
-            }
+        $auth = StaticContainer::get('Piwik\Auth');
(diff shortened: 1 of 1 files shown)


plugins/TwoFactorAuth/TwoFactorAuth.php
```diff
@@ -180,14 +180,11 @@ private function requiresAuth($module, $action, $parameters)
...
-            if ($auth && !$auth->getLogin() && method_exists($auth, 'getTokenAuth') && $auth->getTokenAuth()) {
-                // when authenticated by token only, we do not require 2fa
-                // needed eg for rendering exported widgets authenticated by token
-                return false;
-            }
+        $auth = StaticContainer::get('Piwik\Auth');
+        if ($auth && !$auth->getLogin() && method_exists($auth, 'getTokenAuth') && $auth->getTokenAuth()) {
+            // when authenticated by token only, we do not require 2fa
+            // needed eg for rendering exported widgets authenticated by token
+            return false;
```
sgiehl: Wouldn't that actually allow to bypass the 2fa everywhere when using the tokenauth?

tsteur: That should be fine though. You basically can't do 2FA for the token_auth and that's already the behaviour for the API as well. We just need to make sure the user is actually authenticated using the token and not through regular login/session.

Not sure but could also add another check in the `if` like `Piwik::isUserHasAtLeastSomeViewAccess()` but not sure if needed or if it makes sense.

PR #14659, opened 2019-07-15
Remove old unused test files

Findus23:
This is a WIP PR to remove all files in tests/ that are not used anymore:

- `tests/lib/jsdoc-toolkit`
- `tests/lib/chai-1.9.0`
- `tests/lib/mocha-3.1.2`


tsteur:
@Findus23 are you still working on this? If tests pass we could maybe merge?

Findus23:
While there are more things that can be removed, those will need a few changes and a closer look.

So for now this can be merged if the tests pass (as they should)

PR #14660, opened 2019-07-16
In JS tracker generator make sure to reuse same class when creating new file

tsteur:
Background:
Someone might use DI to set a custom file type like this:

```php
  'Piwik\Plugins\CustomPiwikJs\TrackerUpdater' => DI\decorate(function ($previous) {
 
        $to = new \Piwik\Plugins\MyPlugin\CdnFile('mypath/piwik.js');

        $previous->setToFile($to);

        return $previous;
    }),
```

The tracker updater would then eg update the file content on the CDN instead of locally. However, because we have a `piwik.js` and a `matomo.js` that needs to be updated, the `matomo.js` would have still been updated on the local filesystem because it was using `new File()` hard coded where it should have created basically `new CdnFile()`. 

We need this patch on the cloud fyi.

### DIFF
--- a/plugins/CustomPiwikJs/File.php
+++ b/plugins/CustomPiwikJs/File.php
@@ -15,11 +15,16 @@ class File
     /**
      * @var string
      */
-    private $file;
+    protected $file;
 
     public function __construct($file)
     {
         $this->file = $file;
+    }
+
+    public function setFile($file)
+    {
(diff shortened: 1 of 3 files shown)


tsteur:
Merging this one as tests pass and we have it in use on production and works there.

PR #14669, opened 2019-07-18
Prevent unserialization error in DataTableFactory

jajm:
Sometimes $blobRow[$blobName] is set but its value is false and
DataTable::fromSerializedArray will throw an exception if called with a
false value

This patch prevents this by verifying that the value is not 'empty'
instead

### DIFF
--- a/core/Archive/DataTableFactory.php
+++ b/core/Archive/DataTableFactory.php
@@ -397,7 +397,7 @@ class DataTableFactory
             }
 
             $blobName = $dataName . "_" . $sid;
-            if (isset($blobRow[$blobName])) {
+            if (!empty($blobRow[$blobName])) {
                 $subtable = DataTable::fromSerializedArray($blobRow[$blobName]);
                 $this->setSubtables($subtable, $blobRow, $treeLevel + 1);
 


core/Archive/DataTableFactory.php
```diff
@@ -397,7 +397,7 @@ private function setSubtables($dataTable, $blobRow, $treeLevel = 0)
             }
 
             $blobName = $dataName . "_" . $sid;
-            if (isset($blobRow[$blobName])) {
+            if (!empty($blobRow[$blobName])) {
```
tsteur: cheers @jajm . Maybe we should simply check like `isset($blobRow[$blobName]) && is_array($blobRow[$blobName])`? I haven't looked in detail but wonder if we would still want to go in the `if` when there is eg an empty array? I guess if it was an empty array, it would simply set an empty subtable which probably isn't wanted so the patch looks good with the `!empty()` check actually. Maybe @diosmosis can confirm?

diosmosis: I think it should be ok, blobRow[...] should never be set to false/empty array AFAIK.

tsteur (approved):

diosmosis:
Thanks for the first contribution @jajm!

PR #14671, opened 2019-07-18
Save DataTable metadata in records if added during archiving.

diosmosis:
@tsteur what do you think about this change? This would allow me to set a 'is_imported_from_google' metadata on a table before archiving.


core/DataTable.php
```diff
@@ -204,6 +204,14 @@ class DataTable implements DataTableInterface, \IteratorAggregate, \ArrayAccess
     const LABEL_SUMMARY_ROW = -1;
     const LABEL_TOTALS_ROW = -2;
 
+    /**
+     * The ID of the special metadata row. This row only exists in the serialized row data and stores the datatable metadata.
+     *
+     * This allows us to save datatable metadata in archive data.
+     */
+    const ID_ARCHIVED_METADATA_ROW = -3;
+    const LABEL_ARCHIVED_METADATA_ROW = -3;
```
tsteur: could it be an issue when -3 is actually a value? Like usually so far we don't archive the totals row for example so there it was maybe not an issue? Similar to ranking query wondering should that maybe be like "___metadatarow__archive__"?

core/DataTable.php
```diff
@@ -1291,6 +1299,15 @@ public function getSerialized($maximumRowsInDataTable = null,
             $subtableId = 0;
             throw new Exception("Maximum recursion level of " . self::$maximumDepthLevelAllowed . " reached. Maybe you have set a DataTable\Row with an associated DataTable belonging already to one of its parent tables?");
         }
+
+        // gather metadata before filters are called, so their metadata is not stored in serialized form
+        $metadata = $this->getAllTableMetadata();
+        foreach ($metadata as $key => $value) {
+            if (!is_scalar($value) && !is_string($value)) {
```
tsteur: @diosmosis quickly looked through some archivers and dataArrays and it seems like there's usually no metadata on the dataTable so that should be fine. Haven't checked search engine plugin or so. I wondered if we really want to persist by default automatically all metadata or whether we should maybe require to specify whether some/which metadata should be persisted or not? Not sure if it would be always wanted to persist all information. I guess we could also just leave it like this, and eventually change it later when need comes up to whitelist specific metadata.

tsteur (commented):
Looks good in general. Left a comment though. Not sure what happens for aggregated data tables @diosmosis ? During archiving this would be probably handled in the archiver, when `mergeSubtables` or so is called, it would I think keep the metadata of the first table, which should be fine I suppose. At least for our use case. 👍

core/DataTable.php
diosmosis: 👍 makes sense.

This would also be a problem w/ the summary row as well... actually, rows are associated by ID, not label, so there might be some issues in the future, but perhaps not now. I'll use the better label regardless.

core/DataTable.php
diosmosis: I could add a 'metadata_to_serialize' metadata, though maybe that's just too much meta...

tsteur: I think it's not needed for now. Initially thought it was a good idea, but then realising it's probably over-engineering in the  sense that we seem to not have a use case for it just yet and we can always add it later. I reckon be good to keep it simple for now. 

Couldn't find any table that sets metadata during the archiving in core and our plugins. Don't think there would be any third party plugins out there that do it, and if so, it likely doesn't even cause an issue. I suppose if you set metadata, maybe it's what people expect that it gets persisted 👍

core/DataTable.php
tsteur: @diosmosis be good to make the change 👍  BTW: You maybe also want to move it up next to the other `LABEL*` constants

diosmosis: @tsteur Updated.

PR #14672, opened 2019-07-18
fix environment validator test

diosmosis:
@tsteur can you check these changes? They fix the environment validator test, not sure if this is how you'd want it fixed though.


tsteur:
Looks great 👍

Issue #14418, opened 2019-05-06
Row functions not visible


babielcbo:
We are using Matomo on premise. For some users, the buttons for opening transitions are not displayed on hover in a row.
If we delete the account and create a new one, the buttons are displayed again. However, it happens that they disappear again after a while.
![image001](https://user-images.githubusercontent.com/43536414/57224506-f8a39f00-7009-11e9-80c9-cbfbb9f509e0.jpg)

sgiehl:
@babielcbo The row actions are currently disabled for browser performance reasons when more than 255 rows are displayed in a table. So maybe try reducing the limit of rows shown,

tsteur:
@babielcbo  closing as it's likely due to @sgiehl comment. Feel free to comment if that's not the case.

PR #14675, opened 2019-07-18
Show row actions no matter how many results in the data table are shown

tsteur:
refs https://github.com/matomo-org/matomo/issues/14418


diosmosis:
This solves the performance issue by using a single event instead of one on each row, correct?

tsteur:
@diosmosis before it was iterating over all rows whenever the table was loaded and was doing work for each row and setting up event listener for each row which is now indeed only a single event.

diosmosis:
👍

mattab:
Very awesome to have this bug fixed :tada:

PR #14676, opened 2019-07-19
Remove sponsors from the help page

tsteur:
fyi @mattab I think this link can be removed?


Findus23:
Maybe instead just replace it with "get involved"?

tsteur:
@Findus23 added it. Didn't add it initially cause we have this link already in the top:

image

Not sure if it's needed?

PR #14681, opened 2019-07-21
Support configuration of a reader database

### LINKED ISSUE #7554: Add auxiliary database support for big instances
Goal of this ticket is to discuss and hopefully implement solution allowing single Piwik instance to query two replicant databases.

Outline of issue:
- big Piwik instances often have master-slave db replication,
- currently only use (in our case) of slave servers is to perform db backups, and in emergency cases to switch into master to provide HA

However slave servers could be put to much better use if we can redirect archiving queries onto slave db.
Long running and heavy queries could be triggered against slave, then results processed in PHP and write archives to master server.
This would significantly relieve tracker, take away some other problems with db performance when traffic peak meets long running archiving.

For initial scope, maybe only select queries triggered by LogAggregator could be launched against auxiliary db server, and in future we add option for developers to decide which db to use.
Even putting basic LogAggregator queries into slave would give significant boost to performance.

What do you think?

tsteur:
If configured, most archiving, live, and archive reading queries (when browser archiving is disabled) will be read from a reader instead of the master. This means mainly the tracker will hit the master database and when writing archives.

FYI added a system check that connection works if a reader is configured.

fix https://github.com/matomo-org/matomo/issues/7554

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -45,6 +45,19 @@ ssl_no_verify =
 ; Matomo should work correctly without this setting but we recommend to have a charset set.
 charset = utf8
 
+; If configured, the following queries will be executed on the reader instead of the writer.
+; * archiving queries that hit a log table
+; * live queries that hit a log table
+; * fetching of archives when viewing a report
+; You only want to enable a reader if you can ensure there is minimal replication lag / delay on the reader.
+; Otherwise you might get corrupt data in the reports.
+[database_reader]
+host =
+username =
+password =
+dbname =
+port = 3306
+
 [database_tests]
 host = localhost
 username = "@USERNAME@"
--- a/core/Archive.php
+++ b/core/Archive.php
@@ -453,6 +453,31 @@ class Archive implements ArchiveQuery
         return $dataTable;
     }
 
+    private function canUseDbReader()
+    {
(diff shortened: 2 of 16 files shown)


tsteur:
@mattab @sgiehl @diosmosis could you have a quick look at this? Any thoughts? 

Tested it locally and worked quite well. Also adjusted all our premium features to use `LogAggregator::getDb()` and `VisitorDetailsAbstract::getDb() // in a BC way`  so in the future the core can easily decide/control whether to use slave or master etc. 

Once I get the OK will give it a test on a different Matomo.

mattab:
Didn't look at the code in details but the idea sounds great! Many users will have slave databases running doing nothing and this is a great way to leverage this valuable CPU. 

 Just wondering, what happens if the reader goes down, would it maybe help to full back to master automatically or is it better to fail explicitly?

tsteur:
For now it would just fail. Would otherwise need a lot of extra code etc. It's just to get a quick win. Later you can at some point do a lot more with it.

tsteur:
Maybe we could merge this feature already? Considering it will use regular DB if no reader is defined it shouldn't change anything. Can also patch this later though.

diosmosis:
Code looks good to me, seems ok to merge to me

PR #14686, opened 2019-07-22
Move queued tracking lock implementation to core for other uses.

diosmosis:


core/Concurrency/LockBackend/MySqlLockBackend.php
```diff
@@ -0,0 +1,165 @@
...
+
+    public function install()
+    {
+        if (self::$tableCreated) {
+            return;
+        }
+
+        $tableName = self::getTableName();
+
+        DbHelper::createTable($tableName, "
```
tsteur: As the table is installed on demand... If there's a race condition, should we maybe catch a "table already exists" error and fail silently? Or we just install the table the regular way and provide an update for it as well? IT be fine if table is not needed by most users.

core/Concurrency/LockBackend/MySqlLockBackend.php
```diff
@@ -0,0 +1,165 @@
...
+
+namespace Piwik\Concurrency\LockBackend;
+
+
+use Piwik\Common;
+use Piwik\Concurrency\LockBackend;
+use Piwik\Db;
+use Piwik\DbHelper;
+
+class MySqlLockBackend implements LockBackend
```
tsteur: Might be good to write some simple tests for this backend as well? Can be probably copied from Queued tracking

core/Concurrency/LockBackend/MySqlLockBackend.php
diosmosis: I thought the SQL was 'CREATE IF NOT EXISTS', but I'll double check.

core/Concurrency/LockBackend/MySqlLockBackend.php
diosmosis: Ah yes, I forgot about the backend, I'll add the tests.

core/Concurrency/LockBackend/MySqlLockBackend.php
diosmosis: Looks like it's not `CREATE IF NOT EXISTS`, but the exception is already caught: https://github.com/matomo-org/matomo/blob/3.x-dev/core/Db/Schema/Mysql.php#L452

diosmosis:
@tsteur Updated, though can't be merged just yet. I moved the table creation to Schema.php, since the lock could be used during tracking, and `Db::exec` tries to use the profiler.

PR #14687, opened 2019-07-22
Determine in file class if the file content is the same

tsteur:
I will need this one in 3.11.0 ideally. 

Moved the detection whether the file is up to date to a new method `isFileContentSame()`. It looks unneeded, but is useful when you have eg a ` class CdnFile {}` and need to update the content of that file not only on one path, but also for other trusted hosts. Then you need to save the content basically again if just one of the files is out of date. This happens only when using a custom directory for tracker files.

eg /matomo/foo/$trustedHost/matomo.js
eg /matomo/foo/trusted.host.one/matomo.js
eg /matomo/foo/trusted.host.two/matomo.js
eg /matomo/foo/trusted.host.three/matomo.js

Then we don't just need to read/update/delete one file, but multiple times for each host.

For the same reason we need to return an array of which files were actually updated in the `save()` method. We might not just update the file in one path, but actually in multiple paths. And then we need to trigger an event for each updated file.

Ideally we would have this logic better in core somehow, but the way we use it is quite special and usually you wouldn't have a file per trusted host.

### DIFF
--- a/plugins/CustomPiwikJs/File.php
+++ b/plugins/CustomPiwikJs/File.php
@@ -43,6 +43,8 @@ class File
 
     public function isFileContentSame($content)
     {
+        // we determine if file content is the same in here in case a different "file" implementation needs to check
+        // whether multiple files are up to date
         return $this->getContent() === $content;
     }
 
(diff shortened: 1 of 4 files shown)


tsteur:
@mattab needing to merge this one to have it in the next release. It's covered by tests and only a very basic change.

PR #14689, opened 2019-07-22
Quick links widget

### LINKED ISSUE #13979: Add shortcuts to admin homepage
Add a new widget with some buttons to quickly perform some common actions such as 
* Add new website
* Add new user

Those few things can be otherwise hard to find and maybe make users think they cannot add a 2nd website. Especially since they might not know "Measurables" is websites etc.

image

Others items would maybe not be needed for now as most of them can be found easier. Those are two actions we want users to take though in many cases (add user and add website).

katebutler:
Fixes #13979


plugins/CoreHome/Widgets/QuickLinks.php
```diff
@@ -0,0 +1,26 @@
...
+    {
+        $config->setCategoryId('About Matomo');
+        $config->setName('CoreHome_QuickLinks');
+        $config->setOrder(16);
+        $config->setIsEnabled(Piwik::hasUserSuperUserAccess());
+    }
+
+    public function render()
+    {
+        return $this->renderTemplate('quickLinks');
```
tsteur: Here in `render()` we might need to add another `Piwik::checkUserHasSuperUserAccess()` just to be safe

tsteur:
@katebutler similar to the system summary might be good to show icons in front of the links and not have a newline in between.

In general it's starting to get a bit busy in there: 
image

I wonder if it makes sense to show the "Quick Links" above "System Summary"? They should be more important I suppose. Was thinking it be nice to merge eg Quick Links  and System Summary but we disable different widgets on the cloud so that wouldn't be straight forward.

I was thinking maybe we could also add a link to "Add a goal" etc but these are not quite administrative and can be found in the reporting UI fairly easily anyway. So good not to bloat the widget. Most important is certainly "Add a new website" so users discover they can track more of them in one Matomo.

Maybe for now we leave it with those 2 links and over time at more links that make sense 👍

tsteur (changes requested):
See comment

PR #14690, opened 2019-07-23
Tweak title on "no data site" to mention start tracking

tsteur:
Was like this:
image

but I feel like it's not doing a good job at making clear the user actually needs to do something. Changing therefore the title see image below. Maybe could also say `No data has been recorded yet. Start tracking now!` (might be bit too "aggressive")? Not sure what is best. It should make clear to the user that action needs to be done and should motivate to do this.

image


diosmosis:
Some potential alternatives, that may or may not work:

* "no data has been recorded yet, make sure your tracking code is setup"
* "no data has been recorded yet, setup tracking now"
* "no data has been recorded yet, get setup below"
* "no data has been recorded yet, get started now/below"

diosmosis (approved):
LGTM if the wording has been finalized

PR #14696, opened 2019-07-25
Remove no longer needed require statement

tsteur:
Was probably still there from the old days where includes were done manually eg in tracker mode maybe.


sgiehl (approved):
LGTM

PR #14700, opened 2019-07-25
Make Visitor Log live query more performant

tsteur:
Sometimes the live query can be really slow (like many minutes). It could even end up using all database resources when there is a lot of data in the log tables. Especially when using segments on top. Often the system ends up doing queries like these: 

```sql
SELECT sub.* FROM (
            SELECT
                log_inner.*
            FROM     
        (
            SELECT
                log_visit.*
            FROM
                log_visit AS log_visit LEFT JOIN log_link_visit_action AS log_link_visit_action ON log_link_visit_action.idvisit = log_visit.idvisit
            WHERE
                ( log_visit.idsite in ('1') 
                AND log_visit.visit_last_action_time >= '2019-05-25 23:00:00' )
                AND
                ( ( log_link_visit_action.idaction_url IS NOT NULL AND log_link_visit_action.idaction_url <> '' AND log_link_visit_action.idaction_url <> '0' ) )
            ORDER BY
                idsite DESC, visit_last_action_time DESC LIMIT 0, 4010
        ) AS log_inner
            ORDER BY
                idsite DESC, visit_last_action_time DESC
            ) AS sub
            GROUP BY sub.idvisit
            ORDER BY sub.visit_last_action_time DESC
        LIMIT 401
```

Where it fetches data over several months. This obviously can take a very long time, and it might create big temporary tables, full table scans, etc. As we're only wanting the most recent results, it be smarter to first look only at the last day and if this query returns enough results, then we can directly return the result making it a LOT faster.

In this PR I'm executing up to 5 queries to get the result instead of doing it in one query:
* Look one day back
* Look another 1 week back (without fetching first day again)
* Look 1 month back (without looking at the 1week + 1day we already fetched again)
* Look 1 year back (without looking at the data we already requested)
* Remaining time 

If only 1 week is being requested, it would only execute 2 queries. If 2 years are requested, it would execute up to 5 queries but stop as soon as the requested data has been found. So we might be actually only looking at one day compared to 2 years! 

The logic can be only done if there is a data limit. Otherwise we need to return all results anyway and this can still be a problem fyi @mattab 

It cannot be executed when there is a limit plus an offset @mattab . So it won't solve too many problems possibly. Unless we executed each query twice see this logic:

```
Imagine we have `Limit 5 Offset 5`
Query 1 finds: 2 rows  => should use Limit 5, Offset 5. The result of the method would return an empty array. Because of the offset, we can't know that 2 rows have been skipped unless we executed every query twice when no result has been returned? First we would need to execute `Limit 5, Offset 5`, then we would need to execute `Limit $offset, Offset 0` to find out how many rows were skipped
Query 2 finds: 0 rows  => should use Limit 5, Offset 3 
Query 3 finds: 7 rows  => should use Limit 5, Offset 3 and basically should skip the first 3 rows and then return the next 4
Query 4 finds: 0 rows: => should use Limit 1, Offset 0: that one is easy... as soon as at least one row has been found we use offset => 0 and limit can be calculated easily too (originalLimit - foundRows).
```
Another workaround for offset that slightly improves it could be that when `Offset <= 3 * LIMIT` then we apply the offset in PHP and not in MySQL. Means we fetch a lot more rows but then discard them in PHP. Would make this logic work for a few more pages.

And another workaround be, if there's an offset, we simply check if the first day gives results and if not fetch all other days.

### DIFF
--- a/plugins/Live/Model.php
+++ b/plugins/Live/Model.php
@@ -89,6 +89,7 @@ class Model
             }
 
             if ($virtualDateStart->isEarlier($virtualDateEnd)) {
+                // need to use ",endDate" in case startDate is not set in which case we do not want to have any limit
                 $queries[] = array($dateStart, $virtualDateEnd->subSeconds(1));
             }
         } else {
--- a/plugins/Live/Model.php
+++ b/plugins/Live/Model.php
@@ -61,8 +61,13 @@ class Model
             $virtualDateEnd = Date::now()->addDay(1); // matomo always adds one day for some reason
         }
 
+        $virtualDateStart = $dateStart;
+        if (empty($virtualDateStart)) {
+            $virtualDateStart = Date::factory('1990-01-01'); // matomo always adds one day for some reason
+        }
+
         $queries = [];
-        $hasStartEndDateMoreThanOneDayInBetween = $dateStart && $dateStart->addDay(1)->isEarlier($virtualDateEnd);
+        $hasStartEndDateMoreThanOneDayInBetween = $virtualDateStart && $virtualDateStart->addDay(1)->isEarlier($virtualDateEnd);
         if ($limit
             && $hasStartEndDateMoreThanOneDayInBetween
         ) {
@@ -70,20 +75,20 @@ class Model
                 $virtualDateEnd = $virtualDateEnd->subDay(1);
                 $queries[] = array($virtualDateEnd, $dateEnd); // need to use ",endDate" in case endDate is not set
             }
-            if ($dateStart->addDay(7)->isEarlier($virtualDateEnd)) {
+            if ($virtualDateStart->addDay(7)->isEarlier($virtualDateEnd)) {
                 $queries[] = array($virtualDateEnd->subDay(7), $virtualDateEnd->subSeconds(1));
                 $virtualDateEnd = $virtualDateEnd->subDay(7);
             }
-            if ($dateStart->addDay(30)->isEarlier($virtualDateEnd)) {
+            if ($virtualDateStart->addDay(30)->isEarlier($virtualDateEnd)) {
                 $queries[] = array($virtualDateEnd->subDay(30), $virtualDateEnd->subSeconds(1));
                 $virtualDateEnd = $virtualDateEnd->subDay(30);
             }
-            if ($dateStart->addPeriod(1, 'year')->isEarlier($virtualDateEnd)) {
+            if ($virtualDateStart->addPeriod(1, 'year')->isEarlier($virtualDateEnd)) {
                 $queries[] = array($virtualDateEnd->subYear(1), $virtualDateEnd->subSeconds(1));
                 $virtualDateEnd = $virtualDateEnd->subYear(1);
             }
 
-            if ($dateStart->isEarlier($virtualDateEnd)) {
+            if ($virtualDateStart->isEarlier($virtualDateEnd)) {
                 $queries[] = array($dateStart, $virtualDateEnd->subSeconds(1));
             }
         } else {
--- a/plugins/Live/Model.php
+++ b/plugins/Live/Model.php
@@ -56,57 +56,7 @@ class Model
 
         list($dateStart, $dateEnd) = $this->getStartAndEndDate($idSite, $period, $date);
 
-        $virtualDateEnd = $dateEnd;
-        if (empty($dateEnd)) {
-            $virtualDateEnd = Date::now()->addDay(1); // matomo always adds one day for some reason
-        }
-
-        $virtualDateStart = $dateStart;
-        if (empty($virtualDateStart)) {
-            $virtualDateStart = Date::factory(Date::FIRST_WEBSITE_TIMESTAMP); // matomo always adds one day for some reason
-        }
-
-        $queries = [];
-        $hasStartEndDateMoreThanOneDayInBetween = $virtualDateStart && $virtualDateStart->addDay(1)->isEarlier($virtualDateEnd);
-        if ($limit
-            && $hasStartEndDateMoreThanOneDayInBetween
-        ) {
-            if ($hasStartEndDateMoreThanOneDayInBetween) {
-                $virtualDateEnd = $virtualDateEnd->subDay(1);
-                $queries[] = array($virtualDateEnd, $dateEnd); // need to use ",endDate" in case endDate is not set
-            }
-
-                if ($virtualDateStart->addDay(7)->isEarlier($virtualDateEnd)) {
-                    $queries[] = array($virtualDateEnd->subDay(7), $virtualDateEnd->subSeconds(1));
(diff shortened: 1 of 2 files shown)


tsteur:
> Another workaround for offset that slightly improves it could be that when Offset <= 3 * LIMIT then we apply the offset in PHP and not in MySQL. Means we fetch a lot more rows but then discard them in PHP. Would make this logic work for a few more pages.

Maybe not the best idea actually.

> And another workaround be, if there's an offset, we simply check if the first day gives results and if not fetch all other days.

Kind of implemented something like this but haven't pushed yet. It be a compromise between things maybe and possibly still be somewhat efficient for high traffic websites.

plugins/Live/Model.php
```diff
@@ -43,20 +43,123 @@ public function queryLogVisits($idSite, $period, $date, $segment, $offset, $limi
...
+        list($dateStart, $dateEnd) = $this->getStartAndEndDate($idSite, $period, $date);
+
+        $virtualDateEnd = $dateEnd;
+        if (empty($dateEnd)) {
+            $virtualDateEnd = Date::now()->addDay(1); // matomo always adds one day for some reason
+        }
+
+        $virtualDateStart = $dateStart;
+        if (empty($virtualDateStart)) {
+            $virtualDateStart = Date::factory(Date::FIRST_WEBSITE_TIMESTAMP); // matomo always adds one day for some reason
```
diosmosis: Is this code comment correct?

plugins/Live/Model.php
```diff
@@ -43,20 +43,123 @@ public function queryLogVisits($idSite, $period, $date, $segment, $offset, $limi
...
+
+        $queries = [];
+        $hasStartEndDateMoreThanOneDayInBetween = $virtualDateStart && $virtualDateStart->addDay(1)->isEarlier($virtualDateEnd);
+        if ($limit
+            && $hasStartEndDateMoreThanOneDayInBetween
+        ) {
+            if ($hasStartEndDateMoreThanOneDayInBetween) {
+                $virtualDateEnd = $virtualDateEnd->subDay(1);
+                $queries[] = array($virtualDateEnd, $dateEnd); // need to use ",endDate" in case endDate is not set
+            }
```
diosmosis: is this `if` needed? Above you check for `$limit && $hasStartEndDateMoreThanOneDayInBetween`, so `$hasStartEndDateMoreThanOneDayInBetween` should be true.

plugins/Live/Model.php
```diff
@@ -43,20 +43,123 @@ public function queryLogVisits($idSite, $period, $date, $segment, $offset, $limi
...
+                // need to use ",endDate" in case startDate is not set in which case we do not want to have any limit
+                $queries[] = array($dateStart, $virtualDateEnd->subSeconds(1));
+            }
+        } else {
+            $queries[] = array($dateStart, $dateEnd);
+        }
 
-        $visits = Db::getReader()->fetchAll($sql, $bind);
+        // uncomment to see the dateTimes that have been generated.
+        //$debugDateTime = array_map(function ($query) { return ($query[0] ? $query[0]->getDatetime() : '') . ' ' . ($query[1] ? $query[1]->getDatetime() : ''); }, $queries);
```
diosmosis: Instead of a comment here, maybe we could extract the above code into a public method and write some unit tests?

plugins/Live/Model.php
```diff
@@ -43,20 +43,123 @@ public function queryLogVisits($idSite, $period, $date, $segment, $offset, $limi
...
+                }
+            }
+
+            if ($virtualDateStart->isEarlier($virtualDateEnd)) {
+                // need to use ",endDate" in case startDate is not set in which case we do not want to have any limit
+                $queries[] = array($dateStart, $virtualDateEnd->subSeconds(1));
+            }
+        } else {
+            $queries[] = array($dateStart, $dateEnd);
+        }
```
diosmosis: Here's an idea that might make it a bit faster (assuming I'm reading this correctly). What the code is doing now is this:

- check dateStart => dateStart + 1 day
- still no visits? check dateStart => dateStart + 7 days
- still no visits? check dateStart => dateStart + 30 days
- ... etc.

In each query we're checking the same rows as the previous one, we could change it to something like:

- check dateStart => dateStart + 1 day
- still no visits? check dateStart + 1 day => dateStart + 7 days
- still no visits? check dateStart + 7 days => dateStart + 30 days
- ... etc.

diosmosis:
I wonder if this is similar to `Db::segmentedFetchAll()` and if an approach like that could be used to lower the required resources. For example, check one day, check one week, then query one month at a time, instead of doing a whole year at once.

plugins/Live/Model.php
tsteur: Nope 👍 will remove it

plugins/Live/Model.php
tsteur: Good point, will remove it

plugins/Live/Model.php
tsteur: @diosmosis it should already do what you suggested. Unless I'm not understanding it right. But basically the code does not fetch the same day or week twice for example.

diosmosis: Ah ok, `$virtualEndDate` changes, I didn't notice that.

plugins/Live/Model.php
tsteur: Definitely 👍  just did

tsteur:
It's quite similar @diosmosis just that `Db::segmentedFetchAll()` maybe doesn't know anything re limit/offset/bind and has some first/last/step which I don't need.

diosmosis:
@tsteur I was actually more just wondering if it would save db resources to go month by month, instead of querying for a whole year (or the whole range). I haven't done any sort of testing, so I have no idea if it would help or not.

tsteur:
@diosmosis we were thinking if we go month by month it might be in the end too many queries and taking too long. The though was more like improving performance for high traffic Matomo's. And they likely find eg 500 visitors within the first day / week or month. And if there weren't 500 visitors in the first month, likely they also won't be found in the next months so it might be better to query the remaining date range at once.

We could be a bit "smart" maybe and have some logic if we find eg 25% of the results in one month, then we could assume that we will have all rows found if we query 3-4 more months instead of say the remaining 2 years. Is this maybe something you mean?

diosmosis:
I just figured the high load was due to the number of rows being counted overall, and if so going over smaller chunks might be less of an issue. However, if that isn't what's causing the high load, then it wouldn't be a good idea.

Here's another crazy idea that's too difficult to implement: call VisitsSummary.get for multiple days over the last couple months or so, and use the number of visits to in rows to figure out where an offset starts. (I just wanted to write this idea since it seems strange yet workable, I don't think you should work on it :) )

Anyway, code looks good to me. Don't know if @mattab needs to look at it.

tsteur:
@diosmosis that might not help when a segment is used and when the segment is not preprocessed for example.

diosmosis (approved):

Issue #14085, opened 2019-02-10
Return http status 403 on failed login


fdellwing:
Would it be possible to return a 403 http status code on the login post request if invalid credentials are used? This way the webserver and other software can capitalize on this information more easily.

mattab:
Thanks for the suggestion :+1: 
 Wondering if there could maybe be side effects on some server configuration, but probably not?

mattab:
Pull request welcome @fdellwing :+1:

fdellwing:
I took a look into it and was not able to fiddle with the return codes. If anyone can hint me a method or example, will look into it further.

Findus23:
Solved in https://github.com/matomo-org/matomo/pull/14708

PR #14708, opened 2019-07-29
Send 403 header on failed Login

pebosi:
Proposal for #14085

### DIFF
--- a/plugins/Login/Controller.php
+++ b/plugins/Login/Controller.php
@@ -143,6 +143,10 @@ class Controller extends \Piwik\Plugin\ControllerAdmin
                 $messageNoAccess = $this->getMessageExceptionNoAccess();
             }
         }
+        
+        if ($messageNoAccess) {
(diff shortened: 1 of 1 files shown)


Findus23:
Hi,

Thanks for the PR.
I don't think you can assume everyone is using `HTTP 1.0`.
Thankfully since PHP 5.4 there is a function for this: [`http_response_code()`](https://php.net/http-response-code).

Findus23:
Or maybe use the exiting helper function:
https://github.com/matomo-org/matomo/blob/65b5523adf92c686f1448e858fa428b57d68a674/core/Common.php#L1196

@tsteur Is there any reason why there is a custom function?

tsteur:
Looking at the content of the method it definitely seems like there's a reason fir it by the looks. But maybe `http_response_code()` was never used and that's why it was needed. Not sure.

pebosi:
Changed to use http_response_code

diosmosis:
Works for me, thanks for the contribution @pebosi!

PR #14712, opened 2019-07-30
 Adds command to update iso region names

### LINKED ISSUE #14462: Cities do not correspond to regions
Hi, I'm using matomo 3.9.1 with GeoIP 2 (PHP)
The problem is that Moroccan cities are not matching regions (also the map is showing incorrect regions for cities) as you can see in the picture below :
map
I also tested some IP address here:
https://www.maxmind.com/en/geoip-demo
and seems like everything is OK, so I do not know where the problem comes from.
**PS: I can help correct Moroccan cities to match their regions**

sgiehl:
and updates region names.

fixes #14462


plugins/GeoIp2/Commands/UpdateRegionCodes.php
```diff
@@ -0,0 +1,135 @@
...
+{
+    public $source = 'https://salsa.debian.org/iso-codes-team/iso-codes/raw/master/data/iso_3166-2.json';
+
+    protected function configure()
+    {
+        $this->setName('usercountry:update-region-codes');
+        $this->setDescription("Updates the ISO region names");
+    }
+
+    public function isEnabled()
```
tsteur: is this on purpose?

tsteur: I see this is used for us 👍

tsteur:
Haven't tested it but code looks 👍

plugins/GeoIp2/Commands/UpdateRegionCodes.php
sgiehl: exactly

PR #14717, opened 2019-07-31
When loading a template url from a custom plugin directory, prefix the relative directory

tsteur:
In https://github.com/matomo-org/matomo/pull/14051 we started supporting multiple plugin paths.

There we already rewrite the path to the JS and CSS assets and load them from a different path. But it looks for some reason that a `directive templateUrl` is not loaded correctly when the plugin is located in a different path. It's still trying to load the template file eg from `$matomoUrl/matomo/plugins/PluginName/angularjs/widgetname/widgetname.html` when instead  it should be maybe loaded from `$matomoUrl/matomo/../extensions/PluginName/angularjs/widgetname/widgetname.html`

### DIFF
--- a/plugins/CoreHome/angularjs/http404check.js
+++ b/plugins/CoreHome/angularjs/http404check.js
@@ -24,7 +24,7 @@
                     var urlParts = config.url.split('/');
                     if (urlParts && urlParts.length > 2 && urlParts[1]) {
                         var pluginName = urlParts[1];
-                        if (pluginName && pluginName in piwik.relativePluginWebDirs) {
+                        if (pluginName && pluginName in piwik.relativePluginWebDirs && piwik.relativePluginWebDirs[pluginName]) {
                             urlParts[0] = piwik.relativePluginWebDirs[pluginName];
                             config.url = urlParts.join('/');
                         }
--- a/plugins/CoreHome/angularjs/http404check.js
+++ b/plugins/CoreHome/angularjs/http404check.js
@@ -16,7 +16,7 @@
 
         return {
             'request': function(config) {
-                if ('object' === typeof matomoRelativeWebDirs
+                if ('object' === typeof piwik.relativePluginWebDirs
                     && config && config.url && config.url.indexOf('plugins/') === 0
                     && config.url.indexOf('.html') > 0
                     && config.url.indexOf('/angularjs/') > 0) {
@@ -24,8 +24,8 @@
                     var urlParts = config.url.split('/');
                     if (urlParts && urlParts.length > 2 && urlParts[1]) {
                         var pluginName = urlParts[1];
-                        if (pluginName && pluginName in matomoRelativeWebDirs) {
-                            config.url = matomoRelativeWebDirs[pluginName] + config.url;
+                        if (pluginName && pluginName in piwik.relativePluginWebDirs) {
+                            config.url = piwik.relativePluginWebDirs[pluginName] + config.url;
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


nabiltntn:
I am not sure if it's related , I am using custom plugins folder for some of premium plugins like ABTesting, or SessionsRecording,

When Selecting them from the left menu, nothing is displayed in the main content ( no errors in browser's console / network )

Capture d’écran 2019-08-23 à 16 43 48



For ABTesting plugin, the following files are 404, i guess because it's not not taking in account custom folders relative path:
- $matomohost/plugins/AbTesting/libs/abtestingicons/fonts/abtestingicons.ttf?5qxh2m
- $matomohost/plugins/AbTesting/libs/abtestingicons/fonts/abtestingicons.woff?5qxh2m
 
Thanks.

tsteur:
Yes it's related to that. Feel free to give the patch a try and let me know if it works 👍

nabiltntn:
Unfortunately, i am on aDocker environnement right now, so will wait for the official `3.12.0` image to test if this PR will be included in.

Thanks

plugins/Morpheus/templates/_jsGlobalVariables.twig
```diff
@@ -16,14 +16,13 @@
...
+
     {% if userLogin %}piwik.userLogin = "{{ userLogin|e('js')}}";{% endif %}
 
     {% if idSite is defined %}piwik.idSite = "{{ idSite }}";{% endif %}
 
-    {% if siteName is defined %}
-    // NOTE: siteName is currently considered deprecated, use piwik.currentSiteName instead, which will not contain HTML entities
-    piwik.siteName = "{{ siteName|e('js') }}";
-    piwik.currentSiteName = {{ siteNameDecoded|json_encode|raw }};{% endif %}
+    {% if siteName is defined %}piwik.siteName = "{{ siteName|e('js') }}";{% endif %}
```
diosmosis: I think the code deleted on the left shouldn't be deleted, right?

tsteur: Indeed that should not be removed. Was probably due to some later merge. Will fix that

tsteur: fixed @diosmosis

plugins/CoreHome/angularjs/http404check.js
```diff
@@ -15,6 +15,23 @@
...
+            'request': function(config) {
+                if ('object' === typeof piwik.relativePluginWebDirs
+                    && config && config.url && config.url.indexOf('plugins/') === 0
+                    && config.url.indexOf('.html') > 0
+                    && config.url.indexOf('/angularjs/') > 0) {
+
+                    var urlParts = config.url.split('/');
+                    if (urlParts && urlParts.length > 2 && urlParts[1]) {
+                        var pluginName = urlParts[1];
+                        if (pluginName && pluginName in piwik.relativePluginWebDirs) {
```
diosmosis: The `pluginName` check in the if is redundant here since `urlParts[1]` is checked above.

diosmosis: Would it be safer to check `if (piwik.relativePluginWebDirs[pluginName])` instead of `pluginName in piwik.relativePluginWebDirs` in case the value is falsey?

tsteur: Added another check. I think we could remove `plugin in ...` but had sometimes problems with it in past not sure why.

PR #14718, opened 2019-08-01
Do not cache error page

tsteur:
refs DEV-1725

Ensuring that when the page is reloaded, it will be fetched again.

### DIFF
--- a/core/testMinimumPhpVersion.php
+++ b/core/testMinimumPhpVersion.php
@@ -116,6 +116,7 @@ if (!function_exists('Piwik_GetErrorMessagePage')) {
 
         if (!headers_sent()) {
             header('Content-Type: text/html; charset=utf-8');
+            header('Cache-Control: private, no-cache, no-store');
 
             $isInternalServerError = preg_match('/(sql|database|mysql)/i', $message);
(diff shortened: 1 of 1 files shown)


core/testMinimumPhpVersion.php
```diff
@@ -116,6 +116,7 @@ function Piwik_GetErrorMessagePage($message, $optionalTrace = false, $optionalLi
 
         if (!headers_sent()) {
             header('Content-Type: text/html; charset=utf-8');
+            header('Cache-Control: private, no-cache, no-store');
```
diosmosis: Is there a reason for using multiple cacheability directives? Seems like we'd only need to use `no-store` (from reading https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Cacheability).

tsteur: @diosmosis that's what we also use in production see eg https://stackoverflow.com/a/2068407/1447375 . In this case we don't set the Expires or Pragma no-cache but at least the Cache-Control.

diosmosis: I see it's for different browsers. Fun.

PR #14722, opened 2019-08-02
When loading templates detect if an ad blocker is active

tsteur:
fix DEV-1680
The status seems to be -1 when an ad blocker is blocking the loading of an html template (eg HeatmapSessionRecording)

image

Should avoid error like this:

image

### DIFF
--- a/plugins/CoreHome/angularjs/http404check.js
+++ b/plugins/CoreHome/angularjs/http404check.js
@@ -7,7 +7,7 @@
 
         function isClientError(rejection)
         {
-            if (rejection.status === 500 || rejection.status === -1) {
+            if (rejection.status === 500 || rejection.status <= -1) {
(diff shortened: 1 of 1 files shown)


plugins/CoreHome/angularjs/http404check.js
```diff
@@ -7,7 +7,7 @@
 
         function isClientError(rejection)
         {
-            if (rejection.status === 500) {
+            if (rejection.status === 500 || rejection.status === -1) {
```
diosmosis: Can we do <= 0 to potentially catch more problems?

tsteur: Done. Although not sure when it would be 0 or lower than that but we can always change it again later if needed for some reason 👍

PR #14727, opened 2019-08-02
Two small changes for GA importer

diosmosis:
* Add Visualization.beforeRender event. I need to add a footer message after the datatable is loaded, so I can't use VIewDataTable.configure (since it's posted in the constructor).
* Don't overwrite existing table metadata entries when loading datatables from blobs so metadata that is archived is returned through the API.

### DIFF
--- a/core/Plugin/Visualization.php
+++ b/core/Plugin/Visualization.php
@@ -190,7 +190,9 @@ class Visualization extends ViewDataTable
             $this->applyFilters();
             $this->addVisualizationInfoFromMetricMetadata();
             $this->afterAllFiltersAreApplied();
+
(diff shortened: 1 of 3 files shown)


core/Plugin/Visualization.php
```diff
@@ -723,6 +725,14 @@ public function beforeRender()
         // eg $this->config->showFooterColumns = true;
     }
 
+    private function fireBeforeRenderHook()
+    {
+        /**
+         * @ignore
```
tsteur: Could we make this an official event and add a small description?

tsteur:
LGTM... be just good to add some comment for the event

PR #14728, opened 2019-08-02
Add note about google analytics importer to site without data card.

diosmosis:
The note is added to core since the importer will be released on the marketplace, so we can't add it through an event.

This cannot be merged until the importer is released on the marketplace.


plugins/SitesManager/templates/siteWithoutData.twig
```diff
@@ -77,6 +77,9 @@
 
                 <h3>{{ 'CoreAdminHome_HttpTrackingApi'|translate }}</h3>
                 <p>{{ 'CoreAdminHome_HttpTrackingApiDescription'|translate('<a href="https://developer.matomo.org/api-reference/tracking-api" rel="noreferrer noopener" target="_blank">','</a>')|raw }}</p>
+
+                <h3>{{ 'CoreAdminHome_ImportFromGoogleAnalytics'|translate }}</h3>
+                <p>{{ 'CoreAdminHome_ImportFromGoogleAnalyticsDescription'|translate('<a href="https://plugins.matomo.org/GoogleAnalyticsImporter" rel="noopener noreferrer" target="_blank">','</a>')|raw }}</p>
```
tsteur: On cloud we would rather want a message re getting in touch with support or so? or maybe even hide it... 
And maybe if plugin is activated the message should be also different?

diosmosis: @tsteur 

For changing the message if it is activated, I think we can just check if `isPluginActivated(...)` in the Controller.

For hiding the element, I can either:

- add an event like `CoreAdminHome.disableGoogleAnalyticsImporterMessage` or something
- or just use a div w/ an ID so it can be hidden through css

How does all of that sound to you?

tsteur: Either is fine. Can use the CSS I suppose 👍 

@mattab might also need to think re whitelabel? Should it be removed then as well? I reckon eg if "RemoveLinks" setting is disabled then this would need to be removed? Maybe also in general? (not sure)... in this case it would need to be an event though @diosmosis so we can remove it on demand.

Or we move the addition of this into the `ProfessionalServices` plugin (and then showing it using the template event). It's not a commercial notice the GA importer as it is free but could be still moved in there? Not sure

diosmosis: The event seems like it would be more flexible, I'll just use that.

tsteur: Sounds good. Can you maybe let me know once this is done so I can adjust white label and cloud?

diosmosis:
@tsteur / @mattab modified the PR to use an event

tsteur:
LGTM if tests pass @diosmosis

PR #14738, opened 2019-08-07
Show marketplace in reporting menu

tsteur:
Currently the marketplace is very hidden (like the 15th link in the Admin). Ideally we would want users to discover more features to make the most out of their Matomo.

refs MAR-573


tsteur:
We put also in a link "Premium Features" but could also remove it. It's just that we had the view/widget for it already there.

PR #14752, opened 2019-08-11
Add JS event invoked after a widget is loaded for client side manipulation of widgets.

diosmosis:
New JS event is used in the google analytics importer to add footer text to reports like transitions or usersflow which aren't viewdatatables


tsteur:
Looks good.

PR #14753, opened 2019-08-11
Add some debugging info to log notifications so they can be more easily reproduced.

diosmosis:


plugins/Monolog/Handler/WebNotificationHandler.php
```diff
@@ -58,4 +59,15 @@ protected function write(array $record)
             // Silently ignore the error.
         }
     }
+
+    private function getLiteDebuggingInfo()
+    {
+        $module = Common::getRequestVar('module', false);
+        $action = Common::getRequestVar('action', false);
+        $method = Common::getRequestVar('method', false);
+        $trigger = Common::getRequestVar('trigger', false);
```
tsteur: what is `trigger`? Do you mean archive trigger?

plugins/Monolog/Handler/WebNotificationHandler.php
```diff
@@ -58,4 +59,15 @@ protected function write(array $record)
             // Silently ignore the error.
         }
     }
+
+    private function getLiteDebuggingInfo()
+    {
+        $module = Common::getRequestVar('module', false);
```
tsteur: Might need to use an empty string here by default so something is visible also for the others? like `'""'` or even better not show the info if no value is present?

plugins/Monolog/Handler/WebNotificationHandler.php
diosmosis: Yes, thought it would be useful to see if it's from within an archiving request.

plugins/Monolog/Handler/WebNotificationHandler.php
diosmosis: I can remove it if no value is present.

plugins/Monolog/Handler/WebNotificationHandler.php
diosmosis: Since that could be within a climulti request.

plugins/Monolog/Handler/WebNotificationHandler.php
tsteur: Be good 👍

plugins/Monolog/Handler/WebNotificationHandler.php
tsteur: 👍  makes sense. Wasn't quite sure initially where we use it but then kind of remembered.

diosmosis:
Updated.

tsteur:
Haven't tested it but looks good @diosmosis can you just check it escapes things before merging?

diosmosis:
Verified the notification escapes the content before displaying; adding `<script>alert(1);</script>` did not create an alert.

Issue #14119, opened 2019-02-20
DB: Split log_* Tracker tables into hot/cold for faster tracking API requests & reports archiving 


mattab:
In order to improve overall performance of Matomo (both at Tracker and Reports processing) an idea we have is to split log_* Tracker tables into two sets of tables: hot data  VS cold data. 

eg hot data would be data for the last 48 hours or so. Cold data would be any data older than 48 hours. As a result the "hot data" table could be 100 times smaller than the cold data (depending on data retention settings see https://matomo.org/docs/managing-your-databases-size/ )

Having tables split in this way means that:
* Tracking requests is faster (inserting into smaller tables takes less time)
* Archiving reports is faster (querying the same datasets but over smaller log_ tables will be faster)

There will be a few changes needed eg.  need a process to move data from hot to cold, need to update SQL query generators to query the right tables as applicable (and possibly getting data from the two tables eg. Visitor log requests over multiple days), etc.

tsteur:
It'll be also beneficial for when there are DB updates on log tables.

tsteur:
Before working on this or as part of this issue may be good to see if horizontal MySQL partitioning can help like ` PARTITION BY HASH( MONTH(visit_last_action_time) )`.
I think in most queries where performance matters we already do mention the `visit_last_action_time` and/or `server_time` but that would need to be checked.

mattab:
It would be great to maybe spend 2-4 days on this project and see what is possible to achieve. Making this change in the database would greatly increase performance of most SQL queries executed by Matomo. (also results in less CPUs usage overall/ more efficient.)

tsteur:
BTW: MySQL Parititions would have been great. Like we could have done `PARTITION BY HASH(YEAR(time) Month(time))    PARTITIONS 24;` which would have put data for each month in a separate table eg when using a data retention of 24 months. But, we cannot use any of the time columns for partitions since that column would need to be present in all unique indexes including the primary key.

tsteur:
clientid google 75290852256-39s6vao0u2f9877vebv05apb032pkapc.apps.googleusercontent.com
client secret tXiOs4lnuTS6V3m-otcBhqhY


dev token Mc4ar6CsVm2L9Y0oIgZnyw


I can think here of many possible solutions. Many of the solutions have quite a few downsides or are not easy to develop. Only mentioning some therefore. And partitions is not really an option as we've already seen. Two of them I would like to mention in more detail.

[...]

PR #14761, opened 2019-08-13
Faster segment archiving

### LINKED ISSUE #11900: When segmenting by page URL, Events/Downloads/Outlinks/Site search reports always show no data
Hi

I'm currently starting to use the events feature, and right now I'm able to track and store some test data including event categories, event actions and event names. 

The problem is on the report, when I go to Dashboard -> Actions -> Events, all 3 tabs show that "there is no data for this report". I checked the database and I can see the data correctly stored on log_link_visit_action and log_action tables with my events, but reports just won't display anything, even when I can see visitor logs correctly, the issue is only with events reports.

Of course I made sure to pick the correct site and date for the events activity I'm tracking, but reports are just not showing up.

Am I missing some setting to enable this reports somewhere else? or is there any other method or data I should implement/track to be able to see the reports?

Thanks in advance

**Edited**:

-> More experimentation has revealed that this issue only occurs when a page URL or action URL based segment is used. In a certain way this makes sense because the events themselves don't have any URL.

tsteur:
Seeing roughly 2-6 times faster segment archiving using this patch.
Generally: The less visitors a segment matches, the much faster the archiving query will become (say even 100 times faster)... if all visitors match the segment, there's a chance it takes slightly longer. The more complex the segment be, the faster this is as well.

fix https://github.com/matomo-org/matomo/issues/11900
refs https://github.com/matomo-org/matomo/issues/14119#issuecomment-519675939

Current problems: 
* In the log aggregator we are creating a temporary table. In the log aggregator, we can however not know if all the plugins are using the same DB connection. When a DB reader is configured, then this temporary table needs to be created on the reader. However, if a plugin archiver does `Db::get()->query()` instead of `$this->logAgreggator->getDb()->query`() then the archiving will fail. I reckon this is edge case though since the DB reader is a new feature and not yet wildly used. All our plugins including premium features are compatible
* Another DB reader thing: I don't know if all MySQL replications are happy with creating temporary tables... the reader needs to be able to create a temporary table there
* `create temporary table`: The syntax is like `create temporary table logtemp select idvisit from log_visit ...` where the `select` is generated depending on the configured segment. I'm not sure if this works with older MySQL versions. I reckon so but not sure.
* I will need someone to check the test fixture changes
* Will then need to disable the segment config and create new tests with the segment enabled

In general it works like this when a segment is defined for archivers (not live query as there is no benefit to it): 

```sql
create temporary table logtemp$segmentHash select idvisit from log_visit where idsite = 1 and visit_last_action_time >= ? and visit_last_action_time <= ? AND ... // segment query appended... sometimes also more joins etc

-- now each archiver query uses that temp table

select $archiveSelect from logtemp$segmentHash left join log_link_visit_action where id_action_event_category is not null  -- just an example... We do not exuecute the ` idsite = 1 and visit_last_action_time >= ? and visit_last_action_time <= ?` again in this query

-- drop table (will also automatically deleted on connection end)
```

In other words:
```php
// traditionally be 
$segment->getSelectQuery($select, $from, $where, ...);

// now we do
$segmentTable = 'logtemp445454545';
'create temporary table ' . $segmentTable . $segment->getSelectQuery('idvisit', 'log_visit', 'idsite=1 and time > ? and time < ?', ...);
array_unshift($segmentTable, $from);
$noSegment->getSelectQuery($select, $from);
// because we execute the segment query in advance, we can always execute basically the raw archiver query (inner joined with the segment table)
```

From what I see it should pretty much get rid of subqueries etc. and we basically always run the original archiving query and no longer join random tables.


config/global.ini.php
```diff
@@ -209,6 +209,12 @@
 enabled_periods_UI = "day,week,month,year,range"
 enabled_periods_API = "day,week,month,year,range"
 
+; whether to enable segment archiving cache
+; Note: if you use any plugins, this need to be compliant with Matomo and
+; * depending on the segment you create you may need a newer MySQL version (eg 5.7 or newer)
+; * use a reader database for archiving in case you have configured a database reader
+enable_segments_cache = 1
```
tsteur: we would disable this before the release

tsteur: and we'll need to write tests with this setting enabled and then check the segmentation works

tsteur:
FYI: For now not creating temporary tables as if table be larger than the memory then we would need to execute the select another time and try to create the table on disk... if table result is > allowed memory the create table simply fails

tsteur:
Just did a test btw where a query took 177s, and when using engine=memory the same archiver query was actually not much faster (171s).

tsteur:
All the tests now succeed. There's one failing test in the bandwidth plugin related to this change, all the other test failures are unrelated. I will fix the bandwidth test after merging as otherwise things get complicated re the submodule etc.

core/DataAccess/LogAggregator.php
```diff
@@ -174,18 +187,107 @@ public function setQueryOriginHint($nameOfOrigiin)
...
+
     public function generateQuery($select, $from, $where, $groupBy, $orderBy, $limit = 0, $offset = 0)
     {
+        $segment = $this->segment;
         $bind = $this->getGeneralQueryBindParams();
-        $query = $this->segment->getSelectQuery($select, $from, $where, $bind, $orderBy, $groupBy, $limit, $offset);
+
+        if (!$this->segment->isEmpty() && $this->isSegmentCacheEnabled()) {
+            // todo when we create table for query we cannot know if the archiver will use that same Db writer or Db reader
+            // if plugin doesn't support reader just yet... there will be a problem!
```
diosmosis: Does this TODO need to be turned into an issue for followup?

core/DataAccess/LogAggregator.php
```diff
@@ -174,18 +187,107 @@ public function setQueryOriginHint($nameOfOrigiin)
...
+                            }
+                            $bind = array_slice($bind, 3);
+                            break;
+                        }
+                    }
+
+                }
+            } catch (\Exception $e) {
+                throw $e;
+            }
```
diosmosis: Given it just rethrows, is this try-catch necessary?

diosmosis: Same for below.

core/DataAccess/LogQueryBuilder/JoinGenerator.php
```diff
@@ -190,8 +191,20 @@ public function generate()
...
+                    $joinName = 'INNER JOIN';
+                    // edge case when there is eg `logtmpsegment57cd546b7203d68a41027547c4abe1a2` as first table...
+                    // then we join log_conversion for example... if we didn't use INNER JOIN we would as a result
+                    // get rows for visits even when they didn't have a conversion. Instead we only want to find rows
+                    // that have an entry in both tables when doing eg
+                    // logtmpsegment57cd546b7203d68a41027547c4abe1a2.idvisit = log_conversion.idvisit
+                }
                 // the join sql the default way
-                $this->joinString .= " LEFT JOIN $tableSql ON " . $join;
+                $this->joinString .= " $joinName $tableSql ON " . $join;
```
diosmosis: Would it make this code simpler to make sure the log table segment is always last? I guess it's possible the segment table being first is a valid use case...

tsteur:
BTW: @mattab just fixed another issue while testing with a DB reader where the ranking query was before not using the Reader DB

core/DataAccess/LogAggregator.php
tsteur: Not needed anymore. We discussed to document this.

core/DataAccess/LogAggregator.php
tsteur: Will remove it again 👍 had it in there for easier debugging

core/DataAccess/LogQueryBuilder/JoinGenerator.php
tsteur: The temporary log table should always be first otherwise things get slow. And for regular archives without segment there is no tmp log table as it would make things slower. Not sure if that helps?

diosmosis: If it's always first, then this isn't an edge case? I was responding to the comment above.

tsteur: Yep logtemptableXYZ is always first when there is a segment applied.

diosmosis: So then the comment about the edge case is incorrect?

tsteur: Do you refer to it being an edge case? I guess it's not an edge case considering that segments will be archived fairly often. Was just thinking of it as an edge case as it is only when there are segments but not for regular archive

diosmosis: seems like the new normal behavior to me. when I read the comment I thought the segment table being first was a strange thing that could happen, not something that should happen if there's a segment.

tsteur: changed it 👍

tsteur:
btw @mattab forgot to mention one thing as a side note...

eg page urls we archive like `log_link_visit_action.server_time < ... > ...`... and goals we archive on `log_conversion.server_time < ...>... ` but this is not 100% correct as  we don't respect the visit time and whether the visit for this was also in the same day or not... 

The segments now include `log_visit.last_visit_action_time` which is correct so there might be slightly different number when having segment `Continent=EUR` and segment `Continuent!=EUR` where the total of those 2 segments might not be 100% the number without the segment.

PR #14779, opened 2019-08-16
Admin controllers do not call setGeneralVariablesView, so we have to …

diosmosis:
…set the property another way.


plugins/Morpheus/templates/_jsGlobalVariables.twig
```diff
@@ -23,7 +23,9 @@
     {% if siteName is defined %}
     // NOTE: siteName is currently considered deprecated, use piwik.currentSiteName instead, which will not contain HTML entities
     piwik.siteName = "{{ siteName|e('js') }}";
+    {% if currentSiteName is defined %} // just to be safe
```
tsteur: is this supposed to be `if siteNameDecoded  is defined`?

PR #14786, opened 2019-08-18
Provide possibility to force not using a subquery

tsteur:
refs https://github.com/matomo-org/matomo/pull/14761

Avoids generating queries like this:

```sql
CREATE TEMPORARY TABLE IF NOT EXISTS logtmpsegmente5bfbc1f135a495c4f5aad6d8047199c (idvisit  BIGINT(10) UNSIGNED NOT NULL) 
            SELECT
                distinct log_inner.idvisit as idvisit
            FROM
                
        (
            
            SELECT
                log_visit.idvisit
            FROM
                log_visit AS log_visit LEFT JOIN log_link_visit_action AS log_link_visit_action ON log_link_visit_action.idvisit = log_visit.idvisit
            WHERE
                ( log_visit.visit_last_action_time >= '2019-04-30 14:00:00'
                AND log_visit.visit_last_action_time <= '2019-05-31 13:59:59'
                AND log_visit.idsite IN ('1') )
                AND
                ( ( log_link_visit_action.idaction_url IN (SELECT idaction FROM log_action WHERE ( name LIKE CONCAT('%', '/admin', '%')  AND type = 1 )) ) )
            GROUP BY
                log_visit.idvisit
            ORDER BY
                NULL
        ) AS log_inner
            ORDER BY
                log_inner.idvisit ASC
```

where we don't need the subquery anymore for create temp table statements anymore as we always do a `distinct idvisit` anyway. 

Also fixes an issue where the bind was wrongly set for roll up reporting

### DIFF
--- a/core/DataAccess/LogAggregator.php
+++ b/core/DataAccess/LogAggregator.php
@@ -249,7 +249,7 @@ class LogAggregator
 
             $logQueryBuilder = StaticContainer::get('Piwik\DataAccess\LogQueryBuilder');
             $forceGroupByBackup = $logQueryBuilder->getForcedInnerGroupBySubselect();
-            $logQueryBuilder->forceInnerGroupBySubselect('');
+            $logQueryBuilder->forceInnerGroupBySubselect(LogQueryBuilder::FORCE_INNER_GROUP_BY_NO_SUBSELECT);
             $segmentSql = $this->segment->getSelectQuery('distinct log_visit.idvisit as idvisit', 'log_visit', $segmentWhere, $segmentBind, 'log_visit.idvisit ASC');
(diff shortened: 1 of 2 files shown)


core/DataAccess/LogAggregator.php
```diff
@@ -274,7 +274,7 @@ public function generateQuery($select, $from, $where, $groupBy, $orderBy, $limit
                         if (stripos($where, 'and ') === 0) {
                             $where = substr($where, strlen('and '));
                         }
-                        $bind = array_slice($bind, 3);
+                        $bind = array();
```
tsteur: We can't just remove 3 entries eg when for roll up reporting there are multiple sites specified. Then `bind` might be actually several idSites plus the start date and end date

PR #14788, opened 2019-08-18
Copy visits into temp table non locking

### LINKED ISSUE #14818: Create temporary table for segments only once instead of twice when archiving
Currently, when we archive, we initialize the `PluginsArchiver` twice. Once for core and once for the plugins here: https://github.com/matomo-org/matomo/blob/3.12.0-b2/core/ArchiveProcessor/Loader.php#L101-L117

In the PluginsArchiver's constructor we then first cleanup any possible temp table that already exists ( https://github.com/matomo-org/matomo/blob/3.12.0-b2/core/ArchiveProcessor/PluginsArchiver.php#L70-L71) and then later create the temp table as soon as a `LogAggregator::generateQuery` is called.

This means what we do is:

* Start archiving core metrics
* Cleanup temp table for segment if for some reason was not removed (eg a fatal error or memory issue etc and old MySQL version is used)
* Create temp table for segment
* archive core metrics
* Finalise archive
* Cleanup temp table for segment ===> This is the step we want to prevent but only if we actually go into the plugins archiver again to archive plugin metrics

* Start archiving plugins metrics
* Cleanup temp table for segment if for some reason was not removed (eg a fatal error or memory issue etc and old MySQL version is used) ===> This is the step we want to prevent
* archive plugin metrics
* Finalise archive
* Cleanup temp table for segment

This will make segmenting again a bit faster as we avoid running the expensive segment query and avoid lots of inserts...

tsteur:
@mattab from what I've been reading this should avoid the issue re a long running `create temporary table select ...` possibly locking some tracking requests.

* Converted it in general into two statements: `create temporary table` and `insert into ...` to avoid a metadata lock
* Using `read-uncommitted` to prevent locks on the source and target DB. Target DB doesn't matter anyway since we know nobody can read from it until it's been created as it's a temporary table
* Note: Doing the `read-uncommitted` only when `innodb` is configured in the config as we can't really know if `innodb` is available and it wouldn't make sense with `myisam`.
* In general `Myisam` might be pretty bad as it locks the entire table... I wonder if we should disable this new `temp table` behaviour if someone is still using `myisam` on `log_visit` (and assume it is the same for other log tables)? Could also add a system check if not already done if myisam is still used?

This issue also fixes: https://github.com/matomo-org/matomo/issues/14818 making archiving of segments again faster


tsteur:
BTW the only other solution I see to make this non-locking is selecting all results in PHP, then doing a batchInsert. Of course this can have memory and performance implications. Could try doing it for MyISAM users though

tsteur:
Added a workaround for MyIsam users to actually fetch the idVisits into PHP and insert them all at once to not have it locking for them. This way we don't need to disable it for them as this behaviour should be used eventually for all users

core/ArchiveProcessor/PluginsArchiver.php
```diff
@@ -68,6 +68,7 @@ public function __construct(Parameters $params, ArchiveWriter $archiveWriter = n
         $this->archiveWriter->initNewArchive();
 
         $this->logAggregator = new LogAggregator($params);
+        $this->logAggregator->allowUsageSegmentCache();
```
tsteur: We only do it during archiving... if other usages want to make use of it, they need to enable it as well... prevents performing the create temporary table logic for some live requests like Heatmaps where it otherwise would possibly apply the segment on a yearly range etc.

tsteur: Also prevents that someone maybe doesn't call the cleanup command etc

tsteur: Also prevents that we possibly create heaps of temporary tables depending on the ui/api request...

tsteur:
FYI: Was reading this morning a new blog the read-uncommitted status might be even fine for our regular archiving queries but this would be a different issue https://federico-razzoli.com/read-only-transactions-in-mysql

mattab:
@sgiehl can you please review this PR and check it all makes sense?

mattab:
ping @sgiehl 

@tsteur we can otherwise merge and test in the next beta this week?

tsteur:
sounds good @mattab most of the code we're actually running on production for a while

PR #14789, opened 2019-08-19
Make sure to include instanceId in batchInsert

tsteur:
For easier debugging etc.


mattab:
Good find. Could have resulted in nasty bugs!

PR #14808, opened 2019-08-25
Fix MatomoSenior string in Tour

Findus23:
just noticed it when checking the translations


tsteur:
@Findus23 I think "Matomo Talent" was actually meant to be the translation, we just didn't change the translation key

Findus23:
But then we got a second „Matomo Talent“ a few lines below. 
Lukas

tsteur:
The translation key was actually not used anymore

Issue #14785, opened 2019-08-18
don't allow updating 0 plugins


Findus23:
Just a minor usability improvement:
When no updates are selected, but `Update Selected` is clicked, the user is redirected to an error page.

![image](https://user-images.githubusercontent.com/6266037/63228227-673eee00-c1f0-11e9-9c7a-5f2107c41f92.png)

![image](https://user-images.githubusercontent.com/6266037/63228237-7625a080-c1f0-11e9-9354-ff2218728d4b.png)

> The parameter 'pluginName' isn't set in the Request, and a default value wasn't provided.

sgiehl:
this was fixed with #14811

PR #14811, opened 2019-08-25
Don't allow update 0 plugin

MrYamous:
Fix for issue #14785 

Button for updating plugins is now disabled by default, enabled if at least one plugin is selected

### DIFF
--- a/plugins/CorePluginsAdmin/templates/macros.twig
+++ b/plugins/CorePluginsAdmin/templates/macros.twig
@@ -10,7 +10,7 @@
         <tr>
             <th>
                 <span class="checkbox-container">
-                    <input type="checkbox" id="select-plugin-all" onclick="boxes_selected()"/>
+                    <input type="checkbox" id="select-plugin-all"/>
                     <label for="select-plugin-all"></label>
                 </span>
             </th>
@@ -74,6 +74,11 @@
 
     <script>
         $(function () {
+            $('span.checkbox-container input').on('click', function() {
+                var isAtLeastOneChecked = $('span.checkbox-container > input[checked]').length >= 1;
+                $('#update-selected-plugins').prop('disabled', !isAtLeastOneChecked);
+            });
+
             $('#update-selected-plugins').on('click', function () {
                 var pluginsToUpdate = [];
                 $('tbody#plugins td.select-cell input').each(function () {
@@ -101,21 +106,6 @@
                 });
             });
         });
-        
-        function boxes_selected(){
-            var checked_boxes = 0;
-            var boxes = document.querySelectorAll('span.checkbox-container > input');
-            
-            for (var i = 0; i < boxes.length; i++){
-                if (boxes[i].checked)
-                    checked_boxes += 1
-            }
-            if (checked_boxes >= 1) {
-                document.getElementById('update-selected-plugins').classList.remove('disabled');
-            } else if (checked_boxes === 0) {
-                document.getElementById('update-selected-plugins').classList.add('disabled');
-            }
-        }
     </script>
 
 {% endmacro %}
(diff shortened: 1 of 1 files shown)


plugins/CorePluginsAdmin/templates/macros.twig
```diff
@@ -101,6 +101,21 @@
...
+            for (var i = 0; i < boxes.length; i++){
+                if (boxes[i].checked)
+                    checked_boxes += 1
+            }
+            if (checked_boxes >= 1) {
+                document.getElementById('update-selected-plugins').classList.remove('disabled');
+            } else if (checked_boxes === 0) {
+                document.getElementById('update-selected-plugins').classList.add('disabled');
+            }
+        }
```
diosmosis: This function should not be added to the global scope and should not be referenced via `onclick=`. Instead, jQuery should be used to bind an anonymous function to the event (as is done above). Can you make this change?

Also, this code can be simplified to just check if at least one input is checked, eg:

```
var isAtLeastOneChecked = $('span.checkbox-container > input[checked]').length >= 1;
$('#update-selected-plugins').prop('disabled', !isAtLeastOneChecked);
```

plugins/CorePluginsAdmin/templates/macros.twig
```diff
@@ -101,6 +101,21 @@
                 });
             });
         });
+        
+        function boxes_selected(){
```
diosmosis: In general we don't use snake case in matomo, we use camel case (so this function name would be `boxesSelected`).

diosmosis (changes requested):
Hi @MrYamous thanks for your first contribution, we appreciate your time and effort! I left a couple comments after reviewing the code, once the changes are made we can merge the pull request.

plugins/CorePluginsAdmin/templates/macros.twig
MrYamous: Hi @diosmosis thanks for your review and your advice
I made a new commit considering your tip :)

plugins/CorePluginsAdmin/templates/macros.twig
```diff
@@ -26,7 +26,7 @@
             <tr {% if plugin.isActivated|default(false) %}class="active-plugin"{% else %}class="inactive-plugin"{% endif %}>
                 <td class="select-cell">
                     <span class="checkbox-container">
-                        <input type="checkbox" id="select-plugin-{{ plugin.name|e('html_attr') }}" {% if plugin.isDownloadable is defined and not plugin.isDownloadable %}disabled="disabled"{% endif %} />
+                        <input type="checkbox" id="select-plugin-{{ plugin.name|e('html_attr') }}" onclick="boxes_selected()" {% if plugin.isDownloadable is defined and not plugin.isDownloadable %}disabled="disabled"{% endif %} />
```
tsteur: @MrYamous it doesn't work for me because of an error when I click on it that this method is not defined... but even when removing this onclick it still seems to not activate the "update selected" button.

MrYamous: @tsteur According to diosmosis's feedback, i removed this onclick and boxes_selectes(). Did you test the last commit ?

tsteur: There is still `onclick="boxes_selected()"` though but no such method. Maybe you forgot to commit something?

MrYamous: Exact, i missed one onclick, fixed in a new commit

plugins/CorePluginsAdmin/templates/macros.twig
```diff
@@ -74,6 +74,11 @@
 
     <script>
         $(function () {
+            $('span.checkbox-container').on('click', function() {
+                var isAtLeastOneChecked = $('span.checkbox-container > input[checked]').length >= 1;
+                $('#update-selected-plugins').prop('disabled', !isAtLeastOneChecked);
```
tsteur: I reckon you might also need to remove the CSS class `disabled`? It's still not working for me here.

MrYamous: Hello, sorry i did some mistakes with selector.

Issue #14821, opened 2019-08-27
Make Matomo compatible with PHP 7.4


Findus23:
similar to #13418

### Status:

All Warnings in Matomo itself and the broken Twig rendering are fixed in Matomo 3.12. All remaining warnings depend on third-party libraries (see below).

Soon PHP 7.4 will be released (RC 1 is planned or September 5th), so it would be great if Matomo would work mostly fine when it is released.

I set up a Matomo instance using the latest beta4 at https://matomo-beta.lw1.at
If you want to get access, just message me.

Current status is that Matomo is nearly completly unusable as it renders invalid HTML with missing whitespaces:
```html
<ul class="right hide-on-med-and-down"><li role="menuitem" class="active"><atitle="Zeige Webanalytik Berichte für Testsite."                  id="topmenu-corehome"
                  href="index.php?module=CoreHome&amp;action=index&amp;idSite=1&amp;period=day&amp;date=yesterday"               target="_self" tabindex="3">Dashboard</a></li><li role="menuitem" class=""><atitle="Vergleichen Sie Webanalytikstatistiken für alle Ihre Websites."                  id="topmenu-multisites"
                  href="index.php?module=MultiSites&amp;action=index&amp;idSite=1&amp;period=day&amp;date=yesterday"               target="_self" tabindex="3">Alle Websites</a></li><li role="menuitem" class=""><atitle=""                  id="topmenu-tagmanager"
```

This seem to be caused by a required change in Twig (https://github.com/twigphp/Twig/pull/3004/commits/1fb0f9701d8443083495cd2645e8a0c45d54c34d), so updating to twig 1.41.0 or later should solve this.

Findus23:
One more bug that breaks archiving: (appears as a response in archiving)

```
[...]

PR #14825, opened 2019-08-28
fix implode argument order (php 7.4 compatibility)

Findus23:
related to https://github.com/matomo-org/matomo/issues/14821
With PHP 7.4 a lot of incorrect usages of functions have been deprecated.
One of them is that `implode()` will now throw a warning if the glue isn't the first parameter.

I'll update this PR as I come across more warnings.

### DIFF
--- a/core/Tracker/Model.php
+++ b/core/Tracker/Model.php
@@ -313,7 +313,7 @@ class Model
     {
         list($updateParts, $sqlBind) = $this->fieldsToQuery($valuesToUpdate);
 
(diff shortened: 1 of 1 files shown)


tsteur:
I'd be happy to merge. Not sure why PR is in draft maybe you want to make more changes?

Findus23:
@tsteur When I made the PR I expected there to be many of those cases, but it turns out that this seems to be the only one.
So I guess this can be merged.

PR #14829, opened 2019-08-28
Change login page message when user has been redirected after auth failure

### LINKED ISSUE #14706: More informative session expiry message
Issue: When the session is timing out, users are getting re-directed to login page "You can't access this resource as it requires view access for the website id =1", which gives an impression that the system is not working properly, as for as user is concern they are logged in. No information regarding timeout is provided to the user.

session_timeout

Suggestion: It would be clear for the user why they are not able to access the dashboard/report, if the information about session timeout is displayed (ex: User Session timed out due to inactivity for over 30min"). If possible, if we could alert the user before 10 or 5min of session expiry, so that the users can go back to Matomo and keep the session active if they wanted to.

katebutler:
Fixes #14706


tsteur:
@mattab could you have a look at this one?

mattab:
Does not work for me, here is what i did:

* Set `login_session_not_remembered_idle_timeout = 30`
* Login without remember me
* Wait 40 seconds
* Click refresh on a widget, saw:
Screenshot from 2019-09-10 11-15-34

* Reload page, saw:
Screenshot from 2019-09-10 11-17-32

Expected to see the correct message instead.

tsteur:
@mattab I think it was in the issue referred to the main page. Should be possible to change it there too. Anyway, what we were wondering was re the message that we're showing that this will be fine. also see the comments in the issue.

tsteur:
@mattab can you have a look at the message etc?

mattab:
Within a widget getting: Screenshot from 2019-09-16 22-09-10

Expected to see a more clear message if possible?

Then when clicking again on "Dashboard", getting the old message as reported in https://github.com/matomo-org/matomo/issues/14706

session_timeout

The point of fixing #14706 was to display a more clear message, maybe it's not working as expected or is the message not improved?

tsteur:
@katebutler can we change the message to `Please log in to access this functionality`? And then we make sure it works in all cases.

core/Access.php
```diff
@@ -698,7 +698,18 @@ public function getCapabilitiesForSite($idSite)
      */
     private static function throwNoAccessException($message)
     {
-        Piwik::checkUserIsNotAnonymous();
+        try {
+            // Will throw Exception if user is anonymous, i.e. not currently logged in
+            Piwik::checkUserIsNotAnonymous();
+        } catch (NoAccessException $ex) {
+            // Try to detect whether user was previously logged in so that we can display a different message
+            if (isset($_SERVER['HTTP_REFERER']) && strpos($_SERVER['HTTP_REFERER'], SettingsPiwik::getPiwikUrl()) === 0) {
```
tsteur: fyi `$referrer = Url::getReferrer()`

fyi `Url::isValidHost(Url::getHostFromUrl(referrer))`

core/Access.php
```diff
@@ -689,6 +699,28 @@ public function getCapabilitiesForSite($idSite)
...
+     * @throws NoAccessException
+     */
+    private function throwNoAccessException($message)
+    {
+        if (Piwik::isUserIsAnonymous()) {
+            $message = Piwik::translate('General_YouMustBeLoggedIn');
+        }
+        // Try to detect whether user was previously logged in so that we can display a different message
+        $referrer = Url::getReferrer();
+        if (Url::isValidHost(Url::getHostFromUrl($referrer)) &&
```
tsteur: @katebutler could we check also that `referrer` is not empty? Not sure if needed but may be good.

tsteur:
@katebutler maybe one last screenshot to update? https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/36139/UIIntegrationTest_not_logged_in.png

core/Widget/WidgetConfig.php
```diff
@@ -258,6 +259,10 @@ public function disable()
     public function checkIsEnabled()
     {
         if (!$this->isEnabled()) {
+            // Some widgets are disabled when the user is not superuser. If the user is not logged in, we should
+            // prompt them to do this first rather than showing them the "widget not enabled" error
+            Access::getInstance()->checkUserIsNotAnonymous();
```
sgiehl: Wondering if that could make problems if a site is set as accessible for anonymous user, but e.g. the dashboard contains a widget that is disabled (for any kind of reason)

tsteur: Was thinking about that as well @sgiehl but it should be actually fine since it goes only in here when the widget is disabled. In this case the user would need to log in to see a more accurate message. We basically don't want to show that message then and also in general a user would actually not really get to see a disabled widget in the first place maybe (unless opened maybe directly not sure)

PR #14830, opened 2019-08-28
Do tracker request validation/auth before start of processing

### LINKED ISSUE #14801: Detect if tracking request needs authentication but is not authenticated as early as possible before doing any DB connection ideally
In https://github.com/matomo-org/matomo/blob/3.12.0-b1/core/Tracker/Visit.php#L123 we already check early whether the site exists before doing any "work". 

The same way we should check if a tracking API parameter is used that requires authentication very early in the request (after the site request). We're wanting to this for tracking API request parameters where we know the request would fail otherwise.

We're wanting to do this after `manipulateRequest()` was called. So basically here: https://github.com/matomo-org/matomo/blob/3.12.0-b1/core/Tracker/Visit.php#L131

Like `cdt`, `region`, `city`, `country`, `long`, `lat` (and other tracking parameters that fail in `Piwik\Plugins\UserCountry\Columns\Base`).

If any of them are set, and request is not authenticated, we fail right away. The tracking failure should be still logged which is done in `Request::authenticateTrackingApi()` and should just work automatically.

To validate `cdt` it may be enough to call `$trackerRequest->getCurrentTimestamp()` which will then trigger the exception.

For the location tracking parameters we would throw an exception like in `Piwik\Plugins\UserCountry\Columns\Base::getUrlOverrideValueIfAllowed()`.

We probably don't need any tests for this as behaviour should just stay the same. need to make sure though  (eg manually) that when using one of those parameters, and we are not authenticated with token_auth that a tracking failure is triggered

katebutler:
Fixes #14801

### DIFF
--- a/core/Tracker/Visit.php
+++ b/core/Tracker/Visit.php
@@ -14,6 +14,7 @@ use Piwik\Common;
 use Piwik\Config;
 use Piwik\Container\StaticContainer;
 use Piwik\Date;
+use Piwik\Exception\InvalidRequestParameterException;
 use Piwik\Exception\UnexpectedWebsiteFoundException;
 use Piwik\Network\IPUtils;
 use Piwik\Plugin\Dimension\VisitDimension;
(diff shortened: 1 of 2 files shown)


tsteur:
well done @katebutler 👍 works

PR #14831, opened 2019-08-28
Don't process tracking requests that are older than data purge cutoff

### LINKED ISSUE #14802: When log deletion is enabled, do not allow import of any older data
Say you configured logs to be deleted after 3 months in the privacy settings, then we don't want to allow the tracking of any data that is older than 3 months as we would delete them again anyway.

We could also add new tracking failure type for this maybe but to be seen.

Kind of similar to https://github.com/matomo-org/matomo/issues/14801 we want to fail early when this is the case. Right after https://github.com/matomo-org/matomo/blob/3.12.0-b1/core/Tracker/Visit.php#L131 we want to get the timestamp, and if a request time is older than a potentially configured log deletion time, then we need to discard the request.

This logic is only performed if log deletion is enabled (`delete_logs_enable = 1`)

And we discard the tracking request using `InvalidRequestParameterException()` only if the `$request->currentTimestamp()` is older than (`delete_logs_schedule_lowest_interval` + `delete_logs_older_than`)`. AFAIK both privacy config values are configured in days.

To make this performant we need to make sure the data is read from the cache. We can set the config values for these settings here: https://github.com/matomo-org/matomo/blob/3.12.0-b1/plugins/PrivacyManager/PrivacyManager.php#L205 like (just pseudo code)

```php
$caeche['privacymanager']['delete_logs_enable'] = PurgeDataSettings[...]
$caeche['privacymanager']['delete_logs_schedule_lowest_interval'] = PurgeDataSettings[...]
$caeche['privacymanager']['delete_logs_older_than'] = PurgeDataSettings[...]
```

And then read them in `Visit.php` like `$cache = Tracker\Cache::getCacheGeneral()` and then do like (just pseudo code)
```php
if (!empty($cache['privacymanager']['delete_logs_enable']) && !empty($caeche['privacymanager']['delete_logs_older_than'])){   
$timeDropRequest = time() - (delete_logs_schedule_lowest_interval + delete_logs_older_than) * 60*60*24);
if ($request->getCurrentTimestamp() < timeDropRequest) {}   
}
```

The log deletion setting can be configured in the privacy manager UI or in the config currently but that's not really relevant for this.

katebutler:
Fixes #14802


core/Tracker/Visit.php
```diff
@@ -130,6 +130,17 @@ public function handle()
 
         $this->visitProperties = new VisitProperties();
 
+        // Break early if request timestamp is older than the cutoff for deleting older log entries
+        $cache = Tracker\Cache::getCacheGeneral();
```
tsteur: 👍 works nicely @katebutler 

One thought: let's maybe move the logic to here: https://github.com/matomo-org/matomo/blob/3.x-dev/core/Tracker/Request.php#L503-L505 

Could we also throw an `InvalidRequestParameterException ` and a message that explains that timestamp is too old and data would be deleted again because of log retention ? A `Common::printDebug` be good as well. 

At first was thinking we should fail silently as otherwise anyone could find out whether a user as raw data retention configured and to how many days the raw data retention is set). However, this is not possible because you need to authenticate the request anyway when tracking a request with a timestamp of older than one day. So only users with at least write access would be possible to find out about this and there it is fine. 

Be good to throw an exception 👍

core/Tracker/Request.php
```diff
@@ -502,6 +502,18 @@ protected function getCustomTimestamp()
             }
         }
 
+        $cache = Tracker\Cache::getCacheGeneral();
+        if (!empty($cache['delete_logs_enable']) && !empty($cache['delete_logs_older_than'])) {
+            $scheduleInterval = $cache['delete_logs_schedule_lowest_interval'];
+            $maxLogAge = $cache['delete_logs_older_than'];
+            $logEntryCutoff = time() - (($maxLogAge + $scheduleInterval) * 60*60*24);
+            if ($cdt < $logEntryCutoff) {
+                $message = "Request skipped as custom timestamp is older than $maxLogAge days";
```
tsteur: @katebutler could we mention log deletion in the exception message somehow? Otherwise people might be contacting us asking why the don't accept custom timestamps older than X days.

Something like `Custom timestamp is older than the configured "delete old raw data" value of X days`

PR #14832, opened 2019-08-28
Various platform tweaks

tsteur:
* Let a user customise from where to load the config file by specifying a function in `$matomoDir/bootstrap.php` like `$GLOBALS['CONFIG_INI_PATH_RESOLVER'] = function () {return 'pathOutsideMatomo'};`
* New config to disallow editing users (useful when syncing users)
* New config to disallow editing sites (useful when syncing sites)
* Configurable `misc/user` folder for GeoIP... this way you can have the geoIP DB outside a Matomo directory
* A new event `'Config.beforeSave'` see documentation to prevent saving some sensitive data in the config. It is also eg useful for https://github.com/matomo-org/plugin-EnvironmentVariables where you would want to prevent saving any potentially sensitive value to the database that is set through an environment variable. Say you set the database password through an environment variable, then you would want to prevent that the password is written to the config file next time the config is saved

These changes were needed for some new feature we're working on. Besides the config files and the new event we wouldn't document the other features I suppose. Could potentially add some tests for disable user/site admin but not sure if needed. Think we're not doing it so far for other areas? And it's quite straight forward but could add some if needed.


plugins/UsersManager/Controller.php
```diff
@@ -205,11 +208,14 @@ public function userSettings()
             $view->defaultReportSiteName = Site::getNameFor($defaultReport);
         }
 
-        $view->defaultReportOptions = array(
-            array('key' => 'MultiSites', 'value' => Piwik::translate('General_AllWebsitesDashboard')),
-            array('key' => $reportOptionsValue, 'value' => Piwik::translate('General_DashboardForASpecificWebsite')),
-        );
+        $defaultReportOptions = array();
+        if (Plugin\Manager::getInstance()->isPluginActivated('MultiSites')) {
```
tsteur: this is a fix where we were showing multisites link even if not enabled

config/global.ini.php
```diff
@@ -663,6 +663,12 @@
 ; - links to Uninstall themes will be disabled (but user can still enable/disable themes)
 enable_plugins_admin = 1
 
+; By setting this option to 0 the users management will be disabled
+enable_users_admin = 1
+
+; By setting this option to 1 the users management will be disabled
```
mattab: `users` should be `websites`

core/Config.php
```diff
@@ -138,6 +138,10 @@ public static function getDefaultLocalConfigPath()
      */
     public static function getLocalConfigPath()
     {
+        if (!empty($GLOBALS['CONFIG_PATH_RESOLVER']) && is_callable($GLOBALS['CONFIG_PATH_RESOLVER'])) {
+            return call_user_func($GLOBALS['CONFIG_PATH_RESOLVER']);
```
mattab: is it secure to run a function defined in GLOBALS? I guess so, but wasn't sure...

tsteur: Yes it is secure. We use it also in various places.

config/global.ini.php
```diff
@@ -663,6 +663,12 @@
 ; - links to Uninstall themes will be disabled (but user can still enable/disable themes)
 enable_plugins_admin = 1
 
+; By setting this option to 0 the users management will be disabled
+enable_users_admin = 1
+
+; By setting this option to 1 the websites management will be disabled
```
diosmosis: Typo here, I think. `1` should be `0`, correct?

tsteur: Yes 👍 fixed

PR #14834, opened 2019-08-29
Add option to archiver to not create archives for segments for today

tsteur:
Unless the segment was created recently (in the last 24 hours).

At first I had a different implementation only in `CronArchiver` and it worked but problem was that when it archives current week, month, year it would still archive today again. This is why a mixed implementation was needed.

It can be triggered using `core:archive --skip-segments-today`

@mattab could you have a look at this?

### DIFF
--- a/core/CronArchive.php
+++ b/core/CronArchive.php
@@ -1772,12 +1772,16 @@ class CronArchive
         return $url;
     }
 
-    protected function wasSegmentCreatedRecently($definition, $allSegments)
+    protected function wasSegmentChangedRecently($definition, $allSegments)
     {
         foreach ($allSegments as $segment) {
             if ($segment['definition'] === $definition) {
                 $twentyFourHoursAgo = Date::now()->subHour(24);
-                return Date::factory($segment['ts_created'])->isLater($twentyFourHoursAgo);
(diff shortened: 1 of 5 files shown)


tsteur:
@mattab if you could have a look here that be great

core/CronArchive.php
```diff
@@ -1760,6 +1772,18 @@ private function makeRequestUrl($url)
         return $url;
     }
 
+    protected function wasSegmentCreatedRecently($definition, $allSegments)
+    {
+        foreach ($allSegments as $segment) {
+            if ($segment['definition'] === $definition) {
+                $twentyFourHoursAgo = Date::now()->subHour(24);
+                return Date::factory($segment['ts_created'])->isLater($twentyFourHoursAgo);
```
mattab: we use `ts_created` but could we instead use `ts_last_edit`? For the use case where user is updating the segment after a few days, and would also expect the segment to be processed soon after the modification (ideally).

mattab:
PR works :+1: 

Does this feature need to be documented in a FAQ or guide? (don't think so)

core/CronArchive.php
tsteur: updated the PR to look at that also

PR #14843, opened 2019-09-02
Minor performance tweak for purge archive for deleted sites

tsteur:
Not expecting a big change in performance here. But generally it's better to have two simple queries that can be fully read from index instead of joining the tables etc. The join should be straight forward though as well because it should be also created based on the index, however the two individual queries should be still faster.

Also another tweak is that for deleted sites we  don't need to look at `ts_archived`. This actually does make things quite a bit faster.

### DIFF
--- a/core/DataAccess/Model.php
+++ b/core/DataAccess/Model.php
@@ -319,16 +319,29 @@ class Model
      * @param string $oldestToKeep Datetime string
      * @return array of IDs
      */
-    public function getArchiveIdsForDeletedSites($archiveTableName, $oldestToKeep)
+    public function getArchiveIdsForDeletedSites($archiveTableName)
     {
-        $sql = "SELECT DISTINCT idarchive FROM " . $archiveTableName . " a "
-            . " LEFT JOIN " . Common::prefixTable('site') . " s USING (idsite)"
-            . " WHERE s.idsite IS NULL"
-            . " AND ts_archived < ?";
+        $sql = "SELECT DISTINCT idsite FROM " . $archiveTableName;
+        $rows = Db::getReader()->fetchAll($sql, array());
 
-        $rows = Db::fetchAll($sql, array($oldestToKeep));
+        if (empty($rows)) {
+            return array(); // nothing to delete
+        }
 
-        return array_column($rows, 'idarchive');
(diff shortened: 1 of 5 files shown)


tests/PHPUnit/Integration/Archive/ArchivePurgerTest.php
```diff
@@ -116,8 +116,8 @@ public function test_purgeNoSiteArchives_PurgesAllNoSiteArchives()
 
         //There are 5 rows for website #3. We leave the other two because they're before our purge threshold.
         $deletedRowCount = $this->archivePurger->purgeDeletedSiteArchives($this->january);
-        $this->assertEquals(3 * RawArchiveDataWithTempAndInvalidated::ROWS_PER_ARCHIVE, $deletedRowCount);
-        self::$fixture->assertArchivesDoNotExist(array(3, 7, 10), $this->january);
+        $this->assertEquals(5 * RawArchiveDataWithTempAndInvalidated::ROWS_PER_ARCHIVE, $deletedRowCount);
```
tsteur: We are now expecting more results as for deleted sites we don't need to look at ts_archived.

core/DataAccess/Model.php
```diff
@@ -319,14 +319,30 @@ public function getSitesWithInvalidatedArchive($numericTable)
...
+        $deletedSites = array_diff($idSitesUsed, $idSitesExisting);
+
+        if (empty($deletedSites)) {
+            return array();
+        }
+        $deletedSites = array_values($deletedSites);
+
+        $sql = "SELECT DISTINCT idarchive FROM " . $archiveTableName . " WHERE idsite IN (".Common::getSqlStringFieldsArray($deletedSites).")";
+
+        $rows = Db::getReader()->fetchAll($sql, $deletedSites);
```
diosmosis: If there are hundreds or thousands of deleted sites, will this be a problem? (since we're using placeholders)

tsteur: I reckon it's unlikely this happens but will change it indeed 👍 doesn't hurt

tsteur: Will also apply this to the `deleteArchiveIds` method that is in the same model

tsteur: Done 👍

PR #14844, opened 2019-09-02
Faster raw log data deletion

tsteur:
This improves the performance for raw data deletion on the log tables further after already making partially suer the index is used in https://github.com/matomo-org/matomo/pull/14840

Before this change, when we delete raw data from log tables, queries like these would be executed:

```sql
SELECT idvisit FROM `log_visit` WHERE idvisit > 290 AND visit_last_action_time < '2019-06-02 03:04:05' AND idsite IN (1,5,7) ORDER BY idvisit ASC LIMIT 1000

SELECT idvisit FROM `log_visit` WHERE idvisit > 1290 AND  visit_last_action_time < '2019-06-02 03:04:05' AND idsite IN (1,5,7) ORDER BY idvisit ASC LIMIT 1000

Where idvisit > 2290 ...
...
```

However, this means for raw data deletion, MySQL needs to look at each visit within that time range  , read the idvisit, store them in memory or tmp table, and order them afterwards. That's not quite efficient when we are deleting because we could simply always execute the same query:

```sql
SELECT idvisit FROM `log_visit` WHERE visit_last_action_time < '2019-06-02 03:04:05' AND idsite IN (1,5,7) LIMIT 1000
```

Now Mysql can just randomly look only at 1000 visits which is fast because of the `idsite, visit_last_action_time` index instead of looking at potentially many millions of visits and sorting them etc. It's a lot more efficient and reduces IO quite a bit. Especially considering we're executing this query VERY often when there are millions of visits to be deleted where we looked say 10,000 times over all visits in that time range just to delete 10M visits (10,000 times executing the query to delete 1000 visits each time). 

Noticed the `forAllLogs()` method is also called from `VisitorGeolocator` where it is not deleting data and therefore it actually needs the `idvisit > ? ORDER BY idvisit` logic as it otherwise can't make sure to apply the callback logic to each method. 

Did my best to keep this logic difference somewhat simple. Would otherwise need two different methods for this but that doesn't make things better in the end.

### DIFF
--- a/core/DataAccess/RawLogDao.php
+++ b/core/DataAccess/RawLogDao.php
@@ -103,15 +103,16 @@ class RawLogDao
      *                            ```
      * @param int $iterationStep The number of rows to query at a time.
      * @param callable $callback The callback that processes each chunk of rows.
+     * @param string $willDelete Set to true if you will make sure to delete all rows that were fetched. If you are in
+     *                           doubt and not sure if to set true or false, use "false". Setting it to true will
+     *                           enable an internal performance improvement but it can result in an endless loop if not
+     *                           used properly.
      */
     public function forAllLogs($logTable, $fields, $conditions, $iterationStep, $callback, $willDelete)
     {
         $lastId = 0;
 
         if ($willDelete) {
-            // it's fine if the reader has a delay and a very few visits are missed...
-            // usually we look at old data and they would be already synced/replicated on the reader for ages
-            $db = Db::getReader();
             // we don't want to look at eg idvisit so the query will be mostly index covered as the
             // "where idvisit > 0 ... ORDER BY idvisit ASC" will be gone... meaning we don't need to look at a huge range
             // of visits...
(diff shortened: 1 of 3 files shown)


core/DataAccess/RawLogDao.php
```diff
@@ -103,23 +103,39 @@ public function countVisitsWithDatesLimit($from, $to)
...
      */
-    public function forAllLogs($logTable, $fields, $conditions, $iterationStep, $callback, $useReader = false)
+    public function forAllLogs($logTable, $fields, $conditions, $iterationStep, $callback, $willDelete)
     {
-        $idField = $this->getIdFieldForLogTable($logTable);
-        list($query, $bind) = $this->createLogIterationQuery($logTable, $idField, $fields, $conditions, $iterationStep);
-
         $lastId = 0;
-        if ($useReader) {
-            $db = Db::getReader();
```
tsteur: we can no longer use the reader for this but that's fine as the query is now fast.

If we were still using the reader, we would risk reading visits again that were already deleted on the master but the delete was not yet replicated to the reader...

PR #14852, opened 2019-09-04
Fix percentage calculation on world map

### LINKED ISSUE #14849: Wrong percentage on Visitors Location page
Hi, there is a diff in unique visitors percent stat on the Visitors Location page (`/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=day&date=yesterday&category=General_Visitors&subcategory=UserCountry_SubmenuLocations`) between the Country map widget and the Country table widget.

See the pictures below (19,7% for France on the map, 23,9% for France on the table, both for the same unique visitors number).

Capture d’écran 2019-09-03 à 11 01 30
Capture d’écran 2019-09-03 à 11 01 05

sgiehl:
fixes #14849

### DIFF
--- a/plugins/UserCountryMap/Controller.php
+++ b/plugins/UserCountryMap/Controller.php
@@ -16,6 +16,7 @@ use Piwik\Container\StaticContainer;
 use Piwik\Piwik;
 use Piwik\Plugins\Goals\API as APIGoals;
 use Piwik\Plugins\VisitsSummary\API as VisitsSummaryAPI;
+use Piwik\SettingsPiwik;
 use Piwik\Site;
 use Piwik\Translation\Translator;
 use Piwik\View;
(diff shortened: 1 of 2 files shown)


plugins/UserCountryMap/Controller.php
```diff
@@ -286,10 +287,7 @@ private function getMetrics($idSite, $period, $date, $token_auth)
         $metrics = array();
         if (!empty($metaData[0]['metrics']) && is_array($metaData[0]['metrics'])) {
             foreach ($metaData[0]['metrics'] as $id => $val) {
-                // todo: should use SettingsPiwik::isUniqueVisitorsEnabled ?
-                if (Common::getRequestVar('period') == 'day' || $id != 'nb_uniq_visitors') {
-                    $metrics[] = array($id, $val);
-                }
+                $metrics[] = array($id, $val);
```
sgiehl: That check should not be needed, as the `API.getMetadata` should only return metrics that are available for the given period.

plugins/UserCountryMap/javascripts/visitor-map.js
```diff
@@ -197,7 +197,7 @@
                 if (metric.substr(0, 3) == 'nb_' && metric != 'nb_actions_per_visit') {
                     var total;
                     if (id.length == 3) total = UserCountryMap.countriesByIso[id][metric];
-                    else if (id == 'world') total = _worldTotal;
+                    else if (id == 'world') total = self.config.visitsSummary[metric];
```
diosmosis: I think we can get rid of the `_worldTotal` variable in the whole file, it doesn't appear to be used anywhere else.

sgiehl: removed the unused vars

PR #14856, opened 2019-09-04
Make it possible to force archiving for custom date ranges

sgiehl:
In some cases it might be very useful to be able to define custom date ranges to be archived by `core:archive`. E.g. if someone always wants to get the last 10 days stats via api, there is currently no way to get that pre-archived.

The new config option would make it possible to define any range that is valid to be preprocessed


viezel:
👏 yes please

core/CronArchive.php
```diff
@@ -1685,9 +1685,17 @@ private function getCustomDateRangeToPreProcess($idSite)
...
+
         if (empty($cache[$idSite])) {
-            return array();
+            $cache[$idSite] = array();
+        }
+
+        $customRanges = array_filter(Config::getInstance()->General['archiving_custom_ranges']);
+
+        if (!empty($customRanges)) {
+            $cache[$idSite] += $customRanges;
```
tsteur: not quite sure but wouldn'T that be an array_merge?

config/global.ini.php
```diff
@@ -362,6 +362,11 @@
 ; or make sure the date ranges users' want to see will be processed somehow.
 archiving_range_force_on_browser_request = 1
 
+; By default Matomo will automatically archive all date ranges any user has chosen in his account settings
+; This is limited to the available options last7, previous7, last30 and previous30
+; If you need any other period, or want to ensure one of those is always archived, you can define them here
+archiving_custom_ranges[] =
```
tsteur: Might need to update a screenshot test I suppose

tsteur (changes requested):

core/CronArchive.php
sgiehl: yes. indeed

sgiehl:
@tsteur updated

PR #14857, opened 2019-09-04
Tweak behaviour of orphaned segment archive purge

### LINKED ISSUE #14828: Tweak behaviour of orphaned segments
Weekly we have a cronjob to delete orphaned segments in `CoreAdminHome\Tasks`. Meaning we delete all segment archives of segments that no longer exist. 

This behaviour currently works like this:
* Get a list of all segments
* Iterate over each archive table
   * Get the idArchives of archives whose done flag is not in the list of existing segments
   * Delete the idArchives

Now that this command ran a few times and we have cleaned up all segment archives, and now that we implement a proper soft delete in https://github.com/matomo-org/matomo/issues/14826 we could make this behaviour more efficient:

* Get all segments where `deleted = 1 and ts_last_edit >= 1 week ago` (assuming we run the task weekly). These are the segments that were deleted last week
* Iterate over all archive tables 
   * Get the idarchives of segments having the done flag for the deleted segment
   * Delete these idarchives from numeric and blob table

I was thinking we would only need to loop over archive tables since `ts_created` but we need to loop likely over all archive tables since some users will have enabled to automatically archive segments for all historical data which is also the default `process_new_segments_from = "beginning_of_time"`

This way things will be faster as we only execute logic if a segment was deleted

katebutler:
Fixes #14828


core/DataAccess/ArchiveWriter.php
```diff
@@ -171,6 +171,7 @@ private function getModel()
     protected function logArchiveStatusAsIncomplete()
     {
         $this->insertRecord($this->doneFlag, self::DONE_ERROR);
+        $this->flushSpool('numeric');
```
tsteur: btw I had removed that one before merging the PR re the spools as after some more investigating found it wasn't needed.

plugins/SegmentEditor/Model.php
```diff
@@ -129,6 +129,74 @@ public function getSegmentByDefinition($definition)
...
+    /**
+     * Gets a list of segments that have been deleted in the last week and therefore may have orphaned archives.
+     * @param Date $date Segments deleted on or after this date will be returned.
+     * @return array of segments. The segments are only populated with the fields needed for archive invalidation
+     * (e.g. definition, enable_only_idsite).
+     * @throws \Exception
+     */
+    public function getSegmentsDeletedSince(Date $date)
+    {
+        $dateStr = $date->toString('Y-m-d');
```
tsteur: Better to use `$date->getDatetime()`

plugins/SegmentEditor/Model.php
```diff
@@ -129,6 +129,74 @@ public function getSegmentByDefinition($definition)
...
+        $sql = "SELECT DISTINCT definition, enable_only_idsite FROM " . Common::prefixTable('segment')
+            . " WHERE deleted = 1 AND ts_last_edit >= '$dateStr'";
```
tsteur: here we want to use bound parameters... meaning `AND ts_last_edit >= ?` and then below we use `Db::fetchAll($sql, array($dateStr))`. In this case it wouldn't be a security issue to not use bound parameters but it's still better in general and makes sure it can't become a security issue in the future

core/DataAccess/Model.php
```diff
@@ -354,50 +354,47 @@ public function getArchiveIdsForDeletedSites($archiveTableName)
...
-        return array_map(function($row) { return $row['idarchive']; }, $rows);
+        return array_column($rows, 'idarchive');
+    }
+
+    private static function getDeletedSegmentWhereClause(array $segment)
+    {
+        $idSite = (int)$segment['enable_only_idsite'];
+        $segmentHash = Segment::getSegmentHash($segment['definition']);
+        // Valid segment hashes are md5 strings - just confirm that it is so it's safe for SQL injection
+        if (!preg_match('/^[a-z0-9A-Z]+$/', $segmentHash)) {
```
tsteur: just btw here we can use `ctype_xdigit($segmentHash)`

core/DataAccess/Model.php
```diff
@@ -354,50 +354,47 @@ public function getArchiveIdsForDeletedSites($archiveTableName)
...
+            throw new Exception($segment . ' expected to be an md5 hash');
+        }
+
+        $nameClause = 'name LIKE "done' . $segmentHash . '%"';
+        $idSiteClause = '';
+        if ($idSite > 0) {
+            $idSiteClause = ' AND idsite = ' . $idSite;
+        } elseif (! empty($segment['idsites_to_preserve'])) {
+            // A segment for all sites was deleted, but there are segments for a single site with the same definition
+            $idSiteClause = ' AND idsite NOT IN (' . implode(',', $segment['idsites_to_preserve']) . ')';
```
tsteur: To be safe re possible SQL injection we should do `$segment['idsites_to_preserve'] = array_map('intval', $segment['idsites_to_preserve'])`

core/DataAccess/Model.php
```diff
@@ -354,50 +354,47 @@ public function getArchiveIdsForDeletedSites($archiveTableName)
...
+            . ' WHERE ts_archived < ?'
+            . ' AND (' . $segmentClauses . ')';
 
         $rows = Db::fetchAll($sql, array($oldestToKeep));
 
-        return array_map(function($row) { return $row['idarchive']; }, $rows);
+        return array_column($rows, 'idarchive');
+    }
+
+    private static function getDeletedSegmentWhereClause(array $segment)
```
tsteur: btw any reason this method is static? Looks like it's not needed?

core/DataAccess/Model.php
```diff
@@ -354,50 +354,47 @@ public function getArchiveIdsForDeletedSites($archiveTableName)
...
-            $idSite = (int)$idSite;
-
-            // Vanilla case - segments that are valid for a single site only
-            $sql = '(idsite = ' . $idSite . ' AND (';
-            $sql .= 'name LIKE "done' . implode('%" OR name LIKE "done', $segments) . '%"';
-            $sql .= '))';
-            $validSegmentClauses[] = $sql;
+        $segmentClauses = [];
+        foreach ($segments as $segment) {
+            $segmentClauses[] = self::getDeletedSegmentWhereClause($segment);
```
tsteur: btw be good to add the segmentClause only if
```
if(!empty($segment['definition'])) {
     $segmentClauses[] = self::getDeletedSegmentWhereClause($segment);
 }
```

because otherwise it would delete archives where `done%`. It shouldn't ever happen but could to have a safety net here just in case in the future somewhere `definition` suddenly gets lost or someone is not using the method correctly.

plugins/SegmentEditor/Model.php
tsteur: Actually seeing it seems to be expecting a timestamp so might have to use `getTimestamp()` alternatively

core/DataAccess/Model.php
tsteur: Actually this will be really important. I just had triggered a case where `definition` wasn't set for some reason. So we need to make sure to not accidentally delete regular archives

tsteur:
Found one bug @katebutler 

I had these segments (among some others):
```
INSERT INTO `piwik_segment` (`idsegment`, `name`, `definition`, `login`, `enable_all_users`, `enable_only_idsite`, `auto_archive`, `ts_created`, `ts_last_edit`, `deleted`)
VALUES
	(48, 'exclude funds library', 'customVariableValue1!@HTTP-Grabber%2520www.fundslibrary.co.uk', 'root', 0, 1, 1, '2019-04-08 02:43:54', '2019-08-30 22:48:49', 1),
	(49, 'tets3', 'actions>=1', 'root', 0, 0, 1, '2019-08-29 02:14:09', '2019-08-30 22:48:49', 0),
	(50, 'test23434', 'actions>=1', 'root', 0, 0, 1, '2019-08-29 02:14:54', '2019-08-30 22:48:49', 1),
	(51, 'tererer', 'actions>=1', 'root', 0, 3, 1, '2019-08-29 02:15:33', '2019-08-30 22:48:49', 0),
	(52, 'tererer', 'actions>=1', 'root', 0, 1, 1, '2019-08-29 02:15:42', '2019-08-30 22:48:49', 1);

```

and it generated this output

```
array (
  0 => 
  array (
    'definition' => 'customVariableValue1!@HTTP-Grabber%2520www.fundslibrary.co.uk',
    'enable_only_idsite' => '1',
    'idsites_to_preserve' => 
    array (
    ),
  ),
  1 => 
  array (
    'idsites_to_preserve' => 
    array (
      0 => '3',
    ),
  ),
)
```

Causing the deletion of regular segments. The problem is that it first was doing `unset($deletedSegments[$i]);` and then in the next iteration it was doing `$deletedSegments[$i]['idsites_to_preserve'][] = $existing['enable_only_idsite'];`. I reckon it might help to `break` the loop after calling `unset($deletedSegments[$i]);`.

tsteur:
BTW Might be good to add a test for this particular case where the logic was failing

core/DataAccess/Model.php
tsteur: @katebutler be good to not have the method static if possible

core/DataAccess/Model.php
tsteur: @katebutler be good to add the definition check just to be 100% safe.

plugins/SegmentEditor/Model.php
```diff
@@ -129,6 +129,75 @@ public function getSegmentByDefinition($definition)
...
+        $dateStr = $date->getDatetime();
+        $sql = "SELECT DISTINCT definition, enable_only_idsite FROM " . Common::prefixTable('segment')
+            . " WHERE deleted = 1 AND ts_last_edit >= ?";
+        $deletedSegments = Db::fetchAll($sql, array($dateStr));
+
+        if (empty($deletedSegments)) {
+            return array();
+        }
+
+        $existingSegments = self::getExistingSegmnetsLike($deletedSegments);
```
tsteur: just btw there is small typo `getExistingSegmnetsLike ` => `getExistingSegmentsLike `

plugins/SegmentEditor/Model.php
```diff
@@ -129,6 +129,75 @@ public function getSegmentByDefinition($definition)
...
+                }
+            }
+        }
+
+        return $deletedSegments;
+    }
+
+    private static function getExistingSegmnetsLike(array $segments)
+    {
+        $whereClauses = array();
```
tsteur: be good to add a little check `if (empty($segments)) { return array(); }`. I know this is done in the method that is calling this method already, be just good to add this here as well in case the method will be used in the future somewhere else too and then it prevents the case of having a broken sql..   
also be good to remove the `static` and make it a regular method

plugins/SegmentEditor/Model.php
```diff
@@ -129,6 +129,75 @@ public function getSegmentByDefinition($definition)
...
+        $bind = array();
+        foreach ($segments as $segment) {
+            $bind[] = $segment['definition'];
+            if ($segment['enable_only_idsite'] == 0) {
```
tsteur: @katebutler but concerned about this code here: https://github.com/matomo-org/matomo/blob/3.12.0-b3/core/Segment.php#L104-L108

Not sure if this can have any effect on the condition. I wonder if there are cases where we would store `actionType==sitesearches` as definition in the DB, and if passed differently in the API if we would also accept / store `"actionType%3D%3Dsitesearches` in case it is double encoded and not just once... according to the logic both segment definitions would be accepted in that class and have the same hash. The segment editor would probably store the segments always in the same way, but if someone uses the API and double encodes it, this might still work by the looks and we would store it urlencoded...

actually I just tested it and this would be the case:
image

See `https://example.com/index.php?module=API&method=SegmentEditor.add&format=JSON&token_auth=88d86212d1967a8565d05e51b1bfaa19&idSite=1&name=foo&definition=actionType%253D%253Dsitesearches`   vs using `actionType%3D%3Dsitesearches` as definition.

So it seems we also need to do `(definition = ? or definition = ? or definition = ?)` with a bind of `$definition, urlencode($definition), urldecode($definition)` for each segment :( because we can't know if the given definition is urlencoded or urldecoded etc.

tsteur (approved):

plugins/SegmentEditor/Model.php
```diff
@@ -129,6 +129,87 @@ public function getSegmentByDefinition($definition)
...
+    /**
+     * Gets a list of segments that have been deleted in the last week and therefore may have orphaned archives.
+     * @param Date $date Segments deleted on or after this date will be returned.
+     * @return array of segments. The segments are only populated with the fields needed for archive invalidation
+     * (e.g. definition, enable_only_idsite).
+     * @throws \Exception
+     */
+    public function getSegmentsDeletedSince(Date $date)
+    {
+        $dateStr = $date->getDatetime();
```
tsteur: @katebutler not 100% sure why but for my MySQL version it works only when using `getTimestamp()` instead of `getDatetime()`. Makes kind of sense since the data is stored as timestamp but not sure how it works somewhere else.

Can you test to use a `getTimestamp()` and see what it does for you if tests still pass? Then this seems to be the safer option

katebutler: As an example `$date->getTimestamp()` is 1567934627 for 2019-09-08. On my localhost when I use that value in the query it matches any record with `ts_last_edit` not null, even if I set it right back as close to epoch as mysql will let me go (1970-01-01 12:01). This causes the relevant test to fail as the code picks up records that have been deleted more than a week ago.
I am on quite an old version of mysql (5.7.27).

tsteur: All good! The date in my virtual machine was outdated :( works 👍

PR #14858, opened 2019-09-04
New INI setting live_query_max_execution_time to allow configuration of max execution time

tsteur:
cc @mattab 

Users can now optionally configure a max_execution_time for live queries to prevent a live query from running for a very long time eg when looking at a year with a segment that doesn't find any visitors.

Added as many tests as I could. Noticed the MySQL version on Travis doesn't support this feature yet otherwise could have tested it there further.


plugins/Live/Model.php
```diff
@@ -112,6 +115,84 @@ public function queryLogVisits($idSite, $period, $date, $segment, $offset, $limi
...
+                 * @ignore
+                 * @internal
+                 */
+                Piwik::postEvent('Live.queryMaxExecutionTimeExceeded', array($params));
+                throw new MaxExecutionTimeExceededException($message);
+            }
+            throw $e;
+        }
+        return $visits;
+    }
```
diosmosis: This code seems important. Would it be possible to unit/integration test this code w/ a mock Db object returned by `Db::getReader()`?

mattab:
LGTM :+1: (Benaka left a comment above)

Findus23:
Unfortunately this syntax isn't supported in MariaDB which has another synax to set the maximum execution time for a query:
https://mariadb.com/kb/en/library/aborting-statements/

tsteur:
thanks @Findus23 I will mention in the config that it likely won't work on MariaDB

tsteur:
Added some tests @diosmosis

PR #14860, opened 2019-09-05
Update shown logo in the app

### LINKED ISSUE #14795: Show Matomo logo in the app
refs MAR-575

We're wanting to show the Matomo logo/icon in the app like this:

image

The mobile version would basically look like this:

image

Here's the high res circle image with transparent background:
image

Problems:
* I don't think we have the icon as an SVG but are showing the logo as an SVG in all places.
* If we decide to show an image instead, we would need to combine above circle and the word "Matomo" into one image. But likely it be blurry in various places.

Ideally we create an SVG version of the logo.

tsteur:
fix https://github.com/matomo-org/matomo/issues/14795


tsteur:
@Findus23 or @diosmosis do you maybe have an idea why the icons would slightly change when nothing else changes and the PR shouldn't change it? There seem to be minor changes will icons in the all the screenshots https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/35792/UIIntegrationTest_dashboard1.png

tsteur:
Before I update all screenshots and they become outdated again etc it be good to review this issue. Looking at https://builds-artifacts.matomo.org/matomo-org/matomo/14795/35795/
 makes it also easier to review the updated screenshots. Once it is fine I will update the screenshots and fix the tests

diosmosis:
The icons look the same to my eye, maybe a change here subtly changes how chrome renders the icons. The size is the same and it basically looks the same, so I suppose it's ok to just copy over the screenshots?

tsteur:
Updated the screenshots @diosmosis

PR #14866, opened 2019-09-07
Fix user path was not working as it should (allow placing config file outside webroot)

### LINKED ISSUE #13037: global(.ini).php in local PIWIK_USER_PATH not updated (composer)
Hi,

We update matomo with composer and host a central piwik/matomo, with local config files for each directadmin account. 

The constant 'PIWIK_USER_PATH' is set to the right config-dir inside an user-account. That works fine!

Matomo also needs 'global.php' and 'global.ini.php' to be in this local dir. In the above configuration these file are not updated, when composer updates the central matomo. 
Since it is adviced NOT to change these global files, what is the purpose of keeping them in a local user-path? 

This issue could be fixed if these file would be (also) loaded from the central matomo. Or that you could change the absolute paths in bootstrap.php.
Same for common.ini.php > here an option to set the path would be best, so you can have a own central common.ini.php somewhere else on your server.

When looking into the code of matomo, I see that the config files are chained. 
Why not have a chain like, load:
- from matomo vendor dir: /config/config.ini.php (holds the global (updated) settings)
- if exists /your/global/path/to/global/config/config.ini.php  > instead of common.ini.php
- PIWIK_USER_PATH/config/config.ini.php

Can this be changed in matomo? 
Or what can we do to have the global files inside the central matomo-dir?

Thanks in advance, flexjoly

tsteur:
fix https://github.com/matomo-org/matomo/issues/13037

As user described in the issue the user path behaviour didn't quite make sense. For example in an update we copied over `config/global.php` and `config/global.ini.php` to the user path. This may fail though due to write permission and is generally not needed cause the only files it should load from there is `config/config.php` and `config/config.ini.php` and `config/common.config.ini.php`. 

This makes things a lot simpler and now there is an easy way to put the config file outside Matomo webroot.

Should probably also create an FAQ for this afterwards.

### DIFF
--- a/core/Filechecks.php
+++ b/core/Filechecks.php
@@ -23,7 +23,7 @@ class Filechecks
         if (!is_writable(PIWIK_INCLUDE_PATH . '/') ||
             !is_writable(PIWIK_DOCUMENT_ROOT . '/index.php') ||
             !is_writable(PIWIK_INCLUDE_PATH . '/core') ||
-            !is_writable(PIWIK_USER_PATH . '/config/global.ini.php')
+            !is_writable(PIWIK_DOCUMENT_ROOT . '/config/global.ini.php')
         ) {
             return false;
         }
@@ -40,7 +40,9 @@ class Filechecks
     {
         $resultCheck = array();
         foreach ($directoriesToCheck as $directoryToCheck) {
-            if (!preg_match('/^' . preg_quote(PIWIK_USER_PATH, '/') . '/', $directoryToCheck)) {
+	        if (!preg_match('/^' . preg_quote(PIWIK_USER_PATH, '/') . '/', $directoryToCheck)
+	            && !preg_match('/^' . preg_quote(PIWIK_DOCUMENT_ROOT, '/') . '/', $directoryToCheck)
+	            && !preg_match('/^' . preg_quote(PIWIK_INCLUDE_PATH, '/') . '/', $directoryToCheck)) {
                 $directoryToCheck = PIWIK_USER_PATH . $directoryToCheck;
             }
 
--- a/js/tracker.php
+++ b/js/tracker.php
@@ -22,9 +22,19 @@ if ($_SERVER['REQUEST_METHOD'] == 'POST'
  *
  * @see core/Piwik.php
  */
-define('PIWIK_INCLUDE_PATH', '..');
 define('PIWIK_DOCUMENT_ROOT', '..');
-define('PIWIK_USER_PATH', '..');
+
+if (file_exists(PIWIK_DOCUMENT_ROOT . '/bootstrap.php')) {
+    require_once PIWIK_DOCUMENT_ROOT . '/bootstrap.php';
+}
+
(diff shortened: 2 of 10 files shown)


core/Filechecks.php
```diff
@@ -40,7 +40,9 @@ public static function checkDirectoriesWritable($directoriesToCheck)
     {
         $resultCheck = array();
         foreach ($directoriesToCheck as $directoryToCheck) {
-            if (!preg_match('/^' . preg_quote(PIWIK_USER_PATH, '/') . '/', $directoryToCheck)) {
+	        if (!preg_match('/^' . preg_quote(PIWIK_USER_PATH, '/') . '/', $directoryToCheck)
```
tsteur: those 3 constant might point to different directories so we need to only add another directory if none of these are set

js/tracker.php
```diff
@@ -22,9 +22,19 @@
...
  */
-define('PIWIK_INCLUDE_PATH', '..');
 define('PIWIK_DOCUMENT_ROOT', '..');
-define('PIWIK_USER_PATH', '..');
+
+if (file_exists(PIWIK_DOCUMENT_ROOT . '/bootstrap.php')) {
+    require_once PIWIK_DOCUMENT_ROOT . '/bootstrap.php';
+}
+
+if (!defined('PIWIK_INCLUDE_PATH')) {
```
tsteur: same behaviour now as in `index.php` as `bootstrap.php` could change the user path

misc/cron/updatetoken.php
```diff
@@ -15,12 +15,8 @@
 use Piwik\Tests\Framework\TestingEnvironmentManipulator;
 use Piwik\Tests\Framework\TestingEnvironmentVariables;
 
-if (!defined('PIWIK_INCLUDE_PATH')) {
```
tsteur: these constants are now removed as they are detected and set by `index.php`

plugins/CoreUpdater/Updater.php
```diff
@@ -298,11 +298,6 @@ private function installNewFiles($extractedArchiveDirectory)
             $model->removeGoneFiles($extractedArchiveDirectory, PIWIK_DOCUMENT_ROOT);
         }
 
-        // Config files may be user (account) specific
-        if (PIWIK_INCLUDE_PATH !== PIWIK_USER_PATH) {
```
tsteur: no longer needed to copy the global config files over as there might not be write permission anyway and it's not needed. They should be always in document root.

mattab:
> Should probably also create an FAQ for this afterwards.

@tsteur if you have a FAQ draft let me know, happy to publish it :+1:

tsteur:
It was already there kind of

florianjacob:
This made it impossible to use `PIWIK_USER_PATH` as an instance-specific writable location, separated from a shared, read-only `PIWIK_DOCUMENT_ROOT`: https://github.com/matomo-org/matomo/issues/15087

PR #14868, opened 2019-09-08
Rename Generic widget to KPI Metric

### LINKED ISSUE #14612: Make the "Generic > metric" widget more visible in the dashboard
A super valuable feature is the ability to add a KPI to the dashboard. It's hidden under `Generic > Metric`:

Screenshot from 2019-07-02 17-08-52

Instead:

* could we maybe rename `Generic` to `KPI Metric` 
* could we rename `Metric` to `KPI Metric`
* move this to be the first entry in the list, above `Visitors`

This way it will be more visible to users that they can add any KPi to their users.

katebutler:
Fixes #14612

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -244,6 +244,7 @@
         "InvalidResponse": "The received data is invalid.",
         "IP": "IP",
         "JsTrackingTag": "JavaScript Tracking Code",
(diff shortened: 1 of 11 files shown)


tsteur:
@katebutler haven't tested it yet but a few tests are failing. Be good to update them 👍

PR #14869, opened 2019-09-08
Stop urlencoding dashboard names before sending to API

### LINKED ISSUE #14820: Dashboard title with space
## Summary

It is possible to create several custom Dashboard without any restriction into the title of them.

## Current behavior

When I create a new Dashboard with a space into the title, the space is encoded (replaced by a `%20`)

Screen Shot 2019-08-19 at 19 15 39

## Expected behavior

Several possibilities are available

1. Prevent user to create a Dashboard with some special chars (useless because the dashboard title is not used into the URL)
2. Allow special char but display them without encodage

katebutler:
When sent as plain text, the dashboard names are passed through `htmlspecialchars` before they are written to the DB, so they can still be used in their raw form in the UI.
Fixes #14820

### DIFF
--- a/plugins/Dashboard/javascripts/dashboard.js
+++ b/plugins/Dashboard/javascripts/dashboard.js
@@ -21,7 +21,7 @@ function createDashboard() {
             format: 'json'
         }, 'get');
         ajaxRequest.addParams({
-            dashboardName: encodeURIComponent(dashboardName),
+            dashboardName: dashboardName,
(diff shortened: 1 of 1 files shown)


jbrule:
I'm seeing url encoded dashboard names and I wonder if this PR is related. I am on v3.13.3.

Screen Shot 2020-03-10 at 1 26 46 PM

PR #14870, opened 2019-09-08
Make it easier to search help resources from within Matomo

### LINKED ISSUE #14626: Make it easy to search Matomo.org user guides and FAQs within the app
The goal of this issue is to make it very easy and obvious for users to search through our great Matomo.org resources (user guides, FAQs) directly from within the app. While keeping it very simple for us to implement. 

Suggested places where people can access Matomo.org search:

* When clicking on the Help icon, the Help page opens, and at the top of this Help page, we'd display a big search field. 
* When searching in the "Awesome search" field (at the top of the left menu), a new dropdown element would appear at the bottom.

## Help page design

At the top of [our Help page](https://demo.matomo.org/index.php?module=Feedback&action=index&idSite=1&period=week&date=today), above the "About Matomo 3.10.0" block, we would:

*  add a new block "How can we help?" with below a big search field. The search field would show a grey text like "Search Matomo.org help resources" or so. 
* add a link to `Browser all user guides` and to `Browse all FAQs`

It could look like Slack help page:

Screenshot from 2019-07-03 17-10-16

(we probably don't need the "Popular help topics" section for now)

## Awesome menu

When any term is entered in the awesome menu, the dropdown would include a new element, at the very bottom, called eg `Help resources`, with a sub-item called eg. `Search KEYWORD_HERE on Matomo.org`

Example suggested mockup:
Screenshot from 2019-07-05 12-31-53

## Behaviour

In both cases, when people enter a search term, and click the button or link, a new tab would open in their browsers, and open the Matomo.org search result page. For example if you search for `custom dimension` it would open: https://matomo.org/?s=custom+dimension

We don't want to include the help resources within the app because it is too much work to implement. Rather we want to do the easiest solution that will still provide great value to users when they need help.

As always, thoughts/suggestions welcome!

katebutler:
Fixes #14626


plugins/CoreHome/angularjs/quick-access/quick-access.directive.html
```diff
@@ -36,5 +36,12 @@
                 ng-click="quickAccess.selectSite(site.idsite)"
                 ng-repeat="site in quickAccess.sitesModel.sites"><a ng-bind="site.name"></a></li>
         </ul>
+        <ul>
+            <li class="quick-access-category helpCategory">{{ 'General_HelpResources' | translate }}</li>
+            <li ng-class="{selected: search.index == 'help'}"
+                ng-mouseenter="search.index='help'">
+                    <a href="https://matomo.org?s={{ urlEncode(search.term) }}" target="_blank">Search '{{ search.term }}' on matomo.org</a>
```
tsteur: @katebutler this string needs to be translatable... like `{{ 'foobar_xyz'|translate:search.term }}`

tsteur: Might also need to check if `ng-href` is a better fit. Might be more secure and might not need the `urlencode` then. Not 100% sure though but think so

plugins/Feedback/templates/index.twig
```diff
@@ -8,6 +8,24 @@
 {% block content %}
 
     <div id="feedback-faq" class="admin">
+        <div piwik-content-block content-title="{{ "Feedback_HowCanWeHelp"|translate|e('html_attr') }}" class="feedback-help-resources">
+            <form target="_blank" method="GET" action="https://matomo.org/">
+                <div class="search-matomo-form">
+                    <div>
+                        <span class="icon-search"></span>
+                        <input type="text" placeholder="{{'Feedback_SearchHelpResources'|translate|e('html_attr')}}" name="s"/>
```
tsteur: This isn't using our form widgets like `<div piwik-field uicontrol="text"....>`

tsteur: Might need also generally need to look more like a search field maybe like but with the icon on the left 
image

BTW we're using MaterializeCSS see http://archives.materializecss.com/0.100.2/shadow.html so we could use classes like `z-depth-1`. If we make the search look like above we would maybe not need to reuse our builtin widgets.

In general be good to improve the look a bit. Could we maybe make it look like this?

image  (the search icon needs to be vertically centered)

plugins/Feedback/templates/index.twig
```diff
@@ -8,6 +8,24 @@
 {% block content %}
 
     <div id="feedback-faq" class="admin">
+        <div piwik-content-block content-title="{{ "Feedback_HowCanWeHelp"|translate|e('html_attr') }}" class="feedback-help-resources">
```
tsteur: I reckon be good to move this section under "About Matomo". 

We then need to see how to further optimise it as there will be three different help sections...  Community help also partially repeats the guides and FAQs. People might not notice the community help and professional help section when we show the "how can we help" section... 
image

@katebutler 
could you merge the three sections into one section similar to this?
image

PR #14873, opened 2019-09-09
fix magic quotes deprecation warning

Findus23:
reported in https://forum.matomo.org/t/quickform2-php-109-deprecated-function-get-magic-quotes-gpc-is-deprecated/34121
related to https://github.com/matomo-org/matomo/issues/6339
related to https://github.com/matomo-org/matomo/issues/14821 (deprecation warning is new in 7.4)

`get_magic_quotes_gpc()` seems to return if magic quotes are enabled. As it has been removed with PHP 5.4 it can be simply replaced with a hardcoded `False`.


tsteur:
👍

PR #14875, opened 2019-09-09
remove Zend Config

Findus23:
I think (and PHPstorm tends to agree) that this code isn't used. It seems to consist of old PHP code that implements things that are already done differently in Matomo and uses deprecated PHP features like `each()`.

The tests will show if this is really unused.


tsteur:
LGTM for me. I see `Zend_Config` is being used itself a few times but you're keeping that file in the repo so should be all fine.

PR #14876, opened 2019-09-10
Allow custom DB adapter in tracker mode

tsteur:
In regular mode we already use the above logic see https://github.com/matomo-org/matomo/blob/3.x-dev/core/Db/Adapter.php#L68-L74

This means in regular mode any plugin can already define a custom adapter and configure it. This was not working in tracking mode where the two PDO and MySQLI modes were hard coded. This was likely because back in the days we weren't using auto loading in tracker mode which we are doing now already for many years.

### DIFF
--- a/core/Tracker/Db.php
+++ b/core/Tracker/Db.php
@@ -261,18 +261,13 @@ abstract class Db
          */
         Piwik::postEvent('Tracker.getDatabaseConfig', array(&$configDb));
 
-        switch ($configDb['adapter']) {
-            case 'PDO\MYSQL':
-            case 'PDO_MYSQL': // old format pre Piwik 2
-                require_once PIWIK_INCLUDE_PATH . '/core/Tracker/Db/Pdo/Mysql.php';
-                return new Mysql($configDb);
+        $className = 'Piwik\Tracker\Db\\' . str_replace(' ', '\\', ucwords(str_replace(array('_', '\\'), ' ', strtolower($configDb['adapter']))));
 
-            case 'MYSQLI':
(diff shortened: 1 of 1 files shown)


core/Tracker/Db.php
```diff
@@ -261,18 +261,13 @@ public static function factory($configDb)
...
-        switch ($configDb['adapter']) {
-            case 'PDO\MYSQL':
-            case 'PDO_MYSQL': // old format pre Piwik 2
-                require_once PIWIK_INCLUDE_PATH . '/core/Tracker/Db/Pdo/Mysql.php';
-                return new Mysql($configDb);
-
-            case 'MYSQLI':
-                require_once PIWIK_INCLUDE_PATH . '/core/Tracker/Db/Mysqli.php';
-                return new Mysqli($configDb);
+        $className = 'Piwik\Tracker\Db\\' . str_replace(' ', '\\', ucwords(str_replace(array('_', '\\'), ' ', strtolower($configDb['adapter']))));
```
diosmosis: Does this mean it's not possible to define a db adapter in a plugin (because the namespace has to be in `Piwik\Tracker\Db`? I'm wondering if getting the class name through DI or an event might be better.

tsteur: It works when you load the file where the class for this adapter is loaded yourself. DI could work as well, I just didn't want to change too much logic as in the regular DB factory it works already like this and also has the namespace `Piwik\Db...`. I reckon it's unlikely other people write their own adapters so maybe it's not too important, but could of course also change it towards DI but actually not sure if DI is already loaded at the time the DB is created? Likely it is ... but maybe not during installation etc? Not sure...

diosmosis: No worries, if it's already an established method I can merge

PR #14885, opened 2019-09-10
Show unique visitors metric only if available

### LINKED ISSUE #14881: Unique Visitors metric appear in Graph metric selector for periods for which the metric is disabled
Unique Visitors metric appear in Graph metric selector for periods for which the metric is disabled, causing confusion amongst users. And also many support requests to our team.

--> When [Unique Visitors metric is disabled](https://matomo.org/faq/how-to/faq_113/) for the currently selected period, let's hide it from the evolution graph metric picker.

This is what it currently looks like (no data for Unique Visitors): 
Screenshot from 2019-09-10 16-09-37

sgiehl:
fixes #14881

### DIFF
--- a/plugins/VisitsSummary/Controller.php
+++ b/plugins/VisitsSummary/Controller.php
@@ -15,6 +15,7 @@ use Piwik\DataTable\Row;
 use Piwik\FrontController;
 use Piwik\Piwik;
 use Piwik\Plugins\CoreVisualizations\Visualizations\Sparklines;
+use Piwik\SettingsPiwik;
 use Piwik\Site;
 use Piwik\Translation\Translator;
 use Piwik\View;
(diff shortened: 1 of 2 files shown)


plugins/VisitsSummary/Controller.php
```diff
@@ -103,6 +104,12 @@ public function getEvolutionGraph()
             'avg_time_generation'
         );
 
+        $currentPeriod = Common::getRequestVar('period', false);
+
+        if (!SettingsPiwik::isUniqueVisitorsEnabled($currentPeriod)) {
+            $selectableColumns = array_diff($selectableColumns, ['nb_uniq_visitors']);
```
tsteur: does it also need to remove nb_users?

tsteur: I'm not sure but there might be a method to get the unique metrics?

sgiehl: true, `nb_users` needs to be removed as well. But I didn't find any method returning unique metrics...

tsteur: Should we maybe add one? Makes it just easier in the future... seeing similar behaviour in `VisitsSummary\Api` etc and be maybe good to no longer hard code these metrics

sgiehl: Actually I'm not sure if that would the optimal solution. I would prefer having a Metric class for each available metric, so we could use a `isAvailable` method to check each metric. Also that would make it easier to get the translation, documentation and stuff like that for each metric.
But that's a bit a bigger task we could maybe do for Matomo 4.

PR #14890, opened 2019-09-11
Ensure all plugins are archived when range archiving is triggered by core:archive

### LINKED ISSUE #14889: Range archives might not be fully processed by core:archive
When archiving runs for a range like `last30` it's expected that all archives (for all plugins) are being processed. Currently it seems not all reports are being processed. For example all Event reports are not processed. 

Steps to reproduce:

* Set up a matomo site containing some tracked events for the last 30 days.
* Disable browser triggered archiving in UI or in config 
* set `[General] archiving_range_force_on_browser_request = 0` to disable automatic range archiving in normal requests
* to ensure the range `last30` is archived, selected it as default for a user or add `[General] archiving_custom_ranges[] = last30` to config  (last one only works with latest beta)
* run `./console core:archive --force-idsites=1` and ensure something like this is in the output: 
```
INFO [2019-09-11 13:38:32] 2049  Will pre-process for website id = 1, period = range, date = last30
INFO [2019-09-11 13:38:32] 2049  - pre-processing all visits
INFO [2019-09-11 13:38:41] 2049  Archived website id = 1, period = range, 0 segments, 1543 visits in last 30 ranges, 1543 visits this range, Time elapsed: 8.978s
```
* Open Matomo and look at the events reports or requests them using API. Result will be empty, but should actually contain the event data for the range. Other reports work as expected

sgiehl:
fixes #14889


core/ArchiveProcessor/Rules.php
```diff
@@ -57,7 +57,7 @@ public static function getDoneStringFlagFor(array $idSites, $segment, $periodLab
 
     public static function shouldProcessReportsAllPlugins(array $idSites, Segment $segment, $periodLabel)
     {
-        if ($segment->isEmpty() && $periodLabel != 'range') {
+        if ($segment->isEmpty() && ($periodLabel != 'range' || SettingsServer::isArchivePhpTriggered())) {
```
sgiehl: Not sure if the range check makes sense at all. But at least all reports should be archived when triggered by `core:archive`

diosmosis: I think the check is just to make browser triggered archiving of ranges faster. Otherwise, I guess it might take a bit longer for the user to see reports. Not sure how much time it would actually add though...

diosmosis:
Code looks good, can we add a test for this in ArchiveCronTest?

sgiehl:
@diosmosis added a test

Issue #12963, opened 2018-05-23
Warning when using Matomo + WordPress: "Warning: session_cache_limiter(): Cannot change cache limiter when session is active ..."


aclion111:
I've installed
```
	Matomo v3.5.0
```

on Nginx v1.14.0

It's working as expected as a standalone instance.

I've also installed

```
	WordPress 4.9.6
	WP-Matomo (WP-Piwik) plugin, v1.0.19
```

It's connected & tracking at my site

```
	"WP-Piwik 1.0.19 is successfully connected to Piwik 3.5.0. You are running WordPress 4.9.6."
```

In my WP site's Admin UI, I see this warning on each page:

```
	Warning: session_cache_limiter(): Cannot change cache limiter when session is active in /srv/www/piwik/core/bootstrap.php on line 32
```

I asked about this is WP-Matomo support forum @

> session_cache_limiter warnings when WP-PIWIK is enabled

https://wordpress.org/support/topic/session_cache_limiter-warnings-when-wp-piwik-is-enabled/#post-10311647

where the response by plugin's dev @braekling included

```
	"...
	There is nothing WP-Matomo can do about this. Matomo itself and WordPress or other plugins should check if the session is started already before calling session_cache_limiter.
	..."
```

So, I'm checking here next.

Notice: Constant PIWIK_ENABLE_SESSION_START already defined in /srv/www/piwik/bootstrap.php on line 8

I did find this old issue by @robocoder

> Should dispatcher always start/resume session? #885

https://github.com/matomo-org/matomo/issues/885

which seems to be (?) related, but looks like it was fixed long ago.

I did try adding

```
	define('PIWIK_ENABLE_SESSION_START', true);
```

or

```
	define('PIWIK_ENABLE_SESSION_START', false);
```

to Matomo's ```piwik/bootstrap.php```, but that doesn't change the behavior.  I just get

```
	Notice: Constant PIWIK_ENABLE_SESSION_START already defined in /srv/www/piwik/bootstrap.php on line 8

	Warning: session_cache_limiter(): Cannot change cache limiter when session is active in /srv/www/piwik/core/bootstrap.php on line 32
```

What in Matomo core code or config needs to be changed to correct this session_cache_limiter issue when used with WP/WP-Matomo plugin?

Or is this _not_ a Matomo issue either, and needs to be raised @ WP Core?

robocoder:
[...]

PR #14896, opened 2019-09-17
If session was already started, do not start it again

tsteur:
refs https://github.com/matomo-org/matomo/issues/12963
refs https://forum.matomo.org/t/an-error-occurred/32500

A session may be already started by some other platform when Matomo is embedded like this: https://developer.matomo.org/guides/querying-the-reporting-api

### DIFF
--- a/core/Session.php
+++ b/core/Session.php
@@ -174,7 +174,11 @@ class Session extends Zend_Session
 
     public static function close()
     {
-        parent::writeClose();
+        if (self::isSessionStarted()) {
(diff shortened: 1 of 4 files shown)


diosmosis:
Can a system test be added for this? Something like:

- use Platform.initialized in a test case to manually start a session
- send an http request to matomo that would fail w/o these changes

tsteur:
@diosmosis done... without the patch the page would have a fatal error

PR #14899, opened 2019-09-18
Better handling of setting transaction level

tsteur:
FYI @diosmosis maybe you can give it a test if that works for you?

I tried to force it failing with the error you had with a test like this but I couldn't get it failing. Instead of having it configurable I thought it be good to basically detect it automatically.

```php
	public function test_setUncommitted_returnsFalseWhenItFails()
	{
		$this->db->query('SET SESSION binlog_format = \'STATEMENT\';');
		$result = $this->level->setUncommitted();

		// value remains unchanged
		$value = $this->db->fetchOne('SELECT @@TX_ISOLATION');
		$this->assertSame('REPEATABLE-READ', $value);

        $this->assertFalse($result);
	}
```

### DIFF
--- a/core/DataAccess/LogAggregator.php
+++ b/core/DataAccess/LogAggregator.php
@@ -277,11 +277,22 @@ class LogAggregator
             throw $e;
         }
 
+        $transactionLevel = new Db\TransactionLevel($readerDb);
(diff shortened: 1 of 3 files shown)


diosmosis:
Working for me.

PR #14900, opened 2019-09-19
Fixed SQL Exception in UsersManager\API->getUsersPlusRole()

MichaelRoosz:
If `$loginsToLimit` is an empty array, because  `$this->model->getUsersWithAccessToSites($adminIdSites)` returned no results an SQL Exception is triggered:
> Uncaught exception in API: /libs/Zend/Db/Statement/Pdo.php(234): SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ') GROUP BY u.login ORDER BY u.login ASC ' at line 4

This change fixes this problem and sends an empty user list in this case.

Normally $loginsToLimit  will never be empty, however it can happen if access is managed by another plugin, for example the GroupPermissions plugin (https://plugins.matomo.org/GroupPermissions).

### DIFF
--- a/plugins/UsersManager/API.php
+++ b/plugins/UsersManager/API.php
@@ -323,19 +323,27 @@ class API extends \Piwik\Plugin\API
                 $loginsToLimit = $this->model->getUsersWithAccessToSites($adminIdSites);
             }
 
-            list($users, $totalResults) = $this->model->getUsersWithRole($idSite, $limit, $offset, $filter_search, $filter_access, $loginsToLimit);
+            if ($loginsToLimit !== null && empty($loginsToLimit)) {
+                // if the current user is not the superuser, and getUsersWithAccessToSites() returned an empty result,
+                // access is managed by another plugin, and the current user cannot manage any user with UsersManager
+                Common::sendHeader('X-Matomo-Total-Results: 0');
+                return [];
 
-            foreach ($users as &$user) {
-                $user['superuser_access'] = $user['superuser_access'] == 1;
-                if ($user['superuser_access']) {
-                    $user['role'] = 'superuser';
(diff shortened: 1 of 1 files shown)


plugins/UsersManager/API.php
```diff
@@ -323,19 +323,27 @@ public function getUsersPlusRole($idSite, $limit = null, $offset = 0, $filter_se
...
-                    $user['role'] = 'superuser';
-                    $user['capabilities'] = [];
-                } else {
-                    list($user['role'], $user['capabilities']) = $this->getRoleAndCapabilitiesFromAccess($user['access']);
-                    $user['role'] = empty($user['role']) ? 'noaccess' : reset($user['role']);
-                }
+            if ($loginsToLimit !== null && empty($loginsToLimit)) {
+                // if the current user is not the superuser, and getUsersWithAccessToSites() returned an empty result,
+                // access is managed by another plugin, and the current user cannot manage any user with UsersManager
+                Common::sendHeader('X-Matomo-Total-Results: 0');
```
tsteur: @MichaelHeerklotz is the header maybe still there from debugging?

MichaelHeerklotz: It is intentional, some lines above there is another "fast" return case that sets the header, and at the bottom the header is set also for the default case. So i adjusted the new return case to work like the others. I did not work with these custom Matomo before however, so I not 100% sure when exactly to set them and how.

tsteur: Cheers, didn't know :)

tsteur:
Cheers @MichaelHeerklotz

PR #14901, opened 2019-09-19
Fixed #14662 and fixed decimal separator translation

MichaelRoosz:
- Fixed #14662 (non well formed numeric value encountered warning in Sparkline lib)
- Fixed wrong translation key (Intl_NumberSymbolDecimal)

Issue #14662, opened 2019-07-17
“non well formed numeric value encountered” upon upgrading to 3.10 and php7.3


DeamonMV:
**What is my ENV:**

`VM server on Hetzner `
`Matomo version: 3.10.0`
`MySQL version: 5.7.26`
`PHP version: 7.3.6-1+0~20190531112735.39+stretch~1.gbp6131b7`

**What is a problem**

Sometimes i get those warning

```
WARNING: /var/www/matomo/vendor/davaxi/sparkline/src/Sparkline/FormatTrait.php(193): Notice - A non well formed numeric value encountered```

tsteur:
@DeamonMV do you know if that happens when viewing a specific report or sparkline ?

Kervinou:
same issue for me, my new server use MariaDB last version on Debian Linux Buster

MichaelHeerklotz:
I did some research on this.
The problematic line is
`$value = round(($value / $max) * $height);`
$value should be fine because its value comes from some lines above: `$value = (int)$value;`
$height should be fine because it comes from getInnerNormalizedHeight() which does `return $this->getInnerHeight() * $this->ratioComputing;`, forcing it to be numerical
this leaves us with $max which  comes from getMaxValue() with either returns $this->base which we never set or max($this->data);
So the offending value must be the max value $this->data which we set in Piwik\Visualization\Sparkline. 

Thus, this warning most likely can be avoided by ensuring we only pass numerical values in the $sparkline->setData($values); call in  Piwik\Visualization\Sparkline.

I will do some more research and come back if I can track it down some more.

MichaelHeerklotz:
I added some debug code to my setup:
`            if (!is_numeric($value)) {
                file_put_contents("debug.log.txt", var_export($value, true) . PHP_EOL, FILE_APPEND);
            }`

It found these values:
`'78 '
'79,5 '
'81,63 '
'80,88 '
'84,05 '
'84,04 '
'81,37 '
'82,16 '
'75,38 '
'79,91 '
'82,33 '
'81,88 '
'83,38 '
`
The problem with these values is the space character at the end which makes them non-numerical for php.
To resolve the problem a simple trim() should do the trick.

While debugging I also found this code:
`$thousandSeparator = Piwik::translate('Intl_NumberSymbolGroup');
$decimalSeparator = Piwik::translate('Intl_NumberSymbolGroup');
$value = str_replace($thousandSeparator, '', $value);
[...]

PR #14908, opened 2019-09-24
Sparkline: Fix value sanitizing 

MichaelRoosz:
My last pull request (https://github.com/matomo-org/matomo/pull/14901) tried to fix the problem of values with spaces at the end in the Sparkline class (making the values "non-numeric").

While doing some more debugging I also encountered "non-breaking" spaces in some values, which the trim() function (my previous fix) does not remove.

These special spaces are hard to remove, because they sometimes are encoded in ISO/ASCII (hex 0xA0) as well as UTF-8 (hex 0xC2 0xA0). Without making use of the mb_* functions it will be hard to remove all of them while not damaging utf-8 strings.

This PR tries to solve the problem by making use of filter_var() to sanitize the value and remove all non-numeric characters.

See also https://github.com/matomo-org/matomo/issues/14662#issuecomment-534505048

### DIFF
--- a/core/Visualization/Sparkline.php
+++ b/core/Visualization/Sparkline.php
@@ -53,33 +53,29 @@ class Sparkline implements ViewInterface
 
         $sparkline = new \Davaxi\Sparkline();
 
-        $seconds = Piwik::translate('Intl_NSecondsShort');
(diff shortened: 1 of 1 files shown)


tsteur:
Cheers @MichaelHeerklotz very appreciated

PR #14913, opened 2019-09-26
Add a configurable row limit to report export popover

katebutler:


tsteur:
@katebutler just tested and it looks when no max filter limit is defined, I can no longer enter any value like 500 or something? Can you try that? Otherwise works 👍

PR #14920, opened 2019-09-30
Fix transaction level capability check

diosmosis:


core/Db/TransactionLevel.php
```diff
@@ -37,12 +40,15 @@ public function canLikelySetTransactionLevel()
 
     public function setUncommitted()
     {
+        $backup = $this->db->fetchOne('SELECT @@TX_ISOLATION');
```
tsteur: @diosmosis does this need to be `$this->statusBackup` instead of `$backup`?

diosmosis: That's done below after successfully changing it. If it fails for some reason to set it then we don't need to set it back.

tsteur: makes sense 👍

tsteur:
@diosmosis fyi I'm now getting an error 

`SQLSTATE[HY000]: General error: 1193 Unknown system variable 'TX_ISOLATION' - in plugin Goals` see https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-3.html

with MySQL8. Will investigate and create PR.

PR #14921, opened 2019-09-30
HTTP method needs to return true when a destination path is given

tsteur:
In https://github.com/matomo-org/matomo/pull/14877 I added a new feature so plugins can hook into HTTP. Now noticed the GeoIP download was not working because it didn't return true when the response was loaded into a file.

### DIFF
--- a/core/Http.php
+++ b/core/Http.php
@@ -244,16 +244,19 @@ class Http
 		     * described below
 		     * @ignore
 		     */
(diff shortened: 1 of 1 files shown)


sgiehl (commented):
Haven't tested, but code looks fine. Should we maybe add a test to check that true is returned in that case, to avoid future failures when the method is changed again?

PR #14929, opened 2019-10-01
Removed closing PHP tag due to PSR-2 definition

Crease29:
While reading your Matomos contributing guide, I saw, that Matomo is sticking to PSR-1, PSR-2 and PSR-4. I digged into the code and saw some closing PHP tags, which must be omitted (see: https://www.php-fig.org/psr/psr-2/#22-files).

There are also closing PHP tags in `libs/HTML` but I wasn't sure, if these should be touched.

### DIFF
--- a/plugins/API/config/config.php
+++ b/plugins/API/config/config.php
@@ -1,3 +1,2 @@
 <?php
 return array();
-?>
--- a/plugins/API/config/tracker.php
+++ b/plugins/API/config/tracker.php
@@ -1,3 +1,2 @@
 <?php 
 return array(); 
-?>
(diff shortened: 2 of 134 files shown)


tsteur:
@Crease29 thanks for this. You updated the custom alerts submodule which makes the tests fail see https://travis-ci.org/matomo-org/matomo/jobs/592136762#L533 could you revert the submodule change?

Crease29:
Sorry for the amount of pushes. I messed something up here when reverting the previous commit. Should be fine now 🙂

tsteur:
Cheers @Crease29 for fixing the PR will merge now. 

Sure we appreciate such PRs as long as they don't change any logic and are easy to review and quick to merge before any merge conflicts happen :)

PR #14936, opened 2019-10-01
Show all sites item only if multi sites is enabled

tsteur:
To prevent showing a not working link to the all sites page.

There are a few more usages of siteselector but they aren't as important and visible for now. Also sometimes I think they don't link to the all sites page but handle it differently.

### DIFF
--- a/core/Plugin/Controller.php
+++ b/core/Plugin/Controller.php
@@ -751,6 +751,7 @@ abstract class Controller
 
         $pluginManager = Plugin\Manager::getInstance();
         $view->relativePluginWebDirs = (object) $pluginManager->getWebRootDirectoriesForCustomPluginDirs();
+        $view->isMultiSitesEnabled = Manager::getInstance()->isPluginActivated('MultiSites');
 
         self::setHostValidationVariablesView($view);
     }
--- a/plugins/CoreHome/templates/_siteSelectHeader.twig
+++ b/plugins/CoreHome/templates/_siteSelectHeader.twig
@@ -1,3 +1,3 @@
 <div class="top_bar_sites_selector piwikTopControl">
-    <div piwik-siteselector show-selected-site="true" show-all-sites-item="{{ (isMultiSitesEnabled|default) ? 'true' : 'false' }}" class="sites_autocomplete"></div>
+    <div piwik-siteselector show-selected-site="true" show-all-sites-item="{{ isMultiSitesEnabled ? 'true' : 'false' }}" class="sites_autocomplete"></div>
 </div>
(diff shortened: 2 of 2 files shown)


kniziol:
Is the issue fixed or still available to work on it?

tsteur:
The issue should be fixed afterwards.

core/Plugin/Controller.php
```diff
@@ -751,6 +751,7 @@ protected function setBasicVariablesNoneAdminView($view)
 
         $pluginManager = Plugin\Manager::getInstance();
         $view->relativePluginWebDirs = (object) $pluginManager->getWebRootDirectoriesForCustomPluginDirs();
+        $view->isMultiSitesEnabled = Manager::getInstance()->isPluginActivated('MultiSites');
```
kniziol: Make sure you've got a boolean value:
```php
$view->isMultiSitesEnabled = (bool)Manager::getInstance()->isPluginActivated('MultiSites');
```

plugins/CoreHome/templates/_siteSelectHeader.twig
```diff
@@ -1,3 +1,3 @@
 <div class="top_bar_sites_selector piwikTopControl">
-    <div piwik-siteselector show-selected-site="true" class="sites_autocomplete"></div>
+    <div piwik-siteselector show-selected-site="true" show-all-sites-item="{{ (isMultiSitesEnabled|default) ? 'true' : 'false' }}" class="sites_autocomplete"></div>
```
kniziol: ...and the [default filter](https://twig.symfony.com/doc/2.x/filters/default.html) won't be necessary:
```twig
<div
    class="sites_autocomplete"
    piwik-siteselector show-selected-site="true"
    show-all-sites-item="{{ isMultiSitesEnabled ? 'true' : 'false' }}">
</div>
```

kniziol:
@tsteur I've reviewed your 539e39 commit

core/Plugin/Controller.php
tsteur: btw the method always returns a bool

plugins/CoreHome/templates/_siteSelectHeader.twig
tsteur: I had the `default` in there as I couldn't guarantee the `basic variables` method in the controller is always called so just wanted to be sure to not break anything. I could remove it as it's likely not needed. It just means if someone did like `$view = new View('...'); $view->render()` without calling the basic variables in the controller then things could break when embedding that template. From what I see it shouldn't be needed though. Will check again.

tsteur: From what I can see it can be removed indeed cheers 👍  will wait for the tests to confirm it

core/Plugin/Controller.php
kniziol: OKay. So the `(bool)` is not necessary.

plugins/CoreHome/templates/_siteSelectHeader.twig
kniziol: Great! :)

kniziol (approved):

PR #14937, opened 2019-10-02
Do not proceed with archiving if a valid archive exists for the parameter combination.

### LINKED ISSUE #13387: Archiver should skip creating a new "temporary" archive when there was no new visit/traffic since the last "temporary archive"

diosmosis:
Fixes #13387


diosmosis:
@mattab / @tsteur seems to work, not sure about all the edge cases in archiving though

core/CronArchive.php
```diff
@@ -462,7 +465,7 @@ public function run()
                     try {
                         if ($this->isWebsiteUsingTheTracker($idSite)) {
 
-                            if(!$this->hadWebsiteTrafficSinceMidnightInTimezone($idSite)) {
+                            if (!$this->hadWebsiteTrafficSinceLatestArchive($idSite)) {
```
tsteur: @diosmosis I don't think we can do this check. Eg just because an archiver finished last time at eg `2019-10-02 10:00:00` and there has been no visit since then, doesn't mean we shouldn't archive. 

* Eg the site has 20 segments or even when no segment then eg the day archive might have finished at `2019-10-02 09:50:00` and there could have been visits since then.
* Also we sometimes skip some archives when another archiver is currently processing the same site... like archiver launches and archives idsite=1&period=day... then another archiver launches and skips all `idsite=1` archivers because it's waiting for `period=day` to finish first. At the end the `archiving finished ts` will be updated though. I reckon the `OPTION_ARCHIVING_FINISHED_TS` is something we could almost remove as it the time is pretty much useless in terms of anything. 

Note: I may be understanding things wrong though as we know what a complicated thing it is. In general the only thing we know though is when actually requesting the archive for a specific `idsite, period, date, segment`... then we can look when this particular archive was last archived (the time the archiving started), and then we need to look if there was a visit since the archiving started

core/CronArchive.php
```diff
@@ -874,53 +877,61 @@ protected function processArchiveDays($idSite, $lastTimestampWebsiteProcessedDay
...
-        if (empty($content)
-            || !is_array($daysResponse)
-            || count($daysResponse) == 0
-        ) {
-            // cancel marking the site as reprocessed
-            if ($websiteInvalidatedShouldReprocess) {
-                $store = new SitesToReprocessDistributedList();
-                $store->add($idSite);
-            }
+        list($visitsToday, $isThereArchive) = $this->isThereAValidArchiveForPeriod($idSite, 'day', $segment = '');
```
tsteur: @diosmosis wondering... in the method `isThereAValidArchiveForPeriod` do we always need to call `invalidateArchivedReportsForSitesThatNeedToBeArchivedAgain()` first in case there is a race condition? Like currently the archiver starts and calls `invalidateArchivedReportsForSitesThatNeedToBeArchivedAgain()`... the archiving may take many seconds, minutes, hours... by the time we get here there might be more `getRememberedArchivedReportsThatShouldBeInvalidated` because there were visits since the archiver started.

core/CronArchive.php
```diff
@@ -996,15 +1020,22 @@ private function archiveReportsFor($idSite, $period, $date, $archiveSegments, Ti
...
-            $request = new Request($url);
-            $request->before(function () use ($self, $url, $idSite, $period, $date) {
-                if ($self->isAlreadyArchivingUrl($url, $idSite, $period, $date)) {
-                     return Request::ABORT;
-                }
-            });
-            $urls[] = $request;
-            $this->logArchiveWebsite($idSite, $period, $date);
+            list($visits, $isThereArchive) = $this->isThereAValidArchiveForPeriod($idSite, $period, $segment);
+            if ($isThereArchive) {
```
tsteur: should it be `! $isThereArchive `?

tsteur (changes requested):

core/CronArchive.php
diosmosis: 👍 yes, makes sense

core/CronArchive.php
diosmosis: I'm not 100% sure why I put this here. Seems like the code should work w/o it, will remove.

diosmosis: I see, I was just renaming the method. There was no functional change.

core/CronArchive.php
diosmosis: 👍

diosmosis:
Tested locally, seems to work. Ready for another review.

core/CronArchive.php
```diff
@@ -1815,9 +1858,16 @@ private function getUrlsWithSegment($idSite, $period, $date)
                 continue;
             }
 
+            list($visits, $isThereArchive) = $this->isThereAValidArchiveForPeriod($idSite, $period, $segment);
```
tsteur: @diosmosis I wonder if we should move this into line `1907` after 

```php

                if ($self->isAlreadyArchivingSegment($urlWithSegment, $idSite, $period, $segment)) {
                    return Request::ABORT;
                }
```
and then make it like
```php
list($visits, $isThereArchive) = $this->isThereAValidArchiveForPeriod($idSite, $period, $segment);
if ($isThereArchive) {
   return Request::ABORT;
}
```

Reason being there might be 20 segments for example, each segment could take an hour to archive, and by the time it actually executes a segment it is invalidated. So we'd need to check it really only shortly before executing a request.

core/CronArchive.php
```diff
@@ -1008,15 +1034,20 @@ private function archiveReportsFor($idSite, $period, $date, $archiveSegments, Ti
...
-            $request = new Request($url);
-            $request->before(function () use ($self, $url, $idSite, $period, $date) {
-                if ($self->isAlreadyArchivingUrl($url, $idSite, $period, $date)) {
-                     return Request::ABORT;
-                }
-            });
-            $urls[] = $request;
-            $this->logArchiveWebsite($idSite, $period, $date);
+            list($visits, $isThereArchive) = $this->isThereAValidArchiveForPeriod($idSite, $period, $segment);
+            if (!$isThereArchive) {
```
tsteur: wondering here as well similar to https://github.com/matomo-org/matomo/pull/14937/files?w=1#diff-c94ca2937035751f5e1ea942c529495dR1861 should we maybe move this into `1042` after ?
```php
 if ($self->isAlreadyArchivingUrl($url, $idSite, $period, $date)) {
                        return Request::ABORT;	                        return Request::ABORT;
                    }	                    }
```

tsteur (changes requested):
@diosmosis looks good in general, would just move some code slightly. I like how easy the logic is :)

core/DataAccess/ArchiveSelector.php
```diff
@@ -67,6 +67,9 @@ public static function getArchiveIdAndVisits(ArchiveProcessor\Parameters $params
 
         $doneFlags      = Rules::getDoneFlags($plugins, $segment);
         $doneFlagValues = Rules::getSelectableDoneFlagValues();
+        if (!$includeInvalidated) {
```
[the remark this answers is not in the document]
diosmosis: Yes, I also noticed I didn't actually use this parameter in CronArchive :).

core/CronArchive.php
diosmosis: 👍

core/CronArchive.php
diosmosis: 👍

diosmosis:
Applied more review feedback.

core/CronArchive.php
```diff
@@ -946,6 +956,21 @@ protected function processArchiveDays($idSite, $lastTimestampWebsiteProcessedDay
         return $dayArchiveWasSuccessful;
     }
 
+    private function isThereAValidArchiveForPeriod($idSite, $period, $segment = '')
```
tsteur: @diosmosis wondering: Wouldn't we need to take into consideration the date as well we want to archive? Only realising this now... like what if a user is tracking visits a few days back and this archive was invalidated but not today... then we would need to detect this right? But then I'm wondering if things get in the end slower than we benefit from it. @mattab 

There are a few complexities:
* Date might be `last2` or `last52` for example... if it was say last52 does this maybe mean we need to iterate over 52 dates/archives and check if any of them is invalidated meaning doing below logic max 52 times? (also adding lots of IO)
* If date is `last2`, but there are reports to be invalidated in the past in `invalidateArchivedReportsForSitesThatNeedToBeArchivedAgain()`, does this mean we need to dynamically need to change the date that will be used in the query to say `lastXYZ` to make sure the invalidated reports will be rearchived? 

The original purpose was to make archiving faster when you have say 1000 sites and only say 200 of those sites are tracking regularly. Meaning for 800 sites we could mostly skip launching the archiver at all. But now wondering if in the end we make all other cases slower? In general it seems like the archiver should maybe behave like this though. Goes maybe also a bit in the direction of https://github.com/matomo-org/matomo/issues/15117

tsteur: FYI also created https://github.com/matomo-org/matomo/issues/15127

diosmosis: I think we could potentially use multiple dates w/ one query (`SELECT COUNT(*) WHERE date1 >= X AND date2 <= Y and period = Z AND ... == $rangePperiod->getSubperiodCount()`). Think that would work?

tsteur: I reckon we'd still need to go over several tables though and we'd need to know which data is stored where etc. So we could probably do it in a few queries for example.

diosmosis: Just noticed another issue, this doesn't affect the original case where 800 sites w/o visits shouldn't have archiving launch, since there won't be any archives for those sites even though they have no visits. I guess we should also check for a single visit in a period for a site? Maybe only for the non-segment requests, then if there are visits for those we'd know there could be visits for segments.

diosmosis: @tsteur  for the no visits optimization I wonder if we can just change `hadWebsiteTrafficSinceMidnightInTimezone()` to check whether there has been a visit for the site since the last time archiving was run for the site? We would save the archive start time in a new option like `CronArchive.startArchiveSite.$idSite` in `archiveSingleSite()`.

tsteur: @diosmosis not 100% sure. We would probably need to fix first https://github.com/matomo-org/matomo/issues/15127

You mean then add an empty archive when no visits for today so far basically?

diosmosis: I see, yes, I guess it can't work w/o #15127. I guess we can add and look for a `'start'` flag that's inserted when an archive starts archiving (in ArchiveWriter), as was done in a previous version of this PR then check for a single visit since then.

diosmosis: Actually I think we don't have to worry about specifically looking for no visits... if a site has a visit the archive will be invalidated, otherwise there might be an archiving launch for the first time in a day, but after that there will be an archive... Ok, think this is ready for another review.

core/CronArchive.php
```diff
@@ -1851,12 +1912,18 @@ private function getUrlsWithSegment($idSite, $period, $date)
...
             $logger = $this->logger;
             $self = $this;
-            $request->before(function () use ($logger, $segment, $segmentCount, &$processedSegmentCount, $idSite, $period, $urlWithSegment, $self) {
+            $request->before(function () use ($logger, $segment, $segmentCount, &$processedSegmentCount, $idSite, $period, $date, $urlWithSegment, $self) {
 
                 if ($self->isAlreadyArchivingSegment($urlWithSegment, $idSite, $period, $segment)) {
                     return Request::ABORT;
                 }
 
+                $isThereArchive = $this->isThereAValidArchiveForPeriod($idSite, $period, $date, $segment);
```
tsteur: @diosmosis probably not needed as part of this PR... do you reckon, we could dynamically change the lastN date that is applied in `$request->url` if something was invalidated? Like maybe `isThereAValidArchiveForPeriod ()` could return the `minInvalidatedArchiveDate`, based on this information we create the `lastN` string and then do something like `$request->url = str_replace(date=$date, date=newLastN, $request->url)`? Not sure it's clear what I mean? Just meaning in case we know we invalidated an older archive then we could adjust the date right away?

diosmosis: Yes, I think it's possible, will look into it.

diosmosis: Note: this code does the lastN queries on archive tables we were discussing on slack.

diosmosis: @tsteur made the change, it will change the lastN to be `min(oldestInvalidatedDateDaysSinceToday, originalLastN)` (so it will still respect the max dates in CronArchive).

core/CronArchive.php
```diff
@@ -1010,13 +1088,24 @@ private function archiveReportsFor($idSite, $period, $date, $archiveSegments, Ti
...
+                    return Request::ABORT;
+                }
+
+                list($isThereArchive, $newDate) = $this->isThereAValidArchiveForPeriod($idSite, $period, $date, $segment);
+                if ($isThereArchive) {
+                    $this->logArchiveWebsiteSkippedValidArchiveExists($idSite, $period, $date);
+                    return Request::ABORT;
                 }
+
+                $url = $request->getUrl();
```
tsteur: @diosmosis any chance to add a test for this? could maybe move this somehow into the request class and then call like 

```php
$request->changeDate($newDate)
```

and this changeDate method could be then public and testable? with regexp I always get nervous so be good to have a basic test for it. Otherwise there doesn't seem anything missing anymore?

diosmosis: Moved and added a test

tsteur:
Worked nicely for me @diosmosis . The only thing I would maybe change is to maybe instead of using `last1` using at least `last2` always or could maybe always add `+1 period`? not needed maybe though but could prevent some random edge cases and be maybe safer re timezone or other hard to find issues. Might not be needed though as it would maybe only hide a problem. Up to you. 

We'll merge this after the next 3.13. release just to give it more time to test during a beta.

diosmosis:
I'll add this to 3.13.1

tsteur:
@diosmosis seems there are 2 more integration tests failing although not sure if the 2nd test is maybe fixed by now as it was maybe caused by some other test? https://travis-ci.org/matomo-org/matomo/jobs/616834354#L826-L879 . 

If you have time, could you update the test and then merge?

tsteur:
@diosmosis when you have some time could you have a look at this?

diosmosis:
@tsteur yes, forgot about this

PR #14941, opened 2019-10-02
Fix XML parser warning handling in RssWidget

### LINKED ISSUE #14814: Handle broken RSS feed more gracefully
see also https://github.com/matomo-org/matomo/issues/13062
As the Matomo RSS feed is currently down (https://github.com/matomo-org/matomo/issues/14813), Matomo instances show the following error:
```
plugins/RssWidget/RssRenderer.php(59): Warning - simplexml_load_string(): Entity: line 2: parser error : Extra content at the end of the document - Matomo 3.11.0 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)
```

I guess instead of a non-userfriendly xml error we should show information that fetching the RSS feed didn't work.

yurihs:
Hi! This should fix #14814.

I had to temporarily set an error handler to promote the warnings generated by `simplexml_load_string()` to exceptions. That way, they're handled gracefully.

Please let me know if this can be improved.

### DIFF
--- a/plugins/RssWidget/RssRenderer.php
+++ b/plugins/RssWidget/RssRenderer.php
@@ -56,7 +56,21 @@ class RssRenderer
         if (!$output) {
             try {
                 $content = Http::fetchRemoteFile($this->url);
+
+                $promoteWarningToException = function ($errno, $errstr, $errfile, $errline, $errcontext) {
+                    // if the error has been suppressed by the @ we don't handle the error
+                    if (error_reporting() == 0) {
+                        return false;
+                    }
+                    if ($errno !== E_WARNING) {
+                        return false;
+                    }
+                    throw new \Exception($errstr, $errno);
+                };
+                set_error_handler($promoteWarningToException, E_WARNING);
                 $rss = simplexml_load_string($content);
+                restore_error_handler();
+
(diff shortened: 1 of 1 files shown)


plugins/RssWidget/RssRenderer.php
```diff
@@ -56,7 +56,21 @@ public function get()
...
+                    // if the error has been suppressed by the @ we don't handle the error
+                    if (error_reporting() == 0) {
+                        return false;
+                    }
+                    if ($errno !== E_WARNING) {
+                        return false;
+                    }
+                    throw new \Exception($errstr, $errno);
+                };
+                set_error_handler($promoteWarningToException, E_WARNING);
```
tsteur: this would be overwriting our error handler? 

Can we maybe just do something like `$rss = @simplexml_load_string($content)`
and trigger an exception if $rss is empty?

yurihs: I restored it right afterwards with `restore_error_handler()`. But yeah, I guess just supressing it would be simpler. I'll do that.

tsteur:
Cheers @yurihs  very appreciated 👍

PR #14942, opened 2019-10-02
Skip unnecessary validation on report unsubscribe

### LINKED ISSUE #14925: Unsubscribing from email report leads to error
Unsubscribe does not work, here are the steps to reproduce:

* receive an email report from Matomo Cloud 
* click the Unsubscribe link
* Then confirm the unsubscription
* Got: the next page shows an error message "Report FormAnalytics_get is unknown or not available for report type 'email'."
* Expected to see a success message instead

katebutler:
Fixes #14925, which was caused by the subscription containing a report from a plugin which had been deactivated.

### DIFF
--- a/plugins/ScheduledReports/API.php
+++ b/plugins/ScheduledReports/API.php
@@ -194,6 +194,36 @@ class API extends \Piwik\Plugin\API
     }
 
     /**
+     * Updates an existing report.
+     *
+     * @see addReport()
+     */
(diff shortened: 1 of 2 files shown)


tsteur:
@katebutler some tests are failing see https://travis-ci.org/matomo-org/matomo/jobs/593226856#L907

It seems it's not clearing the scheduled reports cache in `API::public static $cache `

PR #14949, opened 2019-10-03
update icons submodule

Findus23:
As it seems like #14860 reverted the matomo-icons submodule, I'll update it to the latest version for the 3.12 release.


tsteur:
Cheers @Findus23 might just need to update the tests?

Findus23:
@tsteur Should be fixed, but there are a few broken UI tests left.

PR #14952, opened 2019-10-03
Remove duplicate decodeURIComponent call in segmentcontroller

### LINKED ISSUE #14943: Segmenteditor fails
While testing https://github.com/matomo-org/matomo/pull/14884 I noticed an error in the segment editor. Likely it's a regression.

It fails with the error

```
angular.min.js?cb=21177de8310fa51319dd14c492d0bdea:123 URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at findAndExplodeByMatch (segmentgenerator.controller.js?cb=21177de8310fa51319dd14c492d0bdea:50)
    at SegmentGeneratorController.setSegmentString (segmentgenerator.controller.js?cb=21177de8310fa51319dd14c492d0bdea:262)
    at SegmentGeneratorController (segmentgenerator.controller.js?cb=21177de8310fa51319dd14c492d0bdea:273)
    at Object.invoke (angular.min.js?cb=21177de8310fa51319dd14c492d0bdea:44)
    at O.instance (angular.min.js?cb=21177de8310fa51319dd14c492d0bdea:94)
    at q (angular.min.js?cb=21177de8310fa51319dd14c492d0bdea:69)
    at angular.min.js?cb=21177de8310fa51319dd14c492d0bdea:81
    at angular.min.js?cb=21177de8310fa51319dd14c492d0bdea:136
    at m.$digest (angular.min.js?cb=21177de8310fa51319dd14c492d0bdea:147)
```
in `plugins/SegmentEditor/angularjs/segment-generator/segmentgenerator.controller.js` around line 50 where the value of `newMetric` is `{"segment":"browserCode","matches":"==","value":"<'>/\"#$I*%"}`

image

In the DB I have the segment defined like this 

```sql
INSERT INTO `piwik_segment` (`idsegment`, `name`, `definition`, `login`, `enable_all_users`, `enable_only_idsite`, `auto_archive`, `ts_created`, `ts_last_edit`, `deleted`)
VALUES
	(28, '&lt;&#039;&gt;/&quot;#$I*%', 'browserCode==%3C%27%3E%2F%22%23%24I*%25', 'root', 0, 1, 0, '2016-09-29 10:36:08', '2016-09-29 11:36:22', 0);
```

I suppose the problem is that the value might be decoded twice? not sure. When I debug it does the decode twice though (`if (minPos < metric.length) {`)

image

katebutler:
Fixes #14943
This line is not needed as the `decodeURIComponent` is already done whenever `newmetric.value` is set.

### DIFF
--- a/plugins/SegmentEditor/angularjs/segment-generator/segmentgenerator.controller.js
+++ b/plugins/SegmentEditor/angularjs/segment-generator/segmentgenerator.controller.js
@@ -47,7 +47,6 @@
             }
         }
 
-        newMetric.value = decodeURIComponent(newMetric.value);
         return newMetric;
     };
 
--- a/plugins/SegmentEditor/angularjs/segment-generator/segmentgenerator.controller.js
+++ b/plugins/SegmentEditor/angularjs/segment-generator/segmentgenerator.controller.js
@@ -47,6 +47,13 @@
             }
         }
 
+        try {
+            // Decode again to deal with double-encoded segments in database
+            newMetric.value = decodeURIComponent(newMetric.value);
+        } catch (e) {
+            // Expected if the segment was not double-encoded
+        }
+
         return newMetric;
     };
 


tsteur:
@katebutler it seems to work for my case but now failing for this case: https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/36240/SegmentSelectorEditorTest_saved_details.png

I wonder if that's because some segments might be double encoded while some might not or so... 

I reckon we could just try to decode it there, and if it fails return the regular version... like

```js
try {
return decode...(newMetric.value)
} catch (e){
return newMetric.value
}
```

katebutler:
@tsteur Have added a second `decodeUIRComponent` in a try/catch as suggested, build passes now.

PR #14961, opened 2019-10-06
Log stack traces for PHP warnings etc

katebutler:
Includes the full stack in the log message when writing to file only.  The log message for other handlers (web notifications and database) will not be changed.

### DIFF
--- a/core/ErrorHandler.php
+++ b/core/ErrorHandler.php
@@ -161,10 +161,22 @@ class ErrorHandler
             case E_DEPRECATED:
             case E_USER_DEPRECATED:
             default:
+                // Need to create and throw an Exception so that we can get the stack trace
+                $e = new \Exception();
+                $message = self::getHtmlMessage($errno, $errstr, $errfile, $errline, $e->getTraceAsString());
                 try {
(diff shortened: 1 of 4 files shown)


tsteur:
BTW @katebutler there are some failing tests https://travis-ci.org/matomo-org/matomo/jobs/594801494 and https://travis-ci.org/matomo-org/matomo/jobs/594801495#L818

and as discussed be good to remove the newline.

tsteur:
@katebutler I'm thinking this failing test might be due to this change? https://travis-ci.org/matomo-org/matomo/jobs/594907204#L817

core/ErrorHandler.php
```diff
@@ -161,8 +161,12 @@ public static function errorHandler($errno, $errstr, $errfile, $errline)
             case E_DEPRECATED:
             case E_USER_DEPRECATED:
             default:
+                $context = array('trace' => debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS));
```
tsteur: @katebutler could we add a limit here of say max 15 last traces? There is a second parameter `$limit` that can be set.

PR #14962, opened 2019-10-06
Remove JSON2 from Admin UI

tsteur:
All browsers that we support should have JSON support. The file was added > 7 years ago when not all browsers had it yet.


tsteur:
Test failures don't seem related to this

PR #10024, opened 2016-04-08
Live API when a Segment is used: Select 10 times more rows so that resultset after grouping is more likely to contain enough rows

mattab:
refs https://github.com/piwik/piwik/issues/9200

If the build passes, this may be worth to merge to help with #9200

PR #14963, opened 2019-10-07
Fix mismatched visitors (Overview / Visitor Log) when segment is applied

### LINKED ISSUE #13861: Mismatched visitors (Overview / Visitor Log) when segment is applied
I have created a basic segment to show this unexpected behavior.

When I apply the segment as shown in above image, I expect to see 5 visits in Visitor Log. But, when the segment is applied, there are only 3 visitors in Visitor Log as shown in the image below.

I have observed this behavior in more complex segments, but I can reproduce it in basic segments when I use the options Action URL or Page URL.

I'm doing something wrong or this is really a bug!? I'm using Matomo 3.7.0, but this occurs in release 3.5.1 too.

Best regards.

tsteur:
fix https://github.com/matomo-org/matomo/issues/13861 

also refs https://github.com/matomo-org/matomo/issues/9200 and now we no longer need workaround in https://github.com/matomo-org/matomo/pull/10024

Tried various different solutions until I realised we can simplify this actually all quite a lot by no longer using any subselects at all and likely also making regular visitor logs without segment a bit faster with some logic I added recently.

Now we basically have no more subqueries in the visitor log (similar to archiving yay). Instead we apply the group by, order, and limit directly. More on this in the comments...

You can see in the tests how this simplifies queries. When segments were applied we had before 3 subqueries (or one query with 2 nested inner queries), and when no segment was applied we had 2 subqueries. Now there's always only one query. It also removes a group by when no segment is applied.


core/Segment.php
```diff
@@ -368,8 +370,23 @@ public function getSelectQuery($select, $from, $where = false, $bind = array(),
             $limitAndOffset = (int) $offset . ', ' . (int) $limit;
         }
 
-        return $this->segmentQueryBuilder->getSelectQueryString($segmentExpression, $select, $from, $where, $bind,
-            $groupBy, $orderBy, $limitAndOffset);
+        try {
+            if ($forceGroupBy && $groupBy) {
+                $this->segmentQueryBuilder->forceInnerGroupBySubselect(LogQueryBuilder::FORCE_INNER_GROUP_BY_NO_SUBSELECT);
```
tsteur: this was a feature I added recently... as we're using a `group by idvisit` we basically don't ever need to use a subselect...

plugins/Live/Model.php
```diff
@@ -469,25 +469,27 @@ public function makeLogVisitsQueryString($idSite, $startDate, $endDate, $segment
...
         }
 
-        $orderBy .= "visit_last_action_time " . $filterSortOrder;
-        $orderByParent = "sub.visit_last_action_time " . $filterSortOrder;
+        $orderBy .= "log_visit.visit_last_action_time " . $filterSortOrder;
 
-        // this $innerLimit is a workaround (see https://github.com/piwik/piwik/issues/9200#issuecomment-183641293)
-        $innerLimit = $limit;
-        if (!$segment->isEmpty()) {
-            $innerLimit = $limit * 10;
```
tsteur: we no longer need that workaround since things should work correctly now

plugins/Live/Model.php
```diff
@@ -496,21 +498,15 @@ public function makeLogVisitsQueryString($idSite, $startDate, $endDate, $segment
             // for now let's not apply when looking for a specific visitor
             $maxExecutionTimeHint = '';
         }
-
-        // Group by idvisit so that a given visit appears only once, useful when for example:
-        // 1) when a visitor converts 2 goals
-        // 2) when an Action Segment is used, the inner query will return one row per action, but we want one row per visit
-        $sql = "
```
tsteur: This subquery was actually never needed when no segment was applied yet we had it there... and instead of having the group by in the outer query, we now have only one query and the group by applied there.

plugins/Live/tests/Integration/ModelTest.php
```diff
@@ -197,20 +197,13 @@ public function test_makeLogVisitsQueryString()
                 $minTimestamp = false,
                 $filterSortOrder = false
         );
-        $expectedSql = ' SELECT  sub.* FROM
-                (
-                    SELECT log_visit.*
+        $expectedSql = ' SELECT log_visit.*
```
tsteur: that's how it simplifies the queries when no segment is applied. No more subquery, and we don't even need a group by anymore.

plugins/Live/tests/Integration/ModelTest.php
```diff
@@ -319,29 +298,18 @@ public function test_makeLogVisitsQueryString_whenSegment()
...
-                        ORDER BY idsite DESC, visit_last_action_time DESC
-                        LIMIT 10, 1000
-                        ) AS log_inner
-                    ORDER BY idsite DESC, visit_last_action_time DESC
-                 ) AS sub
-                 GROUP BY sub.idvisit
-                 ORDER BY sub.visit_last_action_time DESC
-                 LIMIT 100
-        ';
+        $expectedSql = ' SELECT log_visit.*
```
tsteur: Here you can see how it simplifies those 3 queries into one.

plugins/Live/tests/System/ApiTest.php
```diff
@@ -78,6 +78,19 @@ public function getApiForTesting()
             )
         );
 
+        $apiToTest[] = array(array('Live.getLastVisitsDetails'),
```
tsteur: added this test to ensure it works besides the already existing tests that generates the SQL where we can see in the SQL query that it will now work

tsteur:
@sgiehl or @mattab if someone could have a review that be great? @mattab be good you can confirm it should be fine from a performance perspective. From what I see it should be fine... or there is at least no other way to have segmenting fully work with the visitor log. In regards to your comment in https://github.com/matomo-org/matomo/issues/13861#issuecomment-457159967 we basically only added this to the visitor log and only when segmenting. Considering we remove a few subselects it might not even be any slower or even faster?

mattab:
From performance perspective it should be fine :+1: 
Haven't reviewed the exact code in detail, but overall the SQL change looks great and I can't think of a way this could create a problem.

PR #14975, opened 2019-10-08
Allow noreply email address/from to be set in admin panel

### LINKED ISSUE #12426: Mail from must equal authorized user

katebutler:
Fixes #12426


tsteur:
@katebutler it's not working for me. I think the problem is that you're trying to set a config value directly which doesn't work

>  Config::getInstance()->General['noreply_email_name'] =  Common::unsanitizeInputValue(Common::getRequestVar('mailFromName', ''));

instead something like this:
```php
$general = Config::getInstance()->General
$general['noreply_email_name'] = Common::unsanitizeInputValue(Common::getRequestVar('mailFromName', ''));
...
Config::getInstance()->General = $general;

```

tsteur:
fyi @katebutler merging now... works now... made a minor tweak that when no address is configured it will use the default value as otherwise sending mails likely fails and making sure to validate the configured address that it looks like an email etc.

PR #14979, opened 2019-10-09
Avoid warning headers already sent when setting session cache limiter

tsteur:
Not sure this could be causing any issues?

### DIFF
--- a/core/bootstrap.php
+++ b/core/bootstrap.php
@@ -29,7 +29,7 @@ if (!defined('PIWIK_VENDOR_PATH')) {
 // NOTE: the code above must be PHP 4 compatible
 require_once PIWIK_INCLUDE_PATH . '/core/testMinimumPhpVersion.php';
 
-if (session_status() !== PHP_SESSION_ACTIVE) {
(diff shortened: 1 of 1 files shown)


tsteur:
@sgiehl or @mattab any thoughts on this? Was reading https://www.php.net/manual/en/function.session-cache-limiter.php

I understand it won't send eg the `Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0 .... Pragma: no-cache ...` if headers were already sent. We don't really want this, but when this happens, we can't change things anymore anyway. The default is `nocache`  see https://www.php.net/manual/en/session.configuration.php#ini.session.cache-limiter so it should all be fine even if we can't set it specifically. It would only be an issue if the script is eg embedded in WP-Matomo.

mattab:
Code change looks fine. just checked and this code is from 11 years ago so maybe the default changed and now it's not that important :watch: https://github.com/matomo-org/matomo/commit/23ad6dfa4e9413ee0fdb571bf8c69efe75ec5265

PR #14989, opened 2019-10-10
Disable update of some resources when auto update is disabled

tsteur:
When auto update is disabled, we probably also don't want to update these resources from GitHub.

If needed I can also introduce a new config setting but reckon auto update also kind of falls into this category maybe? Can also create a new config setting...  If so, any thoughts on the wording?

### DIFF
--- a/plugins/CoreAdminHome/Tasks.php
+++ b/plugins/CoreAdminHome/Tasks.php
@@ -74,7 +74,7 @@ class Tasks extends \Piwik\Plugin\Tasks
         $this->daily('cleanupTrackingFailures', null, self::LOWEST_PRIORITY);
         $this->weekly('notifyTrackingFailures', null, self::LOWEST_PRIORITY);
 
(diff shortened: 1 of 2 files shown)


sgiehl:
Not sure if we should put that into a new config setting. If not we should at least mention it somewhere in the config file and maybe in an FAQ or similar so it's well documented

tsteur:
@mattab any thoughts?

mattab:
I have no strong preference so could merge. As Stefan said we would just need to update faqs and the INI setting doc to clarify.

tsteur:
I was just adding documentation for it and was looking up some things when I noticed it would actually cause a regression as we would always use outdated lists... as soon as an entry is stored in the DB, it would always use the entry from the DB and not the file anymore. So if someone adds `enable_auto_update` or adds `internet features disabled` then those users end up always having an outdated list and I would regress this for users who have already disabled enable_auto_update. I'll hook into scheduled tasks instead.

PR #14992, opened 2019-10-11
Remove not needed angular mocks library from release

tsteur:
It seems this is only needed for tests but we actually exclude the tests from the release so we can also exclude this in the package.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -56,3 +56,4 @@ tests/ export-ignore
 .scrutinizer.yml export-ignore
 .travis.yml export-ignore
 PULL_REQUEST_TEMPLATE export-ignore
+/libs/bower_components/angular-mocks export-ignore


.gitattributes
```diff
@@ -56,3 +56,4 @@ tests/ export-ignore
 .scrutinizer.yml export-ignore
 .travis.yml export-ignore
 PULL_REQUEST_TEMPLATE export-ignore
+/libs/bower_components/angular-mocks export-ignore
```
sgiehl: Is this lib required for development or why don't we remove it completely then?

tsteur: @sgiehl it seems to actually be required in some JS tests for dev. So it would just not be needed in the release.

sgiehl: ah ok

sgiehl (approved):

PR #15007, opened 2019-10-15
do not call API methods directly

Kaustubh72:
do not call API methods directly


sgiehl:
@Kaustubh72 If there is any good reason for opening a PR with a very old commit that is not even yours, please let us know. Closing this til then

PR #15015, opened 2019-10-16
Fix some sparkline issue fatal errors

tsteur:
Failed for me locally when switching metric in evolution graph

### DIFF
--- a/core/Visualization/Sparkline.php
+++ b/core/Visualization/Sparkline.php
@@ -62,7 +62,7 @@ class Sparkline implements ViewInterface
         $thousandSeparator = Piwik::translate('Intl_NumberSymbolGroup');
         $decimalSeparator = Piwik::translate('Intl_NumberSymbolDecimal');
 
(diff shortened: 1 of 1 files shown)


tsteur:
@mattab now some tests are failing see https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/36568/

What are those points actually meaning in the sparklines?

mattab:
The changes are not expected, there shouldn't be any change in the sparkline in non-comparison mode :thinking:  eg. this one https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/36568/SingleMetricView_goal_metric.png or this one https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/36568/Comparison_visitors_overview_no_compare.png 


* Red point represents the min value,
* green point the max value,
* Blue point the last value

tsteur:
Got some more errors... just realised vendor was out of date :(

PR #15016, opened 2019-10-16
Fix build failures

katebutler:


plugins/Referrers/tests/System/ApiTest.php
```diff
@@ -63,7 +63,7 @@ public function getApiForTesting()
                 'date' => '2010-01-01',
                 'periods' => 'year',
                 'testSuffix' => 'phpSerialized',
-                'format' => 'original',
+                'format' => 'php',
```
tsteur: @katebutler as discussed we probably need to adjust instead the original handler to respect `hideIdSubDatatable`. The test was added to ensure when the dataTable object is serialized that no random information is being added see https://github.com/matomo-org/matomo/commit/46ed8c2eef4990688c1188b31bc8d8229087daab  . But now an array is serialized and the actual logic is no longer tested. If possible be good to still use `original` and also add a comment that this test is testing the serialized DataTable

plugins/API/Renderer/Original.php
```diff
@@ -100,8 +100,20 @@ private function serializeIfNeeded($response)
...
+                $response = $this->removeIdSubDataTables($response);
+            }
+            return $response;
         }
         return $response;
     }
+
+    public function removeIdSubDataTables($response)
+    {
+        $response = preg_replace('/"subtableId";i:[\d]+;/', '', $response);
```
tsteur: @katebutler to possibly avoid regexes would it be better to do something similar as in PHP renderer where we iterate over all rows and then call in this case `$row->removeSubtable()` like

```php
foreach ($table->getRows() as $row) {
 if ($this->hide...) {   $row->removeSubtable(); }
}
```

haven't tested but could work

PR #15022, opened 2019-10-19
Update README.md

pratik07101999:


Findus23:
Copying text from other peoples websites (https://www.g2.com/products/matomo-formerly-piwik/competitors/alternatives) isn't really a meaningful contribution to the README

PR #15026, opened 2019-10-20
Use thousands separator in single metric view

### LINKED ISSUE #14945: In Comparison mode, numbers are not formatted (thousands separator)
See following screenshots where in Comparison mode, numbers next to Sparklines are not formatted (missing thousands separator). Be great to use the same code as when displaying standard sparklines.

Screenshot from 2019-10-03 16-50-50
Screenshot from 2019-10-03 16-50-34
Screenshot from 2019-10-03 16-49-53

katebutler:
Fixes #14945

### DIFF
--- a/plugins/CoreVisualizations/angularjs/single-metric-view/single-metric-view.component.html
+++ b/plugins/CoreVisualizations/angularjs/single-metric-view/single-metric-view.component.html
@@ -6,7 +6,7 @@
     </piwik-sparkline>
     <div class="metric-value">
         <span title="{{ $ctrl.metricDocumentation }}">
-            <strong>{{ $ctrl.metricValue }}</strong> {{ ($ctrl.metricTranslation || '').toLowerCase() }}
+            <strong>{{ $ctrl.metricValue.toLocaleString() }}</strong> {{ ($ctrl.metricTranslation || '').toLowerCase() }}
         </span>
         <span class="metricEvolution"
               ng-if="$ctrl.pastValue !== null"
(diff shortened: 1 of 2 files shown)


tsteur:
@katebutler not sure if this only fixes the issue for the generic single widget or for all reports? In this case @mattab meant for example the Acquisitions => Overview report. 

In 3.11 this was formatted, eg 
image
and in 3.12 this is not the case anymore
image

The solution for this specific widget seems to work but we usually format metrics server side to have consistent formatting everywhere depending on the chosen language / locale AFAIK. Wonder if somewhere maybe some `format_metrics: 'all'` or similar is missing for example or if there is some other issue.

tsteur:
BTW also possible for example that the `|number` filter in a sparkline visualisation template is missing.

katebutler:
Have found and fixed the Twig macro that was outputting the raw numbers on Acquisitions => Overview.
Regarding the single metric view, I've looked into the API calls that it makes and it fetches the metrics both with `format_metrics=all` and with `format_metrics=0`.  In both responses the integer values are unformatted.  `format_metrics=all` calls `$metric->format($value)` for the processed metrics only, and also does currency formatting for other metrics if the column name contains "revenue".

KarthikRaja1388:
@tsteur This issue seems to be happening. Customer is using 3.13.6 and this happens when the language is set to Deutsche

image

tsteur:
@KarthikRaja1388 this seems to only apply here to single metric view and not to a regular report. Can you create an  issue with steps on how to reproduce it? Be good to check if this applies to any Matomo report or only custom reports. If it's only custom reports then we'd create an internal issue.

PR #15028, opened 2019-10-21
Fix possible error in rowevolution for reports with fixed view types

sgiehl:
When a report has a fixed view type (by implementing `alwaysUseDefaultViewDataTable`) the row evolution popover might throw an error as it tries to render this default view type instead of the rowevolution graph.

fixes DEV-81


tsteur:
restarted the UI build https://travis-ci.org/matomo-org/matomo/jobs/600589983 hoping it runs through in time ...

tsteur:
Restarting the job again as it didn't finish in time... most tests succeeded though from the ones that finished.

Issue #4481, opened 2014-01-06
Display Inline help icon linking to User Guide, for all features with a user guide


mattab:
Matomo lets you see inline help for some of the features eg:
1. Hover on a title 
![Screenshot from 2020-04-03 16-31-35](https://user-images.githubusercontent.com/466765/78321502-a6c41a00-75c8-11ea-9289-ae46958ada0f.png)
2. Click on the help icon:
![Screenshot from 2020-04-03 16-31-37](https://user-images.githubusercontent.com/466765/78321516-b8a5bd00-75c8-11ea-87ab-33ebcb98f447.png)

Some of the features already have a Help icon in the Piwik software, linking to the relevant user guide. eg. Transitions, Row Evolution

But currently, many of our reports don't link to the user guide.

Yet almost all features and reports are documented in their own pages.
- [View list of all user guides](http://piwik.org/docs/)
- [DB size](http://piwik.org/docs/managing-your-databases-size/)
- [Segmentation](http://piwik.org/docs/segmentation/)
- [Page speed](http://piwik.org/docs/page-speed/)
- [Scheduled email reports](http://piwik.org/docs/email-reports/)
- [Analytics Privacy](http://piwik.org/docs/privacy/)
- [Analytics Security](http://piwik.org/docs/how-to-secure-piwik/)
- [Configure Piwik for speed](http://piwik.org/docs/optimize-how-to/)

Ideally all doc pages should be linked from the Help icon for consistency. 

When this ticket is closed, all features that have a user guide on matomo.org will be clearly linking to this user guide. This will help users make better use of Matomo and help them discover new features!

mattab:
Be awesome to have all Matomo reports show the Help icon linking to the user guide, whenever the report/plugin has a user guide. Over the last months we've added user guides for almost all core features, free plugins, premium plugins and it would be valuable to surface the guides within the product.

Findus23:
@mattab I guess for this #14530 needs to be solved first.

sgiehl:
I had a rough look through the reports and create two PRs where the guide urls were missing.
@mattab Can we maybe close this issue afterwards and check if the guide urls for the reports are set in Matomo whenever we publish a new/updated guide?

mattab:
btw here is a new guide we published recently: https://matomo.org/docs/dashboards/

> check if the guide urls for the reports are set in Matomo whenever we publish a new/updated guide?

Good idea, could you please create a ticket in KB to mention this? (we'll add it to the workflow)

[...]

PR #15030, opened 2019-10-21
Adds links to online guides in reports help

sgiehl:
In order to show links to the online guide in all reports where one is available, I've implemented a new report property `onlineGuideUrl`. If it's defined for a report, a link will be shown in the report help.
In addition that url will also be available in the metadata and thus when fetch all reports meta data through API.

Would currently look like this if a documentation and a archive date is set for the report:

image

refs #4481

### DIFF
--- a/plugins/CoreHome/templates/_dataTable.twig
+++ b/plugins/CoreHome/templates/_dataTable.twig
@@ -42,7 +42,7 @@
 
     <div class="reportDocumentation">
         {% if properties.documentation|default is not empty %}<p ng-bind-html="{{ properties.documentation|json_encode|e('html_attr') }}"></p>{% endif %}
-        {% if properties.onlineGuideUrl|default is not empty %}<a href="?module=Proxy&action=redirect&url={{ properties.onlineGuideUrl|e('html_attr') }}" target="_blank" class="onlineGuide">{{ 'CoreHome_ReadMoreOnlineGuide'|translate }}</a>{% endif %}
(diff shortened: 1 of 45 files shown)


tsteur:
Looks good but haven't looked at tests yet. Be good to not merge this in 3.12 but 3.13 or so (which you already added as milestone)

plugins/CoreHome/templates/_dataTable.twig
```diff
@@ -42,7 +42,8 @@
 
     <div class="reportDocumentation">
         {% if properties.documentation|default is not empty %}<p ng-bind-html="{{ properties.documentation|json_encode|e('html_attr') }}"></p>{% endif %}
-        {% if reportLastUpdatedMessage is defined and reportLastUpdatedMessage %}<span class='helpDate'>{{ reportLastUpdatedMessage|raw }}</span>{% endif %}
+        {% if properties.onlineGuideUrl|default is not empty %}<a href="?module=Proxy&action=redirect&url={{ properties.onlineGuideUrl|e('html_attr') }}" target="_blank" class="onlineGuide">{{ 'CoreHome_ReadMoreOnlineGuide'|translate }}</a>{% endif %}
```
diosmosis: Is there a reason to use Proxy here instead of `<a rel="noreferrer noopener">` link?

tsteur: Would we also need to add `|safelink` filter? Might not be needed for our own links but I reckon cannot hurt?

sgiehl: Can change that. Thought might be better to use the proxy so no third party plugin could add malicious here

Issue #14835, opened 2019-08-30
Page Overlay - JavaScript Error - TypeError: location.match(...) is null


localhorst:
When opening page overlays and selecting "today" or "yesterday", the page does not load. I get this error in Firefox Console: 

Firefox 68.0.2 (64-bit)
```
TypeError: location.match(...) is nullPiwik_Overlay.js:148:33
    setIframeOrigin https://matomo.mydomain.com/plugins/Overlay/javascripts/Piwik_Overlay.js?cb=b7a7ffc1b1f573a45213fcfecfcdc04d:148
    hashChangeCallback https://matomo.mydomain.com/plugins/Overlay/javascripts/Piwik_Overlay.js?cb=b7a7ffc1b1f573a45213fcfecfcdc04d:170
    init https://matomo.mydomain.com/plugins/Overlay/javascripts/Piwik_Overlay.js?cb=b7a7ffc1b1f573a45213fcfecfcdc04d:278
    invoke https://matomo.mydomain.com/index.php?module=Proxy&action=getCoreJs&cb=b7a7ffc1b1f573a45213fcfecfcdc04d:163
    init https://matomo.mydomain.com/plugins/Overlay/javascripts/Piwik_Overlay.js?cb=b7a7ffc1b1f573a45213fcfecfcdc04d:273
    <anonymous> https://matomo.mydomain.com/index.php?module=Overlay&period=week&date=today&idSite=4:260
    i https://matomo.mydomain.com/index.php?module=Proxy&action=getCoreJs&cb=b7a7ffc1b1f573a45213fcfecfcdc04d:4
    fireWith https://matomo.mydomain.com/index.php?module=Proxy&action=getCoreJs&cb=b7a7ffc1b1f573a45213fcfecfcdc04d:4
    ready https://matomo.mydomain.com/index.php?module=Proxy&action=getCoreJs&cb=b7a7ffc1b1f573a45213fcfecfcdc04d:4
    J https://matomo.mydomain.com/index.php?module=Proxy&action=getCoreJs&cb=b7a7ffc1b1f573a45213fcfecfcdc04d:4
```

A similar error occurs in Chrome 76.0.3809.100 (64-Bit)
```
Uncaught TypeError: Cannot read property '0' of null
    at setIframeOrigin (Piwik_Overlay.js?cb=b7a7ffc1b1f573a45213fcfecfcdc04d:148)
[...]

PR #15032, opened 2019-10-21
Some overlay improvements

tsteur:
refs #14835 

First fix improves the loading when you have eg configured

https://fooBar.com as a site url, and the domain in the url is foobar.com

The second fix is event.data related where in this case a website was firing an event and the site was posting event.data as an object

They don't fix the actual issue but make it more reliable


sgiehl (approved):
Changes look legit and tests are still working so :+1:

PR #15037, opened 2019-10-22
Prevent warning array to string conversion in console renderer

tsteur:
See https://forum.matomo.org/t/notice-array-to-string-conversion-matomo-3-11-0/34644

As we don't have more information yet I couldn't reproduce it. Looking at the code it seems though like some metadata has an array in there. I didn't even quite know we had this renderer and from what I can see we're using it to convert a dataTableMap to a string. Not sure when and why. It should fix the warning though.


tsteur:
BTW I was seeing this error heaps of times in the logs.

Noticed it comes from twig when checking if an object is empty

image

in the dataTable twig

image
(could be a performance tweak to adjust this as the dataTable would maybe always return a string anyway)

Issue #15033, opened 2019-10-22
3.12.0-rc1: No dashboard graph anymore?


dertuxmalwieder:
When updating my Matomo installation from the last 3.12.0 beta version to 3.12.0-rc1, the graph of recent visitors in my dashboard stopped working:

![HEgfZvJ](https://user-images.githubusercontent.com/125629/67253375-9df6e600-f477-11e9-97b4-27431fd43322.png)

This does *not* mean that suddenly nobody cares about my website anymore. The graph on the Visitors page works:

![ox8Gsct](https://user-images.githubusercontent.com/125629/67253385-ab13d500-f477-11e9-8f7d-29aa37fd17cb.png)

So does the other visitors dashboard graph:

![wjTlavF](https://user-images.githubusercontent.com/125629/67253406-c848a380-f477-11e9-86eb-c5ac24d5852f.png)

I really prefer the "Graph der letzten Besuche" though. It is less *roomy*.

tsteur:
Only thing I can image is that it's related to https://github.com/matomo-org/matomo/compare/3.12.0-b7...3.12.0-rc1#diff-2c49596f288bc707bfda47f4b648bc57R241 but I can't reproduce it and it would be maybe only an issue if unique visitors was selected before. @sgiehl any thoughts?

tsteur:
@dertuxmalwieder does it work when you remove the widget and add it again?

dertuxmalwieder:
Sadly, it does not.

tsteur:
Do you know if you have any custom configurations in your `config/config.ini.php`? Any chance you have eg `enable_processing_unique_visitors_day` disabled? By looking at the image I suppose you're viewing day when this happens?

tsteur:
Any chance you could try this patch @dertuxmalwieder ? https://github.com/matomo-org/matomo/pull/15038/files

dertuxmalwieder:
The patch fixes it! :-) Thank you.

No custom configuration involved.

PR #15038, opened 2019-10-22
Request different columns when no columns will be available

tsteur:
refs https://github.com/matomo-org/matomo/issues/15033 
it might fix that issue

fixes a regression from https://github.com/matomo-org/matomo/pull/14996

When you have the Visits over time for example in the dashboard, view day period, select unique visitors, then select year period, it will then show visits metric instead of unique visitors because unique visitors is not available for year. The visits metric will not show any data though because it had the `&columns=nb_unique_visitors` param applied and the visits metric was never fetched. That's because in https://github.com/matomo-org/matomo/pull/14996 the metric was replaced after the dataTable was being requested. Trying to do this now before requesting dataTable too. Not sure what other regressions this could potentially introduce...


tsteur:
Tests are succeeding and user confirms it's working so merging the PR. @sgiehl if you could later maybe have a quick look at the PR be great

PR #15048, opened 2019-10-24
Send queued requests slightly less often and have method to change interval

tsteur:
Eg instead of waiting for 1.75s we now wait for 2.5seconds. Since we're falling back to sendBeacon anyway there should be no risk in missing any information.

And there is a new method so plugins or users can change the interval.


sgiehl (approved):
Haven't tested, but code looks fine and tests are still passing :+1:

PR #15055, opened 2019-10-24
Updates device detector to 3.12.1

sgiehl:
ping @Findus23   the new device type `smart speaker` might need an icon


Findus23:
Is the material.io speaker icon okay?

image

sgiehl:
should be ok for now. we can change it later again if needed

Findus23:
Added the icon. Can you please checkout the latest submodule and check if it works.

PR #15056, opened 2019-10-24
Fix error core JS proxy couldn't be decoded when gzipped

tsteur:
refs https://github.com/matomo-org/wp-matomo/pull/40

This happens when eg WordPress might be triggering a notice before we do the "readfile" in which case the notice will be mixed with the gzipped content in the response and therefore the browser cannot read the gzipped JS/CSS. You can easily reproduce this by eg replacing the `ob_get_clean` with an `echo 'foobar';`

Something I don't quite understand yet is why it works when echoing something after the `_readfile`. In theory we would need to exit right away.

### DIFF
--- a/core/ProxyHttp.php
+++ b/core/ProxyHttp.php
@@ -170,7 +170,13 @@ class ProxyHttp
             Common::sendHeader('Content-Encoding: ' . $encoding);
         }
 
-        @ob_get_clean();
+        // in case any notices were triggered before this point (eg in WordPress) etc.
+        // it would break the gzipped response since it would have mixed regular notice/string plus gzipped content
+        // and would not be able to decode the response
+        $levels = ob_get_level();
+        for ( $i = 0; $i < $levels; $i++ ) {
+            ob_end_clean();
+        }
 
         if (!_readfile($file, $byteStart, $byteEnd)) {
             Common::sendResponseCode(500);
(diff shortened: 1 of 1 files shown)


core/ProxyHttp.php
```diff
@@ -170,6 +170,8 @@ public static function serverStaticFile($file, $contentType, $expireFarFutureDay
             Common::sendHeader('Content-Encoding: ' . $encoding);
         }
 
+        @ob_get_clean();
```
sgiehl: As the buffer content does not matter and is not used, guess a `ob_end_clean()` might be better

tsteur: In WordPress the buffer may be used so didn't want to end them as wasn't sure if it could have any side effects. Shouldn't really make much of a difference? Although I suppose `ob_end_clean` would make sure it'll be printed I suppose? This is currently done in WordPress on shutdown by looping through all of them and then printing them basically. Eg ob_end_clean might call some callbacks that then will be executed and could break things when expecting some wrong output maybe not sure https://www.php.net/manual/en/function.ob-end-clean.php#42939

sgiehl: According to https://www.php.net/manual/en/function.ob-get-clean.php `ob_get_clean` also executes `ob_end_clean`. So guess it doesn't make much difference...

tsteur: 👍 didn't know. Changed it to ob_end_clean and also now iterating over multiple output bufferings

PR #15057, opened 2019-10-24
Prevent notice in tracker code generator

tsteur:
> /core/Tracker/TrackerCodeGenerator.php(254): Notice - Undefined index: host - Matomo 3.11.0

Eg happens from what I can see when no main url is defined

### DIFF
--- a/core/Tracker/TrackerCodeGenerator.php
+++ b/core/Tracker/TrackerCodeGenerator.php
@@ -245,16 +245,29 @@ class TrackerCodeGenerator
         $websiteHosts = array();
         $firstHost = null;
         foreach ($websiteUrls as $site_url) {
+            if (empty($site_url)) {
+                continue;
+            }
+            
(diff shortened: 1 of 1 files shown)


core/Tracker/TrackerCodeGenerator.php
```diff
@@ -245,16 +245,29 @@ private function getJavascriptTagOptions($idSite, $mergeSubdomains, $mergeAliasU
...
+            }
             if (!empty($referrerParsed['path'])) {
                 $url .= $referrerParsed['path'];
             }
+            
+            if (!empty($url)) {
+                $websiteHosts[] = $url;
+            }
+            
             $websiteHosts[] = $url;
```
sgiehl: shouldn't that line be removed? otherwise each url would be added twice.

tsteur: 👍 fixed

sgiehl (approved):

PR #15058, opened 2019-10-25
More efficient heart beat timer

### LINKED ISSUE #15049: More efficient heart beat timer
Currently the heart beat timer is pinging the site every 15s if enabled. 

Since we're now using sendBeacon and can quite ensure a request will make it through onPageUnload, we don't really need to ping every 15 seconds. I reckon it's fine to only ping

* onWindowBlur
* every 15 minutes (since a visit is usually active for say 30 minutes anyway)
* stop pinging after maybe 1 hour? or earlier? I think heart beat is not keeping the visit alive anyway so we should stop after 30 minutes from page view start by default (or whatever is configured as visit length).
* When tab / window closes (if the window was not focussed before (a blur happened) we would maybe only send the ping event if the window had focus after a blur for at least 10 seconds... and only if the tab is being closed within the visit length of eg 30 minutes)

This will significantly reduce the footprint of this feature and would possibly allow us to enable this feature by default in https://github.com/matomo-org/matomo/issues/8225

tsteur:
fix https://github.com/matomo-org/matomo/issues/15049

Currently, when the heart beat timer is enabled, we ping every 15s the Matomo server. If you have 1000 visitors on your page that makes 66req/s. It also sends a request every time you focus the window after you were on another tab.

This PR implements pretty much the opposite behaviour:
* Ping only on window blur (never on focus)
* On window blur send only a request if the user was on the page for at least 15 seconds (or whatever is configured as heart beat timeout). Otherwise user maybe was just quickly switching between tabs but not actively on the page
* Ping on unload of browser window but only when user was on page > 15 seconds (or configured heart beat timeout)
* Stop sending ping requests 30 minutes after the first request (as it would not have any effect anyway)
* Now requiring at least 5 seconds to be configured as heart beat timer as every second be quite extreme and we want to prevent this.

This means instead of possibly sending heaps of requests every few seconds we typically would maybe send one or two ping requests per pageview. Most of the times it be one (on page unload) or even none (when user was only a few seconds on a page).

Behaviour is otherwise the same, it's just more efficient by not sending as many requests. Instead we only send requests when needed. Which is either when the user is leaving the window, or closes the window.


tests/javascript/index.php
```diff
@@ -4243,28 +4244,28 @@ function generateAnIframeInDocument() {
...
+            // test ping not sent on focus
             tracker.setCustomData('token', 6 + tokenBase);
 
-            triggerEvent(window, 'focus'); // ping request sent here
+            triggerEvent(window, 'focus'); // no ping request sent here
 
             tracker.disableHeartBeatTimer(); // flatline
 
-            return Q.delay(1000); // for the ping request to get sent
+            return Q.delay(1000); // for a ping request to get sent if there was one
```
diosmosis: Should we add a test here for when the time goes past `setVisitStandardLength()`? For example, we set `setVisitStandardLength()` to 4 or something, then wait 5 seconds and make sure a ping isn't sent.

tsteur: Good point 👍 Added that test and also resolved the merge conflicts.

ZupZ3r0:
Hi, 
First of all I appreciate this very effective change!
But did you consider Single Page Applications? Blur or unload may never happen in these cases. What I see currently:

- open the page initially: heartbeat ping gets fired when the page loses focus (I go to another tab or whatever) -> this works as described here and expected
- but any pagechanges inside this app (via its navigation) do not trigger pings

Maybe you can ping at on hashchange event or even better: provide a method to force ping. This would be helpful for matomo integrations in different frameworks or if we have own (custom)events that we are listening to.

tsteur:
@ZupZ3r0 AFAIK this should still work because all it does is update the overall visit time but not anything else.

tsteur:
@ZupZ3r0 btw there is also a tracker method named `_paq.push(['ping'])`

ZupZ3r0:
@tsteur apperently it doesn't work out of the box, because the heartbeat sends only the first ping on pageload but never again (or only if you leave the active tab). This doesn't really **break** something, but results in a visit time of 1s for every action/event.
Anyway thanks for the hint with `_paq.push(['ping'])`, thats what I was looking for and helps in this situation :-)

tsteur:
@ZupZ3r0 be great to let me know if using `ping` changes anything for you. Technically it shouldn't change anything from my understanding because no action/event will be updated in a ping request. Be great to give it a try though.

miiimooo:
I have been testing for a few days now with sending a ping (`_paq.push(['ping'])`) but I don't notice any difference in Avg Time on Page measurements.

PR #15062, opened 2019-10-28
Make getDenyHtaccessContent method public

tsteur:
Needing it in https://github.com/matomo-org/wp-matomo/pull/42/files


sgiehl (approved):
Guess wouldn't hurt to merge it for 3.12 already

PR #15077, opened 2019-10-30
When checking if a segment archive table exists, try to read data from an actual column

tsteur:
I'm trying to update WP-Matomo to 3.12.0 and noticed mysqli is not triggering an error when doing a `select 1`. At least in WordPress. Not sure it's an issue with our mysqli implementation as well but might be (haven't seen anything in tests though)


tsteur:
Shouldn't be needed

PR #15089, opened 2019-10-31
 Remove the SiteSpecificUserAgentExcludeEnabled setting

### LINKED ISSUE #14880: Visits with a User-Agent specified in 'Website > Excluded User Agents' are sometimes still tracked
Specifying User-agent to exclude for a website does not work.
Visits with a User-Agent specified in 'Website > Excluded User Agents' are sometimes still tracked.

### Reproduce

* In Administration > Global websites settings > Global list of user agents to exclude, check that `Enable website specific user agent exclusion` is un-checked (this is the default)
* Go to Administration > Websites > Edit
* Under "Excluded User Agents", enter `iplabel`  and click Save
* Visit `matomo.php?idsite=1&rec=1&test&debug=1&ua=Mozilla/5.0 (X11; CrOS x86_64 8172.45.0 ;iplabel) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.64 Safari/537.36&debug=1` to generate a request with a user agent containing `iplabel`
* Got: the visit is tracked. 
* Expected instead: visit excluded.

### Bug origin

Once we click the checkbox `Enable website specific user agent exclusion.` under  Administration > Global websites settings > Global list of user agents to exclude, then the User agent specified in the "edit website" works as expected and the visit is excluded.

### Solutions 

* either remove that checkbox completely and always enable this feature 
* or leave the checkbox and hide the "excluded user agent" setting in the Website Edit screen. + also throw an exception in the API if the value is set on a website.

katebutler:
Fixes #14880


tsteur:
@katebutler as discussed by good to add an entry to the changelog under the deprecations section that we will remove the API method in Matomo 4 and we would also add a test that the deprecated method will actually be removed then.

plugins/SitesManager/API.php
```diff
@@ -1066,28 +1065,22 @@ public function setGlobalExcludedUserAgents($excludedUserAgents)
      * only the global user agent substrings (see @setGlobalExcludedUserAgents) will be used.
      *
      * @return bool
+     * @deprecated
```
tsteur: could we change the comment to something like `@deprecated Will be removed in Matomo 4.0`?

plugins/SitesManager/API.php
```diff
@@ -60,7 +60,6 @@ class API extends \Piwik\Plugin\API
     const OPTION_SEARCH_KEYWORD_QUERY_PARAMETERS_GLOBAL = 'SitesManager_SearchKeywordParameters';
     const OPTION_SEARCH_CATEGORY_QUERY_PARAMETERS_GLOBAL = 'SitesManager_SearchCategoryParameters';
     const OPTION_EXCLUDED_USER_AGENTS_GLOBAL = 'SitesManager_ExcludedUserAgentsGlobal';
-    const OPTION_SITE_SPECIFIC_USER_AGENT_EXCLUDE_ENABLE = 'SitesManager_EnableSiteSpecificUserAgentExclude';
```
sgiehl: Wondering if we should create Updates to remove already existing options from the database if we completely remove one.  ping @tsteur @mattab

sgiehl (commented):
I wonder why the expected system test files changed. Can't see any changes causing that.
If those test changes aren't related to that PR, it might be better to not include the changes here...

plugins/SitesManager/API.php
tsteur: Could do as it's easy. Not 100% needed though.

katebutler:
The expected file changes are all related to the change in the `OneVisitorTwoVisits` fixture, which was using the site-specific useragent exclude setting.

sgiehl:
> The expected file changes are all related to the change in the OneVisitorTwoVisits fixture, which was using the site-specific useragent exclude setting.

Then I guess I don't understand why that should change the test fixtures. Before the OneVisitorTwoVisits might have set a site specific useragent exclude, but disabled the site specific exclude globally in that case before. So it actually shouldn't have had any effect on the test fixture. So removing that part should also not have any effect. Or did I get anything wrong? :thinking:

tsteur:
@sgiehl I haven't tested it yet but code looks generally good. I was expecting the change eg because of https://github.com/matomo-org/matomo/pull/15089/files#diff-9cceb103a81e46f4388517c3b4c33427L245 maybe?

PR #15095, opened 2019-11-03
Fix zend session hides actual error when there are two or more errors…

tsteur:
…/notices/warnings

Need to append each warning.

### DIFF
--- a/libs/Zend/Session/Exception.php
+++ b/libs/Zend/Session/Exception.php
@@ -58,7 +58,7 @@ class Zend_Session_Exception extends Zend_Exception
         if (!isset(self::$sessionStartError)) {
             self::$sessionStartError = '';
         }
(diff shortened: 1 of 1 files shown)


libs/Zend/Session/Exception.php
```diff
@@ -55,7 +55,10 @@ class Zend_Session_Exception extends Zend_Exception
      */
     static public function handleSessionStartError($errno, $errstr, $errfile, $errline, $errcontext)
     {
-        self::$sessionStartError = $errfile . '(Line:' . $errline . '): Error #' . $errno . ' ' . $errstr;
+        if (!isset(self::$sessionStartError)) {
+            self::$sessionStartError = '';
+        }
+        self::$sessionStartError .= $errfile . '(Line:' . $errline . '): Error #' . $errno . ' ' . $errstr;
```
diosmosis: Might be good to separate the errors by newlines/spaces for readability? Otherwise LGTM.

PR #15098, opened 2019-11-03
Invert change comparison when displaying percents in the UI …

diosmosis:
and add extra _change_from metric to support API calculation of from change percent when there are more than two periods.

Changes:
- add invert_compare_change_compute query parameter that computes change to the selected period. this is done so clients don't have to do any strange work to get the change_from value from the other comparison series. useful for the mobile app, eg.
- add `_change_from` metrics to the output. This allows users to get the new metrics that are displayed through the API when any number of periods are compared.


tsteur:
@diosmosis is this UI change expected? https://builds-artifacts.matomo.org/matomo-org/matomo/comparison-percents-ui/36774/Comparison_totals_tooltip.png

In https://builds-artifacts.matomo.org/matomo-org/matomo/comparison-percents-ui/36774/Comparison_visits_overview_widget.png

image

do we still expect the first row to show the `minus` or should maybe the second have the `plus` percentage? I guess we wouldn't want to mix the behaviour as it would be inconsistent. (not 100% sure why we're making this change that's why I'm asking I guess)

BTW there's also at least one system test failing.

Code looks good otherwise

diosmosis:
You're comparing 12-17 w/ 1-31, so it's showing the change from 1-31 => 12-17, ie, jan 12-17 is 42% lower when compared with jan 1-31, so it's correct.

PR #15102, opened 2019-11-04
Fix Insight visualization layout

sgiehl:
The layout of the insight visualization is currently kind of broken, due to various changes. See for example https://builds-artifacts.matomo.org/matomo-org/plugin-CustomDimensions/fixbuild/772/CustomDimensions_report_action_insights.png

### DIFF
--- a/plugins/Insights/javascripts/insightsDataTable.js
+++ b/plugins/Insights/javascripts/insightsDataTable.js
@@ -60,9 +60,6 @@
                 $('td.label', domElem).width(parseInt(width * 0.50, 10));
             }
 
-            var showScrollbarIfMoreThanThisPxOverlap = 32;
-            this.overflowContentIfNeeded(domElem, showScrollbarIfMoreThanThisPxOverlap);
-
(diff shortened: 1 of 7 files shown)


plugins/Insights/javascripts/insightsDataTable.js
```diff
@@ -60,9 +60,6 @@
                 $('td.label', domElem).width(parseInt(width * 0.50, 10));
             }
 
-            var showScrollbarIfMoreThanThisPxOverlap = 32;
```
tsteur: @sgiehl is removing this maybe causing https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/36806/UIIntegrationTest_dashboard2.png ?

sgiehl: not sure. need to investigate that

sgiehl: should be fixed with the last commit

PR #15103, opened 2019-11-04
Ensure insight reports uses the correct compare period number for periods other than day

sgiehl:
When viewing a insight report for a period other than day it might happen, that the report shows an error like this:

image

This happens when someone chooses to compare to same day in last year when viewing a day report and switching to another period afterwards.

When switching the compared date this setting is stored in the report options. When switching to another period like month, it tries to compare the current month to the month 365 months ago, as the stored option will be used.

As all periods except day should always compare to the last period before. This fix should solve this issue.

### DIFF
--- a/plugins/Insights/Visualizations/Insight.php
+++ b/plugins/Insights/Visualizations/Insight.php
@@ -49,7 +49,7 @@ class Insight extends Visualization
             'reportUniqueId' => $report,
             'minImpactPercent' => $this->requestConfig->min_impact_percent,
             'minGrowthPercent' => $this->requestConfig->min_growth_percent,
-            'comparedToXPeriods' => $this->requestConfig->compared_to_x_periods_ago,
+            'comparedToXPeriods' => $this->getComparedToXPeriodsAgo(),
             'orderBy'  => $this->requestConfig->order_by,
             'filterBy' => $this->requestConfig->filter_by,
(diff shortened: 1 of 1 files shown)


plugins/Insights/Visualizations/Insight.php
```diff
@@ -59,6 +59,17 @@ public function beforeLoadDataTable()
         );
     }
 
+    private function getComparedToXPeriodsAgo()
+    {
+        $period = Common::getRequestVar('period', null, 'string');
```
tsteur: @sgiehl took me a while to understand what is happening here and seeing no `Range::getDateXPeriodsAgo()` is converting eg indeed `365` to 365 months ago. From what I see `week` and `year` only can compare to the previous period, but `compared_to_x_periods_ago` can be `1` or `12` for months. Would we need to convert eg a number `365` to `12` instead of 1 when the period is month?

Also probably not needed but could maybe have a side effect... (probs not and only if plugins where to maybe hook into it somehow or so)... Would we maybe need to change `$this->requestConfig->compared_to_x_periods_ago` in above code and set the updated value to `comparedToXPeriods`? I don't think it's actually used in any way but may be better.

sgiehl: added another condition for month to fix that

PR #15107, opened 2019-11-04
Prevent test cookie from being created when cookies are disabled

### LINKED ISSUE #15000: _pk_testcookies set while using "disableCookies"
I'm usign Matomo 3.12.0-b5 and several "_pk_testcookie" set by the hasCookie() function despite I am using "_paq.push(['disableCookies']);".

Now I'm wondern if it is a bug? Feature? Or am I missing something? Because setting cookies without the users consent could be a potential GDPR issue.

**Tracking code example:**
```javascript
        var _paq = window._paq || [];
        _paq.push(['disableCookies']);
        _paq.push(['trackPageView']);
        (function() {
            var u="URL";
            _paq.push(['setTrackerUrl', u+'matomo.php']);
            _paq.push(['setSiteId', '1']);
            var websiteIdDuplicate = 2;
            _paq.push(['addTracker', piwikUrl = u+'matomo.php', websiteIdDuplicate]);
            var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
            g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'matomo.js';
            s.parentNode.insertBefore(g,s);
        })();
</script>
```

**Cookies:**
matomo cookies

tsteur:
fix https://github.com/matomo-org/matomo/issues/15000


diosmosis:
Can a tracker JS test be added for this?

tsteur:
@diosmosis I don't think it's really possible since this would need to be tested kind of "on load" I think

dirk-graetz:
The cookie is still set!

tsteur:
@dirk-graetz make sure you are on the latest version of Matomo. Then I recommend you clear all cookies and reload the page. The testcookie should have a very old expiry date meaning it's set but invalidated at the same time (aka deleted). It appearing in the list, depending on the browser, is unavoidable. 

If you were to remove the tracking code afterwards for a test, you would notice the cookie is gone.

PR #15110, opened 2019-11-05
Fixes a couple segment handling regressions in the UI

### LINKED ISSUE #15080: Comparison feature displays segment definition instead of segment name when segmenting action based custom dimensions
action-based

diosmosis:
FIxes #15080

Changes:
* When trying to see whether we should decode a segment, favor encoded segment since it appears to be more common. (This fixes issue in #15080)
* Replace existing param in URL if it exists but empty (eg, segment=). (this fixes issue that occurs when on All Visits and editing/adding a segment, the URL ends up containing `&segment=&segment=...`).
* If editing segment during comparison, remove comparison in case edited segment is one being compared.

### DIFF
--- a/core/Segment.php
+++ b/core/Segment.php
@@ -123,7 +123,7 @@ class Segment
             // ignore
         }
 
-        if ($subexpressionsRaw > $subexpressionsDecoded) {
+        if ($subexpressionsRaw >= $subexpressionsDecoded) {
             $this->initializeSegment($segmentCondition, $idSites);
             $this->isSegmentEncoded = false;
         } else {
--- a/core/Segment.php
+++ b/core/Segment.php
@@ -66,6 +66,11 @@ class Segment
     protected $string = null;
 
     /**
+     * @var string
+     */
+    protected $originalString = null;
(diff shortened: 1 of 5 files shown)


tsteur:
@diosmosis there are quite a few system tests failing: https://travis-ci.org/matomo-org/matomo/jobs/607490406#L849

Not sure this is expected?

diosmosis:
Tried looking for the cause of the regressions, but couldn't figure out the other encoding issue so I ended up solving this a different way.

tsteur:
@diosmosis looks good. There's a failing system tests still though https://travis-ci.org/matomo-org/matomo/jobs/607961524#L810  probably from the fixture change in the PR. Not sure that fixture change is still needed or expected? Do you reckon it's possible to add a test for the case it fixes?

diosmosis:
Shouldn't be a fixture change here, that's probably something that happens locally on os x. Maybe.

tsteur:
@diosmosis if a test can be added for this that be great, otherwise feel free to merge.

PR #15114, opened 2019-11-05
Better detection of a succesfull start of a session

tsteur:
Differentiate between errors and warnings in session error handler and only assume session was not started if there was an error triggered. It should not throw an exception if there was only a notice or a warning.

Needed for WordPress. Eg there a caching plugin was trying to delete some cache after we executed a DB query, and did cache deletion on the file system triggered a warning. Then Zend_Session assumed the session was not started correctly even though it was and the warning was not related to session_start at all

image

### DIFF
--- a/libs/Zend/Session.php
+++ b/libs/Zend/Session.php
@@ -479,14 +479,14 @@ class Zend_Session extends Zend_Session_Abstract
                 restore_error_handler();
             }
 
-            if (!$startedCleanly || Zend_Session_Exception::$sessionStartError != null) {
+            if (!$startedCleanly || !empty(Zend_Session_Exception::$sessionStartError)) {
                 if (self::$_throwStartupExceptions) {
                     set_error_handler(array('Zend_Session_Exception', 'handleSilentWriteClose'), $errorLevel);
                 }
                 session_write_close();
                 if (self::$_throwStartupExceptions) {
                     restore_error_handler();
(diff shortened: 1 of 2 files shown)


libs/Zend/Session.php
```diff
@@ -479,14 +479,14 @@ public static function start($options = false)
                 restore_error_handler();
             }
 
-            if (!$startedCleanly || Zend_Session_Exception::$sessionStartError != null) {
+            if (!$startedCleanly || !empty(Zend_Session_Exception::$sessionStartError)) {
```
tsteur: ideally we would even only go into that logic if `!$startedCleanly` and not look at `Zend_Session_Exception::$sessionStartError` at all but I reckon we could still change that later.

libs/Zend/Session/Exception.php
```diff
@@ -55,10 +56,12 @@ class Zend_Session_Exception extends Zend_Exception
      */
     static public function handleSessionStartError($errno, $errstr, $errfile, $errline, $errcontext)
     {
-        if (!isset(self::$sessionStartError)) {
-            self::$sessionStartError = '';
+        $message = $errfile . '(Line:' . $errline . '): Error #' . $errno . ' ' . $errstr;
+        if (E_ERROR === $errno || E_CORE_ERROR === $errno || E_COMPILE_ERROR === $errno) {
```
tsteur: only trigger a session start exception when one of these 3 errors occur. Those 3 should be enough and no need to handle recoverable error etc.

Issue #15120, opened 2019-11-06
Considering the facebook/xhprof package is deprecated


peter279k:
As title. Firstly, we look at the `require-dev` block on `composer.json`, we can notice that it requires the `facebook/xhprof`.
After visiting this package page via the [Packagist link](https://packagist.org/packages/facebook/xhprof), I aware that following deprecated warning message:

```
This package is abandoned and no longer maintained. The author suggests using the lox/xhprof package instead. 
```

And I think we should also use the new package to replace the `facebook/xhprof ` :).

tsteur:
👍 totally. Feel free to create a PR 👍

tsteur:
Our profiler supports for example tideways profiler (not the remove service but the php extension)

tsteur:
Actually, closing this for now as xhprof was updated just last month.

PR #15122, opened 2019-11-06
Resolves issue #15120

peter279k:
# Changed log
- Resolves issue #15120.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,38 +54,12 @@
     "require-dev": {
         "aws/aws-sdk-php": "2.7.1",
         "phpunit/phpunit": "~4.8",
-        "facebook/xhprof": "dev-master",
         "phpseclib/phpseclib": "~0.3.8",
         "symfony/var-dumper": "~2.6.0",
-        "symfony/yaml": "~2.6.0"
+        "symfony/yaml": "~2.6.0",
+        "lox/xhprof": "dev-master"
     },
     "repositories": [
-        {
-            "type": "package",
-            "package": {
-                "name": "facebook/xhprof",
-                "type": "library",
-                "description": "XHProf: A Hierarchical Profiler for PHP",
-                "keywords": ["profiling", "performance"],
-                "homepage": "http://pecl.php.net/package/xhprof",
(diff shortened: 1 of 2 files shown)


sgiehl:
Thanks for the PR. We also need to update the composer.lock in this case.

peter279k:
Hi @sgiehl, thanks for your reply.

My question is: How to update the `composer.lock` file?

Using the `php-5.5.x`  to execute `composer` and using update command to update cached dependency information on `composer.lock` file?

sgiehl:
actually you would need to let composer update the composer.json & composer.lock by executing those commands (without your changes). The PHP version you are running doesn't matter as long as it satisfies the requirements (>5.5.x):
```
composer remove facebook/xhprof
composer require --dev lot/xhprof dev-master
```
After that I guess the additional repository entry for facebook/xhprof can be removed manually from composer.json

peter279k:
Hi @sgiehl, I've update the `composer.lock` file and please review this PR again.

Thanks.

composer.json
```diff
@@ -54,10 +54,10 @@
     "require-dev": {
         "aws/aws-sdk-php": "2.7.1",
         "phpunit/phpunit": "~4.8",
-        "facebook/xhprof": "dev-master",
         "phpseclib/phpseclib": "~0.3.8",
         "symfony/var-dumper": "~2.6.0",
-        "symfony/yaml": "~2.6.0"
+        "symfony/yaml": "~2.6.0",
+        "lox/xhprof": "dev-master"
```
sgiehl: could you also remove the package definition for `facebook/xhprof` the lines below (63-88).

peter279k: I've removed them.

sgiehl (approved):

sgiehl:
@peter279k  I'll close this one, as we won't change that for 3.x-dev branch anymore. Would you mind recreating the PR against 4.x-dev branch, if it still makes sense? Thx!

peter279k:
@sgiehl, thanks for your reply. I will recreating one PR for `4.x-dev` branch :).

PR #15131, opened 2019-11-07
accept numbers for setUserId, setEcommerceView, addEcommerceItem, removeEcommerceItem

### LINKED ISSUE #14714: JS Tracker: setUserId  and addEcommerceItem  sku should work when User ID is a number
Calling `setUserId` with a number instead of a string, results in the call to `setUserId` being ignored.

Reproduce with the following code:
```
_paq.push(['setUserId', 1234]);
_paq.push(['trackPageView']);

```
We expected same result as:
```
_paq.push(['setUserId', '1234']);
_paq.push(['trackPageView']);
```
But instead, the User id `1234` was not set and silently ignored.

A fix would be to simply allow numbers, something like this probably:

```js
--- a/js/piwik.js
+++ b/js/piwik.js
@@ -5964,10 +5964,9 @@ if (typeof window.Piwik !== 'object') {
* @param string User ID
*/
this.setUserId = function (userId) {
- if(!isDefined(userId) || !userId.length) {
- return;
+ if (isDefined(userId) && (typeof userId === 'number' || !userId.length)) {
+ configUserId = '' + userId;
}
- configUserId = userId;
};
```

Note:
* Also it would be great to look if other methods in the JS tracker may be affected and if so, also fix these, so that all methods could accept numbers/strings as one would expect an easy to use API to do :+1:

tsteur:
fix https://github.com/matomo-org/matomo/issues/14714

We ran into this issue ourselves actually eg in Matomo for WordPress etc https://github.com/matomo-org/wp-matomo/pull/67/files therefore moved it into 3.13.0 so userId and ecommerce tracking is easier.


js/piwik.js
```diff
@@ -7268,29 +7281,31 @@ if (typeof window.Piwik !== 'object') {
...
+                if (!isDefined(category) || category === null || category === false || !category.length) {
                     category = "";
                 } else if (category instanceof Array) {
                     category = JSON_PIWIK.stringify(category);
                 }
 
                 customVariablesPage[5] = ['_pkc', category];
 
-                if (isDefined(price) && String(price).length) {
+                if (isDefined(price) && price !== null && price !== false && String(price).length) {
```
tsteur: BTW I'm not sure it's on purpose we allow tracking price when there is only category set? Could create another issue for it... maybe it's on purpose ... @mattab

tsteur: BTW `String(false).length` would return `5` which is why I added specific check for false and null etc.

tests/javascript/index.php
```diff
@@ -3970,6 +3975,32 @@ function generateAnIframeInDocument() {
...
+        deepEqual( tracker3.getCustomVariable(5, "page"), ["_pkc", '343'], "All numbers Ecommerce view Category");
+        tracker3.deleteCustomVariables('page');
+
+        // all false
+        tracker3.setEcommerceView( false, false, false, false );
+        deepEqual( tracker3.getCustomVariable(2, "page"), false, "All numbers Ecommerce view price");
+        deepEqual( tracker3.getCustomVariable(3, "page"), false, "All numbers Ecommerce view SKU");
+        deepEqual( tracker3.getCustomVariable(4, "page"), false, "All numbers Ecommerce view Name");
+        deepEqual( tracker3.getCustomVariable(5, "page"), ["_pkc", ''], "All numbers Ecommerce view Category");
+        tracker3.deleteCustomVariables('page');
```
diosmosis: Should we add a similar test w/ `null`s?

tsteur: @diosmosis I don't think it's needed. Then we would need to start doing this kind of for all methods...

diosmosis: 👍

js/piwik.js
mattab: Can't remember if it's on purpose or not, but seems one should be able to track price even without a category set @tsteur

tsteur: Actually, if I see this right you can track just a category, or a category with a price. Not sure why you would track a category with a price but not an SKU or name... sounds to me like price makes only sense when also name or SKU is set?

PR #15134, opened 2019-11-08
Fix header alignment for multi-dimension reports

### LINKED ISSUE #15118: Column Title alignment in Custom report is inconsistent
Issue: Column title alignment in custom report is inconsistent. When viewing a custom report with multiple dimensions by making the report flat and choosing "show dimensions separately", first Column title is left aligned and others are right aligned.

image

Steps to reproduce:

1. Create a Custom Report with more than one dimension
2. View the report
3. Click on "Cog icon" on the bottom left of the report
4. Click on "Make it flat"
5. Click on "Cog icon" on the bottom left of the report
6. Click on "Show dimensions separately"

katebutler:
Fixes #15118


sgiehl:
That kind of breaks the scrolling if the table doesn't have enough space as all `label` columns are set `sticky`

katebutler:
Good point @sgiehl, have added CSS to ensure only the leftmost column will be made sticky.

plugins/CoreHome/stylesheets/dataTable/_dataTable.less
```diff
@@ -153,10 +153,13 @@ table.subActionsDataTable td.label,
 table.actionsDataTable td.label {
   border-top: 0;
   border-left: 0;
-  position: sticky; position: -webkit-sticky;
   left: 0;
   z-index: 1;
   background-clip: padding-box;
+
+  &:first-child {
```
tsteur: @katebutler I reckon this overwrites https://github.com/matomo-org/matomo/pull/15129/files#diff-8b20fa1b843d28bf8988e7c3c93a54b2R661 and breaks the other feature? 

Could we maybe simply add a new class to those column headers that does a textalign:left  ? This should reduce the risk of breaking things

tsteur:
@katebutler can you update below screenshot which seems maybe due to this change and then merge? Thanks

https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/37165/Comparison_subtables_loaded.png

tsteur (approved):

tsteur:
@katebutler any chance this related to your change? https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/37231/UsersManager_permissions_bulk_access_set.png

katebutler:
I wouldn't have thought so - it looks OK in my browser and the heading that's disappeared doesn't have the `first` class so shouldn't be affected.

Issue #15084, opened 2019-10-30
Document Mysql user needs new "CREATE TEMPORARY TABLES" permission


tsteur:
See https://dev.mysql.com/doc/refman/5.7/en/privileges-provided.html#priv_create-temporary-tables
Since Matomo 3.12 it seems this new permission is needed. Not sure if we somewhere document needed permissions? Would either need to add it, or document the list in general.

AFAIK this requirements are needed at least

* CREATE, DELETE, DROP, INSERT, SELECT, UPDATE, LOCK TABLES, CREATE TEMPORARY TABLES, ALTER, INDEX

diosmosis:
Looks like it's already documented? https://matomo.org/docs/requirements/#mysql-user-requirements

tsteur:
👍

PR #15139, opened 2019-11-10
Adds additional system check for CREATE TEMPORARY TABLES permission

### LINKED ISSUE #15115: LoadDataInfile Diagnostics Check is done even in config this setting is disabled
When disabline the LOAD DATA Infile usage in config still it executes the Check, but in my eyes it should skip it and show the correct message.

### LINKED ISSUE #15132: Extend Diagnostic Checks with Temporary Table Feature of mysql
in this commit  first time a new mysql feature about temporary Tables was used: https://github.com/matomo-org/matomo/commit/7795db37be3dd75d0a26e617ffc4ed3c2af83587

I think this should be added as new Check in the Diagnostic Page because not all platforms (expecially when you use mysql loadbalancers like proxysql) may support this.

sgiehl:
Also hides the `LOAD DATA INFILE` check if that feature is disabled in config

fixes #15132
fixes #15115
refs #15084

### DIFF
--- a/plugins/Diagnostics/Diagnostic/DatabaseAbilitiesCheck.php
+++ b/plugins/Diagnostics/Diagnostic/DatabaseAbilitiesCheck.php
@@ -15,7 +15,7 @@ use Piwik\Translation\Translator;
 /**
  * Check if Piwik can use LOAD DATA INFILE.
  */
-class LoadDataInfileCheck implements Diagnostic
+class DatabaseAbilitiesCheck implements Diagnostic
 {
     /**
      * @var Translator
@@ -35,8 +35,15 @@ class LoadDataInfileCheck implements Diagnostic
             return array();
         }
 
-        $label = $this->translator->translate('Installation_DatabaseAbilities');
+        $result = new DiagnosticResult($this->translator->translate('Installation_DatabaseAbilities'));
+        $result->addItem($this->checkLoadDataInfile());
+        $result->addItem($this->checkTemporaryTables());
 
+        return [$result];
+    }
+
+    protected function checkLoadDataInfile()
+    {
         $optionTable = Common::prefixTable('option');
         $testOptionNames = array('test_system_check1', 'test_system_check2');
 
@@ -61,7 +68,7 @@ class LoadDataInfileCheck implements Diagnostic
         Db::exec("DELETE FROM `$optionTable` WHERE option_name IN ('" . implode("','", $testOptionNames) . "')");
 
         if ($loadDataInfile) {
-            return array(DiagnosticResult::singleResult($label, DiagnosticResult::STATUS_OK, 'LOAD DATA INFILE'));
+            return new DiagnosticResultItem(DiagnosticResult::STATUS_OK, 'LOAD DATA INFILE');
         }
 
         $comment = sprintf(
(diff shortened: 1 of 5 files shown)


plugins/Diagnostics/Diagnostic/DatabaseAbilitiesCheck.php
```diff
@@ -82,6 +93,25 @@ public function execute()
...
+        return new DiagnosticResultItem(DiagnosticResult::STATUS_WARNING, $comment);
+    }
+
+    protected function checkTemporaryTables()
+    {
+        $status = DiagnosticResult::STATUS_OK;
+        $comment = 'CREATE TEMPORARY TABLES';
+
+        try {
+            Db::exec("CREATE TEMPORARY TABLE `piwik_test_table_temp` (
```
tsteur: btw the issue was not just about temporary tables probably (which probably works most of the times but also whether we can create temporary table, change the transaction status level, then do a write to temporary table, and several reads from it succeed in case there's a load balancer with multiple readers). For now that check be fine but I think the original issue was a bit more complicated where temporary tables itself weren't an issue.

tsteur: it's not important though for now and can schedule this also for 4.1

plugins/Diagnostics/lang/en.json
```diff
@@ -1,6 +1,7 @@
 {
     "Diagnostics": {
         "ConfigFileTitle": "Config file",
+        "MysqlTemporaryTablesWarning": "MySQL Permission CREATE TEMPORARY TABLES is required for Matomo to work properly.",
```
mattab: could we maybe link to [this FAQ](https://matomo.org/faq/how-to-install/faq_23484/)?
refs  #15237

sgiehl: done

plugins/Diagnostics/Diagnostic/DatabaseAbilitiesCheck.php
tsteur: fyi @sgiehl

sgiehl: I've added a check that writing and reading from the temporary table works.
Not sure if we should add checking if transactions work correctly to a new check, as it should be unrelated to temporary tables at all?

tsteur: We will need to check transaction status level not transactions itself. This can be done in a different check though.

sgiehl: > check transaction status level

actually I'm not sure how to do that :man_shrugging:

tsteur: I think a check could look like this:

```php
$level = StaticContainer::get(TransactionLevel::class);
if (!$level->setUncommitted()) {
    // warning: Changing transaction isolation level not supported. Archiving will still work but it may be slower and it is recommended to change for example the `binlog_format` to `row` if possible.
} else {
$level->restorePreviousStatus();
}

sgiehl: added an additional check for that now

tsteur:
👍 
works 
image
Seems there is no screenshot for system check? At least nothing is failing. Feel free to merge @sgiehl

tsteur (approved):

sgiehl:
Added a UI test for the diagnostic page. Will merge, once it passed...

tsteur:
👍

PR #15153, opened 2019-11-12
Update composer dependencies

diosmosis:
Refs #14821

Includes:
* update of tcpdf w/ php 7.4 fixes
* update to branch of matomo-org/sparkline (fork of davaxi/sparkline) w/ @Findus23 's php 7.4 fix


sgiehl:
Is the failing system pdf test related to the tcpdf update?

diosmosis:
@sgiehl looks like it, i checked one of the PDFs and it appears identical to the expected.

Issue #13347, opened 2018-08-25
Visitor maps not showing regions correctly in most countries


CGsama:
I use the city db for geoip(php), all fine when using wordwide view. However, it couldn't show on reguion and city map. But correct at realtime map.
![5 ac r9qwjo91 un 6d 2o](https://user-images.githubusercontent.com/15722562/44621079-caac6f00-a86d-11e8-81c2-c1fd181cc5c1.png)
![s37 x s73o3r0 yh5x0xamp](https://user-images.githubusercontent.com/15722562/44621074-bd8f8000-a86d-11e8-8587-c95ca449af36.png)
![o6a xx h 0_snnmz k4](https://user-images.githubusercontent.com/15722562/44621075-bf594380-a86d-11e8-9309-5186b1d1fc5e.png)

CGsama:
I'm using the GeoIP2(php). Thanks for clarifying that. And for my IP which in Canada, I can see the region map shows correct.

parruc:
Same problem here: for Italy with GeoIP2 (php) I can see the cities correctly but all regions are marked as unknown.

sgiehl:
For GeoIP2 it will still show regions correctly if the FIPS and ISO code for a region is the same. That applies for US, CA and some smaller countries. We are working on updated maps to fully support the ISO codes GeoIP2 is using...

mattab:
Since the upgrade to Geoip2, most of countries regions are not displayed correctly in the Maps. 
Maybe we should consider disabling the feature of Region mapping on our maps, since it doesn't work mostly -> Moving into the milestone so we can take a decision.

 Because It feels bad to have a non working feature it may be easiest to disable region mapping completely

or should we "partially disable region mapping" if it still work for some countries? @sgiehl

thinkof4:
Will this bug ever be fixed, I mean this is an one year old issue...

dfabreguette-ap:
Same bug here ! Any workaround ?
![Capture d’écran 2019-10-17 à 09 39 43](https://user-images.githubusercontent.com/7055931/66988158-57456c80-f0c2-11e9-8c5c-1e010b30e981.png)

mirkomaty:
@sgiehl, @mattab 
What exactly is the problem with fixing this bug? These are the data we get from the maxdb:

{
	"city": {
...
		}
	},
	"continent": {
...	},
	"country": {
		"geoname_id": 2921044,
		"is_in_european_union": true,
		"iso_code": "DE",
[...]

PR #15168, opened 2019-11-18
Fix region mappings for map widget where possible

sgiehl:
Due to switching from FIPS to ISO with GeoIP2, almost all region maps are currently not showing any visits.
By adding the ISO codes to the existing maps (where an exact mapping is possible), the visits for regions should be shown for more regions again.
Some of the maps will still be broken or show incorrect data, as the iso codes can't be added everywhere.

refs #13347


sgiehl:
This should fix/improve the region mappings for most countries. For some countries a proper mapping isn't possible as they were completely reorganized, they will still show no data for GeoIp2. For some others the new mapping might not be fully exact, but it should be at least better than having none.

tsteur:
@sgiehl as the tests pass I reckon it's maybe good to merge? Or should we maybe wait to Matomo 4 since it's not a priority right now and the original issue is scheduled for matomo 4? Not sure how to test it so feel free to merge.

sgiehl:
@tsteur  We still need to create completely new maps for Matomo 4. But as that might take some more time, I thought it might be useful to fix most of the region maps to work correctly with GeoIP2. It was not possible for all countries, as a mapping from FIPS to GeoIP2 ist not possible (e.g. for France).
To test that, you can simply generate some visits and check if the regions on the maps are highlighted. (Only US and CA were already working before this change)

tsteur:
image
image

worked.

PR #15172, opened 2019-11-19
Add test to make sure woff2 file is up to date...

diosmosis:
... and update file.


tsteur:
fyi @diosmosis seeing an error here: https://travis-ci.org/matomo-org/matomo/jobs/613953629#L765 is this test still needed btw?

diosmosis:
> this test still needed btw?

It looks like the woff2 file isn't currently up to date (the test failed for me at first), so seems like it might be a useful check to have.

diosmosis:
Fixed the issue w/ tests.

PR #15183, opened 2019-11-21
Update spyc to 0.6.X

tsteur:
@sgiehl any thoughts on that? I reckon shouldn't be an issue?

I was getting this error in WordPress:

```
Cannot declare class Spyc, because the name is already in use
```

Because it doesn't use the regular autoloader and Spyc's autoloader/class is loaded on every request. Only in 0.6.X they are checking if the class is already loaded and don't redeclare it https://github.com/mustangostang/spyc/blob/0.6.3/Spyc.php#L46 compared to before they always tried to redeclare it https://github.com/mustangostang/spyc/blob/0.5.1/Spyc.php#L57

This is basically needed for WordPress to:
* Have better compatibility with other WP plugins
* Support WP-CLI

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "twig/twig": "~1.0",
         "leafo/lessphp": "dev-php74-compat",
         "symfony/console": "~2.6.0",
-        "mustangostang/spyc": "~0.5.0",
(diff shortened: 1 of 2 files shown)


sgiehl (approved):
Yes. shouldn't be a problem to update it.

Issue #14395, opened 2019-05-01
Warning in Chrome console: A cookie was set without the SameSite attribute.


neunzehnachtneun:
**Note: More info why this should be fixed in a soon Matomo release below: https://github.com/matomo-org/matomo/issues/14395#issuecomment-529198618**

Hi together,
this is my first issue in this project, so hopefully I do it right.
I would like to request a new feature to improve privacy. At the moment it is possible yet to set the cookie lifetime to a shorter timespan and to set the cookie secure.
I suggest to also set the cookie attributes HttpOnly and SameSite=Strict, so privacy would be improved and cross site request attacks forgery could be prevented.

Thank you!

fdellwing:
As far as I'm aware of every cookie that can be httponly is httponly. Can you show one that is not that way?

tsteur:
What do you refer to here @neunzehnachtneun ? To the JS tracker? In JS you cannot set a cookie as httponly

neunzehnachtneun:
Thank you for answering! Yes, I refer to the JS-Tracker and, yes, you are totally right. I read a bit about javascript and cookies and it's totally useless to set the cookie header to httponly.

But also I read about the [cookie header 'samesite=strict'](https://javascript.info/cookie#samesite), which is possible to use with javascript and which I would like to use with matomo. What do you think about that?

fdellwing:
While using `samesite` you need to really take care of your cookie domain or you will run into a hell of a lot of trouble. This is especially important if you use a reverse proxy.

tsteur:
You would also need to have matomo running on the very same domain? not even a subdomain? or would a subdomain still work?

MichaelHeerklotz:
samesite is a origin related setting and should be used with caution. it would have effects on the tracked data and most certainly break the 3rd party cookie in in many use cases

Findus23:
Seems like this is becoming far more important: Chrome is planning to make all cookies that don't have an explicit SameSite option have SameSite=Lax by default. At the moment it is an optional flag (`#same-site-by-default-cookies`), but chrome 80 (release in the beginning of January 2020) will have it enabled by default.
https://www.chromestatus.com/feature/5088147346030592

(Disclaimer: I haven't fully understood the topic and the explanations are all a bit vague, so this might be incorrect)
[...]

Issue #6505, opened 2014-10-23
Feature request: Separate opt-out for each site


akiebs:
Since we can have multiple sites tracked by one piwik installation it would be good to have a separate opt-out for each site. 
Turning tracking off on one site shall not switch tracking of another site tracked by the same piwik instance. It might confuse users coming to a site they never were and finding tracking turned off "by default". Just in case they turn it on, i.e. to get better recommendations on a specific site, and tracking switches automatically on another site as well even having turned it off before might cause troubles for the site owners.

Andreas

mattab:
It was suggested in #8578 that first party cookies would be useful to have, as we wouldn't have to use a pop-up to set the cookie on the third party domain.  refs #8646

Note: when we would use First party cookie for opt-out, we would effectively have a separate opt-out for each "hostname", not as the issue title says a separate opt-out "for each website in Piwik" (a website in piwik can have many URLs hostnames but still user would need to opt-out from each domain.

matomoto:
Check this and rewrite it for Matomo / Piwik: https://www.penultima.de/optin/google.php

tsteur:
Note: this can be achieved using a custom opt out form https://developer.matomo.org/guides/tracking-javascript-guide#optional-creating-a-custom-opt-out-form 

It's pretty much the only way to make this work reliably.

tsteur:
I believe this is the standard now with the new JS opt out. If this is not the case, please reopen the issue. It may help to configure the cookie domains to make this work.

PR #15184, opened 2019-11-21
Set a first-party cookie when user opts out of tracking

katebutler:
Part of #14395 (allowing opt-out to work in Chrome on HTTP sites), also related to #6505


js/piwik.js
```diff
@@ -7706,6 +7712,32 @@ if (typeof window.Piwik !== 'object') {
         // initialize the Piwik singleton
         addEventListener(windowAlias, 'beforeunload', beforeUnloadHandler, false);
 
+        window.addEventListener('message', function(e) {
+            try {
+                var data = JSON.parse(e.data);
```
tsteur: haven't tested it yet... do we need to check if the message is for us just to avoid any issues with potentially other messages? like include some 'matomo_optout' in the string and then check if `e.data && typeof e.data === 'string' && e.data.indexOf('matomo_optout') >= -1`? (not tested that particular code... and would need to make sure we post this string in the opt out)

js/piwik.js
```diff
@@ -7706,6 +7712,32 @@ if (typeof window.Piwik !== 'object') {
...
+            }
+
+            if (typeof data.loaded != 'undefined') {
+                // We received a response to our message sending the initial state to the iframe
+                // We can clear the interval so that we don't send any more
+                stopInitializationInterval();
+                return;
+            }
+
+            if (typeof data.opted_in == 'undefined') {
```
tsteur: btw we have some method for this `isDefined(data.opt_in)` or so is the name

js/piwik.js
```diff
@@ -7706,6 +7712,32 @@ if (typeof window.Piwik !== 'object') {
...
         addEventListener(windowAlias, 'beforeunload', beforeUnloadHandler, false);
 
+        window.addEventListener('message', function(e) {
+            try {
+                var data = JSON.parse(e.data);
+            } catch (e) {
+                return;
+            }
+
+            if (typeof data.loaded != 'undefined') {
```
tsteur: same here be good to use our isDefined method for this maybe

js/piwik.js
```diff
@@ -7934,8 +7966,39 @@ if (typeof window.Piwik !== 'object') {
...
                     apply(missedCalls[i]);
                 }
+            },
+
+            /**
+             * Send a message to the iframe which holds the opt-out form, telling it the current value of the 
+             * first-party opt-out cookie.
+             */
+            updateOptOutForm: function() {
+                if (typeof window.postMessage === 'undefined') {
```
tsteur: here as well be good to use isDefined method

js/piwik.js
```diff
@@ -7934,8 +7966,39 @@ if (typeof window.Piwik !== 'object') {
...
+                        var iframe = iframes[i];
+                        iframe.contentWindow.postMessage(JSON.stringify(optOutStatus), '*');
+                    }
+                    numAttempts++;
+                    // 10 times per second * 1200 = 2 minutes
+                    // If the iframe hasn't finished loading by now, it ain't gonna, so let's stop trying
+                    if (numAttempts > 1200) {
+                        clearInterval(optOutTimer);
+                    }
+                }, 100);
```
tsteur: As DOM interactions are quite expensive in terms of performance, be good to check this maybe only about every second or so?

plugins/CoreAdminHome/javascripts/optOut.js
```diff
@@ -1,22 +1,73 @@
 function submitForm(e, form) {
-    if (e.preventDefault) { // IE8 and below do not support preventDefault
```
tsteur: I haven't tested the feature yet but would have expected the preventDefault code and also below the setInterval is still needed?

js/piwik.js
katebutler: Have added some namespacing of the message attributes to make it less likely that we'll accidentally pick up messages intended for other scripts.  The best practice would be to also check the message origin against the domain we expect to receive it from, but I can't find a way to do this since the tracker and the tracked site could be on any domain.
I don't think I need to explicitly check that `e.data` is defined, is a string etc as I am catching the exceptions from `JSON.parse()` in situations like this.

js/piwik.js
katebutler: Sweet, using this throughout now

js/piwik.js
katebutler: The problem is that the optout iframe might be displaying incorrectly to the user (e.g. saying they haven't opted out when they have) until it receives this message.  The first time through the loop is usually unsuccessful because the iframe hasn't finished loading yet, and increasing the delay even to 200ms means there is a noticeable pause before the optout iframe updates itself.  Maybe we need a new approach - calling `postMessage` on the iframe before it finishes loading is also causing an error message to be printed to the console which isn't ideal.

plugins/CoreAdminHome/javascripts/optOut.js
katebutler: This code was to prevent the checkbox state from toggling until after the API call had been made, and then refresh the iframe to show the new checkbox state.  It seems a bit unnecessary when we are working with the first-party cookies, but if we do want to retain that workflow (i.e. sync the iframe form checkbox with the real world optout value after every change, to make sure that it reflects what will actually happen) I could put back the `preventDefault` and then fire a message back to the iframe after the first-party cookie has been set in the tracker JS.

js/piwik.js
tsteur: Could use `setTimeout` instead... and each time use something like `interval = numAttempts * 100`

tsteur:

js/piwik.js
```diff
@@ -7706,6 +7719,54 @@ if (typeof window.Piwik !== 'object') {
         // initialize the Piwik singleton
         addEventListener(windowAlias, 'beforeunload', beforeUnloadHandler, false);
 
+        window.addEventListener('message', function(e) {
+            var tracker = Piwik.getTracker();
+
+            var originHost = getHostName(e.origin);
+            var matomoHost = getHostName(tracker.getPiwikUrl());
```
tsteur: Would maybe need to adjust `getHostName` to check if `url` is set... otherwise if the tracking code is embedded on the page, but no async tracker is created (which some users do), then it might not return a string? Not sure though... looking at the code it seems `getPiwikUrl` always returns a string so it is likely fine

tsteur:
I was just testing the opt out. I was embedding the opt out into an html page and opted out. The third party `ignore` cookie was set. Then I added the tracking code to the same site. I then reloaded the page and it was suddenly saying

> You are not opted out

Even though it should say 

> You are opted out

The problem is that it seems to only check for the first party cookie but not for the third party cookie status here:

image

This will be pretty much the behaviour for all Matomo sites as none of the visitors that previously unsubscribed would have the first party ignore cookie yet. So it is important to only show "You are not opted out" if neither the first nor the third party ignore cookie  is set. If either of them is set, we assume the user is opted out.

I then opted out, and it was saying "opted out" after I clicked on it. However, when I reloaded the page, it was saying again  "not opted out".
* We should double check that the user was actually opted out before changing the status of the checkbox ideally. And if it didn't change the status, maybe we could show a message explaining the user it didn't work and to retry or to contact the website owner? Maybe could even adjust the message if the opt out is running on http? @mattab Any thoughts? It would basically not work when the site is embedded using http and the user does not have the tracking code on the same page or the opt out iframe domain does not match the tracker domain... In other cases either the first or the third party cookie should work... This message would be shown to actual visitors... 
* I noticed it was actually not changing the status because I had `disableCookies` in my tracking code (https://matomo.org/faq/general/faq_157/). I wonder if we need to still execute the methods `forgetConsentGiven` and `rememberConsentGiven` even when cookies are disabled? @Findus23 @mattab  

Stopped testing afterwards.

mattab:
> We should double check that the user was actually opted out before changing the status of the checkbox ideally. 

that'd be great.

> And if it didn't change the status, maybe we could show a message explaining the user it didn't work and to retry or to contact the website owner? Maybe could even adjust the message if the opt out is running on http? @mattab Any thoughts? It would basically not work when the site is embedded using http and the user does not have the tracking code on the same page or the opt out iframe domain does not match the tracker domain... 

If it's possible then be great to explain in the error message why it doesn't work eg `Opt-out feature is unfortunately not working because this site is not using https and the tracking code cannot be found on this page. Please contact the website administrator for help.`  or  `Opt-out feature is unfortunately not working because this opt-out iframe domain $DOMAIN does not match the analytics service domain $DOMAIN2. Please contact the website administrator for help.`


> I noticed it was actually not changing the status because I had disableCookies in my tracking code (https://matomo.org/faq/general/faq_157/). I wonder if we need to still execute the methods forgetConsentGiven and rememberConsentGiven even when cookies are disabled? @Findus23 @mattab

not sure of the code details, but the idea would be that consent mechanism should be completely independant of cookies and work the same whether or not cookies are enabled in tracker.

tsteur:
Note: we still need to ignore disableCookies setting when opting out or in...

tsteur:
FYI made a few tweaks... 

* making sure it works with more browsers as eg not all browsers support `startsWith` etc.
* Showing an error message if a user has popup blocker enabled (which it is by default on chrome) when changing the status
* Showing an error message  right away if cookies are disabled since it means it will most certainly not work I reckon
* If when the popup closes, the optOut using postMessage did not work, then we may show a warning if we are on http. Generally we only show warnings if we are on HTTP as in all other cases things should work. 
* We can't fully detect if unsubscribe works because when the popup changes the ignore cookie, the cookies in the opt out frame are not updated and as such only a reload would help. However, we cannot really just reload since we don't really know if we will receive a confirmation message of the opt out through postMessage or not... 

@diosmosis @Findus23 could you maybe have a look as well?

plugins/CoreAdminHome/javascripts/optOut.js
```diff
@@ -1,22 +1,220 @@
...
+
+// Strips off protocol and trailing path and URL params
+function getHostName(url)
+{
+    // scheme : // [username [: password] @] hostame [: port] [/ [path] [? query] [# fragment]]
+    var e = new RegExp('^(?:(?:https?|ftp):)/*(?:[^@]+@)?([^:/#]+)'),
+        matches = e.exec(url);
+
+    return matches ? matches[1] : url;
+}
```
diosmosis: Would using:

```
document.createElement('a');
a.href = url;
return a.hostname;
```

be a bit safer here?

diosmosis:
@tsteur looks ok to me though I didn't test locally and it is a fairly complicated PR. I like the interesting use of postMessage here, that seems like it could potentially have other use cases in the tracker (not just opt out and overlay).

It would be nice to be able to unit test the opt out JS if possible (maybe by loading it in karma and testing w/ angular unit tests?). Might be a bit of work to make that possible though; if it's a good idea we could create an issue for now.

PR #15185, opened 2019-11-21
Use appropriate SameSite values for cookies

katebutler:
Part of #14395

### DIFF
--- a/core/Cookie.php
+++ b/core/Cookie.php
@@ -8,6 +8,8 @@
  */
 namespace Piwik;
 
+use DeviceDetector\DeviceDetector;
+
 /**
  * Simple class to handle the cookies:
  * - read a cookie values
@@ -132,7 +134,7 @@ class Cookie
      * @param bool $Secure
      * @param bool $HTTPOnly
      */
-    protected function setCookie($Name, $Value, $Expires, $Path = '', $Domain = '', $Secure = false, $HTTPOnly = false)
+    protected function setCookie($Name, $Value, $Expires, $Path = '', $Domain = '', $Secure = false, $HTTPOnly = false, $sameSite = null)
     {
         if (!empty($Domain)) {
             // Fix the domain to accept domains with and without 'www.'.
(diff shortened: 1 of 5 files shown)


core/Cookie.php
```diff
@@ -132,7 +134,7 @@ protected function getDefaultExpire()
      * @param bool $Secure
      * @param bool $HTTPOnly
      */
-    protected function setCookie($Name, $Value, $Expires, $Path = '', $Domain = '', $Secure = false, $HTTPOnly = false)
+    protected function setCookie($Name, $Value, $Expires, $Path = '', $Domain = '', $Secure = false, $HTTPOnly = false, $sameSite = null)
```
tsteur: not too important but usually the other default values that aren't set are `false`... (null be generally better it's just for consistency be good to use `false`)

tsteur:

core/Cookie.php
```diff
@@ -414,4 +422,41 @@ public static function isCookieInRequest($name)
...
+     * for some requests which are deemed CSRF-safe, although other requests may have broken functionality.
+     * - On Safari, the "None" value is interpreted as "Strict".  In order to set a cookie which will be available
+     * in all third-party contexts, we have to omit the SameSite attribute altogether.
+     * @param string $default The value that 
+     * @return string SameSite attribute value that should be set on the cookie. Empty string indicates that no value
+     * should be set.
+     */
+    private static function getSameSiteValueForBrowser($default)
+    {
+        $sameSite = ucfirst($default);
```
tsteur: just fyi might be good to use `ucfirst(strtolower($default))`? Just in case...

core/Cookie.php
```diff
@@ -153,7 +156,8 @@ protected function setCookie($Name, $Value, $Expires, $Path = '', $Domain = '',
             . (empty($Path) ? '' : '; path=' . $Path)
             . (empty($Domain) ? '' : '; domain=' . $Domain)
             . (!$Secure ? '' : '; secure')
-            . (!$HTTPOnly ? '' : '; HttpOnly');
+            . (!$HTTPOnly ? '' : '; HttpOnly')
+            . (!$sameSite ? '' : '; SameSite=' . $sameSite);
```
tsteur: just to be safe we maybe want to rawurlencode the samesite value... and just to be safe maybe also domain and path even though they are values coming from our system?

core/Tracker/IgnoreCookie.php
```diff
@@ -61,7 +61,7 @@ public static function setIgnoreCookie()
             $ignoreCookie->delete();
         } else {
             $ignoreCookie->set('ignore', '*');
-            $ignoreCookie->save();
+            $ignoreCookie->save('None');
```
tsteur: should we move this for now into this PR that handles the opt out? https://github.com/matomo-org/matomo/pull/15184/files

tsteur (changes requested):

PR #15186, opened 2019-11-21
Use appropriate SameSite value for session cookie

katebutler:
Part of #14395

### DIFF
--- a/core/Session.php
+++ b/core/Session.php
@@ -201,15 +201,15 @@ class Session extends Zend_Session
      */
     public static function writeCookie($name, $value, $expires = 0, $path = '/', $domain = '/', $secure = false, $httpOnly = false, $sameSite = 'lax')
     {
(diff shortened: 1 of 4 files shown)


core/Session.php
```diff
@@ -185,4 +185,41 @@ public static function isSessionStarted()
...
+     * @param string $path
+     * @param string $domain
+     * @param bool $secure
+     * @param bool $httpOnly
+     * @param string $sameSite
+     * @return string
+     */
+    public static function writeCookie($name, $value, $expires = 0, $path = '/', $domain = '/', $secure = false, $httpOnly = false, $sameSite = 'lax')
+    {
+        $headerStr = 'Set-Cookie: ' . $name . '=' . $value;
```
tsteur: should the name and value by rawurlencoded for security etc? Ideally maybe also some other values?

tsteur:

PR #15189, opened 2019-11-23
Support CliMulti on OS X and print note about whether CliMulti is used in archive output

diosmosis:


tsteur:
@diosmosis looks good to merge. Only need to fix a test: https://travis-ci.org/matomo-org/matomo/jobs/615942754#L834

PR #15196, opened 2019-11-25
Add more debug info to archiving queries

tsteur:
Otherwise it is hard to find out  for what  period etc the query was. Ideally we would add segmentId instead of segmentHash but would need to refactor some code from segmentQueryDecorator to the segmentClass (the method that finds the segmentId for a segment)... 

example:
```sql
SELECT /* 2019-11-25,2019-11-26 sites 1 segmenthash 6fb490cfce17323e01f4d11d61ea9679 */ /* Core */
				count(distinct log_visit.idvisitor) AS `1`, 
			count(*) AS `2`, 
			sum(log_visit.visit_total_actions) AS `3`, 
			max(log_visit.visit_total_actions) AS `4`, 
			sum(log_visit.visit_total_time) AS `5`, 
			sum(case log_visit.visit_total_actions when 1 then 1 when 0 then 1 else 0 end) AS `6`, 
			sum(case log_visit.visit_goal_converted when 1 then 1 else 0 end) AS `7`, 
			count(distinct log_visit.user_id) AS `39`
			FROM
				piwik_logtmpsegmentde86627ef24972fd077a35a9eb0dc563 AS logtmpsegmentde86627ef24972fd077a35a9eb0dc563 INNER JOIN piwik_log_visit AS log_visit ON log_visit.idvisit = logtmpsegmentde86627ef24972fd077a35a9eb0dc563.idvisit 
```

### DIFF
--- a/core/DataAccess/LogAggregator.php
+++ b/core/DataAccess/LogAggregator.php
@@ -366,6 +366,11 @@ class LogAggregator
             $query['sql'] = 'SELECT /* ' . $this->queryOriginHint . ' */' . substr($query['sql'], strlen($select));
         }
 
+        if (!$segment->isEmpty() && is_array($query) && 0 === strpos(trim($query['sql']), $select)) {
+            $query['sql'] = trim($query['sql']);
+            $query['sql'] = 'SELECT /* ' . $this->dateStart->toString() . ',' . $this->dateEnd->toString() . ':' . implode(array_map('intval', $this->sites)). ' */' . substr($query['sql'], strlen($select));
+        }
+ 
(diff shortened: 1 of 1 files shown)


diosmosis:
I was looking at the code for the SegmentQueryDetector code and it seems like it should be called no matter what (since Segment::getSelectQuery() is called which calls getSelectQueryString() on the LogQueryBuilder. Is it not doing that currently?

tsteur:
I think the segmentQueryDecorator maybe gets only called when the `create temporary table select segmentquery` is being executed?

diosmosis:
I see, there's an `if` I didn't notice...

PR #15197, opened 2019-11-27
tracker debug fixes in CLI command + extra logs for Lock

diosmosis:
Fix tracker debug when used in a CLI command (like queuedtracking:process).


plugins/Monolog/config/tracker.php
```diff
@@ -10,6 +10,14 @@ function isTrackerDebugEnabled(ContainerInterface $c)
 
 return array(
 
+    'ini.log.log_writers' => DI\decorate(function ($previous, ContainerInterface $c) {
+        if (isTrackerDebugEnabled($c)) {
+            $previous[] = 'screen';
```
tsteur: Do we really always want to add logging to screen? I reckon many users would have an interest to only log this to file but not screen. Maybe we shouldn't force it? If someone can enable tracker debug in the config file, then they can maybe also add the logging to screen if needed?

tsteur: ping @diosmosis

diosmosis: I think this is just for CLI processes, so I'll just add a check for whether in the CLI command as well.

PR #15218, opened 2019-12-02
remove unnecessary number format in visit summary

### LINKED ISSUE #15200: Incorrect/corrupted number format in dashboards
This affects Matomo  3.13.0

Integer numbers like the total number of visitors are correctly shown in graphs, but are crippled by one place and shown with a thousand-separator(?) in the wrong place.
So 2014 visitors are shown as 2,01 vistors in the sparkline.

Please see the following screenshot:

Anmerkung 2019-11-27 120140

sgiehl:
number formatting is meanwhile done somewhere else. For some local formats this second formatting breaks the numbers...

fixes #15200


tsteur:
There seems to be some regression: https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/37206/UnprocessedSegmentTest_unprocessed_segment.png

sgiehl:
I guess that is kind of expected. The generation time is a computed metric. But when no generation time has been tracked, the metric is not computed and thus the value is actually `false`. Because of that the metric won't get formatted in https://github.com/matomo-org/matomo/blob/3f26e785f015d30d0aeea66aaf7484111b0dbfa9/core/Metrics/Formatter.php#L201-L203
But as the sparklines are using a `number` filter for the values it gets converted to `0` in https://github.com/matomo-org/matomo/blob/f1d6cc7618221813699c0f8c0d97e5c41c48d323/plugins/CoreVisualizations/templates/macros.twig#L37-L41

tsteur:
OK might need to update the tests then 👍

sgiehl:
tests are updated now

PR #15219, opened 2019-12-02
fix PNG interlacing

### LINKED ISSUE #15216: Our logo triggers lib png when sending scheduled report
> libpng warning: Interlace handling should be turned on when using png_read_image

This seems to be caused by our own logo which is included in PDF scheduled report. We can reproduce it when custom logo is disabled and not reproduce it when custom logo is enabled.

Findus23:
fixes #15216

According to https://tex.stackexchange.com/a/472462/66733 this bug is triggered in libpng when opening PNG files with [Interlacing](https://en.wikipedia.org/wiki/Interlacing_(bitmaps)).

Find all interlaced PNGs in `plugins/Morpheus/images`:
```bash
identify -verbose *.png | grep Interlace
```

Remove interlacing:
```bash
mogrify -interlace none logo*
```

Remove execution rights (while we are at it):
```bash
chmod -x logo*
```

optimize the PNG **losslessy**:
```bash
optipng -o 9 logo*.png
```

Let's hope noone has a internet connection slow enough that they can watch single pixel appear while loading a 4KB image (who would benefit from interlacing).


tsteur:
@Findus23 cheers for that. Looks like a few installation screenshots need to be updated: https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/37208/ Will you have time for this?

Findus23:
@tsteur Should be done :heavy_check_mark:

tsteur:
Cheers @Findus23

Issue #15210, opened 2019-12-01
More PHP 7.4 warnings 


Findus23:
As reported in https://forum.matomo.org/t/error-php-deprecated-auto-archiving/35190 and https://github.com/matomo-org/matomo/issues/14821#issuecomment-525816696 (the next time I'll create a sparate issue for every bug/warning).
I can't reproduce it on my empty testinstance anymore, but it still seems to happen.

```
/var/www/matomo-beta/core/DataTable/Filter/ColumnDelete.php(164): Deprecated - array_key_exists(): Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead - Matomo 3.11.0
```
https://github.com/matomo-org/matomo/blob/b16a791aa3650d85af829156129c2bd44c7cb075/core/DataTable/Filter/ColumnDelete.php#L164

tsteur:
I suppose there is no chance to get a stack trace? Any chance we can find out when this happens or on which URL this is triggered?

Ideally, for performance reasons, we would avoid needing to call `is_array|isobject` for each row there but may not be possible to workaround this. In general, we might not really support returning objects so be good to find out where this is triggered.

sgiehl:
Wouldn't it be enough to check for `isset($row[$column])` instead of the `array_key_exists`? As `$row` should implement `ArrayAccess`, `isset` should work as well and avoid the warning.

tsteur:
`isset` might just ignore null values... not sure if that's wanted?

tsteur:
Created https://github.com/matomo-org/matomo/pull/15224 for now just to be safe...

Eckybrazzz:
```
PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated in /home/nginx/domains/analytics.domain.com/public/libs/upgradephp/upgrade.php on line 180
PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated in /home/nginx/domains/analytics.domain.com/public/libs/upgradephp/upgrade.php on line 194
PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated in /home/nginx/domains/analytics.domain.com/public/libs/upgradephp/upgrade.php on line 194
PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated in /home/nginx/domains/analytics.domain.com/public/libs/upgradephp/upgrade.php on line 200
[...]

PR #15222, opened 2019-12-02
Several PHP 7.4 fixes

brainfoolong:
On a fresh 3.13.0 install with php 7.4.0 following error occurs when tracker does track a page view, when using Mysqli as database driver. 

`Notice: Trying to access array offset on value of type int in /srv/0x.at/matomo/core/Tracker/Db/Mysqli.php on line 50`

This fix this error

### DIFF
--- a/core/DataTable/Filter/ColumnDelete.php
+++ b/core/DataTable/Filter/ColumnDelete.php
@@ -161,7 +161,7 @@ class ColumnDelete extends BaseFilter
             }
             foreach ($this->columnsToRemove as $column) {
 
-                if (!array_key_exists($column, $row)) {
+                if (!property_exists($column, $row)) {
                     continue;
                 }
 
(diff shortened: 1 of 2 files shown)


core/DataTable/Filter/ColumnDelete.php
```diff
@@ -161,7 +161,7 @@ protected function removeColumnsFromTable(&$table)
             }
             foreach ($this->columnsToRemove as $column) {
 
-                if (!array_key_exists($column, $row)) {
+                if (!property_exists($column, $row)) {
```
tsteur: Could you maybe undo this change? We will investigate this as part of https://github.com/matomo-org/matomo/issues/15210

brainfoolong: Yes, done.

tsteur:
Awesome, thanks for quick change. Very appreciated.

tsteur (approved):
Need to wait for tests before merging though

PR #15223, opened 2019-12-02
Hide visitor map controls if no data is available

### LINKED ISSUE #8505: Visitor map shows info icon but no tooltip when there is no data
See title. 

Expected: Either show a tooltip or do not show an icon at all

visitormap

sgiehl:
fixes #8505


tsteur:
@sgiehl 
When I view it now, it looks like this:

image

I wonder if we should maybe add a title to the widget? Or maybe only when there is no data? Or incorporate it somehow into the no data message?

No preference. Might otherwise just be confusing that no title or indication is shown which report has no data.

tsteur (changes requested):
see comment

sgiehl:
It should now show the widget title if there is now data for the map

tsteur:
Looks good 👍  
could you check why it seems the info icon is now gone in https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/38014/VisitorMap_cities.png  and https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/38014/VisitorMap_regions.png 

Once they are fixed feel free to merge

sgiehl:
Those changes are valid. The icon shows the number of unlocated visits on hover. The icon is now hidden, if there are no unlocated visits, which is the case in the UI tests. Will update the screenshots and merge afterwards

PR #15224, opened 2019-12-02
PHP 7.4 warning in ColumnDelete

### LINKED ISSUE #15210: More PHP 7.4 warnings

tsteur:
fix https://github.com/matomo-org/matomo/issues/15210

### DIFF
--- a/core/DataTable/Filter/ColumnDelete.php
+++ b/core/DataTable/Filter/ColumnDelete.php
@@ -160,11 +160,15 @@ class ColumnDelete extends BaseFilter
                 continue;
             }
             foreach ($this->columnsToRemove as $column) {
-
-                if (!array_key_exists($column, $row)) {
-                    continue;
+                if (is_array($row )) {
+                	if (!array_key_exists($column, $row)) {
+		                continue;
+	                }
+                } else {
+                	if (!property_exists($row, $column)) {
+		                continue;
+	                }
                 }
-
                 if ($this->deleteIfZeroOnly) {
                     $value = $row[$column];
                     if ($value === false || !empty($value)) {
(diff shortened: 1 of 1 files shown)


core/DataTable/Filter/ColumnDelete.php
```diff
@@ -160,11 +160,15 @@ protected function removeColumnsFromTable(&$table)
                 continue;
             }
             foreach ($this->columnsToRemove as $column) {
-
-                if (!array_key_exists($column, $row)) {
-                    continue;
+                if (is_array($row )) {
```
Findus23: Seems like there is a space too much here:
```suggestion
                if (is_array($row)) {
```

tsteur: changed @Findus23

core/DataTable/Filter/ColumnDelete.php
```diff
@@ -160,11 +160,15 @@ protected function removeColumnsFromTable(&$table)
                 continue;
             }
             foreach ($this->columnsToRemove as $column) {
-
-                if (!array_key_exists($column, $row)) {
-                    continue;
+                if (is_object($row)) {
```
tsteur: @sgiehl I reckon it's not working maybe because we need to use indeed `isset()` or so as in the row object  it might not have the property... 
wondering if we even need this code at all as maybe unset would do the work anyway no matter if the property/key is set?

tsteur: We couldn't remove it actually.

Issue #15161, opened 2019-11-15
testcookies are always set


thisisjulien:
Hi there,
since we're using Matomo v3.12.0 `_pk_testcookies` are always set. A condition has been removed in https://github.com/matomo-org/matomo/pull/14495/files. 
This could be a GDPR issue if we don't declare them in our privacy policy. After the update we're wondering if we should edit our privacy policy's on several sites or if this get's fixed within the next update. So, is this a bug or a feature? :)
Can u maybe remove the generated cookies after doing the test or can you restore the condition?  
Thank u and best regards

Findus23:
Hi,

This sounds like https://github.com/matomo-org/matomo/issues/15000 which has been fixed in https://github.com/matomo-org/matomo/pull/15107

thisisjulien:
Hmm yes almost :D But as I understood for #15000 & #15107 the issue was that the testcookies were generated while `disableCookies` was set. We don't set this option and it happens anyway.

Findus23:
Hi,

The only way to detect if a browser is supporting cookies is to set one and check if it is set. So Matomo does need to set this cookie to work (unless you disable cookies).

tsteur:
It's actually a feature and was needed to avoid some warnings or so in IE. The cookie is immediately deleted again so it shouldn't be an issue. And it's a cookie that isn't tracking you or anything. It's just to check if cookies are enabled so reckon it shouldn't be an issue. Happy to reopen the issue if it is an issue.

thisisjulien:
Thank you for your quick response. Ok so I think we have an issue on our page, because the testcookies are not deleted and thats why I was confused.

tsteur:
The cookies are deleted but you can't see this right away. To see that they were deleted you would basically need to remove the tracking code, and then reload the page. You can otherwise also see the cookie that it has been deleted by looking at the expiry date which should be in the past.

digitalbricks:
I checked this on two sites but in may cases the expiry date of __pk_testcookie_ is updated even after calling ` _paq.push(["deleteCookies"]);`. Well, always to the current time so the expiry date is in the past but i would expect that ` _paq.push(["deleteCookies"]);` also deletes the __pk_testcookie_ – as it does with __pk_ses_ and __pk_id_. So it stays a little bit confusing to me.

tsteur:
[...]

PR #15225, opened 2019-12-02
Fix test cookie may be set multiple times

### LINKED ISSUE #15209: Too many cookies on one domain result in weird browser behaviour
Since #14495 is merged `navigatorAlias.cookieEnabled` is not utilized any more, which results in a greater number of testcookies, which leads to strange browser-problems.

I am currently using Matomo on one domain with >100 different websites `/subdivision-a`, `/subdivision-b`. Some of the websites have multiple names: `/subdivision-a` = `/alias-a`)

`setCookieDomain` is set to `*.example.org`, `setCookiePath` is set to the specific path of a subdivision or alias (`subdivision-a`, `alias-a`) 

Once a user visited too many pages and accumulated too many cookies their browser starts to act weird. Reloading the page does not work, or results in an error. (Deleting the cookies resolves the problem.)

Is there a chance that `navigatorAlias.cookieEnabled` will be used again if applicable, e.g. not on Modals in IE 11 #11507?
Or that the testcookie does not use [`getCookieName`](https://github.com/matomo-org/matomo/blob/6125a0838839b11b95fd2043cc47cd4949de766c/js/piwik.js#L3918)?
Is there a reason to use both [`configTrackerSiteId` or `domainHash`](https://github.com/matomo-org/matomo/blob/6125a0838839b11b95fd2043cc47cd4949de766c/js/piwik.js#L3907) for the testcookie?

I am aware that less testcookies don't solve my problem entirely, but reducing the number of cookies from `n` to `1` would be a big help. (If there are other solutions, I would appreciate any help.)

Thanks in advance!

tsteur:
* Make sure to not set a test cookie per site but instead use one test cookie overall
* Only set test cookie on IE or browsers that support modal dialog
* Delete the test cookie after it has been set https://github.com/matomo-org/matomo/issues/15161#issuecomment-560543852 

fix https://github.com/matomo-org/matomo/issues/15209


dirk-graetz:
_pk_test cookie..undefined is still be set

utrautmann:
Two test cookies are still set in Firefox with Matomo 3.13.3.
I do not use the disable-cookies setting, but multiple domains.
These are the activated settings:
_paq.push (["setCookieDomain", "* ..."]);
         _paq.push (["setDomains", ["* ....", "* ...."]]);
         _paq.push (["enableCrossDomainLinking"]);
         _paq.push (['trackPageView']);
         _paq.push (['enableLinkTracking']);

PR #15233, opened 2019-12-03
Fix Malformed URL in real-time-visitors module

### LINKED ISSUE #15232: Malformed URL in real-time-visitors module
While reviewing the actions section in the real-time-visitors module at the main board, I discovered the link to the downloaded content was broken. The issue there is that the generated URL for that link is not well formed. In this case, the system adds the protocol incorrectly and that makes the link to bring you nowhere.

As you can see in the attached screenshot, Matomo starts the URL with "http://https//" which takes the browser to "https//[uri-to-resource]" (missing the colon after the protocol), and that results in a broken link.

Thanks for taking care of this.

20191203_screenshot_matomo_malformed-URL

tsteur:
This issue was created in WP Matomo so was investigating and could reproduce this eg by tracking

```js
_paq.push(['trackLink', 'http://mydomain.co.uk/mailto/Agent', 'download']);
```

It adds the http:// prefix twice causing the links to not work. Maybe this was a regression from when refactoring things into visitorDetails? It's probably more a workaround since the actual issue might be that it shouldn't set the url prefix when it keeps the http url, or maybe it should remove the http in the action when it sets a url prefix... 

This might be a more proper fix but not sure if it would regress anything

```diff
--- a/app/plugins/Actions/Actions/ActionDownloadUrl.php
+++ b/app/plugins/Actions/Actions/ActionDownloadUrl.php
@@ -34,7 +34,7 @@ class ActionDownloadUrl extends Action
     {
         return array(
             // Note: we do not normalize download URL
-            'idaction_url' => array($this->getActionUrl(), $this->getActionType())
+               'idaction_url'  => $this->getUrlAndType()
         );
     }
 ```

(it would actually use the wrong type Page URL and would need to adjust this...) I reckon we better don't change this in a minor/patch release.

fix https://github.com/matomo-org/matomo/issues/15232

### DIFF
--- a/plugins/Actions/VisitorDetails.php
+++ b/plugins/Actions/VisitorDetails.php
@@ -138,10 +138,12 @@ class VisitorDetails extends VisitorDetailsAbstract
 
         // Reconstruct url from prefix
         if (array_key_exists('url', $action) && array_key_exists('url_prefix', $action)) {
-            $url = PageUrl::reconstructNormalizedUrl($action['url'], $action['url_prefix']);
-            $url = Common::unsanitizeInputValue($url);
+            if (stripos($action['url'], 'http://') !== 0 && stripos($action['url'], 'https://') !== 0) {
+                $url = PageUrl::reconstructNormalizedUrl($action['url'], $action['url_prefix']);
+                $url = Common::unsanitizeInputValue($url);
+                $action['url'] = $url;
+            }
 
-            $action['url'] = $url;
             unset($action['url_prefix']);
         }
 


diosmosis:
Should we create the original issue open or create a new one for matomo 4 to fix this in a different way later?

tsteur:
@diosmosis  created https://github.com/matomo-org/matomo/issues/15246 for now but not sure we actually need to schedule it just yet. Not sure if there's any issue with setting url_prefix and protocol. So far other things seem to work nicely.

PR #15248, opened 2019-12-08
Make BulkTracking plugin always activated

### LINKED ISSUE #14553: mark BulkTracking plugin as required, not deactivable
Well this was unexpected. Troubleshooting another problem, I decided to upgrade Matomo to the latest beta. It's not a huge installation, so I used update option inside the dashboard. Everything went smoothly, until I logged back in. Now all sites show "No data has been recorded yet".

At first I thought data was wiped. So I went through the database and it does appear data is in the database, but for some reason Matomo isn't recognizing it anymore. The table prefix is correct in the config file, piwik_

It's only tracking data that's not showing up. Websites are still listed, settings are being saved, etc.

I tried running CLI command to reprocess data, that didn't help.

Few details:
- We only import logs into Matomo. No client-side tracking.
- All system checks are good
- PHP 7.2.14
- MariaDB 10.3.12

This is a list of plugins we've deactivated:
```
Annotations
BulkTracking
Contents
CustomPiwikJs
CustomVariables
DBStats
DevicePlugins
DevicesDetection
Ecommerce
Events
ExampleAPI
ExampleCommand
ExampleLogTables
ExamplePlugin
ExampleReport
ExampleSettingsPlugin
ExampleTracker
ExampleUI
ExampleVisualization
Feedback
GeoIp2
Goals
IntranetMeasurable
Live
MobileAppMeasurable
ProfessionalServices
Provider
Resolution
SegmentEditor
TagManager
UserCountry
UserCountryMap
UserId
Widgetize
```

Here's a screenshot of what I'm seeing in database:
12155597-f373-4bc0-8a81-b75901242552

Any ideas?

katebutler:
Fixes #14553

### DIFF
--- a/core/Plugin/Manager.php
+++ b/core/Plugin/Manager.php
@@ -70,6 +70,7 @@ class Manager
 
     // These are always activated and cannot be deactivated
     protected $pluginToAlwaysActivate = array(
+        'BulkTracking',
         'CoreHome',
         'CoreUpdater',
         'CoreAdminHome',
--- a/core/Updates/4.0.0-b1.php
+++ b/core/Updates/4.0.0-b1.php
@@ -34,9 +34,12 @@ class Updates_4_0_0_b1 extends PiwikUpdates
         $migration1 = $this->migration->db->changeColumnType('log_action', 'name', 'VARCHAR(4096)');
         $migration2 = $this->migration->db->changeColumnType('log_conversion', 'url', 'VARCHAR(4096)');
 
+        $migration3 = $this->migration->plugin->activate('BulkTracking');
+
         return array(
             $migration1,
-            $migration2
+            $migration2,
+            $migration3
         );
     }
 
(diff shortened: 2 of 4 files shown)


core/Plugin/Manager.php
```diff
@@ -70,6 +70,7 @@ public static function getInstance()
 
     // These are always activated and cannot be deactivated
     protected $pluginToAlwaysActivate = array(
+        'BulkTracking',
```
sgiehl: Doesn't the plugin need to be added to the list in `global.ini.php` instead?

tsteur: It's about making sure it is always activated and not be able to be deactivated. AFAIK it is in global.ini.php already. Or not sure if you mean something else?

sgiehl: Ah sure. I misread the title :see_no_evil:

PR #15249, opened 2019-12-08
Prevent race conditions when updating plugin settings

### LINKED ISSUE #15234: WARNING: ccore/Settings/Storage/Storage.php(86): Notice - Array to string conversion
image

Noticed this in the plugin settings table. Basically there is an array stored but we expect a string or integer. Not sure how this happens as I think it does set only a string in the application. Maybe some race condition?

This breaks the archiving for us.

Looking only quickly at the code I don't really understand how this could happen: https://github.com/matomo-org/matomo/blob/3.13.0/plugins/Tour/Engagement/Challenge.php#L142-L148

katebutler:
Fixes #15234


core/Settings/Storage/Backend/PluginSettingsTable.php
```diff
@@ -48,6 +53,7 @@ public function __construct($pluginName, $userLogin)
 
         $this->pluginName = $pluginName;
         $this->userLogin = $userLogin;
+        $this->lockWrapper = new Lock(new LockBackend\MySqlLockBackend(), 'PluginSettingsTable');
```
tsteur: @katebutler as mentioned in chat we should ideally use DI here like : `StaticContainer::getContainer()->make(Lock::class, array('lockKeyStart' => 'myPrefix'));`

tsteur (changes requested):
@katebutler we'll need the changes for `MeasurableSettingsTable` as well

Issue #12519, opened 2018-01-29
Rename our composer packages


sgiehl:
Our composer packages are still listed under the vendor piwik on packagist. As there is no way to rename them, we need to create new packages and mark the old ones as abandoned

I've already started and will try to rename the other over the next days

- [x] `piwik/cache` -> `matomo/cache` (https://github.com/matomo-org/component-cache/pull/17)
  -> released as version 2.0.0; usage needs to be implemented in Matomo 4
- [x] `piwik/decompress` -> `matomo/decompress` (https://github.com/matomo-org/component-decompress/pull/12)
  -> released as version 2.0.0; usage needs to be implemented in Matomo 4
- [x] `piwik/device-detector` -> `matomo/device-detector`
- [x] `piwik/ini` -> `matomo/ini` (https://github.com/matomo-org/component-ini/pull/8)
  -> released as version 2.0.0; usage needs to be implemented in Matomo 4
- [x] `piwik/network` -> `matomo/network` (https://github.com/matomo-org/component-network/pull/14)
- [x] `piwik/piwik` -> `matomo/matomo`
   -> needs to be done in Matomo 4
- [x] `piwik/piwik-php-tracker` -> `matomo/matomo-php-tracker` (https://github.com/matomo-org/matomo-php-tracker/pull/51)
- [x] `piwik/referrer-spam-blacklist` -> `matomo/referrer-spam-blacklist` (#12518)
- [x] `piwik/searchengine-and-social-list` -> `matomo/searchengine-and-social-list` (#12518)

Renaming the php packages does only make sense with Matomo 4, so we can rename all namespaces and classes without the possibility to break any plugins. Imho it doesn't make sense to name the package `matomo/*` but the classes would still need something like `use Piwik\*`

AndoniLarz:
Hi, I'm willing to use the piwik/piwik package in my project so I'm wondering : if you rename it to matomo/matomo, will I still be able to use piwik/piwik in my composer.json or should I wait until you rename it and use matomo/matomo ?

sgiehl:
@AndoniLarz  what exactly are you going to use? It will take a while until we will rename this package, as we will likely do that in Matomo 4. As most of the remaining rename tasks will be done than. See #12420

AndoniLarz:
Well I wanted to integrate this package into my Symfony project, and I wanted to know if you were going to rename it promptly so I should wait a bit, or if you were going to let it named piwik for some time.
[...]

PR #15253, opened 2019-12-09
Use rebranded composer components

sgiehl:
refs #12519


sgiehl:
@diosmosis @tsteur @mattab Switching to the rebranded packages (with new namespaces) might maybe cause problems while updating, as some plugins might still use the old namespace until they are updated.
Any ideas how we should solve that issue? Should we update the plugins before, so they use the actual available namespace? Or shall we still include both packages in Matomo 4 (old and rebranded ones), so both namespaces are still available - and remove the old one in Matomo 4.1 or so?

tsteur:
> and remove the old one in Matomo 4.1 or so?

I reckon this won't be possible since some people will update from older 3.x or 2.X to 4.1.

Updating plugins first can cause issues too if anything fails there. 

Ideally the components would have both the new Matomo and the old Piwik classes but that might not be easily doable. Although the components typically maybe don't have too many classes so could be done maybe. Otherwise I suppose it should be fine maybe as long as the plugins don't access any component during update? They might though... also people manually updating core etc will have problems. 

Do we even need to update the components from Piwik to Matomo or wait until we do this also for core?

diosmosis:
> Ideally the components would have both the new Matomo and the old Piwik classes but that might not be easily doable. Although the components typically maybe don't have too many classes so could be done maybe.

Could this be done via inheritance? Ie:

```
namespace Piwik;

class Abc extends Matomo\Abc {
  // empty
}
```

sgiehl:
> Could this be done via inheritance? Ie:

I guess that should work. Should we add those classes to the packages or to Matomo itself?

tsteur:
Not sure what's easier?

tsteur:
I suppose we would need to make sure the autoloader still finds the files etc.

tsteur:
Very crazy thought... can the auto loader do that? Recognise Piwik namespace, then look for the matomo file that includes Matomo class, then do dynamically a `if (!class_exists($piwikNamespaceClass, $doNotLoadFile)) { class $piwikNamespaceClass extends $matomoNamespaceClass }`

Not sure if we can extend classes that way dynamically...

sgiehl:
@tsteur I'll give it a try and implement it that way if possible

sgiehl:
@tsteur found an even simpler solution using `class_alias`:

https://github.com/matomo-org/matomo/blob/42266510197ddf9ef40362fbe4d08d26721f1943/LegacyAutoloader.php#L15-L23

tsteur:
That's pretty cool 👍

sgiehl:
Should be ready for review. Remaining package is `piwik/device-detector`. But I will do some more refactorings there before releasing a new rebranded version.

LegacyAutoloader.php
```diff
@@ -0,0 +1,26 @@
...
+        if (strpos($className, 'Piwik\\') === 0) {
+            $newName = 'Matomo' . substr($className, 5);
+            if (class_exists($newName)) {
+                class_alias($newName, $className);
+            }
+        }
+    }
+}
+
+LegacyAutoloader::register();
```
tsteur: Maybe I missed it but could we add maybe a few basic tests that this works? 

In general behaviour seems otherwise good: https://3v4l.org/TAS8d

Can we assume, that if a file containing a Matomo class was loaded manually through `include/require`, then the logic would not work? (Probably not an issue but something to be aware of eg if we wanted to do that for all classes in core).

sgiehl: No, that works as well. Added some simple tests for it. See https://github.com/matomo-org/matomo/pull/15253/commits/de2fea66381aa5d5d2631bf535da68cd2b42d5cb

Issue #15205, opened 2019-11-28
spike after updating matomo to 3.13


Risingson:
![image](https://user-images.githubusercontent.com/5108701/69827062-1f2f5e80-1216-11ea-9fd7-29034869962a.png)
seems like actions of a single user are counted as unique visits

tsteur:
I would suspect this is due to https://github.com/matomo-org/matomo/pull/14360 maybe? Are you using the userId feature?

Risingson:
You are right i am using the userID and i am sure this is related. But still: Triggering an action should not count a unique visit, should it?

tsteur:
Not sure what you mean? Triggering an action should indeed not create a new unique visit and it doesn't unless your visitors were to regenerate the visitorId on every request somehow. Not sure how you embed the tracking code into your mobile app? It seems your using maybe a different visitorId each time

Risingson:
I do not intentionally create a new visitorId. I am not deep enough into this. I can provide code if you want me to.

Risingson:
These are 2 action calls a few seconds apart counting as 2 visits obviousely.
![image](https://user-images.githubusercontent.com/5108701/69833614-519c8400-1235-11ea-9513-a75b8907b04b.png)

Normally -if the system recognises the same visitorID (xor after this merge userID?)- it would look like this and count one unique visit:
![image](https://user-images.githubusercontent.com/5108701/69833649-8dcfe480-1235-11ea-9af7-d0c8de9bbcd3.png)

Risingson:
Any news on this ?

tsteur:
It really depends how you track your mobile app. Are you using one of our android or iOS SDKs? Then I would create an issue and ask there. If you implemented it yourself, then maybe have a look at the Tracking API docs: https://developer.matomo.org/api-reference/tracking-api . Maybe you need to set the `_id ` parameter

Risingson:
So this is intentional behaviour of a minor release that rips functionality provided up until 3.13. 
Nice one.

mattab:
@Risingson Are you using the official Matomo SDK? If so, it would be a bug in the SDK and it would be great if you can open a bug report there, or we can move this report to the Android SDK project?

Risingson:
I am not using the Android SDK. 
It is a mobile app using webviews (html+js>phonegap). I can send the tracking code if you wish. 
[...]

PR #15258, opened 2019-12-10
Trust cookie only when visitorId is set

tsteur:
refs https://github.com/matomo-org/matomo/issues/15205

Since we no longer use the UserId as VisitorId I reckon this needs changing. I haven't tested anything but it seems logical to me. Otherwise there might be issues where no visitorId is set but a userId and then it always creates new visits.


diosmosis:
LGTM, though there's a test failure in the PDF file. Strange it's just the PDF files...

tsteur:
@diosmosis that test is failing in some other PRs and I think 3.x-dev too

PR #15265, opened 2019-12-13
ProxySite related changes to let comparisons work properly

diosmosis:
Changes:
- Make sure ProxySite will disable post processor in Visualization where API Proxy is called directly. Otherwise the target.
- Use `Request::process()` for `SegmentEditor.getAll` calls.


tsteur:
@diosmosis looks good in general but wondering if https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/37332/UIIntegrationTest_goals_individual_goal_updated.png
and 
https://travis-ci.org/matomo-org/matomo/jobs/624430881#L992
 is maybe failing because of this change? Or another change?

diosmosis:
Yes it's failing due to `disable_datatable_post_processing` since it will get propagated to other requests. We'll probably want to get rid of this on staging.

diosmosis:
Modified the parameter to only apply to the top level API request.

diosmosis:
Tried a different approach, if the tests pass, would be useful to have on staging.

tsteur:
@diosmosis this tests is still failing and not sure it's due to this PR ? https://travis-ci.org/matomo-org/matomo/jobs/625560724#L923

tsteur:
@diosmosis the change is on staging

diosmosis:
The updated changes seem to work.

tsteur (approved):

tsteur:
Looks like tests are passing now?

BTW do you know why those tests are failing? Must be from some previous change? https://travis-ci.org/matomo-org/matomo/jobs/628335284#L1176-L1439

diosmosis:
That's from another PR, shouldn't be related to this one (plugin builds not updated).

Tests are passing, but I changed the code a bit to do that, can you take another look?

tsteur:
Looks good @diosmosis

PR #15268, opened 2019-12-14
If fingers crossed handler is used make sure to print exception traces in CLI

diosmosis:


tsteur:
Works when I add the fingers crossed handler manually in `config.php` 👍

PR #15269, opened 2019-12-14
Make segment title clickable and add segment definition to tooltip.

diosmosis:
Couple small changes after talking to @mattab


tsteur:
Are there any important fixes in there? Otherwise rather move it to 4.0 maybe.

diosmosis:
No important fixes, just seemed small enough for 3.13.1. Will move it.

plugins/CoreHome/angularjs/comparisons/comparisons.component.html
```diff
@@ -4,8 +4,8 @@ <h3>{{ 'General_Comparisons'|translate }}</h3>
     <div class="comparison card" ng-repeat="comparison in $ctrl.comparisonsService.getSegmentComparisons() track by $index">
         <div class="comparison-type">{{ 'General_Segment'|translate }}</div>
 
-        <div class="title" title="{{ comparison.title }}">
-            {{ comparison.title }}
+        <div class="title" title="{{ comparison.title }}&lt;br/&gt;{{ comparison.params.segment }}">
```
mattab: not sure, but does it maybe need some escaping for XSS, as a segment definition can contain any string?

diosmosis: angularjs does this by default, so should be ok.

tsteur: @diosmosis this is what it looks for me

image

with this segment

image
in the DB the segment is stored as `pageUrl=@b`

tsteur (changes requested):
See comment, there might be encoding issues?

sgiehl:
@diosmosis any update on this?

tsteur:
@diosmosis be good to look into this when you can

diosmosis:
@sgiehl / @tsteur pushed a fix

tsteur:
@diosmosis seems some tests are failing now because the font changes slightly?

https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/39205/Comparison_dashboard_last_period.png

https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/39205/Comparison_visitors_overview.png

Wondering if it is because there's now no newline anymore and the browser renders it slightly differently? https://github.com/matomo-org/matomo/pull/15269/files#diff-35a447f52ea628304209b9dcf28812baR8 I suppose be fine to add this newline again?

diosmosis:
@tsteur could re add it and check, or could just copy over the screenshots? I don't really have an opinion.

tsteur:
Not really. I'll test and add the spaces. Maybe it fixes the screenshots then it would avoid conflicts with other PR that updates same screenshots.

tsteur:
@diosmosis didn't seem to change anything. Seems to be maybe because of adding a title?

diosmosis:
probably, maybe the rendering changes order or something and there's some tiny subtle change. can just update the screenshots?

tsteur:
sounds good 👍

Issue #8444, opened 2015-07-28
Ensure that the archive algorithm cannot be triggered multiple times for a same site/period/segment


tsteur:
refs #1938 

I've seen this here in UI tests:

![33131fd2-24a8-11e5-9644-e01aa4b831c3](https://cloud.githubusercontent.com/assets/273120/8932062/f22888a6-353a-11e5-80eb-7d9821124d70.png)

As one can see the archive was created multiple times

quba:
But actually it should work so - e.g. for current day/week/month/year. Then temporary archives are purged once a day. Could you check the `done` flag?

Additionally sometimes it makes sense if there's some missing data. E.g. if one updates from old Piwik version (e.g. 1.6 -> latest) then it makes sense to rearchive some data (e.g. returning visitors because latest Piwik is using a segment to calculate it).

tsteur:
I do not have the database anymore. Should have made a dump :( 

It happened when starting a UI test when inserting the fixture so there should be no reason for duplicated archives

mattab:
@quba on the screenshots there are two done flags with different `idarchive` for the same idsite, same dates and period. This is a bug as Piwik should not archive twice the exact same archive.

quba:
So it's not possible when e.g. you archive two times today's period (so those will be only temporary archives)?

mattab:
> So it's not possible when e.g. you archive two times current day (so those will be only temporary archives)? 

Not sure what you meant, but what I mean is:
 * if one loads dashboard, and 10 widgets call the API at the same time (same second), we expect that Piwik will trigger archiving only once (in one thread) and other threads will wait for this process, then  when archiving is finished all widgets loading will be unblocked and will  read this same archive

quba:
Yes, that's one case. What I meant was the fact, that multiple archives for same date and period with different idarchive can be created at the same day (e.g. temporary archives created today for period=day,week,month,year). Is that correct? If yes, then the provided screenshot doesn't have to mean that there's a bug. On the other hand, those archives are for the year 2012 so it may be also connected with concurrency.

mattab:
>  What I meant was the fact, that multiple archives for same date and period with different idarchive can be created at the same day

this works well as long as you don't make concurrent requests. Piwik will only create one archive when necessary, eg. temporary archives will be re-created if the last temporary is older than N seconds (as per general settings).

>  If yes, then the provided screenshot doesn't have to mean that there's a bug.

the screenshot is clearly a bug because:
[...]

PR #15272, opened 2019-12-15
Lock when archiving and avoid invalidating sites that have archiving in progress

### LINKED ISSUE #15170: Wrong data in visitors overview over time
In Matomo 3.12.0 a bug appeared that i have described here:
https://forum.matomo.org/t/visitors-overview-not-updating/34804/7

diosmosis:
When archiving for a site begins, we create a Lock and only release when archiving finishes or fails.

The lock's expire time is updated on each query (the TTL can be configured via DI).

There is also a change to log each archiving query (previously this was just in LogAggregator, but now it will affect all archivers).

I think we should deploy this to demo and see if it works after a day or two (after getting tests to pass of course).

Note: with this implementation it's possible for archiving to start while an invalidation is occurring, which would cause issues. I'm not sure how big of a risk it is since the invalidation queries are fairly simple. There could be a few of them though.

Fixes #15170


core/ArchiveProcessor/Loader.php
```diff
@@ -73,12 +77,21 @@ private function prepareArchiveImpl($pluginName)
...
-            return $idArchive;
+        try {
+            list($visits, $visitsConverted) = $this->prepareCoreMetricsArchive($visits, $visitsConverted);
+            list($idArchive, $visits) = $this->prepareAllPluginsArchive($visits, $visitsConverted);
+
+            if ($this->isThereSomeVisits($visits) || PluginsArchiver::doesAnyPluginArchiveWithoutVisits()) {
+                return $idArchive;
+            }
+        } finally {
+            $archivingStatus->archiveFinished($lock);
```
tsteur: fyi `archiveFinished ` does not take an argument  and above the `archiveStarted ()` does not return a value

core/Archive/ArchiveInvalidator.php
```diff
@@ -206,7 +213,11 @@ public function markArchivesAsInvalidated(array $idSites, array $dates, $period,
             $periodDates = $this->getPeriodDatesByYearMonthAndPeriodType($periods);
         }
 
+        // TODO: it's possible an archiving request could begin during invalidation, though probably unlikely that it would cause true disruption... just note in issue.
+
         $periodDates = $this->getUniqueDates($periodDates);
+
+        $idSites = $this->removeSitesWithInProgressArchiving($idSites);
```
tsteur: @diosmosis not sure I understand why we would filter the idSites here and not just invalidate all of them? Is this because we assume they were already invalidated / will be invalidated? Would there be any harm in still invalidating them or the issue that this would cause the remember flag to be removed?

core/Concurrency/Lock.php
```diff
@@ -21,12 +21,19 @@ class Lock
...
-    public function __construct(LockBackend $backend, $lockKeyStart)
+    public function __construct(LockBackend $backend, $lockKeyStart, $defaultTtl = null)
     {
         $this->backend = $backend;
         $this->lockKeyStart = $lockKeyStart;
         $this->lockKey = $this->lockKeyStart;
+        $this->defaultTtl = $defaultTtl;
+    }
+
+    public function reexpireLock()
```
tsteur: should it be maybe `reExpireLock`? Not too important though...

core/DataAccess/LogAggregator.php
```diff
@@ -1160,6 +1159,9 @@ public static function makeArrayOneColumn($row, $columnName, $lookForThisPrefix
 
     public function getDb()
     {
-        return Db::getReader();
+        /** @var ArchivingStatus $archivingStatus */
+        $archivingStatus = StaticContainer::get(ArchivingStatus::class);
+        $archivingLock = $archivingStatus->getCurrentArchivingLock();
+        return new ArchivingDbAdapter(Db::getReader(), $archivingLock, $this->logger);
```
tsteur: @diosmosis is there a chance you forgot to commit this file?

core/Archive/ArchiveInvalidator.php
tsteur: I wonder maybe we could still invalidate archives for the other idSites as well just not remove the remember flag?

tsteur (changes requested):
@diosmosis left few comments.  Be good in general to test this on demo for few days.

core/Archive/ArchiveInvalidator.php
diosmosis: We remove them because we know they are currently being archived. We also keep the idSites in the remember flags so they will eventually be invalidated when they are not being archiving.

core/Concurrency/Lock.php
diosmosis: Yeah, it looks weird to me too. But so did `reExpireLock` and I couldn't choose :)

core/DataAccess/LogAggregator.php
diosmosis: a chance? no... I definitely forgot to commit it :) added now

core/ArchiveProcessor/Loader.php
diosmosis: fixed

diosmosis:
I'll add this to demo soon

core/Archive/ArchiveInvalidator.php
tsteur: Just wondering if there could be meanwhile some other previously finished archives for those idSites that need to be invalidated?

diosmosis: There could be, but because of the way the function parameters are (idSites + list of dates + period type), it's a bit harder to control exactly what to invalidate. We'd have to loop over idSites & dates and check each individually, and invalidate each archive individually instead of with one query (which is what we do now I think).

tsteur: Not sure I understand... it would be safe to invalidate all the data related to that site/period and all parent periods as long as we keep the remember flag? Or would there be an issue?

diosmosis: The issue is if we invalidate while archiving is ongoing, then it's possible for archives to re-initiate archiving for child periods re-doing archiving unnecessarily.

Eg, if the week period initiates archiving and finishes, then a day in the week is invalidated, and the month period starts archiving, it will re-archive the month and the day. Since we invalidate on new visits, this means for high traffic sites we'll always re-archive these subperiods as soon as an archive request goes through (since invalidation is not just done in core:archive but in Archive.php).

tsteur: Isn't that though what we still want it to behave? Whether it will be rearchived right away depends on `time_before_today_archive_considered_outdated` and `time_before_week_archive_considered_outdated` etc in that case or does it behave differently? I'm not sure how it behaves when data is invalidated... whether it still takes those values into consideration

diosmosis: That's used by `getMinTimeProcessedForTemporaryArchive()` which is no longer used because we got rid of temporary archives... I guess we can re-add that logic?

diosmosis: Incidentally I don't see `time_before_week_archive_considered_outdated` being used anywhere.

diosmosis: @tsteur Ok I found where the period configs are used but they're not used anymore either because of the temporary archive change. This is likely the cause of the bug, and probably a less overarching change.

However, the lock does allow telling if a specific archive is being built (well not quite, but we can modify it if we increase the lock key column length). What do you think should happen? Make sure the time_before_ INI config is used now and keep the locks but don't do anything different in ArchiveInvalidator?

tsteur: @diosmosis it is very important we respect these configured values. We basically need to check if the last archive was done more than the configured time ago even if the archive is invalid (unless the archive is invalid and it is a new day I think then we don't respect these values but not sure? @mattab )

> and keep the locks but don't do anything different in ArchiveInvalidator?
I'm not quite fully understand this part but be good in general to maybe still invalidate an archive in ArchiveInvalidator? Not sure of consequences of not using Locks?

diosmosis: > Not sure of consequences of not using Locks?

It's just an easy way to check if an archive is on-going no matter what machine/process it's on. It's not really necessary, but it might be useful in the future.

tsteur: I reckon be good to keep it 👍

mattab: > It's just an easy way to check if an archive is on-going no matter what machine/process

@diosmosis it would be big value if it would help with this issue: https://github.com/matomo-org/matomo/issues/8444 -> maybe it would be fixed?

> unless the archive is invalid and it is a new day I think then we don't respect these values but not sure? @mattab

correct, these INI settings are only used when the date range includes Today.

tsteur: just wanting to make sure we're not going to work on #8444 as part of this issue :)

diosmosis:
@tsteur added back old functions and added a test for the Loader method that uses them

tsteur:
@diosmosis maybe we can run this code on demo for a while?

diosmosis:
@tsteur it should be running on demo2

tsteur:
@diosmosis do you know of any issue on the demo? Maybe we could merge and release a new beta?

diosmosis:
The metrics for the last two week seem accurate, I'm not aware of any issues (though I don't get emails if demo2 has an issue).

mattab:
There was no error from demo2 in the last few days. @tsteur @diosmosis can you merge this and let me know, i'll release the next beta?

PR #15275, opened 2019-12-16
Force using DB sessions once Matomo is installed

tsteur:


core/Session.php
```diff
@@ -101,10 +101,7 @@ public static function start($options = false)
 
             @ini_set('session.save_handler', 'files');
             @ini_set('session.save_path', $sessionPath);
-        } elseif (self::isSessionHandler('dbtable')
-            || self::isSessionHandler('files')
```
sgiehl: the method `isSessionHandler` is now unused and can be removed.
Also it might make sense to remove the config `session_save_handler` completely, as it has no effect anymore.

sgiehl (changes requested):

PR #15276, opened 2019-12-16
Fix mysql table name too long while archiving a segment

tsteur:
fix error 
> WP DB Error: Identifier name 'wp_foooobarrrrrr_matomo_logtmpsegment07171d02820b83eab14316299645c2e9' is too long SQL: CREATE TEMPORARY TABLE wp_foooobarrrrrr_matomo_logtmpsegment07171d02820b83eab14316299645c2e9 (idvisit BIGINT(10) UNSIGNED NOT NULL)

### DIFF
--- a/core/DataAccess/LogAggregator.php
+++ b/core/DataAccess/LogAggregator.php
@@ -195,7 +195,8 @@ class LogAggregator
     private function getSegmentTmpTableName()
     {
         $bind = $this->getGeneralQueryBindParams();
-        return self::LOG_TABLE_SEGMENT_TEMPORARY_PREFIX . md5(json_encode($bind) . $this->segment->getString());
+        $tableName = self::LOG_TABLE_SEGMENT_TEMPORARY_PREFIX . md5(json_encode($bind) . $this->segment->getString());
+        return Common::mb_substr($tableName, 0, 64);
     }
 
     public function cleanup()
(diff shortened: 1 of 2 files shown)


core/DataAccess/LogAggregator.php
```diff
@@ -195,7 +195,8 @@ public function setQueryOriginHint($nameOfOrigiin)
     private function getSegmentTmpTableName()
     {
         $bind = $this->getGeneralQueryBindParams();
-        return self::LOG_TABLE_SEGMENT_TEMPORARY_PREFIX . md5(json_encode($bind) . $this->segment->getString());
+        $tableName = self::LOG_TABLE_SEGMENT_TEMPORARY_PREFIX . md5(json_encode($bind) . $this->segment->getString());
+        return Common::mb_substr($tableName, 0, 64);
```
diosmosis: Can we make 64 a constant? We could also shorten `self::LOG_TABLE_SEGMENT_TEMPORARY_PREFIX` since the table names are about 5 characters too long it seems.

tsteur: Pushed something, is that what you mean?

diosmosis: Yep, was also wondering if shortening `self::LOG_TABLE_SEGMENT_TEMPORARY_PREFIX` was an option since `wp_foooobarrrrrr_matomo_logtmpsegment07171d02820b83eab14316299645c2e9` was 69 characters.

tsteur: @diosmosis now I understand what you mean. I guess it's not really needed I would say as the next one might again have an even longer prefix etc. Even if the table name is truncated by 16 characters it would still not be an issue since a temporary table only exists for one session and is not visible to any other mysql session... and after an archive the table is being deleted so it's unlikely there will two tables having the same name... we could shorten it though to `logtmpseg` if that helps and still clear what it is about?

diosmosis: I guess it's not really needed. Will merge.

PR #15277, opened 2019-12-17
Remove output that can be ignored in climulti

### LINKED ISSUE #982: Process Isolation and phpunit.phar install
While using phpunit with the phar distribution, some project cannot resolve PHPUnit/Autoload.php file as the file does not exist in the include path.

Here an error:

```
The error is 9) JMS\SecurityExtraBundle\Tests\Functional\TwigIntegrationTest::testIsExprGranted
PHPUnit_Framework_Exception: PHP Warning:  require(PHPUnit/Autoload.php): failed to open stream: No such file or directory in - on line 9
Warning: require(PHPUnit/Autoload.php): failed to open stream: No such file or directory in - on line 9
PHP Fatal error:  require(): Failed opening required 'PHPUnit/Autoload.php' (include_path='.:/usr/share/php:/usr/bin/phpunit') in - on line 9
Fatal error: require(): Failed opening required 'PHPUnit/Autoload.php' (include_path='.:/usr/share/php:/usr/bin/phpunit') in - on line 9

```

The test is using 

```
        $client = $this->createClient(array('config' => 'all_voters_disabled.yml'));
        $client->insulate();
```

The insulate method make the Client (Symfony2) create a dedicated process to run the test, which fail to include the `PHPUnit/Autoload.php` file.

I tried different solutions with the default include path: (include_path='.:/usr/share/php:phar:///usr/bin/phpunit/') ... but with no luck ...

Does it mean there is no solution if a phar cannot be in a include_path ? So the only valid installation solution for PHPUnit is by using the Pear install.

tsteur:
Not sure it's really needed and how it happens but I reckon it cannot hurt. We've seen some archiving errors in Matomo for WordPress when CLI archiving is used where the output starts with the console hashbang ... 

Seeing this also in PHPUnit in https://github.com/sebastianbergmann/phpunit/blob/8.5.0/src/Util/PHP/AbstractPhpProcess.php#L268-L273 which was added to fix https://github.com/sebastianbergmann/phpunit/issues/982 but the change doesn't seem related to that.

### DIFF
--- a/core/CliMulti/Output.php
+++ b/core/CliMulti/Output.php
@@ -8,6 +8,7 @@
 namespace Piwik\CliMulti;
 
 use Piwik\CliMulti;
+use Piwik\Common;
 use Piwik\Filesystem;
 
 class Output
@@ -58,7 +59,14 @@ class Output
 
     public function get()
     {
-        return @file_get_contents($this->tmpFile);
+        $content = @file_get_contents($this->tmpFile);
+        $search = '#!/usr/bin/env php';
+        if (!empty($content)
+            && is_string($content)
+            && Common::mb_substr(trim($content), 0, strlen($search)) === $search) {
+            $content = trim(Common::mb_substr(trim($content), strlen($search)));
+        }
+        return $content;
     }
 
     public function destroy()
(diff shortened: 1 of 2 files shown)


tsteur:
See https://github.com/matomo-org/matomo/pull/15285 this could for example happen when using litespeed by the looks

PR #15283, opened 2019-12-18
Improve detection of cgi in cli multi

tsteur:
Not sure if it's possible but maybe a regular request is not using CGI, but then the CLI request is if `findPhpBinary` returns a cgi script?

See https://www.php.net/manual/en/function.shell-exec.php
In Matomo for WordPress seeing various reports where the output starts with `#!/usr/bin/env php`. 

image

refs https://github.com/matomo-org/matomo/pull/15277
refs https://github.com/matomo-org/wp-matomo/issues/135#issuecomment-566766633

Trying to reproduce this locally this happens:

```bash
php -r "echo shell_exec(\"php-cgi app/console climulti:request -q  'module=API&method=API.get&idSite=1&period=day&date=last19&format=php&trigger=archivephp' --superuser\");"
X-Powered-By: PHP/7.3.12
Content-type: text/html; charset=UTF-8
```

Where it prints x-powered by etc

### DIFF
--- a/core/CliMulti.php
+++ b/core/CliMulti.php
@@ -239,7 +239,8 @@ class CliMulti
      */
     public function supportsAsync()
     {
-        $supportsAsync = Process::isSupported() && !Common::isPhpCgiType() && $this->findPhpBinary();
+        $phpBinary = $this->findPhpBinary();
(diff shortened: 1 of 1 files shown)


tsteur:
just seeing this should be already done in https://github.com/matomo-org/matomo/blob/3.x-dev/core/CliMulti/CliPhp.php#L79-L84 ... but it may not be done for PHP binary... will adjust PR

PR #15285, opened 2019-12-18
Fix cli archiving might not work when using litespeed

tsteur:
replaces https://github.com/matomo-org/matomo/pull/15283

Noticed the user's instance was using listespeed https://www.php.net/manual/en/install.unix.litespeed.php which returned following

``` 
phpbinary: /usr/local/bin/lsphp
phpbindir: /opt/alt/php72/usr/bin
which php: /usr/local/bin/php
```

We would have used lsphp but this seems to print the hashbang and have other issues. see #15283

### DIFF
--- a/core/CliMulti/CliPhp.php
+++ b/core/CliMulti/CliPhp.php
@@ -80,7 +80,8 @@ class CliPhp
         return !empty($path)
         && false === strpos($path, 'fpm')
         && false === strpos($path, 'cgi')
-        && false === strpos($path, 'phpunit');
+        && false === strpos($path, 'phpunit')
+        && false === strpos($path, 'lsphp');
     }
 
     private function getPhpCommandIfValid($path)


skizzerz:
This change **broke** archiving on my matomo instance, because the path to my cli version of php is `/usr/local/lsws/lsphp72/bin/php`.

You really shouldn't be trying to guess what flavor of PHP is being used by testing what strings appear in the path. Instead, run `/path/to/php -r 'echo php_sapi_name();'` and check that output. If it says "cli" then you're good to go. If it says anything else (such as "cgi" or "lsphp") then it's the wrong binary.

For a common-case fast-path, you can run `php_sapi_name()` directly in the `PHP_BINARY` branch to see if the currently-running PHP is cli or not. If it is, you can use that without needing to call to shell_exec (which may be important for some customers that have the function disabled).

tsteur:
cheers @skizzerz created https://github.com/matomo-org/matomo/pull/15611

PR #15286, opened 2019-12-18
Refactor search category/count to not use custom variables

### LINKED ISSUE #12072: Site Search tracking should not use Custom Variables
Currently, our Site Search feature uses Custom Variables in the backend to track site searches and site search categories. We need to rewrite this feature without using Custom Variables (see https://github.com/piwik/piwik/issues/11524) so that we can deprecate Custom Variables in the future.

Since it would likely require new fields in the database, assigning to Piwik 4.

katebutler:
Fixes #12072


sgiehl:
Guess Kate won't finish this PR, so I will take over...

sgiehl:
I've updated the PR and fixed some obvious bugs, adjusted the tests and also made search category and count available as segments

Should be ready for a review now

core/Columns/Discriminator.php
```diff
@@ -46,7 +46,7 @@ public function __construct($table, $discriminatorColumn = null, $discriminatorV
     public function isValid()
     {
         return isset($this->discriminatorColumn)
-            && (is_numeric($this->discriminatorValue) || $this->discriminatorValue == ActionSiteSearch::CVAR_KEY_SEARCH_CATEGORY);
+            && (is_numeric($this->discriminatorValue));
```
diosmosis: minor: the parentheses isn't needed anymore

plugins/Actions/Archiver.php
```diff
@@ -165,6 +166,25 @@ protected function archiveDaySiteSearchActions($rankingQueryLimit)
         }
     }
 
+    protected function archiveDaySearchCategoryActions()
+    {
+        $where = "%s.search_cat != ''";
```
diosmosis: should we also check if `search_cat IS NOT NULL`?

plugins/CustomVariables/VisitorDetails.php
```diff
@@ -41,7 +41,11 @@ public function extendActionDetails(&$action, $nextAction, $visitorDetails)
         for ($i = 1; $i <= $maxCustomVariables; $i++) {
             if (!empty($action['custom_var_k' . $i])) {
                 $cvarKey                 = $action['custom_var_k' . $i];
-                $cvarKey                 = static::getCustomVariablePrettyKey($cvarKey);
+
+                if (in_array($cvarKey, ['_pk_scat', '_pk_scount'])) {
+                    continue; // ignore old site search variables
+                }
```
diosmosis: Is it currently supported to send these custom variables manually in tracking requests? Would this break that use case?

tests/PHPUnit/System/expected/test_ManyVisitorsOneWebsiteTest__Live.getLastVisitsDetails_month.xml
```diff
@@ -964,18 +960,14 @@
...
 				<title>Site Search</title>
-				<subtitle>Bring on the party</subtitle>
+				<subtitle>Bring on the party - Search Category: CAT - Search Results Count: 36</subtitle>
 				
-				<customVariables>
-					<row>
-						<customVariablePageName4>Search Category</customVariablePageName4>
-						<customVariablePageValue4>CAT</customVariablePageValue4>
-					</row>
-				</customVariables>
```
diosmosis: This removal from the API is a breaking change, correct? Perhaps we should mention it in the CHANGELOG?

diosmosis (commented):
Left a couple comments, code looks good and tests pass, can be merged after comments dealt w/.

core/Columns/Discriminator.php
sgiehl: changed

plugins/Actions/Archiver.php
sgiehl: makes sense as the column is nullable

tests/PHPUnit/System/expected/test_ManyVisitorsOneWebsiteTest__Live.getLastVisitsDetails_month.xml
sgiehl: Will add a changelog entry

plugins/CustomVariables/VisitorDetails.php
sgiehl: Actually it seems it wasn't restricted to send those variables manually. Not sure if that is something we should handle 🤔 Don't think someone used those variables accidentally.

diosmosis: I guess our options are:

* just stop handling them
* report a tracking error if found so users can fix their tracking code
* keep handling them in some way (ie, not recording as custom var, but as search dimension value)

What do you think?

cc @tsteur

tsteur: I reckon we can keep the logic there as is and doubt indeed someone uses these values. And in worst case these values would just not be visible in the visitor log.

diosmosis:
There are some tests that are failing, including one UI test that looks non-trivial to fix, otherwise looks good to merge to me.

sgiehl:
@diosmosis the failing plugin tests simply needs to be update in the submodule. Would do that before merge. But which UI test are you referring to? I think all of those failing here might also fail on 4.x-dev. (some of them started failing after merging 3.x-dev to 4.x-dev)

diosmosis:
@sgiehl there was an actions report that was failing but I don't see it anymore, I'm not sure what happened there. I guess ignore my comment.

PR #15294, opened 2019-12-20
MergeDataTables does not correctly merge if table2 has more data than table1

diosmosis:
... & fix build.

When fixing tests I noticed a bug exposed by the new VisitFrequency metrics. MergeDataTables will only iterate over the first table when merging tables, which means if there are tables that exist in table2 only, they won't get added.


plugins/API/DataTable/MergeDataTables.php
```diff
@@ -52,4 +52,23 @@ public function mergeDataTables($table1, $table2)
         }
     }
 
+    private function makeNewDataTable(DataTable\DataTableInterface $subTable2)
+    {
+        if ($subTable2 instanceof DataTable\Map) {
```
tsteur: @diosmosis wondering if we would need to create a clone and keep any filters? Not sure though

plugins/API/DataTable/MergeDataTables.php
```diff
@@ -25,13 +24,14 @@ public function mergeDataTables($table1, $table2)
...
         if ($table1 instanceof DataTable\Map && $table2 instanceof DataTable\Map) {
-            $subTables2 = $table2->getDataTables();
-            foreach ($table1->getDataTables() as $index => $subTable1) {
-                if (!array_key_exists($index, $subTables2)) {
-                    // occurs when archiving starts on dayN and continues into dayN+1, see https://github.com/piwik/piwik/issues/5168#issuecomment-50959925
-                    continue;
+            $subTables1 = $table1->getDataTables();
+            foreach ($table2->getDataTables() as $index => $subTable2) {
+                if (!array_key_exists($index, $subTables1)) {
+                    $subTable1 = $this->makeNewDataTable($subTable2);
```
tsteur: @diosmosis reading the previous comment in  https://github.com/piwik/piwik/issues/5168#issuecomment-50959925 I wonder if this could cause any issues with potentially wrong data when archiving overlaps another day basically? Not sure I fully understand #5168 .  I suppose if it was a problem, it would need to be handled somewhere else maybe and not in this class as when merging you expect both tables to be merged as it is implemented in this PR

plugins/API/DataTable/MergeDataTables.php
diosmosis: I guess we could. Might be an issue if filters hold references to the old table or to some other object that isn't relevant anymore (ie, like ones that calculate percent over a total or specific column value in another table or something like that). Actually, it's probably better to require this class be used before filters are run, since if metrics are formatted they won't be able to be added properly...

plugins/API/DataTable/MergeDataTables.php
diosmosis: So:

- archive starts on one day sending request for 'today'
- new day begins, archive finishes, 'today' value changes

then some bug happens? I'm not sure I understand either.

tsteur: maybe @mattab knows?

plugins/API/DataTable/MergeDataTables.php
tsteur: It shouldn't hold a reference to the table but possibly to some other parameters.

plugins/API/DataTable/MergeDataTables.php
mattab: as you said, "if it was a problem, it would need to be handled somewhere else maybe and not in this class as when merging you expect both tables to be merged as it is implemented in this PR" -> so sounds good to proceed with this proper fix and we can see if the problem in #5168 starts appearing again

mattab:
Would it be possible adding some unit tests or so?

plugins/API/DataTable/MergeDataTables.php
diosmosis: Some filters do hold references to other tables, eg, CalculateEvolutionFilter. It is conceivable that they could hold onto a lot of data.

Ideally it would be great to deprecate queued filters and instead move that logic to a method in the Report class, like, `formatApiResponse()`, similar to `configureView()` or whatever it's called.

diosmosis:
👍 can add unit tests

mattab:
@diosmosis looks like you forgot to add the unit tests in https://github.com/matomo-org/matomo/pull/15294/commits/1c83ed2ed1f8b5a6b89987aa70b7a579f83f3b3d ?

diosmosis:
argh

diosmosis:
@mattab added in https://github.com/matomo-org/matomo/pull/15323

PR #15296, opened 2019-12-20
Make sure unprocessed segment notification is still shown and include notifications in UI test.

diosmosis:
Looks like the unprocessed segment notification has been broken for a bit. This should fix it.


tsteur:
LGTM. Not sure if any of the UI test failures is maybe due to this? https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/37434/UIIntegrationTest_visitors_overview_segment.png 

Feel free to merge

diosmosis:
That one's a travis-ci issue, same thing that happens to the system test build. I did some digging a while back and it happens to another project on php 5, but not php 7.

PR #15304, opened 2019-12-22
Compute row percentages in PHP before metrics are formatted to avoid percentage formatting WARNINGS.

### LINKED ISSUE #14540: warning Piwik.php(122): Warning - A non-numeric value encountered
WARNING: www/demo.piwik.org/core/Piwik.php(122): Warning - A non-numeric value encountered - Matomo 3.10.0-rc1 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)

diosmosis:
Fixes #14540


tsteur:
@diosmosis where can this be reproduced the previous issue?

diosmosis:
@tsteur open any product report in the UI, eg, https://demo.matomo.org/index.php?module=CoreHome&action=index&idSite=62&period=day&date=yesterday#?idSite=62&period=day&date=yesterday&segment=&category=Goals_Ecommerce&subcategory=Goals_Products , then look in the settings page for the warning notifications.

plugins/CoreVisualizations/templates/_dataTableViz_htmlTable_ratio.twig
```diff
@@ -1,20 +1,19 @@
 {% if column in properties.report_ratio_columns and (column in totals|keys or forceZero|default) -%}
     {% set reportTotal   = totals[column]|default(0) %}
-    {% if siteTotalRow|default is not empty %}
+
+    {% if siteTotalRow is defined and siteTotalRow is not empty %}
```
tsteur: @diosmosis  do you understand why we calculate the percentage differently 
* The percentage value we actually show is based on reportTotal
* The percentage value in tooltip is based on siteTotal

I wonder if we can't always use the same logic? Eg always use reportTotal? Then we wouldn't even need to do the API call in the HTML table maybe. I suppose the reportTotal may be formatted as well though. Actually I think I get it... the siteTotal is without segment while reportTotal may have a segment...

Looks good to merge @diosmosis

tsteur (approved):

plugins/CoreVisualizations/templates/_dataTableViz_htmlTable_ratio.twig
diosmosis: > Actually I think I get it... the siteTotal is without segment while reportTotal may have a segment...

Interesting, I had no idea why this was there :)

tsteur: Actually, I just tested it and it does actually not seem to be the case. I reckon the `siteSummary` might not be needed at all and we can just use the same percentage as we show in the table based on reportTotal?

diosmosis: Looks like this is the commit that differentiated between totals & siteSummary: https://github.com/matomo-org/matomo/commit/bb0e02d21541125eca23ba0256b3a3cde3db262c

Maybe `siteSummary` was supposed to get the data w/o segment, but because we weren't using `Request::processRequest(..., $default = [])` or something similar, it wasn't working as expected? I wonder if this is actually a bug and we can fix it by making siteSummary work the way you thought it worked.

tsteur: Sounds great to add the default 👍

tsteur:
LGTM @diosmosis seems like some UI tests maybe need to be updated as for some reason the hidden ratio maybe is taking more space? Maybe those tests are failing though also in 3.x... I'll restart the PR travis jobs

diosmosis:
@tsteur looks like the precision changed: https://builds-artifacts.matomo.org/matomo-org/matomo/ratio-percent-unformatted/37553/ViewDataTableTest_14_visits_percent.png

odd... will look into it

tsteur:
Funny indeed since I thought this percentage was only used for the title but not for the shown value

PR #15309, opened 2019-12-22
Ignore disableCookie setting when giving consent / removing consent

tsteur:
Will be needed for https://github.com/matomo-org/matomo/pull/15184

No preference if we merge or not. I  would say it makes sense to merge though since it is in the interest of the user to know whether the user has opted out or removed consent etc. The cookie itself is just a 0/1 and can't identify anyone particularly and it is in the clear interest of the user to persist this cookie I would say. It is pretty much a preference cookie.

Any thoughts @mattab @Findus23


mattab:
+1 to merge it

Note, we would also need to update the faq explaining how to disable cookies to mention that this cookie will not be disabled and why.

mattab:
added some doc to https://matomo.org/faq/general/faq_157/ and https://matomo.org/faq/general/faq_146/

PR #15312, opened 2019-12-23
Fix log tmp segment may be too long

tsteur:
Noticed https://github.com/matomo-org/matomo/pull/15276 does not actually work since it doesn't take into consideration the DB prefix.


mattab:
@tsteur LGTM. Although did you mean to include the 2fa improvements in the PR?

 Looking forward to having this in a new release :+1:

Issue #15308, opened 2019-12-22
MaxMind GeoIP DB no longer available / license changes etc


tsteur:
see https://blog.maxmind.com/2019/12/18/significant-changes-to-accessing-and-using-geolite2-databases/

* This is valid from December 30, 2019 even though it was only announced two days ago (and that during the season where most developers are on holidays)
* One can't simply download the free GeoIP databases anymore, instead one has to register an account
* The file is still free of charge
* But they also use this change to sneakily change the license of the database from the Creative Commons Attribution-ShareAlike to their own end-user license agreement
*  This end-user license agreement is not even available yet, but they are estimating they will publish it on December 23.
* They reason that they need to collect the user data of all of their users to comply to privacy laws (especially the new Californian one).

* What does this mean for Matomo:
The normal GeoIP setup will stop working for new users in a week. For existing Matomo users it will continue to work, but will get out of data.
* What they don't mention: They can't relicense the existing files, so we can continue to distribute the last database before the change (even though it will become out of date and distributing it might be a challenge)

I reckon as a first step we could temporarily host the latest version of the DB on builds.matomo.org so at least existing downloads won't fail? 

Then start implementing an alternative asap? Like eg https://db-ip.com/db/download/ip-to-city-lite 
Are there any other suggestions?

mattab:
Other suggestions:

* write a blog post explaining: the situation and what it means for Matomo users (geolocation will get less and less accurate over time if they don't upgrade Matomo), whether people need to do anything, what we're doing to address it, when will it be addressed roughly
* look into db-ip.com but also this other DB provider: https://lite.ip2location.com - the company and people behind the service have actually built the Matomo plugin themselves https://plugins.matomo.org/IP2Location
* Shall we release 3.13.1 with the updated URLs very soon then?

tsteur:
I suppose we don't want to host the DB ourselves to be GDPR compliant etc for the same reason why MaxMind no longer provides the download directly.

mattab:
> I suppose we don't want to host the DB ourselves to be GDPR compliant etc for the same reason why MaxMind no longer provides the download directly.

I think we wouldn't want to host it too long, but it would be fine to do it for a few weeks or as long as it takes for us to have an alternative solution eg. ip-db or ip2location? 

fyi also asked in https://github.com/maxmind/geoipupdate/issues/61 how it would affect the debian package

diosmosis:
[...]

Issue #15317, opened 2019-12-24
GeoIP2 City South America "Continent" reported as Broken


aabenitez:
Error: Matomo tried to checking the location of a known IP address (194.57.91.215), but your server returned 'Unknown, Unknown, FR'. If this provider was configured correctly, it should return 'Besançon, BFC, FR'.

Matomo version: 3.13.0
MaxMind DB: GeoIP2-City-South-America.mmdb 
(Cities, regions and countries for Ip's of south america; Countries data for Ip's of other continents)

Why it tries to check with a France city IP if I installed a South America City database?

Best Regards.

![geoip2_south_america_db_matomo_broken](https://user-images.githubusercontent.com/7979587/71419611-108a6a80-264f-11ea-83e3-1419434217d1.PNG)
![XML_api_geoip_south_america_db](https://user-images.githubusercontent.com/7979587/71419614-1718e200-264f-11ea-926c-23a547c22a52.PNG)

sgiehl:
Thanks for the report. It's right, we are currently checking a France IP only. Something we need to change for databases not containing Europe

aabenitez:
Ok thank you, It Will be great!

sgiehl:
@aabenitez  does the problem still exist with the latest release? I don't have access to the continent databases to test that...

aabenitez:
Hello, I will test..

aabenitez:
Now it recognizes my continent maxmind database.. Thank you!

Alcides B.

PR #15319, opened 2019-12-27
Allow using and auto-updating dbip databases and default to db-ip.com lite in the UI

diosmosis:
refs https://github.com/matomo-org/matomo/issues/15308


sgiehl:
@mattab could you maybe get in touch with DB-IP, to get access to their paid databases? maybe also the one including ISP data? Would be good if we could test those as well.

plugins/GeoIp2/LocationProvider/GeoIp2.php
```diff
@@ -38,10 +39,19 @@ abstract class GeoIp2 extends LocationProvider
      * @var array
      */
     public static $dbNames = array(
-        'loc' => array('GeoIP2-City.mmdb', 'GeoIP2-City-Africa.mmdb', 'GeoIP2-City-Asia-Pacific.mmdb', 'GeoIP2-City-Europe.mmdb', 'GeoIP2-City-North-America.mmdb', 'GeoIP2-City-South-America.mmdb', 'GeoIP2-Enterprise.mmdb', 'GeoIP2-Country.mmdb', 'GeoLite2-City.mmdb', 'GeoLite2-Country.mmdb'),
+        'loc' => array('GeoIP2-City.mmdb', 'DBIP-City.mmdb', 'DBIP-City-Lite.mmdb', 'DBIP-Country-Lite.mmdb', 'DBIP-Country.mmdb',
+            'dbip-city-lite-\d{4}-\d{2}.mmdb', 'GeoIP2-City-Africa.mmdb', 'GeoIP2-City-Asia-Pacific.mmdb', 'GeoIP2-City-Europe.mmdb',
+            'GeoIP2-City-North-America.mmdb', 'GeoIP2-City-South-America.mmdb', 'GeoIP2-Enterprise.mmdb', 'GeoIP2-Country.mmdb',
+            'dbip-country-lite-\d{4}-\d{2}.mmdb', 'GeoLite2-City.mmdb', 'GeoLite2-Country.mmdb'),
```
sgiehl: We might need to add the file names for the paid databases here. not sure if the include the date as well...

tsteur:
BTW haven't looked at the code but do we want to auto update the URL for people using the free maxmind DB?

sgiehl:
guess it would make sense to do that in an update script

plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -186,8 +200,10 @@ public static function unzipDownloadedFile($path, $dbType, $unlink = false)
             $fileToExtract = null;
             foreach ($content as $info) {
                 $archivedPath = $info['filename'];
-                if (in_array(basename($archivedPath), LocationProviderGeoIp2::$dbNames[$dbType])) {
-                    $fileToExtract = $archivedPath;
+                foreach (LocationProviderGeoIp2::$dbNames[$dbType] as $dbName) {
+                    if (basename($archivedPath) === $dbName || preg_match($dbName, basename($archivedPath))) {
```
sgiehl: Running the updater this throws a warning:
```
WARNING [2019-12-29 20:09:57] 31960  /srv/matomo/plugins/GeoIp2/GeoIP2AutoUpdater.php(204): Warning - preg_match(): Delimiter must not be alphanumeric or backslash
```

plugins/GeoIp2/LocationProvider/GeoIp2.php
diosmosis: 👍 not sure what they are actually named. I'll wait for @mattab's input.

mattab:
> could you maybe get in touch with DB-IP, to get access to their paid databases? maybe also the one including ISP data? Would be good if we could test those as well.

Just requested to db-ip team if they're willing to send us a commercial DB set, will let you know when we hear back. Btw i noticed the commercial DBs are quite large, see:
Screenshot from 2019-12-30 09-49-09

Would this still work fine even though the DB can be up to 4DB large?

diosmosis:
Looks like those are for the CSV files, the mmdb files are smaller. Though they still seem to be quite large (~350mb for city db). The download might take a while but for once a month it should be ok?

plugins/GeoIp2/GeoIP2AutoUpdater.php
diosmosis: Should be fixed

diosmosis:
Updated the PR

plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -186,8 +200,12 @@ public static function unzipDownloadedFile($path, $dbType, $unlink = false)
             $fileToExtract = null;
             foreach ($content as $info) {
                 $archivedPath = $info['filename'];
-                if (in_array(basename($archivedPath), LocationProviderGeoIp2::$dbNames[$dbType])) {
-                    $fileToExtract = $archivedPath;
+                foreach (LocationProviderGeoIp2::$dbNames[$dbType] as $dbName) {
+                    if (basename($archivedPath) === $dbName
+                        || preg_match('/' . $dbName . '/', basename($archivedPath))
```
tsteur: You might want to use `preg_quote($dbName, '/')  ` just to be safe?

plugins/GeoIp2/LocationProvider/GeoIp2.php
```diff
@@ -178,7 +189,7 @@ public static function getGeoIPDatabaseTypeFromFilename($filename)
     {
         foreach (self::$dbNames as $key => $names) {
             foreach ($names as $name) {
-                if ($name === $filename) {
+                if ($name === $filename || preg_match('/'.$name.'/', $filename)) {
```
tsteur: Same here re `preg_quote`... probably not needed but may be good to have

plugins/GeoIp2/GeoIP2AutoUpdater.php
diosmosis: It's not supposed to have preg_quote, the values in `$dbNames` are either the file name or a regex, so they should already be quoted appropriately, or they're just string names. I could separate them into two static variables perhaps...

tsteur: It's just that you need to know how the regex is escaped where the regexes are defined. Like where you define the regexes is different place to where the preg_match is defined and by using preg_quote you don't need to know it uses a `slash`. It's a detail though not too important.

diosmosis: Yes, but that would also make a regex like `dbip-city-lite-\d{4}-\d{2}.mmdb` match the exact string `"dbip-city-lite-\d{4}-\d{2}.mmdb"` instead of match as a regex. Or am I missing something?

tsteur: True... it be more like a `str_replace('/', \/, $regex)`... not needed though

core/Updates/3.13.1-b2.php
```diff
@@ -0,0 +1,46 @@
...
+    public function __construct(MigrationFactory $factory)
+    {
+        $this->migration = $factory;
+    }
+
+    public function getMigrations(Updater $updater)
+    {
+        $optionTable = Common::prefixTable('option');
+        $migration = $this->migration->db->boundSql("UPDATE $optionTable SET option_value = ? WHERE option_name = ? AND option_value = ?",
+            [GeoIp2::getDbIpLiteUrl(), GeoIP2AutoUpdater::LOC_URL_OPTION_NAME, GeoIp2::GEO_LITE_URL]);
```
sgiehl: Wondering if we maybe should also replace the lite version of the country db, if someone has set it up
`https://geolite.maxmind.com/download/geoip/database/GeoLite2-Country.tar.gz`
But guess that should be less common case....

tsteur: Does DB IP have a country DB? I suppose we can't really update automatically since some people prefer country DB for privacy reasons.

diosmosis: dbip does have a country DB, we can do another query to update if the user is using geolite2-country

diosmosis: Updated.

sgiehl: :+1:

mattab:
FYI @diosmosis our affiliate link for db-ip.com is     https://db-ip.com/?refid=mtm or especially appending `?refid=mtm` to all links to db-ip.com. With maxmind we had a similar affiliate link. Can you check and add the parameter to all links within the app to db-ip.com?

diosmosis:
@sgiehl @tsteur updated to work w/ purchased dbip urls. tested w/ all of them and tested w/ geolite2 city just to make sure it still works.

tsteur:
BTW: Not sure if important... once it failed for me and then I got this notice:

> Notice: Undefined index: region_code in plugins/GeoIp2/LocationProvider/GeoIp2.php on line 101

diosmosis:
@tsteur should be fixed, let me know if you still get it

plugins/GeoIp2/LocationProvider/GeoIp2.php
```diff
@@ -38,10 +41,19 @@ abstract class GeoIp2 extends LocationProvider
...
+            'dbip-city-lite-\d{4}-\d{2}.mmdb', 'GeoIP2-City-Africa.mmdb', 'GeoIP2-City-Asia-Pacific.mmdb', 'GeoIP2-City-Europe.mmdb',
+            'GeoIP2-City-North-America.mmdb', 'GeoIP2-City-South-America.mmdb', 'GeoIP2-Enterprise.mmdb', 'GeoIP2-Country.mmdb',
+            'dbip-country-lite-\d{4}-\d{2}.mmdb', 'GeoLite2-City.mmdb', 'GeoLite2-Country.mmdb', 'DBIP-Enterprise.mmdb'),
+        'isp' => array('GeoIP2-ISP.mmdb', 'GeoLite2-ASN.mmdb', 'DBIP-ISP.mmdb', 'GeoIP2-Enterprise.mmdb', 'DBIP-Enterprise.mmdb'),
     );
 
+    public static function getDbIpLiteUrl($type = 'city')
+    {
+        $today = Date::today();
+        return "https://download.db-ip.com/free/dbip-{$type}-lite-{$today->toString('Y-m')}.mmdb.gz?refid=mtm";
```
sgiehl: guess the refid param doesn't make sense here?

plugins/GeoIp2/LocationProvider/GeoIp2/Php.php
```diff
@@ -233,12 +298,18 @@ public function getSupportedLocationInfo()
             switch ($reader->metadata()->databaseType) {
```
sgiehl: We should move setting country and continent availability (few lines above) into this if statement. Otherwise it will report `country` as available if even no geoip file is available (or only an isp file is available).

plugins/GeoIp2/LocationProvider/GeoIp2.php
```diff
@@ -136,12 +146,12 @@ public static function getPathForGeoIpDatabase($filename)
...
      * @return array eg. array('1.2.3.4', array(self::COUNTRY_CODE_KEY => ...))
      */
-    private static function getTestIpAndResult()
+    protected function getTestIpAndResult()
     {
         static $result = null;
         if (is_null($result)) {
             $expected = array(self::COUNTRY_CODE_KEY => 'FR',
-                self::REGION_CODE_KEY  => 'BFC',
+                self::REGION_CODE_KEY => 'BFC',
```
sgiehl: The DB-IP Lite database returns no region for this IP, so the provider is reported as broken.
In relation to #15317 we maybe shouldn't use only one IP to check if the results are ok. Maybe we could have one IP per continent and go through the list until one matches exactly? Any idea where to get some IPs with static results?

sgiehl: Or we might use different ips/results depending on the provider

diosmosis: What about using the test database we include for system tests? Could we load that so we can control the IP/location? I started on supporting multiple ip addresses, but we'd have to have quite a few and they might change...

diosmosis: I guess that wouldn't work for server modules... though it looks like servermodule overrides isWorking(). If we move that logic to the Php class it might work I think.

diosmosis: I think I'm going to make the check less strict and just check that there's at least some data in the provider. Users can always see their own IP geolocated to check for accuracy.

mattab:
@diosmosis @sgiehl As soon as this is merged we'll release another beta for testing, looking forward to getting it merged today or tomorrow :rocket:

PR #15330, opened 2019-12-31
Fix archiving when getmypid is disabled

tsteur:
See https://github.com/matomo-org/wp-matomo/issues/163#issuecomment-569846882

Seems this fixes some PHP bug or so. More explanation in above link

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -15,7 +15,9 @@ use Piwik\CronArchive\SitesToReprocessDistributedList;
 use Piwik\DataAccess\ArchiveTableCreator;
 use Piwik\DataAccess\Model;
 use Piwik\Date;
+use Piwik\CliMulti\Process;
 use Piwik\Option;
+use Piwik\Common;
 use Piwik\Piwik;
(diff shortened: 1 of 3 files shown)


core/Archive/ArchiveInvalidator.php
```diff
@@ -130,7 +132,11 @@ private function buildRememberArchivedReportIdForSiteAndDate($idSite, $date)
     private function buildRememberArchivedReportIdProcessSafe($idSite, $date)
     {
         $id  = $this->buildRememberArchivedReportIdForSiteAndDate($idSite, $date);
-        $id .= '_' . getmypid();
+        if (Process::isMethodDisabled('getmypid')) {
+	        $id .= '_' . Common::getRandomString();
```
diosmosis: Should we cache the value of `Common::getRandomString()` so the same value gets used next time (to simulate a real PID)?

tsteur: @diosmosis this method should be executed max once per request as the next time the option is set... could even always use getRandomInt in theory

diosmosis: I see, makes sense 👍 will merge

PR #15333, opened 2019-12-31
fix tests broken since #15265

sgiehl:
Guess those changes were accidentally included in #15265  ?

ping @diosmosis


diosmosis:
Yes, thanks for the fix!

PR #15337, opened 2020-01-01
Match configId even when userId is found

### LINKED ISSUE #15320: Multiple unique visit for each action for same User ID
I use latest matomo 3.13.0

I read:

> ............................................................
> Tracking HTTP API: API Reference - Matomo Analytics (formerly Piwik Analytics) - Developer Docs - v3
> https://developer.matomo.org/api-reference/tracking-api
> 
> JavaScript Tracking Client: Integrate - Matomo Analytics (formerly Piwik Analytics) - Developer Docs - v3 - User ID
> https://developer.matomo.org/guides/tracking-javascript-guide#user-id
> 
> User ID
> 
> User ID is a feature in Matomo that lets you connect together a given user's data collected from multiple devices and multiple browsers.
> ............................................................

I want to use this "User ID feature" and set code in my site:
```
............................................................
<?php
...
@$uid = substr(sha1($_SERVER['REMOTE_ADDR'].$_SERVER['HTTP_USER_AGENT'].$_SERVER['HTTP_ACCEPT_LANGUAGE']), 0, 16);
...
?>

<!-- Matomo -->
  var _paq = window._paq || [];
  _paq.push(["setCookieDomain", "*.example.com"]);
  <?php echo sprintf("_paq.push(['setUserId', '%s']);", $uid); ?>
  _paq.push(['trackPageView']);
  ...
</script>
<!-- End Matomo Code -->

<noscript>
<!-- Matomo Image Tracker -->
<img src="https://example.com/piwik.php?idsite=2&amp;rec=1&amp;url=<?php echo rawurlencode(JUri::getInstance()); ?>&amp;action_name=<?php echo rawurlencode('[noscript] '.$document->title); ?>&amp;send_image=0&amp;urlref=<?php echo rawurlencode(@$_SERVER['HTTP_REFERER']); ?>&amp;uid=<?php echo rawurlencode($uid); ?>&amp;cid=<?php echo rawurlencode($uid); ?>" style="border:0" alt="" />
<!-- End Matomo -->
</noscript>
............................................................
```

In config.ini.php window_look_back_for_visitor var I set to 86400, see my full config.ini.php: https://pastebin.com/KG8FBQHy

But, when I looked statistic:
```
............................................................
Суббота, 28 декабря 2019 г. - 23:24:12
IP: 188.127.251.xx
Россия

Profile ID (aka uid): d2f22ac701961993
Visitor ID (aka cid): 8cbf24c93d0f4628

Прямой вход
Провайдер: Неизвестно

...............

Суббота, 28 декабря 2019 г. - 23:42:17
IP: 188.127.251.xx
Россия

Profile ID (aka uid): d2f22ac701961993
Visitor ID (aka cid): c3f6fec89ef58cb4

Прямой вход
Провайдер: Неизвестно
............................................................
```

I see, that same one visitor from same IP, with device and same "Profile ID" (d2f22ac701961993) counted/tracked as unique visitor instead as "Returned visitor" ;(

**User ID is feature in Matomo or bug? How to fix this bug?**

**This bug appear for user which disable cookies, then matomo for same uid (d2f22ac701961993) generate different cid (8cbf24c93d0f4628 | c3f6fec89ef58cb4) for each visit.**

Related topics for this problem:

New visit created for each action with userid populated since 3.13 · Issue #15299 · matomo-org/matomo · GitHub
https://github.com/matomo-org/matomo/issues/15299

New visit created for each action with userid populated since 3.13 · Issue #28 · matomo-org/piwik-java-tracker · GitHub
https://github.com/matomo-org/piwik-java-tracker/issues/28

tsteur:
fix https://github.com/matomo-org/matomo/issues/15320
refs https://github.com/matomo-org/matomo/pull/15337
fyi @mattab @diosmosis 

I wonder if maybe the query should be actually and `user_id is null or user_id = $userIdFromRequestIfGiven`? 

Edit: Added the `or user_id = ?`

### DIFF
--- a/core/Tracker/Model.php
+++ b/core/Tracker/Model.php
@@ -398,10 +398,13 @@ class Model
         } elseif ($shouldMatchOneFieldOnly) {
             $visitRow = $this->findVisitorByConfigId($configId, $select, $from, $configIdWhere, $configIdbindSql);
         } else {
-            $visitRow = $this->findVisitorByVisitorId($idVisitor, $select, $from, $visitorIdWhere, $visitorIdbindSql);
+            if (!empty($idVisitor)) {
+                $visitRow = $this->findVisitorByVisitorId($idVisitor, $select, $from, $visitorIdWhere, $visitorIdbindSql);
+            } else {
+                $visitRow = false;
+            }
 
             if (empty($visitRow)) {
-                $configIdWhere .= ' AND user_id IS NULL ';
                 $visitRow = $this->findVisitorByConfigId($configId, $select, $from, $configIdWhere, $configIdbindSql);
             }
         }
--- a/core/Tracker/Model.php
+++ b/core/Tracker/Model.php
@@ -366,7 +366,7 @@ class Model
         return $wasInserted;
     }
 
-    public function findVisitor($idSite, $configId, $idVisitor, $fieldsToRead, $shouldMatchOneFieldOnly, $isVisitorIdToLookup, $timeLookBack, $timeLookAhead)
+    public function findVisitor($idSite, $configId, $idVisitor, $userId, $fieldsToRead, $shouldMatchOneFieldOnly, $isVisitorIdToLookup, $timeLookBack, $timeLookAhead)
     {
         $selectCustomVariables = '';
 
@@ -405,6 +405,13 @@ class Model
             }
 
             if (empty($visitRow)) {
+                $configIdWhere .= ' AND ( user_id IS NULL ';
+                if (!empty($userId)) {
+                    $configIdWhere .= 'OR user_id = ? )';
+                    $configIdbindSql[] = $userId;
+                } else {
+                    $configIdWhere .= ')';
+                }
                 $visitRow = $this->findVisitorByConfigId($configId, $select, $from, $configIdWhere, $configIdbindSql);
             }
         }
--- a/core/Tracker/Model.php
+++ b/core/Tracker/Model.php
@@ -405,12 +405,9 @@ class Model
             }
 
             if (empty($visitRow)) {
-                $configIdWhere .= ' AND ( user_id IS NULL ';
                 if (!empty($userId)) {
-                    $configIdWhere .= 'OR user_id = ? )';
+                    $configIdWhere .= 'AND ( user_id IS NULL OR user_id = ? )';
                     $configIdbindSql[] = $userId;
-                } else {
-                    $configIdWhere .= ')';
                 }
                 $visitRow = $this->findVisitorByConfigId($configId, $select, $from, $configIdWhere, $configIdbindSql);
             }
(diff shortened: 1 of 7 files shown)


diosmosis:
code looks good to me, and the user_id = ? addition makes sense to me after thinking about it. might be good to have a test (if current tests aren't already affected), otherwise looks good to me.

core/Tracker/Model.php
```diff
@@ -398,10 +398,20 @@ public function findVisitor($idSite, $configId, $idVisitor, $fieldsToRead, $shou
...
-            $visitRow = $this->findVisitorByVisitorId($idVisitor, $select, $from, $visitorIdWhere, $visitorIdbindSql);
+            if (!empty($idVisitor)) {
+                $visitRow = $this->findVisitorByVisitorId($idVisitor, $select, $from, $visitorIdWhere, $visitorIdbindSql);
+            } else {
+                $visitRow = false;
+            }
 
             if (empty($visitRow)) {
-                $configIdWhere .= ' AND user_id IS NULL ';
+                $configIdWhere .= ' AND ( user_id IS NULL ';
```
tsteur: @diosmosis @mattab noticing this would create a new visit when userId was previously set, and then user logs out using resetUserId. I suppose this is not what we want as we also don't create a new visit when userId is set initially. I suppose we maybe really only always want to add ` AND ( user_id IS NULL OR user_id = ? )` when userId is not empty?

tsteur: Also when user is found by idVisitor, then we would also not create a new visit after resetUserId so we need to make sure to have consistent behavior... I will later change the query I reckon to only compare userId when a userId is set

diosmosis: If userId is null, we'd want to look for ( user_id IS NULL ) though, correct? Otherwise it would match the logged out user. Then again, ( user_id IS NULL OR user_id = ? ) would also match the user after they login (ie, logged out => login), so maybe both behaviors are desired?

tsteur: That's current desired/ implemented behaviour AFAIK and that's how it also works when it matches the visit by visitorId.  See thread in https://github.com/matomo-org/developer-documentation/pull/320

We just need to make sure to always have consistent behaviour on log in and log out. Either we always create a new visit on log in and log out, or never.

diosmosis: I thought the point of separating user ID from visitor ID was so you could tell when a user logged in/out. Ie, it would count as the same visitor, but could as different users?

tsteur: I'm not sure? If that's the case, then I think this wasn't implemented.

diosmosis: I think you're right it was never implemented. It does seem more useful to me though. I guess it could go in another issue.

tsteur:
@diosmosis I'll merge for now. We'll see what comes out of https://github.com/matomo-org/developer-documentation/pull/320 re forcing new visits

PR #15348, opened 2020-01-05
Do not use get_results in mysqli as it requires mysqlnd

### LINKED ISSUE #15345: Matomo 3.13.1.b2 : Call to undefined method mysqli_stmt::get_result()
Hello,

This occur since last beta update.

Matomo version: 3.13.1-b2

 PHP Fatal error: Call to undefined method mysqli_stmt::get_result() in (...) piwik/core/Tracker/Db/Mysqli.php

Tracking is completely down : all request are throwing 500 errors.

Thanks,

Eric

tsteur:
fix https://github.com/matomo-org/matomo/issues/15345

### DIFF
--- a/core/Tracker/Db/Mysqli.php
+++ b/core/Tracker/Db/Mysqli.php
@@ -149,6 +149,39 @@ class Mysqli extends Db
     }
 
     /**
+     * @param \mysqli_stmt $stmt
+     * @param $fields
+     * @return array|bool|false
+     */
+    private function fetchResult($stmt, $fields) {
+
+        $values = array_fill(0, count($fields), null);
+
+        $refs = array();
+        foreach ($values as $i => &$f) {
+            $refs[$i] = &$f;
+        }
+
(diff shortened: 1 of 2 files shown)


core/Tracker/Db/Mysqli.php
```diff
@@ -148,6 +148,39 @@ public function disconnect()
...
+     * @return array|bool|false
+     */
+    private function fetchResult($stmt, $fields) {
+
+        $values = array_fill(0, count($fields), null);
+
+        $refs = array();
+        foreach ($values as $i => &$f) {
+            $refs[$i] = &$f;
+        }
```
diosmosis: Is `$refs` needed? It doesn't seem to be used anywhere.

tsteur: It was done like this in Zend I suppose it turns it into a reference.

diosmosis: that's weird, but ok zend

PR #15351, opened 2020-01-06
hasConsent() is used in tracker so can't be a debug method

diosmosis:
Otherwise there'll be a JS error when the opt out frame is used w/ the minified JS.


tsteur:
Looks good if tests pass

PR #15358, opened 2020-01-08
Use tracker cache in ArchiveInvalidator so extra query is not needed.

### LINKED ISSUE #15338: Additional options query per tracking request
On every regular request we are now performing an option select query in https://github.com/matomo-org/matomo/blob/3.13.1-b1/core/Archive/ArchiveInvalidator.php#L71 caused by report invalidation in https://github.com/matomo-org/matomo/blob/3.13.1-b1/core/Tracker/Visit.php#L581 

This was changed when removing temporary archives in https://github.com/matomo-org/matomo/pull/14639 since every request might now invalidate a previous archive. Before this was only executed when tracking data into a previous day that is not today and therefore rather rare. 

Can we maybe use this `option like %` query use a cache? That would mean though we also need to invalidate the cache when this flag is removed.

Not sure if there is a different alternative.

@mattab marking this for now as a regression since we don't want additional queries like these for each tracking request. Feel free to change it.

diosmosis:
Fixes #15338

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -23,6 +23,7 @@ use Piwik\Plugins\CoreAdminHome\Tasks\ArchivesToPurgeDistributedList;
 use Piwik\Plugins\PrivacyManager\PrivacyManager;
 use Piwik\Period;
 use Piwik\Segment;
+use Piwik\SettingsServer;
 use Piwik\Site;
 use Piwik\Tracker\Cache;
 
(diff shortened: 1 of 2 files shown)


core/Archive/ArchiveInvalidator.php
```diff
@@ -90,7 +103,21 @@ public function rememberToInvalidateArchivedReportsLater($idSite, Date $date)
             // that happen to run in overlapping transactions.
             $mykey = $this->buildRememberArchivedReportIdProcessSafe($idSite, $date->toString());
             Option::set($mykey, '1');
+            Cache::clearCacheGeneral();
+        }
+    }
+
+    private function getRememberedArchivedReportsOptionFromTracker($idSite, $dateStr)
+    {
+        $cacheKey = self::TRACKER_CACHE_KEY;
```
tsteur: Ideally, it would check whether we are in tracker mode and then read it from cache, or otherwise use the result directly. Otherwise we have to basically launch Matomo in tracker mode during a regular request. Seems to be eg used in DataSubjects which is not a tracking request.

core/Archive/ArchiveInvalidator.php
```diff
@@ -68,16 +71,26 @@ public function __construct(Model $model, ArchivingStatus $archivingStatus)
...
         // MyISAM installations)
-        $value = Option::getLike($key . '%');
+        $values = Option::getLike($this->rememberArchivedReportIdStart . '%');
+
+        $all = [];
+        foreach ($values as $name => $value) {
+            $suffix = substr($name, strlen($this->rememberArchivedReportIdStart));
+            list($idSite, $dateStr) = explode('_', $suffix);
+
+            $all[$idSite][$dateStr][$name] = $value;
```
tsteur: Probably not too important but `[$name]` could be removed? It doesn't look at this one in the cached tracker value and it would also prevent that when 200 tracking requests at the same time use the `option::set` that we have 200 keys in there for the same idSite,dateStr combination

PR #15364, opened 2020-01-09
Avoid using getmypid

tsteur:
@diosmosis was getting this error a few times:

image

I wonder if we could simply generate a random number instead of using `getmypid`? I noticed the class is always used through DI so in the end the instance should be only created once and therefore the `pid` should always be the same even though the property is not static. Could also make it static though if that helps and only set it once? 

Not too important for 3.13.1 but be good. We can move it to 4.0 though and I patch later Matomo for WordPress after the 3.13.1 release

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -171,11 +171,7 @@ class ArchiveInvalidator
     private function buildRememberArchivedReportIdProcessSafe($idSite, $date)
     {
         $id  = $this->buildRememberArchivedReportIdForSiteAndDate($idSite, $date);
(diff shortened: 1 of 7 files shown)


diosmosis:
It might need to be cached and the same one used for the entire run (not too sure, just thinking)

tsteur:
@diosmosis was adding the cache but then noticed this was making things worse as cache can be reset any time and then new PID was generated. The original simple logic worked actually fine after looking into it a bit more but figured better to create a method in `Common` class for this purpose since we had similar logic somewhere else. First had it `Common::getMyPid` but this could lead to wrong assumption when not reading the text so named it `getProcessId`.

PR #15370, opened 2020-01-10
Save concatenated less file on compiling error so we can better debug issues.

diosmosis:


core/AssetManager/UIAssetMerger/StylesheetUIAssetMerger.php
```diff
@@ -46,6 +47,9 @@ protected function getMergedAssets()
         try {
             $compiled = $this->lessCompiler->compile($concatenatedAssets);
         } catch(\Exception $e) {
+            // save the concated less files so we can debug the issue
+            $this->saveConcatenatedAssets($concatenatedAssets);
```
tsteur: @diosmosis would that be only something for dev mode? Although I suppose it should be fine and not cause much of an issue.

diosmosis: This is to help debug a cloud error, so it would ideally be available in a prod setup. Without the concatted less files, debugging the error is basically impossible.

tsteur: 👍

Issue #15313, opened 2019-12-23
Set secure flag for _pk_testcookies


cundd:
The test-cookie is not explicitly set as "secure". This generates a warning with our PCI Security Scan.

```
THREAT:
The cookie does not contain the "secure" attribute.
IMPACT:
Cookies with the "secure" attribute are only permitted to be sent via HTTPS. Cookies sent via HTTP expose an unsuspecting user to sniffing attacks that could lead to user impersonation or compromise of the application account.
SOLUTION:
If the associated risk of a compromised account is high, apply the "secure" attribute to cookies and force all sensitive requests to be sent via HTTPS.
RESULT:
1GEThttps://domain.comSG9ovKg0K_pk_testcookie.11.e5c9=1; path=/; domain=domain.com
```

A possible fix could be to pass `configCookieIsSecure` to the `setCookie()` function.

```
@@ -3925,7 +3925,7 @@ if (typeof window.Piwik !== 'object') {
 
                 // for IE we want to actually set the cookie to avoid trigger a warning eg in IE see #11507
                 var testCookieName = configCookieNamePrefix + 'testcookie';
-                               setCookie(testCookieName, '1');
+                setCookie(testCookieName, '1', undefined, configCookiePath, configCookieDomain, configCookieIsSecure);
```

Is this a viable solution? Should I send a PR?

tsteur:
@cundd be great. You probably want to respect the `configCookieIsSecure` setting when setting the cookie within the `hasCookies` method.

cundd:
Should be fixed

PR #15377, opened 2020-01-10
Respect cookie config for the test (path, domain, isSecure flag) (#15313)

cundd:
This PR uses the `configCookiePath`, `configCookieDomain` and `configCookieIsSecure` configurations when the test-cookie is set.


cundd:
Unfortunately I wasn't able to correctly preform the UI test suites. Are there tests for the tracking JS that need to pass/I need to modify to make the tests pass?

tsteur:
Cheers @cundd the JS tests are parsing. You can execute them by opening `https://yourMatomodomain.com/tests/javascript`.

There were some integration tests failing as it is currently checking to not have an empty line at the end for some reason so tweaked it. Will merge it soon.

Tested it and worked 👍

PR #15384, opened 2020-01-13
Updates search engine and social definitions

sgiehl:
use latest release of matomo/searchengine-and-social-list


sgiehl:
some tests are failing until the missing icons will be added. Will merge it anyway as it's better to have the new detections included in upcoming release than waiting for the icons

PR #15390, opened 2020-01-14
Store session ID hashed in the DB

tsteur:


sgiehl:
Looks good and works locally. Wondering if that might make any problems while updating. If the user is currently logged in, I guess he will be automatically logged out with the code changes as the session can't be found anymore 🤔

tsteur:
@sgiehl it should be fine if they are logged out since they will still be able to complete the updater. Once updating is started basically anyone can complete it.

Issue #7893, opened 2015-05-12
Add possibility to create Mobile Apps in Sites Management  (and later any type)


tsteur:
This issue needs to be solved for #7131 and #4734

A "type" could be a plugin that defines a type "Website" or "MobileApp". Different types need not only different reports and different naming of reports but also different fields in the Sites Manager. The Sites Manager might become at some point a Property Manager or something else. Naming is not important for now.

A mobile app would not define any URLs but an App-ID such as `org.piwik.mobile` or `737216887 `. 

We would maybe also allow to select an App store? Eg Google Play, ... Not needed by core actually but by a App Analytics plugin later. Not sure how to handle this!

**Things to consider**
* In the Sites Manager are currently columns shown like "URLs" which does not apply to all kind of types. After #7587 was implemented we could maybe show only columns that apply to actually all "types" like "Excluded IPs" etc.  URLs would not be shown in the overview but when the site is expanded. I know this lowers the usability a bit since it is nice to see the URLs at a glance. We could do this only if there are different types etc. 
* The type "Website" would define a URLs field. This would be not shown to other types by default.
* The "JavaScript Tracking Code" would be only needed for sites but not for mobile apps. Here we would want to see either nothing (when no SDK is available) or links / code for Android / iOS SDK tracking generator code refs #7803
* We would detect if a plugin defines a "Type". If so, when creating a new type/property/website, we would also show all available plugins on the marketplace defining a new type. This means the user will be directly able to install new plugins from there etc. (in a later version)
* When a type defines a new field such as "App identifier" then we need to create a column in the site table for that. Alternatively we will store such things as JSON in a "custom" field (more likely we go with this). 
* TBD how to handle URLs for Mobile Apps. I think some features depend on URL's.

tsteur:
We actually need the possibility to require plugins

tsteur:
So far we came to the following definition.

A property (eg `Piwik Mobile 2`) has a name, a created timestamp, some other fields but also it has a type (eg `MobileApp`).

Some more of my thoughts:

In general we might differentiate between `attributes`, `setting` and `feature`:
* A feature would be for example "Site Search" as it provides new functionality 
* An attribute would be for example "Currency" or "App-ID" as a property has a currency or app-id and it is just a plain attribute more or less.
* A setting or config could be for example `action_title_category_reverse` as it changes existing functionality. Eg it reverses the action titles but it doesn't provide the feature of `action titles` itself. 
[...]

PR #15394, opened 2020-01-14
Enable MobileAppMeasurable plugin to allow users to track mobile apps

mattab:
This would not provide much value but would make it clear that Matomo can be used to track mobile apps. Also the plugin already exists and be great to leverage it. For example it would be valuable to hide some specific menus for mobile app measurables, where we know the menu will never have data (eg. Referrers, Forms, Media, Sessions, Heatmaps...)

refs https://github.com/matomo-org/matomo/issues/7893

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -955,6 +955,7 @@ Plugins[] = CoreAdminHome
 Plugins[] = CoreHome
 Plugins[] = WebsiteMeasurable
 Plugins[] = IntranetMeasurable
+Plugins[] = MobileAppMeasurable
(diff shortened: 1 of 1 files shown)


tsteur:
Voting to not include it by default as it makes the experience bit worse. Currently, there is not really any difference between mobile and web and we won't be working on this any time in the near future. 

It makes at the same time UI and usability worse since it will for example rename Websites to Measurables and making it harder for example for users to find the right link to manage websites etc.

tsteur:
@mattab can we close this PR? Currently it would make things a lot worse than better re UX etc.

PR #15395, opened 2020-01-15
Removes some more unneeded code

sgiehl:


misc/composer/build-xhprof.sh
```diff
@@ -2,7 +2,7 @@
 
 if ! type phpize &> /dev/null; then
     echo "phpize missing, skipping build"
-    echo "If you installed PHP via Aptitude, you can install phpize w/ 'sudo apt-get install php5-dev'"
+    echo "If you installed PHP via Aptitude, you can install phpize w/ 'sudo apt-get install php7-dev'"
```
tsteur: does xhprof even still work with PHP7? AFAIK not?

tsteur: I think we could just remove the script... anyone can install it manually anyway if needed

sgiehl: It seems to work with php7 meanwhile: http://pecl.php.net/package/xhprof

sgiehl: shall I remove the whole package from composer then, or only the scripts to be executed?

tsteur: Sweet, looks like it was made compatible just last month maybe. We can also keep it then as it is. No Preference. Feel free to merge the PR

sgiehl: Have no preference either. But could you have a look at the composer script. Guess that needs a change. But I'm not using xhprof at all, so it's hard for me to check what needs to be done there:

https://github.com/matomo-org/matomo/blob/b16a791aa3650d85af829156129c2bd44c7cb075/core/Composer/ScriptHandler.php#L30-L33

tsteur: Removed the `isPhp7orLater` for a test and it at least didn't fail for me on PHP 7.3

sgiehl: ok. removed the checks

PR #15396, opened 2020-01-15
Remove LoginHttpAuth submodule

tsteur:
We decided we no longer support it in Matomo 4 for various reasons such as security and wanting to focus more on the essential things and being able to focus. I suppose we could create an issue on LoginHttpAuth if someone wanted to take over the project and continue supporting it? A few changes would need to be made to make it compatible with Matomo 4 because of #6559

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,10 +18,6 @@
     path = plugins/TasksTimetable
     url = https://github.com/matomo-org/plugin-TasksTimetable.git
     branch = master
-[submodule "plugins/LoginHttpAuth"]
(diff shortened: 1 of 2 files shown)


sgiehl:
Guess that line can be removed as well then:

https://github.com/matomo-org/matomo/blob/444e8415ed5e0771213be972bd96249f091fdbf1/tests/PHPUnit/Framework/TestingEnvironmentVariables.php#L107

tsteur:
👍 made that change

sgiehl (approved):

PR #15405, opened 2020-01-16
enable send beacon by default

### LINKED ISSUE #13681: JS Tracker: enable alwaysUseSendBeacon by default
assuming that the trial went well, then in Matomo 4.0.0 we would like to enable the feature `alwaysUseSendBeacon` by default. 

see https://github.com/matomo-org/matomo/pull/13451 https://github.com/matomo-org/matomo/issues/6641

tsteur:
Will update minified JS  once PR is approved 

fix https://github.com/matomo-org/matomo/issues/13681


sgiehl:
Do we need to mention that in the changelog or update some other related faqs or docs later?
Otherwise looks good to merge

sgiehl (approved):

tsteur:
@sgiehl added changelog entry 👍

PR #15406, opened 2020-01-16
No longer redirect when interacted with content

### LINKED ISSUE #6265: Content-tracking: I want to be able to disable piwik.php redirect and instead use the link tracking 500ms solution.
Especially useful for single page applications but also if someone does not want to have piwik.php links in his page.

piwik.php redirects are currently used if a link goes to the same domain.

### LINKED ISSUE #12227: Content tracking doesn't track interaction correctly when using multiple trackers but outlink
Content tracking does currently replace the link in the `a.href` when clicking on it (when the link is supposed to be tracked as an interaction). This means we will first load eg `piwik.php?redirect_url=...` which then goes to the actual redirect_url. This redirect will track the interaction happened.

However, when there are 2 trackers, the first tracker will replace the actual target URL with `piwik.php?redirect_url=...` and the second tracker will track the link like this as an outlink: 

image

Ideally, these redirect_urls can be disabled optionally see #6265, and ideally, when using multiple trackers, we disable them automatically.

The result is for example `0` interactions in the second Piwik, and the interactions are shown in the first piwik.

tsteur:
fix https://github.com/matomo-org/matomo/issues/6265
fix https://github.com/matomo-org/matomo/issues/12227

Will update minified JS once PR is approved.

We will also need to merge https://github.com/matomo-org/plugin-QueuedTracking/pull/120


js/piwik.js
```diff
@@ -4062,81 +4058,6 @@ if (typeof window.Piwik !== 'object') {
...
-                    var contentTarget = block.target;
-
-                    if (!query.hasNodeAttributeWithValue(targetNode, content.CONTENT_TARGET_ATTR) || targetNode.wasContentTargetAttrReplaced) {
-                        // make sure we still track the correct content target when an interaction is happening
-                        targetNode.wasContentTargetAttrReplaced = true;
-                        contentTarget = content.toAbsoluteUrl(url);
-                        query.setAnyAttribute(targetNode, content.CONTENT_TARGET_ATTR, contentTarget);
-                    }
-
-                    var targetUrl = buildContentInteractionTrackingRedirectUrl(url, 'click', contentName, contentPiece, contentTarget);
```
sgiehl: Is there a reason to keep the method `buildContentInteractionTrackingRedirectUrl` (and its tests)?

tests/PHPUnit/Integration/Tracker/RequestSetTest.php
```diff
@@ -128,49 +128,6 @@ private function buildRequest($idsite)
         return $request;
     }
 
-    public function test_shouldPerformRedirectToUrl_shouldNotRedirect_IfNoUrlIsSet()
```
sgiehl: I guess the class `TestRequestSet` in the beginning of the file can now be adjusted and the redirect stuff can be removed from it.

core/Tracker/RequestSet.php
```diff
@@ -145,41 +133,6 @@ protected function getAllSiteIdsWithinRequest()
...
-        if (!$this->hasRedirectUrl()) {
-            return false;
-        }
-
-        if (!$this->hasRequests()) {
-            return false;
-        }
-
-        $redirectUrl = $this->getRedirectUrl();
-        $host        = Url::getHostFromUrl($redirectUrl);
```
sgiehl: `use` statement for `Url` class could be removed now...

core/Tracker/RequestSet.php
```diff
@@ -145,41 +133,6 @@ protected function getAllSiteIdsWithinRequest()
...
-
-        if (empty($host)) {
-            return false;
-        }
-
-        $urls     = new SiteUrls();
```
sgiehl: `use` statement for `SiteUrls` class could be removed now...

sgiehl (changes requested):

core/Tracker/RequestSet.php
tsteur: fixed

core/Tracker/RequestSet.php
tsteur: fixed

js/piwik.js
tsteur: fixed

tests/PHPUnit/Integration/Tracker/RequestSetTest.php
tsteur: fixed now as well

tsteur:
Cheers @sgiehl made the changes

sgiehl (approved):

PR #15408, opened 2020-01-16
Fix event value is truncated

### LINKED ISSUE #14934: Matomo trackEvent bug, event value is truncated
We have encountered an issue with TrackEvent, when you send an event value like 12225057, the value is becoming 12225100 on matomo dashboard.

When you send an event value like 719 then it is recorded correctly.

tsteur:
fix https://github.com/matomo-org/matomo/issues/14934

should require 2 bytes for the right part, and about 7 byes for the left part.

### DIFF
--- a/plugins/API/tests/System/expected/test_AutoSuggestAPITest_eventValue__VisitsSummary.get_range.xml
+++ b/plugins/API/tests/System/expected/test_AutoSuggestAPITest_eventValue__VisitsSummary.get_range.xml
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <result>
-	<nb_visits>0</nb_visits>
-	<nb_actions>0</nb_actions>
-	<nb_visits_converted>0</nb_visits_converted>
+	<nb_visits>4</nb_visits>
+	<nb_actions>20</nb_actions>
+	<nb_visits_converted>4</nb_visits_converted>
 	<bounce_count>0</bounce_count>
-	<sum_visit_length>0</sum_visit_length>
-	<max_actions>0</max_actions>
+	<sum_visit_length>6492</sum_visit_length>
+	<max_actions>5</max_actions>
 	<bounce_rate>0%</bounce_rate>
-	<nb_actions_per_visit>0</nb_actions_per_visit>
-	<avg_time_on_site>0</avg_time_on_site>
+	<nb_actions_per_visit>5</nb_actions_per_visit>
+	<avg_time_on_site>1623</avg_time_on_site>
 </result>
--- a/core/Db/Schema/Mysql.php
+++ b/core/Db/Schema/Mysql.php
@@ -225,7 +225,7 @@ class Mysql implements SchemaInterface
                                         idvisit BIGINT(10) UNSIGNED NOT NULL,
                                         idaction_url_ref INTEGER(10) UNSIGNED NULL DEFAULT 0,
                                         idaction_name_ref INTEGER(10) UNSIGNED NULL,
-                                        custom_float DECIMAL(14,4) NULL DEFAULT NULL,
+                                        custom_float DOUBLE NULL DEFAULT NULL,
                                           PRIMARY KEY(idlink_va),
                                           INDEX index_idvisit(idvisit)
                                         ) ENGINE=$engine DEFAULT CHARSET=utf8
--- a/core/Db/Schema/Mysql.php
+++ b/core/Db/Schema/Mysql.php
@@ -225,7 +225,7 @@ class Mysql implements SchemaInterface
                                         idvisit BIGINT(10) UNSIGNED NOT NULL,
                                         idaction_url_ref INTEGER(10) UNSIGNED NULL DEFAULT 0,
                                         idaction_name_ref INTEGER(10) UNSIGNED NULL,
-                                        custom_float FLOAT NULL DEFAULT NULL,
+                                        custom_float DECIMAL(14,4) NULL DEFAULT NULL,
                                           PRIMARY KEY(idlink_va),
                                           INDEX index_idvisit(idvisit)
                                         ) ENGINE=$engine DEFAULT CHARSET=utf8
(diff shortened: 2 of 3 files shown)


tsteur:
FYI Using decimal it causes the values to appear like this: https://travis-ci.org/matomo-org/matomo/jobs/638207041#L5670

We can workaround this by using eg `TRIM(TRAILING '0' FROM custom_float)` AFAIK but adds some additional overhead. Decided to simply switch to `DOUBLE` which should store most values accurately as well if there are no objections while needing same amount of storage basically. Should be also better for comparisons etc.

plugins/API/tests/System/expected/test_AutoSuggestAPITest_eventValue__VisitsSummary.get_range.xml
```diff
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <result>
-	<nb_visits>0</nb_visits>
-	<nb_actions>0</nb_actions>
-	<nb_visits_converted>0</nb_visits_converted>
+	<nb_visits>4</nb_visits>
```
tsteur: Not sure why this test didn't used to work. Ran the test with float and double and in both cases the value looked like they were stored correctly. 
This is what the fixture data looks like
image

sgiehl:
> while needing same amount of storage basically

Doesn't `DOUBLE` need the double amount of space than `FLOAT` 🤔 

According to https://dev.mysql.com/doc/refman/8.0/en/floating-point-types.html it uses 8 instead of 4 bytes.

But nevertheless, the numbers will get stored more accurate...

plugins/API/tests/System/expected/test_AutoSuggestAPITest_eventValue__VisitsSummary.get_range.xml
sgiehl: It didn't work as the comparison `custom_float = '345.678'` doesn't return any results when the field is a float. The number seems not to be stored accurate enough to be able to compare it correctly. When changing the column to `DOUBLE` it gets converted to `345.6780090332031`.

Not sure if it would make sense to truncate some minor fraction digits after changing the type 🤔 

Using a `DOUBLE` the number is stored accurate enough to be compared correctly

tsteur:
> while needing same amount of storage basically

That was compared to the decimal alternative, not compared to float.

plugins/API/tests/System/expected/test_AutoSuggestAPITest_eventValue__VisitsSummary.get_range.xml
tsteur: Could use something like `DOUBLE(15,5)` or so and limit it to 5 numbers after the  dot. I just tested it and the problem is that then a similar issue as with decimal happens that all numbers are formatted to have 5 digits after dot
image

sgiehl: Could we maybe additionally run a script like `UPDATE log_link_visit_action SET custom_float = TRUNCATE(custom_float, 5)` after it was converted to double? That should throw away all fraction digits after the fifth one 🤔

tsteur: @sgiehl from what I see mysql stores also new values like this. 

and truncate seems to ensure there's always that many zeros too?
image

sgiehl: If we are using `DOUBLE` only (without defining any fixed fraction digits), that should work, doesn't it?

tsteur: Yes, that's why I ended up only defining double

sgiehl: Guess that's fine, but it might be useful to truncate the fraction digits on update, so the comparison of values works better after the update 🤔

tsteur: @sgiehl not sure I understand this fully? If we just use `DOUBLE` then this shouldn't be needed?

sgiehl: When we currently have a value like `345.678` and the field is converted to `DOUBLE`, the new value will be something like `345.6780090332031`. That's the reason why I suggested to run something like a TRUNCATE, so we have a similar value as before

tsteur: Which MySQL version is that? This wasn't happening for me on MySQL 8. Or maybe I'm not checking correctly?

sgiehl: mysql 5.7

tsteur: I see, can reproduce it on `10.2-MariaDB`. I just created 3 rows with values 3, 4.1 and 4.12. After converting to float I got this:

image

After executing your suggested query I got this:

image

So it's still changing the values quite a bit. 

It seems there might not be a safe way to change the field? I wonder if it's worth it or maybe we rather create an FAQ for people needing more accuracy and only try to change it for new installs.

@mattab there doesn't seem to be an easy way to change the field for this column. I reckon we could really rather only change it for new installs maybe and create an FAQ for all other cases.

tsteur:
@sgiehl as discussed changing the type only for new installs and created FAQ for existing installs https://matomo.org/faq/how-to/how-do-i-fix-a-truncated-event-value/

sgiehl (approved):
Looks good now. Seems all failing tests are because the branch is a bit outdated...

PR #13443, opened 2018-09-18
Support goal metrics for browser name dimension

tsteur:
fix #13202

This PR will create a column in the conversion table and therefore we need to do it in a Matomo 4.0 release.

PR #15409, opened 2020-01-17
Support goal metrics for browser name dimension 

### LINKED ISSUE #13202: Show "browser" report in goal by dimension report
We already show various reports in a Goals reporting page in the "Goals by dimension" widget... 
image

We should also show browser information here. This is important cause when a goal is not converting as much anymore as it used to do, you want to see if it is maybe related to a specific browser. For example there could be a JS error only in one or few browsers on a page preventing users to achieve a goal. 

Having this browser segment in the goal report be very helpful. 
The "Browser" report be very useful. Could optionally also add "Browser version" report.

tsteur:
replaces https://github.com/matomo-org/matomo/pull/13443
fix #13202

This PR will create a column in the conversion table and therefore we need to do it in a Matomo 4.0 release.

@sgiehl fyi had somehow trouble pushing into the other branch so creating a new one where I will be fixing the tests

Fixing custom dimensions test might be easier after merging


tests/PHPUnit/System/BackwardsCompatibility1XTest.php
```diff
@@ -118,6 +118,7 @@ public function getApiForTesting()
 
             // those reports generate a different segment as a different raw value was stored that time
             'DevicesDetection.getOsVersions',
+            'DevicesDetection.getBrowserFamilies',
```
sgiehl: shouldn't `DevicesDetection.getBrowsers` be in the list as well? Also are the expected files for those methods still needed then? Guess those methods are skipped for testing 🤔

sgiehl (changes requested):
Works locally, but some tests are still failing and need some more adjustments...

tests/PHPUnit/System/BackwardsCompatibility1XTest.php
tsteur: Added it now. The failing test for getBrowsers was initially not shown on travis only after getBrowserFamilies test was fixed. The test is not skipped but instead have a separate fixture so the files be still needed I suppose.

Issue #15368, opened 2020-01-09
Provide ability to restrict auth tokens to site, access, scope


tsteur:
In https://github.com/matomo-org/matomo/issues/6559

I am starting to implement app specific authtokens/passwords. 

I started adding some additional features to further increase the security of tokens:

* Scope: Let users choose if token should be valid for `Reporting API`, and/or `Tracking API`, `Widgets`
* Access: I was going to let users choose what access the token should have. Eg an admin user could decide the token should have only `view` or `write` or `admin` access (but not super user)
* Sites: I was going to let the user choose whether the token should have access to all sites, or only one site.

Of course this way you could create different combination of tokens to lower the risk a lot, eg
* A write token for the tracking API that has only access to one site
* A reporting token with view permission for only a specific site even though the user is super user or write user or admin user
* A token for the exported widgets with only view access which has only access for one site

This way, even if a tracker gets the token, the scope of what they can do is quite restricted. 

It's tricky to implement though. Eg likely we would need to use completely different `Access` class depending on whether user is authenticated through UI, or through token_auth. It me mostly done though by changing maybe the behaviour of `Access:loadSitesIfNeeded` but not sure. Also we would need to check in various places eg in `API::index()` whether the token is allowed for the current scope etc.

Figured I create separate issue for now to simplify #6559.

PR #15410, opened 2020-01-17
App specific token_auths

### LINKED ISSUE #6559: Add support for app specific passwords

tsteur:
fix #6559 

* We now store token_auth no longer in the `user` table but hashed in a new table `user_token_auth`
* To not needing to rewrite the entire app re `piwik.token_auth` basically now storing in the session an extra token auth which is randomly generated when a user logs in. This random token is basically just like an `nonce` and because we need to store it in the session, it is stored in plain text in the database (base64 encoded in the session table). The plain text storage is not an issue though since it is not really a token auth but more an `nonce` and it only works if you also have the sessionID. And if you have the sessionId, you are authenticated anyway
* Also created https://github.com/matomo-org/matomo/pull/15390 so sessionIDs are hashed as well.
* Also removes `login` and the weakly hashed `token_auth` from the cookie.
* To not break BC for existing token_auths we migrate them automatically.
* Token auths still have same permission and access as the actual user. Created https://github.com/matomo-org/matomo/issues/15368 which may be interesting in the future.
* We differentiate between system tokens and regular user tokens.
  * User auth token is a token created by the user and a user can delete them etc.
  * A system auth token is created by the system and not visible to the user. It's used in the background when needed. This tokens have an expiry date and are invalidated/deleted typically within a few hours or days. They were needed because eg when we archive using http (when CliMulti cannot be used), then we need to issue an HTTP request from `CoreArchive` and this request has to be authenticated. In this case `CoreArchive` generates a token auth that is valid for say 2 days and it can use this generated token to authenticate the archiving requests.
* We store the tokens hashed as `sha512`. Was going to use `sha3-512` but so far `sha512` seems more "proven". There is a column in the table storing the algorithm for future migrations.
* Introduces a new page "Security" where these tokens can be managed plus the password and 2FA.

## Todo before release
* We'll need to adjust the mobile app before the release to support this new flow of accessing tokens and also to let them enter a token directly.
* We'll need to adjust some docs/faqs as we now only show the token after creation only.


tsteur:
Ready for a first review...

Will in the mean time need to check why some screenshots like https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/37868/UIIntegrationTest_admin_home.png
and some Dashboard screenshots seem to be failing now.

core/Db/Schema/Mysql.php
```diff
@@ -43,15 +44,27 @@ public function getTablesCreateSql()
             'user'    => "CREATE TABLE {$prefixTables}user (
                           login VARCHAR(100) NOT NULL,
                           password VARCHAR(255) NOT NULL,
-                          alias VARCHAR(45) NOT NULL,
+                          alias VARCHAR(100) NOT NULL,
```
sgiehl: Any reason for changing that? If so it should also be changed by the update script

plugins/CoreHome/angularjs/widget-loader/widgetloader.directive.js
```diff
@@ -113,7 +113,7 @@
                         }
 
                         if (piwik.shouldPropagateTokenAuth && broadcast.getValueFromUrl('token_auth')) {
-                            url += '&token_auth=' + broadcast.getValueFromUrl('token_auth');
+                            url += '&force_api_session=1&&token_auth=' + broadcast.getValueFromUrl('token_auth');
```
sgiehl:
```suggestion
                            url += '&force_api_session=1&token_auth=' + broadcast.getValueFromUrl('token_auth');
```

core/Db/Schema/Mysql.php
tsteur: Forgot to push that which I did now. Shouldn't have been changed and corrected it. Eventually alias will be removed anyway

tsteur:
The tests should now pass.

core/Access.php
```diff
@@ -155,8 +156,24 @@ public function reloadAccess(Auth $auth = null)
...
+        if ($forceApiSession && Piwik::getModule() === 'API' && (Piwik::getAction() === 'index' || !Piwik::getAction())) {
+            $tokenAuth = Common::getRequestVar('token_auth', '', 'string', $_POST);
+            if (!empty($tokenAuth)) {
+                Session::start();
+                $auth = StaticContainer::get(SessionAuth::class);
+                $auth->setTokenAuth($tokenAuth);
+                $result = $auth->authenticate();
+                // if not successful, we will fallback to regular auth
+            }
+        }
```
diosmosis: Is this for using the session during API requests? I remember when changing to sessions we didn't want to do this for some reason, do you remember why? Wondering if it's still relevant. It might've been for performance.

core/Piwik.php
```diff
@@ -262,6 +263,51 @@ public static function checkUserHasSuperUserAccessOrIsTheUser($theUser)
...
+
+        if (isset($token[$reason])) {
+            // note: For now we do not increase the expire time when it is already requested
+            return $token[$reason];
+        }
+
+        $twoWeeksInHours = 14 * 24;
+        if ($validForHours > $twoWeeksInHours) {
+            throw new Exception('The token cannot be valid for so many hours: ' . $validForHours);
+        }
```
diosmosis: Can we just set `$validForHours` to `$twoWeeksInHours` in this case? Might avoid some fatal errors in the future. (I can't imagine a case where someone would do this on purpose in a plugin, so I assume it would always be a strange bug, if it occurs at all.)

core/Piwik.php
```diff
@@ -262,6 +263,51 @@ public static function checkUserHasSuperUserAccessOrIsTheUser($theUser)
...
+     * Note: During one request the token is only being requested once and used throughout the request. So you want to make
+     * sure the token is valid for enough time for the whole request to finish.
+     *
+     * @param string $reason some short string/text explaining the reason for the token generation, eg "CliMultiAsyncHttpArchiving"
+     * @param int $validForHours For how many hours the token should be valid. Should not be valid for more than 14 days.
+     * @return mixed
+     */
+    public static function requestTemporarySystemAuthToken($reason, $validForHours)
+    {
+        static $token = array();
```
diosmosis: Would this ever need to be reset for tests? If so, might be good to keep it as a class static variable.

diosmosis:
Left some comments, have a pending review w/ some TODOs for myself when reviewing. Will test tomorrow and submit the review.

core/Access.php
tsteur: @diosmosis this is only for requests initiated by the UI. A standard API request will still not use the session. Was like the only workaround I could think of that doesn't break too much. It's actually not even so much of a tokenAuth here but more of an nonce.

Note to myself: Add BruteForce logic here.

core/Piwik.php
tsteur: I don't think it's needed for tests. If so, we can refactor it for sure 👍

core/Piwik.php
tsteur: Not sure @diosmosis personally, would prefer to throw an exception. It be only mostly under development anyway if someone calls it wrongly. Otherwise they might pause 4 weeks and don't notice it won't work because it's changed silently which might not be expected.

I could throw the exception though if needed only if Dev Mode is enabled if that helps? And fallback to 2 weeks if dev mode is disabled but then it might be hard to debug some issues.

diosmosis:
Noticed an issue when trying to test, the code seems to be trying to use `matomo_user_token_auth` before the update can run.

diosmosis:
Noticed something else: removing the token_auth column makes downgrading impossible. Not sure if it's an issue for users, but for development it makes it impossible to go back to 3.x-dev and test w/o adding the column back and then setting a token auth manually.

tsteur:
> Noticed something else: removing the token_auth column makes downgrading impossible. Not sure if it's an issue for users, but for development it makes it impossible to go back to 3.x-dev and test w/o adding the column back and then setting a token auth manually.

I could maybe instead of removing the column generate some random token_auth value and keep the existing column?

I can't keep the original `token_auth` value since it be a security issue as the token would then still be stored in plain text. The only way I can think of be to keep the column but set a random value for each token which is basically not in use as part of Matomo 4. Could be bit confusing but would allow to downgrade.

core/Piwik.php
diosmosis: We can always remove the exception later if it becomes an issue, so probably not an issue.

diosmosis:
> I could maybe instead of removing the column generate some random token_auth value and keep the existing column?

That sounds like it could work 👍

core/Session/SessionAuth.php
```diff
@@ -113,7 +117,17 @@ public function authenticate()
 
         $this->updateSessionExpireTime($sessionFingerprint);
 
-        return $this->makeAuthSuccess($user);
+        if (!empty($this->tokenAuth) && $this->tokenAuth !== $sessionFingerprint->getTempTokenAuth()) {
+            return $this->makeAuthFailure();
+        }
```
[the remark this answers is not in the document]
tsteur: I'll rename the method to `getSessionTokenAuth` 👍  it's not regenerated, only once per session

plugins/UsersManager/Model.php
```diff
@@ -43,7 +52,8 @@ class Model
     public function __construct()
     {
         $this->passwordHelper = new Password();
-        $this->table = Common::prefixTable(self::$rawPrefix);
+        $this->userTable = Common::prefixTable(self::$rawPrefix);
+        $this->tokenTable = Common::prefixTable('user_token_auth');
```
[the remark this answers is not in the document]
tsteur: @diosmosis I think we have this in various places like this and it was already there before for `$this->table` so it might not be an issue in this case?

tsteur:
@diosmosis applied the feedback now eg no longer removing the token_auth column. Also fixed the merge issues with 4.X and updated tests

plugins/UsersManager/Model.php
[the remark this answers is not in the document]
diosmosis: It's not a big deal, can always be changed if it becomes a problem.

diosmosis:
@tsteur looks like omnifixture might need an update? The UI tests are failing but I'm not sure why...

sgiehl:
@diosmosis updating omnifixture will likely change nearly all screenshots, due to various changes in tracked visits. Planning to do that in #15520. Will try to finish that somewhen this week. Might make more sense to update the omnifixture here afterwards to see if any test fails due to some changes here...

diosmosis:
@sgiehl makes sense, there was an odd error for this build but it seems unrelated to omnifixture (unless I forgot about another test). I'm in the process of fixing it.

tsteur:
Thanks for your help on this @diosmosis 💯

sgiehl:
@diosmosis @tsteur I'm getting an error now in the browser when the update is triggered (after checking out 4.x-dev branch). Guess the problem might be, that I'm not logged in. Doing a `core:update` on command line works
```
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] Uncaught exception: /srv/matomo/libs/Zend/Db/Statement/Pdo.php(234): SQLSTATE[42S02]: Base table or view not found: 1146 Table 'matomo.matomo_user_token_auth' doesn't exist
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] SQLSTATE[42S02]: Base table or view not found: 1146 Table 'matomo.matomo_user_token_auth' doesn't exist
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #0 /srv/matomo/libs/Zend/Db/Statement.php(300): Zend_Db_Statement_Pdo->_execute(Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #1 /srv/matomo/libs/Zend/Db/Adapter/Abstract.php(479): Zend_Db_Statement->execute(Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #2 /srv/matomo/libs/Zend/Db/Adapter/Pdo/Abstract.php(238): Zend_Db_Adapter_Abstract->query('SELECT * FROM m...', Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #3 /srv/matomo/core/Db/Adapter/Pdo/Mysql.php(309): Zend_Db_Adapter_Pdo_Abstract->query('SELECT * FROM m...', Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #4 /srv/matomo/libs/Zend/Db/Adapter/Abstract.php(755): Piwik\Db\Adapter\Pdo\Mysql->query('SELECT * FROM m...', Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #5 /srv/matomo/plugins/UsersManager/Model.php(334): Zend_Db_Adapter_Abstract->fetchRow('SELECT * FROM m...', Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #6 /srv/matomo/plugins/UsersManager/Model.php(419): Piwik\Plugins\UsersManager\Model->getTokenByTokenAuthIfNotExpired('299b4455fd7fe3b...')
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #7 /srv/matomo/plugins/Login/Auth.php(104): Piwik\Plugins\UsersManager\Model->getUserByTokenAuth('anonymous')
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #8 /srv/matomo/plugins/Login/Auth.php(62): Piwik\Plugins\Login\Auth->authenticateWithLoginAndToken('anonymous', 'anonymous')
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #9 /srv/matomo/core/Access.php(183): Piwik\Plugins\Login\Auth->authenticate()
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #10 /srv/matomo/core/FrontController.php(405): Piwik\Access->reloadAccess(Object(Piwik\Plugins\Login\Auth))
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #11 /srv/matomo/core/dispatch.php(33): Piwik\FrontController->init()
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #12 /srv/matomo/index.php(27): require_once('/srv/matomo/cor...')
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #13 {main},
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] caused by: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'matomo.matomo_user_token_auth' doesn't exist
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #0 /srv/matomo/libs/Zend/Db/Statement/Pdo.php(228): PDOStatement->execute(Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #1 /srv/matomo/libs/Zend/Db/Statement.php(300): Zend_Db_Statement_Pdo->_execute(Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #2 /srv/matomo/libs/Zend/Db/Adapter/Abstract.php(479): Zend_Db_Statement->execute(Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #3 /srv/matomo/libs/Zend/Db/Adapter/Pdo/Abstract.php(238): Zend_Db_Adapter_Abstract->query('SELECT * FROM m...', Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #4 /srv/matomo/core/Db/Adapter/Pdo/Mysql.php(309): Zend_Db_Adapter_Pdo_Abstract->query('SELECT * FROM m...', Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #5 /srv/matomo/libs/Zend/Db/Adapter/Abstract.php(755): Piwik\Db\Adapter\Pdo\Mysql->query('SELECT * FROM m...', Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #6 /srv/matomo/plugins/UsersManager/Model.php(334): Zend_Db_Adapter_Abstract->fetchRow('SELECT * FROM m...', Array)
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #7 /srv/matomo/plugins/UsersManager/Model.php(419): Piwik\Plugins\UsersManager\Model->getTokenByTokenAuthIfNotExpired('299b4455fd7fe3b...')
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #8 /srv/matomo/plugins/Login/Auth.php(104): Piwik\Plugins\UsersManager\Model->getUserByTokenAuth('anonymous')
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #9 /srv/matomo/plugins/Login/Auth.php(62): Piwik\Plugins\Login\Auth->authenticateWithLoginAndToken('anonymous', 'anonymous')
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #10 /srv/matomo/core/Access.php(183): Piwik\Plugins\Login\Auth->authenticate()
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #11 /srv/matomo/core/FrontController.php(405): Piwik\Access->reloadAccess(Object(Piwik\Plugins\Login\Auth))
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #12 /srv/matomo/core/dispatch.php(33): Piwik\FrontController->init()
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #13 /srv/matomo/index.php(27): require_once('/srv/matomo/cor...')
ERROR Piwik\ExceptionHandler[2020-03-18 08:42:37 UTC] [7964a] #14 {main}
```

PR #15423, opened 2020-01-20
Use correct name in update available message

### LINKED ISSUE #15417: In Update pop up venster Piwik is mentioned instead of Matomo
image

Suggestion: Change Piwik into Matomo

sgiehl:
could be included in next 3.x release

fixes #15417


tsteur:
Changed it to 4.x since it is not needed in 3.x but then many other changes would be merged into 4.x. So merging for now into 3.x. In 3.x we pretty much only want important fixes/regressions ideally.

Issue #15414, opened 2020-01-17
Since 3.13.1 autologin in an iframe does not work anymore


igugigu:
hy

I am showing Matomo inside an iFrame to my customers. Since I upgraded from 3.13.0 the auto login does not work anymore (HTTP 403).

If I copy the iframe url to a new tab the autologin works as expected.

Best regards

nik

In matomo.log I get:
```
ERROR API[2020-01-17 18:39:17 UTC] [57710] Uncaught exception in API: Piwik\NoAccessException: Ihre Sitzung ist aufgrund Inaktivität abgelaufen. Bitte melden Sie sich an um fortzufahren. in /srv/websites/tracker/piwik/core/Access.php:728
ERROR API[2020-01-17 18:39:17 UTC] [57710] Stack trace:
ERROR API[2020-01-17 18:39:17 UTC] [57710] #0 /srv/websites/tracker/piwik/core/Access.php(491): Piwik\Access->throwNoAccessException('Ihre Sitzung is...')
ERROR API[2020-01-17 18:39:17 UTC] [57710] #1 /srv/websites/tracker/piwik/core/Piwik.php(545): Piwik\Access->checkUserHasSomeViewAccess()
ERROR API[2020-01-17 18:39:17 UTC] [57710] #2 /srv/websites/tracker/piwik/plugins/LiveTab/API.php(23): Piwik\Piwik::checkUserHasSomeViewAccess()
ERROR API[2020-01-17 18:39:17 UTC] [57710] #3 [internal function]: Piwik\Plugins\LiveTab\API->getSettings()
ERROR API[2020-01-17 18:39:17 UTC] [57710] #4 /srv/websites/tracker/piwik/core/API/Proxy.php(237): call_user_func_array(Array, Array)
ERROR API[2020-01-17 18:39:17 UTC] [57710] #5 /srv/websites/tracker/piwik/core/Context.php(28): Piwik\API\Proxy->Piwik\API\{closure}()
ERROR API[2020-01-17 18:39:17 UTC] [57710] #6 /srv/websites/tracker/piwik/core/API/Proxy.php(328): Piwik\Context::executeWithQueryParameters(Array, Object(Closure))
ERROR API[2020-01-17 18:39:17 UTC] [57710] #7 /srv/websites/tracker/piwik/core/API/Request.php(265): Piwik\API\Proxy->call('\\Piwik\\Plugins\\...', 'getSettings', Array)
[...]

PR #15439, opened 2020-01-23
Use SameSite none for session token when embedded into iframe

tsteur:
fix https://github.com/matomo-org/matomo/issues/15414

Worked for me after enabling embed framed pages. It was setting the `None` as SameSite though instead of `Lax` and I was able to view reports within the frame etc.

Used this:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
</head>
<body>
<iframe width="100%" height="600" src="https://example.com/index.php?module=Login&action=logme&login=root&password=TOKEN"></iframe>

</body>
</html>
```

### DIFF
--- a/core/Session.php
+++ b/core/Session.php
@@ -186,6 +186,17 @@ class Session extends Zend_Session
         return self::$sessionStarted;
     }
 
+    public static function getSameSiteCookieValue()
+    {
+        $config = Config::getInstance();
+        $general = $config->General;
+        if (!empty($general['enable_framed_pages']) && ProxyHttp::isHttps()) {
+            return 'None';
+        }
+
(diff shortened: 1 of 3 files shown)


dm577:
Can this be tested on Chrome 80 with the "Samesite by default cookies" and "Cookies without Samesite must be secure" flags set as enabled? That is what the default behavior will be when Chrome 80 is launched on Feb 4 and if a cookie is set with SameSite=none without being marked as Secure, it will be blocked.

diosmosis:
Works for me after loading on demo2, both w/ and without the settings mentioned by @dm577

PR #15443, opened 2020-01-23
Ensure to close visitor popover correctly

### LINKED ISSUE #15404: Real time map dialog box does not close with x
I did upgrade to the new release Matomo 3.13.1
I have seen this before.
When opening the real time map and selecting one of the dots I can review the activity. When I click on the x in the top right corner nothing happens. I can click away from the box and it does close.

sgiehl:
fixes #15404

### DIFF
--- a/plugins/Live/javascripts/visitorProfile.js
+++ b/plugins/Live/javascripts/visitorProfile.js
@@ -71,14 +71,27 @@
             var self = this,
                 $element = this.$element;
 
-            $element.on('click', '.visitor-profile-close', function (e) {
-                e.preventDefault();
-                try {
-                    $element.tooltip('destroy');
-                } catch (e) {}
-                broadcast.propagateNewPopoverParameter(false);
-                return false;
-            });
+            // if there are no popover params in stack, simply close the popover
+            if (!broadcast.popoverParamStack.length) {
+                $element.on('click', '.visitor-profile-close', function (e) {
+                    e.preventDefault();
+                    try {
+                        $element.tooltip('destroy');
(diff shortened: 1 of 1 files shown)


plugins/Live/javascripts/visitorProfile.js
```diff
@@ -71,14 +71,27 @@
...
 
-            $element.on('click', '.visitor-profile-close', function (e) {
-                e.preventDefault();
-                try {
-                    $element.tooltip('destroy');
-                } catch (e) {}
-                broadcast.propagateNewPopoverParameter(false);
-                return false;
-            });
+            // if there are no popover params in stack, simply close the popover
```
tsteur: Haven't tested this one... but not quite understanding the logic... Eg I would have expected that when there is a popoverParamStack then we call the `Popover.close`? Or is the popover not added to the stack in that case? When is a popover added to the stack? I'm looking at the code for popover etc and it all looks a bit complicated to understand easily.

I would have maybe expected also some check if the profile is shown within a popover or not? like `self.$element.closest('#Piwik_Popover').length` or do we always show it within a popover so far?

Is `popoverParamStack ` basically used when the popover is opened through the hash in the URL?

sgiehl: The stack is modified when `broadcast.propagateNewPopoverParameter` is called and not when a popover is opened directly. When the method is called with a new popover parameter it's added to the stack. When the method is called with `false`, the last stack entry should be removed and the url hash will be updated to the next popover in stack or removed.

PR #15444, opened 2020-01-23
When closing a popover, leave opened any previously opened popover

sgiehl:
While investigating #15404 I discovered that the popover stack "feature" is kind of broken:

When opening popovers (segmented visitor log, visitor profile, ...) it is possible to open a popover from within another one. Those popovers are stored in a popover stack in javascript, to make it possible to open the previous popover if the current one is closed.
That is currently broken. When closing a popover it seems the stack is completely destroyed in any case.

With those changes the behaviour will be:

*Clicking outside of the popover*: complete stack will be closed
*Clicking the close button (x)*: current popover will be closed, previous will be opened (if any)
*Pressing Excape key*: complete stack will be closed

Is that expected, or should that be changed some more?

### DIFF
--- a/plugins/CoreHome/javascripts/popover.js
+++ b/plugins/CoreHome/javascripts/popover.js
@@ -35,6 +35,8 @@ var Piwik_Popover = (function () {
                 }
 
                 $('.ui-widget-overlay').on('click.popover', function () {
+                    // if clicking outside of the dialog, close entire stack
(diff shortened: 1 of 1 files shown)


mattab:
> Is that expected, or should that be changed some more?

 @sgiehl sounds like it is the ideal behavior now :+1:

PR #15445, opened 2020-01-23
Forward instance_id from local config when reseting config during tests.

diosmosis:
If this isn't done and you use a domain when running tests, the instance_id won't be used during tests. So the main PHPUnit execution context will use one tmp dir path (w/o the instance id), but other execution contexts (like climulti) will use a directory with a tmp dir path. Which can result in very hard to debug errors.


tsteur:
Feel free to merge if tests pass

PR #15450, opened 2020-01-24
Process unique visitor metric for Roll-ups (set enable_processing_unique_visitors_multiple_sites = 1)

mattab:
Alternatively, we could remove this setting completely and always assume we want to process unique visitors across websites. I'm not sure there is any reason not to want this enabled?


tsteur:
@mattab can you recreate the PR against 4.x-dev and then also fix the UI test (there is a diagnostics config test failing)

tsteur:
@mattab this PR will very likely cause issues because it will only work when `enable_fingerprinting_across_websites ` is also enabled. This is a privacy issue though. I reckon this PR should be reverted.

PR #15457, opened 2020-01-25
Add event that allows plugins to disable archiving for certain periods/sites if they want.

diosmosis:
New event is not public. Would be useful for the GA importer to make sure day archives are not re-archived, erasing imported data.

Added to 3.13.2 since it would be good for this functionality to work for users of 3.x installs, but I don't require it.

### DIFF
--- a/core/ArchiveProcessor/Rules.php
+++ b/core/ArchiveProcessor/Rules.php
@@ -226,9 +226,24 @@ class Rules
         return !$isArchivingEnabled;
     }
 
-    public static function isRequestAuthorizedToArchive()
+    public static function isRequestAuthorizedToArchive(Parameters $params = null)
     {
-        return Rules::isBrowserTriggerEnabled() || SettingsServer::isArchivePhpTriggered();
+        $isRequestAuthorizedToArchive = null;
+        if (!empty($params)) {
+            /**
+             * @ignore
+             *
+             * @params bool &$isRequestAuthorizedToArchive
+             * @params Parameters $params
+             */
+            Piwik::postEvent('Rules.isRequestAuthorizedToArchive', [&$isRequestAuthorizedToArchive, $params]);
+        }
+
+        if ($isRequestAuthorizedToArchive === null) {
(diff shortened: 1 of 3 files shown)


core/ArchiveProcessor/Rules.php
```diff
@@ -226,9 +226,24 @@ public static function isArchivingDisabledFor(array $idSites, Segment $segment,
...
-        return Rules::isBrowserTriggerEnabled() || SettingsServer::isArchivePhpTriggered();
+        $isRequestAuthorizedToArchive = null;
+        if (!empty($params)) {
+            /**
+             * @ignore
+             *
+             * @params bool &$isRequestAuthorizedToArchive
+             * @params Parameters $params
+             */
+            Piwik::postEvent('Rules.isRequestAuthorizedToArchive', [&$isRequestAuthorizedToArchive, $params]);
```
tsteur: Should we maybe use an event name starting with `Archiving` or so? See https://developer.matomo.org/api-reference/events#archivinggetidsitestoarchivewhennovisits ... just so they are next to each other maybe.

Also, I think usually we mostly would do 

```php
$isRequestAuthorizedToArchive = Rules::isBrowserTriggerEnabled() || SettingsServer::isArchivePhpTriggered();
```

First and put the result of this into the event argument instead of later checking for `null`. Not sure it's clear what you mean? This way eg the plugin can skip complex logic if the request is not authorized already anyway. Eg in the event it can do simple `if (!$isRequestAuthorizedToArchive) { return; }`. I think this is how we usually do it in most events. But no strong preference. Could avoid running GA Importer logic in almost every request for example?

diosmosis: > Should we maybe use an event name starting with Archiving or so?

Yes, that makes sense. `Rules` is a bit generic.

> First and put the result of this into the event argument instead of later checking for null. 

Yes, this should work. I remember deciding against this at first... but I can't remember why.

tsteur:
@diosmosis feel free to merge if tests pass 👍

PR #15462, opened 2020-01-26
Day range archiving issue

diosmosis:


core/Archiver/Request.php
```diff
@@ -65,12 +65,29 @@ public function getUrl()
     public function setUrl($url)
     {
         $this->url = $url;
+
+        // TODO: revisit in matomo 4
+        // period=range&date=last1/period=range&date=previous1 can cause problems during archiving due to Parameters::isDayArchive()
+        if (preg_match('/[&?]period=range/', $this->url)) {
+            $this->changeParam('period', 'day');
```
tsteur: @diosmosis not sure I understand. Wouldn't we only replace the period = range when date is last1 or previous1?

diosmosis: Yes, missed that

core/Archiver/Request.php
```diff
@@ -65,12 +65,29 @@ public function getUrl()
     public function setUrl($url)
     {
         $this->url = $url;
+
+        // TODO: revisit in matomo 4
+        // period=range&date=last1/period=range&date=previous1 can cause problems during archiving due to Parameters::isDayArchive()
+        if (preg_match('/[&?]period=range/', $this->url)) {
```
tsteur: @diosmosis could you maybe add a simple test?

tsteur (changes requested):
@diosmosis I reckon eg in https://github.com/matomo-org/matomo/blob/3.13.1/core/CronArchive.php#L1111 we need to change it to something like 

```php
$urlBefore = $request->getUrl(); 
$request->changeDate()...
 if ($request->getUrl() !== $urlBefore && $self->isAlreadyArchivingUrl($request->getUrl(), $idSite, $period, $date)) {
      return Request::ABORT;
}
```

The idea behind it be basically, that if it suddenly changes from `range` to `day`, then we need to check again if a `day` archive is already running and if so, abort it. 

Probably also needed here: https://github.com/matomo-org/matomo/blob/3.13.1/core/CronArchive.php#L1979-L1984

diosmosis:
@tsteur added a test and applied review feedback

core/CronArchive.php
```diff
@@ -1976,9 +1976,17 @@ private function getUrlsWithSegment($idSite, $period, $date)
                     return Request::ABORT;
                 }
 
-                $url = $request->getUrl();
-                $url = preg_replace('/([&?])date=[^&]*/', '$1date=' . $newDate, $url);
+                $urlBefore = $request->getUrl();
```
tsteur: @diosmosis does this also need to be done in https://github.com/matomo-org/matomo/blob/3.13.1/core/CronArchive.php#L1111 ?

core/Archiver/Request.php
```diff
@@ -68,9 +68,34 @@ public function setUrl($url)
     }
 
     public function changeDate($newDate)
+    {
+        $this->changeParam('date', $newDate);
+    }
+
+    public function makeSureDateIsNotSingleDayRange()
```
tsteur: Not sure if this changed in recent commit... shouldn't `setUrl` basically call this method?

core/CronArchive.php
diosmosis: Yes, will add

core/Archiver/Request.php
diosmosis: I thought about that then thought it might be a bad idea to do it every time the request url is set. Seems like it could cause strange side effects, and if we did, we'd have to check for an already running job after every setUrl() call.

diosmosis: I can move it back if needed, though.

tsteur: Shouldn't be needed I reckon 👍

PR #15464, opened 2020-01-27
Use latest davaxi/sparkline release

sgiehl:


sgiehl:
Latest version currently generates an warning. Create a PR to fix that.

PR #15465, opened 2020-01-27
Use latest davaxi/sparkline release

### LINKED ISSUE #15458: Update davaxi/sparkline composer dependency
Currently `davaxi/sparkline` is set to the version `dev-multiple-series` in `composer.json`

https://github.com/matomo-org/matomo/blob/5322d058d5afcd2c40e76a13762038e05f0795f3/composer.json#L51)

This results in the following error when running `composer require piwik/piwik`

```
Problem 1
    - piwik/piwik 3.13.1 requires davaxi/sparkline dev-multiple-series -> no matching package found.
```

I believe `davaxi/sparkline` should instead be set to `~1.2`, to install the [latest release](https://github.com/davaxi/Sparkline/releases/tag/1.2.0) that includes the changes merged to support [displaying multiple series](https://github.com/davaxi/Sparkline/pull/11).

sgiehl:
fixes #15458

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     "require": {
         "php": ">=7.2.0",
         "composer/semver": "~1.3.0",
-        "davaxi/sparkline": "dev-multiple-series",
+        "davaxi/sparkline": "~1.2",
         "geoip2/geoip2": "^2.8",
         "leafo/lessphp": "dev-php74-compat",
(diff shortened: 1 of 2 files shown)


sgiehl:
Latest version currently generates an error. Create a PR to fix that. See https://github.com/davaxi/Sparkline/pull/15

PR #15472, opened 2020-01-28
Use safemode when running CLI commands

diosmosis:
This would allow the controller method in plugins to be used when an error or exception occurs during a CLI command (which should result in an email on cloud, but I haven't tested).

### DIFF
--- a/console
+++ b/console
@@ -1,5 +1,8 @@
 #!/usr/bin/env php
 <?php
+
+use Piwik\FrontController;
+
 if (!defined('PIWIK_DOCUMENT_ROOT')) {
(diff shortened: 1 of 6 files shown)


mattab:
Is it possible to write a test for this @diosmosis ?

diosmosis:
Maybe, I'll give it a shot.

tsteur:
BTW Prefer to not have this in 3.x
We could always deploy it on cloud separately. or include it in next release. Just in case it maybe breaks something (eg around archiving, or in WP Matomo, etc)

mattab:
moved to Matomo 4 milestone

Issue #15477, opened 2020-01-28
Matomo can't track IE <9 any more


Findus23:
caused by https://github.com/matomo-org/matomo/pull/15184
reported in https://forum.matomo.org/t/update-3-13-1-no-tracking-in-internetexplorer-error-addeventlistener/35826

IE <9 doesn't support `addEventListener()`, instead a wrapper is used to make matomo.js compatible. But #15184 added a non-wrapped eventlistener which seems to break tracking.

https://github.com/matomo-org/matomo/blob/907a3d6f78693bcaadde7460fe0c2224161de63b/js/piwik.js#L6817

mattab:
This seems fixed, so closing :+1:

PR #15480, opened 2020-01-29
Make sure tracking works in IE9 and lower

tsteur:
fix https://github.com/matomo-org/matomo/issues/15477

Worked for me (couldn't see any error)
image


Findus23 (approved):
This change is confirmed to be working in https://forum.matomo.org/t/update-3-13-1-no-tracking-in-internetexplorer-error-addeventlistener/35826/6?u=lukas

PR #15483, opened 2020-01-29
fix: enable user-defined value for insertIntoTempIterationStep

toredash:


toredash:
I just noticed that 4-x.dev is the new default branch. Do you want me to change the target branch ?

tsteur:
@toredash that would be good. We're not accepting such PRs for 3.x anymore. Be good to close it and create new for 4.x

PR #15487, opened 2020-01-29
Sort Packages

millnut:
Enabled composer sort-packages config, DX improvement


tsteur:
@limitstudios cheers. Be great to just quickly mention the benefit of doing it this way. Does it help avoid conflicts or so?

limitstudios:
It's more of a readability thing, so when packages are added/removed they are sorted correctly

tsteur:
👍  thanks

PR #15491, opened 2020-01-30
fix: enable user-defined value for insertIntoTempIterationStep

toredash:


tsteur:
@toredash I believe there are some other commits there probably from 3.x-dev. Could you maybe create a PR with only the actual changes? That be great

Issue #15484, opened 2020-01-29
Missing documentation of deletelogs->delete_logs_unused_actions_schedule_lowest_interval


toredash:
Hi,

I'm unable to find any documentation on this config option:
delete_logs_unused_actions_schedule_lowest_interval

Feature was introduced in #13822

I've checked the forums and global.ini.php for help but there is no docs on its usage:
https://github.com/matomo-org/matomo/blob/162513d7608a43d53f178f744543e30dafc15f9b/config/global.ini.php#L910-L921

Reading from the code, I get that it is related to how often one should purge data in table `log_action`.

I've made a PR to add some docs to this, https://github.com/matomo-org/matomo/pull/15492, but also to add a user-definable value for how many rows to analyze in one iterations. Our `log_action` table contains >100M rows, so pruning data takes a while.

tsteur:
It's basically how much time is in between matomo tries to delete unused actions again

toredash:
Created a new PR for 4-x.dev branch. Original post updated.

Good to get a verification on the intended usage for the variable.

toredash:
closed by https://github.com/matomo-org/matomo/pull/15492

PR #15492, opened 2020-01-30
New INI setting delete_logs_unused_actions_max_rows_per_query useful for speeding up deletion process for very large sites

toredash:

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -910,12 +910,15 @@ enable_sql_profiler = 0
 [Deletelogs]
 ; delete_logs_enable - enable (1) or disable (0) delete log feature. Make sure that all archives for the given period have been processed (setup a cronjob!),
 ; otherwise you may lose tracking data.
-; delete_logs_schedule_lowest_interval - lowest possible interval between two table deletes (in days, 1|7|30). Default: 7.
+; delete_logs_schedule_lowest_interval - lowest possible interval between two table deletes, for table log_visit (in days, 1|7|30). Default: 7.
 ; delete_logs_older_than - delete data older than XX (days). Default: 180
+; delete_logs_unused_actions_schedule_lowest_interval - lowest possible interval between two table deletes, for table log_action (in days, 1|7|30). Default: 30.
+; *_max_rows_per_query - can be increased for large sites to speed up delete processes
 delete_logs_enable = 0
 delete_logs_schedule_lowest_interval = 7
 delete_logs_older_than = 180
 delete_logs_max_rows_per_query = 100000
+delete_logs_unused_actions_max_rows_per_query = 100000
 enable_auto_database_size_estimate = 1
 enable_database_size_estimate = 1
 delete_logs_unused_actions_schedule_lowest_interval = 30
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -912,7 +912,7 @@ enable_sql_profiler = 0
 ; otherwise you may lose tracking data.
 ; delete_logs_schedule_lowest_interval - lowest possible interval between two table deletes, for table log_visit (in days, 1|7|30). Default: 7.
 ; delete_logs_older_than - delete data older than XX (days). Default: 180
-; delete_logs_unused_actions_schedule_lowest_interval - lowest possible interval between two table deletes, for table log_action (in days, 1|7|30). Default: 30.
+; delete_logs_unused_actions_schedule_lowest_interval - lowest possible interval between two table deletes, for tables named log_* (in days, 1|7|30). Default: 30.
 ; delete_logs_max_rows_per_query and delete_logs_unused_actions_max_rows_per_query can be increased for large sites to speed up delete processes
 ;
 ; The higher value one assign to *_schedule_lowest_interval, the longer the data pruning/deletion will take. This is caused by the fact there is more data to evaluate and process every month, than every week.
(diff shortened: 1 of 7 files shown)


config/global.ini.php
```diff
@@ -910,12 +910,15 @@
 [Deletelogs]
 ; delete_logs_enable - enable (1) or disable (0) delete log feature. Make sure that all archives for the given period have been processed (setup a cronjob!),
 ; otherwise you may lose tracking data.
-; delete_logs_schedule_lowest_interval - lowest possible interval between two table deletes (in days, 1|7|30). Default: 7.
+; delete_logs_schedule_lowest_interval - lowest possible interval between two table deletes, for table log_visit (in days, 1|7|30). Default: 7.
```
tsteur: this would affect all log tables, not just log_visit

config/global.ini.php
```diff
@@ -910,12 +910,15 @@
...
 ; delete_logs_older_than - delete data older than XX (days). Default: 180
+; delete_logs_unused_actions_schedule_lowest_interval - lowest possible interval between two table deletes, for table log_action (in days, 1|7|30). Default: 30.
+; *_max_rows_per_query - can be increased for large sites to speed up delete processes
```
tsteur: btw... the longer you set the value, the longer the delete will take, the longer the table might be locked (which will cause incoming tracking requests to pile up since they can't finish and are waiting for the lock). It's probably not an issue with InnoDB since it should have row locking though.

tsteur: btw not seeing this one used anywhere, are you still working on the PR?

plugins/PrivacyManager/API.php
```diff
@@ -218,14 +218,15 @@ public function setDeleteReportsSettings($enableDeleteReports = 0, $deleteReport
...
-        $settings['delete_logs_max_rows_per_query']      = PiwikConfig::getInstance()->Deletelogs['delete_logs_max_rows_per_query'];
+        $settings['delete_reports_keep_basic_metrics']             = (int) $keepBasic;
+        $settings['delete_reports_keep_day_reports']               = (int) $keepDay;
+        $settings['delete_reports_keep_week_reports']              = (int) $keepWeek;
+        $settings['delete_reports_keep_month_reports']             = (int) $keepMonth;
+        $settings['delete_reports_keep_year_reports']              = (int) $keepYear;
+        $settings['delete_reports_keep_range_reports']             = (int) $keepRange;
+        $settings['delete_reports_keep_segment_reports']           = (int) $keepSegments;
+        $settings['delete_logs_max_rows_per_query']                = PiwikConfig::getInstance()->Deletelogs['delete_logs_max_rows_per_query'];
+        $settings['delete_logs_unused_actions_max_rows_per_query'] = PiwikConfig::getInstance()->Deletelogs['delete_logs_unused_actions_max_rows_per_query'];
```
tsteur: AFAIK it shouldn't be needed to add this here?

config/global.ini.php
tsteur: I see now, you're only adding it for documentation purposes 👍

tsteur (changes requested):

config/global.ini.php
toredash: Yes :) I'll add some more docs as well as I think it is valuable to have there

plugins/PrivacyManager/API.php
toredash: I'll remove it then :) thanks

toredash:
closes #15484

config/global.ini.php
tsteur: Not sure why it was resolved. Should we mention it is applied to all `log_*` tables meaning all tables that start with `log_`?

plugins/PrivacyManager/Controller.php
```diff
@@ -59,14 +59,15 @@ private function getPurgeSettingsFromRequest()
...
-        $settings['delete_logs_max_rows_per_query']      = PiwikConfig::getInstance()->Deletelogs['delete_logs_max_rows_per_query'];
+        $settings['delete_reports_keep_basic_metrics']             = Common::getRequestVar("keepBasic", 0);
+        $settings['delete_reports_keep_day_reports']               = Common::getRequestVar("keepDay", 0);
+        $settings['delete_reports_keep_week_reports']              = Common::getRequestVar("keepWeek", 0);
+        $settings['delete_reports_keep_month_reports']             = Common::getRequestVar("keepMonth", 0);
+        $settings['delete_reports_keep_year_reports']              = Common::getRequestVar("keepYear", 0);
+        $settings['delete_reports_keep_range_reports']             = Common::getRequestVar("keepRange", 0);
+        $settings['delete_reports_keep_segment_reports']           = Common::getRequestVar("keepSegments", 0);
+        $settings['delete_logs_max_rows_per_query']                = PiwikConfig::getInstance()->Deletelogs['delete_logs_max_rows_per_query'];
+        $settings['delete_logs_unused_actions_max_rows_per_query'] = PiwikConfig::getInstance()->Deletelogs['delete_logs_unused_actions_max_rows_per_query'];
```
tsteur: That entry might not be needed here? I suppose there is also no harm to have it here though

toredash: I'm not sure, I only added it since I saw there was an entry for `['delete_logs_max_rows_per_query']`. It didn't really make sense for me that it was there.

@tsteur do you want me to remove it or ?

tsteur: Just had a look . Actually, it's fine to have it there and might be even needed after all.

config/global.ini.php
toredash: This is fixed in the latest commit

plugins/PrivacyManager/API.php
toredash: Given that we have the new setting present in Controller.php, I assume we need it here as well.

toredash (commented):
I believe everything you have pointed out is resolved, @tsteur

tsteur:
Thanks @toredash very appreciated

PR #15493, opened 2020-01-30
Fixed lang sk typo in translation.

branislavballon:
Sk typo in translations file for key Outlinks. In SK there is no "Konkajšie" but Vonkajšie. Word mens outer or external.


tsteur:
Thanks for the PR @branislavballon . We don't manage translations here in Github unfortunately and can't even merge it since it would cause issues :( 

Our translations are managed through a system called Transifex. If you're interested in changing translations in Matomo be great if you could go to https://matomo.org/translations and reach out to us via this page. We can invite you to our translation system.

cc @sgiehl

PR #15496, opened 2020-01-31
Updated supported currencies: Added Bitcoin

cd2357:
For sites that denominate their Goals in BTC

### DIFF
--- a/core/Intl/Data/Resources/currencies.php
+++ b/core/Intl/Data/Resources/currencies.php
@@ -176,4 +176,7 @@ return array(
     'XOF' => array('Fr', 'West African CFA franc'),
     'YER' => array('﷼', 'Yemeni rial'),
     'ZMW' => array('ZK', 'Zambian kwacha'),
+    
+    // Crypto-currencies
+    'BTC' => array('₿', 'Bitcoin'),
 );


tsteur:
Looks good to me. Any thoughts @sgiehl ?

sgiehl:
the file currently only holds currencies, that have a valid iso code. None of the crypto currencies currently has a valid iso code.
Also Bitcoin is already predefined in global.ini.php. See https://github.com/matomo-org/matomo/blob/b247f8538cd3b67b9145997f5aca929b87f4eef5/config/global.ini.php#L267-L269

So as long as that config value is not overwritten somewhere, it should already be available without that change.

@cd2357 wasn't bitcoin available for you without that change? Or what's the purpose?

cd2357:
Well, I had just recently installed a fresh version of Matomo 3.13.1. I noticed that under Websites > Manage, there's a currency field ("The Currency symbol will be displayed next to Goals revenues.")

Now since I wanted to track goal revenues in BTC, I figured I'd select or somehow enter that. But the list of available currencies didn't show it. So I figured I'd add it. That's the background behind the commit.

sgiehl:
Interesting. For me `Bitcoin` appears in the list of currencies without that change 🤔

tsteur:
It's definitely already there for me as well. Maybe double check your `config/global.ini.php` whether it mentions that currency? I'll close this for now as it should work. Happy to reopen or further discuss if there's a problem @cd2357

Issue #12420, opened 2018-01-10
Finish renaming Piwik to Matomo everywhere


mattab:
In January 2018 [we announced that Piwik is now Matomo](https://matomo.org/blog/2018/01/piwik-is-now-matomo/) :tada: 

As part of the renaming efforts we're updating all our websites, tools, core app, plugins, ie. everything! 

Renaming checklist left to do:

* [ ] HTTP API methods eg. `ExampleAPI.getPiwikVersion`
* [ ] plugin `CustomPiwikJs` to be renamed eg. `CustomJsTracker`
* [ ] HTTP Tracking API endpoints: `piwik.js` and `piwik.php` 
* [ ] JavaScript tracking sdk with objects like `Piwik` and `window.Piwik` and methods like `getPiwikUrl`
* [ ] Plugin developers, core platform uses namespace `Piwik` + some objects called "Piwik", an event `Piwik.getJavascriptCode`
* [ ] cookie names eg `piwik_auth`
* [ ] URL parameter like `piwik_campaign` to also support `matomo_campaign`
* [ ] DB prefix like `piwik_`
* [ ] Marketplace URL parameters like "piwik_version"
* [ ] plugin.json parameters like require "piwik"
* [ ] ... and more

In the case of Matomo core platform we want to keep things backward compatible for users including HTTP APIs users or users of the JavaScript SDKs. 

Assigning to Matomo 4.0.0 but hopefully we can tackle a few of the list, and make progress on the others, during Matomo 3.x :rocket:

linusg:
Hey there,
not sure if there's a separate repo for the website (matomo.org, didn't found one), but I think it will fit in here.

On the main page, there's this animation:

![image](https://user-images.githubusercontent.com/19366641/42729324-d483a52a-87d4-11e8-9b66-0149cc4a8234.png)

On the truck it say's Piwik, so that should be changed at one point in the future. :-)

Cheers!

tsteur:
As part of this issue it be important to look at which ones are easy and quick to do. The rest we'll do as part of Matomo 5.

tsteur:
for tracker created https://github.com/matomo-org/tracker-proxy/pull/61

tsteur:
I reckon we can move the rest to Matomo 5? The only question would be wether we want to release another Matomo 2 and Matomo 3 update with the legacy auto loader: https://github.com/matomo-org/matomo/blob/4.x-dev/LegacyAutoloader.php#L17

[...]

PR #15504, opened 2020-02-03
Removes Piwik is now Matomo from update screen

sgiehl:
refs #12420


tsteur:
👍 feel free to merge, might need to update UI test?

Issue #7054, opened 2015-01-20
core:archive does not work on Solaris


mkobel:
The command `ps ex` (in core/CliMulti/Process.php in Method isProcessStillRunning) cannot be executed on Solaris.
Changing this command to `ps -e` it works.

tsteur:
It used to use `-e` but the dash was removed while working on #5842 see https://github.com/piwik/piwik/commit/ca84f227dc1de30fd96e76ab5e7f61d5d5adca87

@diosmosis do you remember why it was removed? If we have to execute it this way maybe we could run `ps ex` first and if it fails use `ps -ex`?

mnapoli:
Could we consider using something like [Symfony's Process](http://symfony.com/doc/current/components/process.html)? Or is there a particular reason we don't use it? They seem to have an alternative solution [using `proc_get_status()`](https://github.com/symfony/Process/blob/master/Process.php#L1282-L1296), that would (hopefully) prevent us maintenance and bugfixes like this.

tsteur:
Probably not, the latest version requires `PHP 5.5.9`

diosmosis:
There was an issue w/ using `'-...'` on BSD. I can't quite remember what happened, but wikipedia says "Most systems derived from BSD fail to accept the SUS and POSIX standard options because of historical conflicts (for example, the "e" or "-e" option will cause the display of environment variables)." [1]. Sounds familiar.

[1] http://en.wikipedia.org/wiki/Ps_%28Unix%29

mnapoli:
Symfony 3 will require 5.5, but Symfony 2 requires 5.3.3.

tsteur:
We could give it a try then. We don't use it much so would be easy to replace for a test although there are a few warnings in the docs to consider which makes it maybe a bit more work. How long will they support Symfony 2? I'm pretty sure we looked that time into it and there was something why we couldn't use it. 

It would be nice if it still worked on the same systems as it does today. I can't find anything about this in docs or composer so presume it just works everywhere. By looking at the code we'll have to patch their [PHP executable finder](https://github.com/symfony/Process/blob/master/PhpExecutableFinder.php) as it seems to not correctly find a valid php that we can use (eg cgi, fpm, ...). Probably we could also inject another dependency there and use our own instead. We should also make sure their code makes sure it detects the end of a process in all cases to prevent ending in a never ending loop. We might have to make more adjustments here.

mattab:
[...]

PR #15508, opened 2020-02-03
Fix archiving with console on SELinux: use `ps x` instead of `ps ex`

ghost:
Resolves ps errors when archiving as an account does not have access to /proc/ (typical on SELinux).  Change 'ps ex' to 'ps x' for compatibility with SELinux.  
Example of this issue in forum:  https://forum.matomo.org/t/selinux-errors-about-ps/14540

### DIFF
--- a/core/CliMulti/Process.php
+++ b/core/CliMulti/Process.php
@@ -242,7 +242,7 @@ class Process
 
     public static function getListOfRunningProcesses()
     {
-        $processes = `ps ex 2>/dev/null`;
+        $processes = `ps x 2>/dev/null`;
         if (empty($processes)) {
             return array();
         }
@@ -254,7 +254,7 @@ class Process
      */
      public static function getRunningProcesses()
      {
-         $ids = explode("\n", trim(`ps ex 2>/dev/null | awk '! /defunct/ {print $1}' 2>/dev/null`));
+         $ids = explode("\n", trim(`ps x 2>/dev/null | awk '! /defunct/ {print $1}' 2>/dev/null`));
 
          $ids = array_map('intval', $ids);
          $ids = array_filter($ids, function ($id) {


tsteur:
I'm not sure re possible consequences of that. Eg I reckon that for some users this could maybe break things where a different job runs under a different user/session? 

Maybe someone knows a bit more. The only other thing I could imagine is to maybe first try `ps -ex` and if this doesn't work run `ps -x` maybe?

Maybe someone else knows more? @Findus23 maybe?

diosmosis:
I can't remember if this was the reason for `ex`, but years ago there was an issue on FreeBSD and to fix it we had to change the command. This might break for that or other systems.

core/CliMulti/Process.php
```diff
@@ -242,7 +242,7 @@ private static function isProcFSMounted()
 
     public static function getListOfRunningProcesses()
     {
-        $processes = `ps ex 2>/dev/null`;
+        $processes = `ps x 2>/dev/null`;
```
tsteur: I wonder if we could do something like 

```php
$returnCode = null;
$processes = system(ps ex 2>/dev/null, $returnCode);
if ($returnCode != 0) {
     $processes = system(ps x 2>/dev/null, $returnCode);
}
```

? And same below?

leftmostlane: Running -ex on secure linux increases the size of the audit log.  It is a policy violation.  Perhaps run 'x' first?  Least privileged first?

diosmosis: I guess we could do this? Though if `ps ex` fails for a temporary reason and `ps x` fails because the OS is wrong that could result in some strange error messages. Would be a bit of an edge case though. Probably better than trying to detect the correct system/command to use. (I guess we could also allow the command options to be specified through config or something.)

tsteur: @leftmostlane sorry for getting back to you so late. Any chance you could look at the approach above where we fall back to `ps x` if `ps ex` doesn't work?

leftmostlane: Running 'ex' on secure Linux increases the size of the audit log and is a policy violation. On shared systems this fills up the audit log with access violations.  Perhaps run 'x' first? Test for least privileged first then fall back to 'ex' if that does not work.  
```
$returnCode = null;
$processes = system(ps x 2>/dev/null, $returnCode);
if ($returnCode != 0) {
     $processes = system(ps ex 2>/dev/null, $returnCode);
}
```

tsteur: Actually I did some various tests on different OS and some more research and it should be fine to use only `x`.

I also went back the history (took a while) and noticed I had added `ps -e` about 7 years ago in https://github.com/matomo-org/matomo/commit/113e403a9e5a6ccab2762fc321382b540b51f8d9#diff-daeea28082bcbb1d6765684dc7f99fb2R92

It was initially `-e` as in "Select all processes.".

Then it became `ps -ex`.

And then it became `ps ex` in https://github.com/matomo-org/matomo/commit/ca84f227dc1de30fd96e76ab5e7f61d5d5adca87

Which I think changes the meaning from `select all processes` to `Show the environment after the command.`. However, we don't need the environment. I think it was meant to be changed to `ps -e x` if that makes any sense and I understand things correctly.

The referenced issue says
```
On some systems the command 'ps -e' will not return processes that are run via cron (since they are not attached to a terminal). This causes the ./console climulti:request command to abort early and return no data, which in turn causes ./console core:archive to fail.
```

Which caused the change from `ps -ex` to `ps ex`. Unfortunately the referenced forum link does no longer work.

There is also https://github.com/matomo-org/matomo/issues/7054 which suggests to use `ps -e`. 

I'm planning to merge this PR as is within the next 7 days or so. Hoping it also fixes things on solaris maybe.

Otherwise we'd need to do things like `php_uname ()` and adjust the command based on the OS.

If anyone has some further insights be great to let me know.

leftmostlane: Sounds great!

PR #15512, opened 2020-02-04
Resolves issue #15120

peter279k:
# Changed log
- Recreate PR #15122.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,39 +54,13 @@
     },
     "require-dev": {
         "aws/aws-sdk-php": "2.7.1",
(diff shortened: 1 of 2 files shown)


sgiehl:
@peter279k  seems you updated all packages in the composer.lock. Would be better to keep the currently used versions there and only replace the xhprof package

peter279k:
> @peter279k seems you updated all packages in the composer.lock. Would be better to keep the currently used versions there and only replace the xhprof package

Fix it :).

sgiehl (approved):

Issue #15179, opened 2019-11-21
heartbeat/pingreq. does not update log_visit.visit_last_action_time


toredash:
I believe there is an issue with how ping requests are processed.

According to the source `visit_last_action_time` holds the visit´s end time:
```
/**
 * This dimension holds the best guess for a visit's end time. It is set the last action
 * time for each visit. `ping=1` requests can be sent to update the dimension value so
 * it can be a more accurate guess of the time the visitor spent on the site.
 *
 * Note: though it is named 'visit last action time' it actually refers to the visit's last action's
 * end time.
 */
```
https://github.com/matomo-org/matomo/blob/6b4cd1676d524641167ba32776a999e3e37a6a2a/plugins/CoreHome/Columns/VisitLastActionMinute.php

Looking at how the PingRequestProcessor (https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/Heartbeat/Tracker/PingRequestProcessor.php) handles hearbeats, it seems it does not update `log_visit.visit_last_action_time`, it only updates `log_visit.visit_total_time`.

`log_visit.visit_last_action_time` is only updated if an action is sent. I think the issue is that `PingRequestProcessor` sends an empty `Actions`:
https://github.com/matomo-org/matomo/blob/2.16.0-b6/plugins/Heartbeat/Tracker/PingRequestProcessor.php#L24-L29

When viewing a users visits, the "x actions in <duration>" is correct, as it uses `log_visit.visit_first_action_time` + `log_visit. visit_total_time`.

Example of this:
![image](https://user-images.githubusercontent.com/312038/69317158-a0955880-0c3a-11ea-9ed6-4fff7c8180eb.png)

For us, this becomes an issue an issue when using CustomReports and `Server Time - Hour/Minute (End Of Visit)`, as it using `log_visit.visit_last_action_time` (https://github.com/matomo-org/matomo/blob/3.x-dev/plugins/CoreHome/Columns/VisitLastActionMinute.php#L32) to determine the visis end time. This occurs since ping requests does not update `log_visit.visit_last_action_time`. With the current code, it is not possible to determine when a visit ended for SPA apps, since it seems only an action can that trigger an update of `log_visit.visit_last_action_time`.

I'm not sure if this is a bug in the PingRequestProcessor, or an issue with how `VisitLastActionMinute` is expected to compute the visits end time.

Issues i think are relevant:
#9626 #9748

tsteur:
[...]

PR #15518, opened 2020-02-05
fix: change description for segments visitEndServer*

toredash:
visitEndServer* does not show when a visits ends, but when the last action for the visit was. 

Modifications for other languages should be done as well


toredash:
closes #15179

tsteur:
👍 @mattab can you have a quick look here?
@toredash we'll need to fix some tests like the ones failing here. https://travis-ci.org/matomo-org/matomo/jobs/646313431?utm_medium=notification&utm_source=github_status  Have you done this before? We might be able to merge it and then fix the tests in a separate PR

toredash:
@tsteur  No I haven't done this before. Looked at the builds and I altered one test that is related to this PR. Waiting for the build results now

toredash:
@tsteur I'm not sure where I go in the code to resolve these;
https://travis-ci.org/matomo-org/matomo/jobs/647297552#L966

If you could point me to the correct location that would probably help, thanks

mattab:
@toredash to fix this test, you can do the same as you did for the other test ie. overwrite the files within the folder `plugins/Live/tests/System/expected/test_actionSegment__*.xml`  by the files that were in the `plugins/Live/tests/System/processed` folder on your machine after you ran the tests and generated the new correct responses.

tsteur:
@toredash I'll merge this now in a branch on our side so I can fix the tests afterwards. THanks for this 👍

PR #15520, opened 2020-02-06
Regenerates Omnifixture

sgiehl:
It's been about 9 months since the Omnifixture has been regenerated. As there have been a lot changes to various fixtures, there are a lot changes in the exptected UI files. Especially in the tracked visitors & numbers


sgiehl:
Most tests are working again, but there are a few changes left that don't seem correct.

@tsteur @diosmosis  Do you know if that change is correct:
https://builds-artifacts.matomo.org/matomo-org/matomo/omnifixture/38167/UIIntegrationTest_ecommerce_log_segmented.png

The additional action says something about that it shouldn't be tracked in that visit. Not sure if that changed in the last months or if something is broken 🤔

tsteur:
Looking at the `FewVisitsWithSetVisitorId` it's hard to tell. There is some different visitorId, IP, and User Agent set so would have expected this would be a new visit as also the configId should not match.

sgiehl:
Had another look. Seems that change is related to #14360. Will check and improve the test, so it does wat it is intended to.

plugins/CoreHome/javascripts/broadcast.js
```diff
@@ -804,8 +804,8 @@ var broadcast = {
             var value = url.substring(startPos + lookFor.length, endStr);
 
             // we sanitize values to add a protection layer against XSS
-            // &segment= value is not sanitized, since segments are designed to accept any user input
-            if(param != 'segment') {
+            // &segment= (and &popover=) value is not sanitized, since segments are designed to accept any user input
+            if(param != 'segment' && param != 'popover') {
```
sgiehl: This change was required in order to fix a UI test with XSS values. The popover param made problems for row evolution with labels containing special chars. Closing the popover didn't remove the param from url without that change.

sgiehl:
Omnifixture is up to date with latest 4.x-dev. Some tests are still failing.
* OneClickUpdate -> automatic login after update does not work for any reason (also on 4.x-dev)
* UIIntegrationTest -> metric tooltip fails randomly - didn't find a good way to fix that properly 🤷‍♂ 

and two tests using the segment editor are failing because of #15719. Those should work again afterwards

diosmosis (approved):
The files and image changes look good, I guess we can create an issue to fix the OneClickUpdate test? It's probably some issue that needs to be fixed when updating from 3.x to 4.x via one click update.

PR #15521, opened 2020-02-06
Removes GeoIp Legacy support

### LINKED ISSUE #13957: Remove GeoIP Legacy Implementation and switch users to Db-ip
Maxmind has completely discontinued GeoIP Legacy Lite support. No databases in that format are available for free download anymore. Commercial databases might still be up (don't have access, so can't check if that's the case)

See https://support.maxmind.com/geolite-legacy-discontinuation-notice/

Therefor we completely should remove the GeoIP Legacy support in one of the upcoming version or a least in 4.0.

The GeoIP Legacy Autoupdater should already be removed (or automatically be disabled if a GeoLite DB is configured) in one of the next releases as it can't work anymore.

sgiehl:
Note: For users still using GeoIp Legacy, the update will switch their provider back to the default one.

fixes #13957


core/Updates/4.0.0-b1.php
```diff
@@ -48,11 +50,35 @@ public function getMigrations(Updater $updater)
...
         }
 
+        if ($this->usesGeoIpLegacyLocationProvider()) {
+            // activate GeoIp2 plugin for users still using GeoIp2 Legacy (others might have it disabled on purpose)
+            $migrations[] = $this->migration->plugin->activate('GeoIp2');
+        }
+
+        // remove old options
+        $migrations[] = $this->migration->db->sql('DELETE FROM `' . Common::prefixTable('option') . '` WHERE option_name IN ("geoip.updater_period", "geoip.loc_db_url", "geoip.isp_db_url", "geoip.org_db_url")');
+
```
diosmosis: Should there be a migration that deletes the old GeoIp plugin?

plugins/UserCountry/tests/UI/UserCountry_spec.js
```diff
@@ -0,0 +1,35 @@
...
+describe("UserCountry", function () {
+
+    afterEach(function () {
+        delete testEnvironment.unloadGeoIp2;
+        testEnvironment.save();
+    });
+
+    it('should show geolocation admin without additional providers', async function () {
+        testEnvironment.unloadGeoIp2 = 1;
+        testEnvironment.save();
```
diosmosis: minor suggestion: instead of a `unloadGeoIp2` setting, we could have a `pluginsToUnload` setting that could be re-used in other contexts for other plugins. what do you think of that idea?

diosmosis:
Tested locally, seems to work for me, even going back to 3.x-dev. Left a couple comments otherwise.

core/Updates/4.0.0-b1.php
sgiehl: There was no `GeoIp` plugin. That stuff was included in the `UserCountry` plugin, which still exist

plugins/UserCountry/tests/UI/UserCountry_spec.js
sgiehl: makes sense. change that.

core/Updates/4.0.0-b1.php
diosmosis: 👍 missed that

PR #15125, opened 2019-11-06
Add support for up to 4 fields to multipairfield

MichaelRoosz:
Adds support for up to 4 fields in the multipairfield config template.
Needed for another pull requests for Tag Manager (support for Custom Variables https://github.com/matomo-org/tag-manager/pull/201).

Maybe the changes in "plugins/CorePluginsAdmin/angularjs/form-field/field-multituple.html" are not required? I am not 100% sure.

PR #15531, opened 2020-02-07
Add support for up to 4 fields to multipairfield

sgiehl:
recreates #15125

### DIFF
--- a/plugins/CoreHome/angularjs/multipairfield/multipairfield.controller.js
+++ b/plugins/CoreHome/angularjs/multipairfield/multipairfield.controller.js
@@ -34,8 +34,16 @@
             $scope.field2.templateFile = getTemplate($scope.field2);
         }
 
+        if ($scope.field3 && !$scope.field3.templateFile) {
+            $scope.field3.templateFile = getTemplate($scope.field3);
+        }
+
+        if ($scope.field4 && !$scope.field4.templateFile) {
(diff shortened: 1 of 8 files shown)


sgiehl:
@tsteur remaining UI failures are unrelated. Guess would be good to merge.

plugins/CoreHome/angularjs/multipairfield/multipairfield.controller.js
```diff
@@ -63,6 +79,19 @@
                         hasAny = false;
                     }
                 }
+
```
tsteur: Not sure the above `else if ($scope.field2) {` is still needed? Could maybe remove the feature that only a field2 is defined but no field1? If I understand the logic correct?

BTW could slightly make above block more readable by using eg 
image

tsteur (changes requested):
Left a comment. Be good to look at it. Also there's some other issue see video. You enter a value and then the other fields get the same value 

Feb-10-2020 07-59-44

plugins/CoreHome/angularjs/multipairfield/multipairfield.controller.js
sgiehl: adjusted the code accordingly

sgiehl:
> Left a comment. Be good to look at it. Also there's some other issue see video. You enter a value and then the other fields get the same value

That was because the `keys` of all fields were the same. adjusted that...

PR #15541, opened 2020-02-10
Adds translation updates action

sgiehl:
Adds an action to create a PR with translation updates once a week (Saturday 2:00 UTC)

Attention: In order to get the action working someone with admin access needs to add two secrets to this repository:

`TransifexUsername`: Containing the username of a transifex account with access to matomo
`TransifexPassword`: Password for that account

Those are required to fetch the translations from transifex API

### DIFF
--- a/.github/workflows/translations.yml
+++ b/.github/workflows/translations.yml
@@ -1,0 +1,143 @@
+name: Translation Updates
+
+on:
+  schedule:
+  - cron: "0 2 * * 6"
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    services:
+      mysql:
+        image: mysql:5.7
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: true
+          MYSQL_ROOT_PASSWORD: ""
+          MYSQL_DATABASE: piwik_tests
+        ports:
+          - 3306/tcp
+
+
+    steps:
+    - uses: shivammathur/setup-php@v1
+      with:
+        php-version: '7.3'
+    - uses: actions/checkout@v2
+      with:
+        ref: '4.x-dev'
+        lfs: false
+    - name: Prepare branches
+      run: | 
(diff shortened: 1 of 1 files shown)


tsteur:
What will be the best way to test this? Re the transifex account: Do we need to sign up a new user there for this with limited access?

sgiehl:
I've tested that on a fork as well. The last two translation PRs were actually triggered from my fork. The script here is slightly different, as the PR goes to the same repo...

Actually it needs to be an account which has access to all languages. Once the secret is added it can't be seen anywhere, so guess should be fine to add it. We can use mine, but I can't access the repo settings, so can't add it...

tsteur:
Can we create an access token that basically has only read access? Just thinking if there's some security issue in Github or between github and transifex then someone would have only read access. It's very unlikely something would happen but might be better to be safe? Not sure how easy it be to do it or what transfix can maybe do there already. Or can it use a github token?

sgiehl:
It's possible to generate API tokens on transifex and use them (together with the username `api`), but you can't limit the access. Alternatively there is a direct GitHub integration for transifex, but then we would need to grant transifex access to github, which imho isn't better...

tsteur:
I see. If I understand correctly, we already have some `api` user we can use and this user maybe only has view access? Otherwise could quickly sign up a user on github and connect it to transifex and give "read" access to all languages? Or maybe transifex doesn't even have something like read access in general. Just wanting to avoid somehow being able to delete all languages, all translators and all data in there. We'd still have the files with all translations anyway so that's not an issue anyway but might be an issue to recover all translators and get them in there again? I suppose we have a list anyway so it might be all fine. I'll follow up in slack re the token.

sgiehl:
@tsteur is it fine to merge? we can then simply see if the action runs correctly on the next weekend. If not we can simply adjust or remove it again...

tsteur:
will merge now 👍

PR #15542, opened 2020-02-10
Adds action for building tracker js files

sgiehl:
This action will make it possible to automatically update the minified tracker js files in a Pull Request.
It is triggered with a comment "build js" within a Pull Request.

It (currently) only works for Pull Requests that are based on a local branch.

### DIFF
--- a/.github/workflows/buildtrackerjs.yml
+++ b/.github/workflows/buildtrackerjs.yml
@@ -1,0 +1,82 @@
+name: Build Tracker JS files
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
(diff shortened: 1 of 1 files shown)


tsteur:
What will be the best way to test this?

sgiehl:
Hm... I've developed & tested those actions on a fork. They need to be placed in the default branch (4.x-dev) to be executed

tsteur:
OK feel free to merge tmrw and in worst case simply revert 👍

tsteur:
OK feel free to merge tmrw and in worst case simply revert 👍

PR #15552, opened 2020-02-11
Move Provider plugin to the Marketplace 

### LINKED ISSUE #5341: Move Provider plugin to the Marketplace
The Provider plugin is currently part of core Piwik platform. 

Let's move it to the [Marketplace](http://plugins.piwik.org/) instead.
- Remove plugin from core
- Move file to the [marketplace](http://plugins.piwik.org/)
- On upgrade:
  - if Provider plugin is disabled, then delete Provider files.
  - if Provider plugin is enabled, make sure it still works and that the plugin can be later upgraded to latest marketplace version

tsteur:
fix #5341

plugin is now its own repo https://github.com/matomo-org/plugin-Provider


.gitmodules
```diff
@@ -72,4 +72,7 @@
 	path = plugins/Morpheus/icons
 	url = https://github.com/matomo-org/matomo-icons.git
 
-# Note: do not add new plugin submodules here, but a few lines above
\ No newline at end of file
+# Note: do not add new plugin submodules here, but a few lines above
```
sgiehl: any reason you did't add that in the list above (as the comment suggests)? 🤔

sgiehl (changes requested):
I'm wondering what happens when doing the update. I guess the plugin wouldn't be removed automatically, even though it isn't included in the archive anymore. But actually it would then lack the `plugin.json` and thus it would be reported as not compatible? 🤔

.gitmodules
tsteur: moved it up, git adds it by default at the end

tsteur:
@sgiehl just double checked again it will recognise it's not in core and then ask the Marketplace for an update.

sgiehl:
ok. Then guess it's fine if we release a version of the plugin on the marketplace before releasing Matomo 4.

sgiehl (approved):

PR #15558, opened 2020-02-12
Remove PHP serialization response format

### LINKED ISSUE #8100: Deprecate the PHP serialization response format
A possible response format for APIs is in a PHP serialized string. This isn't good for many reasons:
- readable only by PHP
- prone to not work across systems or across time (e.g. differences between HHVM and the zend engine)
- very fragile (e.g. some plugin replace words in responses, this break this format)
- not standard at all (JSON, XML & co are more than enough for an HTTP API)

We could deprecate it for 3.0 and remove it later.

sgiehl:
fixes #8100


core/CronArchive.php
```diff
@@ -908,7 +908,7 @@ protected function processArchiveDays($idSite, $lastTimestampWebsiteProcessedDay
             $this->logArchiveWebsite($idSite, "day", $date);
 
             $content = $this->request($url);
-            $daysResponse = Common::safe_unserialize($content);
+            $daysResponse = json_decode($content, true);
```
tsteur: btw does it make sense to look here for json decode errors? Like before I remember I've seen quite a few times `'Unable to unserialize a string:` errors/exceptions when eg some other content was printed like warnings/errors or something else before or after the actual output

plugins/API/API.php
```diff
@@ -524,7 +524,7 @@ public function getBulkRequest($urls)
 
         $result = array();
         foreach ($urls as $url) {
-            $params = Request::getRequestArrayFromString($url . '&format=php&serialize=0');
+            $params = Request::getRequestArrayFromString($url . '&format=json');
```
tsteur: wouldn't we use directly the original format? Then the decode might not be needed? But I suppose this wouldn't return the same result maybe as the dataTable etc wouldn't be converted to an array 👍  Haven't checked yet but be good to test there's no change in output (doubt there is but never know)

tsteur (changes requested):
Haven't tested it yet but left a comment re the archiver

plugins/API/API.php
sgiehl: The JSON & XML formatter actually used the PHP Formatter to get the DataTable as an array. I have moved the method to convert the table to an array to the abstract class. So there shouldn't be any change in output when using json format an a `json_decode`

core/CronArchive.php
sgiehl: I don't think json_decode will result in warning when the string is no valid json. At least for PHP 7. There's `json_last_error()` to get the error if there was any.

tsteur: `json_last_error ` that's what I meant 👍

sgiehl: I'm not sure if it's worth to log a warning here if the request doesn't have a valid json response.
I guess the response might only not be valid if an error or warning is printed within, but that should already be logged within the request. Or what benefit would you see there?

tsteur: It's basically forwarding the error from the archiving request to the CronArchive. I've seen quite a few issues with this in the past 6 months and it was quite useful to fail the archiving here should there be any error.

sgiehl: I see your point. But archiving will actually still fail because of the lines below:
https://github.com/matomo-org/matomo/blob/430a7f4c1f0cfceee7ede1d43c75e9e562fb718a/core/CronArchive.php#L910-L927

tsteur: Yes, you just don't know why maybe? A serialisation error be a bit more specific and can help investigate / understand things better

sgiehl: Can't see a big difference between a (specific) serialization error and a `invalid response` error. But might make sense to make the one message already triggered a bit better...

sgiehl: Done. See https://github.com/matomo-org/matomo/blob/b19ecc8d56201f4c1347a561208976c17acdf0f1/core/CronArchive.php#L923-L927

tsteur:
@sgiehl one more test still seems to be failing and then should be good to merge https://travis-ci.org/github/matomo-org/matomo/jobs/655982878#L893

sgiehl:
@tsteur test should be fixed now

Issue #15513, opened 2020-02-04
Loading Page Overlay fails and logs user out


kiwi2101:
In a Matomo 3.13.1 when I'm calling a Page Overlay I always get kicked out of the current session with message "Error: Your session has expired due to inactivity. Please log in to continue."
I can't find any reason for this behavior. Any help would be appreciated!

tsteur:
You may get logged out if you don't select "remember me" on log in and if you're not active for 30 minutes. For how long are you on the overlay page usually when this happens?

kiwi2101:
I did select "remember me" on login. I'm getting kicked out as soon as I call the overlay page.

tsteur:
I've tried to reproduce this on a few Matomo installations but always works for me. Do you know if there is any custom login plugin active, or anything else custom maybe? Eg do you log in regularly through the regular log in screen or maybe use a feature we have called `logme` or something?

tsteur:
Also... wondering... do you maybe have eg on the page you are opening the overlay for maybe `disableCookies` active or so?

kiwi2101:
Matomo is a pretty standard installation but quite a lot of pages getting tracked there (15 in total). No special login or disableCookies on the target pages. I wondered if there is any Header data which interferes but removing them didn't help.
Strange thing was, that I could call the staging-System. No difference there but the staging domain is the same where matomo is running.
So I double checked the output in Chrome and got following warning:

"A cookie associated with a cross-site resource at <URL> was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at <URL> and <URL>."

By the way I can see the iframe content with the called page but the main frame shows the session timeout error message. I also tried the option overlay_disable_framed_mode but looks like the same thing happens, the page is called without any bubbles.

tsteur:
Is your site or Matomo running on `http://`? 

Any chance you could maybe create a login for us so we could reproduce it there and do some investigation? I suppose a user with view access be enough. If so, feel free to send us details to hello at matomo dot org. 

Otherwise be good to know if your site is running http or https

pyuyu:
I also encountered the same problem

BigIron5:
Same problem here.

mattab:
@tsteur the issue can also be reproduced in our instance, it looks like this on loading Overlay:
[...]

PR #15561, opened 2020-02-13
Various fixes for Samesite handling in sessions, iframes

diosmosis:
Fix couple issues w/ samesite handling in session, make sure session start has correct value, write header where needed, and use None for opt out frame so the session ID is sent when embedding the iframe.

fix https://github.com/matomo-org/matomo/issues/15513


core/Session.php
```diff
@@ -190,7 +190,14 @@ public static function getSameSiteCookieValue()
     {
         $config = Config::getInstance();
         $general = $config->General;
-        if (!empty($general['enable_framed_pages']) && ProxyHttp::isHttps()) {
+
+        $module = Common::getRequestVar('module', false);
```
tsteur: just fyi you can use `Piwik::getModule()` and `Piwik::getAction()`. not too important though

diosmosis: modified

js/piwik.js
```diff
@@ -3290,7 +3290,8 @@ if (typeof window.Piwik !== 'object') {
                     (msToExpire ? ';expires=' + expiryDate.toGMTString() : '') +
                     ';path=' + (path || '/') +
                     (domain ? ';domain=' + domain : '') +
-                    (isSecure ? ';secure' : '');
+                    (isSecure ? ';secure' : '') +
+                    (isSecure ? ';SameSite=None' : ';SameSite=Lax');
```
sgiehl: Actually that might maybe cause problems in some specific browser versions that do not accept cookies with `SameSite=None` or treat them as `SameSite=Strict` instead. There's a list of incompatible browser versions. See https://www.chromium.org/updates/same-site/incompatible-clients

sgiehl (changes requested):
Looked through the changes and if I unterstand everything correct it looks mostly good so far. But I'm wondering if those lines need an adjustment as well:

https://github.com/matomo-org/matomo/blob/109926dd5a65244406bede12b897d59a23803d96/core/Cookie.php#L438-L458

First point here is that Chrome is split up into `Chrome`, `Chrome Mobile`, `Chrome Webview`, ... and many browsers are based on Chrome/Chromium and if they update their base to Chrome 80, they will require the correct SameSite values. So basically it's not good to only check for Browser with the name `Chrome` here. It would be better to check the browser family instead. So something like:

```php
$browserFamily = \DeviceDetector\Parser\Client\Browser::getBrowserFamily($deviceDetector->getClient('short_name'))
$isChrome = $browserFamily === 'Chrome';
```

Also Firefox and Edge announced to support the SameSite cookies. Won't they restrict `SameSite=None` without `Secure` as well? Not sure if the check for `Chrome` makes sense here at all.

diosmosis:
great points @sgiehl , I'll them asap.

tsteur:
> Also Firefox and Edge announced to support the SameSite cookies. Won't they restrict SameSite=None without Secure as well? Not sure if the check for Chrome makes sense here at all.

Just fyi they support samesite already for a while. What changes for Chrome is the default when no samesite attribute is specified, then Lax is applied, and when it says None it must be secure. AFAIK Firefox etc that's not the case just yet. We could probably apply same behaviour across browsers already though. Just means opt out might be broken for few more sites on http but might not be too many.

diosmosis:
Updated to always add SameSite=Lax in JS. Seems to work for chrome, firefox & safari. Didn't use old versions of safari of course, though those have issues w/ None only.

PR #15564, opened 2020-02-13
Fix some failing tests

sgiehl:
* Opt-Out UI test was currently failing as the https warning is displayed. That didn't show up in build artifacts as the same screenshot name was reused


plugins/TwoFactorAuth/tests/UI/TwoFactorAuth_spec.js
```diff
@@ -20,13 +20,19 @@ describe("TwoFactorAuth", function () {
...
     }
 
     async function loginUser(username, doAuth)
     {
         // make sure to log out previous session
         await page.goto(logoutUrl);
 
+        var cookies = await page.cookies();
+        cookies.forEach(cookie => {
+            page.deleteCookie(cookie);
```
sgiehl: It seems in tests logging out doesn't kill cookies, and so the some session data is still available. That caused some 2fa tests to fail, as the password verify was still stored and not triggered again

plugins/UsersManager/tests/UI/UsersManager_spec.js
```diff
@@ -210,7 +210,10 @@ describe("UsersManager", function () {
         await page.click('.userEditForm .menuPermissions');
         await page.mouse.move(0, 0);
 
-        expect(await page.screenshotSelector('.usersManager')).to.matchImage('permissions_edit');
+        expect(await page.screenshotSelector('.usersManager')).to.matchImage({
+            imageName: 'permissions_edit',
+            comparisonThreshold: 0.0005,
```
sgiehl: this and the following tests often failed randomly as some pixels of the permission boxes and drop downs were slightly different. Adding that small threshold shouldn't hurt, but they should fail less often

tests/PHPUnit/Framework/Fixture.php
```diff
@@ -77,7 +77,7 @@
 class Fixture extends \PHPUnit_Framework_Assert
 {
     const IMAGES_GENERATED_ONLY_FOR_OS = 'linux';
-    const IMAGES_GENERATED_FOR_PHP = '5.6';
+    const IMAGES_GENERATED_FOR_PHP = '7.2';
```
sgiehl: Because of that some scheduled report tests were not executed at all. Adjusted the expected files as well, so the tests pass again :-)

tests/UI/expected-screenshots/ReportExporting_VisitsSummary.get_exported.php.txt
```diff
@@ -1 +1 @@
-a:5:{i:2008;a:0:{}i:2009;a:0:{}i:2010;a:0:{}i:2011;a:0:{}i:2012;a:9:{s:9:"nb_visits";d:311;s:10:"nb_actions";d:708;s:19:"nb_visits_converted";d:245;s:12:"bounce_count";d:166;s:16:"sum_visit_length";d:96791;s:11:"max_actions";d:20;s:11:"bounce_rate";s:3:"53%";s:20:"nb_actions_per_visit";d:2.2999999999999998;s:16:"avg_time_on_site";s:9:"5 min 11s";}}
\ No newline at end of file
+a:5:{i:2008;a:0:{}i:2009;a:0:{}i:2010;a:0:{}i:2011;a:0:{}i:2012;a:9:{s:9:"nb_visits";d:311;s:10:"nb_actions";d:708;s:19:"nb_visits_converted";d:245;s:12:"bounce_count";d:166;s:16:"sum_visit_length";d:96791;s:11:"max_actions";d:20;s:11:"bounce_rate";s:3:"53%";s:20:"nb_actions_per_visit";d:2.3;s:16:"avg_time_on_site";s:9:"5 min 11s";}}
```
sgiehl: Guess that changed due to the php version change

PR #15567, opened 2020-02-13
Add UI test for JavaScript tracking.

diosmosis:
Forgot that #15374 also contained a UI test for JS tracking so re-creating it. Just changes to tests so adding to 3.13.3 (also updated puppeteer in the samesite PR)


tests/PHPUnit/Fixtures/JSTrackingUIFixture.php
```diff
@@ -0,0 +1,53 @@
...
+        $this->extraTestEnvVars = array(
+            'loadRealTranslations' => 1,
+        );
+        $this->extraPluginsToLoad = array(
+            'ExampleTracker',
+        );
+
+        parent::performSetUp($setupEnvironmentOnly);
+
+        $this->testEnvironment->overlayUrl = UiTestFixture::getLocalTestSiteUrl();
```
sgiehl: The fixture fails to set up here
```
PHP Fatal error:  Class 'Piwik\Tests\Fixtures\UiTestFixture' not found in /home/travis/build/matomo-org/matomo/tests/PHPUnit/Fixtures/JSTrackingUIFixture.php on line 46
```

tests/UI/specs/JSTracker_spec.js
```diff
@@ -0,0 +1,32 @@
...
+        // view another page
+        await page.evaluate(() => $('a:contains(Page 3)')[0].click());
+        await page.waitForNetworkIdle();
+        await page.waitFor(500);
+        await page.waitForNetworkIdle();
+
+        // visit visitor log for correct date
+        await page.goto(visitorLogUrl);
+
+        expect(await page.screenshot({ fullPage: true })).to.matchImage('visitor_log');
```
sgiehl: the expected image for that test is missing

tests/UI/specs/JSTracker_spec.js
```diff
@@ -0,0 +1,32 @@
+/*!
+ * Piwik - free/libre analytics platform
+ *
+ * Bar graph screenshot tests.
```
sgiehl: Guess that could be adjusted a little 🙈

sgiehl (changes requested):

tests/UI/specs/JSTracker_spec.js
sgiehl: Seems something doesn't work here. Last processed screenshot shows en empty visitor log:

https://builds-artifacts.matomo.org/matomo-org/matomo/new-js-tracking-test/38545/processed/JSTracker_visitor_log.png

diosmosis: Yeah, it works when it runs by itself, but not w/ other tests. Will be fun to debug.

PR #15569, opened 2020-02-13
Fix non-numeric value encountered in session recording

tsteur:
fix DEV-1879
In session recording some columns aren't numeric like the time:

image

This caused these notices to be shown in the admin

image

It's the easiest fix I can think of for now. This regressed in https://github.com/matomo-org/matomo/pull/15304 by the looks.


sgiehl (approved):
Guess makes sense to change that. Can't think of a reason why that could cause problems...

PR #15570, opened 2020-02-14
couple proxysite changes for comparison hopefully w/o side effects

### LINKED ISSUE #15527: Demo.matomo.org does not show correct data
Issue reproduced on the demo: https://demo.matomo.org/index.php?module=CoreHome&action=index&date=today&period=week&idSite=62&updated=2#?idSite=62&period=week&date=2020-02-01&category=Referrers_Referrers&subcategory=General_Overview
This issue is visible to everyone on the demo it seems (in a private browser window)
but i could not reproduce it on other instances.

Here is how it looks:

Screenshot from 2020-02-07 13-18-58

* Got evolution graphs showing "Clicks" metrics for Direct entries and Total
* Expected instead evolution graphs showing "Visits" metrics for Direct entries and Total

What could cause such an issue of plotting the wrong metrics?

I could try to [run the SQL in this FAQ](https://matomo.org/faq/troubleshooting/faq_21433/) for "anonymous" user but I expected that anonymous would never get any setting stored (as the settings are expected to be stored in each anonymous' user session)

diosmosis:
Fixes #15527

Not completely sure these changes are complete, but I don't get errors locally w/ comparison or get strange data.

### DIFF
--- a/core/Plugin/ViewDataTable.php
+++ b/core/Plugin/ViewDataTable.php
@@ -8,10 +8,12 @@
  */
 namespace Piwik\Plugin;
 
+use Piwik\API\DataTablePostProcessor;
 use Piwik\API\Request;
 use Piwik\API\Request as ApiRequest;
 use Piwik\Common;
 use Piwik\DataTable;
+use Piwik\Metrics\Formatter\Html;
 use Piwik\Period;
 use Piwik\Piwik;
 use Piwik\Plugins\API\Filter\DataComparisonFilter;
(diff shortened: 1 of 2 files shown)


tsteur:
@diosmosis feel free to merge if tests pass. We can always adjust demo later again as it's hard to test/know if it fully works

diosmosis:
ProxySite has tests but they don't work so well anymore, I can use those to get a better idea.

diosmosis:
Undid one part of the change, had to be fixed in ProxySite, will merge if tests pass.

Issue #15545, opened 2020-02-10
Deadlock found when trying to get lock


eramirezprotec:
Hi!

I'm getting this error every day since I updated Matomo to version 3.13.1:

`INFO [2020-02-09 04:07:32] 9595  Failed to invalidate archived reports: SQLSTATE[40001]: Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction`

I found some tickets about the deadlock situation but none of them speak about the deadlock happening when trying to invalidate archived reports. 

Thank you very much.

tsteur:
seeing here https://github.com/matomo-org/matomo/issues/6398#issuecomment-388370992 reports to invalidate are mentioned as well. Might be the same. How often do you see this error per day? Are you also using the log importer? Is there any chance you could enable mysql error log so we can get more information about the deadlock?

pardvm:
@tsteur,
I've able to relate this problem with this other: #15267 
In that issue, you sugested to have a sleep between the launch of each cron archiver. We did it indeed but it doens't seems to work.

Do you have another idea on how to get several archivers synchronized in order only one site is processed?

tsteur:
@pardvm have you tried to increase the sleep to say one minute (assuming you launch them hourly)? And I suppose the sleep is executed correctly with different sleep intervals for each archiver? Looking at the code this should work. Let me know if this doesn't work or if you confirm it's set up like this. I suppose we could in general also add some random minor sleep interval at the beginning of the script just to avoid these issues in general. @pardvm as a result of this do you also see deadlocks?

pardvm:
@tsteur,
We've increased the sleep to 30s and, yes: each archiver is launched 30s after the previous.

I'll keep observing and I'll come back here with the results.

pardvm:
@tsteur,
We haven't observed the deadlock since last modification regarding the sleep: cross fingers!
Thanks

tsteur:
Awesome @pardvm fingers crossed. Let me know if they come back otherwise.

@eramirezprotec might be worth a try as well in case you're starting multiple archivers at the same time? 

I will keep this issue open to at least at a tiny random sleep at the beginning of the archive script to prevent this issue a bit better.

@diosmosis @mattab thinking to add some sleep to the beginning of the archiver script in `CoreArchiver.php` command.

```php
        // we randomly sleep up to half a second in case multiple archivers are being launched at the same time.
[...]

PR #15577, opened 2020-02-16
During archive invalidation clear general cache less often

tsteur:
I know the code is not ideal since everyone who calls `forgetRememberedArchivedReportsToInvalidate` would need to know to ideally also invalidate the general tracker cache. On the other side we're seeing quite a lot of invalidations like sometimes 50 invalidations within a few ms causing heaps of cache invalidations and cache re-generations across many servers within few ms potentially resulting in locks and temporary performance issues. 

Maybe this helps to prevent also things like #15545

@diosmosis maybe you can have a look? Maybe since the code is only used internally it may be fine like that?

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -125,7 +125,6 @@ class ArchiveInvalidator
 
         $generalCache = Cache::getCacheGeneral();
         if (empty($generalCache[$cacheKey][$idSite][$dateStr])) {
-            Cache::clearCacheGeneral();
             return [];
         }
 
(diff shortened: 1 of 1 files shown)


diosmosis:
Looks good to me, some suggestions:

* We can get rid of the cache clear in `getRememberedArchivedReportsOptionFromTracker()`. It is cleared again anyway at the end of the method that calls it here: https://github.com/matomo-org/matomo/blob/4f879e1db8600f21b3a09711b43bcc63282682a1/core/Archive/ArchiveInvalidator.php#L118.
* If it's possible to just change a single entry in the general cache, then we could also get rid of that clear too. But that might not be do-able.

tsteur:
@diosmosis removed the other clear cache as well 👍

PR #15578, opened 2020-02-16
Add movers and shakers widget to dashboard

### LINKED ISSUE #13153: Add "Movers and shakers" widget to the default dashboard
I think it would be quite valuable to have. I don't know what it shows when there is no data... maybe could explain what the plugin does... because after installation there will be no data and this way users can understand what the widget is for... 

maybe we can also add a few more reports to this report? not sure.

tsteur:
fix https://github.com/matomo-org/matomo/issues/13153

### DIFF
--- a/plugins/Insights/tests/Integration/ApiTest.php
+++ b/plugins/Insights/tests/Integration/ApiTest.php
@@ -246,7 +246,7 @@ class ApiTest extends SystemTestCase
         $this->assertTrue($this->api->canGenerateInsights('2012-12-12', 'month'));
 
         $this->assertFalse($this->api->canGenerateInsights('last10', 'day'));
-        $this->assertFalse($this->api->canGenerateInsights('2012-11-11,2012-12-12', 'range'));
+        $this->true($this->api->canGenerateInsights('2012-11-11,2012-12-12', 'range'));
     }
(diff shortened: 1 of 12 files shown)


plugins/Insights/tests/Integration/ApiTest.php
```diff
@@ -246,7 +246,7 @@ public function test_canGenerateInsights()
         $this->assertTrue($this->api->canGenerateInsights('2012-12-12', 'month'));
 
         $this->assertFalse($this->api->canGenerateInsights('last10', 'day'));
-        $this->assertFalse($this->api->canGenerateInsights('2012-11-11,2012-12-12', 'range'));
+        $this->true($this->api->canGenerateInsights('2012-11-11,2012-12-12', 'range'));
```
sgiehl:
```suggestion
        $this->assertTrue($this->api->canGenerateInsights('2012-11-11,2012-12-12', 'range'));
```

sgiehl (changes requested):
One test is failing and the UI tests will need some updates

tsteur:
Tests should hopefully be fixed now @sgiehl

sgiehl (approved):

PR #15583, opened 2020-02-17
Fixes some comparison issues

### LINKED ISSUE #15193: Compare to: previous year (day) goes back two years
I was testing the new comparing feature inside the calendar and I found this.

1. select period = day (for example 5 nov 2018)
2. check :heavy_check_mark: «Compare in» then select «Previous year»

I expected a compared dashobard of 2018-11-05 day period and 2017-11-05 day period but query is ran with **2016-11-05** CompareDates[] and ComparePeriods[] = range.

Then I've expected I can compare day periods and not custom ranges. 
This means to me Matomo is always calculating a range report. Not using pre-archived periods.

Note: «Compare to: previous period» works like expected

### LINKED ISSUE #15194: Compare to: previous year always returning a range
I've expected to compare a day period with same day period in year before. Two pre-archived reports. So I checked in the calendar-dropdown:

1. 2018-11-05 (day)
2. «compare to:» and «previous year»

I expected to have two day period comparison like happend on «previous period» but worked like «custom», returning a range. 

On huge websites, _ranged report must be calculated on the fly by Matomo_, while day, week, month, year periods are pre-calculated during cron archive.

Screenshot_2019-11-25 ER-portale - 2018-09-03 - Web Analytics Reports - Matomo
Screenshot_2019-11-25 ER-portale - 2018-09-03 - Web Analytics Reports - Matomo(1)

### Then I've expected some missing features:

 - week periods comparison between years (for example 43th week of 2018 vs 43th week of 2017)
- month period comparison between years (november 2018 vs november 2017), not ranged from 2017-11-01 and 2017-11-30

The problem with ranged data is the server gets overloaded on big amounts of data. You can feel it by comparing «period year» with «previous year», then «period year» with «previous period». The second (pre archived) is faster.

sgiehl:
fixes #15194, #15193

### DIFF
--- a/plugins/CoreHome/angularjs/common/services/periods.js
+++ b/plugins/CoreHome/angularjs/common/services/periods.js
@@ -63,7 +63,7 @@
         },
 
         getDateRange: function () {
-            return [this.dateInPeriod, this.dateInPeriod];
+            return [new Date(this.dateInPeriod.getTime()), new Date(this.dateInPeriod.getTime())];
         }
     };
 
--- a/plugins/CoreHome/angularjs/period-selector/period-selector.controller.js
+++ b/plugins/CoreHome/angularjs/period-selector/period-selector.controller.js
@@ -257,9 +257,16 @@
                 currentDateRange[0].setFullYear(currentDateRange[0].getFullYear() - 1);
                 currentDateRange[1].setFullYear(currentDateRange[1].getFullYear() - 1);
 
+                if (vm.selectedPeriod === 'range') {
+                    return {
+                        comparePeriods: ['range'],
+                        compareDates: [piwikPeriods.format(currentDateRange[0]) + ',' + piwikPeriods.format(currentDateRange[1])],
+                    };
+                }
+
                 return {
-                    comparePeriods: ['range'],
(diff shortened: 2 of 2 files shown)


plugins/CoreHome/angularjs/common/services/periods.js
```diff
@@ -63,7 +63,7 @@
         },
 
         getDateRange: function () {
-            return [this.dateInPeriod, this.dateInPeriod];
+            return [new Date(this.dateInPeriod.getTime()), new Date(this.dateInPeriod.getTime())];
```
sgiehl: we need to ensure new date objects are returned here. Otherwise only references for `this.dateInPeriod` were returned, causing problems when they were manipulated. (That caused day periods compared two years back)

diosmosis:
LGTM

tassoman:
Hello there!
 Is there any regression about this bug fixes? I've installed `3.13.5` in staging environment and I still feeling the wrong behaviour.
I thought were fixed some months ago. :thinking:

tassoman:
The «interval» bug is confirmed also for _month_ period

tsteur:
@tassoman this was only fixed in 4.x AFAIK. @tassoman feel free to create a PR for 3.X and I can merge it should you need the patch earlier

PR #15584, opened 2020-02-17
Fix CSV export for idSite=all

### LINKED ISSUE #14362: csv export for VisitsSummary.getVisits with idSite=all and non-superuser not working
Reported twice in https://forum.matomo.org/t/matomo-api-to-fetch-results-for-idsite-all-in-csv-xls-or-tsv-format-not-working-for-normal-users/31663 and reproduced on my Matomo instance

Requesting the `VisitsSummary.getVisits` API when
- the user of the token_auth is not a superuser
- idSite=all is requested
- and format=csv (or xls)

then the following error appears:
> Error: The parameter 'idSite=' is missing from the request.

All other requests seem to work fine (e.g. format=json)

sgiehl:
problem was actually not only the CSV exporter, but also some plugins did failing defining their report metadata when `idSite=all`

fixes #14362


plugins/ImageGraph/ImageGraph.php
```diff
@@ -56,7 +56,7 @@ public function getReportMetadata(&$reports, $info)
         $idSite = $info['idSite'];
 
         // If only one website is selected, we add the Graph URL
-        if (empty($idSite)) {
+        if (empty($idSite) || 'all' === $idSite) {
```
diosmosis: Not super important but we could probably make these checks stronger w/ `!is_numeric($idSite)`. Think it would be a good idea?

sgiehl: changed that

diosmosis:
One other thing, can we add tests for this?

sgiehl:
@diosmosis add a new test

Issue #8567, opened 2015-08-13
Remove code tagged with `@deprecated` 


mattab:

tsteur:
I think so late in the release cycle we shouldn't remove any more methods and rather postpone it to next time. Looking through it there are > 50 deprecated comments and it would require quite a bit of work to make plugins compatible

PR #15586, opened 2020-02-17
Removes deprecated class Piwik\Translate

sgiehl:
refs #8567


tsteur:
BTW: I'm wondering in that case it it's worth removing the class? Not sure there is much of a down side to keep it and at the same time it might force heaps of plugins to change the code maybe? Might be the same for some other classes /methods. Wondering if sometimes we could just keep the old classes for simplicity of BC. In this case it wouldn't hurt us much maybe to keep that old class? 

Of course we'd still ideally change some usages to be a good example and make use of it correctly ideally. 

In general since we manage > 50 plugins it might be easier to sometimes just keep the old classes if they are as simple as the translate class

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)*
 * The deprecated Platform API class `\Piwik\Registry` has been removed. Use `\Piwik\Container\StaticContainer` instead
 * The deprecated Platform API class `\Piwik\TaskScheduler` has been removed. Use `\Piwik\Scheduler\Scheduler` instead
 * The deprecated Platform API class `\Piwik\DeviceDetectorFactory` has been removed. Use `\Piwik\DeviceDetector\DeviceDetectorFactory` instead
+* The deprecated Platform API class `\Piwik\Translate` has been remoced. Use `\Piwik\Translation\Translator` instead.
```
tsteur: btw typo in remoced

sgiehl:
We already deprecated that class in Piwik 2.11 and kept it in 3.x for BC. At some point we need to break BC and remove such stuff otherwise we will mess up the code with more and more such stuff :man_shrugging:

CHANGELOG.md
sgiehl:
```suggestion
* The deprecated Platform API class `\Piwik\Translate` has been removed. Use `\Piwik\Translation\Translator` instead.
```

tsteur:
>  At some point we need to break BC and remove such stuff otherwise we will mess up the code with more and more such stuff 🤷‍♂

Not sure how one file would mess up the code there? Of course it always depends but in this case not seeing it so much. There doesn't seem to be much of a harm as it's only forwarding methods anyway. If it helps keeping plugins easily compatible and less pain for us and developers be great. Personally, I love eg how Android and iOS etc keep BC basically almost forever (not saying we should always do this but when it's easy like there might be worth it)

sgiehl:
OK. What about keeping that class but triggering a deprecation notice
whenever a method is called and remove it in Matomo 5 then?

>

tsteur:
Not sure. I reckon in that case there might not even be a need to remove it in Matomo 5 either. I'm thinking if we can keep BC whenever it's easy for us we probably should. Could even remove the deprecated flag there. In development mode we could trigger a notice though maybe and recommend using a different way. 👍

sgiehl:
@tsteur readded the file, triggering deprecation notices in development mode

tsteur:
👍

core/Translate.php
```diff
@@ -123,7 +133,15 @@ private static function getTranslator()
...
     {
+        self::triggerDeprecationNotice();
         self::loadCoreTranslation();
         Manager::getInstance()->loadPluginTranslations();
     }
+
+    protected static function triggerDeprecationNotice()
+    {
+        if (Development::isEnabled()) {
+            trigger_error('Using \\Piwik\\Translate is deprecated. Use \\Piwik\\Translation\\Translator instead.', E_USER_DEPRECATED);
```
tsteur: btw any reason we're not using Log::warn or so?

sgiehl: No. Guess a warning should be fine as well, as it's only in development mode

sgiehl:
@tsteur let me know if you think it's ready to merge. Would do that in all linked repos then and update the submodules before merge.

tsteur:
@sgiehl I haven't really looked (only very quickly) but reckon it's  fine 👍 Unfortunately won't be able to do any reviews in the next 7 days :(

PR #15590, opened 2020-02-18
Fallback to JSON renderer if no valid format is given

tsteur:
Prevents exceptions like these which we are getting quite often due to security researchers trying different inputs

> Call to a member function getResponseException() on null","file":"\/core\/API\/Request.php","line":279,"request_id":"3e2b6","backtrace":" on \/var\/www\/html\/core\/API\/Request.php(279)\n#0 \/plugins\/API\/Controller.php(41)

Ideally will add a test but probably won't find the time soon.

### DIFF
--- a/core/API/Request.php
+++ b/core/API/Request.php
@@ -276,6 +276,10 @@ class Request
                 'ignoreInScreenWriter' => true,
             ]);
 
+            if (empty($response)) {
+               $response = new ResponseBuilder('console', $this->request);
+            }
+            
             $toReturn = $response->getResponseException($e);
         } finally {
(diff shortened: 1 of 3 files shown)


mattab:
Example request that triggered error: `POST: {"date":"last10","format":"-1 OR 2+931-931-1=0+0+0+1 --","idSite":"1","limitActionsPerStep":"5","method":"UsersFlow.getUsersFlow","module":"API","period":"day","token_auth":"XYZANONYMIZED","translateColumnNames":"1","format_metrics":"bc"}`

tsteur:
Thanks @diosmosis

PR #15591, opened 2020-02-19
Do not initiate auth instance if a user is already logged in in FrontController::init()

### LINKED ISSUE #15550: Notice: Trying to access array offset on value of type null in core/Session/SessionAuth.php on line 63
Not sure if this is only related to WordPress... getting below output when running a CLI command. Started happening with 3.13.2. Maybe caused by https://github.com/matomo-org/matomo/pull/15472 ? Not sure... 

It's just a notice though. Things generally still seems to work.

> Notice: Trying to access array offset on value of type null in core/Session/SessionAuth.php on line 63

> Call Stack:
>     0.0010     397688   1. {main}() console:0
>     0.9315   47844424   2. Piwik\Console->run() console:32
>     0.9343   48033128   3. Piwik\Console->doRun() vendor/symfony/console/Symfony/Component/Console/Application.php:126
>     0.9343   48033128   4. Piwik\Console->doRunImpl() core/Console.php:93
>     1.0943   57948336   5. Piwik\Access::doAsSuperUser() core/Console.php:141
>     1.0954   57967824   6. Piwik\Console->Piwik\{closure:core/Console.php:139-141}() core/Access.php:644
>     1.0954   57968200   7. call_user_func:{core/Console.php:140}() core/Console.php:140
>     1.0954   57968200   8. Piwik\Console->doRun() core/Console.php:140
>     1.0955   57968472   9. Piwik\Console->doRunCommand() vendor/symfony/console/Symfony/Component/Console/> Application.php:195
>     1.0956   57968472  10. Piwik\Plugins\WordPress\Commands\GenerateCoreAssets->run() vendor/symfony/console/Symfony/> Component/Console/Application.php:874
>     1.0957   57970664  11. Piwik\Plugins\WordPress\Commands\GenerateCoreAssets->execute() vendor/symfony/console/> Symfony/Component/Console/Command/Command.php:257
>     1.2942   58002656  12. Piwik\FrontController->init() /Users/thomassteur/Development/wpsingle/wp-content/plugins/> matomo/plugins/WordPress/Commands/GenerateCoreAssets.php:57
>     1.3136   58568120  13. Piwik\FrontController->makeAuthenticator() core/FrontController.php:406
>     1.3139   58568120  14. Piwik\Session\SessionAuth->getLogin() core/FrontController.php:701

diosmosis:
@tsteur think this will solve the problem for the wp-matomo command?

Fixes #15550

### DIFF
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -392,18 +392,21 @@ class FrontController extends Singleton
 
         $loggedIn = false;
 
-        // try authenticating w/ session first...
-        $sessionAuth = $this->makeSessionAuthenticator();
-        if ($sessionAuth) {
-            $loggedIn = Access::getInstance()->reloadAccess($sessionAuth);
-        }
+        // user may already be logged in when init() is called, eg, in a CLI command
+        if (!Access::getInstance()->isUserLoggedIn()) {
+            // try authenticating w/ session first...
+            $sessionAuth = $this->makeSessionAuthenticator();
+            if ($sessionAuth) {
+                $loggedIn = Access::getInstance()->reloadAccess($sessionAuth);
+            }
 
-        // ... if session auth fails try normal auth (which will login the anonymous user)
-        if (!$loggedIn) {
-            $authAdapter = $this->makeAuthenticator();
-            Access::getInstance()->reloadAccess($authAdapter);
-        } else {
-            $this->makeAuthenticator($sessionAuth); // Piwik\Auth must be set to the correct Login plugin
+            // ... if session auth fails try normal auth (which will login the anonymous user)
+            if (!$loggedIn) {
+                $authAdapter = $this->makeAuthenticator();
+                Access::getInstance()->reloadAccess($authAdapter);
+            } else {
+                $this->makeAuthenticator($sessionAuth); // Piwik\Auth must be set to the correct Login plugin
+            }
         }
 
         // Force the auth to use the token_auth if specified, so that embed dashboard
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -392,21 +392,19 @@ class FrontController extends Singleton
 
         $loggedIn = false;
 
-        // user may already be logged in when init() is called, eg, in a CLI command
-        if (!Access::getInstance()->isUserLoggedIn()) {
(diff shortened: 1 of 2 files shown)


core/FrontController.php
```diff
@@ -392,18 +392,21 @@ public function init()
...
 
-        // try authenticating w/ session first...
-        $sessionAuth = $this->makeSessionAuthenticator();
-        if ($sessionAuth) {
-            $loggedIn = Access::getInstance()->reloadAccess($sessionAuth);
-        }
+        // user may already be logged in when init() is called, eg, in a CLI command
+        if (!Access::getInstance()->isUserLoggedIn()) {
+            // try authenticating w/ session first...
+            $sessionAuth = $this->makeSessionAuthenticator();
```
tsteur: @diosmosis I'm not sure re possible side effects or so. 

I wonder... if technically someone could do something like this 

```
$_GET['token_auth'] = ...
FrontController::getInstance()->dispatch()
```

if that could be any issue. Or if the user may be temporarily logged in using `doAsSuperUser` but then the `Access::$login` is not reset or so. 

It might be a great fix I'm just unable to tell re side effects etc it's so tricky! In generally it does sound good though cause there should be a need to authenticate again. Unless maybe for some random use case, not sure. I reckon we could do this only on CLI and maybe only skip the session auth? Or always skip session auth on CLI since there is no session? Might be more "secure" in terms of possible side effects?

tsteur:
btw moving this into 3.13.4 for now. I reckon it's actually not needed in 3.13.3

core/FrontController.php
diosmosis: I guess it makes sense to just skip sessionauth if in cli mode. Can't imagine how it would work then...

diosmosis: @tsteur sound good to change it to:

```
if (!Access::getInstance()->isPhpCliMode()) {
    ...
```

?

tsteur: 👍  Although I reckon you mean `Common::isPhpClimode`?

diosmosis: Yes :) don't know how that got past me :)

diosmosis:
@tsteur ready for another review

Issue #15589, opened 2020-02-18
[Geo] Unknown region metrics increased significantly after update to v3.13.1


pardvm:
After changes in MaxMind licensing model and before upgrading to Matomo v3.13.1, I decided to become a register user of MaxMind in order to continue using their geo database. 

Then, after I finished to upgrade my Matomo instance to v3.13.1, taking into account [the recomendations from Matomo](https://matomo.org/faq/how-to/faq_163/), I decided to give a chance to the new provider db-ip.com.
The experience was very gratefull because the "Unkown city" metrics fell down drastically as shown in the following evolution graph:
(_the vertical line indicates the day of the upgrade to v3.13.1_)
![image](https://user-images.githubusercontent.com/25912392/74739010-4e270080-5258-11ea-950f-6a9dd3ec7839.png)

However, the drawback is that "Unknown region" has raised, also dramatically:
![image](https://user-images.githubusercontent.com/25912392/74739247-bd9cf000-5258-11ea-81aa-4b5b8161f1a6.png)

I was warned about that Geo-IP free databases aren't 100% accurated. However, given the huge difference between the two providers, I'm wondering if it could be a kind of v3.13 bug.
(I reviewed my installation several times but I'm not able to see any clue about the origin of this effect.)

Any help will be wellcome!

tsteur:
I can confirm re city seeing same behaviour on commercial DB-IP. There is basically now known unknown.

We're also seeing a bit less Unknown regions though but we only made the change a couple weeks ago so maybe too early to tell.

On the free DB I can confirm same behaviour. Regions are basically all unknown. Which makes me think this is not an issue in Matomo but related to the free database not including regions possibly.

fyi @mattab

pardvm:
@tsteur,
Let me check if I understand what you say:
Can you confirm that, using commercial DB-IP, the behaviour on regions is the same that the one I described using free DB-IP?

I would need to double-check this because I'm willing to pay an anual subscription to DB-IP but only if the regions are woriking fine.

Thanks for you support.

tsteur:
So far I'm seeing quite a lot of regions using the commercial-DB, but on another Matomo using the free DB I get basically no region.

Ideally, if possible, ping me again in a week just so I can double confirm. This is the graph for the Unknown regions and we configured it like 1-2 weeks ago

![image](https://user-images.githubusercontent.com/273120/74775725-318fd480-52fb-11ea-9b90-5068019c04bc.png)

[...]

PR #15592, opened 2020-02-19
Try to determine the region iso code based on its name if not available

sgiehl:
refs #15589


tsteur:
Moving this into 3.13.4 for now. Be good to add a test @sgiehl 

Could also move it into Matomo 4 if needed but 3.13.4 should be fine.

sgiehl:
Will try to add one

sgiehl:
done

PR #15601, opened 2020-02-20
Fix VisitorFrequency.get API for idsites=all

sgiehl:
The ArchiveCronTest was failing for a while now already. It actually seemed to be a regression from #15128 

not sure if it's worth to cherry-pick it to 3.x as well (ping @mattab @tsteur )


tsteur:
@sgiehl it be not worth it to cherry pick but if it makes the tests work on 3.x feel free to cherry pick

plugins/VisitFrequency/API.php
```diff
@@ -49,7 +49,10 @@ public function get($idSite, $period, $date, $segment = false, $columns = false)
         $columns = Piwik::getArrayFromApiParameter($columns);
 
         /** @var \Piwik\DataTable\DataTableInterface $resultSet */
-        if (Period::isMultiplePeriod($date, $period)) {
+        if ($idSite === 'all') {
+            $resultSet = new DataTable\Map();
+            $resultSet->setKeyName('idSite');
+        } else if (Period::isMultiplePeriod($date, $period)) {
```
diosmosis: Will this still work when there's multiple sites and multiple periods?

diosmosis: Actually from the tests it looks like it does?

PR #15603, opened 2020-02-20
Try to prevent possible deadlock in archive invalidator

tsteur:
See http://mitchdickinson.com/mysql-innodb-row-locking-in-delete/ 

>  A delete actually locks all rows that are scanned in the query.

The idea is that if we delete by primary key, then no other rows will be locked thus avoiding the deadlock hopefully. For deadlock info see https://github.com/matomo-org/matomo/issues/15545#issuecomment-588916822 and below.

Checked other Option::deleteLike but they seem fine. 

fix https://github.com/matomo-org/matomo/issues/15545 

Be great if that could make it into 3.13.3 since we would likely be facing these issues a lot too.

I haven't tested if it actually prevents the deadlock but hoping so. We will otherwise need to reopen #15545 if it doesn't. 

Background: When we set an option, we set basically `$id_$processId`. So this should prevent the lock when we delete the key directly by primary key since the appendix `_$processId` should be different and not prevent the insertion of another key.

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -16,6 +16,7 @@ use Piwik\DataAccess\ArchiveTableCreator;
 use Piwik\DataAccess\Model;
 use Piwik\Date;
 use Piwik\CliMulti\Process;
+use Piwik\Db;
 use Piwik\Option;
 use Piwik\Common;
 use Piwik\Piwik;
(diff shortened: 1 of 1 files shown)


core/Archive/ArchiveInvalidator.php
```diff
@@ -192,7 +192,18 @@ public function forgetRememberedArchivedReportsToInvalidate($idSite, Date $date)
...
+        Cache::clearCacheGeneral();
+    }
+
+    private function deleteOptionLike($id)
+    {
+        // we're not using deleteLike since it maybe could cause deadlocks see https://github.com/matomo-org/matomo/issues/15545
+        // we want to reduce number of rows scanned and only delete specific primary key
+        $keys = Option::getLike($id . '%');
+        foreach ( $keys as $key => $val) {
+            Option::delete($key);
```
diosmosis: Would it make sense to do a `DELETE FROM option WHERE option_name IN (...)` so it's just one query?

tsteur: @diosmosis was thinking about it but then figured it's easier to just reuse existing code for now. Not sure it makes a big difference?

diosmosis: I guess it depends on how many invalidated dates there can usually be at one time for a single site.

tsteur: Shouldn't be too many since we regenerate the tracker cache once the flag is written to avoid another request to set this flag? could be a few if there were many requests in parallel maybe. I'll try to have a look but don't really have time.

tsteur: changed it @diosmosis

PR #15606, opened 2020-02-21
Allow using paid db-ip urls for auto update

### LINKED ISSUE #15580: Geolocation: automatically download latest paid DB-IP databases
Goal of this issue is to automatically download the latest databases using a user-input `Account key`. 

As a DB-IP database customer, one can access the URL to download the databases in  https://db-ip.com/account/db/?tab=updates

Because the mmdb download URL changes regularly,  there is a API to return the valid URL at:
`https://db-ip.com/account/$ACCOUNT_KEY/db/ip-to-location` or at: `https://db-ip.com/account/$ACCOUNT_KEY/db/ip-to-location/mmdb/url`. 

In the meantime the downloads may fail for commercial DB-IPs in some cases.

sgiehl:
This allows adding urls like `https://db-ip.com/account/$ACCOUNT_KEY/db/ip-to-location` or `https://db-ip.com/account/$ACCOUNT_KEY/db/ip-to-location/mmdb/url` directly to the auto updater

fixes #15580

### DIFF
--- a/plugins/GeoIp2/GeoIP2AutoUpdater.php
+++ b/plugins/GeoIp2/GeoIP2AutoUpdater.php
@@ -752,7 +752,7 @@ class GeoIP2AutoUpdater extends Task
 
     private function fetchPaidDbIpUrl($url)
     {
-        $content = Http::fetchRemoteFile($url);
+        $content = trim(Http::fetchRemoteFile($url));
 
         if (0 === strpos($content, 'http')) {
             return $content;
--- a/plugins/GeoIp2/GeoIP2AutoUpdater.php
+++ b/plugins/GeoIp2/GeoIP2AutoUpdater.php
@@ -131,9 +131,9 @@ class GeoIP2AutoUpdater extends Task
 
         $url = trim($url);
 
-        if ($this->isPaidDbIpUrl($url)) {
+        if (self::isPaidDbIpUrl($url)) {
             $url = $this->fetchPaidDbIpUrl($url);
-        } else if ($this->isDbIpUrl($url)) {
+        } else if (self::isDbIpUrl($url)) {
             $url = $this->getDbIpUrlWithLatestDate($url);
         }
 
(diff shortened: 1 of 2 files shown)


plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -741,4 +744,28 @@ public static function isDbIpUrl($url)
...
     }
+
+    private function isPaidDbIpUrl($url)
+    {
+        return !! preg_match('/db-ip\.com\/account\/[0-9a-z]+\/db/', $url);
+    }
+
+    private function fetchPaidDbIpUrl($url)
+    {
+        $content = Http::fetchRemoteFile($url);
```
tsteur: might be good to trim `$content` just to be safe

plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -741,4 +744,28 @@ public static function isDbIpUrl($url)
     {
         return !! preg_match('/db-ip\.com/', $url);
     }
+
+    private function isPaidDbIpUrl($url)
```
tsteur: is it possible to add a test for this?

plugins/GeoIp2/GeoIP2AutoUpdater.php
sgiehl:
```suggestion
        $content = trim(Http::fetchRemoteFile($url));
```

plugins/GeoIp2/GeoIP2AutoUpdater.php
sgiehl: Added some tests

plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -741,4 +744,35 @@ public static function isDbIpUrl($url)
...
+        if (!empty($content['url'])) {
+            return $content['url'];
+        }
+
+        throw new Exception('Unable to determine download url');
+    }
+
+    protected function fetchUrl($url)
+    {
+        return Http::fetchRemoteFile($url);
```
tsteur: @sgiehl just realising we should be checking whether internet is enabled and if not, throw an exception maybe? 

BTW out of curiosity: Any reason using `fetchRemoteFile` over `sendHttpRequest`?

sgiehl: Checking if internet is enabled is done in `__construct`, so need to do that another time.

> BTW out of curiosity: Any reason using fetchRemoteFile over sendHttpRequest?

No. Actually not. Feel free to change that...

PR #15611, opened 2020-02-23
New INI setting to set the absolute path to a PHP binary file in case Matomo cannot detect your PHP binary

tsteur:
see https://github.com/matomo-org/matomo/pull/15285#issuecomment-589980434

Where it matched this path `/usr/local/lsws/lsphp72/bin/php`.

Alternatively, user suggests to exec `/path/to/php -v` to detect it but not sure this always work or has other issues.


diosmosis:
Could we also add a INI or DI config option so users can just set it themselves if they have to?

tsteur:
Yes, I was actually thinking about this a few times and was not sure why we don't have it yet. Maybe because it could become outdated. On the other side things still work even if not configured, will simply not use CLI archiving then. Would we still need to make sure they don't configure eg php-cgi I guess?

diosmosis:
> Maybe because it could become outdated.

This sounds like a general problem, lots of things in a system could change resulting in a broken matomo. Perhaps it might help to regularly run the system check (like once a day or something) and display any errors as error notifications. In another PR of course.

For this I was just thinking of a workaround so next time we don't necessarily have to make a code change to fix it for a user. Could just be in DI?

tsteur:
@diosmosis added a config entry to the global ini and not DI as it's otherwise hard to describe how to change it and hard to expose documentation for it etc. Wasn't sure what to describe in the doc when it is useful since configuring it doesn't ensure that CLI archiving will actually work if other criteria are still not fulfilled. Should be enough though maybe. Will likely need to fix ui tests for the config later.

PR #15615, opened 2020-02-24
Make sure "Event Category" is translated in Visits log API outputs

mattab:
eg. reproduced on any user that has visited events, eg. `index.php?date=2020-02-24&module=API&action=index&visitorId=7a6948677b781fae&idSite=1&period=day&method=Live.getVisitorProfile&format=XML&expanded=1`

 outputs this:
```
<title>Event</title>
<subtitle>Event_Category: "Add review', Action: "click"</subtitle>
```
but we don't expect "Event_Category" in there. It is fixed here.

PS: some tests will need to be fixed


sgiehl (approved):

sgiehl:
will update the test files and merge afterwards...

PR #15616, opened 2020-02-24
Move Archive.php archive invalidation to Loader…

diosmosis:
… so we only invalidate when about to launch archiving.

Not sure if there are tests I can add for this just yet.


sgiehl:
Seems a test is failing now. Otherwise guess looks good to merge

tsteur:
Here's some thoughts maybe that would likely improve our problem but not really 100% sure if it makes sense and would work @diosmosis 

```diff
diff --git a/core/ArchiveProcessor/Loader.php b/core/ArchiveProcessor/Loader.php
index c99dd3d02e..d6f896fee6 100644
--- a/core/ArchiveProcessor/Loader.php
+++ b/core/ArchiveProcessor/Loader.php
@@ -79,15 +79,30 @@ class Loader
     {
         $this->params->setRequestedPlugin($pluginName);
 
+        // no valid archive exists, make sure to invalidate existing archives
+        if (($isBrowserArchiveEnabled || $isArchiveTriggered) && $this->invalidateBeforeArchiving) {
+            $this->invalidatedReportsIfNeeded(); // SHOULD INVALIDATE ALL REPORTS
+        }
+
         list($idArchive, $visits, $visitsConverted) = $this->loadExistingArchiveIdFromDb();
         if (!empty($idArchive)) {
+            // IF REPORT INCLUDES TODAY....
+                // IF ARCHIVE WAS GENERATED WITHIN LAST time_before_today_archive_considered_outdated SECONDS DO NOT RUN
+                // INVALIDATE LOGIC SINCE REPORT WOULD NOT BE REGENERATED ANYWAY
+
+                // IF ARCHIVE WAS GENERATED MORE THAN time_before_today_archive_considered_outdated SECONDS AGO,
+                // RUN INVALIDATION LOGIC AND CALL loadExistingArchiveIdFromDb() again in case it is no longer valid
+
+            // IF REPORT NOT INCLUDES TODAY
+                // IDEALLY WE INVALIDATE REPORTS ONLY THAT AFFECT THAT PERIOD TO NOT SLOW DOWN EG RANGE REQUESTS
+
             return $idArchive;
         }
 
-        // no valid archive exists, make sure to invalidate existing archives
-        if ($this->invalidateBeforeArchiving) {
-            $this->invalidatedReportsIfNeeded();
-        }
+        // NO NEED TO INVALIDATE IF THE ARCHIVE DOES NOT EXIST SINCE IT WOULDN'T MAKE A DIFFERENCE TO THE RESULT. MEANING
+        // IF WE DID INVALIDATE REPORTS, THEN THE OUTPUT BE STILL THE SAME. EG THIS WAY WE AVOID INVALIDATING REPORTS WHEN
+        // RANGE IS SELECTED AND THE REPORT DOES NOT EXIST YET. AS A RESULT OF THIS WE WOULD ALSO NOT INVALIDATE AND
+        // REARCHIVE ANY SUBPERIODS
 
         /** @var ArchivingStatus $archivingStatus */
         $archivingStatus = StaticContainer::get(ArchivingStatus::class);
```

Generally speaking eg on Cloud:
* Browser archiving and segment archiving is disabled so only range periods are causing us trouble and need to ideally avoid invalidating reports for range dates when the archive doesn't exist or when it was created recently
* The problem with invalidation of reports is that it's fairly slow, plus causes the tracker cache general to be emptied which we want to avoid whenever possible.

While many other users have browser archiving disabled, segment archiving is still enabled etc. For these users it be great to invalidate all reports only when `isArchivePhpTriggered()`. In all other cases to keep UI fast and not invalidate tracker caches too often we should only invalidate reports when needed as well see comments. 

When browser archiving is enabled, it'll be a bit of a problem since we often have to invalidate reports in order to get accurate results unless some report was generated recently (say within 900s if default config is used). This might be still fine though. It'll just make Matomo even slower but maybe not possible to avoid it until we are in Matomo 4.

diosmosis:
@tsteur I tried making changes to handle every possible case for when archiving, can you take a look? I haven't modified tests yet, you can probably just look at the latest commit.

core/DataAccess/ArchiveSelector.php
```diff
@@ -61,80 +61,41 @@ public static function getArchiveIdAndVisits(ArchiveProcessor\Parameters $params
...
-                }
-                if (empty($visitsConverted)
-                    && $result['name'] == self::NB_VISITS_CONVERTED_RECORD_LOOKED_UP
-                ) {
-                    $visitsConverted = $value;
-                }
-            }
+        // the archive is too old
+        if ($minDatetimeArchiveProcessedUTC
+            && Date::factory($result['ts_archived'])->isEarlier(Date::factory($minDatetimeArchiveProcessedUTC))
```
tsteur: fyi getting some error here that `ts_archived` is null... and then throws an exception in Date::factory() that the date is not valid. Will try to have a quick look
image

core/DataAccess/ArchiveSelector.php
```diff
@@ -61,80 +61,41 @@ public static function getArchiveIdAndVisits(ArchiveProcessor\Parameters $params
...
-
-        list($visits, $visitsConverted) = self::getVisitsMetricsFromResults($idArchive, $idArchiveVisitsSummary, $results);
-
-        if (false === $visits && false === $idArchive) {
-            return false;
+        $result = self::findArchiveDataWithLatestTsArchived($results, $requestedPlugin, $segment);
+        if (!isset($result['idarchive'])
+            || !isset($result['nb_visits'])
+        ) {
+            return [false, false, false, true];
```
tsteur: btw here it returned for me `isAnyArchiveExists = true` even though there was no archive with a done flag. Only an archive with a `nb_visits`. This can happen if eg the archiving fails before the done flag was written but after nb_visits was written. Not sure it's important.

core/DataAccess/ArchiveSelector.php
tsteur: pushed a fix for it

tsteur (changes requested):

diosmosis:
@tsteur noticed an issue w/ this change, if we invalidate only if there is something to invalidate, we may end up archiving more than once in some situations. For example:

- archive for date 2012-02-02 and site 1 does not exist
- tracking request comes in for that date & site, and pair is added to option list
- archive for date 2012-02-02 and site 1 is requested
- there is no existing archive, so nothing is invalidated
- archiving is initiated for the day 2012-02-02, site 1
- another archive request comes in for 2012-02-02, site 1
- there is now an existing archive, and the date/site is in the list of pairs to archive, so we invalidate it
- archiving is run again for 2012-02-02, site 1

We can't remove the date/site in the list in this case since it could also invalidate other archives that are not currently being requested.

The only way to do this change would be to run the invalidate query as a SELECT first then invalidate if it returns something, but that would of course be pointless.

tsteur:
> Which would only be an issue if the options in the list are removed. We can have a daily task in that case to just run the invalidation I guess.

Daily task for this should work. 👍

tsteur:
Or maybe at the beginning of the archiving we still need to invalidate all reports like we used to for cron archiving. For browser archiving handling would need a task indeed.

diosmosis:
> Or maybe at the beginning of the archiving we still need to invalidate all reports like we used to for cron archiving.

We still do this, I haven't changed that (I disabled the Archive.php code for invalidation if browser archiving is enabled, so it's all done by core:archive, which I think is better IMO).

diosmosis:
@tsteur tests should pass now, ready for a review. I may add some more tests in the meantime.

core/CronArchive.php
```diff
@@ -981,6 +981,16 @@ private function isThereAValidArchiveForPeriod($idSite, $period, $date, $segment
             $periodsToCheck = [Factory::build($period, $date, Site::getTimezoneFor($idSite))];
         }
 
+        $isTodayIncluded = $this->isTodayIncludedInPeriod($idSite, $periodsToCheck);
+        $isLast = preg_match('/^last([0-9]+)/', $date, $matches);
+
+        // don't do this check for a single period that includes today
+        if ($isTodayIncluded
```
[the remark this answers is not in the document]
diosmosis: The idea here was if it was a single period like, `period = week, day = today`, then we don't check for an existing usable archive. If it is for a date in the past, then we check if there is a usable archive and if so we don't need to archive.

For `lastN` values, however, it is more complicated. Here, we want to check for every date except today. So we check for valid archives. If the date is `last5` and for the 5th, 4th and 3rd day we have valid archives, then we change it to last2. The only difference is if we have valid archives for every day, we still do last1 (I think the min is last1 now but we can change it back to last2).

Maybe I can add some tests for this code too.

plugins/CoreAdminHome/Tasks.php
```diff
@@ -81,6 +88,17 @@ public function schedule()
         $this->scheduleTrackingCodeReminderChecks();
     }
 
+    public function invalidateOutdatedArchives()
+    {
+        if (false && !Rules::isBrowserTriggerEnabled()) {
```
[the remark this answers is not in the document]
diosmosis: removed

core/ArchiveProcessor/Loader.php
```diff
@@ -242,4 +271,33 @@ private function getIdSitesToArchiveWhenNoVisits()
 
         return $cache->fetch($cacheKey);
     }
+
+    private function invalidatedReportsIfNeeded()
```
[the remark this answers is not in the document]
diosmosis: added tests

core/CronArchive.php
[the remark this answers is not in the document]
diosmosis: added tests, maybe you can better see what it's trying to do

plugins/CoreAdminHome/Tasks.php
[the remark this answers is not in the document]
tsteur: it is still there @diosmosis ?

core/CronArchive.php
```diff
@@ -1003,36 +1014,59 @@ private function isThereAValidArchiveForPeriod($idSite, $period, $date, $segment
...
-            /** @var Range $newRangePeriod */
-            $newRangePeriod = PeriodFactory::build($period, $oldestDateWithoutArchive . ',' . $todayInTimezone);
+                /** @var Range $newRangePeriod */
+                $newRangePeriod = PeriodFactory::build($period, $oldestDateWithoutArchive . ',' . $todayInTimezone);
 
-            $newDate = 'last' . min($lastNValue, $newRangePeriod->getNumberOfSubperiods());
+                $newDate = 'last' . min($lastNValue, $newRangePeriod->getNumberOfSubperiods());
+            } else if ($isTodayIncluded) {
+                $isThereArchiveForAllPeriods = false;
+                $newDate = 'last1';
```
tsteur: btw not sure but might still need to use last2 here to workaround some archiving problems where it would maybe create a range date for today and some plugins may have trouble with it. I suppose maybe shouldn't hurt to use `last2`?

core/CronArchive.php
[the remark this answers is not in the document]
tsteur: fully getting it now 👍

core/CronArchive.php
```diff
@@ -1003,36 +1014,59 @@ private function isThereAValidArchiveForPeriod($idSite, $period, $date, $segment
...
+            }
```
tsteur: btw... I was thinking... if it is `last1` (aka `today`, `this week`, `this month`), then we could technically get the date when this archive was last generated, and if it was generated less than `time_before_(today|$period)_archive_considered_outdated` seconds ago, return `$isThereArchiveForAllPeriods  = true` and skip the archive this time. 

Only if it's easy to do though. Otherwise could also just launch the archiving. On cloud where we consider archives outdated sometimes only evert 6, 12, 18 hours (depending on config), then it could save us launching quite a few archive launches.

Not sure it's clear what I mean? It be probably similar to what we have in archive launcher.

tsteur:
👍 left another comment @diosmosis few tests are still failing but I suppose you still work in it. eg https://travis-ci.org/matomo-org/matomo/jobs/659112026?utm_medium=notification&utm_source=github_status

core/CronArchive.php
diosmosis: I thought this too, but I wasn't sure... I'll change it just in case.

core/CronArchive.php
diosmosis: I think this is already done, regardless of whether browser archiving is used or not. In Loader.php we always use the result of `getMinTimeArchiveProcessed()`. So it should always look for an archive that is out of date, correct? I could maybe add a debug log for the future...

plugins/CoreAdminHome/Tasks.php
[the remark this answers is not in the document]
diosmosis: huh, must've missed it? not sure how

diosmosis:
@tsteur fixed some tests, there are some unrelated failures though

plugins/Monolog/Processor/ExceptionToTextProcessor.php
```diff
@@ -85,6 +85,10 @@ private function getStackTrace($exception)
 
     public static function getWholeBacktrace(\Exception $exception, $shouldPrintBacktrace = true)
     {
+        if (!$shouldPrintBacktrace) {
```
tsteur: @diosmosis should this one still be needed?

diosmosis: Yes, otherwise the backtrace is always printed. This was an unrelated bug I found.

PR #15618, opened 2020-02-24
Use utf8mb4 character set if possible

### LINKED ISSUE #8790: Tracking API: track Emoji correctly in page URLs and others
Goal: correctly track emoji and all utf8 4-byte characters. This requires a change from our table charset to `utf8mb4`. See discussion in #7766

### LINKED ISSUE #9785: Make Piwik database use utf8mb4 collation by default
When we will require MySQL 5.5.3 in Piwik, it would be desired to make our database use `utf8mb4` collation. This would solve some possibly weird issues and in general, is a best practise for properly managing unicode characters in MySQL. 
## Background information utf8mb4
- https://mathiasbynens.be/notes/mysql-utf8mb4
- Wordpress ticket of utf8mb45 migration https://core.trac.wordpress.org/ticket/21212
## Notes
- we wouldn't need to convert the database 
- only use `utf8mb4` if it's present (in case it could be disabled for some reasons)

### LINKED ISSUE #14485: A fatal error occurred - The string to escape is not a valid UTF-8 string
We are facing the below issues as we our url's have Arabic characters. This happens when track download files and file names are in Arabic.

Firstly we get a bad request response when we track the file download, then we applied the **utf8mb4** setting on mysql as suggested at issue #11136, after that we got a OK response from matomo. 

However, when we open the dashboard or visitors log we are getting error _"Oops… there was a problem during the request....."_.

**Errors:**
(1) The string to escape is not a valid UTF-8 string.
in D:\Matomo\matomo\plugins\Live\templates\getLastVisitsStart.twig line 73  

(2) The string to escape is not a valid UTF-8 string.
in D:\Matomo\matomo\plugins\CoreHome\templates\_dataTable.twig line 67  

**Screenshot:**
matomo-1

**Environment Information:**
Matomo v3.9.1
MySQL 8.0.15 (Community - Win64) 
IIS 10
Windows Server 2019 (1809)

**MySQL Configurations**
mysql-1

**Tracking Url's** to try to reproduce the issue
http://10.10.158.242:6660/piwik.php?idsite=1&rec=1&apiv=1&r=946539&uid=fa%40ecas.com&_idts=1558851102&_idvc=0&_id=909de6088a99b7d5&url=http%3a%2f%2f10.10.158.242%3a4040%2fECASLibrary002%2f%d8%b3%d8%ac%d9%84+%d8%b9%d9%85%d9%84+2019_DMA_00006+(1)%2f%d9%86%d9%85%d9%88%d8%b0%d8%ac+%d8%b9%d8%b1%d8%b6+%d8%a7%d9%84%d8%ac%d9%87%d8%a9.pptx&urlref=http%3a%2f%2f10.10.158.242%3a4040%2findex.aspx&dimension1=8e7449cd-ba1c-e611-80d2-0050568e7fa1&dimension2=External&dimension3=1cfc40ed-7b7f-e911-89cb-005056bb5a0e&dimension4=ec_workitem&download=http%3a%2f%2f10.10.158.242%3a4040%2fECASLibrary002%2f%d8%b3%d8%ac%d9%84+%d8%b9%d9%85%d9%84+2019_DMA_00006+(1)%2f%d9%86%d9%85%d9%88%d8%b0%d8%ac+%d8%b9%d8%b1%d8%b6+%d8%a7%d9%84%d8%ac%d9%87%d8%a9.pptx

http://10.10.158.242:6660/piwik.php?idsite=1&rec=1&apiv=1&r=574682&uid=fa%40ecas.com&_idts=1558851106&_idvc=0&_id=909de6088a99b7d5&url=http%3a%2f%2f10.10.158.242%3a4040%2fECASLibrary002%2f%d8%b3%d8%ac%d9%84+%d8%b9%d9%85%d9%84+2019_DMA_00006+(1)%2f%d8%af%d8%b1%d8%a7%d8%b3%d8%a9+%d8%a7%d9%84%d9%82%d8%b7%d8%a7%d8%b9+-+2017_00020-+SectorAnalysis.docx&urlref=http%3a%2f%2f10.10.158.242%3a4040%2findex.aspx&dimension1=8e7449cd-ba1c-e611-80d2-0050568e7fa1&dimension2=External&dimension3=1cfc40ed-7b7f-e911-89cb-005056bb5a0e&dimension4=ec_workitem&download=http%3a%2f%2f10.10.158.242%3a4040%2fECASLibrary002%2f%d8%b3%d8%ac%d9%84+%d8%b9%d9%85%d9%84+2019_DMA_00006+(1)%2f%d8%af%d8%b1%d8%a7%d8%b3%d8%a9+%d8%a7%d9%84%d9%82%d8%b7%d8%a7%d8%b9+-+2017_00020-+SectorAnalysis.docx

http://10.10.158.242:6660/piwik.php?idsite=1&rec=1&apiv=1&r=075426&uid=fa%40ecas.com&_idts=1558851118&_idvc=0&_id=909de6088a99b7d5&url=http%3a%2f%2f10.10.158.242%3a4040%2fECASLibrary002%2f%d8%b3%d8%ac%d9%84+%d8%b9%d9%85%d9%84+2019_DMA_00006+(1)%2f%d9%86%d9%85%d9%88%d8%b0%d8%ac+%d8%a7%d9%84%d8%b9%d8%b1%d8%b6+%d8%a7%d9%84%d8%aa%d9%82%d8%af%d9%8a%d9%85%d9%8a+%d9%84%d9%84%d9%85%d8%ac%d9%84%d8%b3+%d8%a7%d9%84%d8%aa%d9%86%d9%81%d9%8a%d8%b0%d9%8a.pptx&urlref=http%3a%2f%2f10.10.158.242%3a4040%2findex.aspx&dimension1=8e7449cd-ba1c-e611-80d2-0050568e7fa1&dimension2=External&dimension3=1cfc40ed-7b7f-e911-89cb-005056bb5a0e&dimension4=ec_workitem&download=http%3a%2f%2f10.10.158.242%3a4040%2fECASLibrary002%2f%d8%b3%d8%ac%d9%84+%d8%b9%d9%85%d9%84+2019_DMA_00006+(1)%2f%d9%86%d9%85%d9%88%d8%b0%d8%ac+%d8%a7%d9%84%d8%b9%d8%b1%d8%b6+%d8%a7%d9%84%d8%aa%d9%82%d8%af%d9%8a%d9%85%d9%8a+%d9%84%d9%84%d9%85%d8%ac%d9%84%d8%b3+%d8%a7%d9%84%d8%aa%d9%86%d9%81%d9%8a%d8%b0%d9%8a.pptx

sgiehl:
fixes #9785


sgiehl:
@tsteur @diosmosis  could you maybe have a first look at the PR?
In general the changes should already work. But for sure will need some more excessive testing

What still needs to be clarified:
- When updating to Matomo 4, or installing it newly then, it can still be the case that `utf8` charset will be used, if `utf8mb4` is not available. If that is the case, the update to `utf8mb` will be actually never tried again. Wondering if we should maybe try to do the conversion for every update automatically if `utf8mb4` is not yet configured.

- Using the database charset config to determine if the tables actually use utf8mb4 might not be very save. Someone could maybe change the charset even though the update wasn't performed. Not sure if we should handle that case and maybe store the used charset in the option table instead or maybe simply ignore that case, as it likely shouldn't happen often. Also saving the config in the update process might fail or not work on multi server env 🤔 

- I've changed the length of various fields, so the key length fits again. Imho the new length shouldn't make any problem, but be good if someone else have a thought about that as well...

core/DbHelper.php
```diff
@@ -204,6 +204,42 @@ public static function tableHasIndex($table, $indexName)
...
+
+        if (!empty($result) && $result['Value'] === 'utf8mb4') {
+            return 'utf8mb4'; // database has utf8mb4 charset, so assume it can be used
+        }
+
+        $result = Db::get()->fetchRow("SHOW VARIABLES LIKE 'innodb_file_per_table'");
+
+        if (empty($result) || $result['Value'] !== 'ON') {
+            return 'utf8'; // innodb_file_per_table is required for utf8mb4
+        }
```
diosmosis, line 233: Should we cache the result of this function? And haven't been following this issue, but do we need to do the `innodb_file_per_table` check if `character_set_database` is set to `utf8mb4`, or can we just assume it will work?

diosmosis: Or maybe we can set `Config::getInstance()->database['charset']` if it's not set to the default...

diosmosis: Actually I see this is already done during installation/update, so maybe that's enough

diosmosis:
> Wondering if we should maybe try to do the conversion for every update automatically if utf8mb4 is not yet configured.

It would have to be a major update though, correct? Since it would update a bunch of columns (not sure if those affect log tables). If it's too intensive an operation, maybe it could be a command that users can run when they're ready.

> Using the database charset config to determine if the tables actually use utf8mb4 might not be very save.

Would it help to check the database charset config, and check if a table has the right charset/column length to check if the update was run? This would still be a problem if some tables have different charsets, but that sounds like a real edge case.

> I've changed the length of various fields, so the key length fits again. Imho the new length shouldn't make any problem, but be good if someone else have a thought about that as well...

Only one that looks like it could be an issue is the option_name change. There might be some dynamically created one that somehow gets too long. Maybe we can report when the option name is too long somehow?

Findus23:
Just for reference, this is how Nextcloud handled the upgrade: (They did it in a console command)

https://docs.nextcloud.com/server/18/admin_manual/configuration_database/mysql_4byte_support.html

sgiehl:
Sure. Providing a command to do the migration makes it possible for those people that have shell access. But those using a shared hoster without shell access, wouldn't have the possibility to upgrade later. But maybe we could force the same update in Matomo 5 (for those that haven't switched till then). So using utf8mb4 is kind of optional for Matomo 4, but a requirement for Matomo 5. Would that be reasonably @mattab @tsteur

diosmosis:
The mechanism to make the change could also be in the UI if needed, right? Doesn't have to be a command.

sgiehl:
@diosmosis I've updated the PR.

So here a summary of the current behavior. Maybe @tsteur or @mattab have some thoughts/comments on that as well:

* When updating to Matomo 4
  * all columns used in database indices that are too long will be adjusted in length
  * In case utf8mb4 is available all tables will be converted as well. 
  * The used database charset in config will be automatically set to `utf8` or `utf8mb4` depending on the availability of `utf8mb4`  (when set to `utf8mb4` we assume it's available and do not replace 4 byte chars in tracking requests)
* If the conversion did not take place while updating
  * There is a new console command `core:convert-to-utf8mb4`, which will trigger converting the tables and updating the used charset in config
  * For people not being able to run a console command I would suggest to add a new FAQ explaining how to convert all tables manually and adjust the config afterwards
* New installs will automatically use `utf8mb4` and fallback to `utf8` if not available
* There is an (optional) diagnostic check that checks if `utf8mb4` is available and in use. If not available and/or not in use it will show a warning and suggest to run the command if possible
* There's (already) an update for Matomo 5.0.0-b1, which will force the conversion to utf8mb4 if it was not done before. I'd also suggest to create a follow up issue to remove `utf8` support for Matomo 5 and add it as requirement.

plugins/Diagnostics/lang/en.json
```diff
@@ -9,6 +9,8 @@
         "HideUnchanged": "If you want to see only changed values you can %1$shide all unchanged values%2$s.",
         "Sections": "Sections",
         "DatabaseReaderConnection": "Database Reader Connection",
+        "DatabaseUtf8mb4CharsetRecommended": "Your database doesn't support utf8mb4 charset. This is required to be able to store four byte UTF8 characters. Unless utf8mb4 is available those characters will be replaced.",
```
mattab, line 12: be good to clarify the `will be replaced` so it's clear, maybe something like `will be removed from the Tracking API values. For example special characters such as Emojis, x, y, z will be removed from your Page URLs, Page titles, custom dimension values, etc.`

mattab:
> There's (already) an update for Matomo 5.0.0-b1, which will force the conversion to utf8mb4 if it was not done before. 

just checking, we wouldn't force the conversion if it's not available though, to prevent errors/breaking instance?

plugins/Diagnostics/lang/en.json
sgiehl: Actually those chars are currently replaced with `�`. updated the text and added some examples

sgiehl:
> just checking, we wouldn't force the conversion if it's not available though, to prevent errors/breaking instance?

I've added that for Matomo 5. Matomo 4 won't force anything. But I think until we release Matomo 5 there will be some time gone by and as utf8mb4 is already default collation for newer releases of MySQL I assume we could maybe require it for Matomo 5. But that is something we can see then

core/DbHelper.php
tsteur: Was wondering the same if we need to check for `SHOW VARIABLES LIKE 'character_set_database'` after `innodb_file_per_table `?

core/Updates/4.0.0-b1.php
```diff
@@ -121,6 +142,11 @@ public function doUpdate(Updater $updater)
             // switch to default provider if GeoIp Legacy was still in use
             LocationProvider::setCurrentProvider(LocationProvider\DefaultProvider::ID);
         }
+
+        // update currently used database charset in config
+        $config = Config::getInstance();
```
tsteur, line 147: BTW... I wonder if it be easy enough to do similar to DB and plugin whether we could add a 

```php
$this->migration->config->changeConfigValue('database', 'charset', DbHelper::getDefaultCharset());
```

This command would then print a command like `./config:set --section=database --key=charset --value=XYZ`
in the list of commands to execute just to make sure this one isn't missed. I suppose if this wasn't executed for some reason otherwise it could cause issues?

core/Updates/5.0.0-b1.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        $this->migration = $factory;
+    }
+
+    public function getMigrations(Updater $updater)
+    {
+        $migrations = [];
+
+        // force utf8mb4 conversion in Matomo 5
+        if (DbHelper::getUsedCharset() !== 'utf8mb4') {
```
tsteur, line 38: @sgiehl is there a reason to force it in Matomo 5? Is it easy to still support utf8 basically or is there extra code for it?

core/DbHelper.php
```diff
@@ -204,6 +204,62 @@ public static function tableHasIndex($table, $indexName)
...
+
+        return 'utf8mb4';
```
tsteur, line 235: Similar to https://docs.nextcloud.com/server/18/admin_manual/configuration_database/mysql_4byte_support.html do we also need to check if ` show variables like 'innodb_file_format' == 'Antelope '` and if so prefer utf8?

tsteur:
* Could we also maybe have a diagnostic check that shows if utf8 or utf8mb4 is used and if utf8 is used, point them to an FAQ on how to upgrade using the console command (yes if they are on a hoster they might not be able to execute it, we can explain this in FAQ)

* Would it be useful for users to be able to specify an option to not execute the migration to utf8mb4? There might be some users for some reasons we're not aware yet where they maybe don't want it or where it cannot work? Could have a simple DB option like `database[no_upgrade_utf8mb4=1]`

tsteur:
> Would it be useful for users to be able to specify an option to not execute the migration to utf8mb4? There might be some users for some reasons we're not aware yet where they maybe don't want it or where it cannot work? Could have a simple DB option like database[no_upgrade_utf8mb4=1]

And we might also need to have the possibility to define some global variable or so to disable it. Thinking about WordPress... only realising now this seems to be using so far 
image

Wonder if this already may or may not cause random issues and I'd need to probably be able to control whether Matomo uses utf8mb4 or not (depending on `DB_CHARSET` constant)
Wondering if we already need to do something there ASAP and whether it causes issue when someone uses some other charset (need to add this to Matomo for WP system report). We're using the WordPress DB connection so we cannot control this separately.

Findus23:
I am not sure how many hosts are left without utf8mb4 support. It seems to have been added in 5.5.3 (released in 2010) , which means the last version without it is 5.1 which is end of life since 2013.

But there might still be reasons left for some people to not switch over.

core/DbHelper.php
```diff
@@ -204,6 +204,62 @@ public static function tableHasIndex($table, $indexName)
...
+        $queries[] ="ALTER DATABASE $database CHARACTER SET = utf8mb4 COLLATE = utf8mb4_unicode_ci;";
+
+        foreach ($allTables as $table) {
+            $queries[] = "ALTER TABLE `$table` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;";
+        }
+
+        return $queries;
+    }
+
+    public static function getUsedCharset()
```
tsteur, line 258: btw is this maybe rather something for the `Db/Settings` which we can use in DI and already use eg for table prefix etc

tsteur:
> I am not sure how many hosts are left without utf8mb4 support. It seems to have been added in 5.5.3 (released in 2010) , which means the last version without it is 5.1 which is end of life since 2013.

> But there might still be reasons left for some people to not switch over.

I think in the issue itself there was some mention where it may not work due to limit of file handles. Also Matomo for WordPress would use some limitations as they might support utf8mb4, but use a different charset. Eg utf8. 

If it's easy for us to not force people to use utf8mb4 that be great.

core/Updates/5.0.0-b1.php
sgiehl: It's basically only the replacement of unsupported chars. We can also remove the update for 5.0. and create a new issue to reconsider that later...

core/Updates/4.0.0-b1.php
sgiehl: makes sense... will add that

core/DbHelper.php
sgiehl: Actually not, as we currently do not require `large_prefix_index`, which doesn't work with `Antelope`. If we decide to require `large_prefix_index`, to make it possible to have key length greater than 255, we also need to require `Barracuda`

core/DbHelper.php
sgiehl: ah sure. wasn't aware of that classe

sgiehl:
@tsteur I've applied some review feedback.

> Could we also maybe have a diagnostic check that shows if utf8 or utf8mb4 is used and if utf8 is used, point them to an FAQ on how to upgrade using the console command (yes if they are on a hoster they might not be able to execute it, we can explain this in FAQ)

There is a diagnostic check. If utf8mb4 is available, but not used, it will show the command that can be used to migrate to utf8mb4. We can add a link to an FAQ there, once we have one.
If utf8mb4 is not available, it shows a message that unsupported characters are replaced in tracking requests.

> Would it be useful for users to be able to specify an option to not execute the migration to utf8mb4? There might be some users for some reasons we're not aware yet where they maybe don't want it or where it cannot work? Could have a simple DB option like database[no_upgrade_utf8mb4=1]

We can also remove it from the update script completely and only update to utf8mb4 using the console command. Most users won't have any problems using utf8. And for those asking for emojis and stuff like that, we have a possibility to update using a console command. And we could even provide an FAQ how to do it manually.

core/Updates/5.0.0-b1.php
tsteur: Be great to generally just perform the migration to utf8mb4 again just like in Matomo 4, but not do it if it isn't possible.

core/Updates/4.0.0-b1.php
```diff
@@ -95,6 +95,27 @@ public function getMigrations(Updater $updater)
...
+
+        foreach ($allTables as $table) {
+            if (preg_match('/archive_/', $table) == 1) {
+                $tableNameUnprefixed = Common::unprefixTable($table);
+                $migrations[] = $this->migration->db->changeColumnType($tableNameUnprefixed, 'name', 'VARCHAR(190)');
+            }
+        }
+
+        // Convert all tables to utf8mb4 if it is available
+        if ('utf8mb4' === DbHelper::getDefaultCharset()) {
```
tsteur, line 113: as mentioned for Matomo for WordPress I'd kind of need a way to not execute this unless the user actually uses that charset in WP. 

Also some people will maybe not want to upgrade to utf8mb4 for some reason so we need to have a way to force utf8 (or some other charset) in that method.

sgiehl:
@tsteur  I've spent some more thoughts on that topic. Actually I would prefer to not do the utf8mb4 update automatically at all. I would let new installs use utf8mb4 if it is available and provide the command to convert everything to utf8mb4 (and a FAQ how to do manually). Not using utf8mb4 has only one effect at the moment. And that is the replacing of unsupported characters. As that only affects a minority I actually don't see much benefit in forcing most instances to convert their database to utf8mb4, which might take a while depending on the size.
That way there shouldn't be any problems with that while updating and no problems with wordpress. So we lower the risks but still provide people requiring `utf8mb4` the possibility to use it.

Note: Currently all MySQL versions are handling `utf8` as alias of `utf8mb3`. I would suggest to start a new discussion about automatically updating or forcing `utf8mb4` as soon as MySQL changes the alias of `utf8` to `utf8mb4`  (what they announced to do at some point.)

tsteur:
@sgiehl that would be fine by me. As the update scripts can take quite a while it would also allow high traffic instances to plan better for it (also us on the cloud where we wouldn't want to do this during a regular update). We could show in the diagnostic check that they can upgrade to utf8mb4 if their DB supports it etc. 

Wonder if we'd also offer a one click button to perform the update for users who can't run commands or do it manually and they have not much data so it goes quick? Of course there's a risk it'll be executed during a scheduled task in a tracking request etc and things may time out or something ... but suppose that be also a risk regularly. Or maybe we offer a button for users where we know we can launch a cliMulti request in the background ($cliMulti->supportsAsync == true)

sgiehl:
@tsteur started to change the implementation. The conversion will no longer be done during the update. There will now be an additional setting (under update settings), so trigger the conversion to utf8mb4. The conversion itself will then be done as scheduled task once. Maybe you could have a quick look if it is fine that way. Would adjust / improve the code then tomorrow. The text messages displayed anywhere will need to improved as well, so it's clear it's possible to use the setting or the command...

tsteur:
image

Could maybe also say here alternatively the upgrade can be triggered through the UI in the background... and could also 

image

could we mention here that performing this upgrade can take very long (up to many hours) and cause downtime etc? Do we even need to link to an FAQ or explain everything in there? Should we maybe mention for high traffic sites we recommend disabling the tracker while performing this update? And recommend to not use the UI setting but instead use the command?

In both places in diagnostic checks and the UI setting be great to also explain what they gain by performing this upgrade since most users won't know what UTF8MB4 is, whether it represents an issue for them, whether they should do it or not etc.

Would have here https://github.com/matomo-org/matomo/pull/15618/commits/4f8abcdd1f1e16bfc7a340cd76455939d2afe400#diff-33d27fe974b95654859eba7890184c33R46 maybe expected to see utf8mb4 by default and then have an update that sets it back to utf8? Or how do we make sure that new installs use utf8mb4 when possbile?

sgiehl:
@tsteur will improve the displayed messages now.

> Would have here 4f8abcd#diff-33d27fe974b95654859eba7890184c33R46 maybe expected to see utf8mb4 by default and then have an update that sets it back to utf8? Or how do we make sure that new installs use utf8mb4 when possbile?

We actually can't set `utf8mb4` as default in the global config. Someone doing an update might otherwise end up in a broken Matomo maybe. Let's say someone's database does not support `utf8mb4` and he triggers an update to Matomo 4. If he doesn't have set a charset in his local config `utf8mb4` would now be assumend until the update script sets the local config. But before that some other updates might run trying to create tables with `utf8mb4` or even the database connection itself might fail as the charset set for the connection doesn't exist.
Nevertheless for new install `utf8mb4` should be used, as it sets the database charset in local config when creating the config file here:
https://github.com/matomo-org/matomo/blob/6a17b9cfb7475d307fb45e4df6c1119c3a583640/plugins/Installation/Controller.php#L602
But actually I just realized that the database might be created before the config file is written. So will change the database creation to always use the detected charset. Guess that shouldn't be a problem for wordpress, as the database always exists there, right?

Note: people manually creating the config file or using some other process to install Matomo might end up still using `utf8`, but that's nothing we can circumvent I guess.

sgiehl:
@tsteur There's something else we might need to find a solution for. The conversion currently converts all tables returned by `DbHelper::getTablesInstalled()`. That does only include the tables defined in the Schema and all archive tables. So the conversion won't target any tables additionally created by any plugin. So all those tables would possibly remain with `utf8`.  Wondering if it would make sense to add a new hook into the `getTablesInstalled` method, so plugin can kind of register there tables 🤔

tsteur:
> But actually I just realized that the database might be created before the config file is written. So will change the database creation to always use the detected charset. Guess that shouldn't be a problem for wordpress, as the database always exists there, right?

I guess DB creation wouldn't affect it WP as it is indeed already created. Also in general I think it wouldn't have a big effect as long as we always set encoding when creating a table.

> Wondering if it would make sense to add a new hook into the getTablesInstalled method, so plugin can kind of register there tables 🤔

image

Could maybe change the intersect to `array_unique(array_merge($allMyTables, $allTables))`? I reckon that would be fine... I guess it be an issue though when no table prefix is used... so might not work and a hook be needed. Guess sounds good to add the hook. Bit annoying as it's unexpected for developers but if they missed implementing it, it likely wouldn't cause an issue. Also maybe if a table prefix is defined, we could maybe always add all tables that were found? Maybe not though as it could lead to hard to find/debug issues...

sgiehl:
@tsteur I've add a new event. Nice side effect: the DBStats plugin will finally show really all tables, and not only those created by the core schema 🙈 
Will now create PRs for all our plugins to implement that event...

sgiehl:
@tsteur update the PR with the new event and created PRs for (most of) our plugins. If the event is fine that way, feel free to quickly check the changes in the linked PRs and merge them, so we can update the submodules here again.

plugins/Dashboard/Dashboard.php
```diff
@@ -32,10 +32,21 @@ public function registerEvents()
...
     }
 
+    /**
+     * Register the new tables, so Matomo knows about them.
+     *
+     * @param array $allTablesInstalled
+     */
+    public function getTablesInstalled(&$allTablesInstalled)
+    {
+        $allTablesInstalled[] = Common::prefixTable('user_dashboard');
```
tsteur, line 47: @sgiehl is it possible these tables are installed even when the plugin is deactivated? It wouldn't find this table then? Not sure we can do much about this? guess for core plugins we could at least have them all in corehome for example but would still be an issue for other plugins. Not sure where we call the getDbInstalled event and whether we can in these cases temporarily load all plugins or not. (might cause other issues). Not sure there's a solution to it?

sgiehl: Yes it can happen, that those tables are present, but the plugin is not enabled. On the other hand it can happen that the tables are not yet created, so moving them to corehome might cause other issues. But that also applies for all additional plugins one has installed.
I've adjusted the method so it loads all plugins before posting the event.

tsteur: Went through all usages of the method and might be fine to load all plugins. However, `ArchiveTablesCreator::createArchiveTablesIfAbsent()` might be a problem? The `getNumericTable` and `getBlobTable` seem to cause getting a list of all tables causing maybe a performance regressions. Does methods are used very often in many places and there could be performance regressions. Eg every time we load an archive we might fetch all tables etc. Could you double check this and potentially change the `ArchiveTableCreator` class to not use regular `getTablesInstalled` method or so? Maybe it just needs to check manually for a list of all `Common::prefixTable('archive_')` tables and use this list? To avoid loading and unloading plugins and also to avoid triggering event etc.

sgiehl: implemented a simple new method in ArchiveTablesCreator to only fetch the archive tables.

core/DbHelper.php
```diff
@@ -204,6 +204,57 @@ public static function tableHasIndex($table, $indexName)
...
+     *
+     * @return array
+     */
+    public static function getUtf8mb4ConversionQueries()
+    {
+        $allTables = DbHelper::getTablesInstalled();
+        $database  = Config::getInstance()->database['dbname'];
+
+        $queries   = [];
+        $queries[] ="ALTER DATABASE $database CHARACTER SET = utf8mb4 COLLATE = utf8mb4_unicode_ci;";
```
tsteur, line 249: Actually wondering if this one is really needed @sgiehl ? Eg thinking of WordPress or other applications where the same DB might be used for other applications too this might not be expected. I reckon it be good to remove this if possible. It could have many side effects when people use same DB for multiple apps which many do

tsteur (commented):
@sgiehl there's still a test failing in https://travis-ci.org/github/matomo-org/matomo/jobs/686048413#L668

Would expect that only tables starting with `archive_` are returned.

Otherwise looked good. The only thing I would maybe do is when executing the command instead of showing the table names
image

Maybe we could show all the SQL queries we will be executing in case someone wants to execute these manually? And then show at the end a command `./console config:set --section=database --key=charset --value=utf8mb4`. I reckon some bigger installs like us might prefer to execute these queries manually as it can take a long time.  Eg you want to update one log table over night, then another table another night when there is low traffic etc. (as you have to pause tracking during this time likely).  There could also be a question whether they want the SQL plus config set instructions to be printed in order to execute it manually or whether they want it to be executed automatically?

This also makes me think if there should be an option to automatically disable tracking while the log tables are being migrated. Not sure but this might be even a good idea to do by default? If someone has low traffic, then tracking be down only for short time anyway. If someone has a a big DB, then they might get in trouble when keeping tracking enabled as I'd expect the inserts to be not processed but haven't tested this. It seems converting charset has no online DDL supporthttps://dev.mysql.com/doc/refman/5.6/en/innodb-online-ddl-operations.html#online-ddl-table-operations so would expect it might be best to disable tracking during the migration of log tables. fyi @mattab

core/DbHelper.php
```diff
@@ -204,6 +204,57 @@ public static function tableHasIndex($table, $indexName)
...
+
+        foreach ($allTables as $table) {
+            $queries[] = "ALTER TABLE `$table` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;";
```
tsteur, line 252: btw @sgiehl do we need to specify collation? Not sure if that particular one will always be available?

core/DbHelper.php
sgiehl: will remove it

core/DbHelper.php
sgiehl: should also work without it. Mysql should then use the default collation for utf8mb4

sgiehl:
@tsteur applied some feedback and adjusted the command so it is able to disable tracking and only show all required commands...

plugins/CoreUpdater/SystemSettings.php
```diff
@@ -54,6 +59,13 @@ protected function init()
         $isWritable = $isWritable && PluginUpdateCommunication::canBeEnabled();
         $this->sendPluginUpdateEmail = $this->createSendPluginUpdateEmail();
         $this->sendPluginUpdateEmail->setIsWritableByCurrentUser($isWritable);
+
+        $isWritable = Piwik::hasUserSuperUserAccess() && CoreAdminController::isGeneralSettingsAdminEnabled();
+        $dbSettings = new Settings();
+        if ($dbSettings->getUsedCharset() !== 'utf8mb4' && DbHelper::getDefaultCharset() === 'utf8mb4') {
```
tsteur, line 65: just fyi was worried this might issue DB requests during tracking but double checked looks like these system settings aren't used during tracking 👍  Might be good though to change the `if` to `if ($isWritable && $dbSettings->getUsedCharset() ...)`? This way eg on cloud we avoid these queries there.

tsteur: This should be last thing. feel free to merge if/once tests pass

sgiehl:
@tsteur I've merge all related PRs already, but had to change something here in order to get the tests running. See https://github.com/matomo-org/matomo/pull/15618/commits/7b06248635e3f495f52f358d14af1d94b5d2adff  
Maybe you can have a quick look at that commit and merge the PR if it's fine...

tsteur:
@sgiehl not sure...
in https://github.com/matomo-org/matomo/commit/7b06248635e3f495f52f358d14af1d94b5d2adff#diff-6bfd1a5b94264d4d0e0c03e1ebd70249R446 should it be maybe `getPluginsLoadedAndActivated` instead of only getting loaded plugins? Do you know why `loadActivatedPlugins` didn't work? Bit scared we're loading more plugins than we should.

BTW another option could be when posting the event to try

```php
Piwik::postEvent('Db.getTablesInstalled', [&$allMyTables], false, (Manager::getAllPluginsNames()); // and if this doesn't work could try the array_keys($loadedPlugins)
```

sgiehl:
@tsteur actually it should load the plugins that were loaded before.  But I can try to revert that to `loadActivatedPlugins`. Guess the if statement around was more important, as otherwise `isInstalled` fired the event. will check that

btw. Passing the plugins to the postEvent won't work, as the method only checks loaded plugins.

sgiehl:
@tsteur using `loadActivatedPlugins` works as well now. Changed it again. So feel free to merge if there isn't anything else

PR #15619, opened 2020-02-24
Prevent server gone away error in CronArchive

tsteur:
What happens is basically eg some archive requests run for 6 hours, then it returns back into the `CronArchive` scripts, it tries to execute a query, and results in a server gone away error because the DB connection wasn't used for hours.

Fixes only when a CronArchive happens directly in CronArchive.php script. We've tested this in production for one month and worked nicely.

### DIFF
--- a/core/CronArchive.php
+++ b/core/CronArchive.php
@@ -273,6 +273,8 @@ class CronArchive
      */
     private $isArchiveProfilingEnabled = false;
 
+    private $lastDbReset = false;
+
     /**
      * Returns the option name of the option that stores the time core:archive was last executed.
      *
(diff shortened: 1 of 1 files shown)


core/CronArchive.php
```diff
@@ -1234,12 +1239,30 @@ private function request($url)
...
         }
         return false;
     }
 
+    private function disconnectDb()
+    {
+        if (empty($this->lastDbReset)) {
+            $this->lastDbReset = time();
+            return;
+        }
```
sgiehl: That seems to be dead code. `$this->lastDbReset = time();` is already done in `__construct`, so the value can't ever be empty

sgiehl (changes requested):

tsteur:
@sgiehl updated

sgiehl (approved):

PR #15621, opened 2020-02-25
Fix failing Live system test

sgiehl:
Investigated why one System test of Live plugin is failing on `4.x-dev` but not on `3.x-dev`. Looking back through the builds it turned out the reason actually is #14847, which changed the `referrer_url` column from `TEXT` to `VARCHAR(4000)`. I'm able to reproduce that locally. Changing back to `TEXT` fixes the test again.
I guess that is a sorting issue in MySQL. Most attributes of the visits are identical, and the referrer urls are even empty. Not sure if there is anything we need to adjust. Might be fine to simply update the expected file.

ping @diosmosis @tsteur


tsteur:
Should be fine to merge @sgiehl good find 👍

PR #15624, opened 2020-02-26
Enable fingers crossed handler by default and add optional support in CLI mode

### LINKED ISSUE #15220: Enable fingers crossed handler by default
Now that the fingers crossed handler is available we should enable it by default to be able to get as much info from bug reports as possible.

diosmosis:
Fixes #15220

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -122,6 +122,12 @@ logger_file_path = tmp/logs/matomo.log
 ; NOTE: this only has an effect when logging to a file or when running CLI commands.
 enable_fingers_crossed_handler = 1
 
+; whether to enable the "fingers crossed" logging handler during CLI commands. this will result in much more
+; output on errors, all of which will be identical to normal debug output. this is mostly useful if you run
+; commands in an automated manner (for example running core:archive on a cron) and would like debugging
+; information when an error occurs.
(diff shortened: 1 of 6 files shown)


diosmosis:
@mattab ran into blocking issue w/ DI that I can't figure out just yet, think it would be good to move out of 4.0.

mattab:
What's the issue @diosmosis ? we can move this out if the issue can't be resolved.

diosmosis:
@mattab I couldn't add definitions to the climulti environment to add the fingers crossed handler. The file was always included in php-di and php-di saw the new definitions, but php-di never used them. It would take a while to debug unfortunately.

mattab:
@tsteur maybe you have some thoughts on this?

tsteur:
@mattab best to just close the PR. I'm not seeing a big need for this feature anyway

mattab:
Moved it out of 4.0.0 for now.

Btw the great value of this feature is that for example, when core:archive fails due to a SQL query error, then we can see in the core:archive error output the full SQL query that caused the error (for example a timeout). This is valuable in identifying the root cause of the core:archive error and quickly find a faq or solution to the problem (or reaching out to support with the query).

PR #15627, opened 2020-02-26
Remove gears detection 

pebosi:
Removing detection of old Gears plugin, was developed by Google. It was stopped in 2011: https://gearsblog.blogspot.com/2011/03/stopping-gears.html


sgiehl (changes requested):
Well, development of Gears has been discontinued a while back already and newer browser don't have that anymore. On matomo.org there had been 8 visits with Gears enabled last year. Wondering if it would be still worth to keep old data 🤔 

ping @tsteur @mattab

tests/resources/piwik-1.13-dump.sql
```diff
@@ -445,7 +445,6 @@ CREATE TABLE `log_visit` (
   `config_quicktime` tinyint(1) NOT NULL,
   `config_realplayer` tinyint(1) NOT NULL,
   `config_windowsmedia` tinyint(1) NOT NULL,
-  `config_gears` tinyint(1) NOT NULL,
```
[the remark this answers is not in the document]
pebosi: Reverted

plugins/DevicePlugins/Updates/4.0.0.php
```diff
@@ -0,0 +1,65 @@
...
+
+use Piwik\Updater;
+use Piwik\Updates as PiwikUpdates;
+use Piwik\Updater\Migration;
+use Piwik\Updater\Migration\Factory as MigrationFactory;
+
+/**
+ * Update for version 4.0.0.
+ */
+class Updates_4_0_0 extends PiwikUpdates
```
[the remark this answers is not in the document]
pebosi: Done

core/Updates/4.0.0-b1.php
```diff
@@ -48,11 +51,35 @@ public function getMigrations(Updater $updater)
...
         }
 
+        if ($this->usesGeoIpLegacyLocationProvider()) {
+            // activate GeoIp2 plugin for users still using GeoIp2 Legacy (others might have it disabled on purpose)
+            $migrations[] = $this->migration->plugin->activate('GeoIp2');
+        }
+
+        // remove old options
+        $migrations[] = $this->migration->db->sql('DELETE FROM `' . Common::prefixTable('option') . '` WHERE option_name IN ("geoip.updater_period", "geoip.loc_db_url", "geoip.isp_db_url", "geoip.org_db_url")');
+
```
sgiehl: those change are unrelated to this PR. Guess they are a result of a merge conflict maybe? 🤔

tests/javascript/frameworks/dojo/dojo-1.0.3.js
```diff
@@ -2808,52 +2808,6 @@ if(!dojo._hasResource["dojo._base.window"]){ //_hasResource checks added by buil
 dojo._hasResource["dojo._base.window"] = true;
 dojo.provide("dojo._base.window");
 
-dojo._gearsObject = function(){
```
sgiehl: guess that can be reverted, as it's changing an external library, that's only used for tests...

tests/javascript/frameworks/dojo/dojo-1.1.2.js
```diff
@@ -3168,57 +3168,6 @@ if(!dojo._hasResource["dojo._base.window"]){ //_hasResource checks added by buil
 dojo._hasResource["dojo._base.window"] = true;
 dojo.provide("dojo._base.window");
 
-dojo._gearsObject = function(){
-	// summary:
```
sgiehl: same here

tests/resources/piwik-1.13-dump.sql
```diff
@@ -70,7 +70,7 @@ CREATE TABLE `archive_blob_2012_03` (
...
��%;Rs�?8+3T	�*�\'Nb(R�ӽ;	t�l�q9��3�\0\Z�\Z_�Q?z�)���y1�������\r(�4\'��.t�$a�6zc/Ot��m`�֣�mS�_Q\\r����
{���0�'),(5,'UserSettings_plugin',1,'2012-03-01','2012-03-31',3,'2013-10-08 21:55:44','x����j1�w�{a�U����H�2�:nl4ٺ��;�ŋ0�0L��1!!4ς��D3C��A��z�����\"
��H�@����	��;F�b.�
�58A�n�\n�E�h�³V��p�[�uL뵧�}����F;:Q�VЦ�Dn�3�sT@���_i�^��A�
��B�L���s朔@-� ?mҞ[�
���k�0Ŕ�MKL��$����f�g���D!t����\\�Z[>'),(5,'UserSettings_resolution',1,'2012-03-01','2012-03-31',3,'2013-10-08 21:55:44','x�-��\n�0D�e�B6m�8�zV�e�=
Z���n��73
3��`�|]Ӗ^�IV�˔�������Ѓ������(�4g���������)A��^}P�YVs-*~�2���D`Pj��cU?p
_gd1�'),(5,'UserSettings_wideScreen',1,'2012-03-01','2012-03-31',3,'2013-10-08 21:55:44','x�-���0
D��;R�Ҵ\\Vf@��r�CD�R��!ʿcC�ww>�9�\ng�t�k|�G��M�4��i%0�Bw\n��?t�6�$Ø(,��4�D��Q+�����ޱ����۔�WA�\\�����-|q�2k'),(5,'VisitorInterest_daysSinceLastVisit',1,'2012-03-01','2012-03-31',3,'2013-10-08
```
sgiehl: Those changes should reverted. Same applies to all changes in `tests/resources/OmniFixture-dump.sql`

sgiehl (changes requested):

Findus23:
While we are on it:

Should we maybe also remove Quicktime, Realplayer and Director (which I think is the shockwave player)?

All of those shouldn't work in any modern browser anymore (and as the report notes, detection doesn't work in IE, so there should be nothing left)

tests/javascript/frameworks/dojo/dojo-1.1.2.js
pebosi: Done

tests/resources/piwik-1.13-dump.sql
pebosi: Done

core/Updates/4.0.0-b1.php
pebosi: Merged again

pebosi:
> While we are on it:
> 
> Should we maybe also remove Quicktime, Realplayer and Director (which I think is the shockwave player)?
> 
> All of those shouldn't work in any modern browser anymore (and as the report notes, detection doesn't work in IE, so there should be nothing left)

i would create a pull request per Plugin

plugins/UserCountry/VisitorGeolocator.php
```diff
@@ -189,7 +189,7 @@ public function attributeExistingVisit($visit, $useClassCache = true)
             $this->logger->debug('Updating visit with idvisit = {idVisit} (IP = {ip}). Changes: {changes}', array(
                 'idVisit' => $idVisit,
                 'ip' => $ip,
-                'changes' => json_encode($valuesToUpdate)
+                'changes' => $valuesToUpdate
```
sgiehl: is that change on purpose?

pebosi: Removed

sgiehl:
code changes look fine so far. Need to find some time to check it out locally and do some testing...

mattab:
> Should we maybe also remove Quicktime, Realplayer and Director (which I think is the shockwave player)?

Probably a good idea, any thoughts @sgiehl @tsteur ?

tsteur:
I suppose should be fine. Not sure though re removing that data. You're basically removing historical data there of users. It may be fine as likely nobody looks at this anymore, but also feels a bit wrong.

sgiehl:
Looking at the failing tests it seems all fingerprints changed due to the changes here. While that is fine for the tests and we can simply update them. I'm wondering if that is something we should adjust actually, as fingerprints for already tracked users might change after an update and they might be detected as new users then 🤔 

Maybe we should use a default value `0` for `gears` in the config hash instead of removing it here:
https://github.com/matomo-org/matomo/pull/15627/files#diff-02edccc593e3badd1e9fd8d278fd1d35L113

Or is changing the fingerprints fine? (ping @mattab @tsteur )

mattab:
> Not sure though re removing that data. You're basically removing historical data there of users. It may be fine as likely nobody looks at this anymore, but also feels a bit wrong.

Let's at least mention this clearly in the Developer Changelog? (in a section "Database Schema" or so?)

> Maybe we should use a default value 0 for gears in the config hash instead of removing it here:

:+1:

core/Tracker/Settings.php
```diff
@@ -110,7 +108,7 @@ protected function getConfigHash(Request $request, $os, $browserName, $browserVe
               $os
             . $browserName . $browserVersion
             . $plugin_Flash . $plugin_Java . $plugin_Director . $plugin_Quicktime . $plugin_RealPlayer . $plugin_PDF
-            . $plugin_WindowsMedia . $plugin_Gears . $plugin_Silverlight . $plugin_Cookie
+            . $plugin_WindowsMedia . $plugin_Silverlight . $plugin_Cookie
```
sgiehl: As discussed we should use a `0` as default here instead, to avoid changes in config hashes
```suggestion
            . $plugin_WindowsMedia . '0' . $plugin_Silverlight . $plugin_Cookie
```

tsteur:
@pebosi @sgiehl is there much left to do here?

pebosi:
I thought not.

tsteur:
👍 be great to get this merged @sgiehl not sure if something is missing

sgiehl:
I'll recreate this PR with solved merge conflicts and fixed tests

PR #15630, opened 2020-02-28
Support signing up for WordPress newsletter

tsteur:
Will be needed for https://github.com/matomo-org/wp-matomo/issues/232

Not in changelog as it's only an internal change.


tsteur:
Actually, I think that won't be needed and will look into this if or when needed

PR #15632, opened 2020-02-28
Remove jquery placeholder

pebosi:
All major browsers support the attribute nativly


Findus23 (approved):
Should be fine as support is already very common (even IE 10 and 11):
https://caniuse.com/#feat=input-placeholder

sgiehl:
@pebosi  could you also remove it from the list in `bower.json`: https://github.com/matomo-org/matomo/blob/98fd1f6e4b854e095ac0c7fa551b6fe0ebdbd2c5/bower.json#L29

pebosi:
> @pebosi could you also remove it from the list in `bower.json`:
> 
> https://github.com/matomo-org/matomo/blob/98fd1f6e4b854e095ac0c7fa551b6fe0ebdbd2c5/bower.json#L29

Done

PR #15633, opened 2020-02-28
Fix session cookie expire

### LINKED ISSUE #15625: MATOMO_SESSID expires directive with wrong format
After logging in the MATOMO_SESSID Cookie is changed via the http header set-cookie and an expires directive is added. But the format of the value is unixtime instead of date (https://tools.ietf.org/html/rfc6265#section-4.1)

Example (Version 3.13.3):
```
Set-Cookie: MATOMO_SESSID=380lf4cjmaenc605u73dgklnq9; expires=1583920346; path=/; secure; httponly; SameSite=Lax
```

Found the following code line, perhaps it helps, but did not inspect deeper:
https://github.com/matomo-org/matomo/blob/35467f54cd3e1dae889d88ae4f80d4626f9127ed/core/Session.php#L210

Here it seems to be converted:
https://github.com/matomo-org/matomo/blob/35467f54cd3e1dae889d88ae4f80d4626f9127ed/core/Cookie.php#L155

sgiehl:
fixes #15625

### DIFF
--- a/core/Session.php
+++ b/core/Session.php
@@ -222,7 +222,7 @@ class Session extends Zend_Session
     {
         $headerStr = 'Set-Cookie: ' . rawurlencode($name) . '=' . rawurlencode($value);
         if ($expires) {
-            $headerStr .= '; expires=' . $expires;
+            $headerStr .= '; expires=' . gmdate('D, d-M-Y H:i:s', $expires) . ' GMT';
         }
         if ($path) {
             $headerStr .= '; path=' . $path;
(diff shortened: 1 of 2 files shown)


diosmosis:
Think we can add a unit test for this method? I know there isn't one already and there probably should've be one, thought it would be good to at least test this change imo.

tsteur:
Something like we have in `SessionTest` should work

diosmosis:
@tsteur you mean as an existing test or a place for a new test? I was thinking a simple unit test that tested the output of the method.

tsteur:
Unit test for output makes sense 👍

sgiehl:
added some tests

PR #15634, opened 2020-02-28
accept inavlid ssl cert for development

mikkeschiren:
A suggestion to except invalid ssl cert in curl request for development, I think it could help out people for some functionality.

### DIFF
--- a/core/Http.php
+++ b/core/Http.php
@@ -96,8 +96,13 @@ class Http
         // create output file
         $file = self::ensureDestinationDirectoryExists($destinationPath);
 
+        $acceptInvalidSslCertificate = false;
+        if (isset(Config::getInstance()->Development['accept_invalid_ssl'])) {
+            $acceptInvalidSslCertificate = Config::getInstance()->Development['accept_invalid_ssl'];
+        }
+
         $acceptLanguage = $acceptLanguage ? 'Accept-Language: ' . $acceptLanguage : '';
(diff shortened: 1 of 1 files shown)


sgiehl:
Guess would make sense to add a default value to global.ini.php with a comment what it's good for.
See https://github.com/matomo-org/matomo/blob/35467f54cd3e1dae889d88ae4f80d4626f9127ed/config/global.ini.php#L177-L188

Findus23:
Honestly I think this option should either only work when development mode is enabled or show a permanent, non-dismissable warning in Matomo (or at least the system check).
Otherwise it would be far to easy to accidently undermine a huge part of the security.

mikkeschiren:
Yeah agree - this should only work if Development module is active, and show a warning at system check - my goal with this first version on the pull request is to get the feedback. There are some problems with some development tools for some setups if you can not bypass invalid certs (local development mostly), but like said - it could be a security risk.

core/Http.php
```diff
@@ -96,8 +96,13 @@ public static function sendHttpRequest($aUrl,
         // create output file
         $file = self::ensureDestinationDirectoryExists($destinationPath);
 
+        $acceptInvalidSslCertificate = false;
+        if (isset(Config::getInstance()->Development['accept_invalid_ssl'])) {
```
tsteur, line 100: @mikkeschiren could you add a default value to the `config/global.ini.php` for `accept_invalid_ssl`? And also add `Development::isEnabled()` to the check?

tsteur: @mikkeschiren any chance you could check out my comment above and add the default value?

tsteur:
@mikkeschiren I'll close this PR for now as we haven't heard from you in a while but will be happy to reopen should you be keen to follow up on the other comments. Simply let us know and we'd be happy to review things again.

PR #15636, opened 2020-02-28
Replace jscrollpane with native

pebosi:
Scrolling the Segments list without old libs. Tested on Firefox / Chrome.


bower.json
```diff
@@ -24,8 +24,6 @@
     "ngDialog": "~1.3.0",
     "mousetrap": "~1.4.0",
     "sprintf": "~1.0.0",
-    "jScrollPane": "~2.0.0",
-    "jquery-mousewheel": "~3.1.12",
```
tsteur: @pebosi I suppose mousewheel was removed by accident?

tsteur: I see this was used in combination with jScrollPane.

tsteur:
Worked for me locally as well. I'm restarting the tests and will merge should all be fine. Always great to see when we can remove a big amount of code and use native solutions. Cheers @pebosi

tsteur:
Looks like no tests failed 👍

PR #15649, opened 2020-03-02
By default a segment should be visible to all users

tsteur:
@mattab any thoughts? Got this feedback yesterday on a call and I think it makes sense to share segments across the teams as mostly they will be wanting to use the same segments and it avoids everyone creating their own/similar segments.


mattab:
@tsteur 
Sounds like a good idea :+1: 

Notes:

* sharing a segment with all users is only accessible to Super Users. So we should do this only for Super Users as otherwise it wouldn't work
* will need to update the user guide @ https://matomo.org/docs/segmentation/#creating-a-segment-in-matomo

Screenshot from 2020-03-06 10-11-11

tsteur:
fixed tests

PR #15652, opened 2020-03-03
respect max execution time limit for transitions plugin

### LINKED ISSUE #15609: Limit max execution time of transitions feature
We should apply `live_query_max_execution_time` setting also for transition queries since it is basically also a live query. Ideally, if the limit is reached, similar to visitor log we would then show a message if the limit is reached and suggest to shorten the date range for example, or if a segment is selected mention could be caused by this.

refs https://github.com/matomo-org/matomo/issues/14762

sgiehl:
fixes #15609

### DIFF
--- a/core/DbHelper.php
+++ b/core/DbHelper.php
@@ -213,6 +213,33 @@ class DbHelper
     }
 
     /**
+     * Adds a MAX_EXECUTION_TIME hint into a SELECT query if $limit is bigger than 1
+     *
+     * @param string $sql  query to add hint to
+     * @param int $limit  time limit in seconds
+     * @return string
+     */
+    public static function addMaxExecutionTimeHintToQuery($sql, $limit)
+    {
+        if ($limit <= 0) {
+            return $sql;
+        }
+
+        $sql = trim($sql);
+        $pos = stripos($sql, 'SELECT');
+        if ($pos !== false) {
+
+            $timeInMs = $limit * 1000;
+            $timeInMs = (int) $timeInMs;
+            $maxExecutionTimeHint = ' /*+ MAX_EXECUTION_TIME('.$timeInMs.') */ ';
+
+            $sql = substr_replace($sql, 'SELECT ' . $maxExecutionTimeHint, $pos, strlen('SELECT'));
+        }
+
+        return $sql;
+    }
+
+    /**
      * Returns true if the string is a valid database name for MySQL. MySQL allows + in the database names.
      * Database names that start with a-Z or 0-9 and contain a-Z, 0-9, underscore(_), dash(-), plus(+), and dot(.) will be accepted.
      * File names beginning with anything but a-Z or 0-9 will be rejected (including .htaccess for example).
--- a/plugins/Live/Model.php
+++ b/plugins/Live/Model.php
@@ -16,6 +16,7 @@ use Piwik\Config;
 use Piwik\Container\StaticContainer;
 use Piwik\Date;
 use Piwik\Db;
+use Piwik\DbHelper;
 use Piwik\Period;
 use Piwik\Period\Range;
 use Piwik\Piwik;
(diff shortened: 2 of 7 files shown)


core/DbHelper.php
```diff
@@ -212,6 +212,33 @@ public static function deleteArchiveTables()
...
+
+            $timeInMs = $limit * 1000;
+            $timeInMs = (int) $timeInMs;
+            $maxExecutionTimeHint = ' /*+ MAX_EXECUTION_TIME('.$timeInMs.') */ ';
+
+            $sql = substr_replace($sql, 'SELECT ' . $maxExecutionTimeHint, $pos, strlen('SELECT'));
+        }
+
+        return $sql;
+    }
```
diosmosis: Can we add some unit tests for this logic?

plugins/Live/Model.php
```diff
@@ -130,57 +131,58 @@ private function executeLogVisitsQuery($sql, $bind, $segment, $dateStart, $dateE
...
 	 * @param $dateStart
 	 * @param $dateEnd
 	 * @param $minTimestamp
 	 * @param $limit
+	 * @param $parameters
 	 *
 	 * @throws MaxExecutionTimeExceededException
 	 */
-    public function handleMaxExecutionTimeError($readerDb, $e, $sql, $bind, $segment, $dateStart, $dateEnd, $minTimestamp, $limit)
+    public static function handleMaxExecutionTimeError($readerDb, $e, $segment, $dateStart, $dateEnd, $minTimestamp, $limit, $parameters)
```
diosmosis: If unit tests don't exist for this method can we add some? I'm not sure if it is easy to do however. Some sort of test covering this behavior would be good IMO

diosmosis:
Code looks good to me but I'm not too familiar with the related code, so I think @tsteur may want to take a look as well

plugins/Transitions/API.php
```diff
@@ -81,25 +84,39 @@ public function getTransitionsForAction($actionName, $actionType, $idSite, $peri
...
-                - $this->getTotalTransitionsToFollowingActions()
-                - $report['pageMetrics']['loops'];
+            // derive the number of exits from the other metrics
+            if ($parts == 'all') {
+                $report['pageMetrics']['exits'] = $report['pageMetrics']['pageviews']
+                    - $this->getTotalTransitionsToFollowingActions()
+                    - $report['pageMetrics']['loops'];
+            }
+        } catch (\Exception $e) {
+            Model::handleMaxExecutionTimeError(
```
tsteur: Is there a chance that if the query doesn't fail because of max execution error, the exception is not thrown?

tsteur:
Looks good. Left one comment. Haven't tested it though. Be good to make sure some error message will be shown in the UI if the query fails because of a timeout similar to Visitor Log. I suppose that's what happens automatically anyway. Just need to see how it behaves when it's used in transitions eg through a row action

plugins/Live/Model.php
sgiehl: that method itself is tested here
https://github.com/matomo-org/matomo/blob/148e049bc7665af470514a8f0a573973fc59a52b/plugins/Live/tests/Integration/ModelTest.php#L74-L126

plugins/Transitions/API.php
sgiehl: oh sure. there should be a `throw $e` after that. Will add...

sgiehl:
@tsteur it show the exception in the overlay if loading a transition through a row action:

image

core/DbHelper.php
sgiehl: added some simple tests

plugins/Live/Model.php
tsteur: Wouldn't this test need to be adjusted? Not seeing it in this PR? Would expect seeing this new parameter and calling it static @sgiehl ?

tsteur:
@sgiehl I just tested it and it doesn't work. That's because the ranking query will be applied and therefore it's not after the first `select`.
Not sure how it works for you? Cause they all have ranking query set?

This is my setting
> live_query_max_execution_time = 0.001

tsteur:
btw we could in worst case move this out of 3.13.4 into 3.13.5 if we wanted to release 3.13.4 earlier

sgiehl:
@tsteur updated the code so it works with the ranking query correctly.

PR #15664, opened 2020-03-04
Force using index index_idsite_idvisitor if available

### LINKED ISSUE #15588: MariaDB | Slow Query | High CPU Load
Hi there,
I have struggled with a problem that occurs at the peak of the day's highest load, causing a high CPU load on the database server.

My database server:
```
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                32
On-line CPU(s) list:   0-31
Thread(s) per core:    2
Core(s) per socket:    8
Socket(s):             2
NUMA node(s):          2
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 45
Model name:            Intel(R) Xeon(R) CPU E5-2660 0 @ 2.20GHz
Stepping:              7
CPU MHz:               2200.000
CPU max MHz:           2200.0000
CPU min MHz:           1200.0000
BogoMIPS:              4389.13
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              20480K
NUMA node0 CPU(s):     0-7,16-23
NUMA node1 CPU(s):     8-15,24-31
```
```
MemTotal:       65764440 kB
MemFree:         4493980 kB
MemAvailable:    4999644 kB
Buffers:              16 kB
Cached:           539544 kB
SwapCached:      2550824 kB
Active:         52374964 kB
Inactive:        7595748 kB
Active(anon):   51917680 kB
Inactive(anon):  7528420 kB
Active(file):     457284 kB
Inactive(file):    67328 kB
Unevictable:        8648 kB
Mlocked:            8648 kB
SwapTotal:      16777212 kB
SwapFree:       12554256 kB
Dirty:                52 kB
Writeback:             0 kB
AnonPages:      58823428 kB
Mapped:            30216 kB
Shmem:              8600 kB
Slab:             630112 kB
SReclaimable:     474008 kB
SUnreclaim:       156104 kB
KernelStack:       13600 kB
PageTables:       128872 kB
NFS_Unstable:          0 kB
Bounce:                0 kB
WritebackTmp:          0 kB
CommitLimit:    49659432 kB
Committed_AS:   58100260 kB
VmallocTotal:   34359738367 kB
VmallocUsed:      425568 kB
VmallocChunk:   34325395452 kB
HardwareCorrupted:     0 kB
AnonHugePages:  13486080 kB
CmaTotal:              0 kB
CmaFree:               0 kB
HugePages_Total:       0
HugePages_Free:        0
HugePages_Rsvd:        0
HugePages_Surp:        0
Hugepagesize:       2048 kB
DirectMap4k:      226988 kB
DirectMap2M:     8126464 kB
DirectMap1G:    58720256 kB
```

I have about 20 million page views per month, so, my dedicated database server are complaint with Matomo requirements. 
 
I'm using MariaDB 10.4.10 and this is my.cnf file

```
[mysqld]
datadir=/DB/mysql
socket=/DB/mysql/mysql.sock
tmpdir=/TEMPDB
bind-address=0.0.0.0
general_log_file=/var/log/mariadb/mariadb_acess.log
long_query_time = 10
slow_query_log
slow_query_log_file =/var/log/mariadb/slow-query.log
log_error=/var/log/mariadb/mariadb.err
innodb_flush_method=O_DIRECT
innodb_file_per_table=ON
innodb_flush_log_at_trx_commit = 2
innodb_log_group_home_dir = /TEMPDB

innodb_buffer_pool_size=51G
innodb_buffer_pool_instances=51
innodb_print_all_deadlocks=1
performance_schema=ON
max_allowed_packet=268435456
max_connections = 400
wait_timeout = 14400
interactive_timeout = 14400

[client]
port=3306
socket=/DB/mysql/mysql.sock

[mysqld_safe]
log-error=/var/log/mariadb/mariadb.log
pid-file=/var/run/mariadb/mariadb.pid
```

In the highest peak of the day i see that MySQL get allot slows querys similar to below:

```
# User@Host: matomo[matomo] @ server1 [xxx.xxx.xxx.xxx]
# Thread_id: 2001714  Schema: piwik  QC_hit: No
# Query_time: 16.680956  Lock_time: 0.000157  Rows_sent: 1  Rows_examined: 206832
# Rows_affected: 0  Bytes_sent: 5354
SET timestamp=1582018281;
SELECT visit_last_action_time, visit_first_action_time, idvisitor, idvisit, user_id, visit_exit_idaction_url, visit_exit_idaction_name, visitor_returning, visitor_days_since_first, visitor_days_since_order, visitor_count_visits, visit_goal_buyer, location_country, location_region, location_city, location_latitude, location_longitude, referer_name, referer_keyword, referer_type, idsite, visit_entry_idaction_url, visit_total_actions, visit_total_interactions, visit_total_searches, referer_url, config_device_brand, config_device_model, config_device_type, visit_total_events, visit_total_time, location_ip, location_browser_lang, campaign_content, campaign_id, campaign_keyword, campaign_medium, campaign_name, campaign_source, last_idlink_va, custom_dimension_1, custom_dimension_2, custom_dimension_3, custom_dimension_4, custom_dimension_5, custom_var_k1, custom_var_v1, custom_var_k2, custom_var_v2, custom_var_k3, custom_var_v3, custom_var_k4, custom_var_v4, custom_var_k5, custom_var_v5   FROM piwik_log_visit WHERE idsite = '23' AND visit_last_action_time <= '2020-02-18 10:01:04' AND idvisitor = 'S><FF><DF>J\n-<93>'
                ORDER BY visit_last
_action_time DESC
                LIMIT 1;
```

I don't have any problems running this query's directly, but i have a lot of them in the state "Sending Data" when i run the SHOW FULL PROCESSLIST; 

> ## High Load TOP ##
> top - 09:46:47 up 331 days,  6:56,  1 user,  load average: 42.14, 43.14, 42.75
> Tasks: 485 total,   3 running, 482 sleeping,   0 stopped,   0 zombie
> %Cpu(s): 90.6 us,  0.9 sy,  0.0 ni,  8.5 id,  0.0 wa,  0.0 hi,  0.1 si,  0.0 st
> KiB Mem : 65764440 total,  4386816 free, 60103412 used,  1274212 buff/cache
> KiB Swap: 16777212 total, 12578576 free,  4198636 used.  4996896 avail Mem

Hi attach the "innodb status" file. 

What i'm missing? 

[innodb-status.txt](https://github.com/matomo-org/matomo/files/4218567/innodb-status.txt)

Thanks!
TP

sgiehl:
fixes #15588

### DIFF
--- a/core/DbHelper.php
+++ b/core/DbHelper.php
@@ -178,6 +178,21 @@ class DbHelper
     }
 
     /**
+     * Returns if the given table has an index with the given name
+     *
+     * @param string $table
+     * @param string $indexName
+     *
+     * @return bool
+     * @throws Exception
+     */
+    public static function tableHasIndex($table, $indexName)
+    {
+        $result = Db::getReader()->fetchOne('SHOW INDEX FROM '.$table.' WHERE Key_name = ?', [$indexName]);
+        return !empty($result);
+    }
+
+    /**
      * Get the SQL to create Piwik tables
      *
      * @return array  array of strings containing SQL
--- a/core/DbHelper.php
+++ b/core/DbHelper.php
@@ -188,7 +188,7 @@ class DbHelper
      */
     public static function tableHasIndex($table, $indexName)
     {
-        $result = Db::getReader()->fetchOne('SHOW INDEX FROM '.$table.' WHERE Key_name = ?', [$indexName]);
+        $result = Db::get()->fetchOne('SHOW INDEX FROM '.$table.' WHERE Key_name = ?', [$indexName]);
         return !empty($result);
     }
 
--- a/core/Tracker/Model.php
+++ b/core/Tracker/Model.php
@@ -11,6 +11,7 @@ namespace Piwik\Tracker;
 use Exception;
 use Piwik\Common;
 use Piwik\Container\StaticContainer;
+use Piwik\DbHelper;
 use Piwik\Tracker;
 use Psr\Log\LoggerInterface;
 
@@ -418,7 +419,21 @@ class Model
 
     private function findVisitorByVisitorId($idVisitor, $select, $from, $where, $bindSql)
     {
-        // will use INDEX index_idsite_idvisitor (idsite, idvisitor)
+        $cacheKey = 'log_visit_has_index_idsite_idvisitor';
+        $cache = Cache::getCacheGeneral();
+
+        if (!array_key_exists($cacheKey, $cache)) {
+            $hasIndex = DbHelper::tableHasIndex(Common::prefixTable('log_visit'), 'index_idsite_idvisitor');
+            $cache[$cacheKey] = $hasIndex;
+
+            Cache::setCacheGeneral($cache);
+        }
+
+        // use INDEX index_idsite_idvisitor (idsite, idvisitor) if available
+        if (true === $cache[$cacheKey]) {
+            $from .= ' USE INDEX (index_idsite_idvisitor) ';
+        }
+
         $where .= ' AND idvisitor = ?';
         $bindSql[] = $idVisitor;
 
--- a/core/Tracker/Model.php
+++ b/core/Tracker/Model.php
@@ -11,12 +11,12 @@ namespace Piwik\Tracker;
 use Exception;
 use Piwik\Common;
 use Piwik\Container\StaticContainer;
-use Piwik\DbHelper;
 use Piwik\Tracker;
 use Psr\Log\LoggerInterface;
 
 class Model
 {
+    const CACHE_KEY_INDEX_IDSITE_IDVISITOR = 'log_visit_has_index_idsite_idvisitor';
 
     public function createAction($visitAction)
     {
@@ -419,19 +419,11 @@ class Model
 
     private function findVisitorByVisitorId($idVisitor, $select, $from, $where, $bindSql)
     {
-        $cacheKey = 'log_visit_has_index_idsite_idvisitor';
         $cache = Cache::getCacheGeneral();
 
-        if (!array_key_exists($cacheKey, $cache)) {
-            $hasIndex = DbHelper::tableHasIndex(Common::prefixTable('log_visit'), 'index_idsite_idvisitor');
-            $cache[$cacheKey] = $hasIndex;
-
-            Cache::setCacheGeneral($cache);
-        }
-
         // use INDEX index_idsite_idvisitor (idsite, idvisitor) if available
-        if (true === $cache[$cacheKey]) {
-            $from .= ' USE INDEX (index_idsite_idvisitor) ';
+        if (array_key_exists(self::CACHE_KEY_INDEX_IDSITE_IDVISITOR, $cache) && true === $cache[self::CACHE_KEY_INDEX_IDSITE_IDVISITOR]) {
+            $from .= ' FORCE INDEX (index_idsite_idvisitor) ';
         }
 
         $where .= ' AND idvisitor = ?';
--- a/plugins/CoreHome/CoreHome.php
+++ b/plugins/CoreHome/CoreHome.php
@@ -11,13 +11,16 @@ namespace Piwik\Plugins\CoreHome;
 use Piwik\Archive\ArchiveInvalidator;
 use Piwik\Columns\ComputedMetricFactory;
 use Piwik\Columns\MetricsList;
+use Piwik\Common;
 use Piwik\Container\StaticContainer;
+use Piwik\DbHelper;
 use Piwik\IP;
 use Piwik\Piwik;
 use Piwik\Plugin\ArchivedMetric;
 use Piwik\Plugin\ComputedMetric;
 use Piwik\Plugin\ThemeStyles;
 use Piwik\SettingsServer;
+use Piwik\Tracker\Model as TrackerModel;
 
 /**
  *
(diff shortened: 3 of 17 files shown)


core/DbHelper.php
```diff
@@ -177,6 +177,21 @@ public static function createDatabase($dbName = null)
...
+     *
+     * @param string $table
+     * @param string $indexName
+     *
+     * @return bool
+     * @throws Exception
+     */
+    public static function tableHasIndex($table, $indexName)
+    {
+        $result = Db::getReader()->fetchOne('SHOW INDEX FROM '.$table.' WHERE Key_name = ?', [$indexName]);
```
tsteur: Should probably use `Db::get()` here since the `log_visit` queries go to the master and not the slave. And some versions allow you to define different indexes on master vs reader (which is really cool btw since you need different indexes usually on master and reader)

core/Tracker/Model.php
```diff
@@ -418,7 +419,21 @@ public function findVisitor($idSite, $configId, $idVisitor, $userId, $fieldsToRe
 
     private function findVisitorByVisitorId($idVisitor, $select, $from, $where, $bindSql)
     {
-        // will use INDEX index_idsite_idvisitor (idsite, idvisitor)
+        $cacheKey = 'log_visit_has_index_idsite_idvisitor';
+        $cache = Cache::getCacheGeneral();
+
+        if (!array_key_exists($cacheKey, $cache)) {
```
tsteur: Can we put this into `CoreHome` plugin or some other plugin and enrich the index when  the data is gathered? Otherwise what happens is that `general cache` will be generated, then it will be regenerated here under circumstances and written twice. This becomes especially problematic when you actually want to sync the cache across servers like we do where we generate the cache every hour on one server and sync it to other servers. In this case, we would generate the cache without that field, sync it across all servers, then each server would regenerate the cache again.

Might be actually good to check if there are similar cases in the code for the cache general.

core/Tracker/Model.php
```diff
@@ -418,7 +419,21 @@ public function findVisitor($idSite, $configId, $idVisitor, $userId, $fieldsToRe
...
+            $hasIndex = DbHelper::tableHasIndex(Common::prefixTable('log_visit'), 'index_idsite_idvisitor');
+            $cache[$cacheKey] = $hasIndex;
+
+            Cache::setCacheGeneral($cache);
+        }
+
+        // use INDEX index_idsite_idvisitor (idsite, idvisitor) if available
+        if (true === $cache[$cacheKey]) {
+            $from .= ' USE INDEX (index_idsite_idvisitor) ';
```
tsteur: Might be good to use a constant or so for `index_idsite_idvisitor `

tsteur: wonder if we should use `FORCE INDEX` btw?

tsteur (changes requested):

core/Tracker/Model.php
sgiehl: ok. will change that. Searched for other places where the cache is written. This is the only one I found: https://github.com/matomo-org/matomo/blob/b16a791aa3650d85af829156129c2bd44c7cb075/core/Tracker/ScheduledTasksRunner.php#L65

But guess that can't be changed easily...

core/Tracker/Model.php
sgiehl: guess that makes sense, as it avoid a table scan if the index can be used...

sgiehl:
@tsteur applied the feedback

tsteur:
tested it locally and worked. I suppose these failing tests are not related to this? https://travis-ci.org/matomo-org/matomo/jobs/658576281#L1228

core/DbHelper.php
```diff
@@ -177,6 +177,21 @@ public static function createDatabase($dbName = null)
...
+     *
+     * @param string $table
+     * @param string $indexName
+     *
+     * @return bool
+     * @throws Exception
+     */
+    public static function tableHasIndex($table, $indexName)
+    {
+        $result = Db::get()->fetchOne('SHOW INDEX FROM '.$table.' WHERE Key_name = ?', [$indexName]);
```
tsteur: was thinking could maybe catch an exception here but this would just hide errors I suppose. So be good to keep it for now like this I reckon

tsteur:
feel free to merge if the test failures are unrelated

sgiehl:
Seems the test failures are related, as they are not failing on 3.x-dev. Will investigate that now

sgiehl:
@tsteur I looked into the test failures now for a while. It might be a case that only appears in that test scenario, as all visits have the same visitorId and some even have the same time. Fetching the visitor then returns different results as forcing the index causes the result being sorted by the index order (for same visitorId and datetime). Not using the index sorts those entries by the order they were stored in the table. So basically that "change" only occurs if there are multiple entries for the same visitorid and datetime, which likely shouldn't occur ever I guess 🤔 

I've adjusted the failing test now, so it uses different visitor ids and time. That way it works the same with or without the index...

tsteur:
Not sure I understand @sgiehl . In https://github.com/matomo-org/matomo/pull/15664/files#diff-c27b9aad61da66057fe4abcbdc1c9f67R3 there is quite a big difference in the number of unique visitors for example. This wouldn't be quite expected if it was just a sorting issue?

Seems very tricky. I can imagine as you say that it picks a different visit now when they have same visitorId and datetime, but the number of uniq visitors to be so different is maybe bit unexpected? Not quite sure how to explain this. The other ones can be indeed because if this and this should usually not happen or at least not cause any issue. Thought could sort by `visit_last_action_time, idvisit` but this would be overkill for something that usually doesn't happen and shouldn't do it.

Do you have a thought how the nb_uniq_visitors change can be explained?

sgiehl:
@tsteur that's because the fixture is changed to force new visitorIds and datetimes to avoid having duplicate entries for those at all. The tests would now work with the forced index and without

tsteur:
There are still some tests failing @sgiehl LGTM otherwise

sgiehl:
tests should be fixed now

PR #15666, opened 2020-03-05
Try to prevent a deadlock when writing options

tsteur:
refs https://github.com/matomo-org/matomo/issues/15545#issuecomment-594927218

Invalidator tests passed for me locally. Not sure I missed something maybe?

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -182,7 +182,7 @@ class ArchiveInvalidator
 
     public function forgetRememberedArchivedReportsToInvalidateForSite($idSite)
     {
-        $id = '%_' . $this->buildRememberArchivedReportIdForSite($idSite) . '_%';
+        $id = $this->buildRememberArchivedReportIdForSite($idSite);
         $this->deleteOptionLike($id);
         Cache::clearCacheGeneral();
     }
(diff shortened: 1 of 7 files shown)


core/Archive/ArchiveInvalidator.php
```diff
@@ -78,11 +78,12 @@ public function getAllRememberToInvalidateArchivedReportsLater()
         // we do not really have to get the value first. we could simply always try to call set() and it would update or
         // insert the record if needed but we do not want to lock the table (especially since there are still some
         // MyISAM installations)
-        $values = Option::getLike($this->rememberArchivedReportIdStart . '%');
+        $values = Option::getLike('%' . $this->rememberArchivedReportIdStart . '%');
```
diosmosis: Sometimes I see `'%_' .` and other times `'%' .`, could that be an issue?

tsteur: @diosmosis not really. I just copied behaviour what was done on the right part, to have the same on the left. It'll not make a difference though

diosmosis:
I don't see a change to the `Option::getLike()` in `deleteOptionLike()`, should there be one?

tsteur:
@diosmosis yes there should! Will add it now. I used to have it there but must have been undone by some other reverts

tsteur:
@diosmosis actually it works because it is already partially sent with the `%` when the method is called. I will make that more clear. It's confusing right now

diosmosis:
looks good to merge if tests pass 👍

PR #15670, opened 2020-03-05
activate warning for php versions incompatible with next major release

### LINKED ISSUE #15623: Mention require PHP 7.2 in admin notification
I think we haven't yet enabled this admin notification. The code should be already there but likely disabled. Basically, the goal is, if someone is using PHP 7.1 or older that they should upgrade to PHP 7.2 at least or better even directly PHP 7.4.

sgiehl:
fixes #15623

### DIFF
--- a/core/Plugin/ControllerAdmin.php
+++ b/core/Plugin/ControllerAdmin.php
@@ -207,12 +207,12 @@ abstract class ControllerAdmin extends Controller
     }
 
     /**
-     * PHP Version required by the next major Piwik version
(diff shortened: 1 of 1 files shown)


tsteur:
👍 works. 

Ideally it would show `(PHP 7.3 or newer recommended)` as part of the text as well maybe if easily done? Not too important though. I suppose it would maybe invalidate some translations and we're not updating 3.x translations anymore so not needed.

sgiehl:
Correct. That would mean changing translations

PR #15671, opened 2020-03-05
try to update nonce in out out frame when using new window

### LINKED ISSUE #15629: Using Opt-Out iFrame twice doesn't work
Reported in https://forum.matomo.org/t/fehler-im-optout/36246 and reproduced on demo.matomo.org

Steps to reproduce:
- open https://demo.matomo.org/index.php?module=CoreAdminHome&action=optOut&language=de
- opt out
- don't reload the page and opt-in again
- reload the page
Expected: Tracking is enabled again
Result: Tracking is still disabled

This also works the other way around (having opt-out, enabling and then re-disabling tracking and it won't opt-out correctly)

sgiehl:
fixes #15629

### DIFF
--- a/plugins/CoreAdminHome/templates/optOut.twig
+++ b/plugins/CoreAdminHome/templates/optOut.twig
@@ -33,7 +33,14 @@
      #}
     {% if showConfirmOnly %}
     <p>{{ 'CoreAdminHome_OptingYouOut'|translate }}</p>
-    <script>window.close();</script>
+    <script>
+        {# try to update nonce in iframe, so sending it a second time works #}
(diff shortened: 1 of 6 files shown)


diosmosis:
Maybe we could add this test to the optoutform UI test? I think we could simply add a page reload after one test to check the setting is retained (the `should correctly show the checkbox unchecked after reloading after opting-out` test).

diosmosis:
Tested on chrome, firefox & safari. Works for me. Should probably test on IE.

sgiehl:
@diosmosis added another UI test

Issue #14764, opened 2019-08-13
Page Overlay stuck on "Loading..."


Findus23:
very similar to https://github.com/matomo-org/matomo/issues/11768
not like https://github.com/matomo-org/matomo/issues/12815 (no error or failed request)

Not sure since when, but the page overlay doesn't really seem to work anymore.
It only shows the loading screen but doesn't actually do anything:
![image](https://user-images.githubusercontent.com/6266037/62937372-b50cbe00-bdcc-11e9-91d7-a21c56a0ae06.png)
There are no errors in the console, no failed HTTP requests and not even any HTTP requests apart from the assets of the page (so no calls to the API)

This can also be reproduced [on demo.matomo.org](https://demo.matomo.org/index.php?module=Overlay&period=month&date=today&idSite=62#?l=https$3A$2F$2Fdivezone.net$2Fjobs$3F__enterprise_demo$3Ddemo.matomo.org$26__enterprise_demo_idsite$3D62)

and has been reported on the forum:
https://forum.matomo.org/t/page-overlay-fehler/33851

peterbo:
Same here on 3 different instances (MTM 3.11)

tsteur:
This should have been fixed in https://github.com/matomo-org/matomo/pull/14766

peterbo:
Fix works, thanks, Thomas!

tiptronic85:
I had the same issue, i upgraded to 3.13.1 version, and now i'm getting logged out.
Anyone with this problem?

polarbirke:
I'm seeing the exact same failing behaviour with 3.13.2 in latest Firefox and latest Chrome for several of our websites:

- loading spinner shown, but nothing happens
- no errors in the console
- no failed HTTP requests

I've seen this happen before (August 2019) and I know the fix worked, because the Page Overlay was working fine in November 2019. Could this be another regression?

chrisprog20:
I also want to report that in 3.13.3 the Page Overlay does not work anymore (tested in Chrome, Firefox and Edge browser).
Also: If I want to login with the App "Internet Explorer 11" of Windows 10 I fail to get logged in with my admin user. I then get the message "Error: Your session has expired due to inactivity. Please log in to continue." The login with other browsers works.

chrisprog20:
@polarbirke: Maybe you try the newest version 3.13.3. There the page overlay works - even in Firefox.
[...]

PR #15673, opened 2020-03-05
Always send a referrer-policy header

polarbirke:
Suggested fix / improvement for #14764

### DIFF
--- a/core/View.php
+++ b/core/View.php
@@ -287,6 +287,9 @@ class View implements ViewInterface
         // don't send Referer-Header for outgoing links
         if (!empty($this->useStrictReferrerPolicy)) {
             Common::sendHeader('Referrer-Policy: same-origin');
+        } else {
+            // always send explicit default header
+            Common::sendHeader('Referrer-Policy: no-referrer-when-downgrade');
         }
 
         return $this->renderTwigTemplate();


core/View.php
```diff
@@ -287,6 +287,9 @@ public function render()
         // don't send Referer-Header for outgoing links
         if (!empty($this->useStrictReferrerPolicy)) {
             Common::sendHeader('Referrer-Policy: same-origin');
+        } else {
+            // always send explicit default header
+            Common::sendHeader('Referrer-Policy: no-referrer-when-downgrade');
```
tsteur: @Findus23 do you have maybe some thoughts on this?

Findus23: `no-referrer-when-downgrade` is the default when no header is sent, so sending it when we normally wouldn't send anything shouldn't change anything.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy

tsteur: thanks @Findus23 be good if someone could still test the PR and have a look at tests. Not sure if I find time today.

polarbirke: Aye it's the (implicit) default; this PR caters to our setup where we set a strict policy if nothing is explicitly set by the application (see [this comment](https://github.com/matomo-org/matomo/issues/14764#issuecomment-595180297)). Just in case you want to cater for our way of thinking (not sure if we're the outlier here or if more people might want to set their policies with `setifempty`).

diosmosis:
Tests look fine and matomo seems to work fine locally for me, will merge.

PR #15676, opened 2020-03-05
Make sure to set secure flag when setting samesite none

### LINKED ISSUE #15672: There is no way to enable 'secure' for the 3rd party cookie
A colleague just informed me that it is currently impossible to enable the "secure" flag for the 3rd party cookie.
This triggers a warning in browsers, because it is set to SameSite=None but not secure.

tsteur:
refs https://github.com/matomo-org/matomo/issues/15672 
this PR will also need to be merged to fix #15672 https://github.com/matomo-org/tag-manager/pull/229

@MichaelHeerklotz can you maybe also have a look at this PR

Not sure how to fallback when HTTPS is not used. Should we still set to None and assume it works for a while in some browsers that are not Chrome but eventually it'll fail? I suppose it's not quite a solution. I assume `Lax` maybe won't work so well for 3rd party tracking?

### DIFF
--- a/core/Tracker/Request.php
+++ b/core/Tracker/Request.php
@@ -20,6 +20,7 @@ use Piwik\Network\IPUtils;
 use Piwik\Piwik;
 use Piwik\Plugins\CustomVariables\CustomVariables;
 use Piwik\Plugins\UsersManager\UsersManager;
+use Piwik\ProxyHttp;
 use Piwik\Tracker;
 use Piwik\Cache as PiwikCache;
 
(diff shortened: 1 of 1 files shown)


sgiehl (approved):

diosmosis:
`Lax` doesn't work across domains so in this case we'd have to require https is used?

tsteur:
It would still work maybe if they use 3rd party cookies on same domains I suppose.

diosmosis:
👍 I guess it doesn't hurt to have it then. We may have to document the requirement to put matomo on https for some people then... Actually we shouldn't encourage using http. But that's not related to this issue, really.

tsteur:
We're recommending this here btw: https://matomo.org/blog/2020/02/new-cookie-behaviour-in-browsers-may-cause-regressions/

PR #15677, opened 2020-03-06
Remove director detection

pebosi:
Discussion on #15627 was to decide to remove detection of some plugin detections.


tsteur:
fyi @pebosi same here as in the Gears we would also use a `0` in the fingerprint where director used to be used.

tsteur:
Hi @pebosi are you still working on this?

pebosi:
Will look into this, later this week

sgiehl:
Will finish that in a new PR as well...

PR #15683, opened 2020-03-07
fix #15681 secure cookie flag is not set for referrer and custom dimensions

### LINKED ISSUE #15681: Check that all Matomo cookies are set with the secure flag
Found out _pk_ref-cookie is not secure, despite setSecureCookies is set. All the other _pk-cookies however are.
A quick look into the javascript code the check for a secure cookie is missing on some other cookies, e. g. CustomDimension, too.
Maybe the check if a cookie needs the secure flag can be moved to the setCookie-function instead of doing it individually for every single cookie.

tsteur:
Suppose the minified version should now be generated automatically through github action.

fix https://github.com/matomo-org/matomo/issues/15681


sgiehl:
@tsteur  it's not fully automated yet. You need to post a comment with the content "build js" only, to trigger a build:

Issue #10257, opened 2016-06-26
Feature request: automated install when the required config values are present


dessant:
If `config.ini.php` is present, the installation sequence is skipped entirely.

When the required fields are present in the config file (database host, name, password etc), the installation could happen in a headless way, otherwise the manual istallation page would be loaded on the first visit, and the entered data would be stored in config *without* overwiting the existing contents of the config file.

This would make it easy to provide a custom config during provisioning and fully set up Piwik, without the need for hacks like these: https://github.com/MarvAmBass/docker-piwik/blob/master/startup-piwik.sh.

I am aware of headless installation being available in the Pro version, please note that this issue was opened for the free version of Piwik.

tsteur:
The installation is skipped indeed if all needed fields are there in the config and if database tables exist. I'm not quite sure what the exact problem is? Did you want to install Piwik via command line by providing parameters like "--database-host"?

dessant:
Yeah, the main issue is that headless installation is not supported and one has to use the ui to provide the db connection and bootstrap the database. While this is ok for one-off installations, it becomes problematic when you're doing automated provisioning.

The config way was one idea, ideally the main outcome would be to have a way to install Piwik up to a state where the admin user is created, without requiring manual steps.

tsteur:
got it. I thought I created a similar issue a few years ago when I built a vagrant box for Piwik but cannot find it anymore. Possibly I didn't actually create the issue. It would be a useful feature to have nowadays. 

Basically we would only need to create the config file from the input and validate the entered data. For validation and creation of config file everything is already here so it shouldn't be too complex

mattab:
Hi @dessant 

we may in the future release such a tool as open source. If your company can support us in this area, please get in touch via https://piwik.org/contact/

dessant:
Thanks for considering a possible release, I'll definitely do it when I will have a company or I will work for one. :yum:

Toub:
Let's say the automation can be run with a script named install-piwik.sh, we need 2 different ways to provide the configuration:
- command lines args (e.g. install-piwik.sh --dbuser=myuser --dbpassword=pass ...) => drawback: the tool appears in the command history
- install.ini file (running install-piwik.sh will then remove install.ini, so it can be more secure)

tsteur:
If someone is keen to work on this I'm happy to assist. Be good to have this

nebev:
I had a stab at this yesterday. It's not particularly brilliant, but gets the job done for me.
[...]

PR #15691, opened 2020-03-12
MatomoInstaller : headless installation of matomo

AbcSxyZ:
Diagnostic are runned at the beginning of the installation, and could be a better way to use specific function used for this purpose.

### DIFF
--- a/core/MatomoInstaller.php
+++ b/core/MatomoInstaller.php
@@ -1,0 +1,416 @@
+<?php
+
+namespace Piwik;
+
+use Piwik\Access;
+use Piwik\Common;
+use Piwik\Config;
+use Piwik\Container\StaticContainer;
+use Piwik\Db;
+use Piwik\Db\Adapter;
+use Piwik\DbHelper;
+use Piwik\Plugins\SitesManager\API as APISitesManager;
+use Piwik\Plugins\UsersManager\API as APIUsersManager;
+use Piwik\Plugins\UsersManager\NewsletterSignup;
+use Piwik\ProxyHeaders;
+use Piwik\Updater;
+use Piwik\Url;
+use Piwik\Version;
+use Zend_Db_Adapter_Exception;
+
+//Warning: - Don't know how plugin are installed (not done by installer
+//           actually). Some magics when I created installer.
+//
+// require : https://github.com/matomo-org/matomo/pull/15691
+
+/*
+ * Manage the Matomo server installation. Contain
+ * all basic function to setup database, user, and
+ * site to track.
+ *
+ * (?Could be a parent class to create custom
+ * installation process?)
+*/
+
+class MatomoInstaller{
+    /*
+     * Perform the entire installation of Matomo.
+     * Use user-defined settings.
+     *
+     * @settings is expected to be an array with following
+     * variables.
+     *
+     *  - dbname
+     *  - port (optionnal)
+     *  - dbusername
+     *  - dbpassword
+     *  - dbhost (optionnal?)
+     *  - tables_prefix (optionnal)
+     *  - adapter (optionnal)
+     *  - type (optionnal)
+     *  - adminusername
+     *  - adminpassword
+     *  - email
+     *  - subscribe_newsletter_piwikorg (optionnal)
+     *  - subscribe_newsletter_professionalservices (optionnal)
+     *  - name
+     *  - url
+     *  - ecommerce : (optionnal)
+    */
+    public static function headlessInstall($settings)
+    {
+        if (self::systemCheck())
+        {
+            //Raise Error because DigntosicService report in invalid.
+            //What should I do in case of warning or error ?
+        }
+        
+        self::createDatabaseObject($settings);
+        self::tablesCreation();
+        self::setupSuperUser($settings);
+        self::firstWebsiteSetup($settings);
+    }
+
+    /*
+     * Check if an headless installation is expected.
+     * (Heritance should be related to it)
+     */
+    public static function isHeadlessInstall()
+    {
+        $installSettings = Config::getInstance()->MatomoInstall;
+        return !empty($installSettings);
+    }
+
+    /*
+     * Use settings from MatomoInstall section in
+     * the config.ini.php file and perform installation.
+     */
+    public static function installFromConfig()
+    {
+        //Use settings defined in config.ini.php
+        //in MatomoInstall section.
+        $settings = Config::getInstance()->MatomoInstall;
+        self::headlessInstall($settings);
+
+        //Remove installation config when finished
+        Config::getInstance()->MatomoInstall = array();
+        Config::getInstance()->forceSave();
+
+        //Redirect to matomo homepage
+        Url::redirectToUrl('index.php');
+    }
+
+    /**
+     * Installation Step 2: System Check
+     *
+     * Control if the system is ready for installation,
+     * run Plugins\Diagnostics to get state.
+     * 
+     * @view View (optionnal), view to store DiagnosticReport
+    */
+    //S: Should I perform the system check ?
+    //In : https://github.com/digitalist-se/extratools/blob/master/Lib/Install.php
+    // no system check seem to be performed.
+    public static function systemCheck($view=null)
+    {
+        //Why could delete conf be useful ?? Should I handle
+        //already created config ? + delete conf on failure.
+        // ! Having an error if I remove it.
+        // Multiple entrence for installation, must know where plug
+        // the install.
+        self::deleteConfigFileIfNeeded();
+        // Do not use dependency injection because this service requires a lot of sub-services across plugins
+        /** @var DiagnosticService $diagnosticService */
+        $diagnosticService = StaticContainer::get('Piwik\Plugins\Diagnostics\DiagnosticService');
+        $diagnosticReport = $diagnosticService->runDiagnostics();
+
+        //Store the diagnostic report in the view
+        // !! Should remove Installation plugin mecanism
+        //Must think about the way to handle error here
+        if (!is_null($view))
+        {
+            $view->diagnosticReport = $diagnosticReport;
+            $view->showNextStep = !$diagnosticReport->hasErrors();
+
+            // On the system check page, if all is green, display Next link at the top
+            $view->showNextStepAtTop = $view->showNextStep && !$diagnosticReport->hasWarnings();
+        }
+        //HEADLESS: DO STUFF WITH errors
+        return (!$diagnosticReport->hasErrors());
+    }
+
+    /**
+     * Installation Step 3: Database Set-up
+     * @throws Exception|Zend_Db_Adapter_Exception
+     */
+    public static function createDatabaseObject($settings)
+    {
+        if (empty($settings['dbname']))
+        {
+            throw new Exception("No database name");
+        }
+
+        //Set some default value for database setup.
+        if (is_null($settings['type']))
+            $settings['type'] = Config::getInstance()->database['type'];
+
+        //!! MUST ADD Adapter::getRecommendedApater
+        if (is_null($settings['adapter']))
+            $settings['adapter'] = Adapter::getRecommendedApater();
+
+        if (is_null($settings['port']))
+        {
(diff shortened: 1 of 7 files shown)


AbcSxyZ:
I made a mistake, it was only for commit 8f4fb4a initially. But not so bad, things are related.
I didn't thought pushing to my repo would also add new commit to the PR.

I worked on a feature : https://github.com/matomo-org/matomo/issues/10257 , about Matomo installation in a automated way.

It's for me the beginning of a feature who must be review and adjusted, I wasn't able to manage all. I tried to create it as a depedent block, I juste add it to Matomo in a temporary way (5bb4565) for testing.

Checkout issue, I added some comments. I know it's not finished, I think it can be a good point to start with, I tried to do as most as possible for me (I never used Matomo, don't know some cases). It can be for a new branch.

tsteur:
@AbcSxyZ how would someone trigger the headless installation?

core/MatomoInstaller.php
```diff
@@ -0,0 +1,416 @@
...
+     *  - subscribe_newsletter_professionalservices (optionnal)
+     *  - name
+     *  - url
+     *  - ecommerce : (optionnal)
+    */
+    public static function headlessInstall($settings)
+    {
+        if (self::systemCheck())
+        {
+            //Raise Error because DigntosicService report in invalid.
```
tsteur, line 64: seems there is something to do?

core/MatomoInstaller.php
```diff
@@ -0,0 +1,416 @@
...
+     *  - adminusername
+     *  - adminpassword
+     *  - email
+     *  - subscribe_newsletter_piwikorg (optionnal)
+     *  - subscribe_newsletter_professionalservices (optionnal)
+     *  - name
+     *  - url
+     *  - ecommerce : (optionnal)
+    */
+    public static function headlessInstall($settings)
```
tsteur, line 60: btw ideally we would put these things maybe in a command so someone can specifically trigger the headless installation? We could create a command class and then add this command to https://github.com/matomo-org/matomo/blob/4.x-dev/core/Console.php#L270 so it's available even before any Matomo is installed . Or alternatively such a command be defined in a plugin and then we'd need to configure that plugin name in `[General]always_load_commands_from_plugin` setting.

AbcSxyZ:
> @AbcSxyZ how would someone trigger the headless installation?  

See https://github.com/matomo-org/matomo/issues/10257#issuecomment-598262173 for some explaination. It's using a MatomoInstall section in the config file.

>  seems there is something to do?

Like I said, I'm not a user of matomo, so there is multiple point that I don't know how to do. I commented out those point and it's necessary to dive into the code again with someone who know matomo.
  
> btw ideally we would put these things maybe in a command so someone can specifically trigger the headless installation?

Think only about [MatomoInstaller.php](https://github.com/matomo-org/matomo/blob/5bb4565136fa3b06d18386b94d9ccf3be67a6f54/core/MatomoInstaller.php) file, other files of the PR is for testing, to correct some small stuff incompatible with the installer etc.
I was more thinking about it as a core feature, who can be used for any installation (and could be an helper for the actual one).

You can see the difference because between [`MatomoInstaller::headlessInstall`](https://github.com/matomo-org/matomo/blob/5bb4565136fa3b06d18386b94d9ccf3be67a6f54/core/MatomoInstaller.php#L60) and [`Matomo::installFromConfig`](https://github.com/matomo-org/matomo/blob/5bb4565136fa3b06d18386b94d9ccf3be67a6f54/core/MatomoInstaller.php#L88) have their own purpose. `headlessInstall` function just get all parameters, and perform the installation. `installFromConfig` retrieve the settings inside the MatomoInstaller section of the config file, and give them to the `headlessInstall`.

So what I had in mind is to allow the user to perform the installation in any way he want. He just have to define his own way to retrieve data, and give them to the `MatomoInstaller::headlessInstall`. In this way you [Matomo devs] could create some default way to perform the installation (through config file for exemple) and allow the user to create custom method.

AbcSxyZ:
Because there is a lot of edited files (7 files), I will add explaination about all of them, to detail what are purpose of the modification, hope it will be usefull.

1. commit [8f4fb4afb1232d588d0615a794e6c313b86ed0ff](https://github.com/matomo-org/matomo/pull/15691/commits/8f4fb4afb1232d588d0615a794e6c313b86ed0ff), files ` plugins/Diagnostics/Diagnostic/{DbReaderCheck, ForceSSLCheck, NfsDiskCheck}.php` (3/7)  
**-->** To launch diagnostic, it actually verify if matomo is installed by checking the local config file. Change it by using function `SettingsPiwik::isMatomoInstalled` designed for it, otherwise it create some conflict with a config file containing an installation section.

2. commit [a391f00436c8147bed254d0397438efa8dd2b164](https://github.com/matomo-org/matomo/pull/15691/commits/a391f00436c8147bed254d0397438efa8dd2b164)
    - file `core/Db/Adapter.php` (4/7)
    **-->**  Create a function to get the default adapter instead of hardcoding it in the installer.
    - file ` core/MatomoInstaller.php` (5/7) **[THE MAIN FEATURE]**
    **-->** Class to perform an entire installation. Actually divided by step like it is with the actual GUI (system check, databases installation, site creation ...). Mainly 2 extra function `MatomoInstaller::headlessInstall` and `MatomoInstaller::installFromConfig`, respectively used to perform all step from given settings and to retrieve settings within the config file.

3. commit [5bb4565136fa3b06d18386b94d9ccf3be67a6f54](https://github.com/matomo-org/matomo/pull/15691/commits/5bb4565136fa3b06d18386b94d9ccf3be67a6f54)
    - file `core/SettingsPiwik.php` (6/7)
    **-->** For `SettingsPiwik::isMatomoInstalled`, check if an headless install have to be performed.
    - file `core/Application/Kernel/EnvironmentValidator.php` (7/7)
    **-->** Perform the `MatomoInstaller::installFromConfig`. Probably not the best place to perform this call, but this allowed my to test the installer. From my memory, I was a bit lost because the environnement is called from multiple place, and it was working when called from here. It's like glue.

I hope all comments and this will help to understand my work. Remember, I'm nor a php dev, nor a matomo user so there is multiple unknow stuff to me who could (should ?) be improve. Also in multiple place, I didn't know how to deal with some errors (it should be commented within the code). Finally, to avoid duplicating code, the actual GUI installation could call function from the installer because each installation step is a function of the `MatomoInstaller`. Because it's important feature, I didn't add those modification to the PR, to let the headless installation being a standalone feature, but those modification should be performed at short or long term.

AbcSxyZ:
As an important extra, inside the `MatomoInstaller`, or the installer himself, it have to be rethink to allow inheritance of a couple of functions, at least `MatomoInstall::isHeadlessInstall`, to let the user adding custom installation methods.

tsteur:
@sgiehl maybe you can have a look at this PR tomorrow?

core/MatomoInstaller.php
```diff
@@ -0,0 +1,416 @@
...
+                /* // select the newly created database */
+                @Db::createDatabaseObject($dbInfos);
+            } else {
+                throw $e;
+            }
+        }
+
+        @DbHelper::checkDatabaseVersion();
+
+        @Db::get()->checkClientVersion();
```
sgiehl, line 217: You should try to avoid using `@` so often. If something might throw an Error\Exception us try/catch instead.

core/MatomoInstaller.php
```diff
@@ -0,0 +1,416 @@
...
+/*
+ * Manage the Matomo server installation. Contain
+ * all basic function to setup database, user, and
+ * site to track.
+ *
+ * (?Could be a parent class to create custom
+ * installation process?)
+*/
+
+class MatomoInstaller{
```
sgiehl, line 35:
```suggestion
class MatomoInstaller {
```

core/MatomoInstaller.php
```diff
@@ -0,0 +1,416 @@
...
+        //HEADLESS: DO STUFF WITH errors
+        return (!$diagnosticReport->hasErrors());
+    }
+
+    /**
+     * Installation Step 3: Database Set-up
+     * @throws Exception|Zend_Db_Adapter_Exception
+     */
+    public static function createDatabaseObject($settings)
+    {
```
sgiehl, line 147: Wondering if it would be better to directly configure the database in the database section instead of the new `Installer` section. The Installer could simply check if the database is configured correctly and if no tables exist yet, the installer could create them...

core/MatomoInstaller.php
```diff
@@ -0,0 +1,416 @@
...
+
+        self::updateComponents();
+
+        Updater::recordComponentSuccessfullyUpdated('core', Version::VERSION);
+    }
+
+    /**
+     * Installation Step 5: General Set-up (superuser login/password/email and subscriptions)
+     */
+    public static function setupSuperUser($settings)
```
sgiehl, line 240: Any reason for having all methods static? Imho it would be better to have them non-static. Additionally I would set the settings as private property in `__construct` maybe and use that in all methods instead of passing it to every method

core/MatomoInstaller.php
```diff
@@ -0,0 +1,416 @@
...
+     *
+     * Control if the system is ready for installation,
+     * run Plugins\Diagnostics to get state.
+     * 
+     * @view View (optionnal), view to store DiagnosticReport
+    */
+    //S: Should I perform the system check ?
+    //In : https://github.com/digitalist-se/extratools/blob/master/Lib/Install.php
+    // no system check seem to be performed.
+    public static function systemCheck($view=null)
```
sgiehl, line 114: When moved to a command, we could output the error results and maybe ask the user if the installation should be aborted or continued nevertheless.

sgiehl (changes requested):
Left a few comments, some of them only re coding style.

In general I agree with @tsteur. It might be more useful to have that in a command instead of having a magic process that is triggered automatically once Matomo is opened.

The command could work with a given ini file (doesn't need to be Matomos config), or with command line parameters or if anything is missing the command could even ask for a value...

AbcSxyZ:
@sgiehl I don't know what is the best way to perform installation (command or ini file), the idea was also to let any user create a custom install method. So you can  choose one default method to allow an headless install, and let the user specify if needed.

So I don't know which method would be the best. But in my mind, you can't create a new installation method without (at short or long term) modifying the installation process inside the core. You GUI installation and command installation should call the same interface, otherwise you will have duplicate code.

core/MatomoInstaller.php
AbcSxyZ: I'm not a php developper, and I had some problem with the a php feature where you store initialized class, so I made it like that. Everything can be improve.

core/MatomoInstaller.php
AbcSxyZ: It was like that in the previous version. It seems installation process change since I did it (https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/Installation/Controller.php)

bluikko:
The "ExtraTools" plug-in has support for headless install: https://github.com/digitalist-se/extratools

They use `console matomo:install` to install. It would be great if Matomo could provide headless install out of the box.

sgiehl:
@AbcSxyZ are you going to update this PR any further or is the implementation in the ExtraTools plugin enough for you?

AbcSxyZ:
I won't do it, I just used Matomo once, I was more trying to contribute to an open source project by the time. It took too long to get an answer, I was expecting some help, I'm not involved anymore. Some disappointment honestly.

 So I don't care if this PR is deleted or whatever, but if I were a matomo user I would prefer to have a headless installation rather than a plugin to perform that !
 
 >  It would be great if Matomo could provide headless install out of the box.
 @bluikko 
 
 And Matomo know this since years :)
 Wish you the best for those who will work on that, will be a huge improvement from my point of view.

sgiehl:
@AbcSxyZ  Sorry to hear we weren't able to help you enough to get this PR finished.
Thanks anyway for your effort!
I will close this PR for now, as we won't be working on it soon. If anyone else is keen on implementing this, feel free to use some parts of the changes in this PR to reopen a new one.
Everything else about headless installed will be handled in the issue #10257

PR #15694, opened 2020-03-15
DB-IP databases are free and can be downloaded automatically, but geolocation city results are not as accurate as MaxMind's

### LINKED ISSUE #15589: [Geo] Unknown region metrics increased significantly after update to v3.13.1

diosmosis:
Fixes #15589

### DIFF
--- a/plugins/UserCountry/lang/en.json
+++ b/plugins/UserCountry/lang/en.json
@@ -100,6 +100,7 @@
         "UpdaterWillRunNext": "It is next scheduled to run on %s.",
         "WidgetLocation": "Visitor Location",
         "InstallGeoIp2": "Matomo recommends using the %1$sdbip%2$s databases, but this requires using the GeoIp2 plugin. If you install and activate it (which will require a possibly long running update), you will be able to get started using %1$sdbip%2$s databases. (Ignore this if you are using a third party plugin that provides it's own geolocation functionality.)",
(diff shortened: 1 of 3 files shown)


tsteur:
As mentioned be good if we could explain to users how users can get a MaxMind DB URL by linking to https://matomo.org/faq/how-to/how-do-i-get-the-geolocation-download-url-for-the-free-maxmind-db/  somehow. Not sure if we should also show the message when DB is already configured?

PR #15695, opened 2020-03-16
Update tag manager reference which removes duplicate index

tsteur:
need to wait for tests to pass


sgiehl:
failing tests seem to be unrelated

PR #15701, opened 2020-03-17
Revert "Process unique visitor metric for Roll-ups (set enable_processing_unique_visitors_multiple_sites = 1)"

tsteur:
Reverts matomo-org/matomo#15450

The previous PR will only work when `enable_fingerprinting_across_websites` is also enabled. This is a privacy issue though. As a result archiving will break with roll up reporting etc. I reckon this PR should therefore be reverted as it broke RollUpReporting and fixing that would cause privacy issues.


diosmosis:
Would it also help to emit some kind of error when config has mismatched values?

tsteur:
It does during archiving should a roll up be configured.

diosmosis:
FYI @mattab, merged this

PR #15706, opened 2020-03-18
Avoid update problem if token_auth table does not yet exist

sgiehl:
Regression from #15410

### DIFF
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -402,7 +402,14 @@ class FrontController extends Singleton
         // ... if session auth fails try normal auth (which will login the anonymous user)
         if (!$loggedIn) {
             $authAdapter = $this->makeAuthenticator();
-            Access::getInstance()->reloadAccess($authAdapter);
+            try {
+                Access::getInstance()->reloadAccess($authAdapter);
+            } catch (\Zend_Db_Statement_Exception $e) {
+                // table `user_token_auth` might not yet exist when updating to Matomo 4
+                if (!strpos($e->getMessage(), 'user_token_auth')) {
+                    throw $e;
+                }
+            }
         } else {
             $this->makeAuthenticator($sessionAuth); // Piwik\Auth must be set to the correct Login plugin
         }
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -402,14 +402,7 @@ class FrontController extends Singleton
         // ... if session auth fails try normal auth (which will login the anonymous user)
         if (!$loggedIn) {
             $authAdapter = $this->makeAuthenticator();
-            try {
-                Access::getInstance()->reloadAccess($authAdapter);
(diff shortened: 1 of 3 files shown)


core/FrontController.php
```diff
@@ -402,7 +402,14 @@ public function init()
         // ... if session auth fails try normal auth (which will login the anonymous user)
         if (!$loggedIn) {
             $authAdapter = $this->makeAuthenticator();
-            Access::getInstance()->reloadAccess($authAdapter);
+            try {
+                Access::getInstance()->reloadAccess($authAdapter);
+            } catch (\Zend_Db_Statement_Exception $e) {
+                // table `user_token_auth` might not yet exist when updating to Matomo 4
+                if (!strpos($e->getMessage(), 'user_token_auth')) {
```
tsteur: Would maybe need to additionally check whether the user_token_auth table exists or not for additional security. Ideally maybe the query in the user model (not sure which model is used there) would catch this maybe and on exception check for the text and whether table exists?  Technically could additionally even check the mysql error code maybe? (although that should not be too important)

sgiehl: @tsteur Haven't looked at the code in detail yet. But does the anonymous user need to have a token_auth in the table at all? Wondering if we couldn't simply return `anonymous` as token_auth for the `anonymous` user always 🤔

diosmosis: Tested locally and noticed w/ this code `Access::getInstance()->getSitesIdWithViewAccess()` always returns an empty array. In this case, we could also add a check that the sites w/ access is empty, that might be secure enough? Since we make sure the logged in user (which is no one, not even anonymous w/ this code) has basically no access.

tsteur: > Wondering if we couldn't simply return anonymous as token_auth for the anonymous user always

That should work too.

sgiehl:
@tsteur I've moved and improved the code a bit. Could you have another look

plugins/UsersManager/Model.php
```diff
@@ -141,6 +141,11 @@ public function getUsersLoginWithSiteAccess($idSite, $access)
         return $logins;
     }
 
+    public function tokenTableExists()
+    {
+        return !!$this->getDb()->fetchOne(sprintf("SHOW TABLES LIKE '%s';", $this->tokenTable));
```
tsteur: fyi you could probably use `DbHelper::getTablesInstalled` for this? Or maybe add a new method to DbHelper?

tsteur:
Left a comment. Otherwise should work 👍

plugins/UsersManager/Model.php
sgiehl: Ok. I'll move that to DbHelper. There was a method `tableExists`, but it has been removed as it was unused...

PR #15714, opened 2020-03-20
Remove user alias from DB

### LINKED ISSUE #13547: Remove user alias from database
Now that user alias has been removed from the UI, it can be removed from the DB as well. This should be done for Matomo 4.

tsteur:
fix https://github.com/matomo-org/matomo/issues/13547


plugins/CoreHome/templates/_topBar.twig
```diff
@@ -1,4 +1,4 @@
-{{ postEvent("Template.beforeTopBar", userAlias, userLogin, topMenu) }}
+{{ postEvent("Template.beforeTopBar", userLogin, topMenu) }}
```
diosmosis: Should we mention this change in the changelog?

plugins/ScheduledReports/tests/Integration/SubscriptionModelTest.php
```diff
@@ -20,7 +20,7 @@
  * @group SubscriptionModelTest
  * @group Plugins
  */
-class SubscriptionModelTest extends IntegrationTestCase
+class tSubscriptionModelTest extends IntegrationTestCase
```
diosmosis: typo

tsteur: Thanks, fixed

plugins/CoreHome/templates/_topBar.twig
tsteur: @diosmosis because they are defined in twig itself, they are not official events anyway so shouldn't be needed (see https://developer.matomo.org/api-reference/events where they aren't listed). Also checked in our code we don't seem to use them.

diosmosis:
Merged and created https://github.com/matomo-org/plugin-LoginLdap/issues/202. ActivityLog will also need to be updated I think.

Issue #15617, opened 2020-02-24
Make it possible to run tests against PHP 7.4


sgiehl:
We are currently running tests against PHP 7.2 and 7.3 only. Actually it would be good to run tests against minimum and latest PHP stable release (7.2 and 7.4 at that point).

This won't be directly possible with our current travis setup as PHP 7.4 can't be used with trusty distribution (GD extension doesn't work there)

What actually needs to be done:
- [x] Make travis scripts compatible with newer travis distributions (xenial / bionic) (see https://github.com/matomo-org/travis-scripts/pull/56)
- [x] Update PHPUnit, as currently used version is not compatible with 7.4 (see #15581)
- [x] Update core and plugin .yml files to use a newer distribution and run tests on 7.4

PR #15720, opened 2020-03-23
Run tests on bionic distribution & run AllTests with PHP 7.4

sgiehl:
Note: I've also done various tests with distributions `trusty` and `xenial`. Builds for both will work as well. But `trusty` doesn't support PHP 7.4

Also see https://github.com/matomo-org/travis-scripts/pull/56

refs #15617


diosmosis (dismissed):
LGTM, not sure if this requires other plugin PRs to be merged first or anything?

sgiehl:
Merging the travis scripts will need some updates on several travis.yml files. Will do that next week

sgiehl:
I've updated some more parts of the code and all tests seem to run on PHP 7.4 now.
Unfortunately it seems the version of imagemagick preinstalled on bionic distribution has problems with comparing the very big images. Those compares currently result in a 
```
the 'compare' command output could not be parsed, should be an integer, got: compare: 
../../magick/quantum.c:216: DestroyQuantumInfo: Assertion `quantum_info != (QuantumInfo *) NULL' failed.
```

Seems the policies set for imagemagick differs. While the limitations for image size or memory usage are commented out on trusty builds, they are limited on bionic builds. Maybe that causes that error. Will have a closer look at that tomorrow.

tsteur:
Is this a travis issue maybe? Could create a bug report there and wait for a month or two to see if something happens? Running tests on PHP 7.4 might not be highest priority so it could wait?

I suppose we can't use the old distro for UI tests and run it on older PHP versions easily?

sgiehl:
tried using an own policy for imagemagick, but that doesn't work either. Will check if we can use xenial for the UI build and if the error is the same there...

sgiehl:
Running UI tests only on xenial seems to work. For our plugins we need to see if we are running into the same problem or not. But guess there shouldn't be bigger images to compare and it should run on bionic as well.

@diosmosis could you have another look at the additional change I've made to fix the tests for PHP 7.4. Also the UI test for `dashboard2` has now another result, seems the widgets are ordered in another way. But I really don't understand why that should be the case. I didn't change anything in the code to create the dashboads and the used PHP version only differs in patch version. Can you think of a reason why that's happening?

core/DataTable/Renderer.php
```diff
@@ -265,7 +265,7 @@ protected function getApiMetaData()
 
             $api = \Piwik\Plugins\API\API::getInstance();
             $meta = $api->getMetadata($this->idSite, $apiModule, $apiAction);
-            if (is_array($meta[0])) {
+            if (!empty($meta[0]) && is_array($meta[0])) {
```
diosmosis, line 268: Should this be `!isset($meta[0])` in case `$meta[0]` is `[]`? Not sure if it should be noticed it would be ignored in that case.

diosmosis:
@sgiehl everything I see looks ok (except for one minor thing). Is there anything specifically you're worried about?

core/DataTable/Renderer.php
sgiehl: sure. isset makes sense
```suggestion
            if (isset($meta[0]) && is_array($meta[0])) {
```

sgiehl:
> Is there anything specifically you're worried about?

@diosmosis  as mentioned in the previous comment. The UIIntegrationTest for dashboard 2 seems to have a different result now, but I can't think of a reason. Do you maybe have an idea?

PR #15724, opened 2020-03-23
Add possibility to silently stop the archiver without an error

tsteur:
Needing this for cloud. We're listening there to `CronArchive.init.start` event and under circumstances we may trigger an exception there to not archive any data.

The problem is, that this exception is passed on to the `console` command and then it returns exit code  != 0 meaning we receive an email from crontab every time this happens.

Could have appended another parameter to the postEvent to specifically abort archiving in this event but wasn't sure if that could cause any issue breaking things in 3.X and using the exception you can technically abort the archiving at any time which may be helpful. Won't be an official feature though.

Tested this and worked nicely. We will be using this on Cloud already with 3.13.4 in a patch.


sgiehl (approved):
LGTM

PR #15728, opened 2020-03-24
Ensure to use correct date while archiving

### LINKED ISSUE #15727: Archiving specific date via console won't take date as argument.
When trying to archive data manually via CLI for specific date:

> ./console core:archive --force-date-range=2020-03-24,2020-03-24
or
> ./console core:archive --force-date-range="2020-03-24,2020-03-24"

we are experiencing following error: 

> Error: Got invalid response from API request: ?module=API&method=API.get&idSite=1&period=week&date=&format=php&trigger=archivephp. Response was 'a:2:{s:6:"result";s:5:"error";s:7:"message";s:3044:"Please specify a value for 'date'.

Seems like date parameter is not passed correctly to API call. This issue was introduced with Matomo version 3.13.4. It works correctly on version 3.13.3

sgiehl:
refs https://github.com/matomo-org/matomo/issues/15727

### DIFF
--- a/core/CronArchive.php
+++ b/core/CronArchive.php
@@ -23,7 +23,6 @@ use Piwik\DataAccess\ArchiveSelector;
 use Piwik\DataAccess\RawLogDao;
 use Piwik\Exception\UnexpectedWebsiteFoundException;
 use Piwik\Metrics\Formatter;
-use Piwik\Period\Factory;
 use Piwik\Period\Factory as PeriodFactory;
 use Piwik\CronArchive\SitesToReprocessDistributedList;
 use Piwik\CronArchive\SegmentArchivingRequestUrlProvider;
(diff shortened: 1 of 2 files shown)


core/CronArchive.php
```diff
@@ -991,7 +990,7 @@ public function isThereAValidArchiveForPeriod($idSite, $period, $date, $segment
         if ($isTodayIncluded
             && !$isLast
         ) {
-            return [false, null];
+            return [false, $date];
```
sgiehl: we need to return the date here, as otherwise the date used for archiving will be null here: 
https://github.com/matomo-org/matomo/blob/9969201da8fb875d6953e87b6e818b743c35e791/core/CronArchive.php#L902

core/CronArchive.php
```diff
@@ -978,10 +977,10 @@ protected function processArchiveDays($idSite, $lastTimestampWebsiteProcessedDay
     public function isThereAValidArchiveForPeriod($idSite, $period, $date, $segment = '')
     {
         if (Range::isMultiplePeriod($date, $period)) {
-            $rangePeriod = Factory::build($period, $date, Site::getTimezoneFor($idSite));
+            $rangePeriod = PeriodFactory::build($period, $date, Site::getTimezoneFor($idSite));
```
tsteur: @sgiehl is that change needed or is it actually causing an issue? Just asking out of curiosity.

core/CronArchive.php
tsteur: @sgiehl did you notice this by fixing some test or was something not working? I assume there was some core:archive used with specifically today?

sgiehl: see the linked issue https://github.com/matomo-org/matomo/issues/15727

core/CronArchive.php
sgiehl: Not really needed. But it's a bit messy to import the same class twice with different aliases...

diosmosis:
LGTM, there should be some tests covering this behavior that will need to be udpated.

diosmosis:
@sgiehl this fixes #15727 correct?

sgiehl:
Yes

Issue #11929, opened 2017-08-03
Updated / New Piwik Maps


sgiehl:
Our map wasn't updated for some years and some stuff is shown wrong or is incomplete.
Unfortunately our map generator (https://github.com/piwik/piwik-map-generator) is outdated and doesn't work anymore. 

We either should try to get the old generator working again or look out for some completely new maps.

Would be awesome if we would find some volunteer for this project, as we currently have more important issues and won't be able to handle this soon.

Summary of Issues that should be resolved with new maps:
#11266, #10963, #10123, #9719, #9170, #321 and maybe others

CaptainYukinoshitaHachiman:
So Tibet will become a part of China in the map after this update?
Great to see this XD

sgiehl:
@CaptainYukinoshitaHachiman  No. This be handled before in #11930

fengkaijia:
Adding to the issue board: Hong Kong and Macao are missing from the map (not in either World/Asia/China view). I guess the same goes to a few European city-size country/regions as well, e.g. the Holy See. How about moving to Highmaps (and add an option to allow non-commercial users to switch to it).

mattab:
>  How about moving to Highmaps (and add an option to allow non-commercial users to switch to it).

Their license is not compatible with the GNU/GPL so we can't use it

tassoman:
I dunno if you're interested but why overhelming the project with map building (going outside the matter of analytics) while there are already set mapping projects uut there? 

I mean can be interesting having a ready made map based on [leaflet.js](https://leafletjs.com/examples/choropleth/) filled with [OpenStreetMap ](https://osm.org)open data, or maybe a [d3-geo](https://github.com/d3/d3-collection/blob/v1.0.7/README.md#maps) based project?

Findus23:
@tassoman I don't know any specifics yet, but using one of those is pretty much what is planned.

tsteur:
@tassoman yes 99% likely we will use some existing solution as building a maps is a project in itself which we don't want to do. Better for us to focus on analytics :) 

There is also https://github.com/markmarkoh/datamaps (based on d3 though) and https://github.com/10bestdesign/jqvmap and a few other ones.

mattab:
Suggested next step: research for maybe 1 or 2 days and trying quickly whether any of the libs above could work for us.

fengkaijia:
ECharts (https://github.com/apache/incubator-echarts) might also work, it reads GeoJSON and provides a built-in world map.

[...]

PR #15735, opened 2020-03-26
Create HKG.svg

smallbigcake:
There was no map for Hong Kang which caused an 404 error in the UserCountryMap plugin.
I made one.
I used https://mapshaper.org/ to made it from data of http://www.naturalearthdata.com/downloads/


sgiehl:
@smallbigcake there are various smaller maps missing. The map you made would be displayed I guess, but the regions and cities would not be highlighted correctly on it as the maps lags some metadata.

We are planning to make completely new maps for Matomo 4. Would you maybe be keen to help work on that. If so, feel free to comment on #11929

PR #15736, opened 2020-03-26
New page performance reports

### LINKED ISSUE #12653: Provide more insights into page performance metrics
Performance matters nowadays a lot. Not only for SEO but also with every 100ms a page takes longer to load, you make typically less revenue etc.

To let users monitor and improve performance, it would be great to track more performance metrics such as the DNS load time (`performance.timing.domainLookupEnd - performance.timing.domainLookupStart`). Also the time it took to render the page in the client, the waiting time on the server, how much time was spent during the "ready event" and during the "load event", etc.... be very useful. See https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming

This helps to identify if improvements on performance really work and which effect it has on goal conversions, sales, revenue, etc.

This shouldn't be too complicated to implement but requires likely new DB fields in log tables unless we develop a new log table for this. Because an update for the existing log table be very slow, this would be something for Matomo 4.

sgiehl:
Performance data will be tracked for each page view and will contain the following metrics:
* **Latency (responseStart – fetchStart)**
How long it takes the response to get to the user’s device. This includes the time it takes for the request to get to the server, the time it takes the server to render a response, and the time until the first byte of that response gets back to the user’s device.
* **Transfer (responseEnd – responseStart)**
How long it takes the browser to download the response from the server.
* **DOM Processing to Interactive (domInteractive – domLoading)**
How long the browser spends loading the webpage until the user can starting interacting with it.
* **DOM Interactive to Complete (domComplete – domInteractive)**
How long it takes for the browser to load images/videos and execute any Javascript code listening for the DOMContentLoaded event.
* **Onload (loadEventEnd – loadEventStart)**
How long it takes the browser to execute Javascript code waiting for the window.load event.

Those metrics can be summed up to get the full page load time.

Action reports for at least urls will show the full page load time as new column
image

and an additional row action to show the page load evolution:

image

Also there is a new report page, that shows the overall page load evolution:

image


sgiehl:
The reports for page performance are almost ready, but there are some more adjustments needed for the tracking. 

Currently all the timing data would be sent with the pageview, but that doesn't actually work, as some of the performance timings might not yet be available when the pageview is tracked. Wondering how to solve that best. As sending the page view event after the onload event was finished won't be a usable solution, I currently see two possibilities with sending the timing data later:

- try to update the (last) page view tracked with those performance timings
 That would mean we need a new mechanism to not only update a visits, but also an action. Not sure if that is currently possible at all. Maybe we would need to send some kind of identifier with the pageview so we can identify the action again when sending the timing data, not sure if simply updating the last pageview of that visit would be accurate enough
- move everything into a new log table and track the timing data together with the visitor id, page url (and maybe the page title)
 That would mean the timing data isn't directly related to a specific action anymore. The currently generated reports would still be possible, but it would for example not be possible to show the timing data in the visitor log or profile for a specific action.

@tsteur What would you prefer? Or do you maybe have a different idea? Also maybe you or @mattab have some comments or additional suggestions on the already implemented stuff (see PR description)

tsteur:
@sgiehl which metrics might not be available?

sgiehl:
I actually haven't tested that through yet. But to have all metrics available in the timings api we would need to listen to the window.load event and add the tracking with a timeout of 0, so it's done when onload is finished. So depending on how the matomo tracker is included, I would say at least the onload metric won't be available and maybe also DOM Interactive to Complete, as Matomo might be executed within this process.

tsteur:
>  I would say at least the onload metric won't be available and maybe also DOM Interactive to Complete, as Matomo might be executed within this process.

We could send them with ping events, or other tracking requests if needed. Could even temporarily store data and track it on the next page view for the previous one should they visit another page but that gets too complicated. I would say we would simply document that in some cases the value wouldn't be recorded if the onload event happened yet and if they don't use the ping event or have other tracking requests. We certainly do not want to cause any extra tracking request just for that information.

Something to think about: Would we need to do anything for single page applications that track multiple page views within one page? I suppose only the first page would then get the metrics which be fine for now. Be good to not make this feature too complicated.

As for the update: I suppose either a log table or `log_link_visit_action` would be possible and technically allow the same features if it has an idVisit in there etc. Be great to reuse `log_link_visit_action` though I would say maybe (for simplicity). I suppose the update be like `update log_link_visit_action left join log_action set metric = xyz where idvisit = ? and idpageview = ? and log_action.type = pageview (1 or ...)`?

That might be ok as long as we only execute this query should it not be recorded yet. Meaning the tracker would need to make sure to send the metric only once and the server would only issue this query once and only if the action type is not a page view. When the metric is set in the request, and action type is a page view, then it can be recorded as part of the index. 

We'll need to see what impact this update has on disk I/O later once in production. We may want a feature later to disable recording metrics after the page view was tracked to avoid this update statement and instead if user wants to have these metrics, force them to listen to the onload event for the track page view.

sgiehl:
@tsteur ok. changed the tracking accordingly. If the metrics are already available when the pageview is tracked it will be sent with the pageview, otherwise they will be sent with the next tracking request.

sgiehl:
should be ready for a first review. Some tests of the submodules are still failing, but the expected system files will only need an update later...

tsteur:
@sgiehl in 
image

wonder if we would want an "All tables column visualisation" and then show all page specific performance metrics? 
Many people will never find this though. On the other side we maybe don't want to show all of them all the time so at least they would be able to get it, should they want to.

In this screen should we explain that some metrics might not be always tracked? and link to an FAQ or so?
image

Ideally we would remove the new performance menu item under Behaviour and instead show below screen as part of Visitors -> Overview  (below the visits overview, not as part of the visits overview sparklines)

image

Haven't looked at code yet otherwise or tested too much but looking good! The row evolution screen is quite cool!

sgiehl:
@tsteur 

- added a new table visualization to show the supported page reports with page performance metrics only:
image

- moved the page performance reports below the visits overview

- improved the overlay so it shows the metrics descriptions and a link to the online guide (that has still to be written, shall we create a separate issue here or in jira maybe?)
image

sgiehl:
There are still some weird failures in the UI tests. I'm still trying to investigate why. But a code review can already be done in the meantime...

mattab:
Quick feedback on the UI/UX: Considering the metrics are technical and hard to understand, could we 
* show the definition on hover in tooltips. 
* and also show them in the in-app help section? like on Visitors Overview:
screenshot_from_2020-03-30_23-56-24

* Also ideally these help text would explain both what it means + how to improve it eg.   "Avg latency is X. To make latency better, you can sometimes do X and Y. On the other hand Z will make latency worst." or similar/better.

sgiehl:
> show the definition on hover in tooltips.

That's already done for tables containing those metrics

> and also show them in the in-app help section? like on Visitors Overview:

Already added that: 
image

> Also ideally these help text would explain both what it means + how to improve it eg. "Avg latency is X. To make latency better, you can sometimes do X and Y. On the other hand Z will make latency worst." or similar/better.

Giving advices what to do and what not is not yet done. And that might also be hard to do. If we really want to give the users a good advice what they should do, we could create a simple plugin to get some useful data from Google's PageSpeed API

sgiehl:
The failing UI tests seem to fail because puppeteer fails to render the pages table report correctly. The additional column seems to make the report wider then the window causing puppeteer to render it like this:

image

Updated puppeteer locally, which seems to fix that issue. But it actually changes almost all screenshots slightly. Will do that in another PR, as that would be a bit too much to put it in here...

plugins/PagePerformance/lang/en.json
```diff
@@ -0,0 +1,34 @@
+{
+    "PagePerformance": {
+        "ColumnAveragePageLoadTime": "Avg. page load time",
+        "ColumnAveragePageLoadTimeDocumentation": "Average time how long it takes from requesting a page until the page is fully rendered within the browser",
+        "ColumnAverageTimeLatency": "Avg. latency time",
+        "ColumnAverageTimeLatencyDocumentation": "Average time how long it takes the response to get to the user’s device. Includes the time it takes for the request to get to the server, it takes the server to render a response, and until the first byte of that response gets back to the user’s device.",
+        "ColumnAverageTimeTransfer": "Avg. transfer time",
+        "ColumnAverageTimeTransferDocumentation": "Average time how long it takes the browser to download the response from the server.",
+        "ColumnAverageTimeDomProcessing": "Avg. DOM processing time",
+        "ColumnAverageTimeDomProcessingDocumentation": "Average time how long the browser spends loading the webpage until the user can starting interacting with it.",
```
tsteur, line 10: Be great to mention from what time this is counted. Eg from dom ready, from sending the first byte of the request to the server, from ...

tsteur, line 10: Be great to mention from what time this is counted. Eg from dom ready, from sending the first byte of the request to the server, from ...

plugins/PagePerformance/lang/en.json
```diff
@@ -0,0 +1,34 @@
...
+        "ColumnAverageTimeDomCompletion": "Avg. DOM completion time",
+        "ColumnAverageTimeDomCompletionDocumentation": "Average time how long it takes for the browser to load media and execute any Javascript code listening for the DOMContentLoaded event.",
```
tsteur, line 12: here as well be great to mention from what point this timer is counted so it's more clear

plugins/PagePerformance/lang/en.json
```diff
@@ -0,0 +1,34 @@
...
+        "ColumnAverageTimeOnLoad": "Avg. on load time",
+        "ColumnAverageTimeOnLoadDocumentation": "Average time how long it takes the browser to execute Javascript code waiting for the window.load event.",
```
tsteur, line 14: Same here. Be good to mention eg since the requesting the first byte or since the dom was loaded etc. BTW I suppose this could be also include loading images etc?

plugins/PagePerformance/lang/en.json
```diff
@@ -0,0 +1,34 @@
+{
+    "PagePerformance": {
+        "ColumnAveragePageLoadTime": "Avg. page load time",
+        "ColumnAveragePageLoadTimeDocumentation": "Average time how long it takes from requesting a page until the page is fully rendered within the browser",
+        "ColumnAverageTimeLatency": "Avg. latency time",
+        "ColumnAverageTimeLatencyDocumentation": "Average time how long it takes the response to get to the user’s device. Includes the time it takes for the request to get to the server, it takes the server to render a response, and until the first byte of that response gets back to the user’s device.",
+        "ColumnAverageTimeTransfer": "Avg. transfer time",
+        "ColumnAverageTimeTransferDocumentation": "Average time how long it takes the browser to download the response from the server.",
```
tsteur, line 8: Is this basically the time between receiving the first byte of the response and the last byte?

plugins/PagePerformance/lang/en.json
```diff
@@ -0,0 +1,34 @@
+{
+    "PagePerformance": {
+        "ColumnAveragePageLoadTime": "Avg. page load time",
+        "ColumnAveragePageLoadTimeDocumentation": "Average time how long it takes from requesting a page until the page is fully rendered within the browser",
```
tsteur, line 4: btw not sure I'm seeing it... is there a metric giving us the server time? Meaning eg the time between sending the last byte of the request and receiving the first byte of the response. AKA how long did the server take to start sending a response. This be very valuable.

plugins/PagePerformance/Metrics.php
```diff
@@ -0,0 +1,126 @@
...
+        return $mappedMetrics;
+    }
+
+    /**
+     * @return \Piwik\Plugins\PagePerformance\Columns\Metrics\AveragePerformanceMetric[]
+     */
+    public static function getAllPagePerformanceMetrics()
+    {
+        $metrics = [
+            new AverageTimeLatency(),
```
tsteur, line 54: We should also show the Avg Generation Time here? As mentioned somewhere else a `server time` be great as well if that's not there yet (if it's there already, maybe the naming could be better?) 

`Avg Generation Time` I also always found very confusing the naming. I would maybe think this would be either server time or the time it takes to fully render the page including server time. But not what it currently is... wonder if there's a better name.

js/piwik.js
```diff
@@ -3694,6 +3701,35 @@ if (typeof window.Piwik !== 'object') {
                         && performanceAlias.timing.requestStart && performanceAlias.timing.responseEnd) {
                         request += '&gt_ms=' + (performanceAlias.timing.responseEnd - performanceAlias.timing.requestStart);
                     }
+
+                    if (performanceAvailable && !performanceTracked) {
```
tsteur, line 3705: Not sure... do you reckon it makes sense to track at least some metrics that are already available when eg `trackPageview` is called? And some other metrics that are only available after the load event track them with the next request once `performanceAvailable  = true`? This way would get bit more data and the metrics that were sent become more accurate. Eg `server time`. 
Of course it means we don't always have all information for every user because there might be no second tracking request, but at least there will be some data and more accurate etc.

plugins/PagePerformance/Controller.php
```diff
@@ -0,0 +1,155 @@
...
+use Piwik\Plugins\CoreVisualizations\Visualizations\JqplotGraph\Evolution as EvolutionViz;
+use Piwik\Plugins\PagePerformance\Visualizations\JqplotGraph\StackedBarEvolution;
+use Piwik\View;
+use Piwik\ViewDataTable\Factory as ViewDataTableFactory;
+
+class Controller extends PluginController
+{
+    public function indexPagePerformance()
+    {
+        $view = new View('@PagePerformance/getPagePerformancePopover');
```
tsteur, line 27: I suppose permissions are checked in the API method. Might still be good to call `$this->checkSitePermission();` just like in the other methods.

plugins/PagePerformance/Archiver.php
```diff
@@ -0,0 +1,124 @@
...
+                            case when $table.$column is null
+                                then 0
+                                else 1
+                            end
+                          ) as {$column}_hits";
+        }
+
+        $joinLogActionOnColumn = array('idaction_url');
+
+        $query = $this->getLogAggregator()->queryActionsByDimension([], 'log_action1.type = ' . Action::TYPE_PAGE_URL, $selects, false, null, $joinLogActionOnColumn);
```
tsteur, line 63: Instead of having a separate query, would it be possible to hook into `Actions.Archiving.addActionMetrics` event just like Bandwidth does? Would safe an additional query and maybe this way would also work for page titles automatically? We're seeing not always do people have URLs (eg when mobile app etc)

plugins/PagePerformance/lang/en.json
tsteur: also for all the metrics be good to write eg `Average time in seconds` or `Average time in ms` just in case someone fetches eg the unformatted metrics through the API etc.

plugins/PagePerformance/Columns/TimeDomCompletion.php
```diff
@@ -0,0 +1,74 @@
...
+use Piwik\Plugin\Dimension\ActionDimension;
+use Piwik\Tracker\Action;
+use Piwik\Tracker\ActionPageview;
+use Piwik\Tracker\Request;
+use Piwik\Tracker\Visitor;
+
+class TimeDomCompletion extends ActionDimension
+{
+    protected $columnName = 'time_dom_completion';
+    protected $columnType = 'INTEGER(10) UNSIGNED NULL';
```
tsteur, line 25: Are we storing values as milliseconds? Then maybe a mediumint would still do? Considering `16777215 ms = 280 minutes`... also for the other dimensions

plugins/PagePerformance/javascripts/jqplotStackedBarEvolutionGraph.js
```diff
@@ -0,0 +1,218 @@
...
+                        if (!dataByAxis[axis]) {
+                            dataByAxis[axis] = [];
+                        }
+
+                        var value = self.formatY(valueUnformatted, d);
+                        var series = self.jqplotParams.series[d].label;
+
+                        var seriesColor = self.jqplotParams.seriesColors[d];
+
+                        dataByAxis[axis].push('<span class="tooltip-series-color" style="background-color: ' + seriesColor + ';"/>' + '<strong>' + value + '</strong> ' + piwikHelper.htmlEntities(series));
```
tsteur, line 111: I suppose the `value` here and below is a number but may still be code to convert it to html entities or so just to be extra safe? Also `seriesColor` etc.

plugins/PagePerformance/javascripts/PagePerformance.js
```diff
@@ -0,0 +1,136 @@
...
+        var labelParts = label.split(separator);
+        for (var i = 0; i < labelParts.length; i++) {
+            var labelPart = labelParts[i].replace('@', '');
+            labelParts[i] = $.trim(decodeURIComponent(labelPart));
+        }
+        var delimiter = piwik.config.action_url_category_delimiter;
+        if(apiMethod.indexOf('PageTitles') >= 0) {
+            delimiter = piwik.config.action_title_category_delimiter;
+        }
+        label = labelParts.join(delimiter);
```
tsteur, line 87: In some other places we do eg 

```js

        segmentName = piwikHelper.escape(segmentName);
        segmentName = piwikHelper.htmlEntities(segmentName);
        segmentValue = piwikHelper.escape(segmentValue);
        segmentValue = piwikHelper.htmlEntities(segmentValue);
        segmentName = segmentName.replace(/(&amp;)(#[0-9]{2,5};)/g, '&$2');
        segmentValue = segmentValue.replace(/(&amp;)(#[0-9]{2,5};)/g, '&$2');
```

Wonder if that's needed here too for the label? Considering `setTtile()` might do a `jQuery().html(title)` should the label be longer than 60 characters and maybe a user could have quotes in the title or so? Not sure... (might be good to debug or double check)

plugins/PagePerformance/PagePerformance.php
```diff
@@ -0,0 +1,145 @@
...
+/**
+ */
+class PagePerformance extends \Piwik\Plugin
+{
+    public static $availableForMethods = [
+        'getPageUrls',
+        'getEntryPageUrls',
+        'getExitPageUrls',
+        'getPageUrlsFollowingSiteSearch',
+        'getPageTitles',
```
tsteur, line 25: So far the data for page titles doesn't seem to be archived? But as mentioned in another comment once we hook into actions archiving this should work out of the box

plugins/PagePerformance/Tracker/PerformanceDataProcessor.php
```diff
@@ -0,0 +1,85 @@
...
+                );
+            }
+        }
+
+        if (empty($valuesToUpdate)) {
+            return; // no values to update given with the request
+        }
+
+        $query = sprintf(
+            'UPDATE %1$s LEFT JOIN %2$s ON idaction_url_ref = idaction SET %3$s WHERE idvisitor = ? AND idpageview = ? AND %2$s.type = 1',
```
tsteur, line 75: Would using `idvisit` be much faster than `idvisitor`? It would otherwise need to look at many more rows? It probably also be more accurate cause should someone have multiple visits then a duplicate of idpageview becomes more likely.

Not sure I understand why `idaction_url_ref ` is used? I think I would have expected the join to be `(la.idaction = lva.idaction_url) or (lva.idaction_url = 0 and la.idaction = lva.idaction_name)`. This way it also works eg for mobile apps or in general should only action_name be given. Also I don't think the url_ref is needed but the actual ref?

Also wondering if it be better to have a select query, and then an update query `where idlink_va = X` to prevent locking issues. Cause this update statement would potentially lock quite a few rows in log_action and log_link_visit_action and should there be more tracking requests coming in at the same time, this can cause issues. AFAIK Mysql will lock all rows it needs to look at.

tsteur: For the update itself you could then also use `Tracker/Model::updateAction`

plugins/PagePerformance/Tracker/PerformanceDataProcessor.php
```diff
@@ -0,0 +1,85 @@
...
+        $pageViewId = $request->getParam('pv_id');
+        $visitorId = $request->getVisitorId();
+
+        // ignore requests that can't be associated with an existing page view of a visitor
+        if (empty($pageViewId) || empty($visitorId)) {
+            return;
+        }
+
+        /** @var ActionDimension[] $performanceDimensions */
+        $performanceDimensions = [
```
tsteur, line 49: Be great to add some integration tests for below use case if possible. I suppose they might be already indirectly tested though through system tests. Maybe could just extract things into methods and test these methods.

eg one method fror line `48:68` -> `getPerformanceValuesToUpdate($request)` and `updatePerformanceValues()`

tsteur (changes requested):

plugins/PagePerformance/Archiver.php
sgiehl: That is actually done and it should already work for page titles.... The archiving here are the overall numbers. It's all exactly the same way as Bandwidth plugin does 🤔 

https://github.com/matomo-org/matomo/blob/2f557a0f8479ca7bb16304dbdc0f324073aef5eb/plugins/PagePerformance/PagePerformance.php#L42
https://github.com/matomo-org/matomo/blob/2f557a0f8479ca7bb16304dbdc0f324073aef5eb/plugins/PagePerformance/PagePerformance.php#L128-L131
https://github.com/matomo-org/matomo/blob/2f557a0f8479ca7bb16304dbdc0f324073aef5eb/plugins/PagePerformance/Metrics.php#L81-L124

plugins/PagePerformance/Columns/TimeDomCompletion.php
sgiehl: Didn't spent too much time on thinking about the field type. Will change it to mediumint

plugins/PagePerformance/Metrics.php
sgiehl: The performance timings basically has this metrics:



The current generation time is the time from `requestStart` to `responseEnd`. So it's basically the time from the browser sending the request, to the browser having received the complete answer. To be honest - I think that metric is not very useful as you imho can't work with it in any way. You don't know if the server needs long or if the network was slow. I would actually even suggest to remove it

The time the server needs to process the request is currently included in `Latency`, together with DNS and TCP overhead, which should be very low in most cases. If you think that would be useful, we could split up the Latency time into `Network Latency` and `Page Generation Time`. Even though that would mean the `Network Latency` would be actually `0` quite often, as DNS/TCP gets cached after the first request.

The metrics we currently have can be simply summed up to get the full time it took to render the page completely. That's the reason why I did not include the current page generation time anywhere in the new reports.

plugins/PagePerformance/javascripts/PagePerformance.js
sgiehl: Will add such an encoding, to be sure there can't be any kind of XSS

plugins/PagePerformance/javascripts/jqplotStackedBarEvolutionGraph.js
sgiehl: Besides the total value calculation it's the same as for evolution chart. So should be save...

https://github.com/matomo-org/matomo/blob/3f26e785f015d30d0aeea66aaf7484111b0dbfa9/plugins/CoreVisualizations/javascripts/jqplotEvolutionGraph.js#L101-L118

plugins/PagePerformance/lang/en.json
sgiehl: correct

plugins/PagePerformance/lang/en.json
sgiehl: No. The Onload time is only the time the browser spends to execute the javascript callbacks that were listening to the window.load event. If the javascript requests any stuff synchronous, that would be counted indirectly as well. But for most pages that shouldn't be the case.

plugins/PagePerformance/Tracker/PerformanceDataProcessor.php
sgiehl: Makes sense. Will update it.

plugins/PagePerformance/Tracker/PerformanceDataProcessor.php
sgiehl: will add some simple integration tests to check if updating those values in ongoing requests work as expected

js/piwik.js
sgiehl: ok. will change it that with a page view all data that is already available will be sent. If not all were already available the next request will send the performance data (again) and update the values.

plugins/PagePerformance/Archiver.php
tsteur: Not sure I understand why the extra query is needed then? Is this just to get the overall metrics? Could in that case the sum be already done in the query? Wonder if even a custom query somehow would need to be used since it wouldn't include page views with no page url?

tsteur: Might be ok though if they aren't included

plugins/PagePerformance/Metrics.php
tsteur: @sgiehl agreed. I've always found this metric useless too and not helping. I wonder if there's an alternative to remove it since some users would maybe want it simply for historical reasons. Not sure how to best do that though. Like mentioning in the description this metric is deprecated and only there for historical reasons. We could include one useful performance metric in the regular table report, then have this metric also in the performance report but could list it last or so. And for new installs we could completely remove it based on `DbHelper::getInstallVersion()`. Just a thought

plugins/PagePerformance/Archiver.php
sgiehl: Actually guess we could simply sum up all values regardless of the action type, as the values will only be available for page views, so no need to limit it with the where condition. Will change that

plugins/PagePerformance/Metrics.php
sgiehl: I'm not sure how to handle that best either. Have added the page generation time to the page performance table as last column if Matomo was installed before 4.0. If Matomo was installed later the page generation time won't be shown there and will now also be hidden in the page url / title reports and the visit summary sparkline for it.
Have not yet added some kind of note that the page generation time metric is deprecated. Currently we would still go on with tracking it. Wondering if we should remove that in that case 🤔

tsteur: You mean no longer track it if it was installed from Matomo 4? That could work if it's easy to do. Would need to make sure this option is cached in tracker. Eventually it could allow using the custom_float for something else I suppose. We could maybe also just completely stop tracking it maybe. Considering they could still get kind of the same value (by aggregating the new metrics) and compare it to previous value manually... 
They would keep the historical data also... it may confuse people though when it now always says 0 for average generation time once they upgrade. Could the avg generation time for these people be a computed/processed metric aggregating two metrics like server time = transfer time? Not sure if that be easy to do or make sense...

sgiehl: Ok. To be able to calculate the current page generation time with the performance metrics, we would need to split up latency into `network` and `server` latency. Page generation could then be calculated with `server latency + transfer time`. But actually I don't see any benefits in being able to compute a metric that is still kind of useless. We could also simply sum up latency and transfer time, but that would mean also network latency is included, resulting in a bit higher numbers. If someone really tries to use those number, having higher number after the update of Matomo might be bit confusing...

tsteur: Splitting up latency into network and server might be beneficial independent of this? AKA `server` time = `'(performanceAlias.timing.responseStart - performanceAlias.timing.requestStart);` and network time = `'(performanceAlias.timing.requestStart - performanceAlias.timing.fetchStart);` ?

I guess typically the network time be quite low. Like `2ms` or something. But it could also be higher eg when having users around the world and no CDN, etc and would just help better understand where issues might be?

Calculating that metric indeed wouldn't be needed I'd say. I'd say we could also stop tracking that new metric and if people are still interested in it, then they could be looking at the other metrics to compare things. 

The problem be still that users who installed Matomo before 4.0 would see `0` in some newer reports. 
* We could set a date when they upgraded to Matomo 4 in the options table and if a start date of a report is past that day, remove that metric. 
* Or we could iterate over all columns and should all values be `0` then we remove the column. However this might be a bit slow?

sgiehl: > I guess typically the network time be quite low. Like 2ms or something. But it could also be higher eg when having users around the world and no CDN, etc and would just help better understand where issues might be?

Correct. It should be very low or even zero for most requests, as after the first request DNS/TCP is cached and won't take any time. Can split up that metric if you prefer, but that would mean another couple of hours of work to adjust everything. Let me know, and I'll do it.

> I'd say we could also stop tracking that new metric

Will remove that everywhere. And add a deprecation note

> Or we could iterate over all columns and should all values be 0 then we remove the column. However this might be a bit slow?

That's kind of already done before computing the metric:
https://github.com/matomo-org/matomo/blob/b16a791aa3650d85af829156129c2bd44c7cb075/plugins/Actions/Columns/Metrics/AveragePageGenerationTime.php#L66-L101

But the columns aren't removed from being displayed, as it seems to be done too early. Seems that is broken for longer time already. Will move removing the columns from `columns_to_display` to a later point. Then it should hide the columns if they don't have a value in the report.

sgiehl: > I'd say we could also stop tracking that new metric

Just realized, that it was also possible to manually set that metric, which might maybe be used by some people to track the time the server actually spent. Will remove the part, that calculates the metric based on browser timings. Or shall we also remove the possibility to set that manually?

tsteur: > Or shall we also remove the possibility to set that manually?

I reckon should someone set it, then we could log a warning that this feature is deprecated in favour of the new ones? Would need to get the new metrics also implemented ideally by other SDKs (android/ios/java/...). Some of the metrics may be useful for them eg the load time etc.

>  Can split up that metric if you prefer, but that would mean another couple of hours of work to adjust everything. Let me know, and I'll do it.

It be really great not spending more time on this but I think it's really valuable splitting it up simply because it helps a great deal when trying to act based on this data as network and server are 2 totally different problems and you wouldn't know where the actual problem was. Be good to add and hopefully it's quick to add.

sgiehl:
@tsteur the metrics are now split up as discussed. But it's still bit unclear to me how we should handle the old `page generation` metric. As I tried to describe before, piwik.js has a method `setGenerationTimeMs`, which allows it to set a custom value for the generation time. The new metrics are automatically filled based on the performance timing, with no possibility to set them manually. So users that used to set the value manually can't use the new methods for that anymore.
So how shall we handle that? shall we remove that possibility to set a custom value completely? e.g. remove the method `setGenerationTimeMs` or log a warning when the method is called instead. Also should the `gt_ms` param be processed on Matomo side if the value is set in a request or shall we remove that, as well?

tsteur:
> So how shall we handle that? shall we remove that possibility to set a custom value completely? e.g. remove the method setGenerationTimeMs or log a warning when the method is called instead. 

I would say it should be fine. 

> Also should the gt_ms param be processed on Matomo side if the value is set in a request or shall we remove that, as well?

I'd say we can remove it then as well.

@mattab fyi... we're removing generation time as it was never useful anyway and it'll be replaced by the new metrics see https://github.com/matomo-org/matomo/pull/15736#discussion_r402599498 existing data will still be shown for older reports.

sgiehl:
@tsteur I've updated the PR and all related PRs so it's no longer possible to set the page generation time. Guess this and all related PRs should be ready for a final review...

tassoman:
Cheers team! It's a very good geek improvement but, isn't it a bit tricky report for a target of non-techy editors? Most of all they don’t have any clue about DOM ... :thinking:

tsteur:
fyi @sgiehl there are some conflicts, otherwise looks good

sgiehl:
@tassoman thanks for the comment. There are inline helps that explain those metrics for those who don't understand them directly. Also the default reports will only show the page load time, which should be clear to everyone

AnnoyingTechnology:
Hi and thanks for the new feature.

I've searched for "equivalent", "corresponds", "identical" in this issue, looking for **what would correspond to the old "avg generation time"**, no luck.

**Is there any ?** A client would like to be able to compare old 3.0 matomo statistics with newer 4.0 one.

tsteur:
Hi @AnnoyingTechnology that's unfortunately not quite possible unless you maybe built a custom plugin because the previous generation time is basically a combination of several metrics within the page performance. Sorry about that. If needed, instead of a custom plugin you could also connect to our API https://developer.matomo.org/api-reference/reporting-api and fetch the reports and aggregate this number in a script if needed. Fully aware though this might not be an option in this case. @sgiehl do you remember which performance metrics are basically equal to the previous generation time?

sgiehl:
Assuming that most people only tracked the time the server needs to generate the page there is no value it could be directly mapped to.
The value that matches most is `Latency`. But that one also includes the time it took the request to reach the server.

AnnoyingTechnology:
Thanks to both of you for the clarification.
I wouldn't want to criticize a great tool that I benefit from for free. But this kinda look like an oversight. 

In the case of my client, we migrated his applications to new servers, and upgraded the Matomo instance at the same time. Which prevents us from comparing the before/after performance easily.

Anyways, the new performance reports are so much better, he can deal with that !

PR #15737, opened 2020-03-26
Fix tests for tracker plugins no longer worked

tsteur:
Regression from https://github.com/matomo-org/matomo/pull/15568

When running JS tests like https://foo.bar/tests/javascript/ then the tracker.js of tracking plugins was no longer included in the matomo.js


diosmosis:
should this change be visible in tests?

tsteur:
@diosmosis nope. Only in premium feature JS tests.

PR #15744, opened 2020-03-29
Adds INI setting to disable referrer spam blocklist 

john357smith:
Added option to disable referrer spam blacklist.

PR #15745, opened 2020-03-29
Update global.ini.php

john357smith:
Added option to disable spam blacklist.

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -754,9 +754,6 @@ custom_cacert_pem=
 ; Default is 1.
 enable_tracking_failures_notification = 1
 
-; Disable using referrer spam blacklist.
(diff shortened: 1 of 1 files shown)


config/global.ini.php
```diff
@@ -895,6 +895,9 @@
 ; NOTE: you must also set "[Tracker] debug = 1" to enable the profiler.
 enable_sql_profiler = 0
 
+; Enables using referrer spam blacklist.
+enable_spam_filter = 1
```
tsteur: @john357smith could you make that change as part of the other PR https://github.com/matomo-org/matomo/pull/15744 ?

john357smith:
Changes moved to PR 15744.

PR #15750, opened 2020-03-30
Fix displaying of actions when loading additional visits in visitor profile

sgiehl:
When clicking the `load more visits` link in a visitor profile, the post loaded visits currently do not display any details for the actions.

See bug report in forum: https://forum.matomo.org/t/besucherprofil-nicht-alle-besuche-werden-detailliert-angezeigt/36614


diosmosis:
Is there a way for this to be visible in UI tests?

sgiehl:
@diosmosis added a UI test for that...

PR #15752, opened 2020-03-30
idsegments as option

mikkeschiren:


tsteur:
@mikkeschiren we are currently only merging into 4.x unless it is  some bigger issue. New features we don't really merge. However, we may merge this PR after the PR was merged in 4.X since it is the same change and should be safe. It might just not make it into the next release.

plugins/CoreAdminHome/Commands/InvalidateReportData.php
```diff
@@ -68,7 +71,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-        $segments = $this->getSegmentsToInvalidateFor($input, $sites);
+
+        $segments = array();
+        if ($input->getOption('segment')) {
+            $segments = $this->getSegmentsToInvalidateFor($input, $sites);
+        }
+        if ($input->getOption('idsegments')) {
+            $idsegments = $this->getSegmentsIdToInvalidateFor($input, $sites);
+        } else {
+            $segments = $this->getSegmentsToInvalidateFor($input, $sites);
```
tsteur, line 82: @mikkeschiren wouldn't this set the segments even when not the segment option is set? like a few lines above in https://github.com/matomo-org/matomo/pull/15752/files#diff-72b51ba044329b6585537f817a184a4aR76-R77 it only executes that method if the option table is set.

I suppose it is supposed to be possible to set both `--segment=` and `--idsegments` in the same call?

I suppose it would need to look a bit like below?

```php
 $segments = $this->getSegmentsToInvalidateFor($input, $sites);
if ($input->getOption('idsegments')) {
    $idsegments = $this->getSegmentsIdToInvalidateFor($input, $sites);
    $segments = array_merge($segments, $idsegments);
    $segments = array_unique($segments);
} 
```

plugins/CoreAdminHome/Commands/InvalidateReportData.php
```diff
@@ -223,4 +239,40 @@ private function getSegmentsToInvalidateFor(InputInterface $input, $idSites)
         }
         return $result;
     }
+
+    private function getSegmentsIdToInvalidateFor(InputInterface $input, $idSites)
+    {
+        $idSegments = $input->getOption('idsegments');
+        $idSegments = trim($idSegments);
+        $idSegments = preg_replace('/\s/', '', $idSegments);
```
tsteur, line 247: fyi this preg_replace might not be needed since you already trim all the IDs below?

plugins/CoreAdminHome/Commands/InvalidateReportData.php
```diff
@@ -223,4 +239,40 @@ private function getSegmentsToInvalidateFor(InputInterface $input, $idSites)
...
+        $segments = array_map(function ($segment) {
+            return $segment['definition'];
+        }, $segments);
+
+        $result = array();
+        foreach ($segments as $segmentString) {
+            try {
+                $result[] = new Segment($segmentString, $idSites);
+            } catch (\Exception $e) {
+                return $e->getMessage();
```
tsteur, line 273: we'd actually expect the exception not to be caught just like in the current `getSegmentsToInvalidateFor()` method. This way user will know there is an invalid segment. 
Otherwise you would for example return a `string` here instead of an `array` and the code above would not work correctly anymore.

tsteur (changes requested):
@mikkeschiren looks generally good but left a few comments. Will be a great feature if you could have a look at the comments.

tsteur:
@mikkeschiren I'll close this PR for now as we haven't heard from you in a while but will be happy to reopen should you be keen to follow up on the other comments. Simply let us know and we'd be happy to review things again.

PR #15753, opened 2020-03-30
idsegments as option

### LINKED ISSUE #14763: core:invalidate-report-data command should accept --idsegments (easier to use than --segments)
`core:invalidate-report-data` console command should accept `--idsegments` (easier to use than `--segments`). Suggested by several users and also experienced myself the difficulty of building suitable commands when many special characters are parsed/executed by the shell.

Would be great to have a new `--idsegments` parameter, similar to how we have `--force-idsegments=` in `core:archive` command.

mikkeschiren:
fix https://github.com/matomo-org/matomo/issues/14763

### DIFF
--- a/plugins/CoreAdminHome/Commands/InvalidateReportData.php
+++ b/plugins/CoreAdminHome/Commands/InvalidateReportData.php
@@ -13,6 +13,7 @@ use Piwik\Date;
 use Piwik\Period;
 use Piwik\Period\Range;
 use Piwik\Piwik;
+use Piwik\Plugins\SegmentEditor\Model as SegmentEditorModel;
 use Piwik\Segment;
 use Piwik\Plugin\ConsoleCommand;
 use Piwik\Plugins\SitesManager\API as SitesManagerAPI;
@@ -44,6 +45,8 @@ class InvalidateReportData extends ConsoleCommand
             self::ALL_OPTION_VALUE);
         $this->addOption('segment', null, InputOption::VALUE_IS_ARRAY | InputOption::VALUE_REQUIRED,
             'List of segments to invalidate report data for.');
+        $this->addOption('idsegments', null, InputOption::VALUE_REQUIRED,
+            'List of id of segments to invalidate report data for.');
         $this->addOption('cascade', null, InputOption::VALUE_NONE,
             'If supplied, invalidation will cascade, invalidating child period types even if they aren\'t specified in'
             . ' --periods. For example, if --periods=week, --cascade will cause the days within those weeks to be '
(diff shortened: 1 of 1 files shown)


plugins/CoreAdminHome/Commands/InvalidateReportData.php
```diff
@@ -68,7 +71,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        }
+        if ($input->getOption('idsegments')) {
+            $idsegments = $this->getSegmentsIdToInvalidateFor($input, $sites);
+        } else {
+            $segments = $this->getSegmentsToInvalidateFor($input, $sites);
+        }
+        if (isset($idsegments)) {
+            $segments = array_merge($segments, $idsegments);
+            $segments = array_unique($segments);
+        }
```
diosmosis, line 87: I think the code here can be simplified. `$segments = $this->getSegmentsToInvalidateFor($input, $sites);` is executed no matter what but can be called twice if `--segment` is supplied. And the `if (isset($idsegments)) {` block could be located next to `$idsegments` w/o the `if`. Can you make these changes?

plugins/CoreAdminHome/Commands/InvalidateReportData.php
```diff
@@ -223,4 +239,40 @@ private function getSegmentsToInvalidateFor(InputInterface $input, $idSites)
         }
         return $result;
     }
+
+    private function getSegmentsIdToInvalidateFor(InputInterface $input, $idSites)
+    {
+        $idSegments = $input->getOption('idsegments');
+        $idSegments = trim($idSegments);
+        $idSegments = preg_replace('/\s/', '', $idSegments);
```
diosmosis, line 247: `trim` and `preg_replace` are redundant here. Only the preg_replace is needed.

plugins/CoreAdminHome/Commands/InvalidateReportData.php
```diff
@@ -223,4 +239,40 @@ private function getSegmentsToInvalidateFor(InputInterface $input, $idSites)
...
+        $idSegments = explode(",", $idSegments);
+        $idSegments = array_map('trim', $idSegments);
```
diosmosis, line 249: W/ the preg_replace above, this call isn't necessary.

plugins/CoreAdminHome/Commands/InvalidateReportData.php
```diff
@@ -223,4 +239,40 @@ private function getSegmentsToInvalidateFor(InputInterface $input, $idSites)
...
+        $segments = array_map(function ($segment) {
+            return $segment['definition'];
+        }, $segments);
+
+        $result = array();
+        foreach ($segments as $segmentString) {
+            try {
+                $result[] = new Segment($segmentString, $idSites);
+            } catch (\Exception $e) {
+                return $e->getMessage();
```
diosmosis, line 273: Since this method should return an array of segments, if an error is encountered, can you just output a warning/error and ignore the segment?

plugins/CoreAdminHome/Commands/InvalidateReportData.php
```diff
@@ -223,4 +239,40 @@ private function getSegmentsToInvalidateFor(InputInterface $input, $idSites)
...
+        $idSegments = $input->getOption('idsegments');
+        $idSegments = trim($idSegments);
+        $idSegments = preg_replace('/\s/', '', $idSegments);
+        $idSegments = explode(",", $idSegments);
+        $idSegments = array_map('trim', $idSegments);
+        $idSegments = array_unique($idSegments);
+
+        if (empty($idSegments)) {
+            return array(null);
+        }
```
diosmosis, line 254: What's the reason for returning `[null]`? Should it instead return an empty array so nothing is invalidated? Given it's unexpected we could also output a warning here.

diosmosis (changes requested):
Thanks @mikkeschiren for creating your first contribution! I left some comments in the review, let me know if you have any questions or need some help.

tsteur:
@mikkeschiren thanks a lot for this PR. Very appreciated. Do you mind looking at the comments from @diosmosis ?

tsteur:
@mikkeschiren as mentioned in previous comment could you look at the comments from @diosmosis ?

tsteur:
@mikkeschiren I'll close this PR for now as there has been no response to the comments in quite a while but we'll be very happy to do another review should you continue with this PR

PR #15755, opened 2020-03-31
try to fix two UI tests

### LINKED ISSUE #15725: Fix automatic login after one click update
Our UI tests for automatic login after a one click update currently fails on 4.x-dev
That might be a regression and should be fixed.

diosmosis:
Changes:
* login to matomo after oneclickupdate is done (this is required now I believe due to the app specific token auth change. might not be an issue after 4.x...)
* fixes for a couple UI tests


tests/PHPUnit/Fixtures/LatestStableInstall.php
```diff
@@ -140,6 +140,7 @@ private function cloneMatomoPackageRepo()
     {
         $pathToMatomoPackage = PIWIK_INCLUDE_PATH . '/../matomo-package';
         if (file_exists($pathToMatomoPackage)) {
+            return;
```
sgiehl, line 143: guess that was for debug purpose?

diosmosis: 👍 will remove

diosmosis:
Note: the one click update changes should be merged into 3.13.5 to make sure as many people as possible have an error free one click update to matomo 4. Though we'll have to try it before it's released...

plugins/UserCountry/VisitorGeolocator.php
```diff
@@ -305,6 +305,11 @@ private function getDefaultProvider()
     public static function getDefaultLocationCache()
     {
         if (self::$defaultLocationCache === null) {
+            if (!class_exists(Transient::class)) {
+                $ex = new \Exception();
+                print $ex->getTraceAsString()."\n";
+                exit;
+            }
```
diosmosis, line 312: note to self (or whoever wants to take care of it): need to get rid of this debugging code

sgiehl:
@diosmosis the changes for updater are exactly the same as in the PR for `3.x-dev`, right?
Might make sense to remove them from this PR then as they will be merged over from `3.x-dev` to `4.x-dev` anyways. (Might avoid some merge conflicts though)

tests/UI/specs/OneClickUpdate_spec.js
```diff
@@ -75,6 +75,12 @@ describe("OneClickUpdate", function () {
             }
         }
 
+        await page.waitFor('#login_form_login', { visible: true });
+
+        await page.type('#login_form_login', 'superUserLogin');
+        await page.type('#login_form_password', 'superUserPass');
+        await page.click('#login_form_submit');
```
sgiehl, line 82: So automatic login after update won't work anymore? 🤔

diosmosis: What automatic login are referring to? From my understanding the session is just sent again after the update, but I think it's when updating to 4.x due the app specific token auth change it requires a new login. Otherwise anyone would get logged in even if they aren't authorized to?

sgiehl: That's what I meant. Guess it's not worth to have a closer look if it's possible to not require a new login...

diosmosis: @tsteur might know if the session should still work

sgiehl: if it's ok that the session gets terminated, the changes would be good to merge

tsteur: > but I think it's when updating to 4.x due the app specific token auth change it requires a new login. Otherwise anyone would get logged in even if they aren't authorized to?

Not sure I fully understand the problem @sgiehl @diosmosis 

Is this something that happens for the actual user or is more of a problem only for tests? Generally, when updating, the current token_auth should be automatically be migrated to an app specific token and still work. I suppose the token_auth is not really used for the session though.

What does change is indeed the session cookie maybe as we're removing login and token hash or so from the cookie AFAIK. Not sure if this causes the issue?

If user gets logged out would this cause an issue for https://github.com/matomo-org/matomo/pull/15770/files#diff-adf51e994c74a650d8afec954288d4c3R178 where we require user to be logged in? Not sure if user can be logged out in between (either by this session change or because the cookie expire time which might expire after 30 minutes just in between steps).

Looking at the original PR for app specific tokens, I can only imagine that maybe this session value is missing and causing an issue (random guess) https://github.com/matomo-org/matomo/pull/15410/files#diff-6c322e8c4c05670597358309f152c51aR94

Generally I would say it's fine though if a user gets logged out during the update I would say. As long as it's possible to complete the update fully.

diosmosis: > If user gets logged out would this cause an issue for https://github.com/matomo-org/matomo/pull/15770/files#diff-adf51e994c74a650d8afec954288d4c3R178 where we require user to be logged in?

It might be... though we only make the request internally in Matomo, so maybe we can create a temporary token auth for that request to make sure it works? I'd do this in a separate PR.

tsteur: That could work @diosmosis

diosmosis: @tsteur made the change after merging in changes from 3.x, can review it again

tsteur: 👍

PR #15760, opened 2020-03-31
remove useless return

tolbon:


tsteur:
@tolbon thanks for this PR, very much appreciated. Could you issue that PR against `4.x-dev` as we only merge regressions into `3.x-dev`?

PR #15770, opened 2020-04-02
One click update in two parts so new code is loaded for second.

diosmosis:


plugins/CoreUpdater/Updater.php
```diff
@@ -138,9 +139,45 @@ public function updatePiwik($https = true)
             throw new UpdaterException($e, $messages);
         }
 
+        $partTwoUrl = Url::getCurrentUrlWithoutQueryString() . Url::getCurrentQueryStringWithParametersModified([
```
sgiehl, line 142: Shouldn't the complete block within `if (Marketplace::isMarketplaceEnabled())` now be unneeded in the lines above?

diosmosis: 👍 removed it

plugins/CoreUpdater/Controller.php
```diff
@@ -172,6 +173,17 @@ public function oneClickUpdate()
         return $view->render();
     }
 
+    public function oneClickUpdatePartTwo()
+    {
+        Piwik::checkUserHasSuperUserAccess();
```
tsteur, line 178: Wondering what happens if user was just logged out in between this step and the previous?  Eg maybe because of what we noticed in https://github.com/matomo-org/matomo/pull/15755/files#r405990135 ?

I guess otherwise it's pretty much unlikely that a cookie would expire just between those two requests as we extend the cookie on every request by 30 min AFAIK and previous step shouldn't take more than 30min.

I suppose we would notice such an issue in the one click updater test so things should be all good 👍

diosmosis: Definitely want to test w/ the release candidate, regardless.

sgiehl:
@diosmosis code looks good now I guess. Shall we merge, or do you want to do some more testing before?

diosmosis:
Needs to be merged and in an RC before we can do a final test, so I'll just merge it.

PR #15773, opened 2020-04-03
Notifications with the same id might not be shown at all

sgiehl:
When trying to fix the tests for #15736 I wondered why the `UnprocessedSegmentTest` didn't show the notification anymore. Reproducing that locally showed that, due to the additional report on the tested page, the notification is contained two times. When loading the page directly (not by clicking the menu), the reports seems to be rendered on the same time, causing both notifications to remove each other.

Not sure if the applied fix is the best option, but wasn't sure how to prevent the removal in another way...

### DIFF
--- a/plugins/CoreHome/angularjs/notification/notification.directive.js
+++ b/plugins/CoreHome/angularjs/notification/notification.directive.js
@@ -87,8 +87,9 @@
                 function closeExistingNotificationHavingSameIdIfNeeded(id, notificationElement) {
                     // TODO: instead of doing a global query for notification, there should be a notification-container
                     //       directive that manages notifications.
+                    var notificationStillExists = !!notificationElement.parents('body').length;
                     var existingNode = angular.element('[notification-id=' + id + ']').not(notificationElement);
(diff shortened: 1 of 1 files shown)


diosmosis (approved):
Is this a problem on 3.x-dev? If so would probably want to fix there (cc @mattab). If not, good to merge.

sgiehl:
yes. should be on 3.x-dev as well.

mattab:
@sgiehl agreed with @diosmosis, be great to apply to 3.x-dev as well :+1:

tsteur:
Not sure how often that actually happens in 3.x-dev haven't seen any reports etc unless I'm mistaken so might not need to make that effort for 3.x-dev and rather focus on 4.x

sgiehl:
@tsteur for sure there was no issue yet, as it's hard to know that a notification isn't shown where it should be. That affects for example the "unprocessed segment" notification, that is shown when a segment has been changed, but wasn't reprocessed by the cron archiver yet. Which might also happen on cloud I guess. Isn't hard to move that to 3.x-dev branch. Can do that tomorrow...

PR #15774, opened 2020-04-03
Track seconds instead of days for "days since" dimensions

### LINKED ISSUE #14094: Change "days since last ..." dimensions to save second duration instead of days and compute in PHP
Saving days means cutting off the time of day, which makes the information less accurate. This prevents using, eg, visitor_days_since_first to find the exact time of a visitor's first visit.

Changing this would mean updating every row in the log table so we should likely do it in Matomo 4.

When changing this we can also get rid of the `_idts` query parameter in the tracker. It is possible to compute this in PHP without ever having to issue a SQL query.

After we search for an existing visitor we would:

1. if new visitor, set time_since_last_visit = 0
2. if existing visitor, set time_since_last_visit = last_visit.time_since_first_visit + (this_visit.visit_first_action_time - last_visit.visit_last_action_time)

This would allow computing this value for every type of tracking.

Refs #14125

diosmosis:
Fix #14094


core/Updates/4.0.0-b1.php
```diff
@@ -109,6 +111,18 @@ public function getMigrations(Updater $updater)
         // remove old options
         $migrations[] = $this->migration->db->sql('DELETE FROM `' . Common::prefixTable('option') . '` WHERE option_name IN ("geoip.updater_period", "geoip.loc_db_url", "geoip.isp_db_url", "geoip.org_db_url")');
 
+        // replace days to ... dimensions w/ segments
+        foreach (['log_visit', 'log_conversion'] as $table) {
```
tsteur, line 115: btw we should do this all in one query and not in multiple queries so we group all log_visit and all log_conversion update queries.

plugins/CoreHome/Columns/VisitorSecondsSinceOrder.php
```diff
@@ -0,0 +1,77 @@
...
+use Piwik\Plugin\Segment;
+use Piwik\Tracker\Action;
+use Piwik\Tracker\Request;
+use Piwik\Tracker\Visitor;
+
+class VisitorSecondsSinceOrder extends VisitDimension
+{
+    protected $columnName = 'visitor_seconds_since_order';
+    protected $columnType = 'INT(11) UNSIGNED NULL';
+    protected $segmentName = 'secondsSinceLastEcommerceOrder';
```
tsteur, line 22: btw will you also add all the previous segment names again for days to not break API? Ideally we'd even keep the old dimensions (the old classes) so custom reports wouldn't break either but not sure if possible.

tsteur: Seeing the old segment name is below 👍

mattab:
If custom reports are using one of these dimensions we now remove, can you test it and check what will happen? will the core:archive fail, or will the "bogus" custom report  be skipped and disabled automatically or so?

core/Updates/4.0.0-b1.php
diosmosis: one UPDATE? sure, makes sense

diosmosis:
@mattab 👍 if possible will try to make it work in that case

diosmosis:
@mattab it looks like if one of those dimensions is used, it is ignored and effectively removed from the list. I was hoping to be able to add the dimension back as a view on the seconds one, but it doesn't seem like it's possible to have a dimension that is a "view" of another. I also noticed w/o it, the days dimension disappears and only the seconds one is there. This seems like it would be a problem for most users since seconds is harder to use when they want days. @tsteur can you think of a way around this? Is there a way to add a dimension class w/o it being added to a table but still having an associated column so customreports can aggregate on it?

tsteur:
@diosmosis Custom reports uses as unique ID the dimension id from `Dimension::getId()` which is by default the `PluginName.DimensionClassName`.  Eg `Goals.DaysToConversion`. 

So it could help to recreate the old dimension classes. The SQL segment in that class would be something like 
```
class DaysSince...Dimension {
    protected $type = self::TYPE_NUMBER;
    protected $sqlSegment = 'ROUND(log_visit.days_since_last_visit / 86400)';
    protected $nameSingular = 'General_DaysSinceFirstVisit';
}
```

you could use the same class to add back the previous segments which you are currently adding in `addSegments` manually.

```
class VisitorDaysSinceFirst {
    protected $columnName = 'visitor_days_since_first';
    protected $segmentName = 'daysSinceFirstVisit';
    protected $type = self::TYPE_NUMBER;
    protected $sqlSegment = 'ROUND(log_visit.days_since_last_visit / 86400)';
    protected $nameSingular = 'General_DaysSinceFirstVisit';
}
```

 I think this should the trick. The `seconds since first XYZ` we will later need to disable in Custom Reports since they would be generating too many rows and isn't really useful for Custom Reports and  could cause significant performance problems

mattab:
> we will later need to disable in Custom Reports 

Btw if we could deal with this in general that'd be great. For example a couple of times before we had the same issue as here: a user/customer used to have a plugin (for example Provider) and some custom reports used the dimension in that plugin (eg. Provider name) and then the plugin is deactivated and the archiving then fails/errors. The fix then is to delete custom reports in the DB that contain the dimension (which is not best user experience and requires support assistance)

diosmosis:
ready for review

core/DataAccess/LogAggregator.php
```diff
@@ -1134,13 +1138,13 @@ public static function getSelectsFromRangedColumn($column, $ranges, $table, $sel
 
                 $selectAs = "$selectColumnPrefix$lowerBound-$upperBound";
 
-                $selects[] = "sum(case when $table.$column between $lowerBound and $upperBound $extraCondition" .
+                $selects[] = "sum(case when $column between $lowerBound and $upperBound $extraCondition" .
```
tsteur, line 1141: do you remember why $table was removed?

core/Tracker/Visit.php
```diff
@@ -179,6 +184,7 @@ public function handle()
         }
 
         $isNewVisit = $this->request->getMetadata('CoreHome', 'isNewVisit');
+        $this->previousVisitProperties = new VisitProperties($this->request->getMetadata('CoreHome', 'lastKnownVisit') ?: []);
```
tsteur, line 187: btw did something similar in https://github.com/matomo-org/matomo/pull/15869/files#diff-8a4e24a5a8b6afb6ee880dddb4934c43R114

could actually maybe reuse that logic and have eg `visitProperties->getOriginalProperty($column)` which then looks for a value in `visitInfo[self::KEY_ORIGINAL_VISIT_ROW][$column]` . Or maybe we refactor `self::KEY_ORIGINAL_VISIT_ROW` to use same approach as you implemented if possible as we have eg access to the visitor object here: https://github.com/matomo-org/matomo/pull/15869/files#diff-73996192b2fb6075808b45d6276c3264R72  but would need to tweak https://github.com/matomo-org/matomo/pull/15869/files#diff-8a4e24a5a8b6afb6ee880dddb4934c43R131 to also get the `Visitor` object instead of visit properties. I reckon your solution is likely nicer 👍

core/Updates/1.5-b1.php
```diff
@@ -29,7 +32,12 @@ public function __construct(MigrationFactory $factory)
 
     public function getMigrations(Updater $updater)
     {
-        return array(
+        $tableMetadata = new TableMetadata();
```
tsteur, line 35: are below changes needed? We could otherwise just add them and remove them later again should someone update from an old version?

core/Updates/4.0.0-b1.php
tsteur: @diosmosis are these ones actually needed? I would assume Matomo installs the columns automatically (which is also why you didn't need to add it for new installs to the mysql schema)? 
image

tsteur: oh wait I see it is needed for the update... statement, makes sense... Could technically also record that the component was updated so Matomo won't try to update the columns

image

like `insert into option values option_value = $columnType, option_name = version_log_visit_ visitor_seconds_since_first `

core/Updates/4.0.0-b1.php
```diff
@@ -110,6 +113,34 @@ public function getMigrations(Updater $updater)
         // remove old options
         $migrations[] = $this->migration->db->sql('DELETE FROM `' . Common::prefixTable('option') . '` WHERE option_name IN ("geoip.updater_period", "geoip.loc_db_url", "geoip.isp_db_url", "geoip.org_db_url")');
 
+        // replace days to ... dimensions w/ seconds dimensions
+        foreach (['log_visit', 'log_conversion'] as $table) {
+            $migrations[] = $this->migration->db->addColumn($table, 'visitor_seconds_since_first', VisitorSecondsSinceFirst::COLUMN_TYPE);
+            $migrations[] = $this->migration->db->addColumn($table, 'visitor_seconds_since_order', VisitorSecondsSinceFirst::COLUMN_TYPE);
+        }
+
+        $migrations[] = $this->migration->db->addColumn('log_visit', 'visitor_seconds_since_last', VisitorSecondsSinceLast::COLUMN_TYPE);
```
tsteur, line 122: todo: we may need to test if Matomo will automatically merge this add column with above tables or execute it individually. Otherwise need to move it up into the for loop and potentially use `addColumns()`

core/Updates/4.0.0-b1.php
```diff
@@ -110,6 +113,34 @@ public function getMigrations(Updater $updater)
...
+                . " SET visitor_seconds_since_first = visitor_days_since_first * 86400, 
+                    visitor_seconds_since_order = visitor_days_since_order * 86400,
+                    visitor_seconds_since_last = visitor_days_since_last * 86400");
+        }
+
+        if ($hasDaysColumnInConv) {
+            $migrations[] = $this->migration->db->sql("UPDATE " . Common::prefixTable('log_conversion')
+                . " SET visitor_seconds_since_first = visitor_days_since_first * 86400, 
+                    visitor_seconds_since_order = visitor_days_since_order * 86400");
+        }
```
tsteur, line 142: do I see this right we keep the old columns `visitor_days_since_first` etc?

js/piwik.js
```diff
@@ -3364,10 +3356,7 @@ if (typeof window.Piwik !== 'object') {
 
                 var cookieValue = visitorIdCookieValues.uuid + '.' +
                     visitorIdCookieValues.createTs + '.' +
-                    visitorIdCookieValues.visitCount + '.' +
```
tsteur, line 3367: @diosmosis should we maybe keep BC for existing cookies? Currently, any existing cookie might break as they might have different order eg for the nowTs ?

tsteur (changes requested):
Haven't tested it yet @diosmosis but left some comments to look at.

core/Updates/4.0.0-b1.php
diosmosis: to my understanding deleted dimensions aren't actually removed, so this is ok? otherwise I can delete the column as well.

tsteur: I reckon we delete the columns as there wouldn't be any use for them anymore. Unless I'm not aware of something @mattab maybe knows more.

In https://github.com/matomo-org/matomo/wiki/Matomo-4.0.0-release-notes#downgrading-to-matomo-3 we could add some queries to add back the columns again if needed and we could even migrate the old column values to what they were before basically.

js/piwik.js
mattab: I haven't reviewed the code in detail, but I'm not clear on how we can calculate Visit Count, without storing it in the cookie. I'm probably missing something but how is it possible to keep a count of "visit count" without storing the count in the cookie? (Since our "Window look back" is only 30min by default, we wouldn't be able to lookup past visits)

tsteur: btw we don't just look back 30min AFAIK at least not when visitorId is used. And when configId only is used then we can basically assume anyway it is the first visit. 

visitcount is maybe also quite inaccurate in some cases but mostly works.

diosmosis: > (Since our "Window look back" is only 30min by default, we wouldn't be able to lookup past visits)

this is only for fingerprint matching, not for matching based on idvisitor

diosmosis: > (Since our "Window look back" is only 30min by default, we wouldn't be able to lookup past visits)

this is only for matching by config_id. and since the JS tracker sends a visitor ID, and no other tracker sends the visit count cookie automatically, I don't think this behavior is affected.

diosmosis: i can test this on demo2 via log replay to make sure the data is still accurate

core/Updates/4.0.0-b1.php
diosmosis: Is this the desired approach?

core/Updates/1.5-b1.php
diosmosis: this was for the build I think

core/Updates/4.0.0-b1.php
diosmosis: will do use `addColumns()` just to be safe

core/Updates/4.0.0-b1.php
diosmosis: actually I think the dimensions end up getting added after the main update, so we wouldn't be able to initialize the new dimension values

tsteur: yes that's what I figured too. It's needed to add them here already. Be just good to additionally record the component version so once the DB update completed it won't try to change the type of these columns but it might not be too important since it would get the same type anyway and therefore mysql wouldn't do anything.

core/Updates/1.5-b1.php
tsteur: If possible, be great to undo the changes and see if it still works. It's not really expected that a change is needed here.Technically, the updater should basically first add these columns as part of the 1.5 update, then remove them later as part of the 4.x update.

diosmosis:
build js

core/DataAccess/LogAggregator.php
diosmosis: It's handled above, I've made the code clearer.

core/Updates/1.5-b1.php
diosmosis: I see, here's the error: https://travis-ci.org/github/matomo-org/matomo/builds/690764432

there's a test that checks that every update works if we run it again even w/ the latest changes present (`testUpdateWorksAfterPiwikIsAlreadyUpToDate()`). Since we drop the old columns, it fails.

diosmosis:
@tsteur applied review comments. There's still a failing test for the 1.5 update change. Let me know if you think that's a mistake on the tests or not.

core/Tracker/VisitorRecognizer.php
```diff
@@ -128,14 +127,13 @@ public function findKnownVisitor($configId, VisitProperties $visitProperties, Re
         }
     }
 
-    public function removeUnchangedValues(VisitProperties $visitProperties, $visit)
+    public function removeUnchangedValues(VisitProperties $visitProperties, $visit, VisitProperties $originalVisit = null)
```
tsteur, line 130: seems `$visitProperties`  is no longer used.

core/Updates/4.0.0-b1.php
```diff
@@ -117,6 +148,34 @@ public function getMigrations(Updater $updater)
...
+
+        // remove old days_to_... columns
+        $migrations[] = $this->migration->db->dropColumn('log_visit', 'visitor_seconds_since_first');
```
tsteur, line 173: should `visitor_seconds_since_first ` really be removed?

js/piwik.js
tsteur: @diosmosis just checking whether you checked regarding https://github.com/matomo-org/matomo/pull/15774/files#r423417650 because now `nowTs` is in third position vs before existing cookies would have stored the visitCount there? Not sure if we need to check in `getValuesFromVisitorIdCookie` whether the value looks more like a timestamp?

Also I reckon in `getValuesFromVisitorIdCookie()` the array indexes 4 and 6 might not always be defined?
image

The `currentVisitTs` seems no longer used in general?

tsteur (changes requested):
I see now what the problem was re the 1.5 updates @diosmosis . I reckon we could change the update script to simply not specify the third parameter in the `addColumn/addColumns` methods where it currently sets `visitor_days_since_last/visitor_days_since_first`. The order shouldn't matter so we can simply remove the parameter.

Alternatively, I was thinking we could make the `AddColumns` smarter to automatically add it to the end if that specified column no longer exists because the order should never really matter but shouldn't be needed and maybe there are cases where for some reason the order does matter (although I can't imagine that). So for now could just remove the third parameter in those two places.

diosmosis:
build js

tsteur:
@diosmosis is there much left to do in this PR?

diosmosis:
@tsteur this was ready for another review. I must've forgotten to ping you.

tsteur:
Looks good @diosmosis

diosmosis:
build js

diosmosis:
@tsteur / @sgiehl I updated OmniFixture.sql to have a lower version so the update would run on it fully. This, strangely, causes the ip addresses displayed in the visitor log to change to `0.0.0.0`. I looked in the SQL file, and that's the correct value. When I look in the db for the branch and in 4.x-dev, the ip address for the visits shown are `0.0.0.0`. But for some reason on 4.x-dev, a completely different ip address is shown... because of the update being run I guess... not sure what the issue is, but I'm merging this. Can fix the test if anyone knows what's going on.

Issue #15740, opened 2020-03-27
Pagetitle reports "Page Name not defined" although all PageViews do have a title (action_name)


peterbo:
I see this in multiple instances - a row in the Pagetitle report that shows "Page Name not defined". This row has too many visits to be just a "rogue tracking request" somewhere on the website. Also, this row always reports 0 for exit and bounce rate and 00:00:00 for time on site.

This is also reproducible on the demo:
- navigate to the page titlte report
- there you can see the "Page Name not defined" row, mostly in the top 10 to top 15 visited page titles.

This seems to be a regression where another tracking method seems to write an empty action name to the database.

![PageNameNotDefined](https://user-images.githubusercontent.com/1018973/77785717-dcaf4e80-705c-11ea-9dbb-12e5dec96a52.JPG)

Probably is the same as #14512.

tsteur:
@peterbo I suppose it wouldn't be some 404 error page or so? Any chance you could do some query on the DB to find out more about this (like `idaction_name is null`) or so?

peterbo:
Hi @tsteur - presumably it's not a 404/Error page. I investigated a bit further on the issue and it seems, that this is related to Form Analytics.

Form Analytics sometimes seems to bundle/bulk some requests ("POST /piwik.php?{%22requests%22:[%22?fa_vid=[...]") - which seems to be buggy. Every time a bulk request was sent, there was an extra visit recorded in Matomo (compare the timing from the action in the "main" visit in the middle and the action without page title from the "rogue" visit below):

![fa-bulk1](https://user-images.githubusercontent.com/1018973/77863914-1f705280-7226-11ea-95c4-38bde374a16b.jpg)

tsteur:
Thanks @peterbo 

[...]

Issue #14512, opened 2019-06-05
Page Name not defined


daylicron:
Hey there,

I'm running Matomo 3.9.1

I've found in the page title section the following line:
![image](https://user-images.githubusercontent.com/32066194/58937457-c73a0280-8772-11e9-8419-852d8708127d.png)

As a possible explanation I found https://matomo.org/faq/general/faq_74/ . So I grepped trough my access_log an searched for something without page title. As I understand the page title is the first string after the action name, so I looked for something like 
``
[...]action_name=&idsite=[...]
``

But I didn't find what I was looking for. So I tried to find something inside the database, but coldn't find anything.

I assume I'm running into something like https://stackoverflow.com/questions/22869190/piwik-page-name-not-defined

> Detected Site Search keyword 'ban'. NOTE: The Page URL was changed / removed, during the Site Search detection, was 'http://widgets.digitalmediacommunications.com/widget/embed/index/?p=903&k=ban', now is 'http://widgets.digitalmediacommunications.com/widget/embed/index/?p=903' Action is a SITE_SEARCH, Action name = ban, Action URL = http://widgets.digitalmediacommunications.com/widget/embed/index/?p=903
>I'm think that somehow my k=ban is triggering piwik site search feature. So when I turned off site search it worked! I will reviste this issue if I choose to work with site search someday.

Actually I would guess, that there are some page titles inside the database which are NULL, 0, or something like that. When I find them I should be able to trace back the url(s) which are causing this. I just don't know where to look for it.

Can someone help me out?

tsteur:
@daylicron I reckon it is better you ask this question in the forum: https://forum.matomo.org 

We can't really know what is exactly happening here unfortunately. 

You might find some help there. In general we don't answer questions here. If you find a bug and can tell us steps to reproduce, we're happy to look into it or create another issue for it. You could also get in touch with our support if you're not finding any help in the forum: https://matomo.org/support-plans/

PR #15784, opened 2020-04-06
Force POST for bulk requests, fix alwaysUseSendBeacon not respected for bulk requests

tsteur:
fix https://github.com/matomo-org/matomo/issues/15740
fix https://github.com/matomo-org/matomo/issues/14512
Noticed bulk requests weren't working when they had less than 2000 characters. 

While fixing the issue also noticed that the `alwaysUseSendBeacon` flag was not respected for bulk requests so added them as well. Now we're making sure that any bulk request will be POSTed and not used as a URL parameter.

As a result, bulk requests were considered page views with an empty page title instead of the actual requests. Noticed this issue in almost all our websites.


tsteur:
build js

PR #15785, opened 2020-04-07
Make sure to always set JSON_PIWIK to native JSON when possible

tsteur:
Even if `JSON_PIWIK` is an object, when possible assign the native JSON method again. This way it doesn't go into the `else` should `JSON_PIWIK` already be defined because the file was already loaded. Eg if you otherwise load JS tracker twice, the first time it would assign `JSON_PIWIK = window.JSON`, the second time it would go into the `else` where it then assigns JSON manually. Seeing this on a site that I'm debugging and it appears as then `JSON_PIWIK` might be ending up somehow with the value `{}`.

This seems to be because that site has a method `amd` from jquery

image

It is not using the loader logic
image

and it seems there the JSON/JSON3 value never gets assigned to `exports`:
image

Maybe because of some incompatible `amd` or something I'm not sure how this is supposed to work. Anyway, as we're removing this in Matomo 4 anyway I don't want to spend too much time on it and rather fix this issue in that case by assigning `window.JSON` whenever possible so it doesn't even run into that code.


tsteur:
build js

sgiehl (approved):

Issue #15732, opened 2020-03-25
Faster segmented suggested values when browser archiving is disabled


tsteur:
We can also do this as part of 3.13.6 if needed.

We're getting the feedback around slow segment suggested values a lot and it can potentially even take down Matomo when it needs to look at a lot of data over 60 days. I'm meaning the `API.getSuggestedValuesForSegment` method.

We would like the ability that for some segments to optionally define an API method (eg `Events.getAction`).
* should browser archiving be disabled, then we would get the suggested values by looking at the most common labels instead of issuing the live query
* should browser archiving be enabled, then this can be slow to start archiving so we would prefer the live visitor log query instead to avoid slow archiving

We'll need to make sure to have tests for both ways for this API method. Initially we would probably not set this for visitor segments but action segments were possible (page urls, page titles, events, ...). Could set it for some visitor segments though as well.

Ideally, we would not need to do range archives eg `period=range&date=last2months`. Looking at one full month be fine. However, we can't really look at previous month since then recent values would be missing and we can't look at the current month because we might be only a few days into the month. 
* I suggest looking at this year's data be fine when browser archiving is disabled. 
* When the current month is January, then we would maybe need to look at last 2 months (this and previous month) and accept there will be a range archive. We wouldn't want to use `last 2 years` as it would be slower to aggregate since year archives usually have a lot more data. 

**From a technical point of view:**
The idea is, that a `Dimension` class would be able to define an API method eg `protected $segmentSuggestedValuesApi = 'Actions.getPageUrls'`. This would be automatically passed on to the segment class and in the `API.getSuggestedValuesForSegment` we can then check if browser archiving is disabled and if there is an API defined, then we issue the API request instead of a live query. 

We would also specify all needed parameters like the `filter_limit=X, filter_offset=0, segment = ''` in `Request::processRequest`.

As a side effect, should range period be disabled, this would make the suggested values still work for the months from Feb-Dec. Jan wouldn't work. We need to make sure to have a check for this like 
```
        if (!Period\Factory::isPeriodEnabledForAPI('range')) {
            return array();
        }
```
to prevent any exception.

PR #15786, opened 2020-04-07
Faster segmented suggested values when browser archiving is disabled

tsteur:
fix https://github.com/matomo-org/matomo/issues/15732

Also fixes an issue where the wrong segment metadata was created for Page URL/Title not defined and eg segmented visitor log did not work for these.


core/Plugin/Segment.php
```diff
@@ -296,6 +298,32 @@ public function setSuggestedValuesCallback($suggestedValuesCallback)
...
+
+    /**
+     * Set callback which will be executed when user will call for suggested values for segment.
+     *
+     * @param string $suggestedValuesApi
+     */
+    public function setSuggestedValuesApi($suggestedValuesApi)
+    {
+        if (!empty($suggestedValuesApi) && is_string($suggestedValuesApi)) {
+            if (Development::isEnabled() && strpos($suggestedValuesApi, '.get') === false) {
```
sgiehl, line 318: Guess we can ignore that this might throw a warning in development mode if `setSuggestedValuesApi` is called with something else than a string

plugins/API/API.php
```diff
@@ -580,6 +582,59 @@ public function getSuggestedValuesForSegment($segmentName, $idSite)
...
+            if ($now->toString('m') == '01') {
+                if (Rules::isArchivingDisabledFor(array($idSite), new Segment('', array($idSite)), 'range')) {
+                    $date = $now->subYear(1)->toString(); // use previous year data to avoid using range
+                } else {
+                    $period = 'range';
+                    $date = $now->subMonth(1)->toString() . ',' . $now->addDay(1)->toString();
+                }
+            }
+
+            $flat = strpos($segment['suggestedValuesApi'], 'Actions.') === 0;
```
sgiehl, line 606: Wondering if there would be a better solution than hardcoding that for `Actions` only. Maybe we could always flatten the report if the API method has subtables with the same dimension or something like that 🤔

sgiehl (commented):
Left some comments. Otherwise looks good and works 👍 
Should we maybe create a follow up issue to add that feature to other dimensions as well? 🤔

plugins/API/API.php
tsteur: I had it hard coded flattened initially, but then noticed there was an issue with the content dimensions for example where it then would return `content name - content piece` instead of just `content name`. technically, I suppose we could check the metadata and see if there is a subtableApiMethod defined and whether it is the same API method or a different one. Should there be a different api method we could disable flattening. Was hoping though to not needing to do it simply to keep things simple. I suppose it wouldn't be that complex though maybe. Will check later.

tsteur: Updated it @sgiehl

plugins/API/API.php
```diff
@@ -624,8 +626,6 @@ public function getSuggestedValuesForSegment($segmentName, $idSite)
                     );
                     if (!empty($segment) && preg_match('/^' . implode('|',$remove) . '/', $segment)) {
                         $values[] = urldecode(urldecode(str_replace($remove, '', $segment)));
-                    } else {
-                        $values[] = $row->getColumn('label');
```
sgiehl: removing that actually requires a segment metadata to be set, right? should we maybe mention that in the doc block of `setSuggestedValuesApi`

tsteur: done 👍

PR #15796, opened 2020-04-08
Assume plugins that do not define a upper bound for Matomo version re…

### LINKED ISSUE #14342: Restrict open-ended require to the major Matomo version it defined
I just noticed that my Matomo instance had the `IPv6Usage` plugin running.

The plugin doesn't officially support Matomo 3, [hasn't been adopted by anyone](https://forum.matomo.org/t/adopt-a-plugin/26869) and [removed from plugins.matomo.org](https://plugins.matomo.org/IPv6Usage) 

```json
    "require": {
        "piwik": ">=2.12.0"
    }
}
```

But as the plugin has an open-ended `require` it hasn't been deactivated by Matomo.

I think Matomo should at least show a warning to superusers as subtle bugs may be caused by them and maybe this way more people would try to adopt popular plugins that aren't maintained anymore.

sgiehl:
…quirement as incompatible with Matomo 4

fixes #14342

### DIFF
--- a/core/Plugin/Dependency.php
+++ b/core/Plugin/Dependency.php
@@ -43,6 +43,11 @@ class Dependency
 
         foreach ($requires as $name => $requiredVersion) {
             $currentVersion  = $this->getCurrentVersion($name);
+
+            if (in_array(strtolower($name), ['piwik', 'matomo'])) {
+                $requiredVersion = $this->markPluginsWithoutUpperBoundMatomoRequirementAsIncompatible($requiredVersion);
+            }
+
             $missingVersions = $this->getMissingVersions($currentVersion, $requiredVersion);
 
             if (!empty($missingVersions)) {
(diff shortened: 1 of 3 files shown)


diosmosis:
Should we make a note in the CHANGELOG? And should we note this in the developer documentation?

I wonder if it would be helpful (and even possible) to have an error message displaying the reason for the automatic < 4.0.0-b1? I think some developers may get confused, but I'm not too sure (cc @tsteur).

sgiehl:
Actually I guess we will need to update the marketplace and the developer docs. It seems we currently allow anything semver can parse. See https://developer.matomo.org/guides/distributing-your-plugin
ping @tsteur

tsteur:
@sgiehl I've updated the marketplace for new plugin submissions or updates for existing plugins to require a lower and an upper limit. A developer will get an email with the error message and it will automatically suggest a good require string for that plugin when possible. Could you update the docs on developer site to have upper limits?

sgiehl:
Updated the PR and adjusted the tests. Would be ready for another review

diosmosis (approved):
Guess the only thing left is the developer docs tweak?

tsteur:
Merged the docs 👍

PR #15797, opened 2020-04-08
Fix piwik js minified tests

tsteur:
I think after merging https://github.com/matomo-org/matomo/pull/15784#event-3214586635 the JS tests are broken. In the PR the tests were green but that's because we merged another piwikjs PR that was created afterwards that it is broken now


tsteur:
build js

sgiehl:
@tsteur guess you forgot to add something maybe? you actually only added a whitespace? Also the Javascript tests are running 🤔

PR #15800, opened 2020-04-09
purge all old archives regardless of done value

diosmosis:
we only care about the newest usable one

### DIFF
--- a/core/DataAccess/Model.php
+++ b/core/DataAccess/Model.php
@@ -60,9 +60,7 @@ class Model
                        GROUP_CONCAT(idarchive, '.', value ORDER BY ts_archived DESC) as archives
                   FROM `$archiveTable`
                  WHERE name LIKE 'done%'
-                   AND value IN (" . ArchiveWriter::DONE_INVALIDATED . ','
-                                   . ArchiveWriter::DONE_OK . ','
-                                   . ArchiveWriter::DONE_OK_TEMPORARY . ")
+                   AND AND value NOT IN (" . ArchiveWriter::DONE_ERROR . ")
                    AND idsite IN (" . implode(',', $idSites) . ")
                  GROUP BY idsite, date1, date2, period, name";
 
@@ -73,11 +71,10 @@ class Model
             $duplicateArchives = explode(',', $row['archives']);
             $countOfArchives = count($duplicateArchives);
 
-            $firstArchive = array_shift($duplicateArchives);
-            list($firstArchiveId, $firstArchiveValue) = explode('.', $firstArchive);
-
             // if there is more than one archive, the older invalidated ones can be deleted
             if ($countOfArchives > 1) {
+                array_shift($duplicateArchives); // we don't want to delete the latest archive if it is usable
+
                 foreach ($duplicateArchives as $pair) {
                     if (strpos($pair, '.') === false) {
                         $this->logger->info("GROUP_CONCAT cut off the query result, you may have to purge archives again.");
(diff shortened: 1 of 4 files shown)


core/DataAccess/Model.php
```diff
@@ -46,25 +46,20 @@ public function __construct(LoggerInterface $logger = null)
...
         $sql = "SELECT idsite, date1, date2, period, name,
                        GROUP_CONCAT(idarchive, '.', value ORDER BY ts_archived DESC) as archives
                   FROM `$archiveTable`
                  WHERE name LIKE 'done%'
-                   AND value IN (" . ArchiveWriter::DONE_INVALIDATED . ','
-                                   . ArchiveWriter::DONE_OK . ','
-                                   . ArchiveWriter::DONE_OK_TEMPORARY . ")
-                   AND idsite IN (" . implode(',', $idSites) . ")
-                 GROUP BY idsite, date1, date2, period, name";
+                   AND `value` NOT IN (" . ArchiveWriter::DONE_ERROR . ")
```
tsteur, line 61: btw @diosmosis out of curiosity, do you maybe know why we don't delete an ERROR archive?

diosmosis: I think it might be deleted somewhere else... I guess we can get rid of everything?

tsteur: Merged it for now @diosmosis . @mattab do you maybe remember? I reckon it's otherwise not too important

diosmosis: Oh, I know what it's for, this is the SELECT, so we're looking for archives that are usable and deleting everything older. If we SELECT for DONE_ERROR archives, then the latest might be DONE_ERROR and we end up deleting the usable archives. We can include DONE_ERROR here but we'd have to a little extra processing below.

I think they are currently deleted in another task via `Model::getTemporaryArchivesOlderThan()`.

tsteur: makes sense, thanks 👍

tsteur:
@diosmosis I just ran the task for us and I think it didn't clean up anything maybe because there are no reports marked as to be invalidated

image

and it seems it would only remove the ones that were invalidated 

image

Not sure this is maybe still an issue? But I suppose then the only solution would be to iterate over all archive tables for this which we maybe don't want as it could be quite resource intensive?

Not sure we're doing it already. Maybe it would help to invalidate all previous archives for the same site, date & period when we write a new OK archive?

diosmosis:
@tsteur The fix is to run the command manually, otherwise we'd have to somehow add one date per existing archive table to ArchivesToPurgeDistributedList. Could do that in an update?

diosmosis:
> Maybe it would help to invalidate all previous archives for the same site, date & period when we write a new OK archive?

I guess in the purge task, we can always do today/yesterday, that would solve the biggest source of dupes.

tsteur:
That sounds good to do it in an update 👍  If I see this right in our DB it is only march and april that seems impacted by this so would only need to add two dates for these months maybe?

Currently executing this command to potentially fix it for now for us:

```
console core:purge-old-archive-data "2020-04-15" "2020-03-15" --exclude-ranges --skip-optimize-tables --include-year-archives  '
```

tsteur:
> I guess in the purge task, we can always do today/yesterday, that would solve the biggest source of dupes.

👍  that should work and be quite simple. I guess today would be enough. Yesterday could be done maybe if it belongs to a different month? As `purgeInvalidatedArchivesFrom` looks at the entire month anyway

diosmosis:
> I guess today would be enough. Yesterday could be done maybe if it belongs to a different month?

:+1:

PR #15805, opened 2020-04-11
Fix comparing segments doesn't work with condition startsWith

john357smith:
In case of comparing of two or more segments with condition "startsWith" an error will occur because parameter 'compareSegments' is sanitized resulting to removal of '^' chararcter from start of the parameter value.


sgiehl (approved):
makes sense :+1:

sgiehl:
@mattab does it make sense to backport that to 3.x-dev branch so it will already be fixed in upcoming release?

mattab:
Sounds good to backport @sgiehl since it looks easy enough

PR #15808, opened 2020-04-13
Add more parameters to the computeNbUnique event

tsteur:
Needed for Roll Up Reporting to optionally disable unique visitors when a segment is applied.

Could add it to developer changelog but maybe not even needed as we only just added this event and wouldn't be used by anyone. Looks like we didn't add the event name last time to the changelog. Happy to add it though.


sgiehl (approved):
@tsteur Looks good. Guess makes sense adding the event to changelog if it should be something developers can use. Otherwise we should mark the event as @internal maybe

tsteur:
Done @sgiehl

PR #15812, opened 2020-04-14
Rename interaction_position to pageview_position

### LINKED ISSUE #12124: Rename `log_link_visit_action.interaction_position` to eg `pageview_position`
As it is incremented only on pageviews but not eg on events. At the same time we may want to think about adding a new column `interaction_position` that is also incremented on events etc but not needed for now.

sgiehl:
fixes #12124

### DIFF
--- a/plugins/Actions/Columns/VisitTotalInteractions.php
+++ b/plugins/Actions/Columns/VisitTotalInteractions.php
@@ -83,10 +83,7 @@ class VisitTotalInteractions extends VisitDimension
     {
         $position = self::getCurrentInteractionPosition($request);
 
-        $position = $position + 1;
-
-        // Remove this in Piwik 4
-        return min($position, 32765);
+        return ($position + 1);
     }
 
     /**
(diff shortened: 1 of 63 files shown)


plugins/Actions/Columns/VisitTotalInteractions.php
```diff
@@ -83,10 +83,7 @@ public static function getNextInteractionPosition($request)
     {
         $position = self::getCurrentInteractionPosition($request);
 
-        $position = $position + 1;
-
-        // Remove this in Piwik 4
-        return min($position, 32765);
+        return ($position + 1);
```
diosmosis, line 86: minor: are the parentheses needed?

diosmosis (approved):
LGTM, a new issue might be needed for `interaction_position`

plugins/Actions/Columns/VisitTotalInteractions.php
sgiehl: don't think so
```suggestion
        return $position + 1;
```

PR #15814, opened 2020-04-15
Make sure to set cookie name before loading any cookie

### LINKED ISSUE #14978: Visitor cookie is set before configuration.
The "_pk_id" cookie is set by the `setSiteId()` method before the `setVisitorCookieTimeout()` and `setCookieNamePrefix()` are executed. As a result it is not possible to change the prefix and lifetime for that cookie. The other cookies change with the the prefix and you will get a second id cookie sometimes. 

Tracking Code example:

```
<!-- Matomo -->
  var _paq = window._paq || [];
  _paq.push(['setCookieNamePrefix', 'stats']);
  _paq.push(['setVisitorCookieTimeout', '0']);
  _paq.push(['trackPageView']);
  (function() {
    var u="https://matomo-dev/";
    _paq.push(['setTrackerUrl', u+'matomo.php']);
    _paq.push(['setSiteId', '1']);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
  })();
</script>
<!-- End Matomo Code -->
```

Call Stack:

matomo-visitor-cookie-set-before-prefix

tsteur:
fix https://github.com/matomo-org/matomo/issues/14978

As described in the issue we need to make sure that `setCookieNamePrefix` is executed before `setSiteId`.

Because the method is now executed early, we now longer really need to get the custom variables from cookie there because the method is executed so early that it will be loaded correctly later as part of `loadCustomVariables`. Nonetheless loading them if it was attempted to load them earlier eg when there is a manually created tracker instance that isn't using `_paq.push` but `tracker.setCookieNamePrefix()`


js/piwik.js
```diff
@@ -3102,7 +3102,7 @@ if (typeof window.Piwik !== 'object') {
                 var cookieName = getCookieName('cvar'),
                     cookie = getCookie(cookieName);
 
-                if (cookie.length) {
+                if (cookie && cookie.length) {
```
tsteur, line 3105: `cookie` might be `0` so better to check if it is set

sgiehl (approved):

sgiehl:
build js

PR #15815, opened 2020-04-15
Add test to ensure there are no plugin updates for core plugins

### LINKED ISSUE #11221: when a core plugin was disabled during piwik core update, the core plugin's update was not actually executed?
Error during creating goals on site: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'description' in 'field list'

Piwik version 3.0.1

tsteur:
fix https://github.com/matomo-org/matomo/issues/11221

see https://github.com/matomo-org/matomo/issues/11221#issuecomment-613792193 and below.

Basically things are working except if someone was to manually remove a plugin from `PluginInstalled`. We basically don't want to define plugin updates for core plugins anyway and rather want to ensure to have them in `core/Updates`.

This test worked for me. I was basically copying a test into Goals plugin and then the test failed

image

### DIFF
--- a/tests/PHPUnit/Integration/ReleaseCheckListTest.php
+++ b/tests/PHPUnit/Integration/ReleaseCheckListTest.php
@@ -12,7 +12,6 @@ use Exception;
 use Piwik\Config;
 use Piwik\Container\StaticContainer;
 use Piwik\Filesystem;
-use Piwik\Http;
 use Matomo\Ini\IniReader;
 use Piwik\Plugin\Manager;
 use Piwik\Tests\Framework\TestCase\SystemTestCase;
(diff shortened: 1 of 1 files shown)


tests/PHPUnit/Integration/ReleaseCheckListTest.php
```diff
@@ -617,6 +616,47 @@ public function test_TotalPiwikFilesSize_isWithinReasonnableSize()
...
+                foreach ($files as $file) {
+                    $fileVersion = basename($file, '.php');
+                    if (
+                        version_compare('3.13.0', $fileVersion) != 1
+                    ) {
+                        // since matomo 3.13.0 we basically don't want to see any plugin specific updates for core plugins
+                        // they should be instead in core/Updates/*
+                        $pluginsWithUnexpectedUpdates[$pluginName] = $file;
+
+                        var_export($pluginName . $file);
```
sgiehl, line 645: Guess this debugging code could be removed.

tests/PHPUnit/Integration/ReleaseCheckListTest.php
```diff
@@ -617,6 +616,47 @@ public function test_TotalPiwikFilesSize_isWithinReasonnableSize()
...
+    public function test_noUpdatesInCorePlugins()
+    {
+        $manager = Manager::getInstance();
+        $plugins = $manager->loadAllPluginsAndGetTheirInfo();
+
+        $pluginsWithUnexpectedUpdates = array();
+        $pluginsWithUpdates = array();
+        $numTestedCorePlugins = 0;
+        foreach ($plugins as $pluginName => $info) {
+            if ($manager->isPluginBundledWithCore($pluginName)) {
```
sgiehl, line 628: Wondering if that might start failing if we put updates into TagManager. I guess we still want to have them separate in that case, right?

sgiehl (commented):
Left some comments. Otherwise looks good to merge

tests/PHPUnit/Integration/ReleaseCheckListTest.php
tsteur: removed

tests/PHPUnit/Integration/ReleaseCheckListTest.php
tsteur: Right, changed that 👍

tests/PHPUnit/Integration/ReleaseCheckListTest.php
```diff
@@ -617,6 +616,49 @@ public function test_TotalPiwikFilesSize_isWithinReasonnableSize()
...
+    {
+        $manager = Manager::getInstance();
+        $plugins = $manager->loadAllPluginsAndGetTheirInfo();
+
+        $pluginsWithUnexpectedUpdates = array();
+        $pluginsWithUpdates = array();
+        $numTestedCorePlugins = 0;
+
+        // eg these plugins are managed in a submodule and they are installing all tables/columns as part of their plugin install method etc.
+        $corePluginsThatAreIndependent = array('TagManager');
```
tsteur, line 629: we can add later more plugins when/if needed

PR #15818, opened 2020-04-15
Prevent possible notice and hiding actual error in debug trace

tsteur:
refs https://forum.matomo.org/t/empty-or-invalid-response-php-notice-undefined-index-class-when-archiving/36908/2

Not really needed for 3.x

### DIFF
--- a/plugins/Monolog/Formatter/LineMessageFormatter.php
+++ b/plugins/Monolog/Formatter/LineMessageFormatter.php
@@ -81,12 +81,15 @@ class LineMessageFormatter implements FormatterInterface
             }
 
             $level = $trace[$i];
+            $levelTrace = '';
             if (isset($level['file'], $level['line'])) {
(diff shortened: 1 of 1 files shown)


plugins/Monolog/Formatter/LineMessageFormatter.php
```diff
@@ -81,12 +81,15 @@ private static function formatTrace(array $trace, $numLevels = 10)
             }
 
             $level = $trace[$i];
+            $levelTrace = '';
             if (isset($level['file'], $level['line'])) {
                 $levelTrace = '#' . $i . (str_replace(PIWIK_DOCUMENT_ROOT, '', $level['file'])) . '(' . $level['line'] . ')';
-            } else {
+            } elseif (isset($level['class']) && isset($level['type']) && isset($level['function'])) {
```
sgiehl, line 87: could be simplified to
```suggestion
            } elseif (isset($level['class'], $level['type'], $level['function'])) {
```

tsteur: sure that will work too (might not have much benefit though)

tsteur:
Applied patch

Issue #15117, opened 2019-11-06
Refactor cron archiving for simplicity


diosmosis:
Now that we invalidate archives when a visit is tracked, we have a couple more opportunities for refactoring:

1. during invalidation we can create a new archive w/ done flag value = DONE_INVALIDATED if the archives for the dates do not exist
2. in CronArchive, instead of pulling individual sites and checking if there have been visits for those sites, just pull individual invalidated archives, set the done value to DONE_IN_PROGRESS, and initiate archiving for that one archive.

This should allow splitting archiving of large websites across separate processes and simplify the code quite a bit.

Some things to keep in mind:
* by default we would look at the table for today's date and for everything in the invalidated sites list instead of iterating over every table, though there could be an option to do that.
* we should not even start archiving if raw data has been deleted for an invalidated archive.

tsteur:
Maybe also something to consider:

* https://github.com/matomo-org/matomo/issues/11974
* https://github.com/matomo-org/matomo/issues/15127
* https://github.com/matomo-org/matomo/pull/14937

diosmosis:
Basic implementation would be:

```
core:archive command:
- while true:
   - invalidated archives = getInvalidatedArchives()
   - if no invalidated archives, exit
   - for 1..# of parallel processes to create
      - for each invalidated archive:
        - create archive w/ DONE_IN_PROGRESS for archive if one not already there (removing from invalidated archives in the process)
        - if successful, break, otherwise keep going
      - add archive to list of archive jobs to launch
   - initiate archiving for each DONE_IN_PROGRESS archive via CliMulti

getInvalidatedArchives():
   - if invalidated archives in cache, return it
   - otherwise, loop through every archive and query for invalidated archives
   - set result in cache w/ TTL of 1 hour
```

thoughts @tsteur ?

tsteur:
@diosmosis Wondering... When we create a `done` archive, could we in that moment directly delete any previously existing archive immediately during the archive process?

When we start an archive
* Set start flag (probably what you mean with DONE_IN_PROGRESS)
* archive...
* set done flag & delete previous archive immediately (not sure it's doable but would maybe simplify things and there be never more than one archive for a date...?).

When a tracking request comes in:
* We invalidate that archive so we know next time we need to archive that period again

[...]

PR #15821, opened 2020-04-16
merge 3.x to 4.x

diosmosis:


diosmosis:
@tsteur the ArchiveCronTest failure on 3.x is back now, going to merge this, if it sticks around after #15117 then will look into fixing it.

sgiehl:
@diosmosis  it might be better to push merges from 3.x-dev to 4.x-dev directly to the 4.x-dev branch, without using a PR. The problem with the PR is, that with our merge strategy all commits are squashed into one. That means, the next time some tries to merge 3.x-dev to 4.x-dev all commits in this PR will be tried to merge again, as they are not included in 4.x-dev. That doesn't happen when pushing the merge directly or if we would allow another merge strategy.

diosmosis:
@sgiehl 👍 didn't think of that. Think it would be better to revert this and merge on the CLI manually?

sgiehl:
not needed. Another merge from 3.x-dev to 4.x-dev should be enough as all commits should be tried to merge again, even if the changes have already been merged...

PR #15828, opened 2020-04-19
add a more helpful info to the default matomo.php output

Findus23:
As two people contacted me in a few days because they seem to have mixed up matomo.php and index.php, I propose to add a more direct help to the default output.

### DIFF
--- a/core/Tracker/Response.php
+++ b/core/Tracker/Response.php
@@ -79,7 +79,7 @@ class Response
                 Common::sendResponseCode(400);
             }
             Common::printDebug("Empty request => Matomo page");
-            echo "This resource is part of Matomo. Keep full control of your data with the leading free and open source <a href='https://matomo.org' target='_blank' rel='noopener noreferrer nofollow'>web analytics & conversion optimisation platform</a>.<br>";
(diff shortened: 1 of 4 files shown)


core/Tracker/Response.php
```diff
@@ -79,7 +79,8 @@ public function outputResponse(Tracker $tracker)
                 Common::sendResponseCode(400);
             }
             Common::printDebug("Empty request => Matomo page");
-            echo "This resource is part of Matomo. Keep full control of your data with the leading free and open source <a href='https://matomo.org' target='_blank' rel='noopener noreferrer nofollow'>web analytics & conversion optimisation platform</a>.";
+            echo "This resource is part of Matomo. Keep full control of your data with the leading free and open source <a href='https://matomo.org' target='_blank' rel='noopener noreferrer nofollow'>web analytics & conversion optimisation platform</a>.<br>\n";
+            echo "This file is the endpoint for the Matomo tracking API. If you want to access the Matomo UI or use the Reporting API, please use <a href='index.php'>index.php</a> instead.";
```
tsteur, line 83: @Findus23 I suppose noreferrer nofollow might not be needed in this case since it goes to the same install and there is a noindex?

tsteur: ping @Findus23

Findus23: Sorry, missed the first mail. (need to write a mail filter that marks mentions separately)

I don't think they are needed as they just link to the Matomo main page which would most likely redirect to the login (for random visitors) which is not indexed.

PR #15832, opened 2020-04-20
Always purge today & yesterday and add update for 3.13.5

### LINKED ISSUE #15822: Always purge today and yesterday when purging archives
Yesterday be only needed if yesterday is for a different month.

Also we discussed to have an update script for 3.13.5 that adds this and last months to be purged.
See discussion in https://github.com/matomo-org/matomo/pull/15800#issuecomment-614292176

@diosmosis fyi created this issue so we don't forget

diosmosis:
Fix #15822

### DIFF
--- a/core/Updates/3.13.5-rc1.php
+++ b/core/Updates/3.13.5-rc1.php
@@ -26,15 +26,16 @@ class Updates_3_13_5_rc1 extends PiwikUpdates
     {
         $archivesToPurge = new ArchivesToPurgeDistributedList();
 
-        $startOfProblem = Date::factory('2020-03-01 00:00:00');
+        $startOfProblem = Date::factory('2020-01-01 00:00:00');
 
         $archiveTables = ArchiveTableCreator::getTablesArchivesInstalled(ArchiveTableCreator::NUMERIC_TABLE);
         foreach ($archiveTables as $table) {
             $date = ArchiveTableCreator::getDateFromTableName($table);
             list($year, $month) = explode('_', $date);
 
+            // only add if the table is for jan 2020 or above since tables w/ that date will be most affected
             $dateObj = Date::factory("$year-$month-01 00:00:00");
-            if ($dateObj->isEarlier($startOfProblem)) { // only add if the table is for march 2020 or above since that is when the problem appeared
+            if ($dateObj->isEarlier($startOfProblem)) {
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


Findus23:
Do I understand it correctly that the update script does something similar to this?
```bash
sudo -u www-data php7.3 console core:purge-old-archive-data all
```

Because that took over 10 minutes with my small DB and fast SSD, which means the update will time out for nearly all Matomo users (unless they do it from the cli, which nearly no one does).

diosmosis:
@Findus23  The update adds the dates to the list that is used during scheduled task execution. So the next scheduled task run will run the equivalent of that command (though I think the table optimization is in it's own task).

tsteur:
@diosmosis I reckon it be enough to add this in the update script only for the months since march 2020 since that's when it was an issue?

diosmosis:
@tsteur 👍 will make that change

diosmosis:
@tsteur updated

core/Updates/3.13.5-rc1.php
```diff
@@ -0,0 +1,44 @@
...
+
+        $startOfProblem = Date::factory('2020-03-01 00:00:00');
+
+        $archiveTables = ArchiveTableCreator::getTablesArchivesInstalled(ArchiveTableCreator::NUMERIC_TABLE);
+        foreach ($archiveTables as $table) {
+            $date = ArchiveTableCreator::getDateFromTableName($table);
+            list($year, $month) = explode('_', $date);
+
+            $dateObj = Date::factory("$year-$month-01 00:00:00");
+            if ($dateObj->isEarlier($startOfProblem)) { // only add if the table is for march 2020 or above since that is when the problem appeared
```
tsteur, line 37: I haven't checked it yet... but this would only run it for march if it was executed eg on April 23rd 2020 right? Although I suppose that's fine since we'd add today anyway on every purge and April would be deleted that way 👍

diosmosis: I'm not sure what you mean, but it will run it if it finds an archive table for march or later. Though actually, this won't work because we also have to add the year table. I'll just add it if the year is 2020 or greater.

tsteur: I meant "2020-04-01 00:00:00" would be earlier than 2020-04-23 and therefore it wouldn't do it for April but it's not important since it would do it automatically anyway when the task is executed next

diosmosis: The dates used should always be `-01 00:00:00` in this code, i'm not sure why the execution date of the update matters.

Issue #6535, opened 2014-10-28
Remove all methods out of plugins/*/API.php that have an @ignore annotation


tsteur:
Makes no sense by definition. In API.php should be only methods that are callable via the HTTP API. 

Instead those methods should be moved in different classes. Such as a Model or something more meaningful.

sgiehl:
Just to be sure. That would currently include the following API methods:

* [ ] API.getSVGLogoUrl
* [ ] API.hasSVGLogo
* [ ] Annotations.getDateRangeForPeriod
* [ ] CustomVariables.getReservedCustomVariableKeys
* [ ] DBStats.getArchiveTableYear
* [ ] Events.getActionToLoadSubtables
* [ ] Events.getDefaultSecondaryDimension
* [ ] Events.getSecondaryDimensions
* [ ] Goals.getConversions
* [ ] Goals.getNbVisitsConverted
* [ ] Goals.getConversionRate
* [ ] Goals.getRevenue
* [ ] LanguagesManager.getPluginTranslationsForLanguage
* [ ] Live.getLastVisitsForVisitor
* [ ] MobileMessaging.sanitizePhoneNumber
* [ ] MobileMessaging
* [ ] MobileMessaging
* [ ] MobileMessaging

tsteur:
I'm now wondering how important it actually is. These methods can't be called through the API right? It doesn't seem all that important and I reckon we could close the issue if they can't be called directly.

sgiehl:
They can't be called directly. Nevertheless it might not make any sense to have public methods in an API class that should not be callable. But if it's not important right now, I'll stop working on it for now.
Feel free to close the issue or move it to a later milestone...

tsteur:
Agreed it doesn't make sense but it's also not any important. It really doesn't make any difference at the end of the day. Will close it.

PR #15838, opened 2020-04-21
removes VisitsSummary.getColumns

sgiehl:
refs #6535

### DIFF
--- a/plugins/VisitFrequency/API.php
+++ b/plugins/VisitFrequency/API.php
@@ -120,7 +120,7 @@ class API extends \Piwik\Plugin\API
     protected function prefixColumns($table, $period, $suffix)
     {
         $rename = array();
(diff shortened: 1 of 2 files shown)


plugins/VisitFrequency/API.php
```diff
@@ -120,7 +120,7 @@ protected function unprefixColumns(array $requestedColumns, $suffix)
     protected function prefixColumns($table, $period, $suffix)
     {
         $rename = array();
-        foreach (APIVisitsSummary::getInstance()->getColumns($period) as $oldColumn) {
+        foreach ($table->getColumns() as $oldColumn) {
```
sgiehl, line 123: simply renaming all columns should have the same effect, but makes the `getColumns` method obsolete

PR #15840, opened 2020-04-21
moves Annotations.getDateRangeForPeriod to plugin class

sgiehl:
refs #6535


diosmosis:
some tests are failing, looks like there are some missing `use`s

sgiehl:
@diosmosis should be fixed now

PR #15844, opened 2020-04-22
Run UI tests of submodules with core build if possible

### LINKED ISSUE #9253: UI tests on Travis do not execute UI tests of submodules
When executing UI tests on Travis we do not execute the UI tests that are defined in the submodule plugins. On the contrary, we do execute unit, integration and system tests of these plugins. I think we should be consistent and also include UI tests of plugins that are added in `piwik/piwik` as a submodule.  

When working on these plugins it makes it easier to see if the UI tests of this plugin actually work and when working on Piwik we can see if we break any of these plugins. The ones that are added are some kind of special plugins that are kinda core but we want to provide updates separately etc.

Of course we also have Travis jobs in the plugins that are submodules. The problem is that they run against beta versions of Piwik. This means to make tests work for a plugin we need to release a new beta after each change in Piwik, and then change the `PIWIK_LATEST_STABLE_TEST_TARGET` variable in `.travis.yml` of the plugin.

I am currently developing the `CustomDimensions` plugin and I cannot let Travis test the UI tests. To test the UI tests in `CustomDimensions` Travis we would need to release a new beta version of Piwik. To release a new beta that can be used by this plugin we would need to merge a PR in `piwik/piwik` that includes some changes in Piwik core that are needed to make this plugin work. However, we cannot really merge since we don't know whether the UI tests work etc. 

This issue is to discuss whether we should include these plugins in UI tests or not. And if want to include them, we should probably remove the `--core` flag in https://github.com/piwik/travis-scripts/blob/ab737e8de918bfc06c1fc5234071f60e3be51a60/travis.sh#L76

sgiehl:
Note: Some plugins have special needs in travis setup. Those might not be runable with core build, as they might fail. This currently affects LoginLdap & TagManager

fixes #9253

### DIFF
--- a/tests/lib/screenshot-testing/support/app.js
+++ b/tests/lib/screenshot-testing/support/app.js
@@ -38,8 +38,8 @@ var walk = function (dir, pattern, result) {
 };
 
 var isCorePlugin = function (pathToPlugin) {
-    // if the plugin is a .git checkout, it's not part of core
-    var gitDir = path.join(pathToPlugin, '.git');
+    // skip plugins that have special needs in core build
+    var gitDir = path.join(pathToPlugin, 'tests/travis');
     return !fs.existsSync(gitDir);
 };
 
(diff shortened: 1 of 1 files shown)


tests/lib/screenshot-testing/support/app.js
```diff
@@ -38,8 +38,8 @@ var walk = function (dir, pattern, result) {
 };
 
 var isCorePlugin = function (pathToPlugin) {
-    // if the plugin is a .git checkout, it's not part of core
-    var gitDir = path.join(pathToPlugin, '.git');
+    // skip plugins that have special needs in core build
+    var gitDir = path.join(pathToPlugin, 'tests/travis');
```
diosmosis, line 42: The variable & function name could be changed to be more accurate

diosmosis: Also, I think this method is for running the tests locally so modifying it might affect test running locally w/ git checkouts of pro plugins, eg.

PR #15848, opened 2020-04-22
Reactivates some skipped tests

sgiehl:
Various tests are currently skipped on mysqli because of random failures. As that occurred back on PHP 5 on trusty distribution I think it's time to give them another try.
The test ran more than 5 times now on travis and didn't fail. So guess should be fine. But let's have an eye on that and if any of those tests should fail randomly again, we maybe should investigate them instead of skipping them again.


tsteur:
> we maybe should investigate them instead of skipping them again.

Let's create an issue for this if it happens and schedule it. We wouldn't do this during any of the Matomo 4 work and rather skip it again otherwise as the priority is currently Matomo 4.

PR #15854, opened 2020-04-24
Keep last 100 characters of archive error message when error is too long

tsteur:
Currently investigating some archiving issue where I can only see the first 6000 characters of the serialised archive response which looks good. However, I cannot see the any of the last characters so I can't really tell if the full response looks good or what the issue could be potentially. Therefore, changing the logic to also show the last 100 characters of the response. Tested this locally and worked for me. Makes the message technically 5 characters longer because of ` ... ` but that should be fine.


sgiehl (approved):
Haven't tested much, but code looks good in general

Issue #13655, opened 2018-10-29
Offering the possibility to disable the fingerprint feature


NanorPiwik:
Hi Matomo team,

According to https://matomo.org/faq/general/faq_21418/ we have a feature to enable the fingerprinting across several websites. However we don't have one in order to disable it.
That's a concern in terms of privacy.

clintonblackmore:
Hi. I raised this issue.

I'm starting a business making apps to sell to kids and hired a lawyer to help with my privacy policy (and believe me, I've read over many privacy policies lately). I wondered if I needed to put a notice in my privacy policy about fingerprinting, such as:

> Fingerprinting
> 
> Our website takes a fingerprint of your system’s setup, to imperfectly distinguish between visitors to our site.

My lawyer sent me this page to look at: [The GDPR and Browser Fingerprinting: How It Changes the Game for the Sneakiest Web Trackers](https://www.eff.org/deeplinks/2018/06/gdpr-and-browser-fingerprinting-how-it-changes-game-sneakiest-web-trackers).

I know that Matomo is serious about privacy. Lacking any way to disable the fingerprinting, I've decided to forgo it for tracking hits to my website (but I am planning to use it to track analytics for my games, as I feel that Unity Analytics processes way too much information.)

Thank you for considering this issue.

mattab:
We're wondering whether fingerprinting in Matomo could be considered personal data... Knowing that our fingerprint is quite imprecise and sometimes recognises multiple people as one: https://github.com/matomo-org/matomo/issues/13528

From https://www.eff.org/deeplinks/2018/06/gdpr-and-browser-fingerprinting-how-it-changes-game-sneakiest-web-trackers

> You’ll struggle to find fingerprinting explicitly mentioned in the GDPR—but that’s because the EU has learned from earlier data protection laws and the current ePrivacy Directive to remain technologically neutral.
> Apart from non-binding recitals (like Recital 30, discussing cookies), the GDPR avoids calling out specific technologies or giving exhaustive lists and examples. Instead, it provides general rules that the drafters felt should be neutral, flexible, and keep up with technological development beyond fingerprinting and cookies. 
> [...]
> Thus, when several information elements are combined (especially unique identifiers such as your set of system fonts) **across websites** (e.g. for the purposes of behavioral advertising), fingerprinting constitutes the processing of personal data and must comply with GDPR.

[...]

Issue #12737, opened 2018-04-18
Enable "Super Privacy" mode to not track any personal data, aka "I do not want to be bothered with GDPR" 


mattab:
We are thinking that it might be very useful to have a one click "I don't want to be bothered by GDPR" feature that disables custom variables, custom dimensions, userId, geolocation, anonymizes IP, ... the only thing left be page URLs/titles and referrers which can be ignored likely.

Refs #12600

mattab:
see also our new FAQ: [How do I use Matomo Analytics without consent or cookie banner?](https://matomo.org/faq/new-to-piwik/how-do-i-use-matomo-analytics-without-consent-or-cookie-banner/)

PR #15855, opened 2020-04-24
Anonymise fingerprint

tsteur:

@mattab was quickly working on this 30 min for 3.X. 

The full feature https://github.com/matomo-org/matomo/issues/13655#issuecomment-616892565 could be developed in 4.X. While it's not fully implemented, it would be something though that could already help. The idea is basically to add the current date to the fingerprint, to ignore any visitorId unless it is a userId, and to disable the visitor profile in the visitor log (not in userId feature or API etc which can all be done in Matomo 4)

Not sure if "Disable fingerprint" is the right word since a fingerprint is still needed. Of course there could be also a feature "Disable fingerprint" which generates a new visit on every tracking request. It would basically add the dateTime to the fingerprint as well as some random generated string. That can be useful if people only want to know how often eg a page was viewed etc.

Any thoughts?

### DIFF
--- a/core/Tracker/Settings.php
+++ b/core/Tracker/Settings.php
@@ -8,8 +8,10 @@
  */
 namespace Piwik\Tracker;
 
-use Piwik\Config;
 use Piwik\Container\StaticContainer;
+use Piwik\Date;
+use Piwik\Plugins\PrivacyManager\PrivacyManager;
+use Piwik\Site;
 use Piwik\Tracker;
 use Piwik\DeviceDetector\DeviceDetectorFactory;
 use Piwik\SettingsPiwik;
@@ -118,6 +120,10 @@ class Settings // TODO: merge w/ visitor recognizer or make it it's own service.
 
         if (!$this->isSameFingerprintsAcrossWebsites) {
             $configString .= $request->getIdSite();
+        }
+
+        if (PrivacyManager::shouldAnonymiseFingerprint()) {
+            $configString .= Date::now()->setTimezone(Site::getTimezoneFor($request->getIdSite()))->toString();
         }
 
         $hash = md5($configString, $raw_output = true);
--- a/core/Tracker/VisitorRecognizer.php
+++ b/core/Tracker/VisitorRecognizer.php
@@ -12,6 +12,7 @@ use Piwik\Common;
 use Piwik\EventDispatcher;
 use Piwik\Plugin\Dimension\VisitDimension;
 use Piwik\Plugins\CustomVariables\CustomVariables;
+use Piwik\Plugins\PrivacyManager\PrivacyManager;
 use Piwik\Tracker\Visit\VisitProperties;
 
 /**
@@ -87,7 +88,11 @@ class VisitorRecognizer
 
         $isVisitorIdToLookup = !empty($idVisitor);
 
-        if ($isVisitorIdToLookup) {
+        if (PrivacyManager::shouldAnonymiseFingerprint()) {
+            $isVisitorIdToLookup = false;
+            $idVisitor = false;
+            Common::printDebug("Visitor ID ignored as we anonymise the fingerprint...");
+        } elseif ($isVisitorIdToLookup) {
             $visitProperties->setProperty('idvisitor', $idVisitor);
             Common::printDebug("Matching visitors with: visitorId=" . bin2hex($idVisitor) . " OR configId=" . bin2hex($configId));
         } else {
--- a/plugins/PrivacyManager/templates/privacySettings.twig
+++ b/plugins/PrivacyManager/templates/privacySettings.twig
@@ -56,6 +56,13 @@
              inline-help="{{ 'PrivacyManager_AnonymizeOrderIdNote'|translate|e('html_attr') }}">
         </div>
 
+        <div piwik-field uicontrol="checkbox" name="anonymizeFingerprint"
+             ng-model="anonymizeIp.anonymizeFingerprint"
+             data-title="Anonymize fingerprint"
+             value="{{ anonymizeIP.anonymizeFingerprint }}"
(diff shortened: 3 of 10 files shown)


core/Tracker/Settings.php
```diff
@@ -120,6 +122,10 @@ protected function getConfigHash(Request $request, $os, $browserName, $browserVe
             $configString .= $request->getIdSite();
         }
 
+        if (PrivacyManager::shouldAnonymiseFingerprint()) {
+            $configString .= Date::now()->setTimezone(Site::getTimezoneFor($request->getIdSite()))->toString();
```
tsteur, line 126: todo: generate random string per site per day or so to anonymise it more

plugins/PrivacyManager/templates/privacySettings.twig
```diff
@@ -56,6 +56,13 @@
              inline-help="{{ 'PrivacyManager_AnonymizeOrderIdNote'|translate|e('html_attr') }}">
         </div>
 
+        <div piwik-field uicontrol="checkbox" name="anonymizeFingerprint"
+             ng-model="anonymizeIp.anonymizeFingerprint"
+             data-title="Anonymize fingerprint"
+             value="{{ anonymizeIP.anonymizeFingerprint }}"
+             inline-help="By anonymising the fingerprint, Matomo will still generate a fingerprint. However, this fingerprint will change for each visitor every 24 hours. This means a visitor can no longer be followed over multiple days, no profile can be generated, and an individual cannot be identfied. This setting currently applies to all websites. When enabled, it will ignore any visitorId and as a result, the unique visitor metric will become inaccurate and sometimes different visitors may be falsely identified as the same visitor if the used device is similar. It will partially also disable the visitor profile in the UI. Please note that this feature does not deactivate the UserId feature.">
```
mattab, line 63: Suggested after few minor tweaks:

> By anonymising the fingerprint, Matomo will still generate a fingerprint. However, this fingerprint will change for each visitor every 24 hours. This means a visitor can no longer be followed over multiple days, no profile can be generated, and an individual cannot be identified. This setting currently applies to all websites. When enabled, any Visitor ID collected will be ignored. As a result the Unique visitor metric will become inaccurate, and sometimes different visitors may be falsely identified as the same visitor if the used device is similar. The Visitor profile will also be disabled in the UI. Please note that 'User ID' feature is still available.">

btw I find this not clear, could we explain more clearly? ` sometimes different visitors may be falsely identified as the same visitor if the used device is similar`

mattab (changes requested):
Really great we can have this feature in Matomo 3 and 4 :-) 

Note regarding documentation/letting people know: 
* we could create a new FAQ for this feature (eg easy to find when "no fingerprint analytics")
* document this in https://matomo.org/faq/general/faq_21418/
* also in: https://matomo.org/docs/privacy/
* and in https://matomo.org/features/

plugins/PrivacyManager/templates/privacySettings.twig
tsteur: This is not the final version yet anyway. Only wanting to make sure the concept is working, not details.

mattab:
> Not sure if "Disable fingerprint" is the right word since a fingerprint is still needed. 

Indeed, how else could we call it so it's really clear?

tsteur:
It's hard to say. What we're looking for is maybe not even related directly so much to "fingerprint". 

Saying it "anonymises" the fingerprint kind of suggests we usually don't do that but that'd be also not 100% correct. 

Also wondering what we could say how this maybe impacts cookie banners. Eg can we say as a result, if no personal data is tracked, no cookie consent is needed? So is it maybe more about avoiding "cookie consent" (if also no personal data is tracked)? 

Also be good to have confirmed that we're confident enough once we apply a random token per day that users can really no longer be identified across days.

core/Tracker/Settings.php
mattab: :+1:  and quite a long string so it's not possible to guess if one knew all the other data points

mattab:
Sounds good... 
as we probably couldn't say "Disable fingerprint" as it's still enabled within a day period, so maybe "Limit fingerprint"?

could we maybe name `Avoid cookie consent banner` as a section and under have both `Limit fingerprint` and `Tracking without cookies`?

> Also be good to have confirmed that we're confident enough once we apply a random token per day that users can really no longer be identified across days.

 someone with access to DB could potentially re-create the original hashes since other data points used in fingerprint are stored in the Db afaik, so as long as the daily (or more frequently re-generated, even customisable?) string is random and long, it would be secure?

mattab:
When consent has been given, would we still limit the fingerprint?

core/Tracker/VisitorRecognizer.php
```diff
@@ -87,7 +88,11 @@ public function findKnownVisitor($configId, VisitProperties $visitProperties, Re
 
         $isVisitorIdToLookup = !empty($idVisitor);
 
-        if ($isVisitorIdToLookup) {
+        if (PrivacyManager::shouldAnonymiseFingerprint()) {
+            $isVisitorIdToLookup = false;
+            $idVisitor = false;
```
mattab, line 93: here we wouldn't want to change the logic in "findKnownVisitor" when cookies are used and "limit fingerprint" feature is activated, then the cookie visitor id would need to be used to match visitors accurately. (we assume eg. consent was given when cookies are received by tracking api). 

(Otherwise this would break legit use cases and changes/bc break tracking API, especially as the setting is global and affects all sites)

mattab (changes requested):

tsteur:
> could we maybe name Avoid cookie consent banner as a section and under have both Limit fingerprint and Tracking without cookies?

@mattab that makes afaik no sense as from a privacy and user perspective tracking with cookie or fingerprint is the very same things. All/most privacy laws consider the fingerprint as the same as a cookie and that's why you'd want to ignore any visitorId from a cookie and ideally we'd also disable cookie feature in general in tracker. To avoid easy to understand / simple UI that is useful to users could have one setting "Avoid cookie consent banner" or "cookie less tracking" or similar. But splitting this up makes not really any sense except for making things complicated.

tsteur:
@mattab I'll close this PR for now. Let's discuss further this Wednesday. Wanting to prevent implementing some solution that might be obsolete in Matomo 4 and people need to change things again possibly so be better to work on it first in Matomo 4 and back port it to Matomo 3

mattab:
Might be more clear to use this approach of super privacy mode
https://github.com/matomo-org/matomo/issues/12737

plugins/PrivacyManager/templates/privacySettings.twig
Findus23: Once this is final it should also be translatable.

Issue #9748, opened 2016-02-10
actions/pages "average time on page" broken


hpvd:
actions/pages "average time on page" seems to be broken

please see attachments for example:

2 visits (filtert by shown segment)
- only one site visit
- only pageview
- no events etc...

show visit times in visitorlog (total and for page on hover)
but in actions/pages average time is 0:00

![2016-02-10_17h24_36](https://cloud.githubusercontent.com/assets/5681880/12953894/a6caf3b2-d01c-11e5-9870-f87faafe8dfe.png)

![2016-02-10_17h26_48](https://cloud.githubusercontent.com/assets/5681880/12953916/bdeb70a8-d01c-11e5-9f2b-fc4aea7a9784.png)

![2016-02-10_17h31_59](https://cloud.githubusercontent.com/assets/5681880/12953921/c36e487a-d01c-11e5-92d7-16e4bc1a951e.png)

![2016-02-10_17h25_27](https://cloud.githubusercontent.com/assets/5681880/12953928/c9bf2460-d01c-11e5-9fe7-8b58a09075b3.png)

tsteur:
It's currently like this "by design". We might fix this in Piwik 3.0 but it's not clear yet. We cannot fix it in 2.X as it requires DB schema change for log tables which would take a very long time. 

Background: The avg_time_on_page is not written for last action, only `visit_total_time` is updated for heartbeat. To do proper calculation we always need to know what `last_idlink_va` was and update time spent on the actual action, not on the next action. To have it backwards compatible with existing data will be quite challenging but doable.

hpvd:
thanks for fast and detailed answer.
I think it's quite important to have a possibility to see average time on special types of pages.
Background:
it's important to see when making changes on a page how these are liked by a visitor.
This could be done easily looking on development of average time on page.
If it rises, the visitor likes it. If time is lowered the change works opposite as planned...

(since it's not working as expected, one should give it a bug label. doesn't on?)

tsteur:
It's definitely a bug and a fix is needed at some point indeed. Otherwise the `avg_time_on_page` in the Actions report is simply not correct.

mattab:
Could we maybe fix this in Matomo 4 @tsteur ? assigning tentatively

tsteur:
[...]

PR #15860, opened 2020-04-25
Do not always extend visit for content impressions

tsteur:
@mattab the idea here is that content impressions is not quite an interactive action and as such it should not extend the visit or maybe only less often (if user scrolls after a while on the page we could consider it maybe interactive). The reason is mostly performance related to have less concurrent log_visit updates and therefore less potential locks. Often people have say 10-30 content blocks and either all of them are sent at the same time (eg if trackAllContentBlocks) or many of them are sent in parallel (when trackVisibleBlocksOnly and many become visible while scrolling). In Matomo 4 we can start queueing this requests so they are not sent all at the same time and instead as bulk (maybe they are already not sure) but in general we want to avoid too many concurrent requests here. 

The idea is so far we only extend visit if the last action was more than a minute ago. Could also completely remove it and never extend the visit (only content interactions would extend it). This can avoid many log_visit locks.

Haven't fully tested it yet, but wanted to get feedback on it as I reckon it can help.

For content impressions it will basically no longer update anything on log_visit which be good.

### DIFF
--- a/plugins/Actions/Tracker/ActionsRequestProcessor.php
+++ b/plugins/Actions/Tracker/ActionsRequestProcessor.php
@@ -70,27 +70,7 @@ class ActionsRequestProcessor extends RequestProcessor
 
         if (!empty($action)) { // other plugins can unset the action if they want
             $action->loadIdsFromLogActionTable();
-            if ($action instanceof ActionContent && !$request->getParam('c_i') && !$this->shouldUpdateLastVisit($visitProperties)) {
-                // impressions dont extend visit
-                $request->setMetadata('Actions', 'doNotExtendVisit', true);
-            }
         }
-    }
-    private function shouldUpdateLastVisit(VisitProperties $visitProperties)
-    {
-        $lastActionTime = $visitProperties->getProperty('visit_last_action_time');
-        if (!empty($lastActionTime)) {
-            // it is only numeric when directly being called afterRequestProcessed() and not eg handleExistingVisit
-            // because the VisitLastActionTime dimension will overwrite the original value of the visitor.
-            // we want to make sure to work on the value from the DB
-            $lastActionTimeDate = Date::factory($lastActionTime)->addPeriod(1, 'minutes');
-            if ($lastActionTimeDate->isEarlier(Date::now())) {
-                // we update visit_last_action_time only if visit_last_action_time was updated more than 5 min ago
-                // we do not update all the time or every minute as not needed and to save resources
-                return true;
-            }
-        }
-        return false;
     }
 
     /**
@@ -105,7 +85,11 @@ class ActionsRequestProcessor extends RequestProcessor
      */
     public function onExistingVisit(&$valuesToUpdate, VisitProperties $visitProperties, Request $request)
     {
-        if ($request->getMetadata('Actions', 'doNotExtendVisit')) {
+        /** @var Action $action */
+        $action = $request->getMetadata('Actions', 'action');
+
+        if (!empty($action) && $action instanceof ActionContent && !$request->getParam('c_i')) {
+            // content impressions should not extend the visit, only content interactions
             unset($valuesToUpdate['visit_last_action_time']);
             unset($valuesToUpdate['visit_total_time']);
         }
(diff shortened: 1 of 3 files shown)


mattab:
Sounds great, change makes sense. If users want accurate visit time, heartbeat timer would be a better solution. It would be fine to also remove the 1 minute logic and never update the visit. Be good to mention in developer changelog too.

plugins/Actions/Tracker/ActionsRequestProcessor.php
```diff
@@ -68,6 +70,44 @@ public function afterRequestProcessed(VisitProperties $visitProperties, Request
...
+        }
+    }
+    private function shouldUpdateLastVisit(VisitProperties $visitProperties)
+    {
+        $lastActionTime = $visitProperties->getProperty('visit_last_action_time');
+        if (!empty($lastActionTime)) {
+            // it is only numeric when directly being called afterRequestProcessed() and not eg handleExistingVisit
+            // because the VisitLastActionTime dimension will overwrite the original value of the visitor.
+            // we want to make sure to work on the value from the DB
+            $lastActionTimeDate = Date::factory($lastActionTime)->addPeriod(1, 'minutes');
```
diosmosis, line 86: Should this be `addPeriod(5` or is the comment below inaccurate?

plugins/Actions/Tracker/ActionsRequestProcessor.php
```diff
@@ -68,6 +70,44 @@ public function afterRequestProcessed(VisitProperties $visitProperties, Request
...
+     *
+     * @param array &$valuesToUpdate
+     * @param VisitProperties $visitProperties
+     * @param Request $request
+     */
+    public function onExistingVisit(&$valuesToUpdate, VisitProperties $visitProperties, Request $request)
+    {
+        if ($request->getMetadata('Actions', 'doNotExtendVisit')) {
+            unset($valuesToUpdate['visit_last_action_time']);
+            unset($valuesToUpdate['visit_total_time']);
```
diosmosis, line 110: Just a suggestion, but `doNotExtendVisit` could be handled in Visit.php so any plugin could add behavior like this.

tsteur:
fyi it seems I can't make this work. It seems actually needed to update `visit_last_action_time` as otherwise the `time_spent_ref_action` is calculated wrong.

You can see here some content impressions and their `time_spent_ref_action` where then eg multiple content impressions get a `time_spend_ref_action` of 120, but only the first should get `120` and the remaining to `0` . The same for some other entries around `60s` etc.

image

This means the data in the exported visitor log shows a wrong `time_spent_ref_action`

image

As a result, it also changes eg `sum_time_spent` on page where it aggregates the `time_spent_ref_action`  and it results in an increased value as it tracks eg `120` multiple times instead of just once.

image

Technically, I noticed the page view report should maybe not include `time_spent_ref_action` of content impressions for a given page, but then we'd result in `0 sum_time_spent` and it probably should be included.

So I basically couldn't figure out a way to make this work unfortunately.

FYI @mattab

mattab:
@tsteur 

> You can see here some content impressions and their time_spent_ref_action where then eg multiple content impressions get a time_spend_ref_action of 120, but only the first should get 120 and the remaining to 0 . The same for some other entries around 60s etc.

i'm not sure i fully understand yet, but say if we "forced" content impressions' `time_spent_ref_action` values to be 0, would it mean we could reconsider the PR and performance improvements?

> Technically, I noticed the page view report should maybe not include time_spent_ref_action of content impressions for a given page, but then we'd result in 0 sum_time_spent and it probably should be included.

it would be fine I think to have `sum_time_spent` as 0 in these cases, especially if it solves some major performance problem edge cases... The view would be that if people need accurate time on page, they would better use HeartBeat timer. (although we'd need to fix this other issue: https://github.com/matomo-org/matomo/issues/9748 ...)

tsteur:
Yeah we'd need to implement https://github.com/matomo-org/matomo/issues/9748 to make this work which in the end would cause more issues.

If we forced content impressions to have 0 values for `time_spent_ref_action`, then we could use this PR. However, anyone with content impressions wouldn't get an accurate time on page anymore (which may be fine as events seem to be breaking time on page feature already anyway if I see this right)

PR #15874, opened 2020-04-28
put rexpiring delay logic in Lock class

### LINKED ISSUE #15870: Reexpire archive lock less often
There was https://github.com/matomo-org/matomo/pull/15749 and https://github.com/matomo-org/matomo/issues/15747

But I think it might not fully work actually. In performance insights still seeing these queries executed quote often and taking quite some resources. It seems like might be because in `getDb()` https://github.com/matomo-org/matomo/blob/3.13.5-rc1/core/DataAccess/LogAggregator.php#L1188 we always create a new instance of ArchivingDbAdapter and therefore https://github.com/matomo-org/matomo/blob/3.13.5-rc1/core/DataAccess/ArchivingDbAdapter.php#L37 would be quite often `null`.

BTW should maybe change the default value to `0` so it's not `null + $timeBetweenReexpires` in `reexpireLock()`. 

Maybe would somehow need to remember the last time per archiving lock?

fyi @diosmosis

diosmosis:
Fixes #15870

### DIFF
--- a/core/DataAccess/LogAggregator.php
+++ b/core/DataAccess/LogAggregator.php
@@ -170,11 +170,6 @@ class LogAggregator
     private $allowUsageSegmentCache = false;
 
     /**
-     * @var \Piwik\Tracker\Db|\Piwik\Db\AdapterInterface|\Piwik\Db
-     */
-    private $readerDb;
-
-    /**
      * Constructor.
      *
      * @param \Piwik\ArchiveProcessor\Parameters $params
--- a/core/DataAccess/LogAggregator.php
+++ b/core/DataAccess/LogAggregator.php
@@ -170,6 +170,11 @@ class LogAggregator
     private $allowUsageSegmentCache = false;
 
     /**
+     * @var \Piwik\Tracker\Db|\Piwik\Db\AdapterInterface|\Piwik\Db
+     */
+    private $readerDb;
+
+    /**
      * Constructor.
      *
      * @param \Piwik\ArchiveProcessor\Parameters $params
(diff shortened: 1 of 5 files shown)


tsteur:
looks good if tests pass @diosmosis I haven't done a full test but debugged and looks good. It was applying the correct TTL and basically the lock instance now persists the time across individual archives as the instance is created when the archives starts and added to the stack.

It seems when browser archiving is enabled, it may not 100% work though. In my case it was creating the lock eg every time for each metric
image

I reckon this impacts only browser archiving though (haven't tested cron archiving). Maybe you could quickly double check it only creates the lock once per cron archive?

tsteur:
Actually, I think it might have been because I had archiving forced (`always_archive_data_day=1`) so all good

diosmosis:
@tsteur I can confirm it doesn't get called in core:archive for me, but for it to get called, the archiving job would have to run for a long time (even if I set the timeout to 30s, it would have to run for more than my local jobs run for).

core/DataAccess/LogAggregator.php
```diff
@@ -169,6 +169,11 @@ class LogAggregator
      */
     private $allowUsageSegmentCache = false;
 
+    /**
+     * @var \Piwik\Tracker\Db|\Piwik\Db\AdapterInterface|\Piwik\Db
+     */
+    private $readerDb;
```
tsteur, line 175: btw @diosmosis I suppose this one isn't needed?

diosmosis: removed it

tsteur:
@diosmosis sorry seems one test is now failing: https://travis-ci.org/github/matomo-org/matomo/jobs/680785282#L923

Not sure if the value is expected? Feel free to merge once the test is fixed.

PR #15877, opened 2020-04-28
Ensure oneClickUpdatePartTwo always returns json

### LINKED ISSUE #15876: Warning while upgrading "Warning - array_merge(): Argument #2 is not an array"
WARNING: /home/piwik-demo/storage/www/demo.piwik.org/plugins/CoreUpdater/Updater.php(134): Warning - array_merge(): Argument #2 is not an array - Matomo 3.13.5-rc1 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already) (Module: CoreUpdater, Action: oneClickUpdate, In CLI mode: false)

sgiehl:
That actually solves the issue and should show the exception message in the updater log. 
Or should we throw an exception if `oneClickUpdatePartTwo ` didn't work?

fixes #15876

### DIFF
--- a/plugins/CoreUpdater/Controller.php
+++ b/plugins/CoreUpdater/Controller.php
@@ -175,11 +175,19 @@ class Controller extends \Piwik\Plugin\Controller
 
     public function oneClickUpdatePartTwo()
     {
-        Piwik::checkUserHasSuperUserAccess();
-
         Json::sendHeaderJSON();
 
-        $messages = $this->updater->oneClickUpdatePartTwo();
+        $messages = [];
+
+        try {
+            Piwik::checkUserHasSuperUserAccess();
+            $messages = $this->updater->oneClickUpdatePartTwo();
+        } catch (UpdaterException $e) {
+            $messages = $e->getUpdateLogMessages();
+            $messages[] = $e->getMessage();
+        } catch (Exception $e) {
+            $messages[] = $e->getMessage();
+        }
 
         echo json_encode($messages);
     }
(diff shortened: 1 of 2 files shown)


plugins/CoreUpdater/Controller.php
```diff
@@ -175,11 +175,19 @@ public function oneClickUpdate()
 
     public function oneClickUpdatePartTwo()
     {
-        Piwik::checkUserHasSuperUserAccess();
```
diosmosis, line 178: This shouldn't be removed?

diosmosis: Oh, nvm I see it below.

sgiehl: moved it to the try catch, so the error will be returned as json

plugins/CoreUpdater/Updater.php
```diff
@@ -131,7 +131,9 @@ public function updatePiwik($https = true)
         $response = Http::sendHttpRequest($partTwoUrl, 300);
         $response = @json_decode($response, $assoc = true);
 
-        $messages = array_merge($messages, $response);
+        if (!empty($response)) {
+            $messages = array_merge($messages, $response);
+        }
```
diosmosis, line 136: In case there is an exception, should we throw an UpdaterException? Eg, we could in `oneClickUpdatePartTwo()` return `{"error": true, "messages": [...]}` and throw the exception as it would have been before.

diosmosis: Might be good to add a quick test for this too, like a unit test for the controller that uses a mock updater. Not sure if it's worth the effort though.

sgiehl: Thought about returning an error and throwing an exception in that case. But actually when updating the plugin fails and some plugin would be incompatible then, they would be disabled afterwards. Not sure if we should report that as failed update 🤔

diosmosis: hmm, maybe not

Issue #13829, opened 2018-12-11
Audit Date class and re-structure API


diosmosis:
The logic in the `Date` class is strange and somewhat backwards. Multiple methods have documentation that contradict the actual methods, and sometimes the actual logic just doesn't make sense.

For example:

- `Date::factory('today', $timezone)` takes a timezone, but does not return 'today in $timezone`, it returns 'the time in $timezone that is 'today' in UTC'. then it discards the timezone and treats the result as UTC.
- `Date::$timestamp` is supposed to store the time in UTC, but in `getTimestamp()` it is treated as a time in another timezone.
- getTimestamp() says it returns a UTC time, but it treats the stored time as non-UTC and converts it to UTC. getTimestampUTC() however just returns the timestamp. so `getTimestamp() != getTimestampUTC()` even though both say they return UTC time.

The API should be audited, trimmed and made coherent, but only for Matomo 4 since so much depends on this code.

Things that should be done:

* [ ] timestamps should always be in UTC. timezones are just a view on that timestamp. so Date should not store a timezone and should not do any conversion in methods that return timestamps.
* [ ] methods that take string representations or return string representations should accept timezones, eg: `getDatetime($timezone = 'UTC')`, `toString($timezone = 'UTC')`, `Date::factory($value, $timezone = 'UTC')`
* [ ] since timestamps should always be UTC, `Date::factory(...)` should not accept timestamps (or should throw a if a timezone is specified w/ the timestamp). the semantics of `Date::factory()` should be, give me the posix timestamp of this date in this timezone. (ie, give me the UTC timestamp of `2012-03-04 03:05:23 UTC+5`.
* [ ] ...UTC() methods should be removed

Refs #13799
Refs #13786

diosmosis:
Should also probably make sure the mysql session uses UTC instead of any other configured timezone.

mattab:
How much work would this be do you reckon, probably several days? 
do you know if tackling this debt solve any existing known bug?

diosmosis:
> How much work would this be do you reckon, probably several days?

I don't know really, possibly longer, maybe less. It would just be refactoring Date and getting all tests to pass. Then manual tests to make sure nothing strange happens. This part could take a while.

> do you know if tackling this debt solve any existing known bug?

There were some bugs that this caused that were worked around by some special methods. There may be many bugs we don't know about. @tsteur would have a better opinion on the value of this issue.

tsteur:
[...]

PR #15886, opened 2020-04-29
Limit the fingerprint 

tsteur:
The purpose of the fingerprint is to put all tracking requests within the same visit into the same visit in the DB. A fingerprint is basically valid for up to 30 minutes after the visit ends (unless different visit length is configured). 

By adding a random hash to the fingerprint we make sure when a user disables cookies, a visitor cannot be identified across multiple days and such no cookie consent should be needed.

This change should not really change behaviour for anyone AFAIK. In Matomo 4 we could remove the check for `create_new_visit_after_midnight` potentially. We cannot limit the fingerprint in that case as when someone has `create_new_visit_after_midnight=0` and has cookies disabled, then a new visit would be still created after midnight since the configId would change.

The limited fingerprint will also not apply when enabling `[Tracker]enable_fingerprinting_across_websites = 1` since different sites can have different timezones and therefore we can't add any date to the fingerprint. 


config/global.ini.php
```diff
@@ -761,6 +761,15 @@
 ; Note: setting this to 0 increases your users' privacy.
 enable_fingerprinting_across_websites = 0
 
+; When enabled (set to 1) and cookies are disabled in the tracker, then the fingerprint will be restricted even further
+; to prevent a visitor to be tracked over multiple days. This means typically that no cookie consent is needed when cookies
+; are disabled. You may disable this feature if you don't want to use cookies in the browser, but still be able to get more
```
mattab, line 766: > You may disable this feature if you don't want to use cookies in the browser, but still be able to get more accurate unique visitor numbers.

Actually afaik, you would disable this feature only if you have set a `window_look_back_for_visitor` value higher than 24 hours. By default window_look_back_for_visitor = 0 which means we only look back 30min to match a visit. Returning visitors wouldn't even be matched from 2 hours ago. So it's only when cookies are disabled and window_look_back_for_visitor > say 43200 then it might be useful to deactivate this.

core/Tracker/Settings.php
```diff
@@ -58,6 +61,22 @@ public function getConfigId(Request $request, $ipAddress)
 
         $browserLang = substr($request->getBrowserLanguage(), 0, 20); // limit the length of this string to match db
 
+        if ($this->isSameFingerprintsAcrossWebsites) {
+            $fingerprintSalt = ''; // fingerprint salt won't work when across multiple sites since all sites could have different timezones
```
mattab, line 65: Interesting. be good to add a comment in global.ini.php above the setting like "when you set this to 1, this will also set enable_fingerprinting_limited_when_cookie_disabled=0"

core/Updates/3.13.6-b1.php
```diff
@@ -0,0 +1,30 @@
...
+ * Update for version 3.13.6-b1.
+ */
+class Updates_3_13_6_b1 extends PiwikUpdates
+{
+
+    public function doUpdate(Updater $updater)
+    {
+        $config = Config::getInstance();
+        $tracker = $config->Tracker;
+        $tracker['enable_fingerprinting_limited_when_cookie_disabled'] = 0;
```
mattab, line 26: could we do this only when  window_look_back_for_visitor > say 43200? (if  window_look_back_for_visitor is not custom, then no impact to users afaik)

mattab (commented):
Awesome to see this pr. this will improve privacy and limiting the fingerprint so it's clear that people can use without consent for cookie-less analytics.

PS: didn't review the code

core/Tracker/Settings.php
tsteur: that is already done

config/global.ini.php
tsteur: Are you certain fingerprint is only looked back 30 minutes by default? Then I'd remove the config setting, and also the extra tracking parameter `cd` when cookies are disabled and we simply do apply the shorter fingerprint always unless `window_look_back_for_visitor > 24 hours` (but even then it would be only needed probably when not a new visit should be created after midnight). So we could apply this more privacy friendly fingerprint pretty much always

mattab: looking at the code, yes it looks like the window_look_back_for_visitor is only used for fingerprint and defaults to 30min (defaults to the visit duration setting)

core/Updates/3.13.6-b1.php
tsteur: as mentioned we're creating new visits after midnight anyway... so this should probably not be important... 
unless `create_new_visit_after_midnight = 0`. Right now it looks like we can safely always apply the shorter fingerprint.

core/Tracker/FingerprintSalt.php
```diff
@@ -0,0 +1,88 @@
...
+                $deleted[] = $name;
+            }
+        }
+
+        return $deleted;
+    }
+
+    public function getDateString(Date $date, $timezone)
+    {
+        $dateString = Date::factory($date->getTimestampUTC(), $timezone)->toString();
```
tsteur, line 58: not 100% sure if this is correct? Had initially `Date::factory()->setTimezone($timezone)` but then noticed in tests it might not return correct values unless I'm expecting the wrong values. In ArchiveProcessor we're using for example `setTimezone`: https://github.com/matomo-org/matomo/blob/3.13.5/core/ArchiveProcessor/Parameters.php#L166 so would have maybe expected `setTimezone` is right but then in CronArchive we're also using it as a parameter https://github.com/matomo-org/matomo/blob/3.13.5/core/CronArchive.php#L1504 ... looking at the test results the parameter seems to give expected result

tests/PHPUnit/System/expected/test_CustomEvents__Live.getLastVisitsDetails_day.xml
```diff
@@ -794,8 +794,8 @@
 				<eventCategory>Music</eventCategory>
 				<eventAction>play50%</eventAction>
 				<bandwidth />
-				<timeSpent>60</timeSpent>
-				<timeSpentPretty>1 min 0s</timeSpentPretty>
+				<timeSpent>0</timeSpent>
```
tsteur, line 797: Couldn't figure out where these changes are from. The tests seem to pass in 3.x-dev. 

Not sure how this would change in this PR since in the tests itself the fingerprintSalt is basically not set and nothing changes.

tsteur:
Should be ready for a review. If I see things right, nothing changes for users in the end but hard to say.

sgiehl:
Haven't looked deeper into the code yet. But wouldn't changing the config id will have the effect, that every visitor will be a new visitor after the update? 🤔

tsteur:
@sgiehl only when cookies are disabled. Otherwise not. The fingerprint as @mattab mentioned was always only used to look 30 minutes back. So it would only affect it when cookies are disabled, and a visitor was active in the last 30 minutes. Then one new visit would be created which should be fine.

tsteur:
@sgiehl did you maybe have another look?

tests/PHPUnit/System/expected/test_CustomEvents__Live.getLastVisitsDetails_day.xml
sgiehl: Looking at the tests it seems the changes to the test files can be reverted.

plugins/SitesManager/SitesManager.php
```diff
@@ -13,12 +13,15 @@
 use Piwik\Common;
 use Piwik\Config;
 use Piwik\Container\StaticContainer;
+use Piwik\Date;
 use Piwik\Exception\UnexpectedWebsiteFoundException;
 use Piwik\Option;
 use Piwik\Piwik;
 use Piwik\Plugins\CoreHome\SystemSummary;
 use Piwik\Settings\Storage\Backend\MeasurableSettingsTable;
+use Piwik\Site;
```
sgiehl, line 22: Actually no need to add that use 
```suggestion
```

core/Tracker/FingerprintSalt.php
sgiehl: Guess that's something we should probably check again with #13829

sgiehl (changes requested):
Did some short local testing and it seems to work as expected I guess.

core/Config/IniFileChain.php
```diff
@@ -538,4 +543,22 @@ private function dumpSettings($values, $header)
...
+                $key = $this->replaceInvalidChars($key);
+                if (is_array($arrayValue)) {
+                    $arrayValue = $this->replaceInvalidChars($arrayValue);
+                }
+
+                $result[$key] = $arrayValue;
+            }
+            return $result;
+        } else {
+            return preg_replace('/[^a-zA-Z0-9_-]/', '', $value);
```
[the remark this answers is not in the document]
diosmosis: That wouldn't solve this problem, because `[...]` is valid in this case, but `]asldkjf`, eg, is not. Throwing an exception isn't an issue for me though so if that is more desired than removing the chars then that's fine too.

sgiehl (approved):

sgiehl:
@tsteur I've reverted the test file changes and the tests seem to pass again. From my point a view the changes look fine now. Maybe you can also have a look at the changes from @diosmosis. Otherwise feel free to merge

PR #15888, opened 2020-04-30
Require authentication when sending custom ip with tracking requests

### LINKED ISSUE #13471: Tracking API: when an invalid token_auth is specified, should the request be entirely dropped, or tracked anyway?

sgiehl:
fixes #13471 

Note: matomo-php-tracker already only sends cip param when token_auth is provided. See https://github.com/matomo-org/matomo-php-tracker/blob/ed2de26b2a5710292d09436a246998d1e28b5760/MatomoTracker.php#L1737,L1739

### DIFF
--- a/core/Tracker/Request.php
+++ b/core/Tracker/Request.php
@@ -892,7 +892,7 @@ class Request
 
         if (!$this->isAuthenticated()) {
             Common::printDebug("WARN: Tracker API 'cip' was used with invalid token_auth");
-            return IP::getIpFromHeader();
+            throw new InvalidRequestParameterException("Tracker API 'cip' was used, requires valid token_auth");
         }
 
         return $cip;
--- a/tests/PHPUnit/Integration/Tracker/VisitTest.php
+++ b/tests/PHPUnit/Integration/Tracker/VisitTest.php
@@ -34,6 +34,7 @@ class VisitTest extends IntegrationTestCase
         // setup the access layer
         FakeAccess::$superUser = true;
 
+        Fixture::createSuperUser(true);
         Manager::getInstance()->loadTrackerPlugins();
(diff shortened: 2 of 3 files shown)


tsteur:
@sgiehl are there other things in the tracker that need to be checked? Eg cdt? But I suppose some CDT is valid vs some other isn't so it is tricky

sgiehl:
@tsteur haven't checked all parameters. but `cdt` isn't easy. We would need to check the config for `tracking_requests_require_authentication_when_custom_timestamp_newer_than` and compare it with the value to say if a token auth is required or not.

tsteur:
Thought it wouldn't be easy. All good shouldn't be needed I suppose

sgiehl:
Fixed the failing test. @tsteur is that good to merge then?

PR #15889, opened 2020-04-30
allow invalidating plugin archives only and archiving past data for plugins

### LINKED ISSUE #11974: Allow plugins to generate their reports using historical data (eg. Custom Reports)
In Piwik reports are processed based using the raw data. Once an archive has been processed for a website and a specific date period, and stored in the database (during the `core:archive` process), it can only be [invalidated](https://piwik.org/faq/how-to/faq_155/) (via HTTP API or CLI or InvalidateReports plugin) which will force all plugins to re-process all their respective reports in the next archiving process. This is problematic because it can take a lot of time to reprocess previous reports for 3 or 12 months. 

-> Plugins should be able to easily create their own reports for some of the historical data. this is useful for all plugins which don't create new raw data but re-uses the existing raw data. For example, Funnels plugin would like to process Funnels data in the past 6 months. 

To make it possible for any plugin to append their reports to the existing archives, we need to make some changes to the archiving process. 

Ideas:

* Piwik core to allow an existing archive to be modified later-on so new plugin's reports can be part of and appended in the existing old archive 
* Piwik core to allow each plugin to force re-processing of past data and the ability to specify how far back the data should be re-processed. (working for our two use cases: when using the core:archive console command in a cron, or when using broworser triggered archiving and looking at historical data.)

(not related but a bit similar to https://github.com/piwik/piwik/issues/7573 )

diosmosis:
Fixes #11974

This PR is based off of #15499 .

Changes made:
- we allow invalidating plugin specific archives, this results in entries being added to archive_invalidations for the plugin archive. they will not be added if we invalidate a normal archive, even if as part of that invalidation some dependent plugin specific archives (like ones created by VisitorInterest/Goals) are invalidated.
- in core:archive, if we see a plugin specific archive in the invalidations table, we archive it separately w/ the `pluginOnly=1` query parameter.
- in Rules.php if in core:archive and pluginOnly=1 is used, we force creating a plugin specific archive.
- in Archive.php/Loader.php/etc. we don't just look for one type of archive (ie, `plugin` vs `all`), we look through every one and use the latest archive values. so if there is an `all` archive and a `ExamplePlugin` archive that is more recent, the archive data in `ExamplePlugin` will be used for `ExamplePlugin` data, but other plugin data will come from `all`.
- plugins on activation can then invalidate archives for their own plugin to schedule archiving for data in the past, without initiating archiving for every other plugin at the same time.

Notes:
- **browser archiving is not supported.** If a user does not use core:archive, the invalidations don't do anything unless the plugin specific archive existed before. I think this is ok since premium plugins need this feature and we don't expect them to be used w/o core:archive, correct?

Example PR can be found in funnels


diosmosis:
@tsteur this PR is based off the other one so might be hard to review, but it's ready for an initial one. once that's done and it looks good to be in core, then I'll probably refactor CronArchive to test more of its internals. Don't want to do this now since both PRs are already pretty significant.

tsteur:
@diosmosis just fyi started reviewing but it's bit difficult as it is based on https://github.com/matomo-org/matomo/pull/15499/files so I reckon we will be trying to get #15499 merged first which will make it easier

diosmosis:
@tsteur you should be able to review this one now

tsteur:
@diosmosis had a first look for a while and starting to understand few things. Must have been so hard to develop!

I've had a look at the funnels pull request to see how plugins basically trigger rearchiving of data for that plugin. Do I basically see this right that when eg someone adds or changes a custom report, then custom reports will mark archives as invalid like below?

```php
     $invalidator->markArchivesAsInvalidated($idSites, $dates, 'month', $segment = null, $cascadeDown = false, $invalidateRanges = false,
            $plugin = 'CustomReports');
```

I wonder if we need something similar than `process_new_segments_from` where we configure how far back by default we rearchive things?  like `process_plugin_specific_archives_from=beginning_of_time|lastN`. I guess `last_edit_time` and `creation_time` wouldn't quite work because the creation time is always different for each custom report, each funnel, each ... We could then maybe have a method in core that automatically goes back the number of months as needed and all the plugin would need to do be like `StaticContainer::get(ArchiveInvalidator::class)->rearchivePlugin($pluginName, $idSite)`. Basically, the funnels `activate` method be mostly in core. 

In Funnels plugin we wouldn't run this on activate btw. but instead when a new funnel is added or changed. Some for custom reports when a custom report was added or changed. A/B testing might also trigger later a rearchive when the A/B test was changed etc.

I suppose the plugin would always need to rearchive everything? Say there are 20 custom reports and a new one was added. I suppose we'd then need to rearchive all custom reports for the given date range? I reckon, if we know in the Custom Reports Archiver that a plugin specific archive is requested, we could maybe check which custom reports were changed or created recently and only archive these but not 100% sure yet if that works.

tsteur:
I suppose we would also need to have a flag whether to also reprocess all segments for the given plugin or not.

core/Archive.php
```diff
@@ -800,10 +801,12 @@ private function prepareArchive(array $archiveGroups, Site $site, Period $period
...
             $this->initializeArchiveIdCache($doneFlag);
 
-            $idArchive = $archiveLoader->prepareArchive($plugin);
+            $idArchives = $archiveLoader->prepareArchive($plugin);
 
-            if ($idArchive) {
-                $this->idarchives[$doneFlag][$periodString][] = $idArchive;
+            if (!empty($idArchives)) {
+                foreach ($idArchives as $idArchive) {
+                    $this->idarchives[$doneFlag][$periodString][] = $idArchive;
```
tsteur, line 808: note to myself: do we need to know which idarchives contain only plugin archive etc

tsteur (commented):
@diosmosis looking at it again right now my biggest concern is performance and IO. Do you think with this logic it is possible to only trigger an archive for a specific part within a plugin? Say a funnel with ID 10 is created and we'd only want to archive this funnel with ID 10 for all segments and periods? A few plugins would maybe need to process all data. Eg when you install UsersFlow then we'd want to reprocess the entire UsersFlow archiver. But when adding or changing eg a funnel, we'd only want to reprocess the needed funnel since it's very slow to process funnels, custom reports, ...

Also not sure yet about how it behaves in things like below, eg

* We do a plugin specific archive for funnels (ID 4) at say 8pm
* Then we have a regular archive (ID 5) at say 8:20pm

Will the archiver know which archives need to be regenerated eg 900 seconds (see time_before_today_archive_considered_outdated) later at 8:22pm? (We'd expect the archiver not to trigger a new archive because the last full archive was done at 8:20pm maybe (120 seconds later only)) And will the logic always use the latest archive for a funnels? It would need to use the one from Archive ID 5 in this case and ignore the plugin specific report? (although there could be an edge case maybe depending on when the archive for regular archive ID5 started)

core/Archive.php
diosmosis: Shouldn't need to know for requesting data

diosmosis:
@tsteur 

> looking at it again right now my biggest concern is performance and IO. Do you think with this logic it is possible to only trigger an archive for a specific part within a plugin?

Yes, this is possible via standardizing the `setArchiveOnlyReport()` approach. Plugins will have to handle this themselves though.

> Will the archiver know which archives need to be regenerated eg 900 seconds (see time_before_today_archive_considered_outdated) later at 8:22pm? (We'd expect the archiver not to trigger a new archive because the last full archive was done at 8:20pm maybe (120 seconds later only))

I'm not sure I 100% understand, but here some possible cases:

- an invalidate for all plugins for same idsite/period is set
- an invalidate for funnels (ID 4) is set
- archiver picks up all plugins archive first
- then plugin specific

*what happens:* not sure exactly what happens here, but ideally we would skip the plugin specific archive. in fact we can probably add this logic if it's not there already (or move it if it is) to `canSkipThisArchive()`, which will apply it in every place that matters.

- a plugin specific archive for funnels (funnel ID 4) is archived first at 8pm
- then we start an all plugins archive at say 8:20pm

*what happens:* this logic isn't implemented, but here's what could happen. first `setArchiveOnlyReport()` takes an array of reports. second, in PluginsArchiver before archiving we select every blob/numeric name in the queried idarchives w/ ts_archived that is older than time_before_today_archive_considered_outdated. we set that as `setArchiveOnlyReport()` and archive each plugin. this might be overkill though given the plugin specific re-archive shouldn't happen often. anyway, then the all plugins archive won't contain the still valid data.

> And will the logic always use the latest archive for a funnels? It would need to use the one from Archive ID 5 in this case and ignore the plugin specific report?

That is the intent. When querying archive data we're sorting by idarchive desc, so it should pick the data w/ the latest idarchive. We could make it more specific by sorting by ts_archived desc.

diosmosis:
>  When querying archive data we're sorting by idarchive desc, so it should pick the data w/ the latest idarchive.

Actually we're not doing that but we're supposed to be.

tsteur:
> Yes, this is possible via standardizing the setArchiveOnlyReport() approach. Plugins will have to handle this themselves though.

Does this mean in the archive invalidate logic we'd also store an ID and then somehow instead of $plugin.get we trigger some CoreAdmin.archivePlugin(pluginName, idToArchive) method or so? Then this API for example directly creates an instance of the archiver and triggers the archive for that plugin (and calls `setArchiveOnlyReport`) should an API be specified? I reckon if that exists things could/should work. 

BTW be good to also add things mentioned in https://github.com/matomo-org/matomo/pull/15889#issuecomment-626714794

diosmosis:
> Does this mean in the archive invalidate logic we'd also store an ID and then somehow instead of $plugin.get we trigger some CoreAdmin.archivePlugin(pluginName, idToArchive) method or so? 

Something like this, but it's possible we invalidate an archive that doesn't exist, in which case there won't be an ID.

However if you're assuming the code will just replace the existing report in an archive, then this isn't that simple due to what we discussed in slack. If not, can you think about my comments on creating the new plugin specific archive?

tsteur:
> Something like this, but it's possible we invalidate an archive that doesn't exist, in which case there won't be an ID.

@diosmosis not sure what you mean here? I was thinking eg we could add an ID to the archive invalidate table and a plugin may provide an additional ID when invalidating a report which the CronArchive would then forward when archiving that invalidated archive entry. Not sure it's clear what I mean? We can still add a new archiveId and not alter an existing archive.

diosmosis:
@tsteur finished an implementation for `setArchiveOnlyReport()`. I still want to create a POC PR for CustomReports and need to fix the tests, but you can take a look at the approach.

I introduced a new done value, `DONE_PARTIAL`, for archives w/ only some (or just one) report. It is treated as an invalidated archive when querying, if browser triggered archiving is enabled it ignores them and will re-archive if a full archive is not found. If core:archive is enabled, if there is a partial archive, we look in it along w/ other latest archives when querying. When archiving in core:archive, we ignore it and only look for all or plugin archives (this way if there is a newer partial archive, we still re-archive everything).

tsteur:
@diosmosis was the done_partial needed? For some specific reason maybe?

diosmosis:
>  was the done_partial needed? For some specific reason maybe?

Yes, to only re-archive a specific report, eg, for custom reports. We can't update the existing report because we don't have the logic to update blob reports like that. If we add a new archive we have to make sure it's not treated like a full archive, since there will just be one report.

core/Archive.php
```diff
@@ -789,21 +789,29 @@ private function prepareArchive(array $archiveGroups, Site $site, Period $period
         $invalidateBeforeArchiving = !SettingsServer::isArchivePhpTriggered();
 
         $parameters = new ArchiveProcessor\Parameters($site, $period, $this->params->getSegment());
+
+        $requestedReport = Common::getRequestVar('requestedReport', '', 'string');
+        if ($requestedReport) {
```
tsteur, line 794: should we accept this parameter only if `SettingsServer::isArchivePhpTriggered()`  as it means we know eg the super user is authenticated?

core/ArchiveProcessor.php
```diff
@@ -668,4 +684,17 @@ public function getArchiveWriter()
...
+
+    public function isArchiving(string $reportName)
+    {
+        $onlyReportName = $this->getParams()->getArchiveOnlyReport();
+        return empty($onlyReportName) || $onlyReportName == $reportName;
+    }
+
+    private function checkArchiveOnlyReport(string $name)
+    {
+        if ($name == $this->getParams()->getArchiveOnlyReport()) {
```
tsteur, line 696: not quite sure about the purpose behind this? and does it need to be reset in `this->getPArams()` again? eg after inserting a numeric record?

core/CronArchive.php
```diff
@@ -1384,6 +1398,52 @@ private function makeWebsiteIdArchiveList(array $websitesIds)
...
+         * a plugin when archiving for just one plugin.
+         *
+         * If your plugin has an archiver but does not provide a `get` API method for general metrics,
+         * then you can use this method to specify which API method to use to make sure the plugin data
+         * is archived.
+         *
+         * @param string &$method Set this value to the method if `$plugin` is your plugin.
+         * @param string $plugin The name of the plugin.
+         */
+        Piwik::postEvent('CronArchive.getArchivingAPIMethodForPlugin', [&$method, $plugin]);
```
tsteur, line 1435: Be great if this maybe wasn't needed like if there was an API like `CoreAdminHome.archivePluginSpecificReport($plugin, $requestedReport, $idSite, $period, $date, $segment = false)` which then would maybe directly bypass the archiver and this API method could instantiate the requested archiver directly etc. This would make it easier for plugin developers as such a method wouldn't need to be specified plus some logic would maybe get easier? Not sure... might be also faster though. Guess it would do something like this: https://github.com/innocraft/plugin-CustomReports/blob/4.x-dev/Commands/ArchiveReports.php#L59-L226

Not replacing entries within an existing archive though but adding a new DONE PARTIAL archive.

The advantage could be that this method would be also callable separately for any plugin maybe. The method would also check if `SettingsServer::isArchivePhpTriggered() = true`.

Not sure it's that easy though or if it in the end makes the logic a lot easier. The reading of archives be still as "complicated".

core/DataAccess/Model.php
```diff
@@ -349,7 +375,7 @@ public function getArchiveIdAndVisits($numericTable, $idSite, $period, $dateStar
                                OR name = '" . ArchiveSelector::NB_VISITS_CONVERTED_RECORD_LOOKED_UP . "')
                          $timeStampWhere
                          AND ts_archived IS NOT NULL
-                     ORDER BY ts_archived DESC, idarchive DESC";
+                     ORDER BY ts_archived ASC, idarchive DESC";
```
tsteur, line 378: shouldn't this still be DESC maybe?

core/CronArchive.php
tsteur: Could make it more complicated I guess as it would still need to fetch all archiveIds and check if reprocessing is needed etc and would probably add some duplicated code there maybe. So feel free to ignore it unless you want to explore it.

Just thought be nice not needing this event.

tsteur (commented):
Haven't really tested it as will wait until you have the custom reports example ready. Overall getting bit complicated and hard to understand everything but that's maybe also expected and probably can't be simplified much. With complicated I mean the different done flags, fetching multiple idArchives and knowing which one will be used etc. It just adds extra code in many places and I reckon it would have been awesome if it somehow wouldn't add that much code to many classes but as said I can see why this might be needed unfortunately (eg it's hooking into invalidations, reading, writing, etc so it's all expected kind of probably)

diosmosis:
Might be able to simplify some w/ creation of the new DONE_PARTIAL flag, I'll check.

core/DataAccess/Model.php
diosmosis: The way the rows are processed, the rows encountered latter overwrite the former in the result array (this is in ArchiveSelector.php). So using ASC means we end up using the latest archives. It was just easier to code than DESC where we'd have to stop when the ts_archived changed, but per name. Also, originally it was DESC so I could add a LIMIT, but because of how archives are structured w/ status flags & archive data together (and we need done flags + visits + conversions), there's no way to know how many rows we need to select.

core/CronArchive.php
diosmosis: Well if we're archiving for a specific report, we can just try to call that API method. However, if it's the entire plugin archive that's invalidated, then we don't necessarily know which API method to call.

I guess we could just not do API.get and do just a `API.initiateArchiving` method, but then there would be two code paths for initiating archiving, one through Archive.php and one through this other API method.

core/ArchiveProcessor.php
diosmosis: It's to automatically check if we archived the one report that was requested, and if so, set the value to DONE_PARTIAL. You're right it needs to be reset if a different archive is inserted at any time. I guess it can't be done automatically in an easy way...

core/Archive.php
diosmosis: 👍 should have been that way, maybe I removed the code

diosmosis:
@tsteur simplified the code a bit and finished the poc

plugins/CoreAdminHome/API.php
```diff
@@ -240,6 +243,41 @@ public function getTrackingFailures()
...
+     * @param bool $plugin
+     * @param bool $report
+     * @return mixed
+     * @throws \Piwik\Exception\UnexpectedWebsiteFoundException
+     * @internal
+     */
+    public function archiveReports($idSite, $period, $date, $segment = false, $plugin = false, $report = false)
+    {
+        // if cron archiving is running, we will invalidate in CronArchive, not here
+        $isArchivePhpTriggered = SettingsServer::isArchivePhpTriggered();
```
tsteur, line 260: do we need to check for super user permissions here?

tsteur: todo: need to check later how we can prevent a superuser calling this method on the cloud directly. But it should still be callable for the archiving. to be seen... might be only a matter of checking again that it won't trigger an archive if it was archived recently within TTL and that it won't simply invalidate the report

core/Archive.php
tsteur: btw a comment explaining why we might get multiple idArchives and which one will be used for what might be good?

plugins/CoreAdminHome/API.php
tsteur: I see core is using this method now as well and I suppose when browser archiving is enabled then a regular user might need to be allowed to call this. 

I wonder if we'd need to do something like `if (\Piwik\API\Request::getRootApiRequestMethod() === 'CoreAdminHome.archiveReports') { Piwik::checkUserHasSuperUserAccess() }`?

Might be generally also good to check always at least `Piwik::checkUserHasViewAccess($idSite);`? So a user can't trigger this method for like any date etc.

core/DataAccess/Model.php
```diff
@@ -75,12 +75,26 @@ public function getInvalidatedArchiveIdsSafeToDelete($archiveTable)
...
+                $pair = $duplicateArchives[0];
+                if (strpos($pair, '.') === false) {
+                    continue; // see below
+                }
+
+                list($idarchive, $value) = explode('.', $pair);
 
+                array_shift($duplicateArchives);
+
+                if ($value != ArchiveWriter::DONE_PARTIAL) {
```
tsteur, line 91: How would a done partial archive ever get invalidated I wonder? Eg say we invalidate the last 6 months and then rearchive all data. Would the partial archives still exist?

tsteur (commented):
Left a few comments @diosmosis generally looking good though

core/ArchiveProcessor.php
```diff
@@ -247,8 +247,7 @@ public function aggregateNumericMetrics($columns, $operationToApply = false)
         $metrics = $this->getAggregatedNumericMetrics($columns, $operationToApply);
 
         foreach ($metrics as $column => $value) {
-            $value = Common::forceDotAsSeparatorForDecimalPoint($value);
-            $this->archiveWriter->insertRecord($column, $value);
+            $this->insertNumericRecord($column, $value);
```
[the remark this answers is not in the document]
diosmosis: it's called in `insertNumericRecord()`

plugins/CoreAdminHome/API.php
diosmosis: `Piwik::checkUserHasViewAccess($idSite);` should definitely be called. I don't think we can  do `Piwik::checkUserHasSuperUserAccess()` since then Archive.php would need to set superuser access. I guess we can use the `getRootApiRequestMethod()` check... can't think of an issue w/ that.

diosmosis:
> btw a comment explaining why we might get multiple idArchives and which one will be used for what might be good?

will add a comment. reason is that now we can have full archives (DONE_OK/DONE_INVALIDATED and partial archives w/ single reports, w/ DONE_PARTIAL). so we have to look in both and use the latest data.

core/DataAccess/Model.php
diosmosis: It's always treated as invalidated, we never treat it like it's a full archive so if the latest is just a partial archive, we will initiate archiving when doing the "should we archive this" check.

> Eg say we invalidate the last 6 months and then rearchive all data. Would the partial archives still exist?

The partial archives would exist, but with a older ts_archived. We would not use the data in there, if there is more recent data we can use.

When purging, the partial archives that are older than a DONE_OK/DONE_INVALIDATED archive are removed.

diosmosis:
Updated.

tsteur (commented):
@diosmosis left 2 minor comments. In general it works, however maybe the order these invalidations are processed seemed bit random (aktually it appeared random but notice there is kind of a pattern behind it where it invalidates days, and when possible weeks and months but eg it does this 
one: `INFO [2020-06-15 00:59:06] 8387   Archived website id 1, period = month, date = 2020-03-01, segment = '', plugin = CustomReports, report = 14, 26 visits found. Time elapsed: 1.498s` before it archived the indivdual days?

```
INFO [2020-06-15 00:58:41] 8387   Archived website id 1, period = day, date = 2020-06-13, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.211s
INFO [2020-06-15 00:58:41] 8387   Archived website id 1, period = day, date = 2020-06-11, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.212s
INFO [2020-06-15 00:58:41] 8387   Archived website id 1, period = day, date = 2020-06-09, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.212s
INFO [2020-06-15 00:58:43] 8387   Archived website id 1, period = day, date = 2020-06-08, segment = '', 3 visits found. Time elapsed: 1.510s
INFO [2020-06-15 00:58:45] 8387   Archived website id 1, period = week, date = 2020-06-08, segment = '', 3 visits found. Time elapsed: 1.814s
INFO [2020-06-15 00:58:46] 8387   Archived website id 1, period = day, date = 2020-06-04, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.852s
INFO [2020-06-15 00:58:46] 8387   Archived website id 1, period = day, date = 2020-06-02, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.161s
INFO [2020-06-15 00:58:46] 8387   Archived website id 1, period = day, date = 2020-05-29, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.162s
INFO [2020-06-15 00:58:48] 8387   Archived website id 1, period = week, date = 2020-06-01, segment = '', plugin = CustomReports, report = 14, 8 visits found. Time elapsed: 1.181s
INFO [2020-06-15 00:58:48] 8387   Archived website id 1, period = week, date = 2020-05-25, segment = '', plugin = CustomReports, report = 14, 5 visits found. Time elapsed: 1.182s
INFO [2020-06-15 00:58:48] 8387   Archived website id 1, period = week, date = 2020-05-18, segment = '', plugin = CustomReports, report = 14, 9 visits found. Time elapsed: 1.182s
INFO [2020-06-15 00:58:58] 8387   Archived website id 1, period = month, date = 2020-06-01, segment = '', 3 visits found. Time elapsed: 9.258s
INFO [2020-06-15 00:58:59] 8387   Archived website id 1, period = day, date = 2020-05-28, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 0.860s
INFO [2020-06-15 00:58:59] 8387   Archived website id 1, period = day, date = 2020-05-27, segment = '', plugin = CustomReports, report = 14, 3 visits found. Time elapsed: 1.173s
INFO [2020-06-15 00:58:59] 8387   Archived website id 1, period = day, date = 2020-05-26, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 1.173s
INFO [2020-06-15 00:59:01] 8387   Archived website id 1, period = day, date = 2020-05-22, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.854s
INFO [2020-06-15 00:59:01] 8387   Archived website id 1, period = day, date = 2020-05-21, segment = '', plugin = CustomReports, report = 14, 7 visits found. Time elapsed: 1.164s
INFO [2020-06-15 00:59:01] 8387   Archived website id 1, period = day, date = 2020-05-18, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.165s
INFO [2020-06-15 00:59:03] 8387   Archived website id 1, period = week, date = 2020-05-11, segment = '', plugin = CustomReports, report = 14, 4 visits found. Time elapsed: 0.890s
INFO [2020-06-15 00:59:03] 8387   Archived website id 1, period = week, date = 2020-05-04, segment = '', plugin = CustomReports, report = 14, 4 visits found. Time elapsed: 1.205s
INFO [2020-06-15 00:59:03] 8387   Archived website id 1, period = week, date = 2020-04-27, segment = '', plugin = CustomReports, report = 14, 7 visits found. Time elapsed: 1.205s
INFO [2020-06-15 00:59:04] 8387   Archived website id 1, period = day, date = 2020-05-08, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.864s
INFO [2020-06-15 00:59:04] 8387   Archived website id 1, period = day, date = 2020-05-05, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.865s
INFO [2020-06-15 00:59:04] 8387   Archived website id 1, period = day, date = 2020-04-29, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.186s
INFO [2020-06-15 00:59:06] 8387   Archived website id 1, period = month, date = 2020-05-01, segment = '', plugin = CustomReports, report = 14, 22 visits found. Time elapsed: 1.189s
INFO [2020-06-15 00:59:06] 8387   Archived website id 1, period = month, date = 2020-04-01, segment = '', plugin = CustomReports, report = 14, 19 visits found. Time elapsed: 1.498s
INFO [2020-06-15 00:59:06] 8387   Archived website id 1, period = month, date = 2020-03-01, segment = '', plugin = CustomReports, report = 14, 26 visits found. Time elapsed: 1.498s
INFO [2020-06-15 00:59:08] 8387   Archived website id 1, period = day, date = 2020-04-28, segment = '', plugin = CustomReports, report = 14, 3 visits found. Time elapsed: 0.862s
INFO [2020-06-15 00:59:08] 8387   Archived website id 1, period = day, date = 2020-04-26, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.862s
INFO [2020-06-15 00:59:08] 8387   Archived website id 1, period = day, date = 2020-04-24, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.176s
INFO [2020-06-15 00:59:09] 8387   Archived website id 1, period = day, date = 2020-04-23, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 0.855s
INFO [2020-06-15 00:59:09] 8387   Archived website id 1, period = day, date = 2020-04-22, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 1.168s
INFO [2020-06-15 00:59:09] 8387   Archived website id 1, period = day, date = 2020-04-14, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.168s
INFO [2020-06-15 00:59:11] 8387   Archived website id 1, period = week, date = 2020-04-20, segment = '', plugin = CustomReports, report = 14, 7 visits found. Time elapsed: 1.163s
INFO [2020-06-15 00:59:11] 8387   Archived website id 1, period = week, date = 2020-04-13, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 1.163s
INFO [2020-06-15 00:59:11] 8387   Archived website id 1, period = week, date = 2020-04-06, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 1.164s
INFO [2020-06-15 00:59:13] 8387   Archived website id 1, period = day, date = 2020-04-09, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.876s
INFO [2020-06-15 00:59:13] 8387   Archived website id 1, period = day, date = 2020-04-08, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 1.183s
INFO [2020-06-15 00:59:13] 8387   Archived website id 1, period = day, date = 2020-04-06, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.183s
INFO [2020-06-15 00:59:14] 8387   Archived website id 1, period = day, date = 2020-03-31, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.852s
INFO [2020-06-15 00:59:14] 8387   Archived website id 1, period = day, date = 2020-03-27, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.161s
INFO [2020-06-15 00:59:14] 8387   Archived website id 1, period = day, date = 2020-03-26, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 1.161s
INFO [2020-06-15 00:59:16] 8387   Archived website id 1, period = week, date = 2020-03-30, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 0.864s
INFO [2020-06-15 00:59:16] 8387   Archived website id 1, period = week, date = 2020-03-23, segment = '', plugin = CustomReports, report = 14, 5 visits found. Time elapsed: 1.176s
INFO [2020-06-15 00:59:16] 8387   Archived website id 1, period = week, date = 2020-03-16, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 1.176s
INFO [2020-06-15 00:59:17] 8387   Archived website id 1, period = day, date = 2020-03-24, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.851s
INFO [2020-06-15 00:59:17] 8387   Archived website id 1, period = day, date = 2020-03-18, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.158s
INFO [2020-06-15 00:59:17] 8387   Archived website id 1, period = day, date = 2020-03-17, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 1.158s
INFO [2020-06-15 00:59:19] 8387   Archived website id 1, period = day, date = 2020-03-16, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 0.860s
INFO [2020-06-15 00:59:19] 8387   Archived website id 1, period = day, date = 2020-03-09, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.860s
INFO [2020-06-15 00:59:19] 8387   Archived website id 1, period = day, date = 2020-03-07, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.170s
INFO [2020-06-15 00:59:20] 8387   Archived website id 1, period = week, date = 2020-03-09, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 1.187s
INFO [2020-06-15 00:59:20] 8387   Archived website id 1, period = week, date = 2020-03-02, segment = '', plugin = CustomReports, report = 14, 17 visits found. Time elapsed: 1.187s
INFO [2020-06-15 00:59:20] 8387   Archived website id 1, period = week, date = 2020-02-24, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 1.187s
INFO [2020-06-15 00:59:22] 8387   Archived website id 1, period = day, date = 2020-03-06, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 1.193s
INFO [2020-06-15 00:59:22] 8387   Archived website id 1, period = day, date = 2020-03-05, segment = '', plugin = CustomReports, report = 14, 12 visits found. Time elapsed: 1.193s
INFO [2020-06-15 00:59:22] 8387   Archived website id 1, period = day, date = 2020-03-04, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 1.193s
INFO [2020-06-15 00:59:24] 8387   Archived website id 1, period = day, date = 2020-02-28, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.895s
INFO [2020-06-15 00:59:24] 8387   Archived website id 1, period = day, date = 2020-02-24, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.202s
INFO [2020-06-15 00:59:24] 8387   Archived website id 1, period = day, date = 2020-02-20, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.202s
INFO [2020-06-15 00:59:25] 8387   Archived website id 1, period = day, date = 2020-02-18, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.866s
INFO [2020-06-15 00:59:25] 8387   Archived website id 1, period = day, date = 2020-02-17, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 1.175s
INFO [2020-06-15 00:59:25] 8387   Archived website id 1, period = day, date = 2020-02-14, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.175s
INFO [2020-06-15 00:59:27] 8387   Archived website id 1, period = week, date = 2020-02-17, segment = '', plugin = CustomReports, report = 14, 8 visits found. Time elapsed: 1.172s
INFO [2020-06-15 00:59:27] 8387   Archived website id 1, period = week, date = 2020-02-10, segment = '', plugin = CustomReports, report = 14, 10 visits found. Time elapsed: 1.172s
INFO [2020-06-15 00:59:27] 8387   Archived website id 1, period = week, date = 2020-02-03, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 1.172s
INFO [2020-06-15 00:59:29] 8387   Archived website id 1, period = day, date = 2020-02-13, segment = '', plugin = CustomReports, report = 14, 5 visits found. Time elapsed: 0.852s
INFO [2020-06-15 00:59:29] 8387   Archived website id 1, period = day, date = 2020-02-12, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 1.163s
INFO [2020-06-15 00:59:29] 8387   Archived website id 1, period = day, date = 2020-02-11, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 1.163s
INFO [2020-06-15 00:59:30] 8387   Archived website id 1, period = day, date = 2020-02-04, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.858s
INFO [2020-06-15 00:59:30] 8387   Archived website id 1, period = day, date = 2020-01-31, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.858s
INFO [2020-06-15 00:59:30] 8387   Archived website id 1, period = day, date = 2020-01-29, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.167s
INFO [2020-06-15 00:59:32] 8387   Archived website id 1, period = month, date = 2020-02-01, segment = '', plugin = CustomReports, report = 14, 21 visits found. Time elapsed: 1.165s
INFO [2020-06-15 00:59:32] 8387   Archived website id 1, period = month, date = 2020-01-01, segment = '', plugin = CustomReports, report = 14, 10 visits found. Time elapsed: 1.165s
INFO [2020-06-15 00:59:32] 8387   Archived website id 1, period = month, date = 2019-12-01, segment = '', plugin = CustomReports, report = 14, 20 visits found. Time elapsed: 1.471s
INFO [2020-06-15 00:59:34] 8387   Archived website id 1, period = day, date = 2020-01-27, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.848s
INFO [2020-06-15 00:59:34] 8387   Archived website id 1, period = day, date = 2020-01-23, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.848s
INFO [2020-06-15 00:59:34] 8387   Archived website id 1, period = day, date = 2020-01-13, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.158s
INFO [2020-06-15 00:59:35] 8387   Archived website id 1, period = week, date = 2020-01-27, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 1.177s
INFO [2020-06-15 00:59:35] 8387   Archived website id 1, period = week, date = 2020-01-20, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 1.177s
INFO [2020-06-15 00:59:35] 8387   Archived website id 1, period = week, date = 2020-01-13, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.178s
INFO [2020-06-15 00:59:37] 8387   Archived website id 1, period = day, date = 2020-01-08, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.855s
INFO [2020-06-15 00:59:37] 8387   Archived website id 1, period = day, date = 2020-01-07, segment = '', plugin = CustomReports, report = 14, 3 visits found. Time elapsed: 1.169s
INFO [2020-06-15 00:59:37] 8387   Archived website id 1, period = day, date = 2020-01-02, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.170s
INFO [2020-06-15 00:59:39] 8387   Archived website id 1, period = week, date = 2020-01-06, segment = '', plugin = CustomReports, report = 14, 6 visits found. Time elapsed: 1.171s
INFO [2020-06-15 00:59:39] 8387   Archived website id 1, period = week, date = 2019-12-30, segment = '', plugin = CustomReports, report = 14, 7 visits found. Time elapsed: 1.171s
INFO [2020-06-15 00:59:39] 8387   Archived website id 1, period = week, date = 2019-12-23, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 1.172s
INFO [2020-06-15 00:59:40] 8387   Archived website id 1, period = day, date = 2020-01-01, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 0.858s
INFO [2020-06-15 00:59:40] 8387   Archived website id 1, period = day, date = 2019-12-31, segment = '', plugin = CustomReports, report = 14, 4 visits found. Time elapsed: 1.170s
INFO [2020-06-15 00:59:40] 8387   Archived website id 1, period = day, date = 2019-12-24, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.170s
INFO [2020-06-15 00:59:52] 8387   Archived website id 1, period = year, date = 2020-01-01, segment = '', 109 visits found. Time elapsed: 11.660s
INFO [2020-06-15 00:59:54] 8387   Archived website id 1, period = day, date = 2019-12-23, segment = '', plugin = CustomReports, report = 14, 2 visits found. Time elapsed: 0.891s
INFO [2020-06-15 00:59:54] 8387   Archived website id 1, period = day, date = 2019-12-19, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.210s
INFO [2020-06-15 00:59:54] 8387   Archived website id 1, period = day, date = 2019-12-10, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.210s
INFO [2020-06-15 00:59:55] 8387   Archived website id 1, period = week, date = 2019-12-16, segment = '', plugin = CustomReports, report = 14, 4 visits found. Time elapsed: 1.191s
INFO [2020-06-15 00:59:55] 8387   Archived website id 1, period = week, date = 2019-12-09, segment = '', plugin = CustomReports, report = 14, 3 visits found. Time elapsed: 1.191s
INFO [2020-06-15 00:59:55] 8387   Archived website id 1, period = week, date = 2019-12-02, segment = '', plugin = CustomReports, report = 14, 4 visits found. Time elapsed: 1.192s
INFO [2020-06-15 00:59:57] 8387   Archived website id 1, period = day, date = 2019-12-08, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 0.879s
INFO [2020-06-15 00:59:57] 8387   Archived website id 1, period = day, date = 2019-12-04, segment = '', plugin = CustomReports, report = 14, 0 visits found. Time elapsed: 1.190s
INFO [2020-06-15 00:59:57] 8387   Archived website id 1, period = day, date = 2019-12-03, segment = '', plugin = CustomReports, report = 14, 1 visits found. Time elapsed: 1.191s
INFO [2020-06-15 00:59:58] 8387   Archived website id 1, period = week, date = 2019-11-25, segment = '', plugin = CustomReports, report = 14, 689 visits found. Time elapsed: 0.915s
INFO [2020-06-15 01:00:02] 8387   Archived website id 1, period = year, date = 2019-01-01, segment = '', plugin = CustomReports, report = 14, 1298 visits found. Time elapsed: 3.098s
INFO [2020-06-15 01:00:02] 8387   Finished archiving for site 1, 102 API requests, Time elapsed: 82.389s [1 / 7 done]
INFO [2020-06-15 01:00:02] 8387   Start processing archives for site 2.
```


it also looked like some days were script so I ran it verbose and noticed indeed some days are maybe skipped see `Found archive with different period than others in concurrent batch`. However, the invalidations table was still empty afterwards so not sure if these skipped ones should have been readded? Here's the verbose output:

https://justpaste.it/75r3a

Another issue I'm not 100% sure about it. I looked into the archive tables and under the most recent archives I would have only expected to see `CustomReports_customreport_16_0_nb_visits` and `done` archive names. But it did also archive several other metrics see screenshot. I get we maybe need for our standard logic the nb_visits (ideally we wouldn't need to archive any of that) but maybe we could ignore the other metrics?
image

diosmosis:
@tsteur 

I think I know what the ordering issue is, will look for a fix.

> Another issue I'm not 100% sure about it. I looked into the archive tables and under the most recent archives I would have only expected to see CustomReports_customreport_16_0_nb_visits and done archive names. But it did also archive several other metrics see screenshot. I get we maybe need for our standard logic the nb_visits (ideally we wouldn't need to archive any of that) but maybe we could ignore the other metrics?

I suspect that is due to the Loader.php code needing the visits in order to check if we can skip the archive. I recently added an optimization `Loader::canSkipThisArchive()` which checks if there is a visit in log_visit, I wonder if we can remove the other one that invokes VisitsSummary (in `Loader::prepareCoreMetricsArchive()`)? Will there ever be a need to find exactly how many visits there are before archiving, if we know we can't (or can) find at least one visit in the log table?

core/Updates/4.0.0-b2.php
```diff
@@ -44,6 +44,7 @@ public function getMigrations(Updater $updater)
             'period' => 'TINYINT UNSIGNED NOT NULL',
             'ts_invalidated' => 'DATETIME NOT NULL',
             'status' => 'TINYINT(1) UNSIGNED DEFAULT 0',
+            'report' => 'VARCHAR(255) NOT NULL',
```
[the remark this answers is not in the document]
diosmosis: 👍

config/global.ini.php
```diff
@@ -742,6 +742,10 @@
 ; Default is 1.
 enable_tracking_failures_notification = 1
 
+; Controls how many months in the past reports are re-archived for plugins that support
+; doing this (such as CustomReports). Set to 0 to disable the feature. Default is 6.
+rearchive_reports_in_past_last_n_months = 6
```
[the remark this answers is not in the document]
diosmosis: So should `site_creation_time` be supported?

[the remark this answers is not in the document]
tsteur: Yes definitely.

core/Archive/ArchiveInvalidator.php
```diff
@@ -413,12 +429,54 @@ public function markArchivesOverlappingRangeAsInvalidated(array $idSites, array
...
+     * @api
+     */
+    public function reArchiveReport(array $idSites, string $plugin, string $report = null, int $lastNMonthsToInvalidate = null)
+    {
+        $lastNMonthsToInvalidate = $lastNMonthsToInvalidate ?: Config::getInstance()->General['rearchive_reports_in_past_last_n_months'];
+        if (empty($lastNMonthsToInvalidate)) {
+            return;
+        }
+
+        $lastNMonthsToInvalidate = (int) substr($lastNMonthsToInvalidate, 4);
```
tsteur, line 451: btw not sure if `markArchivesAsInvalidated ` does it maybe but if the site was created only 60 days ago, and configured is to archive 6 months back, then we would only invalidate the last 60days.

tsteur:
> , I wonder if we can remove the other one that invokes VisitsSummary (in Loader::prepareCoreMetricsArchive())? Will there ever be a need to find exactly how many visits there are before archiving, if we know we can't (or can) find at least one visit in the log table?

i'm not 100% sure about why it's there but I presume it might be still needed for browser archiving maybe as it can basically ignore running the entire archive if there are no visits?

Wondering... if we archive only a specific report, in `aggregateDayVisitsMetrics` could we maybe at least skip `$this->archiveProcessor->insertNumericRecords($metrics);`? So at least we safe some IO etc if we still have to query that data? Or only insert `nb_visits` if that is needed for some reason?

The code looks there also quite tricky and not fully understanding it why we potentially run `callAggregateCoreMetrics` twice https://github.com/matomo-org/matomo/blob/4.x-dev/core/ArchiveProcessor/Loader.php#L162-L181 or why it would run it in the first method and in other cases in the second method...

I'm actually not sure why the loader writes these nb_visits as the returned value from `Loader::prepareArchive()` seems to be not really used (except for now in the new ArchiveReports method) but maybe that all wouldn't even be needed? Guess it's needed to determine if we accept the `idArchive` or not but this might be actually wrong behaviour. Eg if there are no visits in https://github.com/matomo-org/matomo/blob/4.x-dev/core/ArchiveProcessor/Loader.php#L138-L140, then it maybe shouldn't ignore the `idarchive`? If I understand it correct it also means it would try to archive every time if there is no visit instead of respecting the configured TTL and eg only archive again 900s later?

Sorry I'm not of better help.

core/Archive/ArchiveInvalidator.php
```diff
@@ -413,12 +429,54 @@ public function markArchivesOverlappingRangeAsInvalidated(array $idSites, array
...
+     * TODO test
+     * @param int[] $idSite
+     * @param Date $date1
+     * @param Date $date2
+     * @param string $plugin
+     * @param string|null $report
+     * @throws \Exception
+     * @api
+     */
+    public function reArchiveReport(array $idSites, string $plugin, string $report = null, int $lastNMonthsToInvalidate = null)
```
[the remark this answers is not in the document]
diosmosis: @tsteur So for now, requirement is to also invalidate segments, and use `max(dateFor(process_new_segments_from), dateFor(rearchive_reports_in_past_last_n_days))` as the start date?

core/Archive/ArchiveInvalidator.php
diosmosis: 👍

diosmosis:
> i'm not 100% sure about why it's there but I presume it might be still needed for browser archiving maybe as it can basically ignore running the entire archive if there are no visits?

It's needed for when segments are archived. If we find we have 0 visits, then we shouldn't initiate archiving. However, maybe we can use the segment table optimization for this?

If we launch archiving for a segment, then we have to do this query to create the temp table, and even if there are no visits, to get nb_visits, we'd have to do this query as well, correct? So we can do this once, then if there are no rows in the temp table skip archiving for the segment. This would mean not having to archive those VisitsSummary metrics first.

Only time it wouldn't be as fast is if for some reason there already is an nb_visits parameter. I guess we can check for that pretty easily though. @tsteur what do you think about this? I guess I can just try to see if it works.

core/Archive/ArchiveInvalidator.php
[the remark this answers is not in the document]
tsteur: @diosmosis for segments it would be maybe rather `min(dateFor(process_new_segments_from), dateFor(rearchive_reports_in_past_last_n_days))`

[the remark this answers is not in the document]
diosmosis: do you mean "use the most recent of he two dates", ie, the greatest timestamp, or the oldest?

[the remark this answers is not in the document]
tsteur: As mentioned in examples if `process_new_segments_from = 100 days` and `rearchive_reports_in_past_last_n_days = 200 days`, then we would process only 100 days back.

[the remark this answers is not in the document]
tsteur: Basically the most recent of the two dates.

diosmosis:
> Wondering... if we archive only a specific report, in aggregateDayVisitsMetrics could we maybe at least skip $this->archiveProcessor->insertNumericRecords($metrics);? So at least we safe some IO etc if we still have to query that data? Or only insert nb_visits if that is needed for some reason?

I think maybe this isn't needed. VisitsSummary metrics should be present already when rearchiving data in the past. If it's not there, then it would have to be development setup. So I think this is a bit of a non-issue?

tsteur:
> I think maybe this isn't needed. VisitsSummary metrics should be present already when rearchiving data in the past. If it's not there, then it would have to be development setup. So I think this is a bit of a non-issue?

Not 100% what you mean with non-issue? I reckon could try to not write these archive entries and assume it might not be needed but probably the `nb_visit` will be needed when fetching that archive next time in the loader.

diosmosis:
> Not 100% what you mean with non-issue? I reckon could try to not write these archive entries and assume it might not be needed but probably the nb_visit will be needed when fetching that archive next time in the loader.

I mean in a normal setup, when archiving the last month for a plugin, nb_visits should already exist. Or was there already a valid archive when you tested and it initiated another done.VisitsSummary archive too?

tsteur:
>  Or was there already a valid archive when you tested and it initiated another done.VisitsSummary archive too?

There was already a valid generic archive. This might not always be the case though. That's why I reckon we always have to write nb_visits but not really the others. I wonder if ideally it could check if there is a valid generic archive already, and if so, use that nb_visits value instead of querying this metrics again from the raw data.

However, this would be maybe not worth the effort for now as it would require quite some extra code and trying to find the idArchive of the generic archive maybe etc. 

I reckon if we do a plugin specific (re)archive, then we could still run that query for core metrics, but at least only insert `nb_visits` metric, not the others.

diosmosis:
@tsteur this can be reviewed again. I ensured the latest archive is always used (tested by new LoaderTest tests). It will still insert other metrics because otherwise matomo will think the value for them is 0.

tsteur (commented):
Sorry @diosmosis I reviewed this 6 days ago and apparently it didn't sent the comments and the big long comment I wrote!

I mentioned few things but mostly one issue (and back then I even had the verbose logs of core:archive etc) which was related to this:

Configure:
```
[General]
```

* Track a request
* Run core:archive
* It will archive day, week, month, year
* Track a request
* Run core:archive again
* Now it will archive day, week, month, year again even though it should archive only day because the 900s TTL for week/month/year. Otherwise we'd archive these periods too often.

core/CliMulti.php
```diff
@@ -366,12 +367,12 @@ private function executeNotAsyncHttp($url, Output $output)
...
             }
 
-            $url .= 'token_auth=' . $tokenAuth;
+            $requestBody = 'token_auth=' . $tokenAuth;
         }
 
         try {
             $this->logger->debug("Execute HTTP API request: "  . $url);
-            $response = Http::sendHttpRequestBy('curl', $url, $timeout = 0, $userAgent = null, $destinationPath = null, $file = null, $followDepth = 0, $acceptLanguage = false, $this->acceptInvalidSSLCertificate);
+            $response = Http::sendHttpRequestBy('curl', $url, $timeout = 0, $userAgent = null, $destinationPath = null, $file = null, $followDepth = 0, $acceptLanguage = false, $this->acceptInvalidSSLCertificate, false, false, 'POST', null, null, $requestBody, $forcePost = true, [], $forcePost = true);
```
tsteur, line 375: @diosmosis I think there's one `forcePost` too much? Maybe the first one isn't needed as it passes `true` for the additional headers?

diosmosis: Yes, I removed this locally forgot to push.

diosmosis: @tsteur pushed

core/Archive/ArchiveInvalidator.php
```diff
@@ -228,16 +238,30 @@ private function deleteOptionLike($id)
...
     {
+        $plugin = null;
+        if ($name && strpos($name, '.') !== false) {
+            list($plugin) = explode('.', $name);
+        }
+
+        if ($plugin
+            && !Manager::getInstance()->isPluginActivated($plugin)
+        ) {
+            throw new \Exception("Plugin is not activated: '$plugin'");
```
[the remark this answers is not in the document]
diosmosis: This specific case shouldn't be an issue, since once the entries are in archive_invalidations, we don't try to invalidate for the plugin.

If the exception could occur during core:archive then the command would fail... but CronArchive doesn't try to invalidate a specific plugin.

In your specific case, the remaining 900 would be pulled, then in CronArchive we would see the plugin is disabled, so we would skip and remove the invalidation entry.

core/CronArchive/SegmentArchiving.php
```diff
@@ -79,6 +79,11 @@ public function __construct($processNewSegmentsFrom, $beginningOfTimeLastNInYear
     }
 
     public function getSegmentArchivesToInvalidateForNewSegments($idSite)
+    {
+        return $this->getSegmentArchivesToInvalidate($idSite, true);
+    }
+
+    public function getSegmentArchivesToInvalidate($idSite, $checkOnlyForNewSegments = false)
```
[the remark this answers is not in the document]
diosmosis: That should be the same code as before. In `getCreatedTimeOfSegment()` there is a check for site via `isSegmentForSite()`. That method checks if `enable_only_idsite == 0 || enable_only_idsite == $idSite`.

diosmosis:
> Reason I'm thinking about is eg when we set up a new cloud account. Then we install several plugins, eg Funnels. It would then potentially create heaps of archive invalidations maybe even though it's a fresh Matomo and nothing was tracked yet.

You are referring to the `Loader::canSkipThisArchive()` optimization, correct? It's currently used in Archive.php as well as in core:archive to skip invalidations that don't need to be archived. We can avoid invalidating in that case too I suppose. I'll add that, let me know if that was your intention.

core/DataAccess/ArchiveWriter.php
```diff
@@ -273,12 +289,16 @@ private function flushSpool($valueType)
 
     protected function getInsertRecordBind()
     {
+        $now = date("Y-m-d H:i:s");
```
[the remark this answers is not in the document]
diosmosis: 👍

diosmosis:
> Reason I'm thinking about is eg when we set up a new cloud account. Then we install several plugins, eg Funnels. It would then potentially create heaps of archive invalidations maybe even though it's a fresh Matomo and nothing was tracked yet.

Added.

> Actually I think I found it in https://github.com/matomo-org/matomo/pull/15889/files#diff-f0dc4d4d133f84005fcebc1e10f155c0R202 and it's there so might be fine to ignore.

Is this referring to what I added? That is in the core:archive side specifically for segment archiving, there is also `Loader::canSkipArchive()` which is run during core:archive and in Archive.php, so empty archives won't be created regardless. But it would still be a small boost to performance to not have to consider these sites anyway.

diosmosis:
@tsteur I did some debugging for the ttl issue and found the issue,

> Configure:
> 
> [General]
> time_before_today_archive_considered_outdated = 1
> 
>    Track a request
>    Run core:archive
>    It will archive day, week, month, year
>    Track a request
>    Run core:archive again
>    Now it will archive day, week, month, year again even though it should archive only day because the 900s TTL for week/month/year. Otherwise we'd archive these periods too often.

The ttl for a period is determined by `Rules::getPeriodArchiveTimeToLiveDefault()`. If the period is day, the today TTL is used. Otherwise the `time_before_%s_archive_considered_outdated` TTL is used. However, if that config isn't set, then the today TTL is used.

I'm guessing this logic is incorrect... can you clarify how you think it should work? `today ttl if date == today` and `period ttl otherwise` and maybe 900 if period ttl is not defined in config? the `today ttl` can also be the UI setting option.

tsteur:
@diosmosis how the config behaves is fine. (that `today` is used when no week, month, year setting is specified)

I just tested it this way:

```ini
[General]
time_before_week_archive_considered_outdated = 900
time_before_year_archive_considered_outdated = 900
```

using the same flow and would have expected `week` and `year` a larger TTL to be recognised.

I basically did this:
* run core:archive
* track a reuqest
* run core:archive again
  * Was expecting only today and month to be archived, but not week and year since it has been only few seconds and not 900 seconds

Here's the verbose output if that helps:

```log
START
Starting Matomo reports archiving...
Checking for queued invalidations...
  Will invalidate archived reports for today in site ID = 1's timezone (2020-07-21 00:00:00).
Found existing valid archive for yesterday, skipping invalidation...
DEBUG [2020-07-21 01:13:06] 84702  Earliest created time of segment 'visitIp>=15.15.15.15;visitIp<=20.20.20.20,visitIp==21.21.21.21,visitIp==22.22.22.22,actionUrl=@foo' w/ idSite = 1 is found to be 2019-02-04. Latest edit time is found to be 2019-02-04.
DEBUG [2020-07-21 01:13:06] 84702  Earliest created time of segment 'actions>=2' w/ idSite = 7 is found to be 2020-03-30. Latest edit time is found to be 2020-03-30.
Done invalidating
Start processing archives for site 1.
DEBUG [2020-07-21 01:13:06] 84702  Processing invalidation: [idinvalidation = 2365, idsite = 1, period = day(2020-07-21 - 2020-07-21), name = done].
DEBUG [2020-07-21 01:13:06] 84702  Found archive with intersecting period with others in concurrent batch, skipping until next batch: [idinvalidation = 2366, idsite = 1, period = week(2020-07-20 - 2020-07-26), name = done]
DEBUG [2020-07-21 01:13:06] 84702  Found archive with intersecting period with others in concurrent batch, skipping until next batch: [idinvalidation = 2367, idsite = 1, period = month(2020-07-01 - 2020-07-31), name = done]
DEBUG [2020-07-21 01:13:06] 84702  Found archive with intersecting period with others in concurrent batch, skipping until next batch: [idinvalidation = 2368, idsite = 1, period = year(2020-01-01 - 2020-12-31), name = done]
DEBUG [2020-07-21 01:13:06] 84702  No next invalidated archive.
DEBUG [2020-07-21 01:13:06] 84702  Starting archiving for ?module=API&method=CoreAdminHome.archiveReports&idSite=1&period=day&date=2020-07-21&format=json&trigger=archivephp
DEBUG [2020-07-21 01:13:07] 84702  Running command: /usr/local/Cellar/php/7.4.6_1/bin/php -q  /Users/thomassteur/Development/piwik/console climulti:request -q --matomo-domain='' --superuser 'module=API&method=CoreAdminHome.archiveReports&idSite=1&period=day&date=2020-07-21&format=json&trigger=archivephp&pid=9ccd666e3db84c5560cae4743d5bc094b7704451dae04f55fe446b14558f401dcc46d5c8ab73ca0baf2608f4b3c2a805852e0&runid=84702' > /Users/thomassteur/Development/piwik/tmp/api5.matomo.cloud/climulti/9ccd666e3db84c5560cae4743d5bc094b7704451dae04f55fe446b14558f401dcc46d5c8ab73ca0baf2608f4b3c2a805852e0.output 2>&1 &
INFO [2020-07-21 01:13:07] 84702  Archived website id 1, period = day, date = 2020-07-21, segment = '', 0 visits found. Time elapsed: 0.595s
DEBUG [2020-07-21 01:13:07] 84702  Processing invalidation: [idinvalidation = 2366, idsite = 1, period = week(2020-07-20 - 2020-07-26), name = done].
DEBUG [2020-07-21 01:13:07] 84702  Found archive with intersecting period with others in concurrent batch, skipping until next batch: [idinvalidation = 2367, idsite = 1, period = month(2020-07-01 - 2020-07-31), name = done]
DEBUG [2020-07-21 01:13:07] 84702  Found archive with intersecting period with others in concurrent batch, skipping until next batch: [idinvalidation = 2368, idsite = 1, period = year(2020-01-01 - 2020-12-31), name = done]
DEBUG [2020-07-21 01:13:07] 84702  No next invalidated archive.
DEBUG [2020-07-21 01:13:07] 84702  Starting archiving for ?module=API&method=CoreAdminHome.archiveReports&idSite=1&period=week&date=2020-07-20&format=json&trigger=archivephp
DEBUG [2020-07-21 01:13:07] 84702  Running command: /usr/local/Cellar/php/7.4.6_1/bin/php -q  /Users/thomassteur/Development/piwik/console climulti:request -q --matomo-domain='' --superuser 'module=API&method=CoreAdminHome.archiveReports&idSite=1&period=week&date=2020-07-20&format=json&trigger=archivephp&pid=c6831c061564d18e27768d2214fd0040eebe919e967c4d93012e4ae074da0a38b735e84d4e156d3786caea624933965cef860&runid=84702' > /Users/thomassteur/Development/piwik/tmp/api5.matomo.cloud/climulti/c6831c061564d18e27768d2214fd0040eebe919e967c4d93012e4ae074da0a38b735e84d4e156d3786caea624933965cef860.output 2>&1 &
INFO [2020-07-21 01:13:08] 84702  Archived website id 1, period = week, date = 2020-07-20, segment = '', 1 visits found. Time elapsed: 1.174s
DEBUG [2020-07-21 01:13:09] 84702  Processing invalidation: [idinvalidation = 2367, idsite = 1, period = month(2020-07-01 - 2020-07-31), name = done].
DEBUG [2020-07-21 01:13:09] 84702  Found archive with intersecting period with others in concurrent batch, skipping until next batch: [idinvalidation = 2368, idsite = 1, period = year(2020-01-01 - 2020-12-31), name = done]
DEBUG [2020-07-21 01:13:09] 84702  No next invalidated archive.
DEBUG [2020-07-21 01:13:09] 84702  Starting archiving for ?module=API&method=CoreAdminHome.archiveReports&idSite=1&period=month&date=2020-07-01&format=json&trigger=archivephp
DEBUG [2020-07-21 01:13:09] 84702  Running command: /usr/local/Cellar/php/7.4.6_1/bin/php -q  /Users/thomassteur/Development/piwik/console climulti:request -q --matomo-domain='' --superuser 'module=API&method=CoreAdminHome.archiveReports&idSite=1&period=month&date=2020-07-01&format=json&trigger=archivephp&pid=08097f6b2594b0d8d8870c0e28928bd1d3201024cf88a37d98bd97500465290faa609dd0685c2033157118fd01760471d36d0&runid=84702' > /Users/thomassteur/Development/piwik/tmp/api5.matomo.cloud/climulti/08097f6b2594b0d8d8870c0e28928bd1d3201024cf88a37d98bd97500465290faa609dd0685c2033157118fd01760471d36d0.output 2>&1 &
INFO [2020-07-21 01:13:20] 84702  Archived website id 1, period = month, date = 2020-07-01, segment = '', 6 visits found. Time elapsed: 11.088s
DEBUG [2020-07-21 01:13:20] 84702  Processing invalidation: [idinvalidation = 2368, idsite = 1, period = year(2020-01-01 - 2020-12-31), name = done].
DEBUG [2020-07-21 01:13:20] 84702  No next invalidated archive.
DEBUG [2020-07-21 01:13:20] 84702  Starting archiving for ?module=API&method=CoreAdminHome.archiveReports&idSite=1&period=year&date=2020-01-01&format=json&trigger=archivephp
DEBUG [2020-07-21 01:13:20] 84702  Running command: /usr/local/Cellar/php/7.4.6_1/bin/php -q  /Users/thomassteur/Development/piwik/console climulti:request -q --matomo-domain='' --superuser 'module=API&method=CoreAdminHome.archiveReports&idSite=1&period=year&date=2020-01-01&format=json&trigger=archivephp&pid=589576f25540db10b81893639dd3e920d3cf17bb0a07d67bc1484c01a9cd11462f629b38220baf5c29ae302101fc1fd2a6000&runid=84702' > /Users/thomassteur/Development/piwik/tmp/api5.matomo.cloud/climulti/589576f25540db10b81893639dd3e920d3cf17bb0a07d67bc1484c01a9cd11462f629b38220baf5c29ae302101fc1fd2a6000.output 2>&1 &
INFO [2020-07-21 01:14:07] 84702  Archived website id 1, period = year, date = 2020-01-01, segment = '', 114 visits found. Time elapsed: 46.698s
DEBUG [2020-07-21 01:14:07] 84702  No next invalidated archive.
```

diosmosis:
@tsteur tests are fixed and the issue you mentioned is fixed, can you review again? there is one apparently random failure that keeps showing up, but not prioritizing it

core/DataAccess/ArchiveWriter.php
```diff
@@ -152,7 +164,12 @@ public function finalizeArchive()
         $numericTable = $this->getTableNumeric();
         $idArchive    = $this->getIdArchive();
 
-        $this->getModel()->updateArchiveStatus($numericTable, $idArchive, $this->doneFlag, self::DONE_OK);
+        $doneValue = $this->parameters->isPartialArchive() ? self::DONE_PARTIAL : self::DONE_OK;
+        $this->getModel()->updateArchiveStatus($numericTable, $idArchive, $this->doneFlag, $doneValue);
+
+        if ($this->parameters->isPartialArchive()) {
```
tsteur, line 170: @diosmosis shouldn't this be more like `if (!$this->parameters->isPartialArchive()) {`? If it's only a partial archive, can we really delete older archives? Or maybe I'm missing something?

tsteur: also do we need to check if `$this->earliestNow` is set?

core/DataAccess/ArchiveSelector.php
```diff
@@ -96,25 +101,17 @@ public static function getArchiveIdAndVisits(ArchiveProcessor\Parameters $params
             $minDatetimeArchiveProcessedUTC = Date::factory($minDatetimeArchiveProcessedUTC);
         }
 
-        if (!empty($minDatetimeArchiveProcessedUTC) && !is_object($minDatetimeArchiveProcessedUTC)) {
```
tsteur, line 99: @diosmosis was this removed on purpose?

tsteur (commented):
@diosmosis added 2 comments otherwise worked for me.

core/DataAccess/ArchiveWriter.php
diosmosis: 👍 will add a test covering both cases

core/DataAccess/ArchiveSelector.php
diosmosis: it's duplicate code from two weird merges. if you look right above it, the same code is there.

tsteur:
@diosmosis some tests seem to be failing now again. Can you check them out and merge once they pass? 🎉

diosmosis:
@tsteur build is passing now

Issue #12007, opened 2017-09-06
Helo command rejected: need fully-qualified hostname


netdesign:
Googling my issue doesn't give any useful result so I submit this issue with attached the solution (as a screenshot). I hope it is useful.

The ISSUE in the suject is reproducible if:

1) You use an external SMTP server to send Email reports
2) Your external SMTP server requires fully-qualified-hostname EHLO/HELO from client to grant sending

The SMTP transport class included in the Piwik sources hardcodes the hostname to "localhost" instead of the true local hostname as required by RFC2821, RFC1869. This way the SMTP server refuses to send the report email.

The "Zend_Mail_Transport_Smtp" constructor provides a way to set custom configuration (such as the hostname) but the Piwik interface isn't able to set this variable.

Here I attach my solution :)

![piwik-smtp-issue](https://user-images.githubusercontent.com/461899/30123023-e85ba500-9330-11e7-83af-11fb5b5554e9.png)

Globulopolis:
`SMTP server address` in `General settings` is not it?

netdesign:
No, it isn't. The "SMTP server address" in "General settings" is the address of the SMTP server while the issue is related to the client (during smtp connection the client connects to the server, the server introduces itself and waits for the client to send its hostname, this kind of handshake is called HELO / EHLO).

You can get many details about the smtp connection steps and commands on http://www.samlogic.net/articles/smtp-commands-reference.htm

Globulopolis:
@netdesign problem not in the core `Zend_Mail_Transport_Smtp`. I see that the `Mail->initSmtpTransport()` do not use `host` value at all.
Piwik config.ini file contain host for mail transport but this is not a client hostname.
Maybe I can do PR for this.

netdesign:
Dear @Globulopolis, add the "client hostname" support in the Piwik config.ini would be an option, but I'm sure that setting "localhost" as default value in "$this->_name" is a fault because "localhost" is meaningless in the SMTP protocol (every host is localhost).

Even if you add the discussed variable in the Piwik config.ini, what would be the default value? What would happen if the user don't set the variable and leaves the software to choose it for him? It most likely will be "localhost" and the probem is here again (except that Piwik forces the user to set the value during the installation).

What I'd suggest is to report the issue to the upstream Zend team and adding, as just you said, the variable to the Piwik config.ini.

[...]

Issue #8613, opened 2015-08-20
Sending email via gmail SMTP requires to "Allow less secure apps: YES" -> is solution to use PHPMailer?


mattab:
This issue is about improving this FAQ: [ How do I send Piwik emails using Gmail SMTP server?](https://piwik.org/faq/how-to/faq_93/)

While setting it up today, I noticed that i was getting an error when sending a scheduled report 

```
An error occured while sending 'HTML Email Report - 2.2015-08-20.101.en.html' to xxxxxxxxxxxxxxx@gmail.com. Error was '5.7.14 <https://accounts.google.com/ContinueSignIn?sarp=1&scc=1&plt=AKgnsbucv 5.7.14 -v6FnaqYk_3-5- 5.7.14 UQGVqm7Gvg1qMTL70m9YF2_kcoVxK-eN_d9aPG5hrzzCtiG 5.7.14 5edES-qhoae2YHyGn3bqFQnBQ 5.7.14 8BFlMxvMs_RBCEP62jWESOR-fuYk> Please log in via your web browser and 5.7.14 then try again. 5.7.14 Learn more at 5.7.14 https://support.google.com/mail/answer/78754 cd16sm31949626wib.19 - gsmtp'#0 [internal function]

```

To resolve this issue I had to go to:
- https://myaccount.google.com/security?pli=1#connectedapps
- then click `Allow less secure app` to `YES`

![allow less secure apps](https://cloud.githubusercontent.com/assets/466765/9394221/98f8522a-4787-11e5-8c3f-850d2b8c0bbb.png)

Steps
- what does it mean to be less secure app? could we change our FAQ instructions and/or implementation so that users do not need to do this extra step?
- if not possible, we should at least document in FAQ that user also need to change the setting on their google security account

tsteur:
I didn't have to do this I think. Did you use `ssl` and port `465`?

mattab:
> Did you use ssl and port 465?

yes, strange you didn't have to do this. I should try with another gmail account if I can reproduce

MagicFab:
Some more context:
http://security.stackexchange.com/questions/66025/what-are-the-dangers-of-allowing-less-secure-apps-to-access-my-google-account

To avoid this in the future and properly support using Google's SMTP, perhaps Piwik could implement OAuth 2: https://developers.google.com/gmail/oauth_overview?hl=en

MagicFab:
The FAQ could link to:

Original announcement:
[...]

PR #15891, opened 2020-04-30
Replace Zend Mail with PHPMailer

### LINKED ISSUE #14841: use a modern PHP library for SMTP
related to https://github.com/matomo-org/matomo/issues/8613 and https://github.com/matomo-org/matomo/issues/12007

inspired by https://forum.matomo.org/t/email-reports-not-sent-broken-pipe-errno-32/34047

Matomo uses a copy of Zend Mail that has last been updated in 2011 (in https://github.com/matomo-org/matomo/issues/2581)

While SMTP isn't the fastest developing protocol, a lot has changed in that time.

So instead of keeping an outdated library or reinventing the wheel, I'd recommend replacing it with a maintained, popular PHP library like PHPmailer or swiftmailer.

Bonus side effect: We would get proper error messages for hard to debug errors instead of things like `fwrite(): send of 6 bytes failed with errno=32 Broken pipe`.

sgiehl:
fixes #14841


Findus23:
If I could wish for one more minor feature:

Would it be possible to add the verbose logging (`$mail->SMTPDebug = SMTP::DEBUG_SERVER; `) to the output of the `core:test-email` command? This way debugging issues with E-Mail delivery would be quite a bit easier.

sgiehl:
@Findus23 thanks for the hint. Did not yet manually test anything only tried to get the tests passing. Guess we will need to do some manual testing with various settings before merging this. Could you maybe give me some common scenarios that should be tested? And maybe also some that didn't work with Zend_Mail if you know any?

core/Mail.php
```diff
@@ -99,15 +142,42 @@ public function setReplyTo($email, $name = null)
...
+        $mailConfig = Config::getInstance()->mail;
+
+        if (empty($mailConfig['host'])
+            || $mailConfig['transport'] != 'smtp'
+        ) {
             return;
         }
 
-        Mail::setDefaultTransport($tr);
+        $this->isSMTP();
```
Findus23, line 153: Would it be possible, to also use PHPMailer for delivering E-Mails without SMTP? That way we could also make sure that the endless amounts of edge-cases that can occur with mail() are handled properly

sgiehl: That should be the case I guess, but it's not needed to set smtp config if it's not using smtp

Findus23: Thanks for the info. It seems like I misread the function then.

Findus23:
I would at least test the following:
- [ ] sending an E-Mail using SMTP via G-Mail (https://github.com/matomo-org/matomo/issues/8613)
- [ ] sending an E-Mail using both SSL and TLS
- [ ] sending an E-Mail with an Attachment
- [ ] sending an E-Mail on a normal webhost without SMTP
- [ ] (probably not possible by you, but someone else could contribe): test if https://github.com/matomo-org/matomo/issues/12007 is fixed
- [ ] sending an E-Mail with various invalid settings (incorrect host/user/password) and check if the user gets a helpful error message (no more `fwrite(): send of 6 bytes failed with errno=32 Broken pipe`)
- [ ] sending an E-Mail with a Matomo not set to English and check if the error messages are localized

sgiehl:
@Findus23 Adding support for Gmail XOAuth would be possible using PHPMailer, but the required packages are not included by default. We would need to require [`thephpleague/oauth2-client`](https://packagist.org/packages/league/oauth2-client), which pulls in some other libs and polyfills. Wondering if it might be better, to put that into a separate plugin as I don't think using gmail is a common use case, is it? 
(The `Mail.send` event could be actually used to inject the oauth settings before sending)

@mattab @tsteur what do you think re OAuth support?

Findus23:
@sgiehl I think this XOAuth support is not needed as it would add a lot of overhead and potential confusion to users for little gain.
Sending E-Mails via Gmail should still be possible "the normal way": https://github.com/PHPMailer/PHPMailer/blob/master/examples/gmail.phps

If people really need it, creating a plugin for it is probably easier.

sgiehl:
Ok. So far I have tested:
* use Gmail with app specific token
* use custom smtp host with TLS   (don't have one still working with ssl)
* use local sendmail
* check attachments are correct (inline images and pdfs)
* check some errors with invalid configs
* ensure phpmailer errors are in current language

Seems to work so far and should be ready for a review now.

Findus23:
Sounds good. The rest could probably be done with general Matomo 4 testing duing the beta releases.

config/global.php
```diff
@@ -171,40 +171,6 @@
 
         return $ipsResolved;
     },
-    'Zend_Mail_Transport_Abstract' => function () {
-        $mailConfig = Config::getInstance()->mail;
```
tsteur, line 175: could this be maybe kept? We're using it on the cloud for example to inject some settings. Haven't looked how phpmailer handles this. Otherwise some DI setting that returns Emails settings could work too. Like 

```php
'mail.settings'  => function () {
   return Config::getInstance()->mail
}
```

We don't really directly want to change the mail config settings using DI on cloud (eg as these values might get saved in the config file etc)

core/Mail.php
```diff
@@ -61,34 +63,76 @@ public function setWrappedHtmlBody($body)
...
     /**
      * Sets the sender.
      *
      * @param string $email Email address of the sender.
      * @param null|string $name Name of the sender.
-     * @return Zend_Mail
+     * @return bool
      */
-    public function setFrom($email, $name = null)
+    public function setFrom($email, $name = null, $auto = true)
```
tsteur, line 104: could we document what the auto parameter does?

config/global.php
tsteur: Actually, we'll need it to be a class probably as we also use it in WordPress see https://github.com/matomo-org/wp-matomo/blob/1.0.6/config/config.php#L88 and https://github.com/matomo-org/wp-matomo/blob/1.0.6/classes/WpMatomo/Email.php#L39

So we'd maybe need some `Piwik\Mail\Transport` class maybe with a method `send()` or something so we can hook into it and eg ignore PHP Mailer completely

plugins/CoreAdminHome/Controller.php
```diff
@@ -118,7 +118,7 @@ public function generalSettings()
             '' => '',
             'Plain' => 'Plain',
             'Login' => 'Login',
-            'Crammd5' => 'Crammd5',
+            'Cram-md5' => 'Cram-md5',
```
tsteur, line 121: not sure how this is used... does this affect any config setting and does anything need to migrated maybe?

plugins/CoreAdminHome/Emails/JsTrackingCodeMissingEmail.php
```diff
@@ -71,9 +71,9 @@ public function getIdSite()
     private function setUpEmail()
     {
         $this->setDefaultFromPiwik();
-        $this->addTo($this->emailAddress);
+        $this->addAddress($this->emailAddress);
```
tsteur, line 74: any chance we could keep the existing API so we don't need to adjust too many plugins etc? That be great

config/global.php
sgiehl: Any reason you can't hook into `Mail.send` event and manipulate the settings just before sending the mail? Similar as the tests are capturing the mail, you could also adjust the settings

https://github.com/matomo-org/matomo/blob/2bd03bc2e0028d00cea1dc1d6a36021921b4ff78/plugins/Login/tests/Integration/PasswordResetterTest.php#L197-L214

core/Mail.php
sgiehl: will copy it over

plugins/CoreAdminHome/Controller.php
sgiehl: will add a simple migration

plugins/CoreAdminHome/Emails/JsTrackingCodeMissingEmail.php
sgiehl: I've added (deprecated) alias methods to the mail class already, so old plugins should still work. Just changed them in core if we decide to remove the old methods at some time...

core/Updates/4.0.0-b1.php
```diff
@@ -121,6 +121,14 @@ public function doUpdate(Updater $updater)
             // switch to default provider if GeoIp Legacy was still in use
             LocationProvider::setCurrentProvider(LocationProvider\DefaultProvider::ID);
         }
+
+        // @todo migrate that to a config migration. See utf8mb4 branch
```
sgiehl, line 125: as per comment. I've already built a config migration for the utf8mb4 branch. We should change that once that's merged. See https://github.com/matomo-org/matomo/pull/15618/commits/8907203134e0caaec482fe05a9b3f3cd4e20d386

config/global.php
sgiehl: For wordpress you might maybe need to use the `Mail.shouldSend` event, so you are able to abort the normal mail class from sending anything...

tsteur: If listening to shouldSend and returning false it won't trigger `Mail.send` and when listening to Mail.send it will still actually send the mail. Would need to change how this works... through DI is a bit nicer in general as it removes some overhead but not too important. 
We are btw not manipulating the mail we're completely changing the `transport`...

sgiehl: Would it be enough to add a `customSend` method to the mail class. e.g.

```
    public function customSend($mimeHeader, $mimeBody)
    {
        Piwik::postEvent('Mail.send.custom', [$mimeHeader, $mimeBody, $this]);
    }
```

That would allow to set `[mail] transport = custom`, causing no mails send with the phpmailer methods, but calling the `customSend` instead.

tsteur: Would need to implement this on Cloud and WordPress to see what it's like. It's maybe not super intuitive in general. Also the `Mail.send` event is not quite clear it indicates you can send the mail there when actually the apps still does it. Could then maybe just change it like this

```
Piwik::postEvent('Mail.send', [$sent= false,$mail]);
if (!$sent) { ... sent mail ... }
```

Would need to test though what happens on cloud there but suppose we could make it work there. Could maybe tweak the above event?

tsteur:
I'm also seeing another problem re WordPress. See https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/#13-plugins-must-use-wordpress-default-libraries

We will need to test that in Matomo for WordPress we can safely remove the `vendor/phpmailer` library and things still work because we're not allowed to ship our plugin with PHPMailer dependency. Ideally, we would have something like 

```
class Mail {
public function send() {
   $phpmailer = new PHPMailer(); $phpmailer->.... send();
}
}
```

instead of `class Mail extends PHPMailer{}`. This would make our mailer implementation more independent of the actual library but of course this has potentially some downsides. Not sure how easy that would be doable... 

Did a quick test... by default WP does not find PHPMailer 
image

It does work when using `include ABSPATH . '/wp-includes/class-phpmailer.php';`

WP is currently using PHPMailer 5.2.27 in their latest release. However, older versions of WP might use different versions and our PHPMailer version might have slightly different definitions causing issues potentially like these (could happen in the future) 
image

Ideally, we'd not extend the PHPMailer class and ideally, if possible, only use PHPMailer as part of the send method or so.

sgiehl:
@tsteur completely refactored the code now. The Piwik\Mail class is now independent and uses a Transport class to send the mail using PHPMailer. The transport class can be replaced, which can be used by wordpress or cloud plugins....

plugins/ScheduledReports/ReportEmailGenerator/AttachedFileReportEmailGenerator.php
```diff
@@ -47,11 +46,11 @@ protected function configureEmail(Mail $mail, GeneratedReport $report)
         $message = $this->getMessageBody($report);
         $mail->setBodyHtml($message);
 
-        $mail->createAttachment(
+        $mail->addAttachment(
             $report->getContents(),
             $this->attachedFileMimeType,
-            Zend_Mime::DISPOSITION_INLINE,
-            Zend_Mime::ENCODING_BASE64,
+            'inline',
```
tsteur, line 52: should we create a constant for this as well @sgiehl ? Could maybe also keep BC by keeping the old method name? Not too important though. Would need to make sure to mention this in developer changelog

plugins/ScheduledReports/ReportEmailGenerator/AttachedFileReportEmailGenerator.php
```diff
@@ -47,11 +46,11 @@ protected function configureEmail(Mail $mail, GeneratedReport $report)
...
+            Mail::ENCODING_BASE64,
```
tsteur, line 53: btw what does the encoding mean here? Looking at the usages it doesn't look like the body is actually eg base64 encoded? eg in plugins/ScheduledReports/ReportEmailGenerator/AttachedFileReportEmailGenerator.php 

Maybe the Mail method can decide on the encoding? Not sure we want to use other encodings? (haven't looked)

core/Mail/Transport.php
```diff
@@ -0,0 +1,141 @@
...
+     * @throws \PHPMailer\PHPMailer\Exception
+     */
+    public function send(Mail $mail)
+    {
+        $phpMailer = new PHPMailer(true);
+
+        PHPMailer::$validator = 'pcre8';
+        $phpMailer->CharSet = PHPMailer::CHARSET_UTF8;
+        $phpMailer->Encoding = PHPMailer::ENCODING_QUOTED_PRINTABLE;
+        $phpMailer->XMailer = 'Matomo ' . Version::VERSION;
```
tsteur, line 37: Any chance we find a way to white label this term if someone configured a different brand name in WhiteLabel plugin? Not too important though. Or could maybe replace with `SettingsPiwik.getMatomoUrl()`? Not sure if/what restrictions there are on this

tsteur: Suppose we'd also not want to directly disclose the version number like this maybe? Not too much of an issue though maybe.

tsteur (commented):
Thanks @sgiehl  I think that's something I will be able to work with (hopefully). I'm expecting some trouble where we might need to add quite some code in Matomo for WordPress around attachments but will try to workaround and can always make further changes later if needed. Left a few other comments to look at. Cheers

core/Mail/Transport.php
sgiehl: You can set whatever you want as XMailer header. If nothing is set it will default to `PHPMailer X.Y.Z (https://github.com/PHPMailer/PHPMailer)`. I'll change that to an whitespace string. According to the PHPMailer code, that should remove the xmailer header

plugins/ScheduledReports/ReportEmailGenerator/AttachedFileReportEmailGenerator.php
sgiehl: I'll remove the disposition completely and let the transport class automatically use the best value. Will make it actually less complex.

plugins/ScheduledReports/ReportEmailGenerator/AttachedFileReportEmailGenerator.php
sgiehl: yes. will remove that. PHPMailer will automatically encode the data based on the given encoding, so should be fine to always use base64

tsteur:
@sgiehl just tested it but noticed there's now a method `addBcc` missing.

The `setReplyTo` method might be good to not deprecate it but instead actually make it a `set` (meaning overwrite any previously set replyTo)?

Otherwise looks good and things worked. Also text mails and various attachments. Be good to make the two changes and then merge 👍

sgiehl:
@tsteur applied the feedback. will merge once the tests passed...

hahayidu:
Hi,

If anyone need to disable $SMTPAutoTLS functionalty because of a getting an error: SMTP Error: Could not connect to SMTP host.


PHPMailer.php

    /**
     * Whether to enable TLS encryption automatically if a server supports it,
     * even if `SMTPSecure` is not set to 'tls'.
     * Be aware that in PHP >= 5.6 this requires that the server's certificates are valid.
     *
     * @var bool
     */
    public $SMTPAutoTLS = true;

You can set true to false.


**This change may cause security vulnerability that we cannot predict. Please also read the ongoing messages.**

Findus23:
Hi @hahayidu,

If I understand it correctly setting this setting to false, means PHPMailer will accept any invalid certificate which means anyone can man-in-the-middle your E-Mail connection and read and modify it and phpmailer will just accept their faked certificate as correct.

So I'd recommend you to fix the actual error instead of disabling security checks.

hahayidu:
Hi @Findus23,

Yes you are right if matomo send mail directly to wide area network without encryption someone can be listen. In particular, sending a password reset request can be more dangerous than reports.

In my case, I wasn't even getting a certificate error. Because our internal smtp relay server doesn't a valid TLS configuration. (Probably doesn't have an any best practices configuration) I know this kind of case is not common. 

I think real problem is PHPMailer override the mail configuration.

This is my configuration.

> [mail]
transport = "smtp"
port = "25"
host = "XXXX.YYYY.com"

If $SMTPAutoTLS is true matomo can't send e-mail. I expect if I don't select any encryption, phpmailer shouldn't encrypt to connection. Empty is also an option in matomo ui.

So, [Opportunistic TLS](https://en.wikipedia.org/wiki/Opportunistic_TLS) (which is default for phpmailler) is not useful some case.

Thank you for your response. I will update first post with your concerns.

PR #15894, opened 2020-05-01
Send onboarding email during installation

tsteur:
Refs MAR-777

image

In my case it went into spam since the sender was basically `foo@mbp.localdomain`. Not sure how often sending the mail actually works when no proper SMTP is configured and how often it would go to spam? Any thoughts @mattab @Findus23 

I'm catching any error that happens so hopefully this doesn't influence


Findus23:
In theory it should never matter if you are using SMTP or PHP mail (and therefore `sendmail`), at least once we are using PHPMailer for both.

But maybe we should make the fact that the from address can not be detected automatically and this FAQ more prominently. Maybe even add a system check that looks up if the from mail ends with `.localdomain` or something like this.

https://matomo.org/faq/how-to/faq_109/

plugins/Installation/Onboarding.php
```diff
@@ -0,0 +1,80 @@
...
+<br>
+';
+
+        $mail = new Mail();
+        $mail->addTo($email);
+        $mail->setSubject('3 quick wins to improve your Matomo right now');
+        $mail->setBodyHtml($message);
+
+        $testMode = (defined('PIWIK_TEST_MODE') && PIWIK_TEST_MODE);
+        if (!$testMode) {
```
sgiehl, line 70: That should actually not be needed, as it's already done in the mail class. See 
https://github.com/matomo-org/matomo/blob/da92367838fa85b806dfd2486b6e91c308cb5e46/core/Mail.php#L127-L135

plugins/Installation/Onboarding.php
```diff
@@ -0,0 +1,80 @@
+<?php
+/**
+ * Piwik - free/libre analytics platform
```
sgiehl, line 3:
```suggestion
 * Matomo - free/libre analytics platform
```

sgiehl (changes requested):

tsteur:
Made the changes @sgiehl

tsteur:
@sgiehl I suppose it is good to merge now?

Findus23:
I still think that the subject line looks like a random SPAM mail.

tsteur:
@Findus23 changed it 👍

sgiehl (approved):
Tested that locally again and it seems to work as expected

Issue #15426, opened 2020-01-21
Offer feature to anonymise or shorten the referrer


tsteur:
I would say we'd want to have different options maybe:

* Full
* Exclude query parameters
* Exclude path (basically only record the subdomain domain)
* Exclude subdomain (eg someone coming from example.matomo.cloud be still identifiable possibly). We would in that case only record `matomo.cloud`
* Nothing, only record the type of referrer type

This can avoid tracking personal data since the page a user comes from, might include it's name, or some ID, etc.

We maybe also want to look generally into some referrers and remove url parameters from most ad and social networks etc.

Not wanting to make this feature too complex but was thinking if it could be also configurable for social search engines specifically and overall. Eg we recognise a few as social acquisition and these we maybe know how to anonymise and therefore could still keep some of the URL parameters (whitelist). Maybe doesn't need to be a separate setting. This way would maybe still get some valuable information for some referrers while removing eg all query parameters for all others.

Besides this as a follow up from https://github.com/matomo-org/matomo/issues/15902 we want to further anonymise some referrers automatically:

Also below... not sure but this looks like it contains personal data... wonder if it makes sense to only store `https://googleads.g.doubleclick.net/pagead/ads`?  or better `https://googleads.g.doubleclick.net/pagead/ads?url=https://example.com/foobar.php`?

```
[...]

PR #15905, opened 2020-05-04
Exclude some url parameters from referrer urls

### LINKED ISSUE #15902: Instagram generates heaps of different referrer urls causing out of memory issues
Seeing referrer urls like 

```
https://l.instagram.com/?u=http://www.example.com/&e=BTPctSgKzf_-3AEudknfwySo4YW52B5wk1evGWzkDvpRCQYf5Kid7P6mQVKbEhsNvwC0WY37vj19ZKQb&s=1
```

It seems this `&e=` seems something unique to instagram as I see this in many different URLs. See also https://webmasters.stackexchange.com/a/114165

The e parameter most likely stands for external or event. In the worst case this identifies a user and represents personal data?

We do not want to store this parameter in the DB and want to completely remove that string. Potentially we also want to move the `s` parameter. Not sure what it stands for.

In some examples this might cause millions of different URLs causing millions of different rows when archiving.

sgiehl:
fixes #15902


core/Tracker/PageUrl.php
```diff
@@ -44,6 +44,11 @@ public static function excludeQueryParametersFromUrl($originalUrl, $idSite)
         $parsedUrl = self::cleanupHostAndHashTag($parsedUrl, $idSite);
         $parametersToExclude = self::getQueryParametersToExclude($idSite);
 
+        if (!empty($parsedUrl['host']) && strpos($parsedUrl['host'], 'instagram.com') !== false) {
```
tsteur, line 47: @sgiehl couldn't find how this method is used in referrer url and also set a debug breakpoint when executing eg `matomo.php?rec=1&idsite=1&action_name=foo&send_image=1&urlref=https%3A%2F%2Fl.instagram.com%2F%3Fu%3Dhttps%253A%252F%252Fexample.com%252Fexample.com%26amp%3Be%3DBTPcuqWixl6Mf5hgYPp6wXIlstuaEdJssdYEvT9s8-6yme_lb275lY2Bwc-YvE-fZNtSKux4QB-v8xNk%26amp%3Bs%3D1` but it wouldn't go in there for the referrer url? (it's a new visit).

Is this executed for the referrer url?

sgiehl: Sorry, my fault. Seems I didn't think that through. Thought that would be similar as the gclid parameter, but for sure that isn't the case. Will update the implementation.

plugins/Referrers/Columns/Base.php
```diff
@@ -137,6 +139,32 @@ protected function getReferrerInformation($referrerUrl, $currentUrl, $idSite, Re
...
+        Piwik::addAction('Tracker.PageUrl.getQueryParametersToExclude', function(&$parametersToExclude) {
+            if (!empty($this->referrerHost) && strpos($this->referrerHost, 'instagram.com') !== false) {
+                $parametersToExclude[] = 'e';
+                $parametersToExclude[] = 's';
+            }
+            if (!empty($this->referrerHost) && strpos($this->referrerHost, 'facebook.com') !== false) {
+                $parametersToExclude[] = 'h';
+            }
+            if (!empty($this->referrerHost) && strpos($this->referrerHost, 'google.') !== false) {
+                $parametersToExclude[] = 'ust';
```
tsteur, line 154: fyi some tests seem to be failing and need to be updated.

In https://travis-ci.org/github/matomo-org/matomo/jobs/683276069#L984

wonder if we can also exclude `bvm` and `ei`.

Re the addAction... wondering if this causes issues eg in BulkTracking. Say it calls

* PageUrl::excludeQueryParametersFromUrl($pageUrl, $this->idsite);
* AddAction to exclude more params
* PageUrl::excludeQueryParametersFromUrl($this->referrerUrl, $this->idsite);
* now it does another tracking request within bulk tracking
* PageUrl::excludeQueryParametersFromUrl($pageUrl, $this->idsite); // now it would potentially also exclude above parameters if referrers was one of these?

tsteur: Could maybe add a third parameter say `excludeQueryParametersFromUrl($originalUrl, $idSite, $extraParamsToExclude = array())`?

plugins/Referrers/Columns/Base.php
```diff
@@ -137,6 +139,45 @@ protected function getReferrerInformation($referrerUrl, $currentUrl, $idSite, Re
...
+    protected function excludeQueryParamsFromReferrerUrl()
+    {
+        // Add additional excludes for specific referrers
+        Piwik::addAction('Tracker.PageUrl.getQueryParametersToExclude', function(&$parametersToExclude) {
+            if (!empty($this->referrerHost) && strpos($this->referrerHost, 'instagram.com') !== false) {
+                $parametersToExclude[] = 'e';
+                $parametersToExclude[] = 's';
+            }
+            if (!empty($this->referrerHost) && strpos($this->referrerHost, 'facebook.com') !== false) {
+                $parametersToExclude[] = 'h';
```
tsteur, line 151: @sgiehl can we also remove `fbclid` here or would this cause any issues with a possible future paid advertising integration for DB?

sgiehl: `fbclid` is included in `url_query_parameter_to_exclude_from_url`

plugins/Referrers/Columns/Base.php
```diff
@@ -149,16 +149,45 @@ protected function excludeQueryParamsFromReferrerUrl()
...
+                $parametersToExclude[] = 'continuation';
+                $parametersToExclude[] = 'feature';
             }
             if (!empty($this->referrerHost) && strpos($this->referrerHost, 'bing.com') !== false) {
                 $parametersToExclude[] = 'cvid';
                 $parametersToExclude[] = 'refig';
                 $parametersToExclude[] = 'elv';
                 $parametersToExclude[] = 'plvar';
+                $parametersToExclude[] = 'setlang';
+                $parametersToExclude[] = 'qs';
```
sgiehl, line 185: sure you want to exclude `setlang` and `cc`? those are set to iso country and language codes based on the users selection. As bing is not using country domains that might be the only way to get that info (if it's relevant for someone)

plugins/Referrers/Columns/Base.php
tsteur: 👍  I see... noticed it in the log_visit table it is eg like `https://l.facebook.com/l.php?u=https://foo.bar/baz?fbclid=IwAR2RmCIQ_Nmn0K9V7B-c8BBMR8nBuEMVud5XW_tRMNP5rLZOcm2b9BsT1QE` so the fcblid seems to be actually part of the url in `u`... guess that's not really removable so easily.

plugins/Referrers/Columns/Base.php
tsteur: Yes, we should be able to see this based on geolocation and detected language I suppose. Removed it cause was seeing heaps of referrers from different bing countries / languages. Then each referrer url had like only 1 visit and weren't visible vs if we remove these parameters, they would have had maybe a count of 4 or 5 and then these referrers would have become visible in the reports because they represent same search term. It's really about trying to group as many referrers as meaningful together as possible.

sgiehl: Sure. But then we could also group together all google tlds 🤷 
Actual I'm wondering if that is the correct way. We are manipulating the tracked referrer and might cut of information that might possibly be useful for anyone.
Where exactly does it cause memory problems having too many different referrer urls? Wondering if it wouldn't be enough to do the referrer url manipulation only while archiving the reports and keep storing the original referrer in the database for live reports.

plugins/Referrers/Columns/Base.php
sgiehl: Changed the implementation to use an additional parameter as suggested.

plugins/Referrers/Columns/Base.php
tsteur: It causes issues in MySQL... like queries taking 6 hours to run because they do `group by "referer_type", "referer_name", "referer_keyword", "referer_url"` resulting in millions of rows vs now we're like doing with 30K rows. 

> We are manipulating the tracked referrer and might cut of information that might possibly be useful for anyone.

Sure some might be interested, and they could maybe write a plugin. But it would be a very small percentage that is impacted by this likely vs say 95% would profit from it. Eventually we will be even cutting out a lot more by default eg for privacy reason as otherwise a lot of personal data is tracked requiring tracking consent see https://github.com/matomo-org/matomo/issues/15426

It's also making the reports more useful as more information is grouped together and things become more accurate. Super important referrers suddenly show up in reports (we can see the effect already) that were invisible before.

AFAIK there wouldn't be much of a way for users to utilise the full referrer url currently anyway? Except for maybe segmenting?

So far it really helps with memory, performance, correctness of data, and even improves privacy.

Of course if needed can keep some parameters like lang or country if really needed, just makes the result maybe less accurate.

sgiehl: Ok. makes sense. Should we implement a possibility to disable that manipulation?
If not, feel free to review and merge if everything is fine now.

tsteur: We can always add a setting later if / when needed. Not wanting to add too many settings that may not be needed.

tsteur:
Cheers @sgiehl

PR #15907, opened 2020-05-04
Removes Zend_Validate

sgiehl:
Zend_Validate is currently only used in Zend_Mail, which will be removed with #15891 and in Provider plugin, which will be adjusted with https://github.com/matomo-org/plugin-Provider/pull/4

Once both are merged, we can finally remove Zend_Validate as well 🎉


Findus23:
Which I think only leaves us with Zend DB and Zend Sessions. I doubt the former is used (it seems to be some kind of simple ORM) and maybe we could also get rid of the latter with Matomo 4 as it was also not updated in a decade.

diosmosis (approved):

sgiehl:
will merge that into the phpmailer branch, as all other requirements are done.

Issue #15780, opened 2020-04-06
Outlink and download tracking not working for links added after DOM ready (which many are these days eg angular, vue.js, react, ... SPA).


thibautguedou3:
Hi there,

We are using Matomo on a VueJS project but we have trouble tracking outlinks.

When a user navigates from our home page to an other page inside the app and then clicks on a link that redirects him outside the app, the link does not appear as an outlink on the Matomo dashboard. However, when the user clicks on an external link directly on the home page, the outlink appears correctly on the Matomo dasboard.

You can find a very simple vue project where you can reproduce the issue: https://github.com/thibautguedou3/matomo-vue-outlink

I can reproduce the issue both on Chrome and Safari

Platform: Mac OS

Could it be a configuration issue in the matomo script ?
```html
<script type="text/javascript">
      var _paq = window._paq || [];
      /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
      _paq.push(["trackPageView"]);
      _paq.push(["enableLinkTracking"]);
      (function() {
        var u = "MATOMO_URL";
        _paq.push(["setTrackerUrl", u + "matomo.php"]);
        _paq.push(["setSiteId", "SITE_ID"]);
        var d = document,
          g = d.createElement("script"),
          s = d.getElementsByTagName("script")[0];
        g.type = "text/javascript";
        g.async = true;
        g.defer = true;
        g.src = "//cdn.matomo.cloud/MATOMO_URL/matomo.js";
        s.parentNode.insertBefore(g, s);
      })();
    </script>
```

thibautguedou3:
@tsteur thanks for your answer! Yes we are using the trackLink method manually for now but we were hopping to remove these manual calls once the bug has been fixed.

Ok I also think this has something to do with the moment when links are added to the DOM. Could it be possible for matomo to expose a method to "update" listeners on links on a page? Or simply delay the listeners addition.

There are still behaviours that I don't understand: I cannot see outlinks on pages that are not the home page even if I try to access them directly. So I think that there might be something else going on here.

tsteur:
I've changed the title for now @thibautguedou3 

I think it should be quite easy fixable by using some similar logic as in the tag manager:

```js
function isClickNode(nodeName)
                {
                    return nodeName === 'A' || nodeName === 'AREA';
                }

                document.body.addEventListener( 'click', function (event) {
                    if (!event.target) {
                        return;
                    }

                    var target = event.target;
                    var nodeName = target.nodeName;

[...]

PR #15915, opened 2020-05-05
Search for links on page load as well, not just on ready

tsteur:
refs https://github.com/matomo-org/matomo/issues/15780

It's absolutely not solving #15780 but it can already greatly improve the situation where links are added between DOM Ready and DOM load. 

Putting this into 3.x as was looking into this as part of Matomo for WordPress https://wordpress.org/support/topic/tracking-downloads-served-through-zotpress/#post-12787301


tsteur:
build js

sgiehl (approved):

PR #15919, opened 2020-05-06
Bind param values manually since PDOStatement::execute() will assume …

diosmosis:
…params are all strings. This causes null values to not be bound correctly.

I checked and mysqli appears to be fine.

### DIFF
--- a/libs/Zend/Db/Statement/Pdo.php
+++ b/libs/Zend/Db/Statement/Pdo.php
@@ -225,10 +225,12 @@ class Zend_Db_Statement_Pdo extends Zend_Db_Statement implements IteratorAggrega
     {
         try {
             if ($params !== null) {
-                return $this->_stmt->execute($params);
-            } else {
-                return $this->_stmt->execute();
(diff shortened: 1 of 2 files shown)


libs/Zend/Db/Statement/Pdo.php
```diff
@@ -225,10 +225,12 @@ public function _execute(array $params = null)
     {
         try {
             if ($params !== null) {
-                return $this->_stmt->execute($params);
-            } else {
-                return $this->_stmt->execute();
+                foreach (array_values($params) as $index => $paramValue) {
```
tsteur, line 228: btw does it make sense to add one integration test for this with a null value?

diosmosis: 👍 was planning to but forgot

tsteur: 👍 code looks good otherwise and tested it locally, worked. feel free to merge once you've added a test. Must have been a hard fine

PR #15929, opened 2020-05-09
Allow setting mock "now" value through out matomo during tests.

diosmosis:
Couple small changes needed for tests of a new premium feature.

### DIFF
--- a/core/Date.php
+++ b/core/Date.php
@@ -556,7 +556,7 @@ class Date
      */
     public static function today()
     {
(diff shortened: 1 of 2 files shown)


tsteur:
@diosmosis looks good generally. Some tests are failing though see eg https://travis-ci.org/github/matomo-org/matomo/jobs/685178687

tsteur:
ping @diosmosis

diosmosis:
working on this one off the clock, may be a bit before I get to it

PR #15938, opened 2020-05-11
remove support for old format of adding custom search engines

sgiehl:
Guess that shouldn't be needed anymore. Before Piwik 2.15 we used to have another array format for defining search engines. So if anyone would still use this old format with the event `Referrer.addSearchEngineUrls`, that might not work anymore. But as that event is not even marked as API, I guess that should be fine.

refs #8567


tsteur:
btw all Events are marked as API unless we specifically say to ignore it. That's why it appears here: https://developer.matomo.org/api-reference/events#referreraddsearchengineurls 

should still be fine to merge though.

PR #15939, opened 2020-05-11
Make it possible to automatically set up geoip2 in installation

### LINKED ISSUE #12811: Activate DB-Ip support by default for all users
Now that we have Geoip2 activated by default for all new users (done in https://github.com/matomo-org/matomo/pull/12699), we should also activate the feature for all existing users who are upgrading. Since Geoip2 requires a DB update for the Region column, we need to wait for Matomo 4 to enable the plugin.

### Suggested steps for final Geoip2 takeover

* Activate the plugin on update to Matomo 4.0.0
* Could we automatically download the database?
* When Geoip1 was enabled, and Geoip2 database successfully downloaded, could we automatically enable Geoip2 provider instead?
* We should also move all the Geoip1 code to a separate plugin. Later (maybe in Matomo 5?) we should then remove/delete this plugin completely.
* Check that all FAQs/ user guides exclusively mention Geoip2 (delete all Geoip1 specifics)
* ...?

sgiehl:
As discussed in #12811 I've added a new checkbox in the last step of the installation.

When checked, Matomo will automatically add the GeoIP2Updater setting, so the dbip database will be donwloaded/updated monthly. In addition it will set the used location provider to geoip2php.

But I'm currently unsure how to handle the automatic download of the database. Doing that within the installation process is a bit difficult to implement and it also might take a while.

With this implementation the updater will be set and thus the automatic task should be scheduled for "tomorrow". That means until then the geolocation would be reported as broken in admin, as geoip2 is configured, but no database is available. (geolocation itself would still work and fallback to the default provider)
~I've changed the geolocation admin a bit, so it will automatic start the download process if autoupdate is configured but no database is available.~
Not sure if that is a suitable solution, but didn't have a better idea right now.
@mattab @tsteur @diosmosis Does anyone of you have a better idea how to do/trigger the download. We could maybe try to run it async, but that might not work in many cases 🤔 

fixes #12811


diosmosis:
what's the issue w/ making the controller requests it to the installation UI?

sgiehl:
> what's the issue w/ making the controller requests it to the installation UI?

That won't work easily with the hooks currently used to add the checkbox I guess.

diosmosis:
I see, I guess the original code probably uses angular or a lot of JS that needs to be loaded? Or would need to be modified to be pre-loaded during installation?

tsteur:
Wouldn't do it directly during installation but only once fully installed. Could we set eg during the installation a flag that we need to download the DB... then we try to complete this download eg as part of an hourly task which would simply be executed eg on the first tracking request or so. Additionally, if `CliMulti::supportsAsync()` returns `true`, then we could launch a command in the background to sync the DB... AFAIK `supportsAsync()` works on many installs and most often it would be downloaded fairly soon after installation is completed (could trigger it on last step for example). For all other cases we'd download it through a task.

Not sure it's somewhat clear what I mean?

sgiehl:
@tsteur change that. it will now be done using a scheduled task and if async is possible, the scheduled tasks will be triggered in the last step of installation

plugins/GeoIp2/GeoIp2.php
```diff
@@ -50,4 +58,54 @@ public function getClientSideTranslationKeys(&$translationKeys)
...
+
+            /** @var Scheduler $scheduler */
+            $scheduler = StaticContainer::getContainer()->get('Piwik\Scheduler\Scheduler');
+            $scheduler->rescheduleTask(new GeoIP2AutoUpdater());
+
+            $cliMulti = new CliMulti();
+
+            if ($cliMulti->supportsAsync()) {
+                $phpCli = new CliMulti\CliPhp();
+                $command = sprintf('%s %s/console core:run-scheduled-tasks > %s/tmp/out.txt 2>&1 &',
```
tsteur, line 105: @sgiehl might be good to run eg `console core:run-scheduled-tasks  --force "Piwik\Plugins\GeoIp2\GeoIP2AutoUpdater.update"` directly? Also is the redirection needed to out.txt needed? Eg on nginx by default the file would maybe be directly accessible. Not a big issue but might just not be needed?

plugins/GeoIp2/lang/en.json
```diff
@@ -1,6 +1,8 @@
 {
   "GeoIp2": {
     "AssumingNonApache": "Cannot find apache_get_modules function, assuming non-Apache webserver.",
+    "AutomaticSetup": "Automatically configure geolocation using a dbip database",
+    "AutomaticSetupDescription": "For a proper geolocation Matomo requires an external database. Using this option, Matomo will automatically be configured to download an use the latest %1$sdbip%2$s city level database.",
```
tsteur, line 5: could maybe say `... city level database which is licensed under the [Creative Commons Attribution 4.0 International License](https://db-ip.com/db/lite.php?refid=mtm)".` or just say `... city level database. [View licensing terms](https://db-ip.com/db/lite.php?refid=mtm)`. Not sure we do this in the GeoIP downloader already. If not, then it might not be too important. Generally be good though if we do it in the admin as well. fyi @Findus23

tsteur: image
 sorry just seeing this is actually already done just not as clear.

tsteur (commented):
Looks all good. Left one comment re launching the command. Might be good to look into it.

plugins/GeoIp2/GeoIp2.php
sgiehl: the output shouldn't be needed, that was for debugging purpose 🙈 
will change that to directly execute the task.

sgiehl:
@tsteur updated the code

tsteur:
Looks good to me 👍

PR #15955, opened 2020-05-15
Do not check if any archive is running when a segment is forced

tsteur:
From a support request. eg user has

```crontab
6 * * * * * console core:archive --force-idsegment=1
6 * * * * * console core:archive --force-idsegment=2
6 * * * * * console core:archive --force-idsegment=3
```

Then currently it would skip some segments because it checks if any archive is already running. After the launching the first one for segment 1, an archive is already running and would probably never launch 2nd and 3rd segment AFAIK.

@mattab

### DIFF
--- a/core/CronArchive.php
+++ b/core/CronArchive.php
@@ -2119,6 +2119,12 @@ class CronArchive
 
     public function isAlreadyArchivingUrl($url, $idSite, $period, $date)
     {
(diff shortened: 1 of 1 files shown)


sgiehl (approved):

tsteur:
@mattab fyi waiting for your confirmation that it fixes the user's issue before merging

PR #15956, opened 2020-05-15
Remove gears detection

sgiehl:
cheers @pebosi for the initial work

replaces #15627


sgiehl:
I've gone through the changes again and reviewed them. There were some tests failing and various changes missing in other repos (see linked PRs). Would be good if someone else could quickly look through the changes.

And just to point it out again. These changes will drop the column that was used to store the gears detection. But as it wasn't possible to segment for that and there shouldn't have many detections been positive in the past anymore, I think we can remove that. Already processed reports might still show gears, only new reports won't contain it anymore...

tsteur:
@sgiehl not sure if this needs another review from someone else as you reviewed it initially?

sgiehl:
@tsteur If removing the column is fine, guess it should be enough to have a look at the linked PRs and maybe the changes I've added to this PR

tsteur:
Had a quick look and looks all good @sgiehl  👍

PR #15957, opened 2020-05-15
Allow using matomo prefixed parameters for campaigns such as matomo_campaig/mtm_campaign

sgiehl:
Shall we maybe also add something like `mtm_campaign`?

refs #12420


diosmosis:
Seems strange to not use allow mtm_ prefix as well.

tsteur:
btw `mtm` stands for `matomo tag manager`. Of course could also stand for `matomo` :) I have not really and preference but maybe wouldn't provide too many options. I can't really think of a reason why it would be an issue for people to still use `pk_` but could also allow `mtm` additionally

sgiehl:
also added the mtm_ prefixes. We maybe should update the documentation for Matomo 4 later so we suggest to use the new parameter names...

diosmosis:
Weird, the build didn't run for this PR?

sgiehl:
build js

sgiehl:
> Weird, the build didn't run for this PR?

It actually did, but it's not linked here 🤷

Issue #15930, opened 2020-05-10
Archive Blob Tables bloating in size


James-Oakley:
This is a follow-up to #15831.

Since upgrading to 3.13.5, I was able to get my bloated archive tables back to their proper size by running `./console core:purge-old-archive-data all`.

However, since then, the tables began to grow in size again.

Things are much better since 3.13.5. I was able to shrink the tables again by running the `purge-old-archive-data` command again. My largest archive blog table then shrank from 50,456 rows and 236 MiB back to 9,910 rows and 7.0 MiB. So I now have a way to solve the problem of the bloated tables (whereas, before 3.13.5, I had no way to shrink those tables).

However, they are surely not supposed to grow to the point that they need pruning with a manual command. The table referred to above is for January 2020, to which no new data is added, so I'd expect that archive table to remain the same size.

It may well be that the problem is actually my end. Having read the documentation, I'm running a maintenance command once every 10 minutes:

`/usr/bin/php ~/public_html/console core:archive --url=https{domain} >> {log-file}`

Is there another command that, when Matomo is behaving correctly, needs running on schedule to keep things running smoothly? (I realise I _could_ run `purge-old-archive-data` with cron, but is it correct behaviour for me to need to do so?)

Or are the tables still bloating when they shouldn't be?

lolobu:
Same issue on my side. After the 3.13.5 update, my database went down back to 25 MiB after having hit 410 MiB with version 3.13.4. Those 20 MiB has been more or less its size for months before v 3.13.4.
But after the expected shrink following the 3.13.5 update, it started again to grow every day by 9 MiB hitting now 112 MiB.

tsteur:
Putting it for now into 3.13.6 so we don't forget to have a look. It might be possible that eg archives are cleaned up only weekly and they used to be cleaned up daily (just a guess)

daeuble:
Same here on our 3.13.5 installation.

panpomaly:
After the archive got shrunk by the update to 3.13.5, we are also experiencing the same growth again as before.

But It seems to only affect the archives for the same month as before: January (year-archives), March and April. These are only the months, when the installation ran with Matomo 3.13.4 plus the year-archives. No abnormal growth for the month after updating (at least yet) and before.

diosmosis:
Can someone experiencing the problem try to run the following command (from the matomo root directory):

```
[...]

PR #15963, opened 2020-05-18
Purge from year table as well.

diosmosis:
Refs #15930


tsteur:
Looks good if tests pass

Issue #9939, opened 2016-03-17
Offline tracking in JavaScript API


PCSun1987:
Current it's not possible to send tracking event offline. 

So one idea would be to extend PIWIK tracking API especially for JavaScript with additional parameter about even time-stamp (optional), so inside JS, you can keep the event locally and once connecting to internet, send all the events with the event happening time. 

For basic tracking, probably can do similar way. 

https://forum.piwik.org/t/does-piwik-work-even-your-offline/7295/9

hpvd:
+1 
would be a good step on way to universal tracker (Piwik 3.0).
It's not only intersting for apps which have temporaly no connection to web,
but also a good thing to make the manual input of "real world" events after they has happened possible and give them the correct place in timeline

tsteur:
This would be a nice feature indeed

PCSun1987:
So...how long we would plan to have this functionality? E.g. 3.0 would be released when?

tsteur:
Piwik 3.0 would be in about a year but this feature is not planned yet. Pull request or suggestions on how to implement it are always welcome :+1:

stehlo:
Just looking for such a solution. This is tremendously important for mobile apps.

In the JS tracking client, I have noticed a method called retryMissedPluginCalls() and the array missedPluginTrackerCalls. It could be interesting to hook it in some way in order to intercept the calls to the server in an offline state. Then, upon "online" event, we would call and retry missed calls.

Your thoughts, @tsteur and @mattab ?

tsteur:
`retryMissedPluginCalls()` is actually a bit different here. Plugins can extend the Piwik JS tracker and there may be cases where either Piwik is loaded first, or the Plugin. If Piwik was loaded first and tries to apply all `_paq.push` calls, it cannot call the methods for the plugins yet as they are not yet loaded. Therefore once the plugin is loaded they try to call all missed plugin calls again.

Offline tracking is super important nowadays for mobile apps, progressive web apps, .... If someone wanted to work on it I'm happy to give some support. I think it needs to be worked out what the best place is to save requests that couldn't be sent because the user is offline (eg localstorage, ...) and then we need to detect whether user is offline and when user goes online again. Some browsers have an API for that.

There might be one problem when tracking the requests later. I think by default Piwik lets you only track requests up to 4 hours in the past without needing an authentication token. The time is customizable AFAIK but it might be something to take into account that requests older than 4 hours may have to be invalidated.

stehlo:
[...]

PR #15970, opened 2020-05-22
JS Offline tracking

tsteur:
refs https://github.com/matomo-org/matomo/issues/9939

early stage... usage:

within a service worker do

```js
self.importScripts('https://your.matomo.domain/offline-service-worker.js');
matomoAnalytics.initialize();
```

to register service worker:

```js
if ('serviceWorker' in navigator) {
  window.addEventListener('load', function() {
    navigator.serviceWorker.register('/your-service-worker.js');
  });
}
```

We discussed to merge this early in the beta even though it might not fully work yet. This way people can test it easily and give us feedback and review etc.


offline-service-worker.js
```diff
@@ -0,0 +1,175 @@
...
+    var maxTimeLimit = options.timeLimit || (60 * 60 * 24); // in seconds...
+    // same as configured in in tracking_requests_require_authentication_when_custom_timestamp_newer_than
+
+    function getQueue()
+    {
+        return new Promise(function(resolve, reject) {
+            // do a thing, possibly async, then...
+
+            if (!indexedDB) {
+                reject(Error('No support for IndexedDB'));
```
diosmosis, line 16: is there a reason `Error(...)` is used and not `new Error(...)`?

tsteur: @diosmosis it appears both seems to work.

tsteur: no particular reason though

core/Tracker/Request.php
```diff
@@ -498,6 +504,10 @@ protected function getCustomTimestamp()
             $cdt = strtotime($cdt);
         }
 
+        if (!empty($cdo)) {
+            $cdt = $cdt - abs($cdo);
+        }
```
diosmosis, line 509: if a request has a cdt, then should we still apply cdo? ie, if for some reason, the request has `cdt=date(2018-12-01 03:04:05)&cdo=30`, then shouldn't the given cdt be used anyway? Since it's hardcoded.

diosmosis (commented):
left one comment on a possible issue.

approach is basically just:

- if not online, queue requests in indexeddb w/ cdo set to time between now & value creation time
- when online, play all queued requests
- in tracker, if cdo is present apply it to timestamp before processing the request

is this correct?

and also it caches piwik.js/matomo.js? is this needed or can it just be cached via an http header?

offline-service-worker.js
tsteur: changed it 👍

offline-service-worker.js
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+            let example = {
+                url: event.request.url,
+                referrer : event.request.referrer,
+                method : event.request.method,
+                referrerPolicy : event.request.referrerPolicy,
+                headers : headers,
+                created: Date.now()
+            };
```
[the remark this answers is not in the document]
tsteur: changed it 👍

core/Tracker/Request.php
tsteur: Initially, when cdt and cdo was used I only used the `cdt` but then changed it in one of the recent commits as it might not be expected and only adds a benefit / more flexibility to support using both. what would happen in this case is that it removes 30 seconds from the set cdt.

eg `2018-12-01 03:04:05 - 30s = 2018-12-01 03:03:35` . This can be actually quite useful in some cases

tsteur:
> and also it caches piwik.js/matomo.js? is this needed or can it just be cached via an http header?

It can't really be cached via http header in case you are fully offline etc. Otherwise the approach is correct 👍

offline-service-worker.js
```diff
@@ -0,0 +1,175 @@
...
+                syncQueue();
+            }
+            caches.open('matomo').then(function(cache) {
+                return cache.match(event.request).then(function (response) {
+                    return response || fetch(event.request).then(function(response) {
+                        cache.put(event.request, response.clone());
+                        return response;
+                    });
+                });
+            })
```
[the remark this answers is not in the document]
tsteur: 👍  makes sense. We can tweak code later once we get feedback the overall idea actually works

tsteur:
@diosmosis applies few changes. Would otherwise love to get this merged to see if it really works and what problems there are in general. Eg I'm not an indexeddb / service worker so I'm sure there might be few things that maybe won't work as expected or could cause race conditions or so maybe.

core/Tracker/Request.php
diosmosis: What I meant was since cdt is meant to force the datetime, it doesn't really make sense to me to apply cdo to it. Ie, if someone sends a request `https://.../matomo.php?idsite=1&cdt=...`, and it gets queued, then we resend it as `https://.../matomo.php?idsite=1&cdt=...&cdo=...`, shouldn't the cdt be respected as is? Ie, we only send the cdt when we want to force the date time to something specific, but cdo should only be applied to the timestamp if we are setting it to `now`.

tsteur: In that case someone would simply not set the `cdo`? I have no big preference though. It could be generally useful though when eg replaying requests maybe for example. Not sure. Right now not seeing really any downside because they don't have to set both?

diosmosis: I guess it depends on whether the system setting cdo needs to be aware of when cdt should be respected or not.

PR #15973, opened 2020-05-23
add missing link to consent help page

Findus23:
This link from https://developer.matomo.org/guides/tracking-javascript-guide#asking-for-consent is missing in the Matomo page version.

Interestingly it didn't get lost in #15557 but was already missing before.


Findus23:
It seems like there are a lot more links missing (two in the first paragraph). Was this intentional?

tsteur:
Probably when we added this these links maybe didn't exist yet? Not sure. Are you planning on adding more links @Findus23 ?

tsteur:
@Findus23  can this one be merged?

Findus23:
@tsteur It should be finished now (apart from the UI tests)

tsteur:
👍 will you update them or should I?

Findus23:
I don't have as much time at the moment and am not up to date with which PRs modified the UI tests, so it would be great if someone else could do it.

tsteur:
@diosmosis do you maybe know why these tests for the PR would always fail: https://travis-ci.org/github/matomo-org/matomo/builds/702234148

Tried to restart the tests and commit them again but they always fail... even merged 4.x-dev into it but still ails...  there is no file change though that would explain this.... could also just merge and see if it then still has a problem but not sure what's happening. 

Funnily the PUSH PR passes and it should be in the end the same as the PR build

diosmosis:
@tsteur I've seen it before but don't know why I happens. the build will continue to fail until a new commit is pushed.

tsteur:
Pushed already 3 times or so... still happening... @diosmosis do you reckon it might be fine once merged? Guess could always revert the PR if it makes the 4.x-dev branch fail later

diosmosis:
@tsteur the push build is fine, so I say merge? If it's just the PR build, it won't have any effect on the 4.x-dev build. And of course for branches we don't need both.

tsteur:
Worked... interestingly there was a screenshot failure now in 4.x-dev which passed in the PULL build which shouldn't have. Something went wrong there with Github :) Will fix that screenshot in a new PR

PR #15975, opened 2020-05-23
add .rtf to download filetypes

### LINKED ISSUE #14579: JS Tracker: add rtf to list of downloads
Hi there,

I'm new here so sorry when I'm mistaken in any etiquette.

I have a little question: is it possible Matomo as default adds rtf to the list of downloads? I know I can add this with a line of javascript, but it's a bit easier when that is already there.

Findus23:
fixes https://github.com/matomo-org/matomo/issues/14579 (it's easier to just add it instead of keeping the issue open).

I can't think of any false-positives ending with `.rtf` so it should be fine.

Let's see if the github action also works for me :)


Findus23:
build js

Findus23:
Yay! :tada:

Findus23:
CSS in the UI test is broken, but I doubt this change caused it.

PR #15979, opened 2020-05-24
no cut-off icons in DataTables

### LINKED ISSUE #13256: Mobile: Icons not fully shown and shown not correctly
This is on Android 7.0 with Via Browser and with Chrome. The arrow icon to sort the table is shown funny and the table icons in the dashboard are not fully visible (maybe also in regular reports).

screenshot_20180807-102642

Findus23:
fixes #13256

Interestingly I can still reproduce the issue with Chromium 81 (noticable in the normal view and quite strongly when using a mobile view. (like in the screenshot of #13256)

Before:
grafik
After:
grafik

I'd wait with merging until #14082 is merged and all UI tests have been updated.

PS: Is it intentional that the vertical scroll bar is always shown?


tsteur:
@Findus23 that scroll bar shouldn't be always shown. I reckon we can merge this PR now though? Will trigger a new build so the screenshots can be updated.

PR #15980, opened 2020-05-24
JS Tracker: require category and action in trackEvent

### LINKED ISSUE #13165: Issue in piwik.js function logEvent
If you call the function

    logEvent(category, action, name, value, customData, callback)

without parameter it shout gives an error.

    (logConsoleError(‘Error while logging event: Parameters category and action must not be empty or filled with whitespaces’);

But it doesn’t. Because the line

     if (trim(String(category)).length === 0 || trim(String(action)).length === 0) { ...

will not work. String(category) returns always 9 (the length of the word undefined) when categorie is empty
So the error will never been showed.

Findus23:
fixes #13165 with the suggesting from there


js/piwik.js
```diff
@@ -4266,7 +4266,7 @@ if (typeof window.Piwik !== 'object') {
             function logEvent(category, action, name, value, customData, callback)
             {
                 // Category and Action are required parameters
-                if (trim(String(category)).length === 0 || trim(String(action)).length === 0) {
+                if (typeof category !== 'undefined' || typeof action !== 'undefined' || trim(String(category)).length === 0 || trim(String(action)).length === 0) {
```
sgiehl, line 4269: Shouldn't that be a `=== 'undefined'` ?

Findus23: You are right. It seems like I forgot to flip them.

js/piwik.js
```diff
@@ -4266,7 +4266,7 @@ if (typeof window.Piwik !== 'object') {
             function logEvent(category, action, name, value, customData, callback)
             {
                 // Category and Action are required parameters
-                if (trim(String(category)).length === 0 || trim(String(action)).length === 0) {
+                if (typeof category === 'undefined' || typeof action === 'undefined' || trim(String(category)).length === 0 || trim(String(action)).length === 0) {
```
tsteur, line 4269: You'd probably want to use `isDefined()` or even better maybe `isNumberOrHasLength()`

Findus23: I was not aware that there were helper functions like this.

It took a few tries, but I think I got it right now.

Findus23:
build js

PR #15986, opened 2020-05-26
Tracker API speed: avoid DB query when recording historical data or other authenticated tracking requests

tsteur:
Was just looking into https://github.com/matomo-org/wp-matomo/issues/288 and then noticed we always first issue a query against the user table before actually checking the cached data.

Should a token be used with admin or write permission, then we actually save the user query.

It shouldn't cause any security issue cause if the token is not a valid `admin` or `write` permission token, then below code would be still executed and eg the brute force check should be executed just like before.

### DIFF
--- a/core/Tracker/Request.php
+++ b/core/Tracker/Request.php
@@ -198,6 +198,19 @@ class Request
             return false;
         }
 
+        // Now checking the list of admin token_auth cached in the Tracker config file
+        if (!empty($idSite) && $idSite > 0) {
+            $website = Cache::getCacheWebsiteAttributes($idSite);
+            $userModel = new \Piwik\Plugins\UsersManager\Model();
+            $tokenAuth = $userModel->hashTokenAuth($tokenAuth);
+            $hashedToken = UsersManager::hashTrackingToken((string) $tokenAuth, $idSite);
+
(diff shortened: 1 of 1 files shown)


core/Tracker/Request.php
```diff
@@ -198,6 +198,19 @@ public static function authenticateSuperUserOrAdminOrWrite($tokenAuth, $idSite)
             return false;
         }
 
+        // Now checking the list of admin token_auth cached in the Tracker config file
+        if (!empty($idSite) && $idSite > 0) {
+            $website = Cache::getCacheWebsiteAttributes($idSite);
+            $userModel = new \Piwik\Plugins\UsersManager\Model();
+            $tokenAuth = $userModel->hashTokenAuth($tokenAuth);
```
sgiehl, line 205: Guess you need to use another variable name here, as it otherwise will overwrite the tokenAuth used for authentication below. Guess that's the reason for all the test failues...

sgiehl (changes requested):
Looking at the code, the changes seems fine, but actually most tracking tests started to fail with this change...

tsteur:
Thanks @sgiehl the change fixed the tests

PR #15991, opened 2020-05-26
Archiving job prioritization and safety precautions

diosmosis:
Refs #15117


core/CronArchive.php
```diff
@@ -235,6 +240,8 @@ public function __construct($processNewSegmentsFrom = null, LoggerInterface $log
         $this->periodIdsToLabels = array_flip(Piwik::$idPeriods);
 
         $this->rawLogDao = new RawLogDao();
+
+        $this->cliMultiRequestParser = new RequestParser($this->makeCliMulti()->supportsAsync());
```
tsteur, line 244: @diosmosis did you maybe forget to commit this file?

diosmosis: added

tsteur:
@diosmosis now getting an exception `isPeriodInThisPeriod` method not defined

diosmosis:
Oh, huh, I forgot to implement that, but the tests were passing for me... interesting, will fix.

core/CronArchive.php
```diff
@@ -1333,4 +1381,110 @@ private function makeWebsiteIdArchiveList(array $websitesIds)
...
+
+        foreach ($inProgressArchives as $archiveBeingProcessed) {
+            $archiveBeingProcessed['periodObj'] = PeriodFactory::build($archiveBeingProcessed['period'], $archiveBeingProcessed['date']);
+
+            if ($this->isArchiveOfLowerPeriod($archiveToProcess, $archiveBeingProcessed)) {
+                return "lower period in progress (period = {$archiveBeingProcessed['period']}, date = {$archiveBeingProcessed['date']})";
+            }
+
+            if ($this->isArchiveNonSegmentAndInProgressArchiveSegment($archiveToProcess, $archiveBeingProcessed)) {
+                return "segment archive in progress for same site/period ({$archiveBeingProcessed['segment']})";
```
tsteur, line 1455: fyi segment might not be defined here 
image

got `Notice - Undefined index: segment`

core/CronArchive.php
```diff
@@ -380,15 +387,38 @@ public function run()
...
             $archivesToProcess = [];
             while (count($archivesToProcess) < $countOfProcesses) {
-                $invalidatedArchive = $this->getNextInvalidatedArchive($idSite);
+                $invalidatedArchive = $this->getNextInvalidatedArchive($idSite, array_keys($invalidationsToExcludeInBatch));
                 if (empty($invalidatedArchive)) {
                     $this->logger->debug("No next invalidated archive.");
                     break;
                 }
 
+                if ($this->hasDifferentPeriod($archivesToProcess, $invalidatedArchive['period'])) {
```
tsteur, line 405: good one 👍

diosmosis:
@tsteur pushed

tsteur (changes requested):
@diosmosis found one more notice. Also not sure if I missed it or if this is covered by a lock. Should the very same archive already be in process (same site, period, date and segment (if given)) to then skip as well?

diosmosis:
> Also not sure if I missed it or if this is covered by a lock. Should the very same archive already be in process (same site, period, date and segment (if given)) to then skip as well?

Covered by two things, first is that if the archive is DONE_IN_PROGRESS we shouldn't handle it. Second is if it is locked and not expired, we shouldn't handle it. We could also just do `<=` when checking period ID and that would also prevent it in a third check. I'll just add that for fun.

diosmosis:
FYI, I'm planning on refactoring/testing all of CronArchive in #11974

tsteur:
@diosmosis don't have to add it unless it's easy. The DONE IN PROGRESS should be fine too

diosmosis:
@tsteur single character change, so definitely easy

tsteur:
👍 feel free to change it and I have another look

diosmosis:
@tsteur you can take another look

tsteur:
@diosmosis some tests are failing now. Otherwise looks good to merge

PR #15998, opened 2020-05-27
Date factory test when date object is passed

tsteur:
As suggested in chat made this change to see if anything breaks.

There are some failing tests in CustomEvents and Content tracking but that's because of how the date was falsely used in the test fixture. Happy to update the tests and add a changelog entry if we are keen to merge this. Not sure what else it would maybe break that maybe isn't covered by tests.

### DIFF
--- a/plugins/Contents/tests/Fixtures/TwoVisitsWithContents.php
+++ b/plugins/Contents/tests/Fixtures/TwoVisitsWithContents.php
@@ -47,7 +47,7 @@ class TwoVisitsWithContents extends Fixture
 
         $this->trackContentImpressionsAndInteractions($vis);
 
-        $this->dateTime = Date::factory($this->dateTime)->addHour(0.5);
+        $this->dateTime = Date::factory($this->dateTime)->addHour(0.5)->getDatetime();
         $vis2 = self::getTracker($this->idSite, $this->dateTime, $useDefault = true, $uselocal);
         $vis2->setIp('111.1.1.1');
         $vis2->setPlugins($flash = false, $java = false);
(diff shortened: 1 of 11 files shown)


tests/PHPUnit/Fixtures/ThreeVisitsWithCustomEvents.php
```diff
@@ -60,7 +60,7 @@ public function trackVisits()
         $this->trackEventWithoutUrl($vis);
         $this->trackMovieWatchingIncludingInterval($vis);
 
-        $this->dateTime = Date::factory($this->dateTime)->addHour(0.5);
+        $this->dateTime = Date::factory($this->dateTime)->addHour(0.5)->getDatetime();
         $vis2 = self::getTracker($this->idSite, $this->dateTime, $useDefault = true, $uselocal);
```
tsteur, line 64: we used to pass a `Date` object here meaning the tracker would do `$date->toString()` which then would only result in `YYYY-MM-DD` and throw away the entire time. So this was before actually not working which is why some tests are failing.

plugins/Contents/tests/Fixtures/TwoVisitsWithContents.php
```diff
@@ -47,7 +47,7 @@ public function trackVisits()
 
         $this->trackContentImpressionsAndInteractions($vis);
 
-        $this->dateTime = Date::factory($this->dateTime)->addHour(0.5);
+        $this->dateTime = Date::factory($this->dateTime)->addHour(0.5)->getDatetime();
```
tsteur, line 50: we used to pass a `Date` object here meaning the tracker would do `$date->toString()` which then would only result in `YYYY-MM-DD` and throw away the entire time. So this was before actually not working which is why some tests are failing.

tsteur:
Any thoughts @sgiehl @diosmosis ?

sgiehl:
In general I would vote for fixing the tests and merging this change. That is actually the behavior I imho would have expected before 🙈

tsteur:
👍 fixed the tests @sgiehl

PR #15999, opened 2020-05-28
Track ecommerce views as new dimensions instead of custom variables

### LINKED ISSUE #12035: Ecommerce Product average price may be incorrect, review implementation without using Custom Variables
When looking to the ecommerce products report, the average product price display incorrect value. The value seems to be multiplied by number of days for which there has been visit on the product page.

I've check that all my visits log product information report correct values directly into the database...

Piwik 3.1.0
Custom e-commerce report plugin (woocommerce)

sgiehl:
fixes #12035


sgiehl:
build js

composer.json
```diff
@@ -35,7 +35,7 @@
         "matomo/cache": "~2.0",
         "matomo/decompress": "~2.0",
         "matomo/ini": "~2.0",
-        "matomo/matomo-php-tracker": "dev-4.x-dev",
+        "matomo/matomo-php-tracker": "dev-ecommerceview",
```
sgiehl, line 38: This needs to be reverted after https://github.com/matomo-org/matomo-php-tracker/pull/68 has been merged

plugins/Ecommerce/tests/System/expected/test_ecommerceOrderWithItemsmultipleDates_andMultipleWebsites__Goals.getItemsCategory_day.xml
```diff
@@ -7,53 +7,89 @@
 				<revenue>2500</revenue>
 				<quantity>3</quantity>
 				<orders>2</orders>
+				<nb_visits>3</nb_visits>
```
sgiehl, line 10: The product view metrics were missing here before because of 
https://github.com/matomo-org/matomo/blob/bc499def7f848983dc85ced1373d7f6c0b41e38d/plugins/Goals/API.php#L792-L796
That's no longer the case, as product views are directly archived...

tests/PHPUnit/System/expected/test_reportLimiting__Goals.getItemsCategory_day.xml
```diff
@@ -2,19 +2,46 @@
 <result>
 	<row>
 		<label>Books Cat #0</label>
+		<nb_visits>5</nb_visits>
```
sgiehl, line 5: The reason why we have more categories here as before is, that custom variables reports are limited by the report limiting in the test. The ecommerce reports do not have any limits when archived, so all categories are shown now. Not sure if that's something we should maybe change.

tests/javascript/index.php
```diff
@@ -3897,37 +3892,10 @@ function generateAnIframeInDocument() {
...
         tracker3.trackPageView("EcommerceView");
 
         tracker3.deleteCustomVariables('page');
 
-        // No data set
-        tracker3.setEcommerceView( );
-        deepEqual( tracker3.getCustomVariable(2, "page"), false, "No data Ecommerce price");
-        deepEqual( tracker3.getCustomVariable(3, "page"), false, "No data Ecommerce view SKU");
-        deepEqual( tracker3.getCustomVariable(4, "page"), false, "No data Ecommerce view Name");
-        deepEqual( tracker3.getCustomVariable(5, "page"), ["_pkc",""], "No data Ecommerce view Category");
```
sgiehl, line 3912: it's no longer possible to receive the product view values, as they are no longer stored in custom variables. Not sure if we should add a `getProductView` method in order to be possible to test that better...

plugins/Ecommerce/Columns/ProductViewName.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function getDbDiscriminator()
+    {
+        return new Discriminator('log_action', 'type', Action::TYPE_ECOMMERCE_ITEM_NAME);
+    }
+
+    public function onLookupAction(Request $request, Action $action)
+    {
+        $sku = Common::unsanitizeInputValue($request->getParam('_pkn'));
+        if ($request->hasParam('_pkn')) {
```
tsteur, line 41: should maybe only getParam for sku if it has the param to not call like `unsanitzeInputValue(null)`

plugins/Ecommerce/Columns/ProductViewCategory.php
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public function getDbDiscriminator()
+    {
+        return new Discriminator('log_action', 'type', Action::TYPE_ECOMMERCE_ITEM_CATEGORY);
+    }
+
+    public function onLookupAction(Request $request, Action $action)
+    {
+        $categories = Common::unsanitizeInputValue($request->getParam('_pkc'));
```
tsteur, line 99: should maybe only getParam for category if it has the param to not call like unsanitzeInputValue(null)

plugins/Ecommerce/Columns/ProductViewPrice.php
```diff
@@ -0,0 +1,40 @@
...
+    protected $type = self::TYPE_MONEY;
+    protected $nameSingular = 'Ecommerce_ViewedProductPrice';
+    protected $columnName = 'product_price';
+    protected $segmentName = 'productViewPrice';
+    protected $columnType = 'DOUBLE NULL';
+    protected $category = 'Goals_Ecommerce';
+
+    public function onNewAction(Request $request, Visitor $visitor, Action $action)
+    {
+        $price = $request->getParam('_pkp');
```
tsteur, line 27: not sure if we would maybe consider `0` as a price? aka free?

plugins/Ecommerce/Columns/ProductViewSku.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function getDbDiscriminator()
+    {
+        return new Discriminator('log_action', 'type', Action::TYPE_ECOMMERCE_ITEM_SKU);
+    }
+
+    public function onLookupAction(Request $request, Action $action)
+    {
+        $sku = Common::unsanitizeInputValue($request->getParam('_pks'));
+        if ($request->hasParam('_pks')) {
```
tsteur, line 41: same here might be good to first check `hasParam` then `getParam`

plugins/Ecommerce/Columns/ProductViewPrice.php
```diff
@@ -0,0 +1,40 @@
...
+use Piwik\Tracker\Request;
+use Piwik\Tracker\Visitor;
+
+class ProductViewPrice extends ActionDimension
+{
+    protected $type = self::TYPE_MONEY;
+    protected $nameSingular = 'Ecommerce_ViewedProductPrice';
+    protected $columnName = 'product_price';
+    protected $segmentName = 'productViewPrice';
+    protected $columnType = 'DOUBLE NULL';
```
tsteur, line 22: not sure if that should be maybe unsigned? I see in other tables it's not unsigned but maybe here it would make sense? not sure... because it's double  and stores big values it might not be needed?

plugins/Goals/API.php
```diff
@@ -331,11 +332,29 @@ protected function getItems($recordName, $idSite, $period, $date, $abandonedCart
         $archive   = Archive::build($idSite, $period, $date, $segment);
         $dataTable = $archive->getDataTable($recordNameFinal);
 
-        $this->enrichItemsTableWithViewMetrics($dataTable, $recordName, $idSite, $period, $date, $segment);
+        // Before Matomo 4.0.0 ecommerce views were tracked in custom variables
+        // So if Matomo was installed before try to fetch the views from custom variables and enrich the report
+        if (version_compare(DbHelper::getInstallVersion(),'4.0.0-b2', '<')) {
+            $this->enrichItemsTableWithViewMetrics($dataTable, $recordName, $idSite, $period, $date, $segment);
```
tsteur, line 338: In the `enrichItemsTableWithViewMetrics` might need to check if the custom variables plugin is active since we will move it to the marketplace plus people might deactivate it?

plugins/Goals/API.php
```diff
@@ -331,11 +332,29 @@ protected function getItems($recordName, $idSite, $period, $date, $abandonedCart
         $archive   = Archive::build($idSite, $period, $date, $segment);
         $dataTable = $archive->getDataTable($recordNameFinal);
 
-        $this->enrichItemsTableWithViewMetrics($dataTable, $recordName, $idSite, $period, $date, $segment);
+        // Before Matomo 4.0.0 ecommerce views were tracked in custom variables
+        // So if Matomo was installed before try to fetch the views from custom variables and enrich the report
+        if (version_compare(DbHelper::getInstallVersion(),'4.0.0-b2', '<')) {
```
tsteur, line 337: btw getInstallVersion might not be set AFAIK as it was only added in 2018

plugins/Goals/API.php
tsteur: Might not be fully understanding it... are we there basically merging the metrics from custom variables with the metrics from log_link_visit_action if Matomo was installed before 4.0?

I suppose this would usually be mostly for data on the day they upgrade to Matomo 4 cause all previous day will have data from custom variables and all other ones from log_link_visit_action.

plugins/CustomVariables/Archiver.php
```diff
@@ -107,7 +108,9 @@ protected function aggregateCustomVariable($slot)
         // then we also query the "Product page view" price which was possibly recorded.
         $additionalSelects = false;
 
-        if (in_array($slot, array(\MatomoTracker::CVAR_INDEX_ECOMMERCE_ITEM_SKU, \MatomoTracker::CVAR_INDEX_ECOMMERCE_ITEM_NAME, \MatomoTracker::CVAR_INDEX_ECOMMERCE_ITEM_CATEGORY))) {
+        // Before Matomo 4.0.0 ecommerce views were tracked in custom variables
+        // So if Matomo was installed before still try to archive it the old way, as old data might be archived
+        if (version_compare(DbHelper::getInstallVersion(),'4.0.0-b2', '<') && in_array($slot, array(3, 4, 5))) {
```
tsteur, line 113: As mentioned below getInstallVersion might not return anything just fyi

tsteur (changes requested):
Left a few comments. Looks good overall. FYI @mattab this will basically add 8 new columns to `log_link_visit_action`

plugins/Ecommerce/Columns/ProductViewPrice.php
sgiehl: makes sense. Will change the check to `is_numeric`

plugins/Goals/API.php
sgiehl: That's basically for BC. If someone installs Matomo after 4 the variables will be directly tracked into the `log_link_visit_action` table, so there is no need to fallback to the old behavior.
If Matomo was installed before it might be possible that there were some ecommerce views tracked as custom variables. If those old data would be archived there wouldn't be any ecommerce views reported, so we try to get those data from the custom variables reports and merge that in. (That was the way it was done before).
Actually we should see in the UI tests that this behaviour works, as I didn't update the OmniFixture.sql, so it still includes ecommerce views as custom variables but is archived with the new logic before the tests run...

plugins/Goals/API.php
sgiehl: That should be fine I guess. `version_compare(null, '4.0.0-b2', '<')` returns `true`

plugins/Ecommerce/Columns/ProductViewPrice.php
sgiehl: Would leave it as signed as in theory someone might maybe want to track a voucher that has a negativ price maybe? 🤔

sgiehl:
@tsteur left some comments and applied the other feedback

tsteur:
@sgiehl tracking etc worked. Not sure if I was supposed to see this in the visits log? 
image
i tracked some data and didn't see it on hover but not sure if it maybe used to be shown? Eg in https://github.com/matomo-org/matomo/issues/12035#issuecomment-621700923 it looks like they used to be shown?

Otherwise worked nicely 👍  Could you maybe add it if needed to the visitor log, resolve merge conflicts and then merge?

PR #16001, opened 2020-05-29
Remove code tagged with `@deprecated` 

### LINKED ISSUE #8567: Remove code tagged with `@deprecated`

tsteur:
fix #8567 

Didn't remove all deprecated methods but many. Some I removed the deprecated flag or added a note they should be kept in the code (eg `API.getPiwikVersion`). Some of the remaining ones we will remove maybe in Matomo 5 or so. To be seen. 

Before the first Matomo 4 beta or RC we'll need to sort the CHANGELOG file :)


core/ArchiveProcessor/PluginsArchiver.php
```diff
@@ -81,11 +81,6 @@ public function __construct(Parameters $params, ArchiveWriter $archiveWriter = n
          *
          * @param bool $shouldAggregateFromRawData  Set to true, to aggregate from raw data, or false to aggregate multiple reports.
          * @param Parameters $params
-         * @ignore
```
tsteur, line 84: still good to have this event

core/Plugin.php
```diff
@@ -184,7 +184,6 @@ private function hasDefinedPluginInformationInPluginClass()
      * - 'theme' => bool                // Whether this plugin is a theme (a theme is a plugin, but a plugin is not necessarily a theme)
      *
      * @return array
-     * @deprecated
```
tsteur, line 187: not sure why this was deprecated but seems useful

core/Plugin/API.php
```diff
@@ -77,7 +77,7 @@ public static function getInstance()
     /**
      * Used in tests only
      * @ignore
-     * @deprecated
+     * @internal
```
tsteur, line 80: same here seems useful... however only for internal usage. deprecated flag was maybe set so it doesn't appear in the developer docs

core/Site.php
```diff
@@ -172,7 +172,6 @@ public static function setSiteFromArray($idSite, $infoSite)
      *                             array('idsite' => '2', 'name' => 'Site 2', ...),
      *                         )
      * @return array The modified array.
-     * @deprecated
```
tsteur, line 175: still needed but not as api

plugins/API/API.php
```diff
@@ -240,56 +227,6 @@ protected function getSegmentValuesFromVisitorLog($segmentName, $table)
...
-    }
-
-    /**
-     * Returns the URL to application SVG Logo
-     *
-     * @ignore
-     * @param bool $pathOnly If true, returns path relative to doc root. Otherwise, returns a URL.
-     * @return string
-     */
-    public function getSVGLogoUrl($pathOnly = false)
```
tsteur, line 276: the first 2 were marked as deprecated so figured I remove all logo related api methods

plugins/ExampleAPI/API.php
```diff
@@ -32,16 +32,6 @@ public function getMatomoVersion()
         return Version::VERSION;
     }
 
-    /**
-     * Get Matomo version
-     * @return string
-     * @deprecated
-     */
-    public function getPiwikVersion()
```
tsteur, line 40: the example plugin is removed in the release anyway so this is safe to remove

plugins/Login/SessionInitializer.php
```diff
@@ -21,11 +21,6 @@
 use Piwik\Session;
 
 /**
- * This SessionInitializer is no longer used, but is kept for backwards compatibility.
- * Session management no longer uses the piwik_auth cookie.
- *
- * @deprecated
- * @api
```
tsteur, line 28: the class is still used in the login class but we no longer consider it as API

plugins/VisitsSummary/Controller.php
```diff
@@ -40,22 +40,22 @@ public function __construct(Translator $translator)
     /**
      * @deprecated used to be a widgetized URL. There to not break widget URLs
      */
-    public function getSparklines()
+    public function index()
```
tsteur, line 43: had them first removed but be good to keep the widget urls working when it's that easy

core/Container/StaticContainer.php
```diff
@@ -13,7 +13,8 @@
 /**
  * This class provides a static access to the container.
  *
- * @deprecated This class is introduced only to keep BC with the current static architecture. It will be removed in 3.0.
+ * @internal
```
tolbon, line 16: I was just wondering how to delete all instances of code that depended on it, do you have any idea or an example of good migrations StaticContainer to another things ?

sgiehl: I guess the solution would be using dependency injection everywhere. We already have that in some places, but it's not yet usable everywhere...

tests/PHPUnit/Integration/Columns/DimensionTest.php
```diff
@@ -278,14 +278,6 @@ public function test_getId_ShouldCorrectlyGenerateIdFromDimensionsQualifiedClass
         $this->assertEquals("Test.DimensionTest", $this->dimension->getId());
     }
 
-    public function test_factory_ShouldCreateDimensionFromDimensionId()
-    {
-        Manager::getInstance()->loadPlugins(array('ExampleTracker'));
-
-        $dimension = Dimension::factory("ExampleTracker.ExampleDimension");
```
sgiehl, line 285: Might be useful to transform that into a test of `DimensionsProvider::factory` instead, as that one seems untested.

sgiehl (changes requested):

tests/PHPUnit/Integration/Columns/DimensionTest.php
tsteur: Done 👍

sgiehl (approved):

PR #16007, opened 2020-06-01
15583 backport

tassoman:
Backporting merge #15583 from 4.x-dev to 3.x-dev.

As superadministrator I would like to compare effective periods (days, weeks, months) instead of data ranges in order to avoid unwanted reports generation


tsteur:
Thanks for the PR @tassoman . Is it possible to undo the composer changes as part of this PR?

tassoman:
Thank you for the review. I'm sorry for being confused at git. That commit is part of another pull request :unamused: 
Maybe happened because I didn't checked out 3.x-dev before committing the code? :thinking:

tsteur:
Yes you'd probably want to do something like `git checkout -b new_branch_name 3.x-dev` so it uses 3.x-dev as a base.

sgiehl:
You can also try to cherry-pick the original merge commit on a 3.x-dev based branch: `git cherry-pick e8cb9c2ae10dceec3fde93acb58569bb8d9c1b59`

tassoman:
Ok. I managed it again, from scratch. Thank you for hinting!

tassoman:
#16015 for reference

PR #16010, opened 2020-06-02
Optional userLogin in getUserPreference API

tsteur:
Needed this for https://github.com/matomo-org/matomo-mobile-2/pull/5406

We're now allowing users to log in using their token auth. This way for example people using 
* LoginLdap
* Two factor authentication
* Matomo for WordPress (to be confirmed)

can also use the mobile app. It basically looks like this:

image

Because the user is no longer logging in with username/password, I no longer have a username to pass to the `getUserPreference` method. We call this API in the mobile app to find out which site to load by default and which date to show by default.

There was no API method to get the userLogin for an auth token. Could have added such an API but that would mean we always have two API calls when instead the method could simply default to the current user log in. 

Adding this to 3.X since the same Matomo mobile app will be used for Matomo 3 and Matomo 4 and by making this work in the latest 3.X we will have less issues with users trying to log in using token auth. This way we also don't need to wait for months to release this new update as people will be able to use it right away. (if they use a recent Matomo version)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)*
 
 ## Matomo 3.13.6
 
-### New API
+### API Changes
 * The first parameter `userLogin` in the `UsersManager.getUserPreference` method is now optional and defaults to the currently authenticated user login.
 
 ## Matomo 3.13.5
@@ -858,6 +858,7 @@ We are using `@since` annotations in case we are introducing new API's to make i
 
 ### Breaking Changes
 ### Deprecations
+### API Changes
 ### New features
 ### New APIs
 ### New commands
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ This is the Developer Changelog for Matomo platform developers. All changes in o
 
 The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)** lets you see more details about any Matomo release, such as the list of new guides and FAQs, security fixes, and links to all closed issues. 
 
+## Matomo 3.13.6
+
+### New API
```
sgiehl, line 9: `New API` doesn't seem to be the best headline here...

sgiehl (approved):
Left a small comment. Otherwise LGTM

CHANGELOG.md
tsteur: The others didn't quite work either which is why I used New API.... will add a new category `API changes`

PR #16017, opened 2020-06-04
Do not generate crossdomain link when cookies are disabled

tsteur:
Noticed this when testing on our site with cookies disabled. Basically in this case there is no visitorId unless a user sets it manually.

`getCrossDomainVisitorId()` was then only returning the `device ID` but not the visitorId and thus the generated link was invalid.

Tried to add a test like 
```
        tracker.setVisitorId(null); // also tried empty string etc
        tracker.setSiteId(99);

        replacedUrl = makeReplaceHrefForCrossDomainLink('http://www.example.com/');
        strictEqual(null, replacedUrl, 'replaceHrefForCrossDomainLink, should not return a link when visitorId not set');
```

but wasn't possible because the site had a cookie already set and Matomo would then always use that cookie overwriting any custom set value when the value is empty 
image

Shouldn't be needed the test anyway.


diosmosis (approved):
LGTM. If desired, we can add a test to JSTracker_spec.js where we disable cookies in puppeteer and make sure tracking generally works.

tsteur:
build js

PR #16020, opened 2020-06-05
Fix regression in HTTP Reporting API for bulk requests

### LINKED ISSUE #16008: Breaking HTTP API change in Matomo 4
see 

https://example.com/index.php?urls%5B0%5D=method%3dUsersManager.getUserPreference%26preferenceName%3ddefaultReport%26format%3dJSON%26userLogin%3droot%26&urls%5B1%5D=method%3dUsersManager.getUserPreference%26preferenceName%3ddefaultReportDate%26format%3dJSON%26userLogin%3droot%26&module=API&date=today&token_auth=YOURTOKEN&period=day&format=json&method=API.getBulkRequest&language=en&

In Matomo 3, this returns `["12","today"]`. Seems to return same with `json2` format.

vs in Matomo 4 this returns 

`[{"value":"12"},{"value":"today"}]`

And therefore breaks eg the Matomo mobile app.

Here's the request's individual params:

image

tsteur:
@sgiehl do you maybe remember why in https://github.com/matomo-org/matomo/pull/15558/files#diff-2d0bb82438495c9259aa4ef3211b6b36R527 we went for json over original?

Just tested eg below and seems an array of dataTables is still returned correctly. That was my first thought that this would maybe not work.

https://foo.bar.example.matomo/index.php?urls%5B0%5D=method%3dActions.getPageUrls%26%26idSite%3d1%26period%3dyear%26date%3dtoday%26preferenceName%3ddefaultReport%26format%3dJSON2%26userLogin%3droot%26&urls%5B1%5D=method%3dUsersManager.getUserPreference%26preferenceName%3ddefaultReportDate%26format%3dJSON2%26userLogin%3droot%26&module=API&date=today&token_auth=66613c5fe7e585bee9dd5ba56721b1b4&period=day&format=json&method=API.getBulkRequest&language=en&

fix https://github.com/matomo-org/matomo/issues/16008


sgiehl:
@tsteur Guess I did that as I thought it should be the smaller change, as the JSON format actually used the PHP format before. But guess it should be fine to use original as well.

tsteur:
fyi tried using `&format=original&serialize=0`

there are 2 problems:

1. It slightly changes formatting for XML when there is eg a map returned:

image

2. The test_getBulkRequest_IsAbleToHandleManyDifferentRequests fails because the first request is a permission error which throws an error. The original formatter won't catch the error and format it in any way meaning the exception is passed on to the `getBulkRequest` method meaning you only get one error compared to 1 error and 3 valid results within the bulk request response. So as soon as one request fails, all would fail.

When using serialize=1 with the original handler then we see the class names in the response: 
image

So far there are four solutions I can think of:

1. Add PHP handler back (but try to use it only within getBulkRequest to keep BC but that might be difficult and sounds like possible future issues as only bulk request uses it so it might not be as much tested etc)
2. Document the breaking change and keep existing code. It probably be hard to find out all possible changes. Maybe there aren't any further changes, maybe there are. Would need to create more tests.
3. A third option be in `getBulkRequest` to convert back an `{value: 5}` response to `5`. But not sure that would always work. Eg if an API returned actually `return array('value' => 5)`. I don't think that would really work.
4. Have an option in JSON renderer (not sure where this is actually done) to somehow ignore the `value` conversion in JSON datatable renderer but not quite sure how. 
image.

#4 sounds best but not really achievable unless we make some bigger changes.

It's also bit hard to know what regressions there are in other formats. Tempted to go with #1 or #2 . 

Any ideas @sgiehl ?

sgiehl:
@tsteur I'd go with 2. Readding the PHP handler wouldn't make things better. Also I think the bulk request feature shouldn't be used frequently and breaking the BC here shouldn't hurt too many people.

tsteur:
@sgiehl created https://github.com/matomo-org/matomo/pull/16041

Issue #16030, opened 2020-06-07
Update individual libraries through npm


diosmosis:
#14082 switched from bower to npm, but did not update every library, which still has to be done.

PR #16034, opened 2020-06-08
Minor npm updates & ui test fixes

sgiehl:
refs #16030


Findus23:
Updating chroma.js should also be possible as even though it is a major change it is only used for the color bars in maps.

sgiehl:
@Findus23  was wondering if that would have any benefit. The library seems to have a lot more features now, which increased the size quite a bit. See https://bundlephobia.com/result?p=chroma-js@0.6.3

sgiehl:
also we are actually planning to replace the maps with something new, so guess that lib might get removed anyways...

Findus23:
@sgiehl As the library is just a collection of equations it should be fine from a security point of view to stay with the old library.

Issue #12777, opened 2018-04-27
No setVisitorId in JavaScript tracking API?


olleharstedt:
Reading the docs, it seems like there's no `setVisitorId` in the JavaScript tracking API. Any reason for this? I have a big problem with a mismatch between the visitor id returned by the PHP API and the one returned by the JavaScript, which makes part of our logic unusable. I was hoping to bridge this mismatch by setting the visitor id = the one I got from the PHP API.

yohannprigent:
Any news on that ? It seems the method is public...

tsteur:
@yohannprigent you can't set the visitorId yet but a userId if that helps: https://matomo.org/docs/user-id

Not 100% sure why the method isn't public yet but maybe cause we'd then need to update the cookie again etc and there could be some complications if the setVisitorId is called too early when cookies aren't configured yet fully (eg cookie path, domain, ...) or if the idSite is still missing etc but we could kind of control this internally when it will be called. I guess it's just a bit more complicated then having this method public.

Would you expect this method to persist the visitorId in a cookie @yohannprigent ? Might be good to let us know a bit more about your use case so we can better understand things.

yohannprigent:
My use case is essentially to keep a consistent visitorId between different websites, on different subDomains. Matomo provides a great `enableCrossDomainLinking` function, but it's sad this functionality automatically stop tracking Outlinks from domains defined with `setDomains`, even if we don't want.
So I was thinking having a way to define a visitorId I would get from the other Matomo cookie on the other website (but on the same cookie domain). This is my use-case today, but I think if Matomo tracker would allow to select if we want outlinks from domainAlias to be tracked, my use case would not need to exist.

So yes, it would persist within the cookie in my opinion, as `setSiteId`does.

tsteur:
> Matomo provides a great enableCrossDomainLinking function, but it's sad this functionality automatically stop tracking Outlinks from domains defined with setDomains, even if we don't want.

Do you know if we have an issue for this already? Are you saying you define eg 

`setDomains(foo.bar, example.bar)`

Then you want `example.bar` to be still tracked an outlink? Is this basically because they aren't tracked into the same site in Matomo but different sites?

yohannprigent:
@tsteur Totally correct. Different domains tracked into different websites.
At my knowledge, we don't have issue for that.
[...]

Issue #15593, opened 2020-02-19
User-ID: Unique Visitor is not recognized even with the same User-ID


peterbo:
I'm setting a User-ID, when a user visits a given Site. On a certain action, I'm triggering a goal serverside with his User-ID as a parameter (and a token). Effect after the Update from 3.12 to 3.13.2 is that the serverside triggered action is not only stored in another visit (which would be ok), but also fails to recognize the web-visitor with the same User-ID. For reference, the before/after screenshots:

Before (Visitor recognized -> new visit but returning visitor):
![userid-before1](https://user-images.githubusercontent.com/1018973/74825044-82a6c500-5309-11ea-8c28-bc30cf131904.jpg)

After (Visitor is not recognized -> Visit is not returning -> new visit and new visitor):
![userid1](https://user-images.githubusercontent.com/1018973/74825088-9a7e4900-5309-11ea-83f9-dbbce2c70ea4.jpg)

Serverside call:
https://example.com/piwik.php?token_auth=XXX&cdt=2019-08-07 18:56:10&idgoal=3&revenue=1234&idsite=X&rec=1&r=13454&uid=1234567890

In config, trust_visitor_cookie is disabled.

The reason for that is this change: https://github.com/matomo-org/matomo/commit/ea5a14bdf8aa9608cdc2ab7d5c8236a5ff1eb3e2#diff-6700aaf1ce500fe51e284b9ec6f01b01

The change works in the right direction, but now, a User-ID is only assigned to the same visitor, when also the config_id matches. This doesn't make sense, because the main use case is for example a user who logs into a website with different devices (GDPR aside, but the User-ID is for example the customer ID). This user should be recognized as the same visitor (not necessarily the same visit, but at least the same visitor). @MichaelHeerklotz 

Refs https://github.com/matomo-org/matomo/pull/14360

MichaelHeerklotz:
This is intended. Since the referenced change the UserID no longer influences the visitor ID or visit ID.

peterbo:
Is this really what we want? In this case, the User-ID has no more added value than a named CustomDimension.

tsteur:
ping @mattab 

I'm not so much into this topic. I suppose it wouldn't help to do something like `setVisitorId(hashedUserId.substr(0,16)) (pseudo code)`?

I suppose in general the idea was maybe also that you can see what a user did before logging in and after as part of the same visit? But indeed tracking cross device is getting more complicated.

mattab:
[...]

PR #16042, opened 2020-06-08
Expose setVisitorId JS tracker method

tsteur:
Noticed only in https://github.com/matomo-org/matomo/issues/15593#issuecomment-640388105 that it's actually not exposed. As we need this in one of our projects ourselves adding this to 3.x-dev. Will reference that PR where we use it as well. Had this already implemented in another repo not knowing it wasn't working.

we'll also need to document this afterwards on the JS tracker API.

fixes https://github.com/matomo-org/matomo/issues/12777


tsteur:
build js

tests/javascript/index.php
```diff
@@ -3216,6 +3217,11 @@ function getVisitorIdFromCookie(tracker) {
         var visitorId = tracker.getVisitorId();
         equal(visitorId, tracker.getVisitorId(), "Visitor ID is the same when called multiple times");
 
+        tracker.setVisitorId('invalid'); // invalid characters
```
tsteur, line 3220: setVisitorId is also tested in many other places in the tests

sgiehl (approved):

PR #16049, opened 2020-06-10
Ensure to create security files for node_modules

### LINKED ISSUE #16048: HtAccess and web config files not created for node_modules
See eg https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/Installation/ServerFilesGenerator.php#L65
but also for IIS we need to create htaccess file / web config file for the `node_modules` directory.

Might need to trigger the `createFilesForSecurity()` method in an update to ensure they exist (or maybe we do this on every update not sure)

sgiehl:
fixes #16048

### DIFF
--- a/core/Updates/4.0.0-b2.php
+++ b/core/Updates/4.0.0-b2.php
@@ -9,6 +9,7 @@
 
 namespace Piwik\Updates;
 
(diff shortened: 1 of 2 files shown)


diosmosis (approved):
Works for me, left a comment otherwise can merge

PR #16052, opened 2020-06-11
Rename Piwik -> Matomo in JS tracker where possible

tsteur:
refs https://github.com/matomo-org/matomo/issues/12420


js/piwik.js
```diff
@@ -1464,22 +1465,26 @@ if (typeof window.Piwik !== 'object') {
 
         var content = {
             CONTENT_ATTR: 'data-track-content',
-            CONTENT_CLASS: 'piwikTrackContent',
+            CONTENT_CLASS: 'matomoTrackContent',
```
tsteur, line 1468: will need to later update docs to mention these new attributes... I don't think we actually deprecate the old ones

js/piwik.js
```diff
@@ -7121,12 +7152,12 @@ if (typeof window.Piwik !== 'object') {
     }
 
     if (window
-        && 'object' === typeof window.piwikPluginAsyncInit
-        && window.piwikPluginAsyncInit.length) {
+        && 'object' === typeof window.matomoPluginAsyncInit
```
tsteur, line 7155: this was never really documented (only a few weeks ago) so barely any plugin would make use of it and we can simply break BC

js/piwik.js
```diff
@@ -7135,12 +7166,16 @@ if (typeof window.Piwik !== 'object') {
         window.piwikAsyncInit();
```
tsteur, line 7166: we probably keep BC for `piwikAsyncInit` a long time past Matomo 5 etc

js/piwik.js
```diff
@@ -7135,12 +7166,16 @@ if (typeof window.Piwik !== 'object') {
...
+    if (!window.Matomo.getAsyncTrackers().length) {
+        // we only create an initial tracker when no other async tracker has been created yet in matomoAsyncInit()
         if (hasPaqConfiguration()) {
             // we only create an initial tracker if there is a configuration for it via _paq. Otherwise
-            // Piwik.getAsyncTrackers() would return unconfigured trackers
-            window.Piwik.addTracker();
+            // Matomo.getAsyncTrackers() would return unconfigured trackers
+            window.Matomo.addTracker();
         } else {
             _paq = {push: function (args) {
```
tsteur, line 7180: be good to keep using `_paq` I reckon.

js/piwik.js
```diff
@@ -7152,16 +7187,16 @@ if (typeof window.Piwik !== 'object') {
         }
     }
 
-    window.Piwik.trigger('PiwikInitialized', []);
-    window.Piwik.initialized = true;
+    window.Matomo.trigger('MatomoInitialized', []);
```
tsteur, line 7190: `PiwikInitialized` was never really documented or supported until few weeks ago so we can easily break BC

js/piwik.js
```diff
@@ -3901,7 +3927,7 @@ if (typeof window.Piwik !== 'object') {
              */
             function getClassesRegExp(configClasses, defaultClass) {
                 var i,
-                    classesRegExp = '(^| )(piwik[_-]' + defaultClass;
+                    classesRegExp = '(^| )(piwik[_-]' + defaultClass + '|matomo[_-]' + defaultClass;
```
tsteur, line 3930: will need to document the support of these new attributes eg `matomo_download`. We won't remove support for the `piwik` class names any time soon I suppose

tsteur:
build js

js/piwik.js
```diff
@@ -7184,11 +7219,11 @@ if (typeof window.Piwik !== 'object') {
  *
  * @param string documentTitle
  * @param int|string siteId
- * @param string piwikUrl
+ * @param string matomoUrl
  * @param mixed customData
  */
 if (typeof piwik_log !== 'function') {
-    piwik_log = function (documentTitle, siteId, piwikUrl, customData) {
+    piwik_log = function (documentTitle, siteId, matomoUrl, customData) {
```
diosmosis, line 7226: Do we want there to be a matomo_log, matomo_ignore_classes, etc. globals?

diosmosis (commented):
Left a comment, otherwise looks good

js/piwik.js
tsteur: @diosmosis that's there for BC so we can't change it. I guess 0.X or 1.X versions used this `piwik_log()`

diosmosis: I see 👍

diosmosis (approved):

PR #16053, opened 2020-06-11
Rename Piwik -> Matomo in cookies

tsteur:
refs https://github.com/matomo-org/matomo/issues/12420


core/Updates/4.0.0-b2.php
```diff
@@ -48,6 +48,8 @@ public function getMigrations(Updater $updater)
         ], ['idinvalidation']);
 
         $migrations[] = $this->migration->db->addIndex('archive_invalidations', ['idsite', 'date1', 'period'], 'index_idsite_dates_period_name');
+        // keep piwik_ignore for existing  installs
```
tsteur, line 51: important otherwise people who opted out previously would be counted now

PR #16058, opened 2020-06-11
Fixes #16057

### LINKED ISSUE #16057: Nonce::verifyNonce() returning FALSE when it should not
Configuration seems OK to me.

Debugging `Nonce::verifyNonce()` I noticed:

1.  `self::getOrigin();` returns `'https://somedomain.org:443'`
2. `self::getAcceptableOrigins()` returns `array('http://somedomain.org', 'https://somedomain.org')`

Thus, the returned origin is not found in the acceptable origins array.

Note the missing ports in the acceptable origins.

Matomo version 3.13.6.

luismsgomes:
Please see https://github.com/matomo-org/matomo/issues/16057 for details on what this commit is fixing.

### DIFF
--- a/core/Nonce.php
+++ b/core/Nonce.php
@@ -127,28 +127,26 @@ class Nonce
     public static function getAcceptableOrigins()
     {
         $host = Url::getCurrentHost(null);
-        $port = '';
-
-        // parse host:port
-        if (preg_match('/^([^:]+):([0-9]+)$/D', $host, $matches)) {
-            $host = $matches[1];
-            $port = $matches[2];
-        }
 
         if (empty($host)) {
             return array();
         }
 
-        // standard ports
-        $origins = array(
-            'http://' . $host,
-            'https://' . $host,
-        );
-
-        // non-standard ports
-        if (!empty($port) && $port != 80 && $port != 443) {
-            $origins[] = 'http://' . $host . ':' . $port;
-            $origins[] = 'https://' . $host . ':' . $port;
+        // parse host:port
+        if (preg_match('/^([^:]+):([0-9]+)$/D', $host, $matches)) {
+            $host = $matches[1];
+            $port = $matches[2];
+            $origins = array(
+                'http://' . $host . $port,
+                'https://' . $host . $port,
+            );
+        } else {
+            $origins = array(
+                'http://' . $host,
+                'https://' . $host,
+                'http://' . $host . ':80',
+                'https://' . $host . ':443',
+            );
         }
 
         return $origins;
--- a/core/Nonce.php
+++ b/core/Nonce.php
@@ -140,6 +140,11 @@ class Nonce
                 'http://' . $host . $port,
                 'https://' . $host . $port,
             );
+        } elseif (Config::getInstance()->General['force_ssl']) {
+            $origins = array(
+                'https://' . $host,
+                'https://' . $host . ':443',
+            );
         } else {
             $origins = array(
                 'http://' . $host,


core/Nonce.php
```diff
@@ -127,28 +127,26 @@ public static function getOrigin()
...
+            $port = $matches[2];
+            $origins = array(
+                'http://' . $host . $port,
+                'https://' . $host . $port,
+            );
+        } else {
+            $origins = array(
+                'http://' . $host,
+                'https://' . $host,
+                'http://' . $host . ':80',
```
tsteur, line 147: Thanks for the PR @luismsgomes 

btw always adding these ports not sure if it's a good idea or has any side effects. Wonder if we could remove the  two lines that always add `:80` and `:443` assuming that your `$host` is like `https://somedomain.org:443`

What is `Url::getCurrentHost(null);` returning for you?

luismsgomes: Hi,

The main motivation for this PR was exactly that `:80` and `:443` were missing from the list of urls.
The url returned by `Url::getCurrentHost(null);` contains `https://` and `:443` in my case.

Cheers.

core/Nonce.php
```diff
@@ -127,28 +127,26 @@ public static function getOrigin()
...
-        // non-standard ports
-        if (!empty($port) && $port != 80 && $port != 443) {
-            $origins[] = 'http://' . $host . ':' . $port;
-            $origins[] = 'https://' . $host . ':' . $port;
+        // parse host:port
+        if (preg_match('/^([^:]+):([0-9]+)$/D', $host, $matches)) {
+            $host = $matches[1];
+            $port = $matches[2];
+            $origins = array(
+                'http://' . $host . $port,
```
tsteur, line 140:
```suggestion
                'http://' . $host,
                'https://' . $host,
                'http://' . $host . $port,
```

core/Nonce.php
```diff
@@ -127,28 +127,26 @@ public static function getOrigin()
...
+                'https://' . $host . $port,
+            );
+        } else {
+            $origins = array(
+                'http://' . $host,
+                'https://' . $host,
+                'http://' . $host . ':80',
+                'https://' . $host . ':443',
```
tsteur, line 148:
```suggestion
```

core/Nonce.php
tsteur, line 147:
```suggestion
```

tsteur: @luismsgomes added few suggestions to prevent we're regressing here on anything.

Basically could even use the logic completely but only change the `if (!empty($port) && $port != 80 && $port != 443) {` to `if (!empty($port)) {`

Basically.... if no port is configured, maybe we don't want to add `80` and `443`. I think this be still fine for your case. It might be fine to add it just not sure and while it's not needed be good to maybe not add it automatically

tsteur (changes requested):

core/Nonce.php
luismsgomes: Sorry, my last reply was incorrect:  getCurrentHost() returns only the hostname, no scheme and no port in my case.
The function getOrigin() does return the scheme and the port.

tsteur: I see 👍 I'll undo my suggestions. @Findus23 I reckon it be fine to also allow the port as we can assume if no port is returned it is running on either port 80 or 443? Maybe if SSL is forced (`[General]force_ssl=1`), then we would not return the URL with port `80`?

luismsgomes: > Maybe if SSL is forced ([General]force_ssl=1), then we would not return the URL with port 80?

Yes, that makes sense.

tsteur: @luismsgomes  can you make that change or do you need help with this?

luismsgomes: Done.

tsteur:
Looks generally good to me. @Findus23 @sgiehl @diosmosis  any other thoughts?

Generally thinking because there are some minor changes in there that could break something we might prefer to have this in Matomo 4 and not Matomo 3 where we now only fix security issues and bigger bugs.

luismsgomes:
Well, this is big for me since without this patch I couldn't login ;-)
That's why I took the effort to fix it.

tsteur:
@luismsgomes sorry about that. There will be only very few releases for 3.X in the future so it might be worth patching it in the worst case. Maybe we could think of a less "risky" version for 3.X where it doesn't change as much. Eg for the case where a port is defined like in https://github.com/matomo-org/matomo/pull/16058/files#diff-e3d1b64ec2cd9e138ad8d5c1a0d09a0eR140 we are now no longer returning the URLs without port. This could be causing trouble to other users maybe (as it's always hard to tell with these things)

tsteur:
@Findus23 can you think of any objections or do you reckon it should be fine?

Findus23:
I know little about how this part works, so I can't comment on it.

diosmosis:
@tsteur missed this issue, looks ok to me, `Nonce::getAcceptableOrigins()` is only used in Nonce so it won't affect anything else. I can't think of a problem w/ treating http://somesite.com & http://somesite.com:80 as equivalent. Might be good to add some tests to NonceTest.

PR #16061, opened 2020-06-12
Document which APIs break in particular for JSON response

### LINKED ISSUE #16009: Breaking HTTP API change in Matomo 4 in getSiteFromId
This time it is `SitesManager.getSiteFromId`. Not sure if related to https://github.com/matomo-org/matomo/issues/16008

Request is eg 
```
idSite=1&module=API&date=today&token_auth=YOURTOKEN&period=day&format=json&method=SitesManager.getSiteFromId&language=en&
```

It returns an object like this:

image

When it should return an array with one object in there.

This breaks for example the Matomo Mobile app.

Should this change be expected because of JSON2 being applied now, then we'd probably need to mention every API method whose output changes now.

tsteur:
fix https://github.com/matomo-org/matomo/issues/16009

These were the API's I could identify. Should there be more then we'll need to add them later too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,13 @@ The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)*
 * The signature of `Dimension::configureSegments()` has been changed. Similar to configuring Metrics it now takes two parameters `SegmentsList $segmentsList` and `DimensionSegmentFactory $dimensionSegmentFactory`.
 * The method `Dimension::addSegment()` has been removed. See new implementation of `DimensionSegmentFactory::createSegment` for a replacement
 * The signature of the event `Segment.addSegments` has been changed. It now has one parameter `SegmentsList $list`, which allows adding new segments to the list
-* The json2 API format has now been deprecated, and the json renderer now behaves as the json2 renderer did. This means when `format=json` is used, arrays like `['a' => 0, 'b' => 1]` will be rendered in JSON as `{"a":0,"b":1}` instead of `[{"a":0,"b":1}]`. The JSON2 renderer will be removed in Matomo 5 and we recommend switching to it.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -75,7 +75,13 @@ The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)*
...
 * The method `Dimension::addSegment()` has been removed. See new implementation of `DimensionSegmentFactory::createSegment` for a replacement
 * The signature of the event `Segment.addSegments` has been changed. It now has one parameter `SegmentsList $list`, which allows adding new segments to the list
-* The json2 API format has now been deprecated, and the json renderer now behaves as the json2 renderer did. This means when `format=json` is used, arrays like `['a' => 0, 'b' => 1]` will be rendered in JSON as `{"a":0,"b":1}` instead of `[{"a":0,"b":1}]`. The JSON2 renderer will be removed in Matomo 5 and we recommend switching to it.
+* The json2 API format has now been deprecated, and the json renderer now behaves as the json2 renderer did. This means when `format=json` is used, arrays like `['a' => 0, 'b' => 1]` will be rendered in JSON as `{"a":0,"b":1}` instead of `[{"a":0,"b":1}]`. The JSON2 renderer will be removed in Matomo 5 and we recommend switching to it. This impacts these API methods:
+  * API.getSettings
+  * Annotations.get
+  * Goals.getGoal
+  * UsersManager.getUser
+  * UsersManager.getUserByEmail
+  * SitesManager.getSiteFromId
```
sgiehl, line 84: Wondering if it would make sense to add tests for those methods. Guess they are currently untested, as we didn't have any changes in expected results 🤔  Not needed for this PR though, but might be worth to create an issue maybe

tsteur: Was wondering this initially as well. Noticed we're using XML format in system tests and tests are actually covered by json renderer test itself. There shouldn't be a need to run each method with every format as be bit too much and making build slower etc. We even knew it was breaking API it was just not clear which ones are affected

PR #16064, opened 2020-06-14
do not prefix url if it is about:blank in ui tests

diosmosis:


sgiehl (approved):
Guess that explains the problems I had a while back with loading a blank page 🙈

PR #16066, opened 2020-06-14
Makes report export overlay compatible with app specific tokens

### LINKED ISSUE #16043: Export of reports no longer working in Matomo 4
Was just testing https://github.com/matomo-org/matomo/pull/15999 and then went to Ecommerce log -> Export and noticed I always get an error `Your session has expired due to inactivity. Please log in to continue.`. I suppose this applies to all reports through the UI report export popover.

Possible this is due to the app specific token change.

sgiehl:
The export popover will now allow to choose between session auth or using a custom app specific token:

image

fixes #16043


tsteur:
image
image

Should the place holder and description be maybe "app specific token auth"?

Could maybe also have a link to manage the tokens so user can create them easily? 
image

Not really sure though if this is actually needed? Not sure what the value be of being able to enter the token auth? It might again rather encourage to share the token?

Code works though 👍 

BTW could we add here: https://github.com/matomo-org/matomo/blob/4.x-dev/core/Access.php#L177 a `Session::close()`? Only seeing this now... makes sure we close the session again immediately which increases performance. Would do it maybe only if authentication using that token was successful.

sgiehl:
> Should the place holder and description be maybe "app specific token auth"?

Was thinking about that as well. But actually we are naming the tokens that way nowhere. In security settings they are simply called "auth tokens".

> Not really sure though if this is actually needed? Not sure what the value be of being able to enter the token auth? It might again rather encourage to share the token?

That's correct. But actually someone needs to enter an auth token and so he is doing that kind of on purpose...

sgiehl:
> Could maybe also have a link to manage the tokens so user can create them easily?

Tried that, but seems not that easy as we are using an angular radio field there, and HTML seems not to work in the option titles

tsteur:
All good about the other comments 👍 

> That's correct. But actually someone needs to enter an auth token and so he is doing that kind of on purpose...

I'm still not sure it's really needed that users can configure their actual token though? Not sure what the purpose is and when they would use it?

sgiehl:
> I'm still not sure it's really needed that users can configure their actual token though? Not sure what the purpose is and when they would use it?

If someone wants to use the generated url for anything outside of Matomo or wants to bookmark it for later usage for example. The session url wouldn't work for this...

tsteur:
@sgiehl we rather not have this option for now to keep the UI easy and simple. I reckon only very few users would actually use it (like less than 5%) that way and then they could still simply replace the token in the URL manually. If few people ask for it later then we could still add it.

sgiehl:
@tsteur  Should we add some kind of note then, that the generated URL will only work in the current session? Might otherwise be confusing for people why it's not possible anymore to use the link somewhere else.

tsteur:
I reckon it's not needed. It's like any other link doesn't work somewhere else either. Of course this one used to work. Not sure if you have anything specific in mind?

sgiehl:
@tsteur We could maybe show something like:
`Note: The generated export URL will only work in the current browser session. If you want to use it somewhere else you need to use an app specific token.`

We could show the note maybe only when someone clicks show export url

tsteur:
@sgiehl let's maybe show it as tooltip when they hover the export button, hide export URL or the textarea that is showing the URL.

Maybe add a sentence: 
> Note: The generated export URL will only work in the current browser session. If you want to use it somewhere else you need to use an app specific token. You can configure these tokens in Admin -> Security -> Token Auths.

sgiehl:
@tsteur applied the changes.

PR #16068, opened 2020-06-15
Mention users should not share their token with anyone

tsteur:
while reviewing https://github.com/matomo-org/matomo/pull/16066 noticed we should mention this.


plugins/UsersManager/lang/en.json
```diff
@@ -18,9 +18,10 @@
         "AuthTokenPurpose": "What are you using this token for?",
         "NoTokenCreatedYetCreateNow": "No token created yet, %1$screate a new token now%2$s.",
         "TokenSuccessfullyGenerated": "Token successfully generated",
-        "ConfirmTokenCopied": "I confirm I copied the token.",
+        "ConfirmTokenCopied": "I confirm I copied the token and understand I should not share this token with anyone else.",
```
sgiehl, line 21: Not sure if that text might not get a bit too long when translated to other languages. Current layout will cause the text to be cut of instead of having a multiline button...

tsteur: Made it multiline 👍

sgiehl (approved):

PR #16069, opened 2020-06-15
Add legacy autoloader to support Matomo namespaces in 3.X 

tsteur:
Refs https://github.com/matomo-org/matomo/issues/12420#issuecomment-644254519

The purpose is to make updates to a future Matomo version smoother where we maybe use `Matomo` namespace instead of `Piwik`

Tested it eg in 3.x-dev by adding a file `core/Foo.php` with content 

```php
<?php
namespace Matomo;

class Foo {
    public function get()
    {
        return '42';
    }
}
```

then in `core/dispatch.php` tested it using 

```php
$x = new \Piwik\Foo();
echo $x->get();
```
 and using 

```php
$x = new \Matomo\Foo();
echo $x->get();
```

In all cases the class was resolved correctly with no issues.

### DIFF
--- a/LegacyAutoloader.php
+++ b/LegacyAutoloader.php
@@ -17,7 +17,7 @@ class LegacyAutoloader
         if (strpos($className, 'Matomo\\') === 0) {
             $newName = 'Piwik' . substr($className, 6);
             if (class_exists($newName)) {
-                class_alias($newName, $className);
+                @class_alias($newName, $className);
             }
         }
     }
--- a/LegacyAutoloader.php
+++ b/LegacyAutoloader.php
@@ -1,0 +1,26 @@
+<?php
+
+class LegacyAutoloader
+{
+    public function __construct()
+    {
+        spl_autoload_register(array($this, 'load_class'));
+    }
+
+    public static function register()
+    {
+        new LegacyAutoloader();
+    }
+
+    public function load_class($className)
+    {
+        if (strpos($className, 'Matomo\\') === 0) {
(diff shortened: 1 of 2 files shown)


LegacyAutoloader.php
```diff
@@ -0,0 +1,31 @@
...
+    public static function register()
+    {
+        new LegacyAutoloader();
+    }
+
+    public function load_class($className)
+    {
+        if (strpos($className, 'Matomo\\') === 0) {
+            $newName = 'Piwik' . substr($className, 6);
+            if (class_exists($newName) && !class_exists($className, false)) {
```
tsteur, line 19: btw the second `class_exists` is needed because things could happen like 

```
* new \Matomo\Foo
* It then does `class_exists(\Piwik\Foo)`
* It then loads `core/Foo.php`
* This class defines `Matomo\Foo`
* Then it adds alias `Piwik\Foo` to `Matomo\Foo`
* This first `class_exists($newName)` basically caused the file `core/Foo.php` to load which defines `Matomo\Foo` so we no longer want to execute `class_alias(\Matomo\Foo, Piwik\Foo)` since the first class_exists actually defined the class already indirectly.
```

sgiehl (approved):
Looks good. We just need to keep in mind, that if we merge this to 3.x-dev, the changes to LegacyLoader would get merged to 4.x-dev later. So if we don't want to have the changes in 4.x-dev, we need to undo them in the next merge to 4.x-dev

tsteur:
👍

PR #16070, opened 2020-06-15
Add legacy autoloader for 2.x-dev for smoother upgrades

tsteur:
refs https://github.com/matomo-org/matomo/issues/12420 see https://github.com/matomo-org/matomo/pull/16069 for explanation.

We need to create a new 2.X release for this.

### DIFF
--- a/LegacyAutoloader.php
+++ b/LegacyAutoloader.php
@@ -1,0 +1,31 @@
+<?php
+
+class LegacyAutoloader
+{
+    public function __construct()
+    {
+        spl_autoload_register(array($this, 'load_class'));
+    }
+
+    public static function register()
+    {
+        new LegacyAutoloader();
+    }
+
+    public function load_class($className)
+    {
+        if (strpos($className, 'Matomo\\') === 0) {
+            $newName = 'Piwik' . substr($className, 6);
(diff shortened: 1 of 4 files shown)


core/Plugin/ControllerAdmin.php
```diff
@@ -207,7 +207,7 @@ private static function notifyIfEAcceleratorIsUsed()
      */
     private static function getNextRequiredMinimumPHP()
     {
-        return '5.5.9';
+        return '7.2.5';
```
tsteur, line 210: benefit is that users will get a better warning

LegacyAutoloader.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        new LegacyAutoloader();
+    }
+
+    public function load_class($className)
+    {
+        if (strpos($className, 'Matomo\\') === 0) {
+            $newName = 'Piwik' . substr($className, 6);
+            if (class_exists($newName) && !class_exists($className, false)) {
+                @class_alias($newName, $className);
```
tsteur, line 20: was added in 5.3.0 so that should be fine https://www.php.net/manual/en/function.class-alias.php

sgiehl (approved):

PR #16073, opened 2020-06-16
Removes double escaping in siteselector tooltip

### LINKED ISSUE #16072: The display problem of piwik-siteselector in asia language environment
Hi,I got a trouble of piwik-siteselector in asia language environment.If my site name is Asian, this control will not display the content correctly when the mouse is moved.I checked the source code and found that the problem was caused by escape of the twig template engine. if remove the escape filter,My site name can be displayed correctly.
微信图片_20200616095428

微信图片_20200616093225

sgiehl:
checked with website titles like `<b>test</b>` or `{{CONSTRUCTOR.CONSTRUCTOR("_X()")()}}` if removing the escape opens up some html or angular execution, but seems the content is still encoded once.

fixes #16072

### DIFF
--- a/plugins/CoreHome/angularjs/siteselector/siteselector.directive.html
+++ b/plugins/CoreHome/angularjs/siteselector/siteselector.directive.html
@@ -16,7 +16,7 @@
     <a ng-click="view.showSitesList=!view.showSitesList; view.showSitesList && !model.isLoading && model.loadInitialSites();"
        piwik-onenter="view.showSitesList=!view.showSitesList; view.showSitesList && !model.isLoading && model.loadInitialSites();"
        href="javascript:void(0)"
-       title="{{ 'CoreHome_ChangeCurrentWebsite'|translate:((selectedSite.name || model.firstSiteName)|escape) }}"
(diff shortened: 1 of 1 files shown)


plugins/CoreHome/angularjs/siteselector/siteselector.directive.html
```diff
@@ -16,7 +16,7 @@
     <a ng-click="view.showSitesList=!view.showSitesList; view.showSitesList && !model.isLoading && model.loadInitialSites();"
        piwik-onenter="view.showSitesList=!view.showSitesList; view.showSitesList && !model.isLoading && model.loadInitialSites();"
        href="javascript:void(0)"
-       title="{{ 'CoreHome_ChangeCurrentWebsite'|translate:((selectedSite.name || model.firstSiteName)|escape) }}"
+       title="{{ 'CoreHome_ChangeCurrentWebsite'|translate:((selectedSite.name || model.firstSiteName)) }}"
```
mattab, line 19: what about using `|e('html')` for the html attribute?

tsteur: you can't use that in an attribute

tsteur: also btw this is angular not twig

PR #16074, opened 2020-06-16
Use Python 3 for Log Analytics

sgiehl:
With Matomo 4 we should consider dropping the support for Python2 in log analytics completely.

This PR is based on the changes of https://github.com/matomo-org/matomo-log-analytics/pull/267
And is primarily there to prove all Matomo tests are running smoothly with Python 3


sgiehl:
was already done with #16095

Issue #8415, opened 2015-07-23
Update our Develop guides for Matomo 4.0.0 


mattab:
The goal of this issue is to go over our [Develop guides](http://developer.piwik.org/develop) and update them to match the new APIs in Matomo 4.0.0. See also: https://github.com/matomo-org/matomo/wiki/New-major-releases#developermatomoorg

tsteur:
Just mentioning it so we don't forget to adjust the examples mentioned in https://github.com/matomo-org/matomo/issues/6548#issuecomment-639116673

tsteur:
Also document no  longer supported browsers in tracking cause we removed JSON and how to still support them maybe

PR #16077, opened 2020-06-17
Organise Changelog better

tsteur:
refs https://github.com/matomo-org/matomo/issues/8415

We will likely tweak it again before the release.


sgiehl (approved):
Looks better this way 👍

PR #16079, opened 2020-06-17
run npm update and include node_modules

### LINKED ISSUE #16030: Update individual libraries through npm

diosmosis:
Fixes #16030

Also includes the node_modules folder as discussed on slack to keep matomo working for users who do not have servers with node and deploy from git.


package.json
```diff
@@ -31,19 +31,19 @@
     "angular-cookies": "^1.8.0",
     "angular-mocks": "^1.8.0",
     "angular-sanitize": "^1.8.0",
-    "chroma-js": "^0.6.3",
+    "chroma-js": "^2.1.0",
```
sgiehl, line 34: Actually not sure re this update. doesn't have any benefits, but increased in size a lot. See https://github.com/matomo-org/matomo/pull/16034#issuecomment-640496350

sgiehl:
@diosmosis guess if we include the node_modules in core, we might need to remove the automatic `npm update` in travis builds again? so we can be sure all files that are needed are actually commited

tsteur:
Yes that be good 👍

tsteur:
👍 looks all good from my perspective... be good if @sgiehl confirms too

sgiehl:
Looks good to me too. Guess we need to merge the change to travis-scripts to master and update the submodule here before merging...

Issue #11524, opened 2017-03-24
Disable Custom Variables feature by default


mattab:
We need to deprecate Custom Variables in favor of Custom Dimensions.

In Piwik 4, maybe we could deactivate the CustomVariables plugin, if no Custom Variable has been tracked so far in this Piwik?

Additionally we would need to find ways to promote the Custom Dimensions feature within the product. By showing some ads or notifications in some way.

Thoughts?

tsteur:
Sounds good. I reckon in general it will be a hard transition to get rid of it in terms of providing a good experience to users who still use it. We could for sure mention it in Custom Variables report etc.

tsteur:
We could disable it for new clients on cloud as well and only enable it on request eg when importing data or when they ask for it. We still mention it in docs quite often and users are probably more aware of Custom Variables than Custom Dimensions. Also need to remove it from tracking code generator, or better: make the generator extensible so that the custom variables plugin adds itself to the generator.

tsteur:
Also what about custom dimensions in scope event?

mattab:
@tsteur AFAIK the same benefits can be achieved with custom dimensions on events?

### Ideas so far for Custom Variables deprecation

* display a message below Custom Variables reports, where we recommend to use Custom Dimensions instead
*  in Piwik 4, for new installations, do not enable Custom Variables
* in Piwik 4, for users upgrading, possibly we can detect whether any custom variable has been used so far in this Piwik server, and if never used, disable the plugin automatically
* possibly we can go as far as removing the source code from Piwik release package, and move this plugin to the Marketplace?

mattab:
* Note: custom variables is also used in the Log Importer to set Http-Status codes, whether the user is a bot or not, etc. This would need some adjustements

mattab:
Before we can disable custom variables in core, we need to change any core features which currently use Custom Variables (all of it is "hacky"):

* Ecommerce view tracking uses custom variables, which leads to bugs such as https://github.com/piwik/piwik/issues/12035 -> we should rewrite Ecommerce tracking to work without custom variables.
* Custom vars also used in the Log Importer to set Http-Status codes, whether the user is a bot or not, etc. This would need some adjustements ([refs](https://github.com/piwik/piwik-log-analytics/pull/170/files))
* Custom variables also used for Site Search tracking, implementation should be rewritten see #12072

mattab:
[...]

PR #16090, opened 2020-06-18
Removes CustomVariables plugin from core

sgiehl:
This PR aims to remove the CustomVariables plugins from core. In a later PR we will readd it as submodule. But as it will no longer be bundled with core, we need to ensure Matomo 4 runs correctly with the plugin removed (completely)

Note: There are still a lot hard requirements for CustomVariables in core. I will try to remove them if they are not needed or move them to the new plugin repo step by step until the tests are running again.

@tsteur Shall we move the custom variables part of `piwik.js` to the plugin as well? Or should that remain in core?

This PR needs to be reviewed together with #16104

refs #11524


tsteur:
@sgiehl lets keep the tracking code in piwik.js for now and create an issue to move it to the plugin in Matomo 5. This should ensure we keep the update to Matomo 4 easier in case some people don't have the matomo.js writable etc.

sgiehl:
Note: There are still some parts left in core (like tracking or configuring custom variables in tracking code generator)

tsteur:
@sgiehl looks generally good to merge once the tests pass (they currently don't by the looks).

* BTW in tracker model can maybe remove this:
image

* Also in `CoreAdminHome/Controller::trackingCodeGenerator()` the custom variables class is still accessed without checking if plugin is activated? (for `maxCustomVariables`). Getting eg this error on tracking code page: `Class 'Piwik\Plugins\CustomVariables\CustomVariables' not found in /Users/thomassteur/Development/piwik/plugins/CoreAdminHome/Controller.php line 221  `

* In Goals VisitorDetails is a use statement that can be removed `use Piwik\Plugins\CustomVariables\CustomVariables;`

* `plugins/API/SegmentMetadata.php` the `sortSegments()` uses a translation key for customvariables but seems that's fine and won't need changing as it wouldn't be set anyway if it's not activated...

sgiehl:
@tsteur I've applied your feedback. Most tests should work correctly now. Currently failing are: Some submodule plugin tests, a mysqli fail (see #16108) and a lot of UI tests, as they changed a lot due to the removed plugin.

How shall we proceed? Should I update all tests, so we can merge this PR and later include the plugin as submodule plugin, which is disabled by default in tests? Or we merge #16104 into this one (after it and https://github.com/matomo-org/plugin-CustomVariables/pull/1 was reviewed), which should fix all failing tests, but would have the plugin as enabled in tests (like Bandwidth and others)

tsteur:
> Or we merge #16104 into this one (after it and matomo-org/plugin-CustomVariables#1 was reviewed), which should fix all failing tests, but would have the plugin as enabled in tests (like Bandwidth and others)

sounds good

tsteur:
@sgiehl one more fixture to update otherwise good to merge 👍 

I suppose this test failure is unrelated https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/41362/UIIntegrationTest_email_reports_editor.png

sgiehl:
Will update and merge that now

PR #16100, opened 2020-06-22
also support python 3.8

Findus23:
followup to #16095

There should be no reason that log analytics doesn't work in 3.8 and the travis test runs for it:
https://github.com/matomo-org/matomo-log-analytics/blob/4.x-dev/.travis.yml#L7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,7 +126,7 @@ The following changes are recommendations and don't have to be done as we will s
 
 ### Other Breaking changes
 
-* The log importer in `misc/log-analytics` now supports Python 3 (3.5, 3.6, 3.7, 3.8), it will no longer run with Python 2. If you have any automated scripts that run the importer, you will have to change them to use the Python 3 executable instead.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -126,7 +126,7 @@ The following changes are recommendations and don't have to be done as we will s
 
 ### Other Breaking changes
 
-* The log importer in `misc/log-analytics` now supports Python 3 (3.5, 3.6 or 3.7), it will no longer run with Python 2. If you have any automated scripts that run the importer, you will have to change them to use the Python 3 executable instead.
+* The log importer in `misc/log-analytics` now supports Python 3 (3.5, 3.6, 3.7 or 3.8), it will no longer run with Python 2. If you have any automated scripts that run the importer, you will have to change them to use the Python 3 executable instead.
```
tsteur, line 129: fyi added a note to https://github.com/matomo-org/matomo/wiki/Matomo-4.0.0-release-notes

Can people maybe still checkout an earlier version from the log importer repository supporting Python 2? Could mention it maybe but should be enough to have this mentioned in the announcement as it's not really developer related.

Findus23: It should definitely be mentioned in the release notes with maybe a link to an FAQ explaining how to switch over.

I don't know of any breaking change that would cause the old script to not work with Matomo  4

PR #16104, opened 2020-06-23
Readds Custom Variables as submodule

sgiehl:
This PR is based on #16090 and should be merged into it before merging it.

Reason for having two PRs is to check that Matomo (the tests) are running without the plugin (see build results from #16090). That's important to prove Matomo 4 will also run without the plugin being installed at all. This PR readds the plugin as a submodule and updates the expected test files again...

Note: Some submodule plugin tests are still failing and should be updated when merging...


diosmosis:
Is there a reason to even keep it as a submodule for 4.0?

tsteur:
Yes we're still supporting it for a while

diosmosis (approved):
some submodule test failures, but looks good

PR #16112, opened 2020-06-24
Prevent phishing attack on reset your password

tsteur:
fix https://github.com/matomo-org/matomo/issues/11071

As discussed in issue as a quick fix we ask for password confirmation before updating the password


diosmosis (approved):
LGTM, there's still a couple test failures

tsteur:
Fixed the tests

Issue #13056, opened 2018-06-11
Add possibility to turn cookies back on, JS enableCookies


WouterFlorijn:
I'm working on a website that allows visitors to toggle cookies on or off. Currently, with Matomo I can turn them off using

```js
window._paq.push(['disableCookies'])
```

I searched for a way to turn them back on, and tried

```js
window._paq.push(['enableCookies'])
```

but that did not work.

Is there a way to turn cookies back on? If not, it would be great if it could be added.

Note from tsteur: Also need to look at this when working on it: https://github.com/matomo-org/matomo/issues/13056#issuecomment-646278244

sgiehl:
If the page is reloaded and `disableCookies` isn't pushed the cookies should be automatically used "again".
But for singe page apps it would make sense to have a method to turn the cookies back on.

WouterFlorijn:
@sgiehl That would be great. Not only for SPAs, but any website where the user can change their cookie consent settings on the fly, which is relevant with the GDPR.

bjoe2k4:
Any Update on this? Interested in this as well!

martin-braun:
The biggest issue on this is not that we cannot re-enable cookies. As a bypass, we can simple reload and prevent pushing 'disableCookies'. However, if we use a [custom opt-out form](https://developer.matomo.org/guides/tracking-javascript-guide#optional-creating-a-custom-opt-out-form), it will not work after disabling cookies, because the opt-out form requires cookies to be enabled. 

It would be great to be able to allow cookies by JS without reload and make it work, so the custom opt-out solution can work. In it's current design, it's impossible to prevent conventional Matomo cookies and still getting the Opt-Out cookie.

It's been 1 1/2 years, this needs to be addressed, to be honest.

ctholho:
I'm also looking for this functionality. Is there any other way to reenable cookies without a reload?

I don't understand how this flies with GDPR and cookie consent regulation in the EU and other territories. I don't want to disable tracking entirely. Simply cookies on/off is enough, I would have thought.

futureweb:
Would also need this Feature, we use Matomo's disableCookies Feature + Klaro! as Consent Manager ... of course if User gives Consent I want to use Cookies again (1-Pager, ...), so it would be benefical if one could re-activate Cookies by JS

utrautmann:
> If the page is reloaded and disableCookies isn't pushed the cookies should be automatically used "again".

[...]

PR #16113, opened 2020-06-25
Add method to enable cookies in JS

tsteur:
We're putting this into 3.X since it's rather urgent eg re some recent court decisions etc.

refs https://github.com/matomo-org/matomo/issues/13056

User confirmed it works... Will need to add some docs once merged and update some pages. Tested it myself as well and worked. It's basically the same as calling this: `_paq.push(['setSiteId', YOUR_IDSITE]);_paq.push(['ping']);`

Basically it's supposed to work like this:

```js
window._paq.push(['disableCookies']);
```

once you have consent you call

```js
window._paq.push(['enableCookies']);
```

On the next page view you can either not call `disableCookies` anymore or you call both but standard behaviour would be to not call `disableCookies` anymore.

```js
window._paq.push(['disableCookies']);
window._paq.push(['enableCookies']);
window._paq.push(['setTrackerUrl', u+'matomo.php']);
window._paq.push(['setSiteId', '1']);
//...
```

Enable cookies always takes precedency over disable cookies to make this flow work. This is simply because by default cookies are enabled anyway.

What would cause issues is to do this on every page view manually:

```js
tracker.setSiteId(5);
tracker.disableCookies();
tracker.enableCookies();
```

If someone was to execute above code the cookie would change on every request as it's not correctly set up. `disableCookies` needs to be set before `setSiteId` as otherwise all existing cookies would be deleted.

Some of the tests are changing because we now always overwrite the visitorId of the current visit when the visitorId changes in the JS tracker. Before visitorId was never updated during any visit. Eg the visitorId changes when:
* Cookies are disabled. The backend generates a random visitorId not known to the JS tracker
* Enable cookies. Now JS generates a random visitorId and we want to update the existing visit to enforce this ID


tsteur:
build js

tsteur:
build js

tsteur:
build js

tsteur:
build js

tsteur:
build js

core/Tracker/Visit.php
```diff
@@ -553,11 +553,16 @@ private function getVisitStandardLength()
...
     private function setIdVisitorForExistingVisit($valuesToUpdate)
     {
-        // Might update the idvisitor when it was forced or overwritten for this visit
         if (strlen($this->visitProperties->getProperty('idvisitor')) == Tracker::LENGTH_BINARY_ID) {
             $valuesToUpdate['idvisitor'] = $this->visitProperties->getProperty('idvisitor');
         }
-
+        
+        $visitorId = $this->request->getVisitorId();
+        if ($visitorId && strlen($visitorId) === Tracker::LENGTH_BINARY_ID) {
```
tsteur, line 561: this will ensure when cookies are enabled to then update the visitorId based on the value which was generated in the JS cookie.

while cookies were disabled it should have used some randomly generated visitorId generated by the backend.

We can't really think of any breaking issue because of this.

diosmosis:
build js

PR #16115, opened 2020-06-25
Configure device detector cache through DI

tsteur:
Lets someone customise the cache. Eg different TTL or store the caches in a different directory which can be beneficial if you run many Matomo instances. Say you have 500K instances, and each create 1 MB  tracker cache files, then you have 500GB cache files for device detector that are all the same and you'd want to have this all in opcache which is obviously not easily doable.

### DIFF
--- a/config/global.php
+++ b/config/global.php
@@ -79,6 +79,8 @@ return array(
     'Piwik\Translation\Loader\LoaderInterface' => DI\object('Piwik\Translation\Loader\LoaderCache')
         ->constructor(DI\get('Piwik\Translation\Loader\JsonFileLoader')),
 
+    'DeviceDetector\Cache\Cache' => DI\object('Piwik\DeviceDetector\DeviceDetectorCache')->constructor(86400),
+
     'observers.global' => array(),
 
     /**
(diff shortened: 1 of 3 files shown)


sgiehl (approved):

config/global.php
```diff
@@ -79,6 +79,8 @@
     'Piwik\Translation\Loader\LoaderInterface' => DI\object('Piwik\Translation\Loader\LoaderCache')
         ->constructor(DI\get('Piwik\Translation\Loader\JsonFileLoader')),
 
+    'DeviceDetector\Cache\Cache' => DI\object('Piwik\DeviceDetector\DeviceDetectorCache')->constructor(86400),
```
diosmosis, line 82: Would this also work in the DeviceDetector's plugn DI config? Seems like a more appropriate place to put it to me.

tsteur: I reckon it would cause issues with the API plugin... if DeviceDetection plugin was disabled... Probably there be other places where it would cause issues too but in the end maybe not too important?

diosmosis: Wouldn't that be an issue anyway if the the plugin was disabled? Or I guess the class is loaded no matter what? Something about that seems odd, but I guess it's not that important.

tsteur: Yes the class would still be loaded correctly...

diosmosis: ok.

PR #16122, opened 2020-06-25
Try to avoid ui test failures due to CSS loading failures

sgiehl:
There are a lot builds failing because the CSS seems not to be loaded. In most cases the test logs even don't have any notes, that loading the CSS failed. The reason for that is simple. We always return a HTTP 200 on the error page (as long as the message does not include database keywords). So if merging the CSS throws an exception the error page is served instead (with HTTP Status 200).

To avoid too many test failures when the CSS isn't loaded correctly I've build some code that automatically tries to add another style tag to the page to load the CSS again. [In most cases](https://travis-ci.org/github/matomo-org/matomo/jobs/702070073#L1072) that helps and the tests are running, but sometimes even [that reload fails](https://travis-ci.org/github/matomo-org/matomo/jobs/702070073#L880).

Nevertheless I think that might help to have a more stable build.

### DIFF
--- a/core/ProxyHttp.php
+++ b/core/ProxyHttp.php
@@ -286,6 +286,10 @@ class ProxyHttp
             $data = gzencode($data, 9);
         }
 
+        if (false === $data) {
+            throw new \Exception('compressing file '.$fileToCompress.' failed');
+        }
+
         file_put_contents($compressedFilePath, $data);
     }
 }
--- a/plugins/Dashboard/tests/UI/expected-screenshots/Dashboard_loaded_token_auth.png
+++ b/plugins/Dashboard/tests/UI/expected-screenshots/Dashboard_loaded_token_auth.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:de5ea31d38bc002b7fbec28d3ce78d0f9f749d330a6cd6d6d0e16289a85305a4
-size 734292
+oid sha256:a5fdd3bc28347e7faa84b36ad92068f802bce1836b4ea6ac33efd9b016762a4a
+size 734293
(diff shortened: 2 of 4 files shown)


diosmosis:
can we print out the response body (if any) when these methods fail? I'm curious what the error behind it is.

diosmosis (approved):
LGTM

tsteur:
Be great to merge @sgiehl 

We can otherwise also further look into it towards the end of year if needed and maybe the logged error messages help us find over time the root cause and find a solution

sgiehl:
Was about to merge it when I realized, that even reloading the CSS didn't work in most cases. Logging the reposes and it header lead me to the right direction I guess. The response was always an empty response with status 200. Actually for both requests, even with an additional parameter. So I've tried with some debugging code, and it seems for some reason compressing the css fails and so an empty compressed file is stored. That empty file is automatically served with the next request.

sgiehl:
I've now added an exception to be thrown when compressing a file does not work. That might actually solve the travis problem, as it should prevent storing an empty file and a direct regeneration of that file, as merging the file will be triggered a second time [here](https://builds-artifacts.matomo.org/innocraft/plugin-InvalidateReports/fixtests/79/)

@diosmosis @tsteur  could someone have a quick look at that again and merge afterwards?

core/ProxyHttp.php
```diff
@@ -286,6 +286,10 @@ private static function compressFile($fileToCompress, $compressedFilePath, $comp
             $data = gzencode($data, 9);
         }
 
+        if (false === $data) {
```
sgiehl, line 289: Not sure if we maybe should also check if the response looks valid, like checking if it has at least 50 chars or something like that 🤔

diosmosis: If it's gzipped, that would be kind of hard, right?

PR #16123, opened 2020-06-26
Set cookie domain if configured on ignore cookie

### LINKED ISSUE #15778: Cookie domain config is not applied in ignore cookie preventing opt out to work
https://github.com/matomo-org/tracker-proxy/issues/58

We need to set similar to the third party cookie https://github.com/matomo-org/matomo/blob/3.13.4/core/Tracker/Request.php#L707-L710 also set the cookie domain here: https://github.com/matomo-org/matomo/blob/3.13.4/core/Tracker/IgnoreCookie.php#L61

tsteur:
fix #15778 

tracker cookie name and path was already set, but not the cookie domain.

### DIFF
--- a/core/Tracker/IgnoreCookie.php
+++ b/core/Tracker/IgnoreCookie.php
@@ -58,6 +58,12 @@ class IgnoreCookie
     public static function setIgnoreCookie()
     {
         $ignoreCookie = self::getIgnoreCookie();
+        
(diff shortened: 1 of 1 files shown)


tsteur:
this is actually done by https://github.com/matomo-org/matomo/pull/13301  so closing it

PR #16170, opened 2020-07-02
Clarify developer changelog as it's not clear that we would keep BC for these

mattab:


CHANGELOG.md
```diff
@@ -24,7 +24,7 @@ The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)*
 
 #### Recommendations for Matomo JS tracker
 
-The following changes are recommendations and don't have to be done as we will still support them for many years. However, it could change in the future.
+These are only recommendations (because we will keep backward compatibility forever), but we recommend you update your code for consistency with the Matomo documentations:
```
tsteur, line 27: @mattab that's because we might break BC in 5 years or so see https://github.com/matomo-org/matomo/pull/16085

mattab: I'm not sure that there would be any reason to remove BC in the future, as they are just aliases and wouldn't conflicnt with other products/class names etc. that's why i thought better we don't plan to deprecate them?

mattab: how about `These are only recommendations (because we will keep backward compatibility for many more years), but we do recommend you update your code for consistency and for future proofing your tracking:`

PR #16172, opened 2020-07-02
Fix behaviour for segments using "not equals" or "not contains" on an action dimension

### LINKED ISSUE #13785: Segments using "Event category is not X" or "Page URL does not contain" etc. will list visits with these actions in the Visits log
Hi guys,
I am using Matomo 3.7.0
when I create a segment such as, **Event Category is not 'my-value'** the visitor log report keep showing me visits which include this event.
Is that a known behavior?

sgiehl:
This PR aims to fix the behaviour for segments using "not equals" or "not contains" on an action dimension as described in #13785. 

@tsteur I tried to implement it like you suggested in https://github.com/matomo-org/matomo/issues/13785#issuecomment-652023031, but that actually caused a lot problems with segments that contain a union of segments.

When a segment is initialized the union of segments would by default be splitted up into the union of segments. But those segment union needs to be in a single subquery as it would have an incorrect results otherwise. For now I've put the code to generate that subquery into the segment initialize, which requires the period to be available. Not an optimal solution as a segment should not require a period to be initialized, but guess we would need to refactor the whole usage of the Segment class otherwise, as we need to do that stuff while parsing, but the parsing is also needed for methods that are independent from the period...

At least it seems to work the way it's currently implemented. A segment like `actionUrl!@piwik;browserName!@Chrome` generates a query for the visitor log like:

```sql
/* idSegments = [5] */
SELECT log_visit.*
FROM matomo_log_visit AS log_visit
WHERE (log_visit.idsite IN (?)
       AND log_visit.visit_last_action_time >= ?
       AND log_visit.visit_last_action_time <= ?)
  AND ((log_visit.idvisit NOT IN
          (SELECT /* 2020-06-29,2020-07-01 sites 1 segmenthash 4e6fb538924981bfe0f0bd7e9440df3b */ log_inner.idvisit
           FROM
             (SELECT log_visit.idvisit
              FROM matomo_log_visit AS log_visit
              LEFT JOIN matomo_log_link_visit_action AS log_link_visit_action ON log_link_visit_action.idvisit = log_visit.idvisit
              WHERE (log_visit.visit_last_action_time >= ?
                     AND log_visit.visit_last_action_time <= ?
                     AND log_visit.idsite IN (?))
                AND (((log_link_visit_action.idaction_url IN
                         (SELECT idaction
                          FROM matomo_log_action
                          WHERE (name LIKE CONCAT('%', ?, '%')
                                 AND TYPE = 1)))
                      OR (log_link_visit_action.idaction_url IN
                            (SELECT idaction
                             FROM matomo_log_action
                             WHERE (name LIKE CONCAT('%', ?, '%')
                                    AND TYPE = 3)))
                      OR (log_link_visit_action.idaction_url IN
                            (SELECT idaction
                             FROM matomo_log_action
                             WHERE (name LIKE CONCAT('%', ?, '%')
                                    AND TYPE = 2)))
                      OR (log_link_visit_action.idaction_url IN
                            (SELECT idaction
                             FROM matomo_log_action
                             WHERE (name LIKE CONCAT('%', ?, '%')
                                    AND TYPE = 10)))))
              GROUP BY log_visit.idvisit
              ORDER BY NULL) AS log_inner
           GROUP BY log_inner.idvisit))
       AND (log_visit.config_browser_name IS NULL
            OR log_visit.config_browser_name NOT LIKE ?))
GROUP BY log_visit.idvisit
ORDER BY log_visit.idsite DESC,
         log_visit.visit_last_action_time DESC
LIMIT 0,
      12
```


plugins/Live/Model.php
```diff
@@ -466,7 +466,7 @@ public function makeLogVisitsQueryString($idSite, $startDate, $endDate, $segment
             $filterSortOrder = 'DESC';
         }
 
-        $segment = new Segment($segment, $idSite);
+        $segment = new Segment($segment, $idSite, Period\Factory::build('day', implode(',', array_filter([$startDate,$endDate]))));
```
sgiehl, line 469: Note: That actually does not seem to create the correct period, as the bound parameter for the query when choosing the date range `2020-06-15,2020-06-21` are this:
```
1 = "2020-06-14 22:00:00"
2 = "2020-06-20 21:59:59"
3 = "2020-06-13 22:00:00"
4 = "2020-06-20 21:59:59"
```

So the inner query actually has other date values. Need to check why...

core/Segment.php
```diff
@@ -297,6 +342,23 @@ protected function getCleanedExpression($expression)
         $segment = $this->getSegmentByName($name);
         $sqlName = $segment['sqlSegment'];
 
+        // Build subqueries for segments that are not on log_visit table but use !@ or != as operator
+        // This is required to ensure segments like actionUrl!@value really do not include any visit having an action containing `value`
+        if ($this->doesSegmentNeedSubquery($matchType, $name)) {
```
tsteur, line 347: btw do you reckon it's possible to group all segments together to have only one `NOT IN` subquery? Not too important if too difficult. If not too difficult that could be a good improvement when using several `NOT IN` and `NOT EQUALS` . Might get difficult though with changing operands etc but maybe the ones that are connected with OR.

eg 
image

I wonder if it was easily possible to not have 4 `NOT IN` queries but only 2 `NOT IN` queries. (implementing only one `NOT IN` would probably get very tricky).

Meaning as soon as we detect a `NOT IN` subquery, and more segments are following with `OR`, then these could be maybe grouped together? Maybe check if that's not too tricky and be good to have tests for it.

tsteur (commented):
@sgiehl I reckon this goes in the right direction looking good so far 👍

core/Segment.php
```diff
@@ -297,6 +342,23 @@ protected function getCleanedExpression($expression)
...
+            $operator = $matchType === SegmentExpression::MATCH_DOES_NOT_CONTAIN ? SegmentExpression::MATCH_CONTAINS : SegmentExpression::MATCH_EQUAL;
```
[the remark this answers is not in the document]
sgiehl: will move the operator inversion to a separate method, so it's more clear.

core/Segment.php
```diff
@@ -297,6 +342,23 @@ protected function getCleanedExpression($expression)
...
+            $stringSegment = $name . $operator . $value;
+            $segmentObj = new Segment($stringSegment, $this->idSites);
+
+            $params = new ArchiveProcessor\Parameters(new Site(is_array($this->idSites) ? reset($this->idSites) : $this->idSites), $this->period, $segmentObj);
```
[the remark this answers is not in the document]
sgiehl: For the Parameters class it's not possible to set multiple idSites, but could actually overwrite it for the LogAggregator...

core/Segment.php
```diff
@@ -111,6 +118,16 @@ public function __construct($segmentCondition, $idSites)
...
         $this->originalString = $segmentCondition;
 
+        if ($period instanceof Period) {
+            $this->period = $period;
+        }
+
+        if (empty($this->period)) {
+            $date = Common::getRequestVar('date', false);
+            $periodStr = Common::getRequestVar('period', false);
+            $this->period = Period\Factory::build($periodStr, $date);
```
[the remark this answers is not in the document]
sgiehl: The period is needed as soon as we try to build a subquery. Added that to get it working in general. We actually should inject the period always when creating a Segment instance, so we can remove that again.

core/Segment.php
sgiehl: Well having `action url does not contain` combined with an OR actually cancels each other out, but I got your point. Will have a look if that's easy to solve

sgiehl: While trying to implement that I realized, that we actually can't combine two action segments in an subquery if they are combined with OR. We could combine them if they are using an AND. But guess that would actually require to parse the tree completely, as we can't combine them, if one of them is concatenated with another segment using OR 🤯

core/Segment.php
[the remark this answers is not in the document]
tsteur: Parameter class might return multiple idSites here: https://github.com/matomo-org/matomo/blob/4.x-dev/core/ArchiveProcessor/Parameters.php#L130-L137 

Eg RollUp might return an array of idSites and then we need to search in all those sites

core/Segment.php
tsteur: >  But guess that would actually require to parse the tree completely, as we can't combine them, if one of them is concatenated with another segment using OR 🤯

Getting complicated then... was hoping we could combine at least the `OR` ones. All good for now if not doable

core/Segment.php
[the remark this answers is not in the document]
tsteur: It be great if we could always reject it indeed to avoid any unwanted side effects through parameters.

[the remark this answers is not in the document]
sgiehl: I think we need to go away from using a Period at all. Might be more useful to inject a start and end date into the constructor. We have some live queries for example that only use a start date or that look back a few hours, which can't be represented using a period. Also we might not need to use the LogAggregator at all. Might be possible to directly use the Segment class to generate the query. Will give that a try, maybe that solves some problems...

sgiehl:
@tsteur I've done some adjustments and removed the usage of LogAggregator in that case. Also added some tests. Maybe you could have another look and check if there might be some cases that might not work anymore...

core/Segment.php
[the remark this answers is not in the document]
tsteur: sounds good to use start date and end date directly as parameter 👍

tsteur:
@sgiehl 

it generated for me a query like this one:

```sql
/* idSegments = [61] */

			SELECT
				log_visit.*
			FROM
				matomo_log_visit AS log_visit
			WHERE
				( log_visit.idsite in (?) 
				AND log_visit.visit_last_action_time >= ? )
                AND
                ( ( log_visit.idvisit NOT IN (/* idSegments = [56] */

			SELECT
				log_inner.idvisit
			FROM
				
        (
            
			SELECT
				log_visit.idvisit
			FROM
				matomo_log_visit AS log_visit LEFT JOIN matomo_log_link_visit_action AS log_link_visit_action ON log_link_visit_action.idvisit = log_visit.idvisit
			WHERE
				( log_visit.idsite IN (?) AND log_visit.visit_last_action_time >= ? )
                AND
                ( ( log_link_visit_action.idaction_url IN (SELECT idaction FROM matomo_log_action WHERE ( name LIKE CONCAT('%', ?, '%')  AND type = 1 )) ) )
			GROUP BY
				log_visit.idvisit
			ORDER BY
				NULL
        ) AS log_inner) ) )
			GROUP BY
				log_visit.idvisit
			ORDER BY
				log_visit.idsite DESC, log_visit.visit_last_action_time DESC LIMIT 0, 12
```

I reckon we could remove one of the inner subquery similar to how we do it in the `LogAggregator`:

```php
            $logQueryBuilder = StaticContainer::get('Piwik\DataAccess\LogQueryBuilder');
            $forceGroupByBackup = $logQueryBuilder->getForcedInnerGroupBySubselect();
            $logQueryBuilder->forceInnerGroupBySubselect(LogQueryBuilder::FORCE_INNER_GROUP_BY_NO_SUBSELECT);
            $query = $segmentObj->getSelectQuery($select, $from, $where, $bind); // could also catch this
            $logQueryBuilder->forceInnerGroupBySubselect($forceGroupByBackup);
```

This should get rid of one subquery. 

Otherwise looks good to me so far @sgiehl

sgiehl:
@tsteur updated the code & tests. Should now be ready for a final review

tsteur:
Looks generally good.

Is the startDate/endDate maybe missing from `core/Archive/ArchiveQueryFactory.php::getSegmentFromQueryParam`?

Or can we maybe not return true for `$this->doesSegmentNeedSubquery()` when no startDate/endDate is given? I just applied a segment and it generated this where it would basically result in a full table scan if this was actually executed:

image

sgiehl:
> Is the startDate/endDate maybe missing from core/Archive/ArchiveQueryFactory.php::getSegmentFromQueryParam?

Actually when looking through the code for me it seems the segment query shouldn't be used when loading an Archive. But can set those parameters to be sure

tsteur:
@sgiehl can we still add `$this->doesSegmentNeedSubquery` to return `true` only when a start and end date is set? I wan't to prevent by accident it doing a full table scan which could be basically the death for any DB. Or alternatively we need to require a start and end date to be always set.

sgiehl:
@tsteur I've adjusted the code, so it will no longer use a subquery if no start date is given. If that is the case and a segment query is built, that would require a subquery a warning will be logged.
Maybe we could remove the logging after beta cycle or with a later release, but it might be helpful to check if we have any segment class usages that would require setting a date...

core/Segment.php
```diff
@@ -252,6 +284,51 @@ private function getExpressionsWithUnionsResolved($expressions)
...
+    private function doesSegmentNeedSubquery($operator, $segmentName)
+    {
+        $requiresSubQuery = in_array($operator, [
+                SegmentExpression::MATCH_DOES_NOT_CONTAIN,
+                SegmentExpression::MATCH_NOT_EQUAL
+            ]) && !$this->isVisitSegment($segmentName);
+
+        if ($requiresSubQuery && empty($this->startDate)) {
+            $e = new Exception();
+            Log::warning("Avoiding segment subquery due to missing start date. Please ensure a start date is set when initializing a segment if it's used to build a query. Stacktrace:\n" . $e->getTraceAsString());
```
tsteur, line 314: I guess you mean when start and end date is empty?

created follow up issue so we don't forget to act on it https://github.com/matomo-org/matomo/issues/16200

tsteur: sorry just realise shouldn't have merged this yet as it needs to also check if `endDate` is empty @sgiehl ?

tsteur: created https://github.com/matomo-org/matomo/pull/16201

Issue #13246, opened 2018-08-02
'requireConsent' has not effect on cookie setting


lbroermann:
Good day,

due to our goal to achieve GDPR compliance, we're trying to implement an opt-in procedure using a commercial cookie-plugin called "Borlabs Cookie" with our WordPress-Network.
We embedded Matomo with the following code/attributes:
```html
<!-- Matomo -->
<script type="text/javascript">
  var _paq = _paq || [];
  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
  _paq.push(['requireConsent']);
_paq.push(['trackPageView']);
_paq.push(['trackAllContentImpressions']);
  _paq.push(['enableLinkTracking']);
  (function() {
    var u="//***/";
    _paq.push(['setTrackerUrl', u+'piwik.php']);
    _paq.push(['setSiteId', '1']);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
  })();
</script>
<!-- End Matomo Code -->
```
Tracking works fine, but Matomo creates _pk_id and _pk_ses at every session's start, **also if we don't set** 
`<script type="text/javascript"> 
_paq.push(['setConsentGiven']);
</script>` at the bottom of the page, using our plugin.
'rememberConsentGiven' is no option, since we want to use our WordPress plugin to handle opt-in/opt-out.

The support-pages weren't helpful as this doesn't seem to be the expected behaviour of "requireConsent". Could you please give me some advice on how to solve this problem? 
Thank you!

tobbecmd:
I can confirm this is still happening in version 3.13.5.
_paq.push([‘requireConsent’]); has no effect and _pk_id and _pk_ses cookies are being set without calling setConsentGiven.

lastant:
Maybe there is another way to make sure Matomo tracking code is loaded only after consent is given? For instance, I tried to add the tracking code dynamically using JS, instead of hard-coding it into the HTML; but tracking wouldn't work this way for some reason..

Findus23:
See also https://github.com/matomo-org/matomo/issues/13056 and https://github.com/matomo-org/matomo/issues/15948

Sven74Muc:
Any chance to get this fixed... after 2 years? With this matomo isn't really GDPR ready

Sven74Muc:
> 
> 
[...]

PR #16173, opened 2020-07-03
Don't set any cookies when no consent is given

tsteur:
fix https://github.com/matomo-org/matomo/issues/13246

this PR is based on https://github.com/matomo-org/matomo/pull/16113/files and should be reviewed and merged soon. The change is this PR is a lot smaller.

This applies when someone calls `requireConsent`. In the past we would have still set a visitor cookie even though if no further call to eg `setConsentGiven` or `rememberConsentGiven` followed.

In the past, the visitorId cookie was set when calling `setSiteId`. I'm not sure why this was the case but it shouldn't be needed to set it so early in the configuration process of a new tracker. In fact we're setting the session and attribution cookie already only when calling `getRequest('...')` (eg `getRequest('action_name=foo')`) so it should be fine to set the visitor cookie also only once `getRequest` is being called which this PR now does.

It does change the `nowTs` which is stored in the visitor cookie but that shouldn't really change anything cause mostly a tracking request follows directly anyway. And if in past there were cookies disabled etc then we wouldn't have been able to set this anyway. 

It's adding a new tracker method `areCookiesEnabled` but not adding it to the changelog as it doesn't actually do anything and you can't really use it with _paq.push but will document it once merged.


tsteur:
build js

tsteur:
build js

js/piwik.js
```diff
@@ -7618,10 +7655,16 @@ if (typeof window.Piwik !== 'object') {
              * Call this method once the user has given consent. This will cause all tracking requests from this
              * page view to be sent. Please note that the given consent won't be remembered across page views. If you
              * want to remember consent across page views, call {@link rememberConsentGiven()} instead.
+             *
+             * It will also automatically enable cookies if they were disabled previously.
+             *
+             * @param bool [enableCookies=true] Internal parameter. Defines whether cookies should be enabled or not.
```
tsteur, line 7661: we could generally expose this parameter as a support official parameter. However, then we'd also need to add this to `rememberConsentGiven ()` method and the problem is that we'd then need to persist in yet another cookie whether `rememberConsentGiven ()` should always enable cookies or not. Easier to consider it an internal parameter for now

PR #16181, opened 2020-07-06
don't use create_function() anywhere

Findus23:
`create_function()` is deprecated since PHP 7.2, not needed anymore since 5.2 and removed in PHP 8. So there should be no reason to use it in Matomo 
https://www.php.net/manual/en/function.create-function.php

I did not test the functions in Date.php as honestly I don't understand what is going on there.


MatomoForumNotifications:

https://forum.matomo.org/t/fehlermeldung-create-function-bei-systemcheck/44795/2

PR #16185, opened 2020-07-06
don't specify optional parameters before required parameters

tsteur:
refs https://github.com/matomo-org/matomo/pull/16182

I suppose eventually we'll go over all of them since the IDE should be able to detect all usages.


diosmosis:
Interestingly, since this is an API method, this ends up being a bit of a BC break (see the test failure). Will change the test.

tsteur:
@diosmosis good one. Actually this needs to stay optional. I can switch the parameters

tsteur:
Fixed the tests @diosmosis

PR #16186, opened 2020-07-07
If available, use SERVER_PORT variable as well as SERVER_NAME.

diosmosis:
Otherwise an install on a non-standard port will not pass the trusted hosts check.

### DIFF
--- a/plugins/ScheduledReports/config/tcpdf_config.php
+++ b/plugins/ScheduledReports/config/tcpdf_config.php
@@ -59,14 +59,7 @@ if (!defined('K_TCPDF_EXTERNAL_CONFIG')) {
             } else {
                 $k_path_url = 'http://';
             }
-            $k_path_url .= $_SERVER['SERVER_NAME'];
-            if (strpos($k_path_url, ':') === false
-                && !empty($_SERVER['SERVER_PORT'])
-                && $_SERVER['SERVER_PORT'] != 80
-                && $_SERVER['SERVER_PORT'] != 443
-            ) {
-                $k_path_url .= ':' . $_SERVER['SERVER_PORT'];
-            }
+            $k_path_url .= \Piwik\Url::getHostFromServerNameVar();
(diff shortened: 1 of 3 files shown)


tsteur:
@diosmosis do we need to adjust other places like https://github.com/matomo-org/matomo/pull/16169/files#diff-ab05063f21f4243a15f897801ef47a7cR211 and https://github.com/matomo-org/matomo/pull/16169/files#diff-719a15b635df457515639ff08ee921b3R826 too?

tsteur:
@diosmosis some tests seem to be failing now: eg https://travis-ci.org/github/matomo-org/matomo/jobs/705628957#L669

Otherwise all good to merge

plugins/ScheduledReports/config/tcpdf_config.php
```diff
@@ -60,6 +60,13 @@
                 $k_path_url = 'http://';
             }
             $k_path_url .= $_SERVER['SERVER_NAME'];
+            if (strpos($k_path_url, ':') === false
+                && !empty($_SERVER['SERVER_PORT'])
+                && $_SERVER['SERVER_PORT'] != 80
+                && $_SERVER['SERVER_PORT'] != 443
+            ) {
+                $k_path_url .= ':' . $_SERVER['SERVER_PORT'];
+            }
```
sgiehl, line 69: Any reason for not directly using
```php
$k_path_url .= Url::getHostFromServerNameVar();
```

diosmosis: I thought this was part of a library so I didn't want it to depend on Matomo, but looking at the path above, I guess it's actually Matomo code?

sgiehl: yes, it's our configuration for tcpdf. Even though I think some parts of that config could maybe meanwhile be removed.

diosmosis:
Updated and fixed the tests.

Issue #16055, opened 2020-06-11
Do not load JS tracking code deferred


tsteur:
Currently we set `async=true;defer=true;`. However, most browsers support async so defer should be no longer set. This is recommended behavior and ensures a browser won't prefer `defer` over `async` (which be bad for performance).

Should be updated in generated tracking code and in docs. Only `async=true` should be set.

Findus23:
it seems like GA also only sets async, but not defer (https://github.com/matomo-org/matomo/issues/16189)

Issue #16189, opened 2020-07-07
assign _paq to window in tracking code


Findus23:
When putting the default Matomo tracking code into a function, tracking stops working as matomo.js expects `window._paq` to exist and the variable is just local inside the function.

For comparison the [default](https://developers.google.com/analytics/devguides/collection/analyticsjs) Google Analytics tracking code:
```javascript
(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

ga('create', 'UA-XXXXX-Y', 'auto');
ga('send', 'pageview');
```
slightly more readable form:
```javascript
window['GoogleAnalyticsObject'] = 'ga';
window['ga'] = window['ga'] || function () {
    (window['ga'].q = window['ga'].q || []).push(arguments)
}, window['ga'].l = 1 * new Date();
a = document.createElement('script'),
    m = document.getElementsByTagName('script')[0];
a.async = 1;
a.src = 'https://www.google-analytics.com/analytics.js';
m.parentNode.insertBefore(a, m)

ga('create', 'UA-XXXXX-Y', 'auto');
ga('send', 'pageview');
```
completly unminified form:
```javascript
window.GoogleAnalyticsObject = 'ga';
window.ga = window.ga || function () {
    (window.ga.q = window.ga.q || []).push(arguments)
};
now = new Date()
window.ga.l = now.getTime();
scriptTag = document.createElement('script');
firstScript = document.getElementsByTagName('script')[0];
scriptTag.async = true;
scriptTag.src = 'https://www.google-analytics.com/analytics.js';
firstScript.parentNode.insertBefore(scriptTag, firstScript)

ga('create', 'UA-XXXXX-Y', 'auto');
ga('send', 'pageview');
```

tsteur:
@Findus23 also created https://github.com/matomo-org/developer-documentation/pull/361 and https://github.com/matomo-org/tag-manager/pull/258

PR #16190, opened 2020-07-07
Generated tracking code should init window._paq and only use async

tsteur:
refs https://github.com/matomo-org/matomo/issues/16189
fix https://github.com/matomo-org/matomo/issues/16055

@Findus23 can you have a look? 

I'm still using `var _paq` to keep the tracking code short. As otherwise would need to prefix every _paq.push code but could also instead prefix every variable with `_paq`. Once merged will need to update the docs on the developer site to use same style. Tested it in IE6 and it worked for me.

refs https://github.com/matomo-org/matomo/pull/13674

### DIFF
--- a/plugins/Morpheus/templates/javascriptCode.twig
+++ b/plugins/Morpheus/templates/javascriptCode.twig
@@ -1,6 +1,6 @@
 <!-- Matomo -->
 <script type="text/javascript">
-  var _paq = window._paq || [];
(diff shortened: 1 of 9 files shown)


Findus23 (approved):
Looks good to me, but I would mention it in the changelog as it is kind of a breaking change (or at least makes people wonder if their old tracking code is "broken").

PR #16195, opened 2020-07-08
Backport support for matomo_ignore class

### LINKED ISSUE #16193: JS tracker should support CSS class matomo_ignore as alias of piwik_ignore
From email feedback:

>  Hi, I'm noticing some strange issues with matomo.js. For example. the js tracking guide indicates that using the class 'matomo_ignore' should result in the matomo ignoring link downloads, however, i am finding that does't work, while the class 'piwik_ignore' does cause the element to be ignored. Is the documentation wrong? or is it possible that the cdn is serving an older version of the tracking code?

Our guide mentions `matomo_ignore` here: https://developer.matomo.org/guides/tracking-javascript-guide

tsteur:
fix #16193


tsteur:
build js

PR #16197, opened 2020-07-09
Make it possible to detect other clients than Browsers

sgiehl:
This PR prepares Matomo for better reports on used clients. Currently Matomo shows all clients that are not detected as a browser as `Unknown`. With this changes the column that currently stores the browser short code will be increased in size, to be able to store full client names. Additionally there will be a new column to store the client type (represented with an int id).

Besides changing the database I've also decided to already store the full name for clients (other than browsers). That way those other clients will already appear in the "Browsers" reports. That might not be fully correct, but imho better than report them as unknown. The client type will also be set already, but should be invisible everywhere (except from custom reports maybe)

Note: It's currently not possible to segment for those other clients, as the browser segment still only accepts the short codes. That's something we can add later, when we possibly also add new reports for each client type or so.

refs #5413


plugins/DevicesDetection/Columns/ClientType.php
```diff
@@ -0,0 +1,69 @@
...
+
+use Piwik\Metrics\Formatter;
+use Piwik\Tracker\Request;
+use Piwik\Tracker\Visitor;
+use Piwik\Tracker\Action;
+
+class ClientType extends Base
+{
+    protected $columnName = 'config_client_type';
+    protected $columnType = 'TINYINT( 10 ) NULL DEFAULT NULL';
```
tsteur, line 19: @sgiehl I suppose `TINYINT (10)` can't quite work since it stores max 255 different values?

plugins/DevicesDetection/Columns/BrowserName.php
```diff
@@ -21,7 +21,7 @@
 class BrowserName extends Base
 {
     protected $columnName = 'config_browser_name';
-    protected $columnType = 'VARCHAR(10) NULL';
+    protected $columnType = 'VARCHAR(40) NULL';
```
tsteur, line 24: Just to better understand can you mention what values be like? Was thinking ideally this was referencing `log_action` so we don't always store the full name but that would mean migrating all existing values which isn't really doable / realistic. I guess we don't want to use any mappings having of `short_name` to `long_name` considering we would have 10 characters available that would give us quite a bit of flexibility already?

plugins/DevicesDetection/Columns/ClientType.php
sgiehl: sure. Actually meant to use `TINYINT( 1 )`. Will change that

plugins/DevicesDetection/Columns/BrowserName.php
sgiehl: Device detector is able to detect various client names, and there are also some dynamic detection for e.g. mobile apps. You can see a full list here: https://github.com/matomo-org/device-detector/blob/f33d196a7790689f721db64766f5ec3eec470284/README.md#list-of-detected-libraries

I would love to go away from the short codes at all at some point, and having it in `log_action` would be awesome. But as you mentioned a migration isn't easily doable.
We could alternatively also add another id column and start storing the client names in log_action separately. We could than maybe switch on using those data for creating the reports in a later release when we already have some data logged, so we maybe don't need too much logic to merge it with the old reports. What do you think?

tsteur: @sgiehl I'd say really depends how easy it is to do... I suppose a segment would then check using `OR` either in the column or in the log_action? Let's maybe for now just keep using the VARCHAR then

plugins/DevicesDetection/Columns/ClientType.php
tsteur: 👍

plugins/DevicesDetection/Columns/BrowserName.php
sgiehl: @tsteur adding and filling that new column is very easy and quickly done. That way we wouldn't also change any of the existing reports & segments. New reports and changes to segments could then be done any time later

plugins/DevicesDetection/Columns/ClientType.php
sgiehl, line 19:
```suggestion
    protected $columnType = 'TINYINT( 1 ) NULL DEFAULT NULL';
```

plugins/DevicesDetection/Columns/BrowserName.php
tsteur: Up to you @sgiehl I reckon either is fine

sgiehl: Will change it then. Makes more sense to have a new id column. Will check how easy we can change it so only the new column is used for fresh Matomo installs...

sgiehl: @tsteur just had another look at that, and realized, that the client actually is no action dimension or shouldn't be one. So all the action id magic wouldn't work out of the box. Also we would need to reserve a action type for it, even though it isn't an action at all. So it might be a bit more effort than I thought... But guess it would still be the better solution to not store the full names in visit table 🤔

tsteur: Let's just store it in full text for now. Seems not worth the effort. If the tests pass feel free to merge @sgiehl

PR #16201, opened 2020-07-10
Follow up change only triggering error when no start and end date is set

tsteur:
refs https://github.com/matomo-org/matomo/pull/16172

fyi @sgiehl does this make sense?


sgiehl:
Then we need to set an end date for this live query: https://github.com/matomo-org/matomo/blob/d72b8bf705dabcdbd4d00876b2a3be9f58056e61/plugins/Live/Model.php#L367

tsteur:
@sgiehl I was meaning it's fine if only end date or only start date is set. but not when both are empty, as this would be causing a full table scan.

sgiehl:
Ok. Then it should be fine that way

sgiehl (approved):

PR #16203, opened 2020-07-13
skip id segments option

mikkeschiren:


core/CronArchive.php
```diff
@@ -626,6 +633,27 @@ public function setSegmentsToForceFromSegmentIds($idSegments)
...
+     * @param int[] $idSegments
+     */
+    public function setSegmentsToSkipFromSegmentIds($idSegments)
+    {
+        /** @var SegmentEditorModel $segmentEditorModel */
+        $segmentEditorModel = StaticContainer::get('Piwik\Plugins\SegmentEditor\Model');
+        $segments = $segmentEditorModel->getAllSegmentsAndIgnoreVisibility();
+
+        $segments = array_filter($segments, function ($segment) use ($idSegments) {
+            return (!in_array($segment['idsegment'], $idSegments));
```
tsteur, line 647: @mikkeschiren looks good so far. The only thing I'm wondering if we could also filter by idSite? 

eg 
```php
in_array($segment['idsegment'], $idSegments) && (empty($segment['enable_only_idsite']) || $idSite == $segment['enable_only_idsite'])
```

And then we could throw an exception if an `idSegments`  were specified, but none matches any. This way it wouldn't by accident set no segments to force, and the user would be aware a wrong ID was set.

sgiehl:
@mikkeschiren are you able to apply the changes suggested by @tsteur ?

mikkeschiren:
Yeah, sorry, been occupied will look into this and change.

tsteur:
@mikkeschiren sorry I have to close this issue. It cannot be merged as the logic changed quite a bit in Matomo 4. It will now need to be implemented in a so called `ArchiveFilter` similar to https://github.com/matomo-org/matomo/blob/4.x-dev/core/CronArchive/ArchiveFilter.php#L64-L68

Here is the method where it sets the segments https://github.com/matomo-org/matomo/blob/4.x-dev/core/CronArchive/ArchiveFilter.php#L124 and this is the code in the command: https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/CoreConsole/Commands/CoreArchiver.php#L53-L56

It might be fairly easy to adopt the current code into this new way. 

Generally I would maybe not set `$this->segmentsToForce` but another property eg `ArchiveFilter::$segmentsToSkip` and a method `ArchiveFilter::setSegmentsToSkip` just to make sure there won't be any side effects as skip should maybe not force all other segments except for the skipped one but really rather only skip any segment that matches the skipped ones.

PR #16205, opened 2020-07-13
Rename composer package to matomo/matomo

sgiehl:
All packages should still be available as piwik/piwik, as this repo should still be used for updating that package.
After this PR was merged, we need to submit the package with the new name on packagist, so it will be available.

refs #12519

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "piwik/piwik",
+    "name": "matomo/matomo",
     "type": "application",
     "description": "the leading free/libre analytics platform",
     "keywords": ["piwik","matomo","web","analytics"],


tsteur:
Looks good @sgiehl will you submit this afterwards?

sgiehl:
@tsteur can submit it. Wondering if we maybe should wait for a first 4.0 beta to publish the new package

tsteur:
👍 moved it to RC

tsteur:
@sgiehl can you merge this PR on Monday or so and then publish the package on packagist?

sgiehl:
@tsteur done. Shall we mark piwik/piwik as abandoned and suggest the new name for usage? The old name would still work and provide all releases, but composer would suggest to switch to the new name...

tsteur:
sounds great @sgiehl

PR #16212, opened 2020-07-15
Fix language selector in mobile menu is not visible

tsteur:
When being anonymous, and opening the top menu in the mobile view then nothing is readable because it's white on light grey


sgiehl:
Should we maybe add a UI test for the mobile menu? 🤔

tsteur:
It's fine it only affected it when someone had public viewing of data `anonymous` enabled.

PR #16215, opened 2020-07-15
Add one-click deploy button

jakemalachowski:
Hello. I work at [Render](https://render.com/). We recently set up a one-click deployment button and guide that allows your users to get up and running with a Matomo server and MySQL database in just a couple minutes. We are looking to make it as seamless as possible to self-host your software. I was hoping you would be willing to add this as an option in your setup guide.  Also, would be open to mentioning us as an option on your [on-pemise](https://matomo.org/matomo-on-premise/) page?   

I'm happy to address and questions or concerns!

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,14 @@ See https://matomo.org/docs/installation/.
 
 (When using Matomo for development you need to [install Matomo from the Git repository](https://matomo.org/faq/how-to-install/faq_18271/)).
 
+## One Click Deployment
+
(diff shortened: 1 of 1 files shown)


tsteur:
Cheers @jakemalachowski have added this to https://matomo.org/faq/how-to-install/faq_108/ which is also linked to from our install guide. We currently don't link to any specific provider from our github readme unfortunately.

PR #16217, opened 2020-07-17
Add new dimension and segment "Profilable" (and new column log_visit.profilable)

### LINKED ISSUE #16192: Add dimension and segment whether cookies were used or not
With Matomo we've always had the ability to disable cookies or use cookies. Now we have added the ability to also enable cookies after they were enabled etc. This lets users track visitors with cookies disabled, but enable cookies when/if they give consent to use cookies.

When cookies are enabled, [some reports become more accurate](https://matomo.org/faq/general/faq_156/). 

When the data is mixed from visitors with and without cookies in the same site, then it's hard to know which referrer data to trust for example.

That's why you'd want to segment the audience by whether cookies were used or not used.

* We need to add a new dimension and segment to the log_visit table
* We need to send a new tracking parameter whether cookies were used or not (based on `configCookiesDisabled` variable). We need to update the log_visit column onNewVisit and onExisitingVisit since it might change during a visit
* We don't need to add a report for now
* We can't reuse the existing cookie flag we have in the visits table because it tracks whether cookies are generally available but not whether they were actually used or not used

Not sure what naming to use. Ideally, the wording could also be applied for iOS/Android SDK etc. Eg they might have the same concept, but don't use cookies for storing a visitorId.

tsteur:
fix https://github.com/matomo-org/matomo/issues/16192

After lots of thinking came up with the dimension name "profilable" to have it working as well for iOS, Android etc which doesn't rely on cookies etc. I guess that's really the essence of it. Which users can be tracked over several visits vs which can't. Happy to change the name if anything is better.

After some thinking also didn't add a new tracking parameter for now. Generally, it should work out of the box using the visitorId. If cookies are disabled, then no visitorId is sent. Technically we can't rely on cookies alone cause if you are disabling cookies but use userId feature, then the user can be still profiled over several visits. I reckon for now it's not needed to introduce a new tracking parameter. Eventually could have like a parameter `_pr` or so if needed maybe to force a certain value? Not sure if there are cases where it wouldn't be accurate based on visitorId?


core/Updates/4.0.0-b1.php
```diff
@@ -153,6 +155,10 @@ public function getMigrations(Updater $updater)
             $migrations[]     = $this->migration->db->sql("UPDATE $visitActionTable SET search_count = custom_var_v5 WHERE custom_var_k5 = '_pk_scount'");
         }
 
+        $logVisit = Common::prefixTable('log_visit');
+        // until 3.14.0 config_cookie was always 0 if cookies were disabled so for most visits this will be fairly accurate.
+        $migrations[] = $this->migration->db->sql("UPDATE $logVisit SET profilable = config_cookie WHERE profilable is null");
```
tsteur, line 160: It be a problem if someone was using Matomo 3.14.0 for a long time, then upgraded to Matomo 4. Then this may not be as accurate. Could also remove this update script.

mattab: it might be fine to leave it, but it would be important to document it in the developer changelor or in the changelog somewhere. btw would it be maybe possible to update only up to the time that they had upgraded to 3.14.0?

mattab (commented):
Profilable as a name sounds good to me :+1:

How would we document this feature, maybe in a FAQ "What is the Profilable dimension?" to start with something basic? (we could also later link to the faq from other relevant user guides and other relevant faqs)

core/Updates/4.0.0-b1.php
tsteur: AFAIK we don't know when they upgraded

tsteur:
@mattab we'll document it later once merged.

sgiehl (changes requested):
Some tests need to be updated/fixed. Otherwise looks good 👍

PR #16221, opened 2020-07-20
Fix start/end time for given period it can skip this archive

### LINKED ISSUE #16220: Matomo 4: Browser archiving issue for today archive
Custom ini settings:

```ini
[Debug]
always_archive_data_day = 1

[General]
```

First I tracked a new page view 

```
matomo.php?rec=1&idsite=1&action_name=foo&send_image=1
```

Timezone of the site is `America/Los_Angeles`

The `log_visit.visit_last_action_time` for this visit is `2020-07-20 01:41:44`. 

I then view the reports like this

```
index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=day&date=2020-07-19&segment=&category=General_Actions&subcategory=General_Pages
```

The date is `2020-07-19`. However, nothing is being archived since `Loader::canSkipThisArchive` returns `false` because `hasSiteVisitsBetweenTimeframe()` returns false:

image

The `to` and `from` date is correct. However, in the UI I can't select the 20th July as this day as not started yet. My local time is 1:40PM July 20th NZST.

image

When using Matomo 3 everything seems to work as expected. Once I disable the `canSkipThisArchive` check, it starts the archiving. And interestingly it uses a different time:

image
image

and it is showing data for this report 
image

It seems bit random that Matomo would use two different times? Somewhere the time is off?

The time that `canSkipThisArchive` used seems correct to me? 
image

The `LogAggregator` class uses this method:
image

image

and `canSkipThisArchive` uses a different way as it uses this logic

image

Something seems off. Then realised `skipThisArchive` was creating the wrong time as it was generated the time for `UTC 0am to 23:59` in the different timezone. When it should have created the UTC time for `US/LA 0am to 23:59`:

image

To make it short, skipThisArchive should use same logic as logaggregator see PR https://github.com/matomo-org/matomo/pull/16221

tsteur:
fix https://github.com/matomo-org/matomo/issues/16220

fyi @diosmosis be good if you could have a look or maybe you remember if that caused any issue


diosmosis:
@tsteur can you modify `Period::getBoundsInTimezone()`? there are unit tests for it.

tsteur:
@diosmosis sure pushing this right now

diosmosis:
the unit tests failed, are they incorrect?

tsteur:
Yes I think they aren't correct 
image

diosmosis (approved):

PR #16224, opened 2020-07-20
Document new path to piwik / matomo tracker file

tsteur:


sgiehl:
@tsteur  the path actually already changed for Matomo 3.14. See #16227

tsteur:
@sgiehl yes but we'll revert that again

tsteur:
I see I thought we maybe also use the piwik component again in https://github.com/matomo-org/matomo/pull/16227 ? If not then can also simply close it. Might not be needed then.

PR #16226, opened 2020-07-21
Backport userId to overwrite visitorId feature

tsteur:
refs https://github.com/matomo-org/matomo/issues/15593 

based on https://github.com/matomo-org/matomo/pull/16124

not adding any extra tests as default behaviour doesn't change and we have added tests in Matomo 4 already for both config values (enabled / disabled)

### DIFF
--- a/core/Tracker/Visit.php
+++ b/core/Tracker/Visit.php
@@ -562,7 +562,18 @@ class Visit implements VisitInterface
             // Might update the idvisitor when it was forced or overwritten for this visit
             $valuesToUpdate['idvisitor'] = $this->request->getVisitorId(); 
         }
-        
+
+        if (TrackerConfig::getConfigValue('enable_userid_overwrites_visitorid')) {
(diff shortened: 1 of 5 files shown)


core/Tracker/Visit.php
```diff
@@ -562,7 +562,18 @@ private function setIdVisitorForExistingVisit($valuesToUpdate)
             // Might update the idvisitor when it was forced or overwritten for this visit
             $valuesToUpdate['idvisitor'] = $this->request->getVisitorId(); 
         }
-        
+
+        if (TrackerConfig::getConfigValue('enable_userid_overwrites_visitorid')) {
```
tsteur, line 566: technically, this `if` might be no longer needed since we added above `if` in Matomo 4. However, the `if` above does not yet update `$this->visitProperties->setProperty('idvisitor', $binIdVisitor);` and I'm not sure why this is done and if it's needed so I added the entire `if` block

PR #16229, opened 2020-07-21
Rename config setting for maximum rows in custom dimension reports

sgiehl:
shall we mention that change in the changelog?

refs #11524


diosmosis:
Makes sense to me to mention in the changelog since this change will require a change on users that use the old setting.

tsteur:
@sgiehl can we keep the `*custom_variables` configs but also add new `*_custom_dimension` settings? and run an update script that copies over existing values that may be configured to the new setting?  

Then we probably don't need https://github.com/matomo-org/plugin-CustomVariables/pull/5 and things are more clear

mattab:
btw, it would also need to be documented in https://matomo.org/faq/how-to/faq_54/

sgiehl:
@tsteur does it make sense to keep the `*custom_variables` settings in core, if we move the plugin out?
Added an update script that copies over the settings if they were configured locally before.

tsteur:
I don't have a big preference, can move it to the plugin. Would then just need to make sure in the plugin that the config values are set when they update as otherwise suddenly the default values would be missing. So need to try and set it maybe on Matomo 4 core update, but also in the plugin update additionally maybe?

sgiehl:
That's reason I've added a fallback to the `*dimensions` config values in the plugin. Those are defined by core after the update and will always be present if someone did not define the `*variables` config values before manually.

tsteur:
👍  can you fix the tests and then merge and close the ticket?

PR #16237, opened 2020-07-23
Set referrer policy on default img tracking code

### LINKED ISSUE #15845: Tracking of visitors without JavaScript (image beacon) does not work with strict referrer-policy
Hi,

We recently installed Matomo 3.13.4 and included the tracking code in the `<head>` section of [our website](https://www.ciphermail.com/). We generated the tracking code with the option 'Track users with JavaScript disabled' checked.

While tracking visitors with JavaScript enabled works perfectly, it looks like visitors without JavaScript are not tracked at all.

**How to reproduce:**

1. On Firefox, make sure that DNT header is not sent and Matomo is not blocked in any way.
2. Disable JavaScript by setting the javascript:enabled config flag to False.
3. Click around on the website. Verify that the image beacon gets loaded correctly.
4. Try to find the visit in Matomo.

**Expected result:**

The webmaster should be able to find the visit in the Matomo web interface, or at least in the database.

**Actual result:**

The visit can neither be found in the Matomo web interface nor in the database.

tsteur:
fix https://github.com/matomo-org/matomo/issues/15845 I can reproduce this when using this script

```php
<?php header('referrer-policy: same-origin');?>
<html>
<head>

</head><body>
<img src="//YOURDOMAIN.com/matomo.php?idsite=1&rec=1" referrerpolicy="no-referrer-when-downgrade" style="border:0;" alt="" />
</body>
</html>
```

This scripts need to be placed on a different domain than the Matomo domain. Without the referrerpolicy, the tracking quest would not include any information and track a request without any page URL. With the attribute, it tracks the request and includes the referrer.

The downgrade part shouldn't be any issue because should a website be HTTPS, and Matomo HTTP, then the request would likely not be executed anyway.  For more information see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy 

image

### DIFF
--- a/plugins/SitesManager/API.php
+++ b/plugins/SitesManager/API.php
@@ -182,7 +182,7 @@ class API extends \Piwik\Plugin\API
 
         $url = (ProxyHttp::isHttps() ? "https://" : "http://") . rtrim($piwikUrl, '/') . '/'.$matomoPhp.'?' . Url::getQueryStringFromParameters($urlParams);
         $html = "<!-- Matomo Image Tracker-->
-<img src=\"" . htmlspecialchars($url, ENT_COMPAT, 'UTF-8') . "\" style=\"border:0\" alt=\"\" />
+<img referrerpolicy=\"no-referrer-when-downgrade\" src=\"" . htmlspecialchars($url, ENT_COMPAT, 'UTF-8') . "\" style=\"border:0\" alt=\"\" />
 <!-- End Matomo -->";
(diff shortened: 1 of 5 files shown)


imrejonk:
@tsteur Looks good, great to see that you picked this up! Hopefully it will prevent webmasters some headaches :)

tsteur:
@imrejonk thanks for mentioning it and providing the solution 👍  that made it easy

PR #16240, opened 2020-07-24
Improve visitor profile styles, so nothing collapses when zooming in/out

### LINKED ISSUE #16238: Segmented Visitor profile: Layout breaks when zooming out
Issue: Layout of the segmented visits log breaks when zooming out in the browser when reaching any percentage less than 100%, visits log is getting listed on the left instead of right.

Steps to reproduce:

1. Click "View visitor profile" on any visit under the 'Visits Log"
2. Zoom out the browser less then 100%

Can be reproduced in [demo](https://demo.matomo.org/) website

sgiehl:
fixes #16238


diosmosis:
Not sure why, but the UI tests are failing w/ extra space on some of the profile (same issue is not present locally).

tsteur:
FYI If it's not easy to fix the screenshot test be good to simply just close the PR for now and we work on this when it's a priority.

sgiehl:
updated the styles a bit more and fixed the UI tests

PR #16244, opened 2020-07-27
Remove update script for new profilable dimension

tsteur:
refs https://github.com/matomo-org/matomo/pull/16217/files#r456216141
refs https://github.com/matomo-org/matomo/issues/16192

Thought it's better to remove this update as the value might be set wrongly in case someone updates from 3.14.0 to 4.0.0. Wanted to remove this in the PR directly but forgot to do it. Think it's better to have the dimension unset before setting it potentially wrongly.  see https://github.com/matomo-org/matomo/pull/16217/files#r456216141

Was also thinking of changing it to `"UPDATE $logVisit SET profilable = config_cookie WHERE profilable is null and visit_last_action_time < 2020-07-08"` because nobody could have used 3.14.0 before first beta was released on 8th July but it would mean some visits have this dimension set while others don't and then it maybe becomes inconsistent. Could also add the date though.

### DIFF
--- a/core/Updates/4.0.0-b1.php
+++ b/core/Updates/4.0.0-b1.php
@@ -155,10 +155,6 @@ class Updates_4_0_0_b1 extends PiwikUpdates
             $migrations[]     = $this->migration->db->sql("UPDATE $visitActionTable SET search_count = custom_var_v5 WHERE custom_var_k5 = '_pk_scount'");
         }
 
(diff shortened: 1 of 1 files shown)


sgiehl:
Guess completely removing it is better in this case. Using a date is inaccurate as well, as someone might wait a long time till updating to 4.0

PR #16245, opened 2020-07-27
Use new tracker mehtod to set a custom dimension

sgiehl:
requires https://github.com/matomo-org/matomo-php-tracker/pull/76


sgiehl:
Some tests and local tracker implementation required some updates, as due to https://github.com/matomo-org/matomo-php-tracker/pull/72 the generated tracker urls do no longer contain the token auth...

PR #16253, opened 2020-07-29
Remove no longer needed/used heartbeat code

tsteur:
Noticed `heartBeatUp` was actually no longer called since we refactored the heartbeat timer so this code should be possible to be removed


tsteur:
build js

PR #16263, opened 2020-08-04
Only allow widgetize and token usage for view users

tsteur:
Not really needed this change I guess but might be better to allow authentication using tokens only for users when the used token has only view access. This may be helpful if otherwise a user tries to embed some screens using token_auth.

Should there be any regressions that we can't fix we could undo it again.


diosmosis:
build is failing, looks like this has some other side effects

tsteur:
@diosmosis tests should pass now

core/FrontController.php
```diff
@@ -408,9 +408,18 @@ public function init()
 
         // Force the auth to use the token_auth if specified, so that embed dashboard
         // and all other non widgetized controller methods works fine
-        if (Common::getRequestVar('token_auth', false, 'string') !== false) {
+        if (Common::getRequestVar('token_auth', '', 'string') !== ''
+            && Request::shouldReloadAuthUsingTokenAuth(null)) {
             Request::reloadAuthUsingTokenAuth();
+            if (($module !== 'API' || ($action && $action !== 'index')) && Piwik::isUserHasSomeWriteAccess() && !Common::isPhpCliMode()) {
```
diosmosis, line 414: would it be a good idea to move `($module !== 'API' || ($action && $action !== 'index'))` and ` && !Common::isPhpCliMode()` to another method? something like `isTokenAuthLimitedToWriteAccess()`

diosmosis (approved):
left one comment, otherwise looks good to me

core/FrontController.php
tsteur: 👍  changed it. will merge if tests pass.

attritionorg:
@tsteur Can you summarize what security impact this has, or what privileges can be gained by exploiting this?

tsteur:
@attritionorg I assume you refer to when someone uses the token of a write, admin or super user? If the widgetized URL is shared with other users and that URL includes the token, then people who that token is shared with would gain the same privileges over the API. As a best practice we're now enforcing to use only tokens for view users. So if someone was to embed the widget into an internal wiki or so for all employees then they only have "view" access. Not sure if this roughly answers the question?

attritionorg:
I think it does, thank you @tsteur. The initial language in this PR made it sound like this could be an exploitable security issue so was looking for clarity.

mattab:
fyi added mention in the guide: https://matomo.org/docs/embed-matomo-reports/ 

> Note: for security reasons, embedding the reports will only work when you use a token of the "View" permission  (if you use a "Write" or "Admin" permission token an error message will be displayed instead.)

PR #16264, opened 2020-08-06
When embedding reports (widgets) into a site, it is no longer possible to use authentication tokens of users with at least write access 

tsteur:
Added missing changelog entry and translation key.

It's based on https://github.com/matomo-org/matomo/pull/16263 which I don't get to pass right now so thought to create this simpler change extra while figuring out how to make the other one work

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,6 +127,7 @@ These are only recommendations (because we will keep backward compatibility for 
 
 ### Other Breaking changes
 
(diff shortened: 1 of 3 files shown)


tsteur:
@diosmosis splitted the other PR https://github.com/matomo-org/matomo/pull/16263 so it's easier to merge this separate change (which is required change compared to the other which isn't 100% required)

PR #16282, opened 2020-08-10
JS Tracker: Fix default value of heart beat timer

sgiehl:
Documentation says the default should be 15, but actually if no value is given actually 5 would be always used.

Could be backported to 3.x-dev as well maybe


sgiehl:
build js

PR #16284, opened 2020-08-11
Reorganise Matomo 4 updates for better performance

tsteur:
For better performance we should combine as many queries into one ALTER table statement where possible. Technically there are also 2 drop user columns that could be combined but this table will be fast anyway and executing them individually may be better.

technically could also add all `add columns and drop columns` for each table together but not sure how much this impacts performance but would potentially make it quite a bit more difficult to implement maybe.

Any other thoughts?

### DIFF
--- a/core/Updater/Migration/Db/DropColumns.php
+++ b/core/Updater/Migration/Db/DropColumns.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * Matomo - free/libre analytics platform
+ *
+ * @link https://matomo.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ */
+namespace Piwik\Updater\Migration\Db;
+
+use Piwik\Db;
+use Piwik\DbHelper;
+
+/**
+ * @see Factory::dropColumn()
+ * @ignore
+ */
+class DropColumns extends Sql
+{
+    public function __construct($tableName, $columnNames)
+    {
+        $table = DbHelper::getTableColumns($tableName);
+
+        // we need to remove all not existing columns. Otherwise if only one of the columns doesn't exist, all of
+        // the columns wouldn't be removed
+        $columnNames = array_filter($columnNames, function ($columnName) use ($table) {
+            return isset($table[$columnName]);
+        });
+
+        if (empty($columnNames)) {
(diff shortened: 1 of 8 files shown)


core/Updater/Migration/Db/DropColumns.php
```diff
@@ -0,0 +1,43 @@
...
+        $table = DbHelper::getTableColumns($tableName);
+
+        // we need to remove all not existing columns. Otherwise if only one of the columns doesn't exist, all of
+        // the columns wouldn't be removed
+        $columnNames = array_filter($columnNames, function ($columnName) use ($table) {
+            return isset($table[$columnName]);
+        });
+
+        if (empty($columnNames)) {
+            parent::__construct('', static::ERROR_CODE_COLUMN_NOT_EXISTS);
```
tsteur, line 30: could alternatively use ```alter table `table_name`;``` but I reckon it's less confusing when no SQL query is shown.

core/Updates/4.0.0-b1.php
```diff
@@ -151,8 +149,7 @@ public function getMigrations(Updater $updater)
 
         if (Manager::getInstance()->isPluginInstalled('CustomVariables')) {
             $visitActionTable = Common::prefixTable('log_link_visit_action');
-            $migrations[]     = $this->migration->db->sql("UPDATE $visitActionTable SET search_cat = custom_var_v4 WHERE custom_var_k4 = '_pk_scat'");
-            $migrations[]     = $this->migration->db->sql("UPDATE $visitActionTable SET search_count = custom_var_v5 WHERE custom_var_k5 = '_pk_scount'");
+            $migrations[]     = $this->migration->db->sql("UPDATE $visitActionTable SET search_cat = if(custom_var_k4 = '_pk_scat', custom_var_v4, search_cat), search_count = if(custom_var_v5 = '_pk_scount', custom_var_v5, search_count) WHERE custom_var_k4 = '_pk_scat' or custom_var_k5 = '_pk_scount'");
```
tsteur, line 152: only one full table scan...

core/Updater/Migration/Db/DropColumns.php
sgiehl: makes sense to show no sql in this case.

sgiehl (approved):

PR #16287, opened 2020-08-11
By default select site creation date if site was created today or if select date is before site creation date

### LINKED ISSUE #13754: When looking at Yesterday's data, and there is data for today, notify users to try to look at Today
A common cause of confusion for new users (who have just setup Matomo, or created a new website), is that by default Matomo shows Yesterday's data. And people often don't realise they should look at Today's to see their data. So it would be valuable to guide new users and help them understand they're looking at the wrong date (small improvement to onboarding).

### Suggested change 

* When a user is looking at "Yesterday's" data, or "Last week" or "Last month" 
* And there is "No data" 
* And there is data for "Today" (or "This week" or "This month" or "This year")
* Then we could display a message (maybe a close-able notification?) to advise users they are viewing reports the previous day/week/month/year and that they can <a>click to view Today/This week/This month/This year's report </a>

What do you think?

tsteur:
fix https://github.com/matomo-org/matomo/issues/13754

implementing this already now as we'll want to patch this on the cloud

### DIFF
--- a/core/Plugin/Menu.php
+++ b/core/Plugin/Menu.php
@@ -223,7 +223,7 @@ class Menu
                     $period = Period\Factory::build($defaultPeriod, $defaultDateObj);
                     $endDate = $period->getDateEnd();
 
-                    if (!$endDate->isLater($siteCreationDate)) {
+                    if ($endDate->isEarlier($siteCreationDate)) {
                         // when selected date is before site creation date or it is the site creation day
                         $defaultDate = $siteCreationDate->toString();
                     }
--- a/core/Plugin/Menu.php
+++ b/core/Plugin/Menu.php
@@ -202,33 +202,34 @@ class Menu
         }
         if (empty($defaultDate)) {
             $defaultDate = $userPreferences->getDefaultDate();
-            if ($defaultPeriod !== 'range' && !empty($defaultDate) && $defaultDate !== 'today') {
-                // not easy to make it work for range... is rarely the default anyway especially when just setting up
-                // Matomo as this logic is basically only applied on the first day a site is created
-                // no need to run logic when today is selected. It basically runs currently only when "yesterday" is selected
-                // as a default date but would also support future new default dates like past month etc.
-                try {
-                    $siteCreationDate = Site::getCreationDateFor($websiteId);
-                    $siteTimezone = Site::getTimezoneFor($websiteId);
+        }
 
-                    if (!empty($siteCreationDate)) {
-                        if (is_numeric($defaultDate)) {
-                            $defaultDate = (int) $defaultDate; //prevent possible exception should defaultDate be a string timestamp
-                        }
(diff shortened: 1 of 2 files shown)


dev-clavis:
Could it be that the Menu test fails because the period is week and for week it shouldn't matter if the day is previous creation date because the date is within the same week?

`
'idSite' => 1,
'period' => 'week',
-'date' => '2020-08-08',
+'date' => '2020-08-07',
`
https://travis-ci.org/github/matomo-org/matomo/jobs/717098658#L736


`
// date is before site creation date but it is the current week so no need to change it
        $default = $this->menu->urlForDefaultUserParams($this->idSiteToday, 'week', '2020-08-06');
        $this->assertEquals([
            'idSite' => $this->idSiteToday,
            'period' => 'week',
            'date' => '2020-08-06',
        ], $default);
`

tsteur:
@dev-clavis I'll have a look when I continue working on this but basically added some code so the week should work as expected which is needed when someone manually sets a very old date. I simply haven't adjusted the expected test result yet I think.

tsteur:
Tests should pass now. There is one test failure and one UI failure in https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/41910/ but seems unrelated to this change.

core/Plugin/Menu.php
```diff
@@ -194,11 +197,40 @@ public function urlForDefaultUserParams($websiteId = false, $defaultPeriod = fal
...
+                    $siteCreationDate = Date::factory($siteCreationDate, $siteTimezone);
+                    $defaultDateObj = Date::factory($defaultDate, $siteTimezone);
+
+                    $period = Period\Factory::build($defaultPeriod, $defaultDateObj);
+                    $endDate = $period->getDateEnd();
+
+                    if (!$endDate->isLater($siteCreationDate)) {
+                        // when selected date is before site creation date or it is the site creation day
+                        $defaultDate = $siteCreationDate->toString();
+                    }
```
diosmosis, line 229: This sets the date to load to the site creation date if the default period end date is past the site creation date correct? So if the site creation date is 2015-02-02, and `$defaultDate` is 2020-01-01 w/ period = year, then the date is always set to 2015-02-02. Is this the intended functionality? It doesn't sound like what the PR title implies.

tsteur: 👍 adjusted the PR title @diosmosis

diosmosis: But then should it be `$endDate->isEarlier($siteCreationDate)`? To set it "if select date is before site creation date"?

tsteur: it's pretty much the same or not? `!$endDate->isLater == $endDate->isEarlier()`? Will change the code ... tests still pass that way

diosmosis: Oh I see, I missed the `!` while reading, LGTM

diosmosis (approved):

PR #16290, opened 2020-08-13
Fix error when primary key is required

tsteur:
refs https://wordpress.org/support/topic/error-in-visits-over-time/

refs https://forum.matomo.org/t/matomo-on-digitalocean-managed-database-cluster-mysql-8-sql-require-primary-key-error/37623

Putting this into Matomo 4 since it is required for Matomo for WordPress as people can't change DB settings on wordpress.com etc.

### DIFF
--- a/core/DataAccess/LogAggregator.php
+++ b/core/DataAccess/LogAggregator.php
@@ -298,7 +298,15 @@ class LogAggregator
             ) {
                 $createTableSql = str_replace($tempTableIdVisitColumn, $tempTableIdVisitColumn . ', PRIMARY KEY (`idvisit`)', $createTableSql);
 
-                $readerDb->query($createTableSql);
+                try {
+                    $readerDb->query($createTableSql);
+                } catch (\Exception $e) {
+                    if ($readerDb->isErrNo($e, \Piwik\Updater\Migration\Db::ERROR_CODE_TABLE_EXISTS)) {
+                        return;
+                    } else {
+                        throw $e;
+                    }
+                }
             } else {
                 throw $e;
             }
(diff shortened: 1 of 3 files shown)


tests/PHPUnit/Integration/DataAccess/LogAggregatorTest.php
```diff
@@ -80,6 +90,108 @@ public function test_generateQuery()
...
+        $this->assertSame($expected, $query);
+    }
+
+    private function setSqlRequirePrimaryKeySetting($val)
+    {
+        try {
+            $this->logAggregator->getDb()->exec('SET SESSION sql_require_primary_key=' . $val);
+        } catch (\Exception $e) {
+            if ($this->logAggregator->getDb()->isErrNo($e, 1193)) {
+                // ignore General error: 1193 Unknown system variable 'sql_require_primary_key'
```
tsteur, line 157: seems to require mysql8 or so. works for me locally on mysql8

core/DataAccess/LogAggregator.php
```diff
@@ -292,8 +293,17 @@ private function createTemporaryTable($unprefixedSegmentTableName, $segmentSelec
         } catch (\Exception $e) {
             if ($readerDb->isErrNo($e, \Piwik\Updater\Migration\Db::ERROR_CODE_TABLE_EXISTS)) {
                 return;
+            } elseif ($readerDb->isErrNo($e, \Piwik\Updater\Migration\Db::ERROR_CODE_REQUIRES_PRIMARY_KEY)
```
sgiehl, line 296: haven't checked that, but which error is thrown earlier? The table exists or the primary key error? If it's the primary key error the second query might throw a table exists, which isn't ignored then...

tsteur: @sgiehl I checked the table exists error would be triggered first but adding it just to be safe 👍

sgiehl (approved):

EmilMoe:
How long will it take before this is released?

tsteur:
@EmilMoe there will be a beta available likely in the next 2 weeks. The final release might be still 2 months or so away

PR #16295, opened 2020-08-14
Number formatting may be wrong when eg German language is used 

tsteur:
In custom reports. fix dev-1973.

Ran the tests in custom reports and seems to work fine

### DIFF
--- a/core/Plugin/ArchivedMetric.php
+++ b/core/Plugin/ArchivedMetric.php
@@ -162,6 +162,11 @@ class ArchivedMetric extends Metric
         return $value;
     }
 
+    public function getType()
+    {
+        return $this->type;
+    }
+
     public function getTranslatedName()
     {
         if (!empty($this->translatedName)) {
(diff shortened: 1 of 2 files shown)


sgiehl:
Should we maybe create a new issue to refactor that whole behavior in a later release?
I guess the problem is, that custom reports use metric classes for all available metrics, while all core reports use string arrays as metrics (except of processed ones). Imho it would make sense to always use metric classes for all reports. That way we would have the same behavior everywhere and could do the formatting for all metrics at the same place maybe? 🤔

tsteur:
It's actually not so much the problem that metric classes were used but that custom reports basically returned processedMetrics for metrics that weren't processed. It's of course still kind of doing it partially and it's not ideal but reckon for now don't need to create an issue or change anything further. Ideally getProcessedMetrics and metricsToFormat be separated eventually.

sgiehl (approved):
worked locally

PR #16302, opened 2020-08-17
New feature to exclude tracking requests via config file: INI setting exclude_requests

tsteur:
Was needed for Cloud to exclude requests on demand without needing to deploy.

Will create FAQ once merged. https://matomo.org/faq/how-to/how-do-i-exclude-specific-tracking-requests-in-matomo/


sgiehl (commented):
Should we maybe also add a system test that excludes some visits using the new setting?
Otherwise looks good to merge.

PR #16311, opened 2020-08-17
Updates php-di to 6.2.1

### LINKED ISSUE #15974: update php-di to latest version (5->6)
see https://github.com/matomo-org/matomo/issues/15961#issuecomment-631015194

branch with simple test by @sgiehl:
https://github.com/matomo-org/matomo/compare/phpdi

This fixes the deprecation warning of a dependency:
> Package container-interop/container-interop is abandoned, you should avoid using it. Use psr/container instead.

Migration guide: https://github.com/PHP-DI/PHP-DI/blob/6.0.0/doc/migration/6.0.md
Changelogs: https://github.com/PHP-DI/PHP-DI/releases

sgiehl:
After merging this we may need to update the DI definitions if various other plugins.
Once this is finished and approved I'll create PRs accordingly...

fixes #15974


sgiehl:
@tsteur  Not sure if updating the lib might cause problems while updating. The method `DI\object` has been removed, which is in use by some plugin configs. So it may cause fatals unless the plugin is updated together with core.

Also did not yet find a easy way to solve this: 

>  there are some issues with having references in functions defined in configuration like here https://github.com/matomo-org/matomo/blob/0187b71729484a9f173d978239ff0a3cb37a5d7b/config/environment/test.php#L101

Might be required to change all event definitions to not use references if there is no other solution.... Will have a closer look at that...

tsteur:
@sgiehl is there any way that we define `\Di\object` ourselves maybe so it won't break? I just had a look it be mostly only PaidAds and SearchEngine plugin if I'm seeing this right? The other ones be mostly cloud related plugins or core plugins.

sgiehl:
guess we could try using autowire as alias, might work in the most cases...

sgiehl:
Wrapping the observer functions into `\DI\value()` seems to fix the problem. Not sure if that has any disadvantages 🤷 
At least the tests seem to pass again.

I've also added a proxy for `\DI\object()` to `\DI\autowire()`, which should circumvent any possible failures due to plugins still using `\DI\object()`

Should be ready for a first review...

tsteur:
@sgiehl can you adjust the docs in https://developer.matomo.org/guides/dependency-injection ? It's showing `DI\object` a few times and it be great to describe when to use `autowire` and when `create`. Not sure if it's also mentioned in other places (likely not).

CHANGELOG.md
```diff
@@ -142,6 +142,9 @@ These are only recommendations (because we will keep backward compatibility for
 * The following dimensions have been removed and replaced with versions that measure seconds: visitor_days_since_first, visitor_days_since_last, visitor_days_since_order
 * The _idvc, _idts, _viewts and _ects tracker parameters are no longer used, the values are calculated server side.
   Note: tracking these values server side means replaying log data in the past will result in inaccurate values for these dimensions.
+* The Dependency Injection library PHP-DI was updated. Some definitions need to be updated:
```
tsteur, line 145: should we link to https://php-di.org/doc/migration/6.0.html ?

tsteur:
The docs mention also `DI\link` which seems now removed? Need to update this in our docs and changelog

tsteur:
@sgiehl also a question. Any reason we're using `create` here: https://github.com/matomo-org/matomo/pull/16311/files#diff-f96b6870514986e412c7f5454b5313a9R21 and `autowire` here: https://github.com/matomo-org/matomo/pull/16311/files#diff-cdccc896f5d7cfc50295cb535e230070R5

Which one is usually the better one to use?

CHANGELOG.md
```diff
@@ -142,6 +142,9 @@ These are only recommendations (because we will keep backward compatibility for
...
+  * The Method `\DI\object()` has been removed. You can use `\DI\autowire()` or `\DI\create()` instead
+  * Defining global observer functions in config now requires the functions to be wrapped in `\DI\value()`
```
tsteur, line 147: maybe add `unless they are a factory`?

tsteur:
Generally looks good otherwise. Be great to clarify usage of create vs autowire, update the docs etc. And be great if you could then also update the other free plugins and premium features after merging

tsteur:
I suppose we might also need to create new releases for plugins that already had a 4.x release

sgiehl:
If I understood that correctly: `DI\autowire` needs to be used if any constructor parameter needs to be set by DI. If the definition already defines all parameters or none is needed you can use `DI\create`. So actually I guess it should be safe to always use autowire maybe 🤷

sgiehl:
> I suppose we might also need to create new releases for plugins that already had a 4.x release

If their DI config need to be updated... For some plugins it's only test config, so a new release might not be needed...

tsteur:
reading eg https://stackoverflow.com/questions/49474983/how-to-use-autowire-create-and-get-in-definitions-for-php-di I reckon autowire is a good choice that should always work.

tsteur:
looks otherwise good to merge. Need to update docs though.

sgiehl:
I've also added a proxy from DI\link to DI\get now, so plugins that could possibly use DI\link won't break on update...

tsteur:
Sweet. Feel free to merge @sgiehl . Looks like some merge conflicts will need to be resolved.

PR #16314, opened 2020-08-17
Tests no memory limit

Findus23:
While testing around with php8, I stumbled across https://github.com/matomo-org/matomo/pull/15472/commits/79603337e81a1a7cfe889c34fdb4e9cb740d253f which is a test that tries to force a `PHP Fatal error:  Allowed memory size of X bytes exhausted` on php-cli.

Unfortunately, on my system (and probably others too) php-cli has a memory_limit of `-1` which means the PC will run out of memory before one can notice what command is running.

This PR adds a slight hack to skip this test in this case.

### DIFF
--- a/tests/PHPUnit/System/ConsoleTest.php
+++ b/tests/PHPUnit/System/ConsoleTest.php
@@ -145,6 +145,16 @@ class ConsoleTest extends ConsoleCommandTestCase
 
     public function test_Console_handlesFatalErrorsCorrectly()
     {
+        $cliPhp = new CliPhp();
+        $php = $cliPhp->findPhpBinary();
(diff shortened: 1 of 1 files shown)


tests/PHPUnit/System/ConsoleTest.php
```diff
@@ -145,6 +145,16 @@ public function test_Console_ReturnsCorrectExitCode_IfCommandDoesNotEmitAnything
 
     public function test_Console_handlesFatalErrorsCorrectly()
     {
+        $cliPhp = new CliPhp();
```
tsteur, line 148: @Findus23 I suppose `ini_get('memory_limit');` didn't work there?

PR #16316, opened 2020-08-18
Fix plugins not being updated when updating core

tsteur:
Seems to have regressed in https://github.com/matomo-org/matomo/pull/15770

Haven't tested it yet but this should work. We'll then also need to fix this in Matomo 4 separately because the token_auth check wouldn't work there because it would check against the new app specific token auth table which wouldn't exist yet at the time this is executed.

Unfortunately this will basically only work on the next update. So if people update to 3.14.1 it will only work when they update from that version. It won't work yet from 3.14.0 to 3.14.1 since the old code would be used there.


sgiehl (approved):

tsteur:
will replace this PR with a better solution

PR #16318, opened 2020-08-18
Custom token check when updating from 3.x to 4.X

tsteur:
Follow up from https://github.com/matomo-org/matomo/pull/16316

The token_auth check wouldn't work there because it would check against the new app specific token auth table which wouldn't exist yet at the time this is executed. At this time only the files would be updated but no DB migration executed yet meaning it can't authenticate the user.

Haven't tested this code yet.  Any other idea @sgiehl @diosmosis ? The only other thing I can think of be to use redirects in https://github.com/matomo-org/matomo/pull/16316/files#diff-f879da4f4ea7f582701f6cd753dad715R133 instead of issuing an HTTP request. But the upgrade to Matomo 4 would maybe destroy the session so this would maybe not work.](url)

### DIFF
--- a/plugins/CoreUpdater/Controller.php
+++ b/plugins/CoreUpdater/Controller.php
@@ -13,6 +13,7 @@ use Piwik\AssetManager;
 use Piwik\Common;
 use Piwik\Config;
 use Piwik\DataTable\Renderer\Json;
+use Piwik\Db;
 use Piwik\DbHelper;
 use Piwik\Filechecks;
 use Piwik\FileIntegrity;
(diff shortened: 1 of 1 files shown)


tsteur:
Or we could generate a random token in step1 in option table and check for that in the second step instead of the check super user access?

diosmosis:
@tsteur instead of a token auth, can we just use a temporary nonce? eg, before calling it in the update process, we create a a random nonce and store it somewhere (option table, maybe) set it in a query param, then in this controller action, check the query param == the stored nonce, and that it was invoked within a couple minutes. if it matches, remove stored nonce and continue updating. this ensures the only way the controller action can be called is during the update process

plugins/CoreUpdater/Controller.php
```diff
@@ -181,7 +182,19 @@ public function oneClickUpdatePartTwo()
...
-            Piwik::checkUserHasSuperUserAccess();
+            try {
+                Piwik::checkUserHasSuperUserAccess();
+            } catch (Exception $e) {
+                $token = Common::getRequestVar('token_auth', '', 'string');
+                $userTable = Common::prefixTable('user');
+                $userTableColumns = DbHelper::getTableColumns($userTable);
+                if (empty($token)
+                    || (int) \Piwik\Updater::getCurrentComponentVersion('core') >= 4
+                    || !isset($userTableColumns[$userTableColumns])
```
sgiehl, line 193: That seems incorrect. Guess the checked index needs to be something else?

sgiehl (changes requested):

tsteur:
will replace this PR with a better solution

PR #16322, opened 2020-08-19
Fix plugins might not be updated when updating core

tsteur:
Seems to have regressed in #15770

Haven't tested it yet fully but this should work. We'll then also need to merge this into Matomo 4. The previous token_auth check wouldn't work there because it would check against the new app specific token auth table which wouldn't exist yet at the time this is executed. Therefore using an nonce which is better in general anyway so the action can be only called when initiated by the first update step.

Unfortunately this will basically only work on the next update. So if people update to 3.14.1 it will only work when they update from that version. It won't work yet from 3.14.0 to 3.14.1 since the old code would be used there.

### DIFF
--- a/plugins/CoreUpdater/Controller.php
+++ b/plugins/CoreUpdater/Controller.php
@@ -13,6 +13,7 @@ use Piwik\AssetManager;
 use Piwik\Common;
 use Piwik\Config;
 use Piwik\DataTable\Renderer\Json;
+use Piwik\Db;
 use Piwik\DbHelper;
 use Piwik\Filechecks;
 use Piwik\FileIntegrity;
@@ -179,8 +180,24 @@ class Controller extends \Piwik\Plugin\Controller
 
         $messages = [];
 
+        $nonce = Common::getRequestVar('nonce', '', 'string');
+        if (empty($nonce)) {
+            return json_encode(['no token']);
+        }
+        $value = Option::get('NonceOneClickUpdatePartTwo');
+        if (empty($value)) {
+            return json_encode(['invalid token']);
+        }
+        $value = json_decode($value, true);
+
+        if (empty($value['nonce'])
+            || empty($value['ttl'])
+            || time() > (int) $value['ttl']
+            || $nonce !== $value['nonce']) {
+            return json_encode(['invalid nonce or nonce expired']);
+        }
+
         try {
-            Piwik::checkUserHasSuperUserAccess();
             $messages = $this->updater->oneClickUpdatePartTwo();
         } catch (UpdaterException $e) {
             $messages = $e->getUpdateLogMessages();
@@ -189,7 +206,7 @@ class Controller extends \Piwik\Plugin\Controller
             $messages[] = $e->getMessage();
         }
 
-        echo json_encode($messages);
+        return json_encode($messages);
     }
 
     public function oneClickResults()
--- a/plugins/CoreUpdater/Controller.php
+++ b/plugins/CoreUpdater/Controller.php
@@ -222,10 +222,10 @@ class Controller extends \Piwik\Plugin\Controller
         } elseif ($error) {
             $view = new View('@CoreUpdater/updateHttpError');
             $view->error = $error;
-            $view->feedbackMessages = safe_unserialize(Common::unsanitizeInputValue(Common::getRequestVar('messages', '', 'string', $_POST)));
         } else {
             $view = new View('@CoreUpdater/updateSuccess');
         }
+        $view->feedbackMessages = safe_unserialize(Common::unsanitizeInputValue(Common::getRequestVar('messages', '', 'string', $_POST)));
 
         $this->addCustomLogoInfo($view);
         $this->setBasicVariablesView($view);
--- a/plugins/CoreUpdater/Controller.php
+++ b/plugins/CoreUpdater/Controller.php
@@ -225,7 +225,11 @@ class Controller extends \Piwik\Plugin\Controller
         } else {
             $view = new View('@CoreUpdater/updateSuccess');
         }
-        $view->feedbackMessages = safe_unserialize(Common::unsanitizeInputValue(Common::getRequestVar('messages', '', 'string', $_POST)));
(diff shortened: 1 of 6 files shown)


plugins/CoreUpdater/Updater.php
```diff
@@ -124,15 +126,20 @@ public function updatePiwik($https = true)
...
+        $nonce = Common::generateUniqId();
+        Option::set('NonceOneClickUpdatePartTwo', json_encode(['nonce' => $nonce, 'ttl' => $validFor10Minutes]));
 
-        $response = Http::sendHttpRequest($partTwoUrl, 300);
-        $response = @json_decode($response, $assoc = true);
+        PiwikCache::flushAll();
 
-        if (!empty($response)) {
-            $messages = array_merge($messages, $response);
+        $cliMulti = new CliMulti();
```
tsteur, line 135: not 100% sure it's a good idea using CliMulti but reckon it might be beneficial in case there Matomo can't be easily reached through the internet. This way it might cause problems only for those where async using CLI doesn't work and the Matomo can't be reached through the internet.

plugins/CoreUpdater/Updater.php
```diff
@@ -124,15 +126,20 @@ public function updatePiwik($https = true)
...
+        $responses = $cliMulti->request(['?module=CoreUpdater&action=oneClickUpdatePartTwo&nonce=' . $nonce]);
+        if (!empty($responses)) {
+            $response = array_shift($responses);
+            $response = @json_decode($response, $assoc = true);
+            if (!empty($response) && is_array($response)) {
+                $messages = array_merge($messages, $response);
```
tsteur, line 139: one problem here is that any error during  `oneClickUpdatePartTwo` won't trigger an exception and therefore no error will be set here https://github.com/matomo-org/matomo/blob/3.14.1-b1/plugins/CoreUpdater/Controller.php#L167 and therefore the user won't see any of the messages should something go wrong. This was already a problem with the original solution but don't know if we want to show error messages for this action or simply ignore them as it affects only marketplace plugins? And they could simply update them afterwards

plugins/CoreUpdater/Updater.php
tsteur: eg in my case the original http call wouldn't have worked because:

> Got invalid response from API request: https://.... Response was 'curl_exec: SSL certificate problem: unable to get local issuer certificate. Hostname requested was: ...'

plugins/CoreUpdater/Controller.php
```diff
@@ -179,8 +180,24 @@ public function oneClickUpdatePartTwo()
 
         $messages = [];
 
+        $nonce = Common::getRequestVar('nonce', '', 'string');
+        if (empty($nonce)) {
+            return json_encode(['no token']);
+        }
+        $value = Option::get('NonceOneClickUpdatePartTwo');
```
tsteur, line 187: not using a constant here as depending how they upgrade or code changes maybe it wouldn't exist. could use one though

plugins/CoreUpdater/Updater.php
diosmosis: Do you mean a fatal error? Or something else? If it's a fatal error, perhaps the safemode hook could be used. Not sure if it's worth the effort though.

plugins/CoreUpdater/Controller.php
diosmosis: 👍

diosmosis (approved):

diosmosis:
left one comment, otherwise lgtm

plugins/CoreUpdater/Updater.php
tsteur: @diosmosis I meant eg if there's an error like I had it `Got invalid response from API request: https://.... Response was 'curl_exec: SSL certificate problem: unable to get local issuer certificate. Hostname requested was: ...'` or if there's suddenly a bug with the `nonce` and the call is not actually working etc. Before the token_auth was missing so the action for the 2nd step would have always returned an error but the error was never shown.

diosmosis: I guess we'd have to handle this when the controller action is called? Ie, "cannot parse json response, response is: curl-exec: ..." or "invalid token, could not complete install". or maybe i'm missing something?

diosmosis: Note: I don't know if it's actually that important.

plugins/CoreUpdater/Updater.php
```diff
@@ -151,6 +182,12 @@ public function oneClickUpdatePartTwo()
     {
         $messages = [];
 
+        if (!Marketplace::isMarketplaceEnabled()) {
```
tsteur, line 185: had marketplace disabled locally and it was actually before resulting in below error but because we don't show messages we never noticed basically

plugins/CoreUpdater/templates/updateSuccess.twig
```diff
@@ -36,6 +36,17 @@
             </div>
         </div>
 
+        <h2>{{ 'CoreUpdater_UpdateLog'|translate }}</h2>
```
tsteur, line 39: now showing a log of messages... this can help troubleshoot issues and would show a message if for example plugins were disabled during the update etc.

tsteur:
@diosmosis updated the PR.

Noticed the second step was actually failing when marketplace was disabled.

Should something go wrong because user has invalid SSL falling back to execute the second action within the same request. This should usually work quite well especially between minor and patch releases. 

And instead of any special error handling I reckon it's actually better to simply always show the log output in the success screen as well. So people notice eg if plugins were disabled. See example in https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/42112/OneClickUpdate_update_success.png

Note that the translation key appears and I think that's because of some caching as I think it updates from the latest stable where the translation key was not present yet. Locally this works for me. See https://builds-artifacts.matomo.org/matomo-org/matomo/3.x-dev/42112/OneClickUpdate_update_success.png  Also note that the first tick is a bit moved to the right which seems to. be because it's the first line right after `<code>✓ foo...`. I reckon it's not too important and not sure if there's much we could do about it except for adding another newline first.
image

PR #16323, opened 2020-08-19
Create plugin.json files if needed for plugins moved from core to marketplace

tsteur:
While doing our testing when updating from Matomo core 3.14.0 or older Matomo won't update the Provider and custom variables plugin. There's no real possible workaround because we can't change the update process so figured the best and only workaround be to continue shipping the plugins for a while with Matomo see https://github.com/matomo-org/matomo-package/pull/112 . The plugins won't be updated because Matomo thinks the plugins are bundled with core in https://github.com/matomo-org/matomo/blob/3.14.1-b1/plugins/Marketplace/Api/Client.php#L170 (could think about removing that check in the future however would make the post request a lot larger and put bit more load on the marketplace potentially but nothing serious I suppose)

When people then upgrade from older Matomo versions to a newer Matomo 4.X version without the Provider and Custom Variables plugin, this change will make sure that the marketplace will look for updates and find them. By adding the plugin.json the plugins will be no longer bogus.

What is still a problem is that when updating from a version before Matomo core 3.14.1 is that the 2 plugins will be deactivated. There's no real workaround since we can't know whether they were activated previously. What happens is that a user eg updates from Matomo 3.10 to Matomo 4.8 (the 2 plugins are no longer shipped with Matomo): 

* Matomo then updates the core
* because it thinks the plugins belong to core it won't look for an update
* after the files were updated it realises the 2 plugins aren't compatible and therefore deactivates them 
* during the migration we then create the plugin.json files 
* afterwards Matomo will show the user the available updates
* however the plugin stays still deactivated and user needs to activate it. (because we can't know whether they were activated previously).

Without this change Matomo would never even show there is an update for the plugins available.

### DIFF
--- a/core/Updates/4.0.0-b1.php
+++ b/core/Updates/4.0.0-b1.php
@@ -210,6 +210,49 @@ class Updates_4_0_0_b1 extends PiwikUpdates
             // switch to default provider if GeoIp Legacy was still in use
             LocationProvider::setCurrentProvider(LocationProvider\DefaultProvider::ID);
         }
+
+        // eg the case when not updating from most recent Matomo 3.X and when not using the UI updater
+        // afterwards the should receive a notification that the plugins are outdated
+        self::ensureCorePluginsThatWereMovedToMarketplaceCanBeUpdated();
+    }
+
+    public static function ensureCorePluginsThatWereMovedToMarketplaceCanBeUpdated()
+    {
+        $plugins = ['Provider', 'CustomVariables'];
+        $pluginManager = Manager::getInstance();
+        foreach ($plugins as $plugin) {
+            if ($pluginManager->isPluginThirdPartyAndBogus($plugin)) {
+                $pluginDir = Manager::getPluginDirectory($plugin);
+
+                if (is_dir($pluginDir) &&
+                    file_exists($pluginDir . '/' . $plugin . '.php')
+                    && !file_exists($pluginDir . '/plugin.json')
+                    && is_writable($pluginDir)) {
+                    file_put_contents($pluginDir . '/plugin.json', '{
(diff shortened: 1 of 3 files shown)


plugins/CoreUpdater/Updater.php
```diff
@@ -168,6 +169,8 @@ public function oneClickUpdatePartTwo()
         try {
 
             if (Marketplace::isMarketplaceEnabled()) {
+                Updates_4_0_0_b1::ensureCorePluginsThatWereMovedToMarketplaceCanBeUpdated();
```
tsteur, line 172: when updating from 3.14.1 to 4.X then this method would be executed... we'll create the needed plugin.json files shortly before checking for updates so the plugins won't be marked as "bogus"

core/Updates/4.0.0-b1.php
```diff
@@ -210,6 +210,50 @@ public function doUpdate(Updater $updater)
             // switch to default provider if GeoIp Legacy was still in use
             LocationProvider::setCurrentProvider(LocationProvider\DefaultProvider::ID);
         }
+
+        // eg the case when not updating from most recent Matomo 3.X and when not using the UI updater
+        // afterwards the should receive a notification that the plugins are outdated
+        self::ensureCorePluginsThatWereMovedToMarketplaceCanBeUpdated();
```
tsteur, line 216: This logic runs for people that upgrade from 3.14.0 and older because there the `oneClickUpdatePartTwo` wouldn't be executed as it either didn't exist yet or it wasn't working. 
This logic runs also for people who update on the command line.

We then create the files during the update step so at least people will be made aware there's an update available on the marketplace.

PR #16325, opened 2020-08-19
Fix formatting of metrics in milliseconds

sgiehl:
Guess that only affects the formatting of some metrics (e.g. new performance metrics) in custom reports...

### DIFF
--- a/core/Columns/Dimension.php
+++ b/core/Columns/Dimension.php
@@ -456,7 +456,11 @@ abstract class Dimension
             case Dimension::TYPE_DURATION_S:
                 return $formatter->getPrettyTimeFromSeconds($value, $displayAsSentence = false);
             case Dimension::TYPE_DURATION_MS:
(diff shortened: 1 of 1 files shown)


core/Columns/Dimension.php
```diff
@@ -456,7 +456,7 @@ public function formatValue($value, $idSite, Formatter $formatter)
             case Dimension::TYPE_DURATION_S:
                 return $formatter->getPrettyTimeFromSeconds($value, $displayAsSentence = false);
             case Dimension::TYPE_DURATION_MS:
-                return $formatter->getPrettyTimeFromSeconds($value, $displayAsSentence = true);
+                return $formatter->getPrettyTimeFromSeconds($value/1000, $displayAsSentence = true);
```
tsteur, line 459: @sgiehl might be good to use same logic as here? https://github.com/matomo-org/matomo/blob/3.14.1-b1/core/Plugin/ArchivedMetric.php#L151-L155

sgiehl: updated it

Issue #16327, opened 2020-08-20
Archiving Error 


reddipravi:
While trying to run archive reports using the command line in Matomo, we are facing errors. 

command: 
php ./htdocs/console core:archive --force-idsites=29 --force-periods=week --url=******

 INFO [2020-08-19 22:16:05] 28 NOTES
 INFO [2020-08-19 22:16:05] 28 - Async process archiving supported, using CliMulti.
 INFO [2020-08-19 22:16:05] 28 - Reports for today will be processed at most every 3600 seconds. You can change this value in Matomo UI > Settings > General Settings.
 INFO [2020-08-19 22:16:05] 28 - Reports for the current week/month/year will be requested at most every 3600 seconds.
 INFO [2020-08-19 22:16:05] 28 - Will process 1 websites (--force-idsites)
 INFO [2020-08-19 22:16:05] 28 ---------------------------
 INFO [2020-08-19 22:16:05] 28 START
 INFO [2020-08-19 22:16:05] 28 Starting Matomo reports archiving...
 INFO [2020-08-19 22:16:06] 28 Will pre-process for website id = 29, period = day, date = last90
 INFO [2020-08-19 22:16:06] 28 - pre-processing all visits
] WARNING [2020-08-19 22:18:15] 28 /home/vcap/app/htdocs/core/Common.php(297): Notice - unserialize(): Error at offset 0 of 133333 bytes - Matomo 3.13.5 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)
 ERROR [2020-08-19 22:18:16] 28 Empty or invalid response '###!/usr/bin/env php
[...]

PR #16330, opened 2020-08-20
Trim multiple hashes from cli archive response

tsteur:
refs https://github.com/matomo-org/matomo/issues/16327

Noticed in the issue there was a user where the response had `###!/usr/bin/env php` in the beginning for some reason. It won't fix the actual issue but figured it's quick to handle it in case others have this issue too.


diosmosis:
Looks like a test is failing now.

tsteur:
Closing this for now as I think it won't fix the issue anyway. Could otherwise rewrite that logic but thinking it might not be needed and I don't want to do too many string operations there on all the archive responses. To be seen. Will recreate if needed.

PR #16333, opened 2020-08-21
show correct transitions info in "Did you know?" section

### LINKED ISSUE #16331: Transitions: Wrong info in the "Did you know" section
Under "Behaviour > Transitions" the "Did you know" section mentions that the transition report is available for Outlinks, Downloads. However these are not supported yet. Could we remove the names of the reports which are not supported yet https://github.com/matomo-org/matomo/issues/3899.

image

Findus23:
fix #16331

@KarthikRaja1388, can you make sure the information is now correct?

grafik


tsteur:
@KarthikRaja1388 can you have a look at above screenshot from @Findus23  and check if it's as expected?

PR #16337, opened 2020-08-22
Let plugin developers modify the password confirmation view: new Template.confirmPasswordContent event

dominik-th:
Similar to the `Template.loginNav` event it would be useful to inject custom elements to the password confirmation view as well.

The event consists of two parts, where the parameter is either `top` or `bottom`, same as in the [Login Template](https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/Login/templates/login.twig#L60-L63)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ These are only recommendations (because we will keep backward compatibility for 
 #### New PHP events
 
 * Added new event `Db.getTablesInstalled`, plugins should use to register the tables they create.
+* Added new event `Template.confirmPasswordContent`, plugins can inject HTML content to the password confirmation view.
 
 #### Breaking changes in PHP events
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,6 @@ These are only recommendations (because we will keep backward compatibility for 
 #### New PHP events
 
 * Added new event `Db.getTablesInstalled`, plugins should use to register the tables they create.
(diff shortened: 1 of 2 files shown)


tsteur:
Thanks for creating the PR @dominik-th  Just to understand things could you let us know about your use case here eg what kind of content you will be adding there? This will help us in the future should there ever be changes.

CHANGELOG.md
```diff
@@ -64,6 +64,7 @@ These are only recommendations (because we will keep backward compatibility for
 #### New PHP events
 
 * Added new event `Db.getTablesInstalled`, plugins should use to register the tables they create.
+* Added new event `Template.confirmPasswordContent`, plugins can inject HTML content to the password confirmation view.
```
tsteur, line 67: Cheers for thinking about adding this to the changelog. There's no need though since we don't officially support template events and having it in the changelog might send the wrong message. We won't be simply removing the events and we'll do our best to support them but should the UI or flows change for some reason they might disappear. It's unlikely this will happen in the next few years but wanted to mention it.

dominik-th:
My plugin allows users to create accounts via an openid connect identity provider. These users don't necessarily have a password and get stuck in the password confirmation view.

I want to add a button which redirects the user to the identity provider so they can do the password confirmation on their side if necessary.


Do you want me to remove the changelog entry? It might confuse others if template events aren't officially supported.

tsteur:
Thanks @dominik-th I have removed the changelog entry so I can merge 👍

dominik-th:
Thank you @tsteur 😊

PR #16340, opened 2020-08-24
Adjustments for CustomReports plugin

sgiehl:
While making CustomReport plugin compatible with Matomo 4, I've noticed a few dimensions/metrics having incorrect values.

* Metrics using a custom sqlSegment like `daysSinceFirstVisit` are archived incorrectly. As archiving currently uses the plain field value all metrics using custom sqlSegments and those based on them are incorrect. Using the `sqlSegment` in ArchiveMetrics::getSql() fixes that problem.

* Dimensions in milliseconds like the new performance metrics are showing 0 values (when used as dimension), as they are grouped incorrectly. The grouping already divided them by 1000, causing the formatting to 0 the values, as it divides them another time. Grouping them in 1000 segments instead seems more correct.

* The pagviews with performance timings metrics had an incorrect type, causing them formatted as time instead of number

### DIFF
--- a/core/Columns/Dimension.php
+++ b/core/Columns/Dimension.php
@@ -418,7 +418,7 @@ abstract class Dimension
             case Dimension::TYPE_BOOL:
                 return !empty($value) ? '1' : '0';
             case Dimension::TYPE_DURATION_MS:
-                return number_format($value / 1000, 2); // because we divide we need to group them and cannot do this in formatting step
+                return number_format($value / 1000, 2) * 1000; // because we divide we need to group them and cannot do this in formatting step
         }
         return $value;
     }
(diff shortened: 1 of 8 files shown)


tsteur:
@sgiehl 

> The pageviews with performance timings metrics had an incorrect type, causing them formatted as time instead of number

wouldn't that be expected? or you mean when requesting the API with format metrics off?

core/Columns/Dimension.php
```diff
@@ -418,7 +418,7 @@ public function groupValue($value, $idSite)
             case Dimension::TYPE_BOOL:
                 return !empty($value) ? '1' : '0';
             case Dimension::TYPE_DURATION_MS:
-                return number_format($value / 1000, 2); // because we divide we need to group them and cannot do this in formatting step
+                return number_format($value / 1000, 2) * 1000; // because we divide we need to group them and cannot do this in formatting step
```
tsteur, line 421: Not sure why we divide by 1000 in the first place then @sgiehl ?

sgiehl:
I guess to group them. Otherwise there would be multiple different values
ending up in the same formatted value...  Eg. 1246ms and 1249ms would both
be formatted as 1.24s

PR #16345, opened 2020-08-26
Allow execution of core and plugin specific JS tests only

tsteur:
eg by executing `http://domain/tests/javascript/?plugin=HeatmapSessionRecording`. In the past I always patched the `index.php` code manually but figured it be easier to have an option for this. That's because some plugin tests might not be compatible with each other as every JS adds different HTML code to the test. Eg when it runs Form Analytics, Heatmaps, Ab Tests, etc all in one go then the tests might not pass but they pass when executed independently. It doesn't fail because of any JS logic but because of the extra HTML they add to the file. It's possible this could be fixed but might be a bit of work. Will see first if I can get them to pass altogether.

Nonetheless sometimes it might be useful to have this opportunity to only add specific tests to the test page. It lets you eg execute tests just like Travis would do (when it runs the tests only for one plugin but not all plugins)

### DIFF
--- a/tests/javascript/README.md
+++ b/tests/javascript/README.md
@@ -8,3 +8,7 @@ The tests will create a database named `tracker_tests` and store several trackin
 ## Execute
 
 Either open http://piwik.example.com/tests/javascript/ in a browser or execute `phantomjs testrunner.js` on the command line. You can download PhantomJS here: http://phantomjs.org/
+
+To execute tests for a specific module use the `module` URL parameter, for example `&module=core`.
+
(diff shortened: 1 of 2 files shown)


tsteur:
note I also now fixed the tracker tests so they work mostly altogether. However, the `plugin` parameter as introduced by this PR will still be needed when trying to reproduce tests that fail on travis as it basically allows executing the tests just like travis:

* Core JS tracker tests
* Tag Manager JS tests
* The JS tests for a specific plugin

without including any JS tests of other plugins and without including the `head.php`  part of any other tests.

tsteur:
fyi I'll merge this one since it's only for tests anyway

PR #16346, opened 2020-08-26
Tweaks for archiving for enterprise demo.

diosmosis:
- Include exception stack trace if enabled in error message.
- If idSite does not exist, ignore it in archive invalidation.

FYI @tsteur, will be merging

### DIFF
--- a/core/DataAccess/Model.php
+++ b/core/DataAccess/Model.php
@@ -203,7 +203,13 @@ class Model
 
         $dummyArchives = [];
         foreach ($idSites as $idSite) {
-            $siteCreationTime = Date::factory(Site::getCreationDateFor($idSite));
+            try {
+                $siteCreationTime = Site::getCreationDateFor($idSite);
+            } catch (\Exception $ex) {
+                continue;
+            }
+
+            $siteCreationTime = Date::factory($siteCreationTime);
             foreach ($allPeriodsToInvalidate as $period) {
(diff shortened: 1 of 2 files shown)


core/DataAccess/Model.php
```diff
@@ -203,7 +203,13 @@ public function updateArchiveAsInvalidated($archiveTable, $idSites, $allPeriodsT
...
         $dummyArchives = [];
         foreach ($idSites as $idSite) {
-            $siteCreationTime = Date::factory(Site::getCreationDateFor($idSite));
+            try {
+                $siteCreationTime = Site::getCreationDateFor($idSite);
+            } catch (\Exception $ex) {
+                continue;
+            }
+
+            $siteCreationTime = Date::factory($siteCreationTime);
```
tsteur, line 212: fyi @diosmosis  in this case $siteCreationTime might be not defined or pointing to the time of a previous site?

tsteur: actually forget it I see you continue above.

diosmosis: It should skip to the next iteration if it can't be set?

tsteur: Totally. sorry about that

diosmosis: oh, github showed me the first comment, but not the second...

PR #16347, opened 2020-08-26
Fetch report metadata and metrics site by site, if multiple sites are given in API.get

sgiehl:
Fetch report metadata or reports with idSites=all might result in an error as some plugins do not support it.
Fetching all reports and metadata site by site instead fixes that.

### DIFF
--- a/plugins/API/API.php
+++ b/plugins/API/API.php
@@ -340,59 +340,71 @@ class API extends \Piwik\Plugin\API
             $columnsMap[$column] = true;
         }
 
-        // find out which columns belong to which plugin
-        $columnsByPlugin = array();
-        $meta = \Piwik\Plugins\API\API::getInstance()->getReportMetadata($idSite, $period, $date);
-        foreach ($meta as $reportMeta) {
-            // scan all *.get reports
-            if ($reportMeta['action'] == 'get'
-                && !isset($reportMeta['parameters'])
-                && $reportMeta['module'] != 'API'
-                && !empty($reportMeta['metrics'])
-            ) {
-                $plugin = $reportMeta['module'];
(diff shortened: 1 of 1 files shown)


diosmosis:
Doesn't this multiply the number of archive queries by number of sites?

sgiehl:
That might be possible. Not sure if all archives are fetched with one query if idSites=all is given to an API method...
But it at least includes all reports. Fetching the report metadata with idSites=all excludes various reports as they don't support it.

diosmosis:
> That might be possible. Not sure if all archives are fetched with one query if idSites=all is given to an API method...

If multiple idSites are given to `Archive::get(`, it should be just one query using `idsite IN (`.

sgiehl:
Ok. Alternatively we would need to make all API methods and reports metadata compatible with multiple idSites. Tried catching errors in the first place instead of iterating through the sites, but that removes a lot of metrics, as the reports aren't available for multiple sites...

tsteur:
@sgiehl @diosmosis we won't be fixing this for now I reckon as it's not a priority and we'll need to see re performance impact etc. Also the API call to getREportMetadata  would need to be executed through `Request::processRequest` potentially etc re segments/dimensions and other things that might be site specific. If we continue to support this feature then we'd likely would need to fix this maybe in the plugins itself or so.

tsteur:
fyi documented this for now in https://github.com/matomo-org/developer-documentation/pull/377

tsteur:
I've created an issue for now in https://github.com/matomo-org/matomo/issues/16366 and we'll schedule this work eventually to see if other APIs are impacted too and what a proper fix be should we keep this feature. I'll close the PR for now but we might reopen it again later.

PR #16350, opened 2020-08-28
Provide more debug information in system check

### LINKED ISSUE #15931: Add more information to system check and be able to copy/paste for debugging purposes
This will help us better understand and debug issues

@Findus23 I think we maybe already have an issue for the copy/paste system check feature but couldn't find it right now.

Similar to Matomo for WP it would be great to have these features included when also developing the copy/paste feature:
* List of customised settings in config.ini.php compared to original global config
* List of enabled plugins and their versions 

We also want to show more information such as
* Browser User Agent
* Browser language
* Ideally when Archive was last started and finished 
image
* More server information eg 
image
* Mysql version 
image
* Number of sites and users
* What the matomo version was when it was installed

Ideally the copied system check would anonymise few things like 
* `replace(SettingsPiwik::getMatomoUrl(), '$MATOMO_URL', $input)`
* `replace(PIWIK_DOCUMENT_ROOT, '$MATOMO_DOC_ROOT', $input)`
* `replace(PIWIK_USER_PATH, '$PIWIK_USER_PATH', $input)`
* token like things `$value = preg_replace('/[[:xdigit:]]{31,80}/', 'TOKEN_REPLACED', $value);`
* salt and DB passwords etc from config
* etc.

We can generally reuse a lot of logic for this from WP-Matomo see https://github.com/matomo-org/wp-matomo/blob/develop/classes/WpMatomo/Admin/SystemReport.php and https://github.com/matomo-org/wp-matomo/blob/1.0.6/matomo.php#L124-L156

tsteur:
fix https://github.com/matomo-org/matomo/issues/15931

Since we will run into few issues once people upgrade to Matomo 4 I reckon this will be actually quite valuable. It proved invaluable so far for any Matomo for WordPress problem. Eventually can add more like how "Diagnostics" plugin can already identify any customised config setting etc.


tests/resources/screenshot-override/override.js
```diff
@@ -18,4 +18,8 @@ $(document).ready(function () {
     } else if ($.Velocity) {
         $.Velocity.mock = true;
     }
+
+    $('.entityTable.system-check tr:contains(Time) td:nth-child(2)').text('Not showing in tests');
```
tsteur, line 22: it works for me when I execute it manually but seems to not yet work in screenshot tests

plugins/Diagnostics/DiagnosticReport.php
```diff
@@ -102,6 +116,144 @@ public function getOptionalDiagnosticResults()
         return $this->optionalDiagnosticResults;
     }
 
+    public function getInformationalResults()
+    {
+        $results = [];
+
+        $results[] = DiagnosticResult::informationalResult('Matomo Version', Version::VERSION);
```
tsteur, line 123: I can put this into DI if needed.

plugins/Diagnostics/Diagnostic/ConfigInformational.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $this->translator = $translator;
+    }
+
+    public function execute()
+    {
+        $results = [];
+
+        if (SettingsPiwik::isMatomoInstalled()) {
+            $results[] = DiagnosticResult::informationalResult('Browser Archiving Enabled', Rules::isBrowserTriggerEnabled());
```
tsteur, line 38: fyi we don't really want to translate them likely cause it's mostly information for us... makes it easier to analyze things having it in english

core/Twig.php
```diff
@@ -434,6 +435,38 @@ protected function addFilter_number()
...
+            // just in case it was somehow show in a text
+            if (SettingsPiwik::isMatomoInstalled()) {
+                $string = str_replace(SettingsPiwik::getPiwikUrl(), '$MATOMO_URL', $string);
+                $string = str_replace(SettingsPiwik::getSalt(), '$MATOMO_SALT', $string);
+
+                $config = Config::getInstance();
+                $db = $config->database;
+                $string = str_replace($db['password'], '$DB_PASSWORD', $string);
+                $string = str_replace($db['host'], '$DB_HOST', $string);
+                $string = str_replace($db['dbname'], '$DB_NAME', $string);
```
diosmosis, line 463: should db user be replaced as well?

tsteur: sure can add it 👍

plugins/Diagnostics/Diagnostic/ReportInformational.php
```diff
@@ -0,0 +1,68 @@
...
+        return $results;
+    }
+
+    private function hadVisitsInLastDays($numDays)
+    {
+        $table = Common::prefixTable('log_visit');
+        $time = Date::now()->subDay($numDays)->getDatetime();
+
+        try {
+            $row = Db::fetchOne('SELECT count(idsite) from ' . $table . ' where visit_last_action_time > ? LIMIT 1', $time );
```
diosmosis, line 57: I could be wrong, but won't this still count every visit after the date? Should it be `SELECT idsite FROM ... LIMIT 1`?

tsteur: great find, will fix that

plugins/Diagnostics/Diagnostic/ReportInformational.php
```diff
@@ -0,0 +1,68 @@
...
+        return $results;
+    }
+
+    private function hadVisitsInLastDays($numDays)
+    {
+        $table = Common::prefixTable('log_visit');
+        $time = Date::now()->subDay($numDays)->getDatetime();
+
+        try {
+            $row = Db::fetchOne('SELECT idsite from ' . $table . ' where visit_last_action_time > ? LIMIT 1', $time );
```
tsteur, line 57: fyi @diosmosis I just realised this won't use an index. will provide a fix tmrw so it uses the index

PR #16355, opened 2020-08-30
use localized country names for segment selector

Findus23:
There might be a reason why this wasn't done initially, but I personally find it more accessible for people to see the country names in their language and ordered alphabetically instead of by country code

Before:
grafik

After:
grafik

If this is intentional, then you can simply close this PR.


sgiehl:
@Findus23 Using translated properties in segments might break easily as the translated name is part of the segment definition. As soon as one switches the languages or tries to share the segment with someone not using the same language the segment won't work anymore

Findus23:
@sgiehl That makes sense. I (baselessly) assumed that only the country code instead of the name would be stored in the segment.

PR #16364, opened 2020-09-01
Fix Custom opt out screen should not imply tracking consent given

### LINKED ISSUE #16362: Custom opt out screen should not imply tracking consent given
See the guide in https://developer.matomo.org/guides/tracking-javascript-guide#optional-creating-a-custom-opt-out-form

It suggests to call `_paq.push(['optUserOut']);` when a user wants to opt out. This would set the `mtm_consent_removed` cookie and is generally fine.

It suggests to call `_paq.push(['forgetUserOptOut']);` when a user wants to opt back in. What this does is remove the previously set `mtm_consent_removed` and then sets the `mtm_consent` cookie. This is generally OK. However, it becomes a problem as soon as someone was to actually switch from "not asking to consent and offering opt out" to a proper cookie consent. Suddenly, the people that opted back in using the custom opt out would be assumed to have consent given. However, this is not the case because they likely would have only seen a small opt out message. People don't give consent for tracking in the same way as a proper consent screen would require (specific, informed, ...)

The solution be likely quite easy. Instead of setting the cookie for consent given it should only remove a previously set `mtm_consent_removed` cookie:

```
            this.forgetUserOptOut = function () {
this.setConsentGiven(false);
            };
```

This way, should you ever change the legal basis for tracking users, then you don't run into any issues and users would be properly asked for consent.

tsteur:
fix https://github.com/matomo-org/matomo/issues/16362

As explained in the issue.

When `optUserOut` is called, it should set the `mtm_consent_removed` cookie.

When calling `forgetUserOptOut` then it should only undo the previous action and remove the `mtm_consent_removed` cookie. It should not set a `consent given` cookie because the user did likely not really give proper tracking consent (they would then instead implement https://developer.matomo.org/guides/tracking-consent ).

Noticed this while implementing a custom opt out screen for Matomo for WordPress


tsteur:
build js

PR #16372, opened 2020-09-02
Make sure query in system check uses index

tsteur:
fyi @diosmosis

### DIFF
--- a/plugins/Diagnostics/Diagnostic/ReportInformational.php
+++ b/plugins/Diagnostics/Diagnostic/ReportInformational.php
@@ -28,6 +28,8 @@ class ReportInformational implements Diagnostic
      */
     private $translator;
 
+    private $idSitesCached;
+
     public function __construct(Translator $translator)
     {
         $this->translator = $translator;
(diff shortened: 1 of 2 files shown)


plugins/Diagnostics/Diagnostic/ReportInformational.php
```diff
@@ -54,7 +56,8 @@ private function hadVisitsInLastDays($numDays)
         $time = Date::now()->subDay($numDays)->getDatetime();
 
         try {
-            $row = Db::fetchOne('SELECT idsite from ' . $table . ' where visit_last_action_time > ? LIMIT 1', $time );
+            $idSites = $this->getImplodedIdSitesSecure();
```
diosmosis, line 59: Not sure about this, but could we also use the index if we just did something like, `AND idsite > 0`?

tsteur: AFAIK Mysql doesn't support using both index parts (idsite and visit_last_action_time) when both are a range. MySQL stops using the index after the first range. So `idsite > 0` would be a range and therefore it would only use the idsite index.

diosmosis: In that case, can `Site::getIdSitesFromIdSitesString('all')` be used (w/ superuser access) to get all of them? Not really necessary, but might use existing cached data and reduce amount of code here.

tsteur: updated it @diosmosis

PR #16378, opened 2020-09-03
Merge 3.X into 4.X

tsteur:


sgiehl:
@tsteur It's better not to use Pull Request to merge changes from 3.x-dev to 4.x-dev. The next time someone tries to do another merge from 3.x-dev the same commits will be tried to merge again, as they were squashed when merging the PR and thus they don't exsist in 4.x-dev. I usual do the merge on a second branch to check the travis build, but push the merge on 4.x-dev directly

tsteur:
👍  forgot about that one. Will do next time. I reckon in this case it was likely (hopefully) the last time we had to merge 3.x into 4.x and instead issue separate PRs if needed or so

PR #16381, opened 2020-09-04
Remove duplicate CORS header handling in tracker (fixes #16380)

### LINKED ISSUE #16380: CORS access controls not applied for non-GET requests to the tracker
Matomo has a [configuration option to restrict CORS requests to specific domains]( https://matomo.org/faq/how-to/faq_18694/) and that's correctly implemented in the [CORSHandler]( https://github.com/matomo-org/matomo/blob/b09093780b6a906a0522bda2a8137ca280ab3f65/core/API/CORSHandler.php#L48-L55) in the case of `GET` requests. However, some [additional code](https://github.com/matomo-org/matomo/blob/8b21dbf2cfb06e9c9ed9ecbbd07e3b6b386bba3b/core/Tracker/Response.php#L108-L113) overrides the CORSHandler in the case of non-`GET` requests.

Unless there is a reason for bypassing the CORSHandler, it would be safer to treat all HTTP methods the same.

fmarier:


tsteur:
Note: This is in the tracker context only

@fmarier If we removed this then POST tracking requests wouldn't work anymore and always fallback to GET.

See eg 
image

This is currently their on purpose and because not really any information can be retrieved from this it shouldn't be any problem.

fmarier:
Thanks for the extra details @tsteur . That's exactly what I was wondering about (couldn't figure it out from the git history). Let's close the issue as wontfix then.

PR #16382, opened 2020-09-04
Urgent update to Danish translation

atjn:
I understand that you usually don't accept translation fixes directly in GitHub, but given that you don't update the translations for 3.x anymore, and that this is a pretty urgent issue, @sgiehl suggested that i should also create a direct PR for this fix.

The danish translation will curently tell the user that they have opted out of tracking, even though they are still opted in. We are basically lying to the user, which is not very nice, and also illegal.

The specific string is `CoreAdminHome.YouAreNotOptedOut`.
The current danish translation says (re-translated back to english):   "You have now opted out of cookies."
My new danish translation says (re-translated back to english):         "You are allowing tracking."

The string `CoreAdminHome.YouAreOptedOut` also has a minor issue. It specifically mentions cookies, but this mechanism does not just disable cookies, it disables all tracking.
The current danish translation says (re-translated back to english):   "You have opted out of cookies for web analysis."
My new danish translation says (re-translated back to english):         "You have opted out of tracking."

The strings `CoreAdminHome.UncheckToOptOut` and `CoreAdminHome.CheckToOptIn` are also updated, simply to align with the new wording.


tsteur:
Thanks @atjn  very much appreciated

PR #16383, opened 2020-09-04
Add new command to update marketplace plugins via console

tsteur:
This was useful when updating Matomo through git.

In this case I did basically 

```bash
git checkout matomo.js
git checkout piwik.js
rm -rf plugins/CustomDimensions
git checkout 4.x-dev
php composer.phar install 
git submodule update --recursive
 git submodule init &&  git submodule update
./console core:update
./console marketplace:update-plugins
```

Otherwise the marketplace plugins would not have been updated and would have been potentially disabled. Should only be interesting if someone updates Matomo in a production system through git (which we don't really recommend since it is less secure etc)

fyi @mattab

### DIFF
--- a/plugins/Marketplace/Commands/UpdateMarketplacePlugins.php
+++ b/plugins/Marketplace/Commands/UpdateMarketplacePlugins.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Matomo - free/libre analytics platform
+ *
+ * @link https://matomo.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ */
+
+namespace Piwik\Plugins\Marketplace\Commands;
+
+use Piwik\Container\StaticContainer;
+use Piwik\Plugin\ConsoleCommand;
+use Piwik\Plugins\CoreUpdater\Updater;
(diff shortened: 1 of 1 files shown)


plugins/Marketplace/Commands/UpdateMarketplacePlugins.php
```diff
@@ -0,0 +1,36 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class UpdateMarketplacePlugins extends ConsoleCommand
+{
+
+    protected function configure()
+    {
+        $this->setName('marketplace:update-plugins');
+        $this->setDescription('Migrates a measurable/website from one Matomo instance to another Matomo');
```
mattab, line 23: could we add a description?

mattab:
Sounds good. Would we want to mention this command in https://matomo.org/faq/how-to-install/faq_18271/ since it's the only time this would be useful maybe?

tsteur:
Actually I close that one again. It actually didn't really help as the marketplace plugins were still deactivated afterwards and could have been activated through the UI. We can always add it when needed.

PR #16387, opened 2020-09-04
Add more debug information for some specific archiving errors

tsteur:
We are currently troubleshooting some archiving issue and this code we are customer to add in order to troubleshoot what is happening. This be great to have in general.

### DIFF
--- a/core/ArchiveProcessor/PluginsArchiver.php
+++ b/core/ArchiveProcessor/PluginsArchiver.php
@@ -180,7 +180,7 @@ class PluginsArchiver
                         $this->params->getSegment() ? sprintf("(for segment = '%s')", $this->params->getSegment()->getString()) : ''
                     );
                 } catch (Exception $e) {
-                    throw new PluginsArchiverException($e->getMessage() . " - in plugin $pluginName", $e->getCode(), $e);
+                    throw new PluginsArchiverException($e->getMessage() . " - in plugin $pluginName with trace: " . $e->getTraceAsString(), $e->getCode(), $e);
                 } finally {
                     self::$currentPluginBeingArchived = null;
                 }
--- a/core/DataTable.php
+++ b/core/DataTable.php
@@ -1882,18 +1882,6 @@ class DataTable implements DataTableInterface, \IteratorAggregate, \ArrayAccess
         return $result;
     }
 
-    private function getDebugInfo()
-    {
-        $source = array();
-        if (!empty($_GET)) {
-            $source = $_GET;
-        }
-        if (!empty($_POST)) {
-            $source = $source + $_POST;
-        }
-        return json_encode($source);
-    }
-
     /**
      * Aggregates the $row columns to this table.
      *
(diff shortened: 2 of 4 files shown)


core/ArchiveProcessor/PluginsArchiver.php
```diff
@@ -180,7 +180,7 @@ public function callAggregateAllPlugins($visits, $visitsConverted, $forceArchivi
                         $this->params->getSegment() ? sprintf("(for segment = '%s')", $this->params->getSegment()->getString()) : ''
                     );
                 } catch (Exception $e) {
-                    throw new PluginsArchiverException($e->getMessage() . " - in plugin $pluginName", $e->getCode(), $e);
+                    throw new PluginsArchiverException($e->getMessage() . " - in plugin $pluginName with trace: " . $e->getTraceAsString(), $e->getCode(), $e);
```
diosmosis, line 183: Note: this should be handled when printing out an exception via `ExceptionToTextProcessor::getWholeBacktrace`

core/DataTable.php
```diff
@@ -1894,7 +1894,7 @@ protected function aggregateRowWithLabel(Row $row, $columnAggregationOps)
     {
         $labelToLookFor = $row->getColumn('label');
         if ($labelToLookFor === false) {
-            throw new Exception("Label column not found in the table to add in addDataTable()");
+            throw new Exception("Label column not found in the table to add in addDataTable() for row " . var_export($row, 1));
```
diosmosis, line 1897: This happened in the GA importer, when debugging I found  it would be more useful to have archiver params here, ie idSite/period/date and maybe report name. What do you think?

tsteur: That sounds good @diosmosis 

btw is that bug issue fixed in the importer? I think the customer might have actually the issue with the importer

diosmosis: If it's the same bug, then yes, it was an issue in the Referrers.getWebsites report, one of them will have a blank label causing period aggregation to fail this way. Can be fixed by re-importing the specific day(s).

Do you want to add the debug code or should I?

tsteur:
@diosmosis added some code for this. Is this what you meant?

core/DataTable.php
```diff
@@ -1894,7 +1895,12 @@ protected function aggregateRowWithLabel(Row $row, $columnAggregationOps)
     {
         $labelToLookFor = $row->getColumn('label');
         if ($labelToLookFor === false) {
-            throw new Exception("Label column not found in the table to add in addDataTable() for row " . var_export($row, 1));
+            $message = sprintf("Label column not found in the table to add in addDataTable(). idSite: %s, Period: %s, Row: %s",
+                DataTableFactory::getSiteIdFromMetadata($this),
+                DataTableFactory::getPeriodStringFromMetadata($this),
+                var_export($row, 1)
```
diosmosis, line 1901: This data is added after querying, I don't think it will be available at this point in the Archiver. I haven't tested though, just saw this.

diosmosis: I think it's needed to get the archive processor params being used, perhaps throwing a custom exception, then catching and decorating it would work. Just thinking off the top of my head.

tsteur:
@diosmosis removed that information again. I don't want to spend too much time on this as we don't have the problem anymore and we never really before had that issue. Getting data from archiveProcessor would require a lot of changes by the looks.

diosmosis:
👍 looks good, will merge if build passes

PR #16389, opened 2020-09-04
Make sure the same 2FA token cannot be used twice

tsteur:


tsteur:
i'll fix the randomly failing test  https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/42269/Diagnostics_page.png in a different PR

PR #16402, opened 2020-09-07
remove same previousPage function

eknis:
This is a small fix.
Since the same one is defined, I deleted it.

https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js#L106-L123


tsteur:
Thank you @eknis

PR #16412, opened 2020-09-09
Remove visit standard length from tracker

tsteur:
This was actually not needed because a ping doesn't extend the visit anyway and it makes things only more complicated when someone has a custom visit length. As discussed with Matt refs https://github.com/matomo-org/developer-documentation/pull/382


tsteur:
build js

PR #16414, opened 2020-09-10
Fix DB column names in upgrade script

tsteur:


tsteur:
fyi @diosmosis will merge this

PR #16416, opened 2020-09-10
Make sure to update plugin information before disabling plugins

tsteur:
I upgraded 3.14.1 to Matomo 4.

Somehow Custom Variables was disabled afterwards. Then noticed:

* We update core 
* We update plugins in a different request `action=oneClickUpdatePartTwo`
* Then we check if there are incompatible plugins but I'm not sure if ever the updated plugin information is ever reloaded after updating the plugins in the previous action?

I'm not sure though. According to this theory I would have also expected that other plugins would have been disabled that were updated through the marketplace. Not sure how it worked for the marketplace plugins but not this one. There must be maybe some extra logic somewhere.

I think this used to work because when the marketplace plugin was updated, it used to reload the plugin see https://github.com/matomo-org/matomo/blob/3.14.1-rc1/plugins/CorePluginsAdmin/PluginInstaller.php#L73-L77

This used to be done in the same request. But since moving this out into a second request, the information is no longer updated. I think this is a regression from https://github.com/matomo-org/matomo/pull/15770

### DIFF
--- a/plugins/CoreUpdater/Updater.php
+++ b/plugins/CoreUpdater/Updater.php
@@ -298,6 +298,12 @@ class Updater
 
     private function disableIncompatiblePlugins($version)
     {
+        $pluginManager = PluginManager::getInstance();
+        $plugins = $pluginManager->getLoadedPlugins();
+        foreach ($plugins as $plugin) {
+            $plugin->reloadPluginInformation();
+        }
+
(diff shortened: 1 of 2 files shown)


diosmosis (approved):
LGTM

tsteur:
tests look good, will merge and also apply to Matomo 4

PR #16419, opened 2020-09-10
Simplify updater further should climulti not work

tsteur:
There should be actually no reason to go through the dispatch process and instead we can directly call the method. This prevents various events being triggered during frontcontroller dispatch which may involve the API proxy etc.

@diosmosis can you have a look at this as well?


tsteur:
@diosmosis fyi I'll merge this. Tested it locally and looks good.

PR #16420, opened 2020-09-10
add documentation to the GetLanguage and GetLanguageCode reports

Findus23:
see https://forum.matomo.org/t/browser-language-language-code/38540

It seems like those two reports never got a documentation. If someone has an idea on how to improve the wording, just comment.

### DIFF
--- a/plugins/UserLanguage/lang/en.json
+++ b/plugins/UserLanguage/lang/en.json
@@ -3,7 +3,7 @@
         "BrowserLanguage": "Browser language",
         "LanguageCode": "Language code",
         "PluginDescription": "Reports the language used by your visitors' browsers.",
-        "getLanguageDocumentation": "This report shows which Language the visitor's browsers are using. (e.g. \"English\")",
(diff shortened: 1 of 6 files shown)


tsteur:
@Findus23 could you update the failed system tests where the new documentation was added? We can otherwise also do it if you're busy

Findus23:
@tsteur I waited with updating the tests in case someone has a better wording.

plugins/UserLanguage/lang/en.json
```diff
@@ -2,6 +2,8 @@
     "UserLanguage": {
         "BrowserLanguage": "Browser language",
         "LanguageCode": "Language code",
-        "PluginDescription": "Reports the language used by your visitors' browsers."
+        "PluginDescription": "Reports the language used by your visitors' browsers.",
+        "getLanguageDocumentation": "This report shows which Language the visitor's browsers are using. (e.g. \"English\")",
```
tsteur, line 6: should language be lower case @Findus23 (also below)? Looks otherwise good

Findus23:
I think (but am not sure) that all tests have been fixed.

PR #16423, opened 2020-09-10
Make sure to initialize a session for the anonymous user

tsteur:
What happened is that every time an anonymous user would view a widget or report in the Matomo UI it would log a failed authentication (for brute force check) and eventually the anonymous user would be blocked. This is because of the code here: https://github.com/matomo-org/matomo/blob/4.0.0-a1/core/Access.php#L171-L174 

The session auth always thinks that the session is expired because for anonymous user we had actually never started a session. When someone logs in, we would start the session in the login controller in https://github.com/matomo-org/matomo/blob/4.0.0-a1/plugins/Login/Controller.php#L309 but for anonymous users there is no log in.

Therefore initialising the session now when needed in the frontcontroller.

### DIFF
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -21,6 +21,7 @@ use Piwik\Http\ControllerResolver;
 use Piwik\Http\Router;
 use Piwik\Plugins\CoreAdminHome\CustomLogo;
 use Piwik\Session\SessionAuth;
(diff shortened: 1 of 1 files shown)


diosmosis:
Will this be an issue if multiple people end up using the same session simultaneously? (Or will they use the same session simultaneously?)

PR #16425, opened 2020-09-11
disable generate:travis-yml when development mode is not enabled

Findus23:
https://github.com/johsin18/DevicePixelRatioMatomoPlugin/issues/1#issuecomment-690887786

This makes the `generate:travis-yml` command only available when development mode is enabled just like all other development-related commands.

### DIFF
--- a/plugins/TestRunner/Commands/GenerateTravisYmlFile.php
+++ b/plugins/TestRunner/Commands/GenerateTravisYmlFile.php
@@ -8,6 +8,7 @@
  */
 namespace Piwik\Plugins\TestRunner\Commands;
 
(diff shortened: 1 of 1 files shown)


tsteur:
I'm thinking this might break plugin tests. Because they would execute this command to compare if the generated travis yml is up to date and error if it's outdated. I don't think travis has development mode enabled?

tsteur:
@Findus23 is this one important? if we merge this, then run some plugin tests in travis and see if it still works. might not be worth the effort since people would usually not use this command?

Findus23:
Makes sense. I guess it is easier to just ignore the command.

PR #16428, opened 2020-09-11
split disabled PHP functions to not break table layout

Findus23:
grafik


tsteur:
👍  Looks good @Findus23 only need to update the screenshot test

tsteur (approved):

Findus23:
I see no UI changes that are caused by my change. https://builds-artifacts.matomo.org/matomo-org/matomo/info-system-check-width/42399/Diagnostics_page.png only changed the number of tables.

PR #16442, opened 2020-09-15
Add dimension "Action URL"

tsteur:
refs DEV-1983

Note that this would need to be used usually with a segment filter because it would also include non-action URLs technically such as content impression URLs but content impressions/interactions aren't actions.

### DIFF
--- a/plugins/Actions/Columns/ActionUrl.php
+++ b/plugins/Actions/Columns/ActionUrl.php
@@ -9,6 +9,7 @@
 namespace Piwik\Plugins\Actions\Columns;
 
 use Piwik\Columns\DimensionSegmentFactory;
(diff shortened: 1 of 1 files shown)


mattab:
Would it be possible to add a description for the dimension, that explains exactly what data it would return? (and mention that it `includes non-action URLs such as content impression URLs (although content impressions/interactions aren't actions).`)

tsteur:
@mattab no AFAIK it's not possible to provide a description for it. Happy to close the PR without merging if needed

mattab:
Ok, forgot it wasn't a feature :see_no_evil: 

Definitely it's worth reviewing & merging :+1:

sgiehl (approved):

PR #16446, opened 2020-09-16
Do not index the opt out iframe in search engines

tsteur:
As suggested by a user. There shouldn't be a need to index it


mattab:
LGTM

PR #16460, opened 2020-09-22
Moved search translation string from the Feedback plugin to the CoreHome one

### LINKED ISSUE #15385: String not loaded in javascript while Feedback plugin is disabled
I've disabled _Feedback core plugin_, when I try searching websites in the upper left box, Matomo can't load translation string about _SearchOnMatomo_.

Screenshot_2020-01-13 All Websites dashboard - 2020-01-13 - Web Analytics Reports - Matomo

Should be hidden the whole _Help Resources_ section if the plugin is disabled? Or should the translation string reside in another plugin? Can't say 🤔

nina-py:
Fixes #15385.

### DIFF
--- a/plugins/CoreHome/CoreHome.php
+++ b/plugins/CoreHome/CoreHome.php
@@ -485,5 +485,6 @@ class CoreHome extends \Piwik\Plugin
         $translationKeys[] = 'General_Custom';
         $translationKeys[] = 'General_PreviousPeriod';
         $translationKeys[] = 'General_PreviousYear';
+        $translationKeys[] = 'Feedback_SearchOnMatomo';
     }
 }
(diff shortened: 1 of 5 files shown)


tsteur:
Thanks for creating the PR @nina-py 

Do you mind also moving the actual translation from `plugins/Feedback/lang/en.json` to `plugins/CoreHome/lang/en.json` and then we use `CoreHome_Feedback` as translation key?

This will ensure the translation key will work should feedback not be enabled.

nina-py:
@tsteur, done - thanks for pointing this out. Also updated the reference to the search string in `plugins/CoreHome/angularjs/quick-access/quick-access.directive.html` so that it works after these changes :).

tsteur:
Thanks @nina-py very appreciated

PR #16463, opened 2020-09-23
Help user configure correct GeoIP DB url by checking for protocol

tsteur:


plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -406,7 +406,9 @@ public static function setUpdaterOptions($options)
 
             $url = $options[$optionKey];
             $url = self::removeDateFromUrl($url);
-
+            if (!empty($url) && strpos(Common::mb_strtolower($url), 'https://') !== 0 && strpos(Common::mb_strtolower($url), 'http://') !== 0) {
```
sgiehl, line 409: Guess we could also use `UrlHelper::isLookLikeUrl` instead of checking for http & https. Even though it might be a non common use case, it could be possible someone wants to use an ftp source as update url maybe 🤔

sgiehl (approved):

PR #16469, opened 2020-09-24
Fix percent symbol fails JavaScript Matomo code

### LINKED ISSUE #16464: Percent symbol fails JavaScript matomo code
If I use encoded percent symbol in campaign name:
?pk_campaign=qwerty2%25-
It generates error in matomo.js:

-----------------------------
```
matomo.js:26 Uncaught URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at f (matomo.js:26)
    at ct (matomo.js:51)
    at b0 (matomo.js:54)
    at matomo.js:70
    at ch (matomo.js:59)
    at Q.trackPageView (matomo.js:70)
    at af (matomo.js:23)
    at ad (matomo.js:74)
    at Object.addTracker (matomo.js:76)
```

tsteur:
fix https://github.com/matomo-org/matomo/issues/16464

eg `https://matomo.org/product-features/?pk_campaign=qwerty2%25-`
results in `results[1] = 'qwerty2%-` which then fails using decodeWrapper but works using safeDecodeWrapper as it falls back to unescape.


tsteur:
build js

sgiehl (approved):
Would it make sense to add or adjust a test for that case?
Otherwise looks good.

tsteur:
Yes definitely 👍  added one

PR #16473, opened 2020-09-25
 Manage Dimensions: Sorting by ID is not numeric

### LINKED ISSUE #16155: Manage Dimensions: Sorting by ID is not numeric
cd-sorting

nina-py:
For some reason custom dimension ids are passed around as strings, and there is at least one place in the plugin JS code where they are converted to integers as needed - perhaps there is a good reason to keep them as strings almost everywhere else. According to AngularJS documentation, a getter function or an expression can be used instead of a field to order by, so I have added a .parseInt() method so that the ids are sorted properly. The performance hit of that is minimal given there won't be hundreds, or even dozens of custom dimensions in a typical Matomo setup.

fixes https://github.com/matomo-org/matomo/issues/16155


tsteur:
Cheers @nina-py very appreciated 👍

nina-py:
No worries @tsteur!

PR #16475, opened 2020-09-26
Replaced 'Piwik' with 'Matomo' in console output when creating extra Custom Dimensions

nina-py:
While adding extra custom dimensions to reproduce #16155, I noticed that the output from the console command still says "Your Piwik..." instead of "Your Matomo...". Updated the copy both in the output and in the corresponding tests.


nina-py:
@tsteur, just a quick question - for a straightforward fix like this that is not directly related to the issue I was working on, is a pull request enough or should I open an issue as well?

tsteur:
Cheers again @nina-py  A separate issue is not needed 👍

nina-py:
Thanks @tsteur!

PR #16480, opened 2020-09-28
The value for session.gc_probability now comes from the config files

### LINKED ISSUE #15274: Don't set session.gc_probality=1, instead read value from config
https://github.com/matomo-org/matomo/blob/76102d2c9b21cd7317ed9d4e6ea7879f83b01474/core/Session.php#L134

As in the code snippet above setting
`session.gc_probability=1` in general causes problems for different environments. For example our session handler is redis and session save_path for webserver is /var/run/redis.sock
In this case automatic garbage cleanup will attempt to cleanup a directory to which it does not have access to, For example, opendir(/var/lib/php/session) failed: permission denied fails in my case.
As a result I see errors like
ERROR Piwik\Session[2019-12-12 09:14:06 UTC] [2267b] Unable to start session
Occasionally which are frequent but hard to reproduce.

Many systems do not give permission to session directory because they have some kind of cron to periodically clear expired session files. Just the php [recommendation](https://www.php.net/manual/en/function.session-gc.php)
When you are using redis for php sessions there is no need for garbage collection since TTL will take care of removing old sessions. Hence session.gc_probability should be zero.

Now, this concerned me because [link to configure redis for matomo sessions ](https://matomo.org/faq/how-to/faq_20521/)

Should we 
1) update code base with something like 
```
@ini_set(Config->getInstance()->General['session_gc_probability']);
```

2.) Just update the [FAQ](https://matomo.org/faq/how-to/faq_20521/) link with help..(I don't know what help. How do we tell matomo not to force gc_probability to 1? Is there an event? Can I do Dependency injection?)

3.) leave it alone for now

nina-py:
Fixes #15274. Added `session_gc_probability` to the global config under [General]; the Session class now uses the config value instead of a hard-coded value.

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -394,6 +394,9 @@ hash_algorithm = whirlpool
 ; it is recommended for security reasons to always use Matomo over https
 force_ssl = 0
 
+; Session garbage collection on (as on some operating systems, i.e. Debian, it may be off by default)
+session_gc_probability = 1
+
 ; (DEPRECATED) has no effect
 login_cookie_name = matomo_auth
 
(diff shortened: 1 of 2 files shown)


diosmosis (approved):
LGTM, thanks @nina-py for another contribution!

PR #16490, opened 2020-09-29
Optimise Matomo 4 updates

tsteur:
* Fix regression in Matomo 4 that dimension updates where no longer shown see https://github.com/matomo-org/matomo/pull/16491
* Have quick required schema changes first so UI works mostly (eg user and invalidations table)
* Group more console updates next to each other by reorganising plugin activates etc
* Make sure other dimension updates are executed right away so they don't show a second update screen like below

image


core/Updates/4.0.0-b1.php
```diff
@@ -53,9 +54,6 @@ public function getMigrations(Updater $updater)
         $columnsToAdd = [];
 
         $migrations = [];
-        $migrations[] = $this->migration->db->changeColumnType('log_action', 'name', 'VARCHAR(4096)');
```
tsteur, line 56: we should avoid slow updates in the beginning and have some required faster ones first

core/Updates/4.0.0-b2.php
```diff
@@ -1,96 +0,0 @@
...
-        }
-
-        return $migrations;
-    }
-
-    public function doUpdate(Updater $updater)
-    {
-        $updater->executeMigrations(__FILE__, $this->getMigrations($updater));
-
-        if ($this->usesGeoIpLegacyLocationProvider()) {
```
tsteur, line 78: this was already present in b1 upgrade

core/Updates/4.0.0-b1.php
```diff
@@ -211,7 +241,23 @@ public function getMigrations(Updater $updater)
             $migrations[] = $this->migration->config->set('mail', 'type', 'Cram-md5');
         }
 
+        // keep piwik_ignore for existing  installs
+        $migrations[] = $this->migration->config->set('Tracker', 'ignore_visits_cookie_name', 'piwik_ignore');
+
         $migrations[] = $this->migration->plugin->activate('PagePerformance');
```
tsteur, line 247: grouping these config and plugin console updates together

core/Updates/4.0.0-b1.php
```diff
@@ -85,10 +83,42 @@ public function getMigrations(Updater $updater)
...
+        ], ['idinvalidation']);
+
+        $migrations[] = $this->migration->db->addIndex('archive_invalidations', ['idsite', 'date1', 'period'], 'index_idsite_dates_period_name');
+
         $migrations[] = $this->migration->db->dropColumn('user', 'alias');
         $migrations[] = $this->migration->db->dropColumn('user', 'token_auth');
+        $migrations[] = $this->migration->db->changeColumnType('log_action', 'name', 'VARCHAR(4096)');
+        $migrations[] = $this->migration->db->changeColumnType('log_conversion', 'url', 'VARCHAR(4096)');
+        $migrations[] = $this->migration->db->changeColumn('log_link_visit_action', 'interaction_position', 'pageview_position', 'MEDIUMINT UNSIGNED DEFAULT NULL');
+        $migrations[] = $this->migration->db->changeColumnTypes('log_visit', array(
```
tsteur, line 109: these dimension updates were missing before

core/Updates/4.0.0-b1.php
```diff
@@ -85,10 +83,42 @@ public function getMigrations(Updater $updater)
...
+            'config_browser_name' => 'VARCHAR(40) NULL',
+            'referer_name' => 'VARCHAR(255) NULL',
+            'visit_total_interactions' => 'MEDIUMINT UNSIGNED DEFAULT 0',
+        ));
 
-        /** APP SPECIFIC TOKEN END */
+        $migrations[] = $this->migration->db->sql("UPDATE `" . Common::prefixTable('option')
+            . "` SET option_value = 'VARCHAR(40) NULL1' WHERE option_name= 'version_log_visit.config_browser_name'");
```
tsteur, line 116: TODO still needs an update for version_log_link_visit_action.pageview_position

core/Columns/Updater.php
```diff
@@ -53,6 +53,16 @@ public function __construct(array $visitDimensions = null, array $actionDimensio
         $this->conversionDimensions = $conversionDimensions;
     }
 
+    /**
+     * @param PiwikUpdater $updater
+     * @return Migration[]
+     * @api
+     */
+    public function getMigrations(PiwikUpdater $updater)
```
tsteur, line 61: Fixes dimension updates were no longer executed

plugins/Actions/Columns/PageViewPosition.php
```diff
@@ -18,7 +18,6 @@
 class PageViewPosition extends ActionDimension
 {
     protected $columnName = 'pageview_position';
-    protected $columnType = 'MEDIUMINT UNSIGNED DEFAULT NULL';
```
tsteur, line 21: I had to move this into the schema. Otherwise it would generate an update like this:

```sql
ALTER TABLE `matomo_log_link_visit_action` ADD COLUMN `pageview_position ` MEDIUMINT UNSIGNED DEFAULT NULL;
```

That's because Matomo thinks it is a new dimension and doesn't realise the dimension was actually renamed. This `add column` alone wouldn't be a problem since the column would already exist and the failure would be ignored. However, it becomes a problem when other dimensions are also changed and the generated SQL looks like this:

```sql
ALTER TABLE `matomo_log_link_visit_action` ADD COLUMN `pageview_position ` MEDIUMINT UNSIGNED DEFAULT NULL, ADD COLUMN `other_dimension ` MEDIUMINT UNSIGNED DEFAULT NULL;;
```

If then the column `pageview_position` already exists, the SQL would fail and it would not add the second column anymore.

tsteur:
Updated the PR so things should work now

core/Updates/4.0.0-b1.php
```diff
@@ -106,10 +128,6 @@ public function getMigrations(Updater $updater)
         // Prepare all installed tables for utf8mb4 conversions. e.g. make some indexed fields smaller so they don't exceed the maximum key length
         $allTables = DbHelper::getTablesInstalled();
 
-        $migrations[] = $this->migration->db->changeColumnType('session', 'id', 'VARCHAR(191)');
```
tsteur, line 109: Moved those further up since they are quick to execute

PR #16491, opened 2020-09-29
Fix when there are dimension updates they aren't shown in the update screen

tsteur:
To test it change eg the option_value of the `option_name=version_log_visit.referer_type` entry. Then reload Matomo. Without this update the SQL queries aren't shown.


tsteur:
will put a fix for this in https://github.com/matomo-org/matomo/pull/16490 as it requires more changes to avoid duplicate executing of updates etc

PR #16493, opened 2020-09-29
Improve style of textareas in plugin settings

### LINKED ISSUE #16410: Styling problem when two textareas are shown after another in brute force settings
image

sgiehl:
fixes #16410


tsteur:
@sgiehl I wonder if we could for example set a `min-height` to make it more clear that it's a textarea and several rows can be entered? It's otherwise not quite clear as it looks like a text field

image

Like a min height of 100px should do:

image

sgiehl:
added a min height of 100px

PR #16494, opened 2020-09-29
Use new parameters in campaign builder

### LINKED ISSUE #16433: Campaign builder should use new campaign parameters
Should use `mtm_*` instead of `pk_`.

This applies to the in app campaign builder in Matomo app. Not on the website

sgiehl:
fixes #16433

### DIFF
--- a/plugins/Referrers/angularjs/campaign-builder/campaign-builder.controller.js
+++ b/plugins/Referrers/angularjs/campaign-builder/campaign-builder.controller.js
@@ -48,22 +48,22 @@
                 this.generatedUrl += '?';
             }
 
-            this.generatedUrl += 'pk_campaign='+campaignName;
+            this.generatedUrl += 'mtm_campaign='+campaignName;
 
             if (this.campaignKeyword) {
-                this.generatedUrl += '&pk_kwd='+encodeURIComponent(this.campaignKeyword.trim());
+                this.generatedUrl += '&mtm_kwd='+encodeURIComponent(this.campaignKeyword.trim());
             }
 
             if (this.campaignSource) {
-                this.generatedUrl += '&pk_source='+encodeURIComponent(this.campaignSource.trim());
+                this.generatedUrl += '&mtm_source='+encodeURIComponent(this.campaignSource.trim());
             }
 
             if (this.campaignMedium) {
-                this.generatedUrl += '&pk_medium='+encodeURIComponent(this.campaignMedium.trim());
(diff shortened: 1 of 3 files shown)


sanchezzzhak:
encodeURIComponent not escape special chars  `!'()*`
helper method safe escape value
```
function buildQueryEscape(param) {
  return encodeURIComponent(param).replace(/[!'()*]/g, x => `%${x.charCodeAt(0).toString(16).toUpperCase()}`);
}
```
To create a key and value string, I use the equivalent of http_query_builder, which is found in PHP
```
function buildQuery(queryData, numericPrefix = null, argSeparator = '&', tempKey = null) {
  if (queryData === null) {
	return '';
  }
  
  let query = Object.keys(queryData).map(k => {
	let res;
	let key = k;
	if (tempKey) {
	  key = tempKey + '[' + key + ']';
	}
	if (typeof queryData[k] === 'object') {
	  res = buildQuery(queryData[k], null, argSeparator, key);
	} else {
	  if (numericPrefix) {
	    let isNum = !isNaN(parseFloat(key)) && isFinite(key);
		key = isNum ? numericPrefix + Number(key) : key;
	  }
	  if (queryData[k] !== '' && queryData[k]!== undefined && queryData[k]!== null) {
		let val = queryData[k];
                // convert boolean value to int
		if (val === true) {
		  val = '1';
		} else if (val === false) {
		  val = '0';
		}
		res = buildQueryEscape(key) + '=' + buildQueryEscape(val);
	  }
	}
	return res;
  });
  
  query = query.filter((w) => {
	return w !== undefined /* w !== null */;
  });
  return query.join(argSeparator).replace(/[!'()*]/g, '');
}
```
test

https://jsfiddle.net/sanchezzzhak/bs3fq7jg/9/

image

tsteur:
@sgiehl looks good 

@sanchezzzhak  thanks for this. Didn't even know. Do these characters need to be escaped in the URL? I was reading https://stackoverflow.com/questions/18251399/why-doesnt-encodeuricomponent-encode-single-quotes-apostrophes/18251730 and it may not be needed?

Note that angular will also escape it further to prevent eg any XSS.

sanchezzzhak:
@tsteur I had problems parsing get string data in nodejs on special characters
then I wondered how php does it
https://3v4l.org/cvjla

tsteur:
Thanks @sanchezzzhak   For now we only rename the parameters to fix the particular issue then we could in a separate PR apply the other fix if you're maybe keen on creating one?

sanchezzzhak:
@tsteur I'm currently busy adding device definitions. Maybe when I have nothing to do.

PR #16498, opened 2020-09-30
Make JS tracker work in strict mode

### LINKED ISSUE #14791: Support piwik in strict mode?
as I am using es6 module and import piwik.js via npm package, every dependency are in js strict mode
It will show error `ReferenceError: JSON_PIWIK is not defined`

Can we wrap all global variables with `window.`? It seems can fix the issue easily, I am ok to open PR to do it

tsteur:
fix https://github.com/matomo-org/matomo/issues/14791


tsteur:
build js

tsteur:
build js

Issue #16161, opened 2020-06-30
New visitor ID generated per action in iframes (SameSite cookie issue)


promediadd:
## Summary

- Tracking inside an iframe when that iframe is on a third-party site is creating a new visitorId for every page visit/action taken by a visitor inside that iframe
- When the site you want to track with Matomo is embedded in an iframe on a third-party site, the [`getCookie()` function](https://github.com/matomo-org/matomo/blob/82d8267b5ecf163e4adfbf4d4dc705b78bf501a5/js/piwik.js#L2409-L2418) cannot find the visitorId cookie inside of the `documentAlias.cookie` string.
- In our (limited) testing, changing the `SameSite` param in [`setCookie()` function](https://github.com/matomo-org/matomo/blob/82d8267b5ecf163e4adfbf4d4dc705b78bf501a5/js/piwik.js#L2385-L2404) from `SameSite=Lax` to `SameSite=None` resolves the issue on most browsers (iOS Safari excluded) (i.e. the `documentAlias.cookie` string now contains our visitorId cookie and things work as expected)

## Replicating the issue

- We first noticed this issue because we saw a large spike in traffic/actions in our Matomo dashboards for a particular site
- We determined this site was being `iframe`d into a third-party site and any visits occuring inside this iframe were being tracked as new visitors for every page visit and action taken (blowing out the tracking statistics)
- To replicate the bug and test/prove our theory we created a test page and set the tracking JS to console log the visitorId every 5 seconds:
```
setInterval(function(){
    _paq.push([function () {
        var piwikVisitor = this.getVisitorId();
        console.log('the visitor ID is: ' + piwikVisitor)
    }]);
},5000)
```
- We then opened up the test page **without using an iframe** and we saw the same visitorId being spat out every 5 seconds (as expected):
![image](https://user-images.githubusercontent.com/13972603/86085396-63534c00-bae2-11ea-86fb-8e21f93bf31f.png)
- We then embedded this test page **as an iframe on the same domain** and we saw the same visitorId being spat out every 5 seconds (as expected)
- We then embedded this test page **as an iframe on a third-party domain** and we saw a new visitorId being spat out every 5 seconds (as expected, but not wanted, and thus proving our theory)
![image](https://user-images.githubusercontent.com/13972603/86085654-fee4bc80-bae2-11ea-9a0d-618dc398aa65.png)

## Our "almost" fix
[...]

PR #16504, opened 2020-10-01
Implement #16161: SameSite cookie attribute can be configured for JS tracker

beger:
With this change you can now set up your JS tracking code like this:

`                    _paq.push(['setCookieSameSite', 'None']);
`

This will set the SameSite attribute for all cookies set by Matomo to the value used above. The default is still SameSite='Lax' if setCookieSameSite is not used.

Implements the enhancement discussed in https://github.com/matomo-org/matomo/issues/16161.


tsteur:
@beger thanks for creating this PR! I was wondering if we could simplify it and set automatically `SameSite=None` when `isSecure` is set? I'm suggesting this because `None` only works when also `Secure` is set see eg https://web.dev/samesite-cookies-explained/#changes-to-the-default-behavior-without-samesite AFAIK

beger:
@tsteur I think this may not be the right solution for every use case: you still may want to enforce a different SameSite policy even if your connection is secure. But I think it probably would make sense to do it the other way around and set `isSecure` to `true` automatically if `SameSite` is set to `None`. What do you think?

tsteur:
@beger that could work. And ideally it was to throw an error and/or log a message if this is called on HTTP ?

tsteur:
@beger be great if you could look at my last comment and let us know if you're still planning to work on this PR?

fealXX:
Hi @beger, we're currently trying to use the code of your PR as a fix until an release happens and found a Problem:
If you're in a scenario where your tracking-domain is my.matomo.org, you're working on othersite.org and set your Cookiedomain to .matomo.org isPossibleToSetCookieOnDomain will fail. We need to set configCookieIsSecure and configCookieSameSite in line 4349 too. 
`            function isPossibleToSetCookieOnDomain(domainToTest)
            {
                var valueToSet = 'testvalue';
                setCookie('test', valueToSet, 10000, null, domainToTest, configCookieIsSecure, configCookieSameSite); [...]`

Which leads me to another issue: This check depends on the order of operations of setCookieDomain, setSecureCookie and setCookieSameSite. If you first set the CookieDomain, then the SameSite/Secure attributes, the check fails. If you first set Secure and Samesite, the Check works fine. Maybe the cookie configuration should be moved into a single paq-Setting, along the lines of this, while keeping the old syntax for backwards compatibility:
` _paq.push(['setCookieConfig', 'my.matomo.org', 'None', true]);`

I'd be happy to help implementing this, if possible.

tsteur:
fyi if you are using one tracker using and init this trackker using `_paq.push()` then Matomo should make sure to call the methods in the right order as the order is defined in `var applyFirst = `. 

To make it fully independent would probably require quite a few more changes as it is already a general problem in Matomo which could therefore also be fixed in a separate PR later. We should document it though at least that eg `setCookiePath` should be called before `setCookieDomain` if the tracker is being initialised manually.

Adding it to `isPossibleToSetCookieOnDomain` be indeed good and respect settings there too.

beger:
Thank you @fealXX and @tsteur for implementing and reviewing this in https://github.com/matomo-org/matomo/pull/16724. I think this PR can now be closed.

tsteur:
Thanks for getting this work started @beger and contributing to this feature, very appreciated 👍

PR #16506, opened 2020-10-01
Provide an option to fully mask the IP address

### LINKED ISSUE #16377: Offer possibility to not store IP address at all
It's already possible to anonymise 3 bytes: 
image

This is however not good enough. We also need to have a setting to basically anonymise 4 bytes and therefore the entire IP address. In this case we'd for example always track `0.0.0.0`. 

Why might 3 bytes not be enough? Eg if you have only few visitors at a certain time, and you have an IP address in Matomo logged starting with eg `192.0.0.0`, then you might still be able to crosslink this with log files to get the full IP address etc.

sgiehl:
fixes #16377


plugins/PrivacyManager/IPAnonymizer.php
```diff
@@ -25,6 +26,11 @@ class IPAnonymizer
      */
     public static function applyIPMask(IP $ip, $maskLength)
     {
+        // IPv6 can't handle anonymizing 4 bytes
+        if ($maskLength == 4 && $ip instanceof IPv6 && !$ip->isMappedIPv4()) {
```
tsteur, line 30: @sgiehl wouldn't we add this maybe to the IPv6 class? (I suppose IPv4 already supports it)

sgiehl: created https://github.com/matomo-org/component-network/pull/15
We will need to publish a new release and update the component here after merging...

tsteur: 👍  left a comment in that PR. Feel free to merge the component PR afterwards, create the release and then update this PR

sgiehl: updated

tsteur:
works

PR #16507, opened 2020-10-01
Always set consent removed cookie even when cookies are disabled

tsteur:
The removed consent cookie would be a required / necessary cookie and should be stored IMO even if cookies are disabled since it is very much in the interest of the user and to have opt out work etc.


tsteur:
build js

sgiehl (approved):
Guess that makes sense....

PR #16512, opened 2020-10-02
Log warning only in developer mode

### LINKED ISSUE #16200: Remove segment warning when no start date is set before release
See https://github.com/matomo-org/matomo/pull/16172/files#diff-c7ecf8d06820af52d6fae72297c6a9afR314

We should only trigger the warning if development mode is enabled.

Until the first few RCs we always show it so it's more visible. Alternatively, we could simply require a date to be always set and directly throw an exception in `Segment::getSelectQuery()` when no start or end date is set.

fyi @sgiehl

sgiehl:
fixes #16200 

Should be merged before final 4.0.0 release. So we maybe get some reports from RC users...

### DIFF
--- a/core/Segment.php
+++ b/core/Segment.php
@@ -310,8 +310,10 @@ class Segment
             ]) && !$this->isVisitSegment($segmentName);
 
         if ($requiresSubQuery && empty($this->startDate) && empty($this->endDate)) {
-            $e = new Exception();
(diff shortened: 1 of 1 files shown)


tsteur:
@sgiehl I'll merge this already now as it will still trigger the exception in development mode anyway and we'll soon release the RC

PR #16513, opened 2020-10-02
Typings and/or PHPDocs SettingsPiwik.php

tolbon:


tsteur:
Thanks @tolbon Is there any particular reason for adding the typings and things we should look out for?

tsteur:
@tolbon be great if you could have a look at previous comment

tolbon:
Hi sorry for delay, I needed offscreen.
No no really, you target PHP 7.2. Php7.2 have typings...
It’s just a no critical class like Date. I think slowly move code with typing is a good things when you have big project like this. I can understand too.
Sometimes typing make appears some « bug » where PHPDoc say nothing. Like php docs say getSalt return a string but finally getSalt in test can return null (wrong test or wrong phpDocs ?).
I am totally open for suggestions/args 👍 I don’t have your vision or roadmap, Maybe you don’t target this (typing)

tsteur (changes requested):
Left 1 comment. I think otherwise the definitions look good.

tsteur:
Cheers @tolbon

PR #16533, opened 2020-10-06
Ensure min width is applied to reports by dimension view

### LINKED ISSUE #16522: Piecharts aren't shown at full width
The Piechart visualisation doesn't render at a usable width. This has been noted on the Events reports and tested in multiple browsers. This issue is present in at least; Matomo Cloud and Matomo for WordPress.

**Appearance in Firefox/Chrome:**

ff-bug@2x

**Appearance in Safari:**

safari-events-piechart-issue@2x

sgiehl:
fixes #16522


tsteur:
created PR for Matomo 3 in https://github.com/matomo-org/matomo/pull/16537

PR #16535, opened 2020-10-06
Fix selectable metrics in pie/bar charts for Events and Contents reports

sgiehl:
When switching Events or Content reports to a pie or bar chart the metric selection in the report currently only has `Visits` as option, which is actually not even available in some of the reports.
Reason is, that selectable columns is filled with defaults if nothing is set. 

Note: This might already have been an issue in Matomo 3.x


tsteur:
@sgiehl tested it and worked. However, some system tests are failing. I suppose this is unrelated to this PR? If so feel free to merge.

Issue #16530, opened 2020-10-06
In Matomo 3 add check for max PHP version


tsteur:
Matomo 3 won't and can't support PHP 8 (eg because of PHP DI not being compatible with PHP 8 and upgrading would require higher PHP version plus break compatibility for plugins). 

Therefore we need to add a check to `testMinimumPhpVersion.php` for max supported PHP version. 

If the PHP version is < `8.0.0` then we show a notice stating that Matomo 3 is not compatible with PHP 8 and that they should upgrade to Matomo 4. I suppose in `version_compare` as version number we should specify `8.0.0-dev` for the check just in case someone is using PHP 8 alpha, beta or RC?

@Findus23 I think you were running Matomo with PHP 8. Do you remember if it was mostly working?

Findus23:
See also https://github.com/matomo-org/matomo/issues/16183

If I remember correctly, the Matomo application is working mostly fine, but if people have enabled PHP errors to be displayed in the output it breaks Matomo (as Deprecation warnings are shown before the output starts).

So I assume it should work well enough that people are able display the application and see the warning there or during the installer.

tsteur:
So ideally we'd test it in a PHP 8 environment. I see in https://github.com/matomo-org/matomo/pull/16536 there was already a PR made which me be good enough for now but ideally additionally we test using a PHP 8 installation if it was any usable/working then we could show the message differently. Plus in the installer.

We should test:
* Does basic tracking work
* Does basic reporting work
* Does basic archiving work

If eg tracking still works we could remove #16536 again and only show a message in the UI so data will be still connected while using PHP 8.

PR #16536, opened 2020-10-06
Check for PHP version above 8.0.0

luggesexe:
Issue #16530

### DIFF
--- a/core/testMinimumPhpVersion.php
+++ b/core/testMinimumPhpVersion.php
@@ -22,14 +22,21 @@ $piwik_errorMessage = '';
 // 3) composer.json (in two places)
 // 4) tests/PHPUnit/Integration/ReleaseCheckListTest.php
 $piwik_minimumPHPVersion = '7.2.5';
+$piwik_maximumPHPVersion = '8.0.0';
 $piwik_currentPHPVersion = PHP_VERSION;
 $minimumPhpInvalid = version_compare($piwik_minimumPHPVersion, $piwik_currentPHPVersion) > 0;
+$maximumPHPInvalid = version_compare($piwik_maximumPHPVersion, $piwik_currentPHPVersion) >= 0;
 if ($minimumPhpInvalid) {
     $piwik_errorMessage .= "<p><strong>To run Matomo you need at least PHP version $piwik_minimumPHPVersion</strong></p>
 				<p>Unfortunately it seems your webserver is using PHP version $piwik_currentPHPVersion. </p>
 				<p>Please try to update your PHP version, Matomo is really worth it! Nowadays most web hosts
 				support PHP $piwik_minimumPHPVersion.</p>";
 } else {
+    if ($maximumPHPInvalid) {
+        $piwik_errorMessage .= "<p><b>Matomo 3 is not compatible with PHP 8.</b></p>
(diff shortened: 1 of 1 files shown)


core/testMinimumPhpVersion.php
```diff
@@ -22,14 +22,21 @@
 // 3) composer.json (in two places)
 // 4) tests/PHPUnit/Integration/ReleaseCheckListTest.php
 $piwik_minimumPHPVersion = '7.2.5';
+$piwik_maximumPHPVersion = '8.0.0';
 $piwik_currentPHPVersion = PHP_VERSION;
 $minimumPhpInvalid = version_compare($piwik_minimumPHPVersion, $piwik_currentPHPVersion) > 0;
+$maximumPHPInvalid = version_compare($piwik_maximumPHPVersion, $piwik_currentPHPVersion) >= 0;
```
luggesexe, line 28: Wait... no 
`$maximumPHPInvalid = version_compare($piwik_maximumPHPVersion, $piwik_currentPHPVersion) <= 0;`

core/testMinimumPhpVersion.php
```diff
@@ -22,14 +22,21 @@
 // 3) composer.json (in two places)
 // 4) tests/PHPUnit/Integration/ReleaseCheckListTest.php
 $piwik_minimumPHPVersion = '7.2.5';
+$piwik_maximumPHPVersion = '8.0.0';
```
tsteur, line 25: Cheers @luggesexe would it need to be `8.0.0-dev`? so it also matches PHP 8 beta versions etc

Findus23:
As this is a message that is probably shown to non-technical users (e.g. on shared hosting setting up Matomo for the first time), would it maybe make sense to make it translatable?
Or is this running before the localisation system has been initialized?

luggesexe:
> ~As this is a message that is probably shown to non-technical users (e.g. on shared hosting setting up Matomo for the first time), would it maybe make sense to make it translatable? Or is this running before the localisation system has been initialized?~
> The code obviously has to run before everything else, so that wouldn't work.

Yes, it runs before anything else, therefore it wouldn't be possible to translate it. Maybe adding a sentence referencing a help article?

core/testMinimumPhpVersion.php
Findus23: I think that syntax is really confusing, so I'd go with the alternative one: (I hope I didn't mix it up)
https://www.php.net/manual/en/function.version-compare.php
```php
$maximumPHPInvalid = version_compare($piwik_maximumPHPVersion, $piwik_currentPHPVersion, '>=');

luggesexe: I think it should be 
`$maxVersionInvalid = version_compare($maxVersion, PHP_VERSION, '<=');` as we are checking if the version is below 8.0.0 rather than above. Otherwise this option should clear things up, yes.

core/testMinimumPhpVersion.php
luggesexe: Yes this should work. I will add it.

core/testMinimumPhpVersion.php
Findus23: You are of course right.

sgiehl:
closing in favor of #16539. Also the source repo is not available anymore...

PR #16538, opened 2020-10-06
don't mention PHP5 in messages

Findus23:
When looking at https://github.com/matomo-org/matomo/pull/16536 I noticed that php5 was still mentioned in a few places, so I replaced them with a general `php`.

### DIFF
--- a/core/CliMulti/CliPhp.php
+++ b/core/CliMulti/CliPhp.php
@@ -50,10 +50,6 @@ class CliPhp
         }
 
         if (!$this->isValidPhpType($bin)) {
-            $bin = shell_exec('which php5');
-        }
-
-        if (!$this->isValidPhpType($bin)) {
             return false;
         }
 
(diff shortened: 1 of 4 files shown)


core/CliMulti/CliPhp.php
```diff
@@ -49,10 +49,6 @@ public function findPhpBinary()
             $bin = shell_exec('which php');
         }
 
-        if (!$this->isValidPhpType($bin)) {
-            $bin = shell_exec('which php5');
```
tsteur, line 53: @Findus23 not sure if it's a thing but could do `which php7` (and maybe `which php8`)?

Findus23: I don't think php7 exists on any (Debian) system, only `php7.4` etc.
Not sure if listing all of them helps (if I understand it correctly this would matter if there was no `php` in PATH, but a `php7.4` which seems quite rare).

tsteur: 👍 not needed

sgiehl (approved):

PR #16542, opened 2020-10-07
Support removing invalidations for multiple sites at one time.

diosmosis:
Refs https://github.com/matomo-org/matomo/issues/11974


diosmosis:
FYI @tsteur will merge if build passes

Issue #16183, opened 2020-07-06
Support PHP 8


Findus23:
It is still quite a bit in the future until PHP 8 will be released, but as it is a major release removing and deprecating a lot of ancient features, one can not start early enough fixing things:

#### In Matomo:
- [x] create_function() has been removed https://github.com/matomo-org/matomo/pull/16181
- [x] core/Http.php(141): Deprecated - Required parameter $aUrl follows optional parameter $method https://github.com/matomo-org/matomo/pull/16182
https://php.watch/versions/8.0/deprecate-required-param-after-optional
- [x] #16550 `Private methods cannot be final as they are never overridden by other classes in /home/lukas/public_html/matomophp8/core/Singleton.php on line 27`

https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/core/Singleton.php#L27-L29
- [x] #16312 plugins/Actions/ArchivingHelper.php on line 374
- [x]  #16313  don't log silenced errors in php8 

#### In 3rd-party-libraries:
- [x] PHP-DI: ReflectionParameter::getClass() is deprecated
```text
[06-Jul-2020 13:56:34 UTC] PHP Deprecated:  Method ReflectionParameter::getClass() is deprecated in /home/lukas/public_html/matomophp8/vendor/php-di/php-di/src/DI/Invoker/FactoryParameterResolver.php on line 43
```
`https://github.com/php/php-src/pull/5209` 
I still don't know DI, so I can't say much about it 

Will be fixed with https://github.com/PHP-DI/PHP-DI/issues/727 and https://github.com/matomo-org/matomo/issues/15974 (PR: https://github.com/matomo-org/matomo/pull/16311)

------------------

to be continued ...

sgiehl:
> Private methods cannot be final as they are never overridden by other classes in /home/lukas/public_html/matomophp8/core/Singleton.php on line 27

Guess we could simply make the method public to prevent that error and maybe throw an Exception in the method that `clone` is not supported or something like that...

sgiehl:
Closing for now as there is nothing remaining atm.

Findus23:
Now that I can test Matomo without getting hundreds of error messages per request, I can check the rest if something remains.
I'll reopen if I find something.

PR #16550, opened 2020-10-08
Make method signature compatible with PHP 8

sgiehl:
Having that method protected should have the same effect. But is compatible with PHP 8...

refs #16183

### DIFF
--- a/core/Singleton.php
+++ b/core/Singleton.php
@@ -24,7 +24,7 @@ class Singleton
     {
     }
 
-    final private function __clone()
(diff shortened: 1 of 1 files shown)


Findus23:
@sgiehl 
Interestingly even with this merged I still get this warning with php8RC2

> [21-Oct-2020 12:55:47 UTC] PHP Warning:  Private methods cannot be final as they are never overridden by other classes in /home/lukas/public_html/matomophp8/core/Singleton.php on line 27

PR #16561, opened 2020-10-12
Adds measuarable settings to disable visits log and visitor profile

### LINKED ISSUE #16259: Allow disabling Visitor Profile feature and visits log feature
It be great if these feature could be disabled. Currently, the workaround is to disable the entire plugin. However, this also disables real time reports. Also on the cloud it is currently not possible to disable plugins so a user cannot disable these features yet (which may be needed to avoid tracking consent in some countries).

sgiehl:
fixes #16259


sgiehl:
@tsteur Do we also need a possibility to define the default value for new websites, or maybe the possibility to disable the feature in config, so it's not possible to enable it in UI?

tsteur:
Haven't looked at the PR so not sure what is implemented. I think a global system setting to disable the feature for all sites be needed. When the feature is disabled globally then it should not be present in measurable settings. 

If it's enabled, then we would show it in measurable settings.

If we use a system setting for the global flag then it'll be automatically configurable through the config file as well and would be then removed in the UI. I think that should do.

plugins/Goals/Controller.php
```diff
@@ -90,6 +92,7 @@ public function goalConversionsOverview()
         $view->conversion_rate_returning = $this->formatConversionRate($goalMetrics, 'conversion_rate_returning_visit');
         $view->conversion_rate_new = $this->formatConversionRate($goalMetrics, 'conversion_rate_new_visit');
         $view->idGoal = $idGoal;
+        $view->visitorLogEnabled = Manager::getInstance()->isPluginActivated('Live') && Live::isVisitorLogEnabled($this->idSite);
```
diosmosis, line 95: Can we check for `Manager::getInstance()->isPluginActivated('Live')` in `Live::isVisitorLogEnabled` or do the checks need to be separate?

plugins/Live/Live.php
```diff
@@ -19,6 +19,9 @@
  */
 class Live extends \Piwik\Plugin
 {
+    protected static $visitorProfileEnabled = null;
+    protected static $visitorLogEnabled = null;
+    protected static $siteIdLoaded = null;
```
diosmosis, line 24: Note: if adding static variables like this, we'd need to make sure they are unset in `Fixture::clearInMemoryCaches()` (though these won't affect any existing tests I think, but they might in the future). If using the transient cache, then nothing needs to change.

plugins/Live/API.php
```diff
@@ -189,6 +189,10 @@ public function getVisitorProfile($idSite, $visitorId = false, $segment = false,
     {
         Piwik::checkUserHasViewAccess($idSite);
 
+        if (!Live::isVisitorProfileEnabled($idSite)) {
+            throw new Exception('Visitor profile has been disabled in website settings');
+        }
```
diosmosis, line 194: Should this be translated? Since the user could see it.

diosmosis (commented):
Left a couple comments. Testing locally, everything seems to work 👍

plugins/Goals/Controller.php
sgiehl: In theory guess we could do that. But might be better if we always check if a plugin is installed & activated before using it.
Not sure if Matomo would even run correctly if the Live plugin would be deleted, but if that would work, the code would break here if we directly use a part of the plugin without checking if it's available...

plugins/Live/Live.php
sgiehl: using transient cache instead sound fine... will change that

plugins/Live/API.php
sgiehl: He should actually only see this when calling the API directly. The UI shouldn't try to load the visitor profile at all when it's disabled

plugins/Live/Live.php
sgiehl: done

plugins/Live/API.php
diosmosis: I can't remember how, but I was able to see the error in the UI a couple times. Once in widgets, once in the visitor log page. This was while switching the setting on/off in one tab and testing in another, however, so it could have been an edge case users wouldn't notice.

plugins/Goals/Controller.php
diosmosis: 👍

plugins/Live/Live.php
```diff
@@ -48,10 +51,77 @@ public function addJsGlobalVariables(&$out)
...
+        return $profileEnabled;
+    }
+
+    private static function getSettings($idSite = null)
+    {
+        if (empty($idSite)) {
+            $idSite = Common::getRequestVar('idSite', 0, 'int');
+        }
+
+        $cache = Cache::getTransientCache();
```
tsteur, line 80: fyi @sgiehl in future we probably wouldn't need to cache these settings. If so, we could add a generic transient cache here to all settings: https://github.com/matomo-org/matomo/blob/4.0.0-b3/core/Settings/Storage/Factory.php#L134 but they shouldn't be majorly slow and partially already transient cached in the backend potentially

PR #16565, opened 2020-10-13
Custom reports: show server hours in site's timezone

tsteur:
fix dev-2020

Wasn't sure where to put the method. Initially had it in the formatter but it's not formatting the hour it is converting. Also re the behaviour with period/date being used it might be actually ok to put this into a dimension class.

Once merged I'd also adjust eg in Media Analytics

### DIFF
--- a/plugins/CoreHome/Columns/VisitLastActionTime.php
+++ b/plugins/CoreHome/Columns/VisitLastActionTime.php
@@ -8,13 +8,15 @@
  */
 namespace Piwik\Plugins\CoreHome\Columns;
 
+use Piwik\Common;
 use Piwik\Date;
+use Piwik\Period;
 use Piwik\Plugin\Dimension\VisitDimension;
+use Piwik\Site;
 use Piwik\Tracker\Action;
 use Piwik\Tracker\Request;
 use Piwik\Tracker\Visitor;
 use Piwik\Metrics\Formatter;
-use Piwik\Tracker\VisitorRecognizer;
 
 require_once PIWIK_INCLUDE_PATH . '/plugins/VisitTime/functions.php';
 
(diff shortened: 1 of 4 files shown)


plugins/CoreHome/tests/Integration/Column/VisitLastActionTimeTest.php
```diff
@@ -58,6 +58,34 @@ private function makeRequest($request)
...
+        $this->assertEquals(5, $hourConverted);
+    }
+
+    public function test_convertHourToHourInSiteTimezone_WithTimezone()
+    {
+        $idSite = Fixture::createWebsite('2020-01-02 03:04:05', $ecommerce = 1, 'Site', $siteUrl = false,
+            $siteSearch = 1, $searchKeywordParameters = null,
+            $searchCategoryParameters = null, $timezone = 'Pacific/Auckland');
+        $hourConverted = VisitLastActionTime::convertHourToHourInSiteTimezone(5, $idSite);
+        $this->assertEquals(18, $hourConverted);
```
sgiehl, line 74: Just out of curiosity: Will this test fail when the time is changed due to daylight savings? Guess Auckland has summer/winter time as well, right?

tsteur: Yes it will. I reckon it'll be easy to change though then. Do you maybe know a timezone that doesn't have daylight savings? happy to change it

sgiehl: Asia shouldn't have daylight saving afaik. Guess you could use eg `Asia/Jakarta` (UTC+7)

plugins/CoreHome/Columns/VisitLastActionTime.php
```diff
@@ -35,9 +37,40 @@ class VisitLastActionTime extends VisitDimension
...
+        $date = Date::now();
+        $dateString = Common::getRequestVar('date', '', 'string');
+        $periodString = Common::getRequestVar('period', '', 'string');
+        if (!empty($dateString) && !empty($periodString)) {
+            try {
+                $date = Period\Factory::build($periodString, $dateString)->getDateStart();
+            } catch (\Exception $e) {
+                // ignore any error eg if values are wrong...
+            }
+        }
```
diosmosis, line 62: Is it needed to use the date/period if we're just getting the hour later? Could we just use a dummy date or today()?

tsteur: @diosmosis  We'll try to use the date in case of daylight savings to make sure it is being considered. If we just used today then it wouldn't always be correct (see method description)

diosmosis: I see, that's unfortunate.

PR #16566, opened 2020-10-13
Set minimal height of all textareas to 100px

### LINKED ISSUE #16560: Global list of Excluded IPs shows only one IP
I am using matomo 4.0.0-b3

I recorded 3 IPs on 3 lines in the Global list of Excluded IPs global settings. Not all IPs are visible and the is no hint they are more.

On Firefox 81.0.1 MacOS:
- When I load the page again, only the first record is visible.
- When I focus on the textarea, only the first record is visible.
- When I hit return, in the textarea all lines are visble including the empty line of the return
- When I change the hight of the textarea to `max-content`, i only see two records
- When I change the hight of the auto to `auto`, i only see two records

On Chrome 85.0.4183.121 MacOS
- When I load the page again, only two records are visible.
- When I focus on the textarea, only two records are visible.
- When I hit return, in the textarea all lines are visble including the empty line of the return
- When I change the hight of the textarea to `max-content`, i only see two records
- When I change the hight of the textarea to `auto`, i only see two ½ records which is acceptable as I have visual feedback that there could be more.

sgiehl:
To ensure all textareas fit the height of their content, we could use this javascript
```
$('textarea').each(function(){M.textareaAutoResize($(this))});
```
But actually I'm not sure where to add it properly....

fixes #16560


tsteur:
@sgiehl if I understand things correctly that would go into `plugins/CorePluginsAdmin/angularjs/form-field/form-field.directive.js`? There could be something like 

```js
if (hasUiControl(field, 'textarea')) {
    M.textareaAutoResize... 
}
```

not sure it works but this be generally speaking the correct place I reckon. If you need the DOM element I think there's also a method `whenRendered` so you can take action once it's fully rendered by angular

sgiehl:
@tsteur good hint. there was actually a autoresize in that file already, that didn't work anymore. It should work again now...

PR #16570, opened 2020-10-13
Ignore tracking requests for disabled plugins

### LINKED ISSUE #16569: Matomo may track a page view even though it was a different request when a plugin was disabled
Say you are using a plugin that sends tracking requests like Form Analytics.

If you disable this plugin, then the tracking code containing Form Analytics may still be cached in some browsers and these clients keep sending form tracking requests. Because the plugin is disabled, no action it would be found to handle this tracking request so it would fall back to tracking a page view even though it isn't a page view. This could result for example in actions being tracked with no action title but the URL where it happened. Of course it would result in more page views.

I suppose there be two ways to fix it:

1. Have the JS tracker caching set so it always verifies the version the browser has cached is the latest version. Downside: Every user needs to make sure to get caching right plus it may result in additional requests to the server causing more server load plus things get slower in the client.
2. Only track a page view when `action_name` tracking URL parameter is set. Downside: Breaking change if someone doesn't send this parameter. Would need to be solved as part of some major release.
3. Custom request parameter: Require all custom tracking plugins to send an additional parameter and if no specific action was found, ignore that request if that parameter is present.

tsteur:
fix https://github.com/matomo-org/matomo/issues/16569

Once merged, I would also implement this in our premium features.

It shouldn't be needed to mention this in our changelog since AFAIK custom actions/request processors aren't a documented/officially supported feature yet.

Already implemented this basically as a test in core for Contents, and Events plugin. So if someone was to disable the events plugin, but the standard JS tracker is still sending requests for events, then these would be ignored. Not doing it eg for actions plugin as this be usually always enabled anyway etc. Tried to implement it for heartbeat plugin but this doesn't work because it kind of relies on using the default page view action and then aborting the request. So if someone was to disable the heartbeat plugin, but still using the feature in JS this would still register page views.

We could then also create issues for other SDKs/trackers and add this there although it might not be as much of an issue there.

Possible problem: Sometimes one request might be sent for different purposes. Need to make sure in these cases we don't break anything. Eg a form view request might be sent along a page view. In this case we should not append the `ca=1` tracking parameter.


tsteur:
build js

tsteur:
There are some failing tests see eg https://travis-ci.org/github/matomo-org/matomo/jobs/735568092#L737-L847

I don't understand how they fail. They work locally and the requests failing makes no sense to me currently as not really any logic was changed. Maybe someone else has an idea what is happening there?

sgiehl:
@tsteur I've looked into the failing tests. The "problem" is, that your new tests are using the `Action::factory` without loading the `Actions` plugin. That fills the static available actions array without those actions, causing the other tests to fail. Always loading the `Actions` plugin should fix that. I'll push a fix...

tsteur:
Cheers @sgiehl  let me know if this is good to merge

sgiehl (approved):
code looks fine. As tests are all passing now, guess should be good to merge

tsteur:
Updated our premium features.

tsteur:
build js

PR #16578, opened 2020-10-16
Check if time limit function is enabled before executing it

tsteur:
In case the function is disabled


tsteur:
fyi merging this one

PR #16582, opened 2020-10-16
Fixes downloading files in chunks when not using curl

stevenseeley:
Hi Guys,

I was having some issues when downloading files as the response headers were not set when php-curl isn’t installed.

Instead of opening an issue, I decided to create a simple patch that fixes the file download bug and set the response headers correctly.

Kind regards,

Steve

### DIFF
--- a/core/Http.php
+++ b/core/Http.php
@@ -534,6 +534,10 @@ class Http
                     $status = (int)$m[2];
                 }
 
+                foreach ($http_response_header as $line) {
+                    self::parseHeaderLine($headers, $line);
+                }
+		    
                 if (!$status && $response === false) {
                     $error = error_get_last();
                     throw new \Exception($error['message']);
--- a/core/Http.php
+++ b/core/Http.php
@@ -533,16 +533,16 @@ class Http
                 if (isset($http_response_header) && preg_match('~^HTTP/(\d\.\d)\s+(\d+)(\s*.*)?~', implode("\n", $http_response_header), $m)) {
                     $status = (int)$m[2];
                 }
-
-                foreach ($http_response_header as $line) {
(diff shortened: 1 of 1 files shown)


core/Http.php
```diff
@@ -534,6 +534,10 @@ public static function sendHttpRequestBy(
                     $status = (int)$m[2];
                 }
 
+                foreach ($http_response_header as $line) {
```
tsteur, line 537: Cheers for this fix @stevenseeley 

wondering... do we maybe need to move it down a few lines so it is also used when using `fopen` or would `http_response_header` not be set in that case?

stevenseeley: Actually yes, outside of the if statement would probably be more ideal. I’ll test and update on Monday this pull.

sgiehl:
@stevenseeley Would you mind explaining your issue a bit more? Which file did you try downloading?
I would like to reproduce that locally in order to check if the changes are solving it

stevenseeley:
> @stevenseeley Would you mind explaining your issue a bit more? Which file did you try downloading?
> I would like to reproduce that locally in order to check if the changes are solving it

Hi @sgiehl, yes when using the `Http::downloadChunk` static method which is used by several of your API's (such as `downloadFreeGeoIPDB`) there is a check for content-length and if its 0 it will thrown an exception [see here](https://github.com/matomo-org/matomo/blob/4.x-dev/core/Http.php#L821-L824)

It's only set when using php-curl, for example when falling back to `fopen` this header (and all the others) is not set. I hope that makes sense!

sgiehl (approved):
Thanks @stevenseeley. Was now able to reproduce the problem locally and can confirm your changes solves the issue.

PR #16588, opened 2020-10-19
Pass string parameters to DB query instead of object

tsteur:
In Matomo for WordPress this caused the tests to fail.  I have now worked around this to ensure it won't cause any issues in Matomo for WordPress anymore (https://github.com/matomo-org/wp-matomo/commit/ffe92d138555bcea5ad4f2fcde9a2caac80b8fa5#diff-c4e75ddbbcdde3ba408b5a87e273371dd3420a5d0b90a87bb4295e81787cdc31R242-R244) when this happens again but figured it be still good to pass strings directly here just in case it somehow causes other issues somehow :)

### DIFF
--- a/core/DataAccess/Model.php
+++ b/core/DataAccess/Model.php
@@ -377,7 +377,7 @@ class Model
 
         $sql = "SELECT idarchive FROM `$numericTable` WHERE idsite = ? AND date1 = ? AND date2 = ? AND period = ? AND name = ? AND ts_archived < ? AND idarchive < ?";
 
-        $idArchives = Db::fetchAll($sql, [$params->getSite()->getId(), $dateStart, $dateEnd, $params->getPeriod()->getId(), $name, $tsArchived, $idArchive]);
+        $idArchives = Db::fetchAll($sql, [$params->getSite()->getId(), $dateStart . '', $dateEnd . '', $params->getPeriod()->getId(), $name, $tsArchived, $idArchive]);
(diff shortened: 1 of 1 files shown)


core/DataAccess/Model.php
```diff
@@ -377,7 +377,7 @@ public function deleteOlderArchives(Parameters $params, $name, $tsArchived, $idA
 
         $sql = "SELECT idarchive FROM `$numericTable` WHERE idsite = ? AND date1 = ? AND date2 = ? AND period = ? AND name = ? AND ts_archived < ? AND idarchive < ?";
 
-        $idArchives = Db::fetchAll($sql, [$params->getSite()->getId(), $dateStart, $dateEnd, $params->getPeriod()->getId(), $name, $tsArchived, $idArchive]);
+        $idArchives = Db::fetchAll($sql, [$params->getSite()->getId(), $dateStart . '', $dateEnd . '', $params->getPeriod()->getId(), $name, $tsArchived, $idArchive]);
```
diosmosis, line 380: Would it be good to use `->getDatetime()`? Otherwise good to merge.

tsteur: 👍  Good one. So maybe partially it could have been buggy before?

diosmosis: Maybe, I remember testing this and it working at least on my local setup...

tsteur: @diosmosis updated it

PR #16591, opened 2020-10-19
replace visit log with visits log

### LINKED ISSUE #16521: Use name "segmented visits log" (plural) to be consistent with "Visits log" naming
The feature is called "Visits log", and we refer to "Segmented Visits log", but currently in the product it says "Segmented Visit Log":

Screenshot from 2020-10-05 13-16-32

Expected to have an `s` appended to visit so it reads `Open Segmented Visits Log` and `Show Visits Log segmented by [...]`

sgiehl:
fixes #16521


tsteur:
Moving this to 4.1. PR looks good but we'll need to update screenshots. Can do it for 4.0 release if there's time left between RC but and the release but might not be.

tsteur:
UI tests will need to be updated before merging

Issue #16363, opened 2020-09-01
Hide features that don't work when the website is configured in cookie-less tracking


mattab:
More and more people are [disabling tracking cookies](https://matomo.org/faq/general/faq_157/) (we sometimes call it cookie-less tracking), because [no cookie consent](https://matomo.org/faq/new-to-piwik/how-do-i-use-matomo-analytics-without-consent-or-cookie-banner/) is needed then. 

When cookie-less tracking is enabled [some features don't work](https://matomo.org/faq/general/faq_156/) for example:
* Unique Visitors metric is very inaccurate
* Goals and Ecommerce conversions will be attributed to the channel used in the visit that converts
* Some reports are very inacurate: Days since last visit, Visits by visit count, Visits to Conversion, Days to Conversion

Because the features become very broken after cookie-less is enabled, people tend to get confused, and ask questions, or they may use the data not knowing it's very wrong! So it would be really useful to somehow hide these reports from users, or make it otherwise crystal clear that they shouldn't use this data. What do you think?

tsteur:
Hiding this data was also suggested in https://github.com/matomo-org/matomo/issues/13655#issuecomment-616892565 but we didn't do it since it can currently not be enforced and is only done client side. There are few more features like Cohorts, Multi Attribution, Visitor Profile, etc. Hiding this features might only be a good idea though if not also the userId feature is used.

This feature will only work when cookies can be disabled server side eg see https://github.com/matomo-org/matomo/issues/16258 .

tsteur:
Here's what @mattab and I discussed:

We archive a new metric "nb_profilable". It is the sum of how many people had cookies (or similar technology such as advertising identifier on iOS etc) enabled to identify them over several visits. If 99% or more of the visits were not "profilable" (had cookies disabled), then we assume the site is tracking cookie-less and remove certain features etc see issue description.

It basically means for every requested date range we probably also need to check if there were "profilable" visits or not. 

We don't need to expose the new metrics to the UI for now but could do this later. This can be quite useful later to be able to see what percentage basically consented to cookies etc. 

@mattab I just looked in the code and `log_visit.profilable=1` also when userId is used. This kind of makes sense because a profile for that visitor is possible. So in the small percentage where users use the userId feature then we would not disable the features mentioned in the issue description. Not sure if this would be an issue or if we would want to workaround it?

mattab:
[...]

PR #16592, opened 2020-10-19
Adds possibility to force cookie less tracking

### LINKED ISSUE #16258: Allow force disable cookies
When you want to make sure that across all sites cookies should never be used. It's otherwise incredibly difficult to ensure all sites implement it correctly.

What this feature would do is 

* Define a custom `tracker.js` in the plugin that makes ideally sure that cookies cannot be enabled in the client (eg we can overwrite some tracker methods like `tracker.setCookieConsentGiven=function(){}` etc and we can automatically call tracker.disableCookies())
* Server side we would in `Tracker/Request` never return any `visitorId`. So even if the client did send a visitorId, we would ignore it and rely on fingerprint.

Out of scope for now be removing metrics like Unique visitors that wouldn't be correct anymore.

sgiehl:
globally and/or per site

fixes #16258


sgiehl:
Should be ready for a first review. @tsteur maybe you could have a quick look if it does what you expected.
Also not sure how to write some useful tests for it 🤔

plugins/PrivacyManager/SystemSettings.php
```diff
@@ -53,12 +61,35 @@ private function createTermsAndConditionUrlSetting()
...
         return $this->makeSetting('showInEmbeddedWidgets', $default = false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
             $field->title = Piwik::translate('PrivacyManager_ShowInEmbeddedWidgets');
             $field->uiControl = FieldConfig::UI_CONTROL_CHECKBOX;
             $field->description = Piwik::translate('PrivacyManager_ShowInEmbeddedWidgetsDescription');
         });
     }
+
+    private function createForceCookielessTrackingSetting(): SystemConfigSetting
+    {
+        return $this->makeSettingManagedInConfigOnly('Tracker', 'force_cookieless_tracking', $default = false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
```
tsteur, line 75: @sgiehl can we make this a regular setting? I don't think it needs to be managed in the config specifically. Any setting can be configured in the config anyway eg in this case it would be `[PrivacyManager]force_cookieless_tracking=0/1`.

By using a normal setting, the advantage is that when configured in the config then it won't appear in the UI anymore and it cannot be overwritten. Thus it is quite enforced. 

I think we would use `makeSettingManagedInConfigOnly ` only for settings where they were initially configured in the config and then moved into the UI for BC.

plugins/PrivacyManager/PrivacyManager.php
```diff
@@ -741,4 +744,42 @@ private function shouldRenderFooterLinks(SystemSettings $settings)
 
         return false;
     }
+
+    public function shouldAddTrackerFile(&$shouldAdd, $pluginName)
+    {
+        if ($pluginName === 'PrivacyManager') {
+            // only check the config value, so no database connection needs to be established
+            $shouldAdd = !!PiwikConfig::getInstance()->Tracker['force_cookieless_tracking'];
```
tsteur, line 752: here we would instead access privacy system setting class instead of the config

plugins/PrivacyManager/tracker.js
```diff
@@ -0,0 +1,35 @@
...
+        // disable cookies and remove them when a tracker is created
+        window.Matomo.on('TrackerSetup', function(tracker) {
+            tracker.setCookieConsentGiven=function(){};
+            tracker.rememberCookieConsentGiven=function(){};
+            tracker.disableCookies();
+        });
+    }
+
+    if ('object' === typeof window.Matomo) {
+        init();
```
tsteur, line 26: Note: This feature will only work correctly if the `matomo.js / piwik.js` is writable. It won't 100% work when it is loaded separately and I reckon people would probably also not load it as a separate file and we would then rather rely on recommending people to call `disableCookies` and ignoring any cookie server side.

To be safe we could iterate over all existing `Matomo.getAsyncTrackers()` and call the setup method as well for each tracker. However, it would not work for any tracker created using `getTracker()` plus at this point it might be too late for any async tracker as they might have created cookies already. That be not a big issue though, we'll make users aware of this in the setting maybe.

plugins/PrivacyManager/SystemSettings.php
```diff
@@ -53,12 +61,35 @@ private function createTermsAndConditionUrlSetting()
...
+            $field->description = Piwik::translate('PrivacyManager_ForceCookielessTrackingDescription');
+        });
+    }
+
+    public function save()
+    {
+        parent::save();
+
+        try {
+            if (Manager::getInstance()->isPluginActivated('CustomJsTracker')) {
```
tsteur, line 87: wonder if we should call the method `updateTracker` on the plugin instance directly? Or have a new event like `CustomJsTracker.updateTracker` to trigger this logic?

plugins/PrivacyManager/PrivacyManager.php
```diff
@@ -741,4 +744,42 @@ private function shouldRenderFooterLinks(SystemSettings $settings)
...
+    /**
+     * Returns if cookie less tracking is forced globally or for the given site
+     *
+     * @param null|int $idSite
+     * @return bool
+     */
+    public static function isCookieLessTrackingForced($idSite=null)
+    {
+        try {
+            if (PiwikConfig::getInstance()->Tracker['force_cookieless_tracking']) {
```
tsteur, line 765: here we would also use the system setting instead of the config (it should be cached in tracker mode automatically)

plugins/PrivacyManager/MeasurableSettings.php
```diff
@@ -0,0 +1,38 @@
...
+        $this->forceCookielessTracking = $this->createForceCookielessTrackingSetting();
+        $this->forceCookielessTracking->setIsWritableByCurrentUser(!PiwikConfig::getInstance()->Tracker['force_cookieless_tracking']);
+    }
+
+    private function createForceCookielessTrackingSetting(): MeasurableSetting
+    {
+        return $this->makeSetting('forceCookielessTracking', $default = false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
+            $field->title = Piwik::translate('PrivacyManager_ForceCookielessTracking');
+            $field->uiControl = FieldConfig::UI_CONTROL_CHECKBOX;
+            $field->description = Piwik::translate('PrivacyManager_ForceCookielessTrackingDescription');
```
tsteur, line 35: Could we check if `matomo.js and piwik.js` is writable and if not, show a message that this feature might not work 100% and that only server side the cookie will be ignored? The check for `writable` could work like in the `TrackerJsCheck` class.

plugins/PrivacyManager/PrivacyManager.php
tsteur: I see the result is cached for tracking anyway so it doesn't need to be cached anyway in the settings class directly

core/Tracker/Request.php
```diff
@@ -713,6 +714,18 @@ protected function getCookieDomain()
...
     {
+        try {
+            if ($this->getIdSiteUnverified()) {
+                $siteAttributes = Cache::getCacheWebsiteAttributes($this->getIdSite());
+            }
+        } catch (\Exception $e) {
+            // ignore any exception so method also works if site wasn't given or doesn't exist
+        }
+
+        if (isset($siteAttributes[PrivacyManager::TRACKER_CACHE_COOKIELESS_FORCED]) && $siteAttributes[PrivacyManager::TRACKER_CACHE_COOKIELESS_FORCED]) {
```
tsteur, line 725: A userId would be still accepted and I think a forced visitor Id as well. Only the third party cookie and the `_id` parameter which comes AFAIK from the cookie would be ignored

plugins/PrivacyManager/SystemSettings.php
```diff
@@ -53,12 +61,35 @@ private function createTermsAndConditionUrlSetting()
...
             $field->description = Piwik::translate('PrivacyManager_ShowInEmbeddedWidgetsDescription');
         });
     }
+
+    private function createForceCookielessTrackingSetting(): SystemConfigSetting
+    {
+        return $this->makeSettingManagedInConfigOnly('Tracker', 'force_cookieless_tracking', $default = false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
+            $field->title = Piwik::translate('PrivacyManager_ForceCookielessTracking');
+            $field->uiControl = FieldConfig::UI_CONTROL_CHECKBOX;
+            $field->description = Piwik::translate('PrivacyManager_ForceCookielessTrackingDescription');
```
tsteur, line 78: here would probably also need to check if matomo.js is writable

tsteur:
@sgiehl left few comments. Goes in the right direction 👍

plugins/PrivacyManager/SystemSettings.php
sgiehl: I had implemented that as a regular setting. But then a lot tests started failing. Seems updating the tracker js is triggered in various cases where no db connection is yet configured. Checking if the `tracker.js` should be included using a SystemSetting then throws an error that even a try catch couldn't catch 🤷 
Not sure if that only happens in tests or might happen somewhen else as well.

plugins/PrivacyManager/tracker.js
sgiehl: Sure. That also won't work when cookie less tracking is "forced" for a single site only. So we need to advice people to call `disableCookies` nevertheless.
Actually I was thinking about having a separate piwik_no_cookies.js that maybe even does not contain any of the code that uses cookies at all. That would make it easy to ensure no cookies would be set... But would be of course cause more work to maintain both files

plugins/PrivacyManager/SystemSettings.php
sgiehl: sure, why not. will change that...

core/Tracker/Request.php
sgiehl: makes sense. changed that

plugins/PrivacyManager/MeasurableSettings.php
sgiehl: In that case we actually don't need to check that, as the website setting has no effect on the matomo.js. Only when set globally the matomo.js would be changed. But I can adjust the description a bit, so it's clear that this setting will only ensure the cookies aren't used server side and it's required to disable cookies in the tracker.

plugins/PrivacyManager/SystemSettings.php
tsteur: Not sure where it fails so bit hard to say. Generally be good though if we didn't have this stored in the config so people can disable/remove this setting from the UI etc.Not sure though how much work it be to fix it. Is it like that there are unit tests that are actually integration tests maybe and that's why it fails? Not sure where we generate tracker.js in tests.

sgiehl: It is triggered whenever a plugin is activated/deactivated and on some other events. Can revert that to a system setting and try fixing the tests again. But not sure how much effort that will be. Was quicker to use the config instead. But will give it a try tomorrow

tsteur: 👍

sgiehl: It took me a bit to figure out what the problem is. Seems `Piwik\Settings\Storage\Factory` was created in an early state where no database connection was yet set up, causing exceptions when it was used later. Recreating the factory in StaticContainer solves that.

plugins/PrivacyManager/SystemSettings.php
sgiehl: What shall we do if it is not writable? Shall we add an additional note, that it won't work fully or should we even hide the setting?

tsteur:
@sgiehl still reviewing but one thing I noticed is that we basically would need to move this feature from General settings page to the "Privacy -> Anonymise data" settings page. Simply to have these privacy related settings in one place. Should have noticed this earlier before we made heaps of work sorry about this.

Also generally noticing the `tracker.js` is actually problematic as it's currently based on a per site basis but we have only one tracker file for all sites. Meaning we can only add the tracker JS code if the feature is disabled for all sites globally.

see https://github.com/matomo-org/matomo/pull/16592/files#diff-9b9b9954067544c5b06d2d3bf22a15d59cc911577001e57d7fa06e7218339250R751

If we decide to still offer the feature to disable cookies on a per site basis then we'd need to rename the measurable setting for this feature to something like "ignore cookies" because it would still set cookies in the client but only ignore them on the server side. I'm not sure there's actually much of a benefit though as users and website visitors wouldn't know this and to users it would still look like cookies are being used and they could get in trouble. It's adding more confusion then it does good.

Technically, we need the possibility to disable this on a per site basis eg for https://github.com/matomo-org/matomo/issues/16363 . 

@sgiehl for now we should remove this feature from measurable settings and only have the global setting under `Privacy -> Anonymise Data` page. Sorry for wasting time here. 

As part of #16363 we will then think of a way to either let users configure on a per site basis that "they disabled cookies (and we disable it additionally server side)".

plugins/PrivacyManager/Controller.php
```diff
@@ -183,6 +185,11 @@ public function privacySettings()
         $view = new View('@PrivacyManager/privacySettings');
 
         if (Piwik::hasUserSuperUserAccess()) {
+            $jsCodeGenerator = new TrackerCodeGenerator();
+            $file = new File(PIWIK_DOCUMENT_ROOT . '/' . $jsCodeGenerator->getJsTrackerEndpoint());
+
+            $view->trackerFileName = $jsCodeGenerator->getJsTrackerEndpoint();
+            $view->trackerWritable = !$file->hasWriteAccess();
```
tsteur, line 192: We need to remove the `!` otherwise it is inverted

plugins/PrivacyManager/lang/en.json
```diff
@@ -83,6 +83,9 @@
         "DoNotTrack_SupportDNTPreference": "Support Do Not Track preference",
         "EstimatedDBSizeAfterPurge": "Estimated database size after purge",
         "EstimatedSpaceSaved": "Estimated space saved",
+        "ForceCookielessTracking": "Force tracking without cookies",
+        "ForceCookielessTrackingDescription": "Enabling this option will automatically update %1$s, so it contains some additional code to ensure all trackers won't use cookies. Additionally Matomo will ignore all tracking cookies on server side.",
+        "ForceCookielessTrackingDescriptionNotWritable": "Your %1$s is not writable. Therefor it can't be updated, to ensure cookies won't be used. Please make the file writable or ensure to call disableCookies manually for all javascript trackers. Otherwise Matomo will only ignore all tracking cookies on server side.",
```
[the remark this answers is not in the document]
tsteur: Could maybe say something like:

```
The JS tracker file "piwik.js" is not writable which is required for this feature to fully work. When the JS tracker file is not writable, then the cookies will be only ignored on the server side but cookies might be still set in the browser. To solve this issue you or a system administrator need to change the file permissions so your webserver can change this file.
```

Ideally there was a small FAQ explaining how this maybe can be done to give people some idea of what needs to happen. It be bit difficult because of the different operating systems etc. Or could we maybe show a command here that they can run? Maybe we have this already in Matomo from other directory write checks? Then we could add at the end: `For example you might be able to fix this by executing the command "chmod g+w /path/to/matomo/matomo/js"` but I suppose we can't really know if the user, the group needs to add write access (we wouldn't want to suggest give write access for `all`). 

Maybe later we could also have a feature for Matomo to try to give write access but not too important for now and often it wouldn't work anyway.

plugins/PrivacyManager/lang/en.json
```diff
@@ -83,6 +83,9 @@
         "DoNotTrack_SupportDNTPreference": "Support Do Not Track preference",
         "EstimatedDBSizeAfterPurge": "Estimated database size after purge",
         "EstimatedSpaceSaved": "Estimated space saved",
+        "ForceCookielessTracking": "Force tracking without cookies",
+        "ForceCookielessTrackingDescription": "Enabling this option will automatically update %1$s, so it contains some additional code to ensure all trackers won't use cookies. Additionally Matomo will ignore all tracking cookies on server side.",
```
tsteur, line 87: Could maybe say `Enabling this option will automatically update the JavaScript tracker %1$s to ensure cookies will be disabled. Additionally Matomo will ignore all tracking cookies on server side.` Just shortened it bit and making it bit more clear this affects only JS tracker but not other SDKs

plugins/PrivacyManager/Controller.php
```diff
@@ -329,6 +336,7 @@ private function getAnonymizeIPInfo()
         $privacyConfig = new Config();
         $anonymizeIP["enabled"] = IPAnonymizer::isActive();
         $anonymizeIP["maskLength"] = $privacyConfig->ipAddressMaskLength;
+        $anonymizeIP["forceCookielessTracking"] = $privacyConfig->forceCookielessTracking;
```
tsteur, line 339: when saving the setting, could we directly update the tracker? I think it didn't happen when I tested it and had to manually trigger the update.

tsteur: or maybe I didn't look correctly

plugins/PrivacyManager/lang/en.json
tsteur: we should also add a note: `Cookies will be disabled even when using the consent methods in Matomo tracker and calling for example the cookie consent methods won't enable cookies`.

tsteur (changes requested):
👍  worked. just left some minor comments.

There's one issue left that when calling `requireConsent` after the tracker has been set up, then cookies may be enabled if tracking consent has been given. That's pretty edge case though maybe. If we could prevent that, it would be great though.

plugins/PrivacyManager/Controller.php
sgiehl: sure. forgot to push that 🙈

plugins/PrivacyManager/Controller.php
sgiehl: Actually that should already be done in the API method that saves the settings...

sgiehl:
> There's one issue left that when calling requireConsent after the tracker has been set up, then cookies may be enabled if tracking consent has been given. That's pretty edge case though maybe. If we could prevent that, it would be great though.

is it enough to simply overwrite `requireConsent` with an empty function?

Applied the other changes.

tsteur:
> is it enough to simply overwrite requireConsent with an empty function?

Actually I just realise it only disables the cookies so it's no problem as it won't enable them. Meaning we need to remove ` tracker.requireConsent=function(){};` again.

core/Tracker/Request.php
```diff
@@ -738,21 +739,26 @@ public function getVisitorId()
...
-        if (!$found) {
-            $useThirdPartyCookie = $this->shouldUseThirdPartyCookie();
-            if ($useThirdPartyCookie) {
-                $idVisitor = $this->getThirdPartyCookieVisitorId();
-                if(!empty($idVisitor)) {
-                    $found = true;
+        $cache = Tracker\Cache::getCacheGeneral();
+
+        // Only check for cookie values if cookieless tracking is NOT forced
+        if (!isset($cache['forceCookielessTracking']) || !$cache['forceCookielessTracking']) {
```
tsteur, line 745: btw here it be good if we used some constants as it's set in a different place and read here

tsteur:
Feel free to merge once the change is made. Could you maybe then create a quick new FAQ for this like `How do I enforce cookieless tracking` (not sure we use cookieless tracking much could also say "disable cookies" or something). We'd need to explain that this affects all sites etc. We could link to this new FAQ from https://matomo.org/faq/general/faq_157/
and https://matomo.org/docs/privacy/ . Feel free to directly make the changes. We can also link to it from other places in the future

core/Tracker/Request.php
sgiehl: will change that to directly use the privacy config (which will use tracker cache)

sgiehl:
@tsteur created https://matomo.org/faq/how-do-i-enforce-tracking-without-cookies/
Feel free to adjust it if something needs to be improved. Also haven't yet added any links to that page.

PR #16595, opened 2020-10-20
Add new INI config [General] enable_framed_allow_write_admin_token_auth…

### LINKED ISSUE #16554: Using auth tokens of "write" or "admin" permissions when embedding
the new security feature in https://github.com/matomo-org/matomo/pull/16263 would break use cases for users who need to be able to create/update entities like goals and custom dimensions from within an iframe. This is the case when users embed the whole Matomo App within their products. [as quickly described in this faq](https://matomo.org/faq/troubleshooting/faq_147/)). Currently they would get the message `This user has at least some write access. Only tokens of users who have only view access can be used.`. 

Initially was thinking we could have yet another INI setting, but actually  it looks better/easier to accept write/admin tokens whenever `enable_framed_pages=1` or `enable_framed_settings=1` are set in the config? 

(need to be added to RC as this would otherwise be a BC break for some)

diosmosis:
… to allow framed matomo use case to still function in Matomo 4.

Fixes #16554

NOTE: faq is updated at https://matomo.org/faq/troubleshooting/#faq_147


config/global.ini.php
```diff
@@ -455,6 +455,12 @@
 ; Default is 0 (i.e., bust frames on the Settings forms).
 enable_framed_settings = 0
 
+; Set to 1 to allow using token_auths with write or admin access in iframes that embed Matomo.
+; Note that the token used will be in the URL in the iframe, and thus will be stored in webserver
+; logs and possibly other places. Using write or admin token_auths can be seen as a security risk,
+; thought it can be necessary in some use cases.
```
tsteur, line 461: is it supposed to be `though` instead of `thought`? Be good to mention specifically like `We do not recommend enabling this setting, for more information view the FAQ: ...`? with the link to the other FAQ

core/FrontController.php
```diff
@@ -421,8 +421,14 @@ public function init()
         // Force the auth to use the token_auth if specified, so that embed dashboard
         // and all other non widgetized controller methods works fine
         if (Common::getRequestVar('token_auth', '', 'string') !== ''
-            && Request::shouldReloadAuthUsingTokenAuth(null)) {
+            && Request::shouldReloadAuthUsingTokenAuth(null)
+        ) {
             Request::reloadAuthUsingTokenAuth();
+
+            if (Access::getInstance()->hasSuperUserAccess()) {
```
tsteur, line 428: could maybe move this into `isTokenAuthLimitedToViewAccess ` method and then rename the method `isTokenAuthLimitedToViewAccess `? Just in case it will ever be reused in other places.

core/API/Request.php
```diff
@@ -460,13 +461,19 @@ private static function forceReloadAuthUsingTokenAuth($tokenAuth)
      */
     public static function isTokenAuthLimitedToViewAccess($module, $action)
     {
+        $allowWriteAmin = Config::getInstance()->General['enable_framed_allow_write_admin_token_auth'] == 1;
```
tsteur, line 464: btw be good to add 2 or 3 simple tests for this method

config/global.ini.php
diosmosis: 👍

core/FrontController.php
diosmosis: Will refactor to merge the logic.

diosmosis:
@tsteur added the tests, took a while to get them to pass due to a weird quirk of phpunit

sgiehl (approved):

sgiehl:
Test test `FrontControllerTest::test_thrownExceptionInFrontControllerPrintsBacktrace` seems to be failing due to the changes here... Otherwise looks good to merge

plugins/Widgetize/lang/en.json
```diff
@@ -2,7 +2,8 @@
     "Widgetize": {
         "OpenInNewWindow": "Open in a new window",
         "PluginDescription": "Display any Matomo report in your website or app with a simple Embed HTML tag.",
-        "ViewAccessRequired": "This user has at least some write access. Only tokens of users who have only view access can be used.",
-        "TopLinkTooltip": "Export Matomo Reports as Widgets and embed the Dashboard in your app as an iframe."
+        "ViewAccessRequired": "This user has at least some write access. Only tokens of users who have only view access can be used. See %1$s for more information.",
+        "TopLinkTooltip": "Export Matomo Reports as Widgets and embed the Dashboard in your app as an iframe.",
+        "TooHighAccessLevel": "This user has too much access for embedding widgets. Please use a user less access."
```
tsteur, line 7: Maybe we could say `This user has super user access. For embedding widgets please use the token of a user that has view access only.`? could then also maybe link to https://matomo.org/faq/troubleshooting/faq_147/ if needed?

diosmosis: This error could also occur if the allow admin/write token auth setting is set, so will change it to`This user has super user access. For embedding widgets super user token auths are not allowed. %1$sSee our faq for more information.%2$s`.

tsteur:
👍  feel free to merge once the tests pass @diosmosis

PR #16598, opened 2020-10-20
Improve possibility to disable visits log / visitor profile

sgiehl:
refs #16259


plugins/Live/API.php
```diff
@@ -277,6 +282,10 @@ public function getFirstVisitForVisitorId($idSite, $visitorId)
     {
         Piwik::checkUserHasSomeViewAccess();
 
+        if (!Live::isVisitorProfileEnabled($idSite)) {
+            throw new Exception('Visitor profile has been disabled in website settings');
```
tsteur, line 286: BTW maybe there could be `Live::checkVisitorProfileEnabled()` to have less duplicate code? Ideally it was to mention whether it is disabled in website settings or globally but not too important

plugins/Live/API.php
```diff
@@ -147,6 +148,10 @@ public function getLastVisitsDetails($idSite, $period = false, $date = false, $s
         }
         Piwik::checkUserHasViewAccess($idSite);
 
+        if (!Live::isVisitorLogEnabled($idSite) && (Request::isCurrentApiRequestTheRootApiRequest() || Request::getRootApiRequestMethod() === 'Live.getLastVisitsDetails')) {
```
tsteur, line 151: would this also for bulk requests? Might need to specifically only allow the method to be called when `API.getSuggestedValuesForSegment` is called as the root api request method. We don't really use that in bulk so it should work I suppose

plugins/Live/Controller.php
```diff
@@ -89,6 +97,10 @@ public function getLastVisitsStart()
     {
         Piwik::checkUserHasViewAccess($this->idSite);
 
+        if (!Live::isVisitorLogEnabled($this->idSite)) {
```
tsteur, line 100: same here and in other places ideally there was a `Live::checkVisitorLogEnabled`

plugins/Live/API.php
sgiehl: Changed that. The method `API.getSuggestedValuesForSegment` now only works when called directly and fails when called in a bulk request (if visits log is disabled)

sgiehl:
@tsteur did the suggested adjustments.

plugins/Live/Live.php
```diff
@@ -57,69 +52,99 @@ public function addJsGlobalVariables(&$out)
...
+     * @param null|int|array $idSite
+     * @throws \Exception
+     */
+    public static function checkIsVisitorProfileEnabled($idSite = null): void
+    {
+        $systemSettings = new SystemSettings();
 
-            $systemSettings = new SystemSettings();
+        if ($systemSettings->activateVisitorProfile->getValue() == false) {
+            throw new \Exception('Visitor profile is deactivated globally');
```
tsteur, line 113: We could maybe add something like `. A user with super user access can enable this feature in the general settings.`?

plugins/Live/Live.php
```diff
@@ -57,69 +52,99 @@ public function addJsGlobalVariables(&$out)
...
+        if (!empty($idSite)) {
+            $idSites = is_array($idSite) ? $idSite : [$idSite];
+
+            foreach ($idSites as $idSite) {
+                $settings = new MeasurableSettings($idSite);
 
-            if ($systemSettings->activateVisitorLog->getValue() === false) {
-                $visitorLogEnabled = false;
+                if ($settings->activateVisitorProfile->getValue() == false) {
+                    throw new \Exception('Visitor profile is deactivated in website settings');
```
tsteur, line 127: We could maybe add something like `. A user with at least admin access can enable this feature in the settings for this website.`?

tsteur (changes requested):
@sgiehl could we add some description to the settings?

the real time map needs to be disabled as well when visitor log is disabled? And when visitor profile is disabled the bubbles in real time map need to be disabled if possible (haven't checked this one yet, might be already done)

would we also need to remove the "test" button in the segment editor? or would we still show it and people see the error message that the visitor log is disabled? currently seeing this 
image would maybe at least need to be bit more specific and mention visitor log was disabled

Be great if we could also add a sentence to intro of the `Privacy -> Anonymise data` page as mentioned in the issue I think with a link to the `Live` settings (if the Live plugin is enabled). 

Noticed an edge case re Roll Up Reporting that need a documentation or maybe we mention it as part of the system setting or not sure... If there is a Roll Up that has the features enabled and includes a children site where this feature is disabled, then the data will be still shown for this children site because of the event `Live.API.getIdSitesString`. Or maybe ideally we could adjust roll up reporting to not add sites in the `addIdSitesToLiveQueries` where this feature is disabled? Then we wouldn't need to explain anything in the UI and we would maybe have an FAQ explaining "Why the logs of some sites might not appear in the roll up visitor log".

sgiehl:
* Realtime map is now disabled together with visits log
* Button to test segments is also hidden then
* Realtime map won't try to open visitor profile if it's disabled

tsteur:
BTW @sgiehl haven't checked yet but do we need to have a check that visit log also requires visit profile to be disabled? Would we need to rename the settings to like below to make this more clear?

* `Disable Visits profile`
* `Disable Visits log and visitor profile`

Like technically, none of the profile features work without the visits log except for maybe the API. In any way it could be misleading maybe to have visitor profile enabled but not the visits log because they wouldn't find a way through the UI to get there

tsteur:
Works otherwise. 

The settings still need a description though.

And the userId feature is still showing the visitor profile feature: 
image

plugins/Live/Live.php
```diff
@@ -57,69 +52,99 @@ public function addJsGlobalVariables(&$out)
...
-            ];
-        }
+            foreach ($idSites as $idSite) {
+                $settings = new MeasurableSettings($idSite);
 
-        $siteIdLoaded = $idSite;
-        $visitorProfileEnabled = true;
-        $visitorLogEnabled = true;
+                if ($settings->activateVisitorLog->getValue() == false) {
+                    throw new \Exception('Visits log is deactivated in website settings. A user with at least admin access can enable this feature in the settings for this website.');
```
tsteur, line 80: could we include the idSite in here? This may be useful when multiple sites are given. Ideally, if multiple sites are given, we would only filter out the idSites that aren't working though and keep all others but for a first release I would say this isn't needed. Be just useful to know which idSite of the many are disabled

plugins/Live/Live.php
```diff
@@ -57,69 +52,99 @@ public function addJsGlobalVariables(&$out)
...
+        if (!empty($idSite)) {
+            $idSites = is_array($idSite) ? $idSite : [$idSite];
+
+            foreach ($idSites as $idSite) {
+                $settings = new MeasurableSettings($idSite);
 
-            if ($systemSettings->activateVisitorLog->getValue() === false) {
-                $visitorLogEnabled = false;
+                if ($settings->activateVisitorProfile->getValue() == false) {
+                    throw new \Exception('Visitor profile is deactivated in website settings. A user with at least admin access can enable this feature in the settings for this website.');
```
tsteur, line 127: here be also great if we could include the idSite that caused this

tsteur:
I think also the link from the privacy -> anonymise data page to this setting is still missing I think?

sgiehl:
> Like technically, none of the profile features work without the visits log except for maybe the API. In any way it could be misleading maybe to have visitor profile enabled but not the visits log because they wouldn't find a way through the UI to get there

There is actually still the visitor profile widget that can be added to the dashboard. And the UserId report still has the possibility to open a profile for the user. 

Another thing I just noticed is that disabling the visitor log currently breaks the GDPR Tools, as they use Live.getLastVisitsDetails to search for visitors. Not sure how to handle that, as imho I guess those tools should actually work even if visits log / profile are disabled. Could maybe proxy the API call through a controller method, so I'm able to circumvent the check somehow.
On the other side would that open up the possibility to get parts of the visits log data for users with admin access even if its disabled... Any suggestion on that?

plugins/Live/Live.php
sgiehl: The filtering can't be done in this method. But can add it to the `Live.getLastVisitsDetails`

tsteur:
> Another thing I just noticed is that disabling the visitor log currently breaks the GDPR Tools, as they use Live.getLastVisitsDetails to search for visitors. Not sure how to handle that, as imho I guess those tools should actually work even if visits log / profile are disabled. Could maybe proxy the API call through a controller method, so I'm able to circumvent the check somehow.
> On the other side would that open up the possibility to get parts of the visits log data for users with admin access even if its disabled... Any suggestion on that?

Just like we whitelisted the `API suggested segment values` could we also white list this method? 

Below some more needed changes in other plugins: They should also apply if the `Live` plugin is disabled ideally.

For Media Analytics: Could you add a PR to disable the real time map widget when visitor log is disabled and also the "audience log widget"?

In Heatmap & Session Recording there is an icon with a link to the profile popup that will need to be removed: 
image

In form analytics will need to remove these links:
image

in funnels: There is also a link to segmented visitor log: 
image

>> Like technically, none of the profile features work without the visits log except for maybe the API. In any way it could be misleading maybe to have visitor profile enabled but not the visits log because they wouldn't find a way through the UI to get there

> There is actually still the visitor profile widget that can be added to the dashboard. And the UserId report still has the possibility to open a profile for the user.

I'm thinking the visitor profile should only be possible to be deactivated when also deactivating the visitor log. While the feature could be used individually, you could argue that the visitor profile uses features of the visitor log. Also while there are few edge case places where it could be used individually, most normal ways be blocked and it would cause in the end more confusion. Meaning visitor profile can be only disabled together with the visitor log.

sgiehl:
> Just like we whitelisted the API suggested segment values could we also white list this method?

Actually not, as the JavaScript directly calls the `Live.getLastVisitsDetails` method. So the only possibility I see is to add a new proxy method in the controller, that returns the relevant content of the API result that is needed for the GDPR tools.

tsteur:
@sgiehl  Instead of a controller method be better to create a new API method in the privacy manager. It would then not even take any date/period as input but only the siteIds and segment filter. It's better to use API so it can be used by other consumers, it supports different formats, automatically has the token check etc. Also we already have `deleteDataSubjects` and `exportDataSubjects` API so be good to also add a `findDataSubjects` API. It could also make sure to only return the needed information so less data will be transferred plus it can't be used to still get visitor log details etc.

sgiehl:
@tsteur implemented the new api method. Should we hide the link to the visitor profile in the GDPR tools if the visits log or profile is disabled? Might be hard to temporarily allow opening a visitor profile for that case...

plugins/PrivacyManager/API.php
```diff
@@ -90,14 +90,43 @@ public function findDataSubjects($idSite, $segment)
...
+        $result = Request::processRequest('Live.getLastVisitsDetails', [
             'segment' => $segment,
             'idSite' => $idSite,
             'period' => 'range',
             'date' => '1998-01-01,today',
             'filter_limit' => 401,
             'doNotFetchActions' => 1
         ]);
+
+        $columnsToKeep = [
```
tsteur, line 102: btw I suppose `showColumns` parameter doesn't work here? Not important though

tsteur:
> Should we hide the link to the visitor profile in the GDPR tools if the visits log or profile is disabled? Might be hard to temporarily allow opening a visitor profile for that case...

sounds good to remove the feature from there for now when it is disabled.

image

I wonder if we should remove it to rather use `Disable Visitor Log & Visitor profile` and then "check" it to disable it? Cause see screenshot above it's not clear when disabling both that visitor profile is also disabled. It becomes invisible. Would then need to name the other setting "Disable visitor profile".

There are still descriptions missing for the settings. Can you add them and then if tests pass merge the PR? Feel free to then also merge all the related PRs

sgiehl:
@tsteur could you check if the descriptions are fine that way? Will update the tests and merge tomorrow. Waiting for the travis build since an hour already 🤷

tsteur:
@sgiehl looks good! I'll merge now so we can include the feature in the next beta. Maybe you can create a PR to fix the ui tests tomorrow? Just fixed some system/integration test now but there might be still one left also

PR #16600, opened 2020-10-21
MobileMessaging twig fix for 'Variable "ajax" does not exist.'

Findus23:
> A fatal error occurred
> The following error just broke Matomo (v4.0.0-b3):
> 
> Variable "ajax" does not exist.
> in /home/lukas/public_html/matomophp8/plugins/MobileMessaging/templates/index.twig line 156            

I am not sure if a twig update broke this or it never worked anyway


sgiehl (approved):
good find 👍

PR #16603, opened 2020-10-23
several changes to emphasize setup and use of auto-archiving rather than real-time processing

### LINKED ISSUE #6210: Detect when stats load slowly and explain to user to disable Browser trigger archiving in Admin
The goal of this issue is to help users who have poor experience of Piwik by detecting when it happens and displaying a help notification to guide them in right direction.

Problem: **Piwik loads stats slowly for many users when Browser trigger archiving is enabled in settings**. It affects mostly

**Suggested tasks**
- In the `Dashboard` and `All Websites` dashboard, **detect when stats take more than a few seconds to load and display inline notification suggesting to solve the issue.**
- The notification message can change depending on the use case:
  1. when browser archiving is enabled, recommend user to setup [Cron archiving](http://piwik.org/docs/setup-auto-archiving/)
  2. when segments are used, and it's not set to pre-processed, explain that it's recommended to set segments to be pre-processed.
  3. when a custom date range is used, explain that they are not pre-processed and can be slow. Recommend user to use Weekly/Monthly/Yearly reports if problem persists. 
     - in case user selects to load by default a date range eg. `Last 30 days` then we should pre-process it anyway, see #6672 

How often should the notification be displayed? if the user closes the notification, maybe it could be displayed again in one week. The notification should be displayed to all users but messaging should change for Super User to recommend them to change the settings or setup cron (for normal users it would recommend to contact the Piwik Administrator to let them know of the issue and suggest them to tweak Piwik).

**Who is affected by issue?**
Almost all users by default will experience this issue sooner rather than later. By default the [automated cron](http://piwik.org/docs/setup-auto-archiving/) is **not** setup and user has  browser archiving enabled in piwik settings. by default we load Today, and All websites can also load the Last 30 days, so the user will request a lot of fresh data most of the time!

Maybe  you have other ideas?

diosmosis:
Fixes #6210 

Changes:

* Add blurb to installation finished twig about cron setup.
* Make cron setup an actual diagnostic instead of an informational result.
* Emphasize system check widget in settings home so items are more likely to be noticed and taken care of.
  * Also make tooltip use first diagnostic result's comment if there is no long message and use jquery tooltip to make them more noticeable.
* Allow view access users to create auto archive segments.
* Show count of pre-processed vs real-time segments in system summary.

TODO:
* [ ] automated tests for segment creation change (allowing view access) once the PR is reviewed

Screenshots:

**System Check Dash**

image

**Installation**

image

**System Summary**

image


plugins/Installation/templates/finished.twig
```diff
@@ -27,6 +27,10 @@
         </p>
     {% endif %}
 
+    <h3>Performance Settings</h3>
+    <p>{{ 'Installation_PerformanceSettingsDesc1'|translate('<a href="https://matomo.org/docs/setup-auto-archiving/" rel="noreferrer noopener">', '</a>')|raw }}</p>
```
tsteur, line 31: could we add a target _blank here for both links @diosmosis ? That be great as when I clicked on it I left the installer

plugins/SegmentEditor/SegmentEditor.php
```diff
@@ -80,7 +80,19 @@ public function addSystemSummaryItems(&$systemSummary)
...
+        foreach ($segments as $segment) {
+            $autoArchive = (int)$segment['auto_archive'];
+            if (!empty($autoArchive)) {
+                ++$numAutoArchiveSegments;
+            } else {
+                ++$numRealTimeSegments;
+            }
+        }
+
+        $systemSummary[] = new SystemSummary\Item($key = 'segments', Piwik::translate('CoreHome_SystemSummaryNSegments', [$numSegments, $numAutoArchiveSegments, $numRealTimeSegments]), $value = null, $url = null, $icon = 'icon-segment', $order = 6);
```
tsteur, line 95: if `browser_archiving_disabled_enforce=1 (Rules::isBrowserArchivingAvailableForSegments()=true)` then we wouldn't show the message of how many are pre-processed/ real time and only show overall number of segments. As most users wouldn't understand what pre-processed etc means (eg on cloud). In that case all segments would be pre-processed

We would then only like previously show the total number of segments.

plugins/SegmentEditor/templates/_segmentSelector.twig
```diff
@@ -43,8 +43,8 @@
 
                 {{ 'SegmentEditor_ThisSegmentIsVisibleTo'|translate }} <div class="enable_all_users"><strong>
                         <select class="enable_all_users_select">
-                            <option value="0">{{ 'SegmentEditor_VisibleToMe'|translate }}</option>
-                            <option selected="1" value="1">{{ 'SegmentEditor_VisibleToAllUsers'|translate }}</option>
+                            <option {% if not isSuperUser %}selected="1"{% endif %} value="0">{{ 'SegmentEditor_VisibleToMe'|translate }}</option>
```
tsteur, line 46: was this change meant @diosmosis ? Seems maybe unrelated to this PR?

tsteur: I think what was meant to happen is that "preprocessed reports" should be selected by default when possible (eg when browser archiving is disabled aka a cron was set up).

"Preprocessed reports" should also only be selectable ideally when browser archiving is disabled.

plugins/SegmentEditor/API.php
```diff
@@ -107,16 +107,7 @@ protected function checkAutoArchive($autoArchive, $idSite)
...
-                // pre-processed segment for a given website requires admin access
-                throw new Exception(
-                    "Please contact Support to make these changes on your behalf. ".
-                    " To modify a pre-processed segment, a user must have admin access or super user access. "
-                );
-            } elseif (!$realTimeSegmentsEnabled) {
-                // we require view access only when only pre processed can be created
-                Piwik::checkUserHasViewAccess($idSite);
-            }
+            Piwik::checkUserHasViewAccess($idSite);
```
tsteur, line 110: I think it should be possible to set `autoArchive=1` only when browser archiving is disabled? Otherwise the segment would be never archived?

We'll also need to adjust the config text in `enable_create_realtime_segments` as it no longer restricts which kinds of users can create segments to be archived on-demand. Could later if needed add a new setting like `require_admin_access_for_preprocessed_segments=0` but not needed for now.

plugins/Installation/lang/en.json
```diff
@@ -131,7 +131,7 @@
         "SystemCheckWriteDirs": "Directories with write access",
         "SystemCheckWriteDirsHelp": "To fix this error on your GNU\/Linux system, try typing in the following command(s)",
         "SystemCheckZlibHelp": "You need to configure and rebuild PHP with \"zlib\" support enabled, --with-zlib.",
-        "SystemCheckCronArchiveProcess": "Archive Cron",
+        "SystemCheckCronArchiveProcess": "Setup Cron",
```
tsteur, line 134: I'm still finding it useful if it could be a bit more clear what the outcome/result is of fixing this warning so users will pay more attention to it as "Setup cron" you would likely simply ignore. Like "Setup Cron (faster report loading)" or something. Not too important but I think it will get the attention of users a lot more

plugins/CoreHome/lang/en.json
```diff
@@ -82,7 +82,7 @@
         "PivotBySubtable": "This report is not pivoted %1$s Pivot by %2$s",
         "SystemSummaryWidget": "System Summary",
         "SystemSummaryNWebsites": "%d websites",
-        "SystemSummaryNSegments": "%d segments",
+        "SystemSummaryNSegments": "%1$d segments (%2$s pre-processed, %3$s processed in real-time)",
```
tsteur, line 85: >  54 segments (7 pre-processed, 47 processed in real-time)

could we maybe rename `%2$s pre-processed` to something like `%2$s processed in background` or so? Although I reckon that doesn't make it much more clear what it means either.

tsteur (changes requested):

plugins/SegmentEditor/SegmentEditor.php
diosmosis: assuming you mean `Rules::isBrowserArchivingAvailableForSegments()==false` since isBrowserArchivingAvailableForSegments has `!$generalConfig['browser_archiving_disabled_enforce']`

plugins/SegmentEditor/templates/_segmentSelector.twig
diosmosis: This is to allow view access users to be able to create segments, otherwise the creation will always fail because the default is "visible to all users".

plugins/SegmentEditor/API.php
diosmosis: > I think it should be possible to set autoArchive=1 only when browser archiving is disabled?

I think you mean when cron archiving isn't set up, but I don't think there's a good way to check for that? If browser archiving is enabled and cli archiving is enabled, then it should be possible to use auto archive.

plugins/CoreHome/lang/en.json
diosmosis: What about `processed periodically`? Or `reports generated periodically` vs `reports generated in real-time`?

plugins/SegmentEditor/SegmentEditor.php
tsteur: 👍

plugins/SegmentEditor/templates/_segmentSelector.twig
tsteur: 👍

plugins/SegmentEditor/API.php
tsteur: I'm not sure we should see having both browser archiving and cli archiving enabled as a use case? Of course few users might do it (eg because they forget to disable browser archiving) but it should be rather edge cause and most of the times there wouldn't be much of a benefit of having both running AFAIK. Unless there would be some?

plugins/CoreHome/lang/en.json
tsteur: I reckon we can actually just leave it as it is. I don't think any other wording makes it much clear. Same for `processed via cron` etc.

plugins/SegmentEditor/API.php
```diff
@@ -107,16 +108,11 @@ protected function checkAutoArchive($autoArchive, $idSite)
...
-                // pre-processed segment for a given website requires admin access
-                throw new Exception(
-                    "Please contact Support to make these changes on your behalf. ".
-                    " To modify a pre-processed segment, a user must have admin access or super user access. "
-                );
-            } elseif (!$realTimeSegmentsEnabled) {
-                // we require view access only when only pre processed can be created
-                Piwik::checkUserHasViewAccess($idSite);
+            if (Rules::isBrowserTriggerEnabled()) {
+                throw new Exception("Pre-processed segments can only be created if browser triggered archiving is disabled.");
```
tsteur, line 112: Could we add a link to https://matomo.org/docs/setup-auto-archiving/  if it's a super user? Like `To disable browser archiving follow the instructions in https://matomo.org/docs/setup-auto-archiving/ .` Just to give them an idea of what to do next.

tsteur (changes requested):
@diosmosis I reckon the only remaining change needed is to set in segment editor "pre-processed" as default when browser archiving is disabled. For me so far it still always shows real time as default 
image

I also created a follow up issue https://github.com/matomo-org/matomo/issues/16646 to eventually improve segment editor as the option is barely visible depending on the resolution and people currently wouldn't know what to select when. 
image

diosmosis:
@tsteur applied some changes

plugins/SegmentEditor/templates/_segmentSelector.twig
```diff
@@ -56,10 +56,10 @@
                     </strong></div>
                 {{ 'General_And'|translate }} <div class="auto_archive"><strong>
                         <select class="auto_archive_select">
-                            {% if createRealTimeSegmentsIsEnabled %}
+                            {% if createRealTimeSegmentsIsEnabled or isBrowserArchivingEnabled %}
```
tsteur, line 59: @diosmosis not sure this change is actually needed?

plugins/SegmentEditor/templates/_segmentSelector.twig
```diff
@@ -56,10 +56,10 @@
...
                             <option selected="1" value="0">{{ 'SegmentEditor_AutoArchiveRealTime'|translate }} {{ 'General_DefaultAppended'|translate }}</option>
                             {% endif %}
-                            <option {% if not createRealTimeSegmentsIsEnabled %}selected="1"{% endif %} value="1">{{ 'SegmentEditor_AutoArchivePreProcessed'|translate }} </option>
+                            <option {% if not createRealTimeSegmentsIsEnabled and not isBrowserArchivingEnabled %}selected="1"{% endif %} value="1">{{ 'SegmentEditor_AutoArchivePreProcessed'|translate }} </option>
```
tsteur, line 62: btw probably not an issue but do we maybe need to remove `selected="1"` from the previous option when selecting this one by default? I suppose the browser in this case automatically selects the later one?

I'm thinking maybe it needs to use an `or` and not an `and`? Like we want to preselect this one either when browser archiving is disabled or when real time segment is not allowed?

plugins/SegmentEditor/templates/_segmentSelector.twig
diosmosis: @tsteur you said:

> I reckon the only remaining change needed is to set in segment editor "pre-processed" as default when browser archiving is disabled.

plugins/SegmentEditor/templates/_segmentSelector.twig
diosmosis: They should be mutually exclusive conditions.

tsteur: @diosmosis I just disabled browser archiving but real time was still the default. It could alternatively be simplified to `if not isBrowserArchivingEnabled`. 

Otherwise it would be only the default if someone was to disable real time segments but only few people would do this.

diosmosis: @tsteur that may be because browser archiving of segments is still enabled. Can you check that?

tsteur: That's possible. I think we'd want to still enable it though by default in that case see https://github.com/matomo-org/matomo/issues/6210#issuecomment-712408748
image

and then in https://github.com/matomo-org/matomo/issues/6210#issuecomment-712464414
image

It wasn't really meant in relation to the segment specific archive setting but just the general browser archiving setting.

diosmosis: Ok, I will ignore segment specific archiving.

diosmosis: @tsteur changed.

plugins/SegmentEditor/templates/_segmentSelector.twig
```diff
@@ -56,10 +56,10 @@
                     </strong></div>
                 {{ 'General_And'|translate }} <div class="auto_archive"><strong>
                         <select class="auto_archive_select">
-                            {% if createRealTimeSegmentsIsEnabled %}
+                            {% if isBrowserArchivingEnabled %}
```
tsteur, line 59: This would need to be reverted to the original `if createRealTimeSegmentsIsEnabled `? If browser archiving is disabled, then you might still want to create a segment for "real time" if you use the segment for example only for the "visits log" or if you view it very rarely. It should just not be selected by default.

tsteur (commented):
Left one more comment @diosmosis and then feel free to merge once the tests pass.

diosmosis:
@tsteur updated again, will merge if tests pass and no further comments

tsteur:
Great. Thanks @diosmosis

Issue #16576, opened 2020-10-15
console (cli): Add config:delete to delete a key


gramakri:
There is already a `config:set` to add a key in a section. Would be helpful to have a config:delete to delete a key in a section as well. This would be especially useful when combined with `--value` to remove an entry from array keys like severs[], Plugins[].

tsteur:
Thanks for suggesting this feature @gramakri 

If someone is keen to work on this: It could work similar to https://github.com/matomo-org/matomo/blob/4.0.0-b3/plugins/CoreAdminHome/Commands/SetConfig.php#L19-L86 in a new `DeleteConfig` class. It's probably quite easy to implement and we will be happy to review.

mwithheld:
I've put in a PR for this at
https://github.com/matomo-org/matomo/pull/17956

PR #16608, opened 2020-10-24
console (cli): Add config:delete to delete a key #16576

javier-de-juan:
This PR tries to add the delete command for config file. For array positions deletion, just admit numeric number because if  the code removes one labeled position from the array, all other keys are lost (dumpConfig method related), so I preferred to avoid bad config generation.

### DIFF
--- a/plugins/CoreAdminHome/Commands/DeleteConfig.php
+++ b/plugins/CoreAdminHome/Commands/DeleteConfig.php
@@ -25,20 +25,16 @@ class DeleteConfig extends ConsoleCommand
     {
         $this->setName('config:delete');
         $this->setDescription('Delete one or more config settings in the file config/config.ini.php');
-        $this->addArgument('assignment', InputArgument::OPTIONAL | InputArgument::IS_ARRAY, "List of config setting assignments, eg, section or section.config_setting_key");
+        $this->addArgument('assignment', InputArgument::OPTIONAL | InputArgument::IS_ARRAY, "List of config setting assignments, eg, section.config_setting_key");
         $this->addOption('section', null, InputOption::VALUE_REQUIRED, 'The section the INI config setting belongs to.');
         $this->addOption('key', null, InputOption::VALUE_REQUIRED, 'The name of the INI config setting.');
         $this->setHelp("This command can be used to remove INI config settings on a Piwik instance.
 
 You can remove config values two ways, via --section, --key or by command arguments.
 
-To use --section, --key, simply supply those options. You can remove one whole section or one key inside the section.
+To use --section, --key, simply supply those options.
 
 To use arguments, supply one or more arguments in the following format:
-
-Remove section:
-$ ./console config:delete 'section'
-'section' is the name of the section,
 
 Remove setting inside section
 $ ./console config:delete 'section.config_setting_key'
@@ -64,17 +60,6 @@ $ ./console config:delete 'section.config_setting_key'
         $config = Config::getInstance();
 
         foreach ($manipulations as $manipulation) {
-            /** @var ConfigDeletingManipulation $manipulation */
-            if ($manipulation->isSectionDeletion()) {
-                $helper = $this->getHelper('question');
-                $question = new ConfirmationQuestion('<question>Do you want to remove the section "' . $manipulation->getSectionName() . '"? (y/n)</question> ' , false);
-
-                if (!$helper->ask($input, $output, $question)) {
-                    $output->writeln( "<comment>\"".$manipulation->getSectionName()."\" will not be deleted</comment>");
-                    continue;
-                }
-            }
-
             try {
                 $manipulation->manipulate($config);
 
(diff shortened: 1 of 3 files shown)


plugins/CoreAdminHome/Commands/DeleteConfig.php
```diff
@@ -0,0 +1,113 @@
...
+        $this->setHelp("This command can be used to remove INI config settings on a Piwik instance.
+
+You can remove config values two ways, via --section, --key or by command arguments.
+
+To use --section, --key, simply supply those options. You can remove one whole section or one key inside the section.
+
+To use arguments, supply one or more arguments in the following format:
+
+Remove section:
+$ ./console config:delete 'section'
```
tsteur, line 38: @javier-de-juan  do you need the feature to remove an entire section? Cause if used wrongly it could do a lot of harm and break Matomo (eg if someone removes `General` or any other sections and forgets to pass a `key`, then Matomo be broken probably). If you need this feature, we'd probably need another flag like `only-section` just to be sure this is wanted

plugins/CoreAdminHome/Commands/DeleteConfig.php
```diff
@@ -0,0 +1,113 @@
...
+'section' is the name of the section,
+
+Remove setting inside section
+$ ./console config:delete 'section.config_setting_key'
+'section' is the name of the section,
+'config_setting_key' the name of the setting
+
+To remove an array setting position, supply an argument like this:
+$ ./console config:set 'section.config_setting_key[position]'
```
tsteur, line 47: @javier-de-juan in case you don't need this feature be great to remove it as well just as it adds a bit of complexity and people might rarely know the correct index of an array etc. Happy to keep it though if needed.

tsteur (commented):
Cheers for the PR @javier-de-juan very appreciated. Added 2 comments to check if we need certain features.

plugins/CoreAdminHome/Commands/DeleteConfig.php
javier-de-juan: Hi @tsteur! In my case it's not needed but I tried to think in all cases. I can remove this feature and keep it simple. I'll do it ASAP, today will be done ;)

plugins/CoreAdminHome/Commands/DeleteConfig.php
javier-de-juan: The same with this, I could remove it. I'm going to wait for someone with different opinion and If there's no comments today I will change it and give just the posibility of remove settings in sections ;)

tsteur: Thanks. Be great to remove it and if needed it can be always added quickly. This way it's less things to maintain that otherwise maybe nobody would use etc and simplifies the documentation, the code etc. Let me know once you made the changes and we'll review 👍  Thanks for this, very appreciated.

javier-de-juan: This feature was completely removed from the code ;)

plugins/CoreAdminHome/Commands/DeleteConfig.php
javier-de-juan: I was thinking about this feature and, in my opinion, it could be useful. Finally I decided to put a confirm question to avoid undesired behaviour. What do you think???

tsteur: Not sure where it would be useful? Not sure there is any section that can be removed without breaking anything actually.

javier-de-juan: Ok @tsteur The feature was also removed. Actually the command will only remove settings in sections 👍

diosmosis:
@tsteur the original issue references having a `--value` key to delete array elements by value. Is this no longer a requirement?

tsteur:
@diosmosis  good one. That could be useful indeed. @javier-de-juan I think previously one had to set a specific index to delete a value within an index. But the correct index is always hard to know. If it was possible to use `--value` to delete a value within an array only when it matches that value that might be indeed quite useful. Sorry about that. Would it be possible to add that?

javier-de-juan:
Ok, there's no problem. But... what's the desired behavior? If 2 settings have the same value the command deletes both? The section is always required, optional? I need a description about what you need to avoid wrong implementation.

tsteur:
I would say deleting value from an array would only work when specifying the key and the section as well.

Basically `--section=General --key=trusted_hosts --value=mydomain.com` .

When a value is set, and the configured value is a string (or a number) then we would only delete that value if the configured value matches the value from the `value` option. AKA if `$config->$section[$key] == $configuredValue`.

If the configured value is an array, then we only remove that one value from the array that matches the value from the option.

Meaning if `[General] trusted_hosts[]="foobar" trusted_hosts[]="barbaz"` is configured, then when using `--section=General --key=trusted_hosts --value=barbaz` then it would remove the last entry from that index. If `--value=helloworld` was used, nothing would be removed because no value matches that entry.

`--section=General --key=trusted_hosts` would remove the entire array.

Does this help?

tsteur:
@javier-de-juan be great to comment once you made the adjustments then we will do another review. Thanks for all this again

tsteur:
@javier-de-juan are you maybe still working on this PR?

sgiehl:
@javier-de-juan do you maybe have some time to apply the review feedback? Would be awesome if we could merge this for the next release.

tsteur:
@javier-de-juan I will close this PR due to inactivity. If you are still keen on it feel free to let us know and we reopen the PR

PR #16610, opened 2020-10-26
only look for done_ok archives when doing usable archive check

### LINKED ISSUE #16594: Invalidated report should be archived
@diosmosis I'm thinking there may be an issue around invalidation.

When executing eg `CoreAdminHomeApi::getInstance()->invalidateArchivedReports($idsite =1, 'today', 'day', false, false);` then it generates 4 entries in the archive invalidations table: 
image

When I then run the CLI archiver, it only processes the first archive for the day. All other ones are ignored in `usableArchiveExists` because it found a usable archive for these periods.  

image

This archive is invalidated though 
image

and therefore it should reprocess the archive and not consider it as usable as such AFAIK.
 
In this case the invalidation was specifically requested and therefore it has to be reprocessed. FYI In my case in Matomo for WordPress I have these TTLs configured:

```
time_before_today_archive_considered_outdated = 1800
time_before_week_archive_considered_outdated = 3500
time_before_month_archive_considered_outdated = 7100
time_before_year_archive_considered_outdated = 14300
time_before_range_archive_considered_outdated = 3600
```

It is different when:

* User tracks data for today
* Thus Matomo stores a record to remember to invalidate a report
* But an existing archive should actually only be invalidated after the configured TTL. 
* After the configured ttl of say 1800s after the report was archived last then it should actually invalidate the report
* And then it should rearchive the report.

This I think works already.

diosmosis:
Fixes #16594

### DIFF
--- a/core/CronArchive/QueueConsumer.php
+++ b/core/CronArchive/QueueConsumer.php
@@ -226,8 +226,10 @@ class QueueConsumer
                 continue;
             }
 
-            if ($this->usableArchiveExists($invalidatedArchive)) {
-                $this->logger->debug("Found invalidation with usable archive (not yet outdated) skipping until archive is out of date: $invalidationDesc");
+            $archivedTime = $this->usableArchiveExists($invalidatedArchive);
+            if ($archivedTime) {
+                $now = Date::now()->getDatetime();
+                $this->logger->debug("Found invalidation with usable archive (not yet outdated, ts_archived of existing = $archivedTime, now = $now) skipping until archive is out of date: $invalidationDesc");
                 $this->addInvalidationToExclude($invalidatedArchive);
                 continue;
             }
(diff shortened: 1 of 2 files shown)


core/CronArchive/QueueConsumer.php
```diff
@@ -544,10 +546,14 @@ public function usableArchiveExists(array $invalidatedArchive)
...
-        $archiveIdAndVisits = ArchiveSelector::getArchiveIdAndVisits($params, $minDateTimeProcessedUTC, $includeInvalidated = true);
+        $archiveIdAndVisits = ArchiveSelector::getArchiveIdAndVisits($params, $minDateTimeProcessedUTC, $includeInvalidated = false);
 
         $idArchive = $archiveIdAndVisits[0];
-        return !empty($idArchive);
+        if (empty($idArchive)) {
+            return false;
+        }
+
+        return Date::factory($archiveIdAndVisits[4])->getDatetime();
```
sgiehl, line 556: Not really related to this PR: But should we maybe change `getArchiveIdAndVisits` to return something more meaningful. Without looking into the method it's hard to understand what the forth element of the returned array actual is. 
Personally I would prefer to have at least named array keys or an object (or class) with defined properties, that would make the code easier to read and to understand...

sgiehl (approved):
Imho the changes look good. But haven't tested it that much

PR #16615, opened 2020-10-27
Automatically add phpunit group name to a plugin php test

### LINKED ISSUE #16612: PHPUnit tests, add group for current plugin automatically
I had enabled travis for a plugin and was wondering why there were no tests being executed. Until after a while I noticed I didn't have the test group set in all plugins (in my case it actually had the old plugin name in there and when renaming the plugin I didn't update the PHPunit `@group` tags). Also sometimes copying tests from a different plugin results in forgetting to adjust the group name 👍 

```php
/**
 * @group MyPluginName
 */
class ApiTest ...
```

It's easily forgotten to add this group name causing tests to not being executed on Travis and then we miss failing tests when we rely on travis.

Is there any chance we could add these tags automatically? I thought this was already the case but seems to not be or maybe it regressed. For example when the classname is `Piwik\Plugins\MyPluginName\tests\Integration\ApiTest` then we could automatically add `MyPluginName` as a group tag?

tsteur:
fix https://github.com/matomo-org/matomo/issues/16612

fixing this here as this way we can restart all plugin tests and make sure they are included in the test runs

### DIFF
--- a/tests/PHPUnit/Framework/TestCase/SystemTestCase.php
+++ b/tests/PHPUnit/Framework/TestCase/SystemTestCase.php
@@ -63,10 +63,9 @@ abstract class SystemTestCase extends TestCase
         $pluginName = explode('\\', get_class($this));
         if (!empty($pluginName[2]) && !empty($pluginName[1]) && $pluginName[1] === 'Plugins') {
             // we assume \Piwik\Plugins\PluginName nanmespace...
-            $pluginName = $pluginName[2];
-        }
-        if (!in_array($pluginName, $groups, true)) {
-            $groups[] = $pluginName;
+            if (!in_array($pluginName[2], $groups, true)) {
+                $groups[] = $pluginName[2];
+            }
         }
 
         parent::setGroups($groups);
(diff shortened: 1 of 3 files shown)


tests/PHPUnit/Framework/TestCase/SystemTestCase.php
```diff
@@ -58,6 +58,19 @@ abstract class SystemTestCase extends TestCase
      */
     public static $fixture;
 
+    public function setGroups(array $groups): void
```
tsteur, line 61: it's officially an `internal` method but seemed for me easiest way to make this work. Not sure there's another way.

tests/PHPUnit/Integration/ReleaseCheckListTest.php
```diff
@@ -38,6 +39,13 @@ public function setUp(): void
         parent::setUp();
     }
 
+    public function test_TestCaseHasSetGroupsMethod()
```
tsteur, line 42: added a test so we would notice when this method is no longer there

sgiehl (approved):

PR #16620, opened 2020-10-28
Add setting to enable/disable use of tracking cookies by default

### LINKED ISSUE #15507: Track cookie less by default

By default, Matomo should not use any cookies ideally. If someone preferred more accurate data (eg unique visitors), they should be enabling cookies.

To not break BC, there could be a system setting in Matomo which is enabled by default for existing installations and disabled for new installations. Depending on that the tracking code be changed. Since the `piwik/matomo.js` might  not be writable for all installations, there's a chance we wouldn't be able to guarantee BC since we do need to modify the tracking code to achieve this. So it would need quite some communication upfront possibly. Mostly, the file should be writable though. Also the generated tracking code would be adjusted depending on this setting so if someone was using WP-Matomo Integration on WordPress where they embed the tracking code into the site through Matomo API, then even if the JS tracker cannot be changed on the filesystem, the tracking code configuration would still enable/disable cookies depending on the setting further reducing the number of people impacted by this.

diosmosis:
WIP

@tsteur this is a poc that works, but I'm a bit unsure about certain parts of the implementation. Can you take a look and review what's here from a product standpoint?

Fixes #15507

### DIFF
--- a/core/Settings/FieldConfig.php
+++ b/core/Settings/FieldConfig.php
@@ -259,4 +259,13 @@ class FieldConfig
      */
     public $validators = [];
 
+    /**
+     * Closure to execute when a value is saved. This can be used to, eg, trigger an event
+     * or regenerate a file, etc.
+     *
+     * The function should accept one argument, the Setting being saved.
+     *
+     * @var \Closure
+     */
+    public $onValueSaved = null;
 }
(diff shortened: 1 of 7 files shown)


tsteur:
closing this PR for now. We'll reopen this and continue when we work on the issue again.

Issue #8413, opened 2015-07-23
Piwik, 2.14.x: Javascript Error: Uncaught URIError: URI malformed with Chrome/IE11 and ISO 8859-1 Encoding


stastka:
If Website has ISO 8859-1 encoding, you have Uncaught URIError Javascript error with Chrome (and IE)  when URL have some special character

Samples with  French character:
URL: /Les-droits-r%E9els
decode: /Les droits réels

or with German Umlaut

/Hasenb%F6hler-Franz
decode:  Hasenböhler-Franz

Detailed Error: 
Uncaught URIError: URI malformedD @ piwik.js:30(anonymous function) @ piwik.js:51(anonymous function) @ piwik.js:52

I used latest Piwik, 2.14.x

mattab:
Hi @stastka 
Thanks for the report.

Let's try to reproduce and then:
 * add a test showing the bug in tests/javascript/
 * fix the issue

stastka:
Hi @mattab 

I create a small testpage to reproduce. The site contains links with french and german characters with ISO 8859-1 encoding. :

http://stastka.ch/piwiktest/

meyway:
Line 2140 of the current piwik.js:
locationHrefAlias = decodeWrapper(locationArray[1])

"decodeWrapper" is "window.decodeURIComponent" and "decodeURIComponent" expects UTF-8.

Since "locationArray[1]" holds "window.location.href" and since that contains the ISO-8859-... - characters, current piwik (IMHO) is not fully compatible with websites that encode Links in ISO-8859-...

Is the use of "unescape" instead of / before using "decodeURIComponent" a solution to the problem?

stastka:
Seems to resolve the issue possible.
I would test a patch on my system

stastka:
@tsteur @mattab
I testet the patch e8bde681ec380b53ed15fa14a994176dffa5d58f and i have no error. The visitor log on Piwik show the visiting Test-Page. The Patch resolve the issue.

tsteur:
Thx for the feedback!

Issue #7220, opened 2015-02-16
piwik.js double encodes the url and referrer when they are detected automatically


tsteur:
`location.href` and `top.document.referer` seem to be already URL encoded. At least in the browsers we tested. When tracking, we do encode them again meaning they are double url encoded which is wrong. Even worse some browsers use uppercase and some browsers use lowercase for the url encoding meaning we store 2 actions for the same URL. See #7209 . It also causes that it is not possible to segment for those URLs see #7218 

Solution should be most likely to always url-decode the url and referrer before url-encoding them again (which is done when the request is sent)

Issue #7218, opened 2015-02-16
Segmentation for url-encoded page urls does not work


tsteur:
Our segmentation logic does always url decode all values so it will never match a url-encoded url like `piwikjapan.org/%E6%A9%9F%E8%83%BD%E8%AA%AC%E6%98%8E/4695`.

If I understand it right one problem is that some actions are stored raw while some other actions are stored URL encoded.

tsteur:
Actually all is good. In this case it was an error that the URL was wrongly tracked (user error). If the URL is stored URL encoded (which happens when sending them double url encoded), then one also has to build a segment filter that is double url encoded.

cosmoblonde:
The error is still there. Testcase: When creating a segment with an "or" combination of 2 or more page URLs - directly after creation of the seqment the report is fine but when revisiting the seqment at a later time it shows: no data for this report.

When copying the generated url into a urldecoder-Tool, decode it and paste it again into the browser - the seqment works. So the double encoding seems to be still an issue. 
Seqments with only one condition work.
 
Tested with Piwik 3.01 and Firefox 51.01

mattab:
@cosmoblonde please try again in Piwk 3.0.2 where we fixed an issue in segmentation. if you have a problem still, open a new issue

PR #16628, opened 2020-10-29
Ensure current url is not decoded too often

### LINKED ISSUE #14715: Site search is truncating the strings which is placed after ampersand (&)
In Site Search, part of keyword is getting ignored/truncated, when there is an ampersand (&) in the keyword (ex: Heatmap & session recording)

Expected: Report should contain the word "Heatmap & session recording"

Result: Can only see "Heatmap" in Visits log and also the Site Search report

Steps to reproduce:
1.     Search for "Heatmap & session recording"
2.     check the Visits Log and Site Search

**Example site search:**
site search error

**Visits log:**
site serach error visits log

### LINKED ISSUE #15060: Internal site serach decoding bug
We have site in russian language in windows-1251 codepage and all working fine except internal site serch keywords. 

E.g. when I try to find ‘окрб’ this keyword turns to url encoded ‘q=%EE%EA%F0%E1’ and then in database it turns to some wierd encoded string ’ Г®ГЄГ°ГЎ’.

There is small 2-HTML page site (windows-1251) where the problem can be reproduced:
[test.zip](https://github.com/matomo-org/matomo/files/3770797/test.zip)

Matomo 3.11.0
Thanks!

sgiehl:
Let's assume we have the search url on the marketplace containing a `&` in search query (see #14715):
`https://plugins.matomo.org/search?query=Heatmap+%26+session+recording&post_type=product`

Currently Matomo tracker js would run a `safeDecodeWrapper` on it. So the resulting url, that would be used for tracking would be 
`https://plugins.matomo.org/search?query=Heatmap+&+session+recording&post_type=product`, which is actually not the same

fixes #14715


sgiehl:
build js

sgiehl:
build js

tsteur:
We might need to check if there's any other way to fix this as this was added because of https://github.com/matomo-org/matomo/issues/8413 in https://github.com/matomo-org/matomo/pull/8510

Anyway, let's for now focus on Matomo 4 release etc.

sgiehl:
build js

sgiehl:
@tsteur This might actually fix some more issues that are related to encoding and page charset.
It shouldn't make any problems like #8413 or #8510, as before `decodeURIComponant` was used, which throwed the error. Now the url would not be decoded. I'm actually still wondering why that should be needed at all. Are there any cases where the `location` object might hold any values that are double encoded?

tsteur:
@sgiehl I think the original urldecode was added because of https://github.com/matomo-org/matomo/issues/7220 and maybe also https://github.com/matomo-org/matomo/issues/7218 . If we now removed that method, we'd have a problem that URL's would be likely stored URL encoded in the actions meaning it will create a new ID action for every URL and it can cause a lot of problems eg if some reports don't group by URL etc. Might need to test this but there's a chance we can't merge this.

sgiehl:
I guess since #7220 a lot has happened in browser development. Seems the url and referrer are no longer url encoded in modern browsers

tsteur:
@sgiehl do you have maybe some reference to that where browsers changed that? I just tested with some URLs and it seems to be still URL encoded. I would assume it could break a lot of things if browsers simply changed that but it's possible they did.

eg viewing `/test.html?x=34L#$3L#4L1>124:56:7&test` then some characters were encoded 
image

sgiehl:
Ok. Now I understand what you are referring to. That's still normal that some special characters are encoded. But it's actually not the full url. So we shouldn't manipulate the complete url using any decodeWrapper.
Let's assume you request something like `test.html?q=test%26q%3Dp`, which is a valid request with one query parameter.
After decoding it you get `test.html?q=test&q=p`, which is obviously something else.

diosmosis:
@sgiehl I tested this PR against 4.x-dev using https://github.com/matomo-org/test-examples/pull/9 to track encoded urls and noticed there are changes in what's tracked now.

For ISO-8859-1 encoded pages:

encoding1

The first action is tracked via 4.x-dev, and the second via this branch.

For other URL encoding:

(4.x-dev encoding)
4 x-dev-urlencoding-test

(this branch encoding)
fixurl-encoding-test

In one case it fixes the bug referenced in the PR, but for others it changes what's tracked. I'm not sure exactly what's ok to change (cc @tsteur), but it might break some segments if we just merge this.

I didn't test referrer URL tracking, but that would be affected too.

sgiehl:
I'm actually not sure regarding this. But a special character like `ö` in an URI is actually not correct and even a browser would convert it to `%C3%B6` in requests, so the tracked url would actually be correct. But (like most browsers) we should maybe show the readable value in UI.
Another possibility would be to adjust the JS so only characters not representing a url reserverd character will be decoded. That would mean we would need to find a way to not decode `%26` to `&` and maybe also `%2B` to `+`. But not sure if that would fix all potential problems with the url manipulation we are currently doing

diosmosis:
I think for a bug fix it would be better to maintain existing behavior if possible. We can't know exactly how much might be affected downstream, it could be more than just what is displayed to the user (eg, segments no longer matching).

I'm also noticing the first test doesn't save the file in test-examples in ISO-8859-1. Not sure exactly how to do that.

tsteur:
@sgiehl as in https://github.com/matomo-org/matomo/issues/15060 the user mentions it's only a problem for site search. Is there a chance to only change this behaviour for site search but not the entire URL? (not needed to look into it for long but maybe you know )

sgiehl:
@tsteur don't think that is possible. We parse the page url to check if some parameters for site search are contained, so if we don't change the tracked page url, site search will stay the same buggy way.

tsteur:
👍  thanks. moving the issue to 4.7 for now as this current fix would likely introduce quite a few regressions (I remember when we added them and there were various regressions re encoding). We'll look into it again when it's a higher priority.

justinvelluppillai:
@tsteur milestone 4.7.0 has arrived, are you wanting this in this milestone now or still deferring for later?

justinvelluppillai:
hmm looks like maybe it is in 4.8.0 since the milestone got renamed or something, so still deferred 👍

tsteur:
@sgiehl be great to explore some other options if there are other ways to fix this in 4.8 milestone without possible regressions

Anisatum:
> @sgiehl be great to explore some other options if there are other ways to fix this in 4.8 milestone without possible regressions

As far as I can tell there's no way to fix it without affecting the logic that was in effect until now. Any custom logic to try and recreate the original encoding would be pretty complex, and likely to introduce new bugs in some unforeseen special case.

The only way out that comes to mind is adding a setting so each user can decide whether they want this fix applied (off by default for existing instances of Matomo, on by default when a new one is created).

Anisatum:
So in the end I agree with @sgiehl that decoding by default is wrong, as it will necessarily lead to the loss of information, if a URL has encoded components, but I also agree that a bug that remains unfixed for long enough becomes expected behavior, and a bug fix shouldn't change it too much. I figured the best way to do this would be to create a tracker plugin, this way if someone wants to keep the old behavior in it's entirety, they can just not install it / keep it disabled. A proof of concept is available [in the feature/disable_url_decoding branch of my fork](https://github.com/Anisatum/matomo/tree/feature/disable_url_decoding).

The decode-disabling can be triggered either from Javascript with a `_paq.push(['DisableUrlDecoding.doNotDecode']);` or globally with a SystemSetting. The Javascript tracker overrides the SystemSetting, so there is also a `_paq.push(['DisableUrlDecoding.doDecode']);` function for the case that decoding is disabled in SystemSettings, but the user wants re-enable it in specific conditions.

Right now all there is in the settings, is a simple on/off switch, but I was thinking of providing extra options, like:
- Only disable for URLs matching a specific Regex
- Only disable the decoding of specific query parameters

But I wanted to get some feedback before I go ahead.

PR #16631, opened 2020-10-30
Update link to demo

tsteur:
We're changing it cloud since it's faster there (no proxysite in the middle)

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ If you do not have a server or don't want to host yourself you can use our Matom
 
 ## Online Demo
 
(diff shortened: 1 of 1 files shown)


sgiehl (approved):

sgiehl:
There are actually a few more occurrences of the demo url in the code. Not sure if it should be updated anywhere else...

tsteur:
@sgiehl it should be fine for now. There's the prefixUrl in some template but not sure what it is used for etc. Can change later if needed

PR #16634, opened 2020-10-30
Make extra argument to reArchiveReport start date.

diosmosis:


diosmosis:
FYI @tsteur for AbTesting change

tsteur:
`Date` should work too. I thought initially `number of days as integer` but this way using `Date` should be easier since we have the data in the DB most of the time and won't need to do any math 👍

PR #16636, opened 2020-10-30
Require new major release of matomo/device-detector

sgiehl:
(instead of piwik/device-detector)


sgiehl:
Some classes have been renamed and some internal handlings have changed in latest major release, so there were some more adjustments needed here...

tsteur:
Didn't really have a look but lgtm. Note to myself: need to update this in cloud as well

PR #16640, opened 2020-10-31
require PIWIK_PRINT_ERROR_BACKTRACE for exception backtraces in logs

diosmosis:


plugins/Monolog/Processor/ExceptionToTextProcessor.php
```diff
@@ -76,6 +76,10 @@ private function getMessage($exception)
 
     private function getStackTrace($exception)
     {
+        if (!\Piwik_ShouldPrintBackTraceWithMessage()) {
```
tsteur, line 79: @diosmosis can we also check all other places where we use `getTraceAsString()`?

core/Common.php
```diff
@@ -300,9 +300,8 @@ public static function safe_unserialize($string, $allowedClasses = [], $rethrow
             }
 
             $logger = StaticContainer::get('Psr\Log\LoggerInterface');
-            $logger->debug('Unable to unserialize a string: {message} (string = {string})', [
-                'message' => $e->getMessage(),
-                'backtrace' => $e->getTraceAsString(),
+            $logger->debug('Unable to unserialize a string: {exception} (string = {string})', [
+                'exception' => $e,
```
sgiehl, line 304: Doesn't using the exception object here cause a cast to string when it's used in the message? And __toString will show the message and the (full) backtrace. Or does the Monologger do some magic here to use the ExceptionToTextProcessor?

diosmosis: to my understanding, ExceptionToTextProcessor should convert the 'exception' value to a string

plugins/Monolog/Processor/ExceptionToTextProcessor.php
```diff
@@ -76,17 +81,34 @@ private function getMessage($exception)
...
+                $trace = self::replaceSensitiveValues($trace);
+                return $trace;
+            } else {
+                return self::BACKTRACE_OMITTED_MESSAGE;
+            }
+        }
+
         if (!$shouldPrintBacktrace) {
-            return $exception->getMessage();
+            $message = $exception->getMessage() . "\n" . self::BACKTRACE_OMITTED_MESSAGE;
```
tsteur, line 109: fyi the message again might not be needed? In my case it causes the exception message to be shown twice 
image

At least it happened when I triggered an exception in `Common::safe_unserialize()` and thne it went into the `ExceptionToTextProcessor::__invoke`. I reckon the trace does not need to append the message again

tsteur: I see it might be needed because of eg https://github.com/matomo-org/matomo/pull/16640/files#diff-7aeb867200acd3d0c23432f4b94edae59f9549e2b20160b83bd165830e5e2042L208 

The method name might be bit misleading as it gets the message and the trace?

PR #16641, opened 2020-10-31
read from the task timetable option more often when executing tasks...

diosmosis:
… to observe less concurrency issues

Refs https://github.com/matomo-org/plugin-CustomAlerts/issues/67

### DIFF
--- a/core/Scheduler/Timetable.php
+++ b/core/Scheduler/Timetable.php
@@ -24,10 +24,7 @@ class Timetable
 
     public function __construct()
     {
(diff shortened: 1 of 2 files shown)


core/Scheduler/Timetable.php
```diff
@@ -143,4 +140,12 @@ public function taskHasBeenScheduledOnce($taskName)
     {
         return isset($this->timetable[$taskName]);
     }
+
+    public function readFromOption()
```
sgiehl, line 144: Maybe something like `refresh` would be a better name.

Issue #14601, opened 2019-07-01
Release Matomo 4.0.0


mattab:
#### Work on Beta cycle, towards Release Candidate 1:

- [x]    Beta release cycle has started with the first beta release https://github.com/matomo-org/matomo/issues/16165
- [x] **Keep track of releases notes/ any important changes in this wiki page: https://github.com/matomo-org/matomo/wiki/Matomo-4.0.0-release-notes**
- [x] Make a new release of PHP SDK Tracker (~there is a 4.x-dev branch there~ 4.x-dev merged into master, 3.0.0 released and 4.x-dev matomo using it)
- [x] Check [this list of Pull requests all Matomo repositories](https://github.com/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Apr+org%3Amatomo-org+org%3Ainnocraft+milestone%3A%22Current+sprint%22+), where the milestone is set to `Current sprint`
- [x] Triage [all PRs](https://github.com/pulls?utf8=%E2%9C%93&q=is%3Aopen+is%3Apr+org%3Amatomo-org+org%3Ainnocraft+no%3Amilestone) not scheduled yet, or close (see also [issues without milestone](https://github.com/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+org%3Amatomo-org+org%3Ainnocraft+no%3Amilestone))
- [x] Check all 3.x and 4.x branches include the same changes in all plugins
- [x] Check each plugin has a release for the most recent changes in 4.x-dev
- [ ]   Tag a new release for `matomo-org/referrer-spam-blacklist` and `matomo-org/searchengine-and-social-list`
- [x]  Fetch our latest and greatest home-made libraries with `composer update matomo-org/referrer-spam-blacklist matomo-org/searchengine-and-social-list matomo-org/matomo-php-tracker matomo-org/component-network,matomo-org/component-ini,matomo-org/component-decompress,matomo-org/component-cache`
- [x]    Fetch the latest composer libraries for all with `composer update`
- [x]   Up to date submodules
- [x]  Make [CI builds](https://travis-ci.com/matomo-org/) of our official plugins `matomo-org/plugin-*` all green 
- [x]  Tag the RC candidate 
- [x]   Check that the build is green
- [x]   [Package & release](https://github.com/matomo-org/matomo-package#core-matomo-package) the Release Candidate
[...]

PR #16649, opened 2020-11-02
Updates composer dependencies

sgiehl:
- Removing symfony/polyfill-util (v1.17.0)
  - Removing symfony/polyfill-php56 (v1.17.0)
  - Removing nikic/php-parser (v4.4.0)
  - Removing jeremeamia/superclosure (2.4.0)
  - Updating composer/ca-bundle (1.2.7 => 1.2.8)
  - Updating maxmind/web-service-common (v0.7.0 => v0.8.0)
     * PHP 7.2 or greater is now required.
     * Added additional type hints.
  - Updating maxmind-db/reader (v1.7.0 => v1.8.0)
     * Fixes for PHP 8.0
  - Updating geoip2/geoip2 (v2.10.0 => v2.11.0)
     * IMPORTANT: PHP 7.2 or greater is now required.
     * Added the `isResidentialProxy` property to `GeoIp2\Model\AnonymousIP` and `GeoIp2\Record\Traits`.
     * Additional type hints have been added.
  - Updating phpmailer/phpmailer (v6.1.7 => v6.1.8)
     * Mark ext-hash as required in composer.json. This has long been required, but now it will cause an error at install time rather than runtime, making it easier to diagnose
     * Make file upload examples safer
     * Update links to SMTP testing servers
     * Avoid errors when set_time_limit is disabled (you need better hosting!)
     * Allow overriding auth settings for local tests; makes it easy to run tests using HELO
     * Recover gracefully from errors during keepalive sessions
     * Add AVIF MIME type mapping
     * Prevent duplicate To headers in BCC-only messages when using mail()
     * Avoid file function problems when attaching files from Windows UNC paths
     * Improve German, Bahasa Indonesian, Filipino translations
     * Add Javascript-based example
     * Increased test coverage
  - Updating symfony/polyfill-ctype (v1.18.1 => v1.20.0)
  - Updating symfony/polyfill-mbstring (v1.18.1 => v1.20.0)
  - Updating twig/twig (v3.0.5 => v3.1.1)
      * Fix "include(template_from_string())"
      * Fix sandbox support when using "include(template_from_string())"
      * Make round brackets optional for one argument tests like "same as" or "divisible by"
      * Add support for ES2015 style object initialisation shortcut { a } is the same as { 'a': a }
  - Updating pear/archive_tar (1.4.9 => 1.4.10)
  - Updating opis/closure (3.5.6 => 3.6.0)
  - Updating php-di/phpdoc-reader (2.1.1 => 2.2.1)
  - Updating phpdocumentor/type-resolver (1.3.0 => 1.4.0)
  - Updating phpdocumentor/reflection-docblock (5.2.1 => 5.2.2)
  - Updating phpspec/prophecy (1.11.1 => 1.12.1)

refs #14601

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "composer/ca-bundle",
-            "version": "1.2.7",
+            "version": "1.2.8",
             "source": {
                 "type": "git",
                 "url": "https://github.com/composer/ca-bundle.git",
-                "reference": "95c63ab2117a72f48f5a55da9740a3273d45b7fd"
+                "reference": "8a7ecad675253e4654ea05505233285377405215"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/composer/ca-bundle/zipball/95c63ab2117a72f48f5a55da9740a3273d45b7fd",
-                "reference": "95c63ab2117a72f48f5a55da9740a3273d45b7fd",
+                "url": "https://api.github.com/repos/composer/ca-bundle/zipball/8a7ecad675253e4654ea05505233285377405215",
+                "reference": "8a7ecad675253e4654ea05505233285377405215",
                 "shasum": ""
             },
             "require": {
@@ -66,11 +66,15 @@
                     "type": "custom"
                 },
                 {
+                    "url": "https://github.com/composer",
+                    "type": "github"
+                },
+                {
                     "url": "https://tidelift.com/funding/github/packagist/composer/composer",
                     "type": "tidelift"
                 }
             ],
-            "time": "2020-04-08T08:27:21+00:00"
+            "time": "2020-08-23T12:54:47+00:00"
         },
         {
             "name": "composer/semver",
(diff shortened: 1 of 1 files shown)


Findus23:
> Mark ext-hash as required in composer.json. This has long been required, but now it will cause an error at install time rather than runtime, making it easier to diagnose

Should we maybe add this here:
https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/plugins/Diagnostics/Diagnostic/PhpExtensionsCheck.php#L58

sgiehl:
@Findus23 let's do that in a new PR. Guess we could also remove `Reflection` and `SPL`,  from that list, as they are actually built in.

sgiehl:
@Findus23 just checked that. The availability of the hash function is checked here: https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/Diagnostics/Diagnostic/PhpFunctionsCheck.php#L61

Findus23:
Thanks for checking. I was confused as they called it `ext-hash`.

tsteur:
@sgiehl @Findus23 can we merge it? tests seem to pass?

sgiehl:
@Findus23 Maybe the check would be better placed in the extension check. But at least it is checked in any way...

@tsteur Guess it can be merged then

Issue #11237, opened 2017-01-19
 Call to undefined function Piwik\fnmatch() 


bluesailor45:
after upgrade to PIWIK 3.0.1 in Joomla, following error appears:
fatal error in PIWIK 3.0.1
Call to undefined function Piwik\fnmatch() in /share/MD0_DATA/.qpkg/Piwik/Piwik/core/FileIntegrity.php 312 using PHP 5.5.26

thereafter it is not possible to logon to PIWIK:
------------------------------
a fatal error occurred:
Please contact the system administrator, or login to Piwik to learn more.

If you are Super User, but cannot login because of this error, you can still troubleshoot further. Follow these steps: 
1) open the config/config.ini.php file and look for the salt value under [General]. 
2) edit this current URL you are viewing and add the following text (replacing salt_value_from_config by the salt value from the config file): 
index.php?i_am_super_user=salt_value_from_config&....
-----------------------------
However this 2) message is not useful for me.

CheweyZ:
What OS are you running this on?

bluesailor45:
Hi CheweyZ
I am running Linux-Firmware Version 4.2.1  in QNap TS-219P. In Joomla I am using Joomla 3.6.5 with PHP 5.5.26. I was upgrading PIWIK from Version 2.6.1 which was installed as a QNAP Application.
Thanks for your help.

bluesailor45:
No solution to that 😳???

CheweyZ:
So not 100% sure but it appears your system is none POSIX compliant which means that fnmatch is not able to run on your machine. To find out more see or confirm you machine-distro is a POSIX supported system as then php call only works on that.

Sorry for the little help.

bluesailor45:
Thank you very much for your hint. Any tip is welcome😉. I will check in this direction.

tsteur:
@bluesailor45  can you try to copy something like this in your "libs/upgradephp/upgrade.php"

```php
if(!function_exists('fnmatch')) {

    function fnmatch($pattern, $string) {
        return preg_match("#^".strtr(preg_quote($pattern, '#'), array('\*' => '.*', '\?' => '.'))."$#i", $string);
    } // end

} // end if
```

I haven't tested it and only copied it from http://php.net/manual/en/function.fnmatch.php

We would need to define this method then in the future if it is not defined

bluesailor45:
@tsteur thank you very much for your help.
I added your suggested code add the end of upgrade.php and it did work!

[...]

PR #16650, opened 2020-11-02
Remove upgradephp

Findus23:
- [x] utf8_enode/utf8_decode
should be fine as it is a core part of PHP 7.2+
- [x] safe_unserialize
should be possible to replace with Common::safe_unserialize() in all cases
- [x] safe_serialize
???
- [x] \Error class
exists in PHP 7+
- [x] PHP_INT_SIZE/PHP_INT_MAX 
exist since PHP 5.0.5
- [x] mysqli_set_charset
exist since early PHP 5
- [ ]  `parse_ini_file()` 
 as matomo-org/component-ini should handle this well. Might be an issue if `parse_ini_file` is used anywhere directly (but shouldn't be the case. Maybe also check if this implementation had fixes that are missing in the component
- [ ] glob/_glob
is used in many places in Matomo. `glob` is checked as an optional function in the system check
- [x] file_get_contents
I'm a bit confused by this polyfill. Wouldn't this break checks if `file_get_contents` was disabled and this was used instead to download files?
- [ ] _readfile
Is used by `ProxyHttp::serverStaticFile`, but only advantage seems to be supporting partial files (which is used nowhere in Matomo)
- [x] mb_strtolower/mb_strlen
possibly still needed. Maybe find a better place for them
- [ ] gzopen64
this seems to only be a bug with a specific 5.x release
- [ ] dump
it seems like this is used to silence the dump() function from development (var-dumper) during release, but it is used nowhere
- [ ] fnmatch
most likely still needed to avoid #11237

This is mostly a proof of concept. It would be great if someone who knows more about why these functions exist could take over.


sgiehl:
> safe_serialize

Guess we could simply use `serialize` instead, as we always should use `safe_unserialize`, which then should block unseralizing bad objects. Or we introduce a `Common::safe_serialize`, that similar to `safe_unserialize` takes `allowed_classes` as new param and blocks all other objects from being serialized.

sgiehl:
> mb_strtolower/mb_strlen
> possibly still needed. Maybe find a better place for them

I'd suggest to require `symfony/polyfill-mbstring` instead

Findus23:
> I'd suggest to require symfony/polyfill-mbstring instead

That's a good idea. As it is already a dependency of twig adding it explicitly shouldn't even add any code.

sgiehl:
@Findus23  might maybe make sense to split that PR into multiple parts and remove that stuff step by step.

sgiehl:
> I'm a bit confused by this polyfill. Wouldn't this break checks if file_get_contents was disabled and this was used instead to download files?

I've just checked that one. If `file_get_contents` is added as disabled function, this piece of code actually crashes with
```
PHP Fatal error:  Cannot redeclare file_get_contents() in /srv/matomo/libs/upgradephp/upgrade.php on line 137
```

PR #16653, opened 2020-11-02
don't install unneeded polyfills with composer

Findus23:
see https://github.com/symfony/polyfill/blob/main/README.md#design

This removes `symfony/polyfill-php56` on the next `composer install`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,6 +59,14 @@
         "symfony/var-dumper": "~2.6.0",
         "symfony/yaml": "~2.6.0"
     },
+    "replace": {
+        "symfony/polyfill-php54": "*",
+        "symfony/polyfill-php55": "*",
+        "symfony/polyfill-php56": "*",
+        "symfony/polyfill-php70": "*",
+        "symfony/polyfill-php71": "*",
+        "symfony/polyfill-php72": "*"
+    },
     "repositories": [
         {
             "type": "package",


sgiehl:
That package actually was already automatically removed with a composer update. See #16649

Findus23:
Ah, I didn't see that PR. 
We could still add those lines in case any other dependency adds polyfills.

sgiehl:
Yes. Shouldn't hurt to prevent installing those polyfills nevertheless

sgiehl (approved):

tsteur:
@Findus23 @sgiehl the tests are failing see https://travis-ci.com/github/matomo-org/matomo/jobs/424314293

sgiehl:
guess merging in the latest changes should fix it...

tsteur:
there's a test now failing. not sure if it is due to this? https://travis-ci.com/github/matomo-org/matomo/jobs/424766921#L925 @sgiehl @Findus23

Findus23:
The same failure seems to be on 4.x-dev without this PR: https://travis-ci.com/github/matomo-org/matomo/jobs/424766869#L921

sgiehl:
That was caused by the cookie PR. See https://github.com/matomo-org/matomo/pull/14444#pullrequestreview-521623030

PR #16660, opened 2020-11-03
Add timezone version to php informational system check output

### LINKED ISSUE #16657: Website time zone setting for Canada - Whitehorse and Canada - Dawson does not observe now-permanent UTC-7
As of March 8, 2020, the Yukon Territory of Canada no longer changes its clocks to observe different standard or daylight saving times. That means Yukon is now permanently UTC-7 (the same as MST at the moment). See https://yukon.ca/time for details.

The "Canada - Whitehorse" and "Canada - Dawson" time zone regions are both in Yukon and should observe that change.

_Please note that "Canada - Dawson Creek" is not in Yukon and not part of this time zone change._

Our Matomo installation insists that Canada - Whitehorse is on Pacific Standard Time (UTC-8) even though the underlying OS has the update for the new Yukon Standard Time and is configured as "America/Whitehorse". This means that web hits are tagged one hour behind the time in Yukon.

The workaround for now is to use the Phoenix, Arizona time zone setting as most of Arizona observes year-round UTC-7.

The same adoption of year-round daylight saving time is expected for British Columbia, Washington, Oregon, and hopefully California before long.

Please update the time zone policy for "Canada - Whitehorse" and "Canada - Dawson" to be permanently set to UTC-7.

Thanks from Canada's North!

tsteur:
### Description:

fix https://github.com/matomo-org/matomo/issues/16657

Information may be useful when troubleshooting some issues. Only needed though when Matomo is installed

### Review

* [x] Functional review done
* [x] Security review done
* [x] Code review done


sgiehl (approved):

sgiehl:
@tsteur would be good to merge. Not sure if that topic might be worth to create a small FAQ for. Something like `What can I do if my selected timezone doesn't use the correct offset`

tsteur:
@sgiehl had a quick look on how to do this but doesn't seem straight forward or can't find any good info on this. Seems the best/fastest way is typically to update the PHP version? see eg https://www.generacodice.com/en/articolo/838267/How-to-update-timezonedb-in-PHP-(updating-timezones-info) unless I'm missing something?

sgiehl:
Instead of updating PHP itself it's also possible to install/update the pecl package `timezonedb`

tsteur:
created https://matomo.org/faq/troubleshooting/how-do-i-fix-if-a-timezone-that-doesnt-use-the-correct-offset/

PR #16661, opened 2020-11-03
add href in link in update message email

aloxe:
### Description:

The link in update email is not clickable. This makes it clickable

### Review

* [ ] Functional review done
* [ ] Security review done
* [ ] Code review done

### DIFF
--- a/plugins/Marketplace/templates/_updateCommunicationEmail.twig
+++ b/plugins/Marketplace/templates/_updateCommunicationEmail.twig
@@ -10,14 +10,14 @@
 {% if hasThemeUpdate %}
 <p>
 {{ 'CoreUpdater_NotificationClickToUpdateThemes'|translate }}<br/>
-<a href="{{ host }}index.php?module=CorePluginsAdmin&action=themes">{{ host }}index.php?module=CorePluginsAdmin&action=themes</a>
+<a href="{{ host|e('html_attr') }}index.php?module=CorePluginsAdmin&action=themes">{{ host }}index.php?module=CorePluginsAdmin&action=themes</a>
 </p>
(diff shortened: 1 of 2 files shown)


sgiehl (changes requested):
@aloxe  Thanks for creating the PR. Seems you actually only changed a test file. So that wouldn't have any effect. To adjust the mails you need to update this file: https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/Marketplace/templates/_updateCommunicationEmail.twig

aloxe:
Hope this is good now.

plugins/Marketplace/templates/_updateCommunicationEmail.twig
```diff
@@ -10,14 +10,14 @@
 {% if hasThemeUpdate %}
 <p>
 {{ 'CoreUpdater_NotificationClickToUpdateThemes'|translate }}<br/>
-{{ host }}index.php?module=CorePluginsAdmin&action=themes
+<a href="{{ host }}index.php?module=CorePluginsAdmin&action=themes">{{ host }}index.php?module=CorePluginsAdmin&action=themes</a>
```
tsteur, line 13: @aloxe we probably need to escape the first `{{ host }}` like this: `{{ host|e('html_attr') }}`. Also in the link below for better security.

aloxe: sorry for the delay, I haven't noticed this comment. I pushed the change just now.

tsteur (approved):

diosmosis:
Thanks for the contribution, @aloxe !

PR #16668, opened 2020-11-05
Always set consent removed cookie even when cookies are disabled

tsteur:
backport https://github.com/matomo-org/matomo/pull/16507
fyi @Findus23


tsteur:
build js

sgiehl (approved):

tsteur:
Just seeing a test is now failing. Would need to look into it before we have another 3.X release

tsteur:
failing test is actually not related

PR #16673, opened 2020-11-05
Move performance reports to a new page / category

### LINKED ISSUE #16669: Make new page performance report more visible
We should make this useful report more visible. 

I wonder how to best do it as there are various ways. 

1. We could show another report on each of the "pages", "entry pages", "exit pages", "page titles" report below the existing report showing the performance. It be ok but it may be still hard to find and notice it's there.
2. I reckon we should add a new page in the reporting menu under "Behaviour" named "Performance". Could put it maybe below Engagement or so.
 * The question is then do we show only one report and have "related reports" to switch eg from page url to page title
 * Or should we have both visible one after another similar to [media analytics](https://demo.matomo.cloud/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=day&date=yesterday&segment=&category=MediaAnalytics_Media&subcategory=MediaAnalytics_TypeVideo)   Both have advantages/disadvantages like related reports are hard to find and easy to miss but showing only one report lets you focus more on this one report maybe and keeps the page simpler. I think either solution be fine for now and we can always change it very quickly.

I would still keep the existing visualisation on the other pages reports so you can use it on "entry pages" and "exit pages". Unless we use "related reports" on this new Performance reporting page and then also show "entry pages" and "exit pages" as a related report (meaning all 4 reports where the performance report currently works). Then it would not be needed anymore to have it as a visualisation.

@sgiehl could you look into this? We'd also need to adjust the FAQ then
@mattab any thoughts?

sgiehl:
### Description:

This PR creates a new page "Performance" as subcategory of Behaviour.
It contains the performance overview (moved from visitor overview) and the two reports "Pages" and "Page Titles".  Both reports have links to the Entry and Exit page reports as related reports.

Note: Exit and Entry page title reports didn't have page performance numbers before, even though they were already calculated. So I've activated them for those reports.

fixes #16669

### Review

* [ ] Functional review done
* [ ] Security review done
* [ ] Code review done


tsteur:
@sgiehl could you move Performance further up say after Transitions? 
image

I think the side by side reports don't work so well since there are so many columns

image

could we show the reports below each other? There should be some flag somewhere like `wide...`. Don't remember the name right now.

Only did a quick functional check.

sgiehl:
@tsteur Updated the PR accordingly.

Note: UI screenshots in some submodules needs to be updated before merging.

tsteur (commented):
LGTM. 

If easily possible (might be tricky) could we remove the Performance visualisation from the "Actions -> Pages" report etc? Just to not have two ways to get this information and to keep things simpler and have less possible issues with this in the future. it might be difficult because they all use the same API so not too important if it's not easy.

If it's easy to do could also rename it to `Page URLs` just to have the differentiation more clear as `Page Titles` is on the same page. I know in the menu we also call it `Pages`. Just thinking it would make it more clear and maybe should rename it in the menu eventually as well. Not too important though only if it's like no effort.
image

Feel free to merge and then would directly need to update the FAQ. That would be great, thanks

sgiehl:
@tsteur I can have a look if that would be easy doable. Should the row action be still in the normal action reports or only on the performance reports. Also that would apply for the dashboard widgets, so we would need to make the performance page widgets available as well (as those reports would otherwise not be accesible on the dashboard) or allow switching to the page performance viz in the widgets only.
Also keep in mind that the performance viz is also available on the "pages (titles) following site search" reports. Not sure if we should move them to the related reports then as well...

tsteur:
I reckon row action could be actually useful in all the pages reports, not just in the performance reports.

I reckon we don't need to have "pages (titles) following site search" as a related report for now.

sgiehl:
@tsteur Ok. So the performance viz is now only hidden on the page reports pages. The page report widgets will still have the possibility to switch to this viz. Let me know if it's fine that way. Will update the expected ui files in the submodules then and merge the PR...

tsteur:
Sounds good @sgiehl feel free to merge once the tests pass

PR #16680, opened 2020-11-08
Fix spelling of "Pie chart"

danieljwonder:
Pie chart should be two words instead of one.

### Description:

@mattab advised to open issue re pie chart wording. 

### Review

None. This looks like it should work, but not sure how you specifically manage string translation.


Findus23:
Hi,

This should work. Just keep in mind that this invalidates all translations meaning it should not be merged shortly before a release when translators don't have time to catch up.

sgiehl:
I'll merge that now and will update the invalidated translations on transifex manually. As the translation would actually all still be valid it imho doesn't make sense to wait until they get translated again.

Issue #16396, opened 2020-09-06
Take advantage of plugin specific archiving


tsteur:
@diosmosis could you maybe look at this one? refs https://github.com/matomo-org/matomo/issues/11974

* [x] **RELEASED**`Custom reports` (not sure if done already both in create and update custom reports). Should only rearchive the specific report
    PR: https://github.com/innocraft/plugin-CustomReports/pull/92
* [x] **RELEASED** `Users Flow`: on activation it should archive all it's reports
    PR: https://github.com/innocraft/plugin-UsersFlow/pull/43
* [x] **RELEASED**`Multi Channel Conversion Attribution`: on activation it should archive all it's reports 
    PR: https://github.com/innocraft/plugin-MultiChannelConversionAttribution/pull/27 
* [x] **RELEASED**`Funnels`: When adding or updating a funnel rearchive that specific funnel
    PR: https://github.com/innocraft/plugin-Funnels/pull/70
* [x] **RELEASED**`AbTesting`:  When updating (not adding) an existing A/B test then it should rearchive the data for A/B testing. Ideally it would do this back until `abtest['created_date']`. (which is often just a few days or weeks instead of multiple months)
    PR: https://github.com/innocraft/plugin-AbTesting/pull/63
* [x]  **RELEASED** `Cohorts`: On activation it should rearchive all it's reports (be good to confirm @diosmosis )
    PR: https://github.com/innocraft/plugin-Cohorts/pull/19

diosmosis:
@tsteur fyi, AbTesting doesn't appear to have a created_date, is there another date that should be used or should created_date be added?

tsteur:
@diosmosis as mentioned in chat `start_date` will need to be used. That's when the first tracking request for an a/b test has been received.

diosmosis:
@tsteur prs have all been created, awaiting review

tsteur:
@sgiehl can you review the linked PRs in the issue description? Basically, when using CLI archiving, then it should rearchive the related plugin specific reports or archive it for reports in the past on plugin activation for example.

sgiehl:
done. Left some comments on some PRs

tsteur:
I suppose we now only need to update the other plugins that were using the activate hook and then release a new version for each plugin (I can do that) and things should be done?

tsteur:
@diosmosis @sgiehl

diosmosis:
@tsteur it's on my list for today

diosmosis:
[...]

PR #16683, opened 2020-11-09
Keep track of last activate/deactivate date for plugins

diosmosis:
### Description:

Refs #16396 

Stores the last time of activation/deactivation of plugins for use in other plugins. Also includes a change capping the amount of dates that are rearchived if a custom date is supplied.

### Review

* [x] Functional review done
* [x] Security review done
* [x] Code review done


core/Archive/ArchiveInvalidator.php
```diff
@@ -460,18 +460,12 @@ public function reArchiveReport($idSites, string $plugin, string $report = null,
...
-            $lastNMonthsToInvalidate = (int) substr($lastNMonthsToInvalidate, 4);
-            if (empty($lastNMonthsToInvalidate)) {
-                return;
-            }
-
-            $startDate = $date2->subMonth($lastNMonthsToInvalidate)->setDay(1);
+            $startDate = $earliestDateToRearchive;
+        } else if (!empty($earliestDateToRearchive)) {
+            // don't allow archiving further back than the rearchive_reports_in_past_last_n_months date allows
+            $startDate = $startDate->isEarlier($earliestDateToRearchive) ? $earliestDateToRearchive : $startDate;
```
sgiehl, line 468: Does that mean there won't be any possibility to rearchive specific reports for dates older than `rearchive_reports_in_past_last_n_months`? Wondering if there might be any use cases where it might make sense to force that somehow. But guess would be enough to implement it if we have such a use case at some point

core/Plugin/Manager.php
```diff
@@ -516,6 +521,7 @@ public function deactivatePlugin($pluginName)
 
         // execute deactivate() to let the plugin do cleanups
         $this->executePluginDeactivate($pluginName);
+        $this->savePluginDeactivationTime($pluginName);
```
tsteur, line 524: @diosmosis bit unrelated probably but could we execute the deactivate logic only if the plugin is actually activated so it can't be triggered with just any random plugin name?

core/Plugin/Manager.php
```diff
@@ -1654,5 +1662,47 @@ public function loadPluginTranslations()
             $translator->addDirectory(self::getPluginDirectory($pluginName) . '/lang');
         }
     }
+
+    /**
+     * @param string $pluginName
+     * @return Date|null
+     * @throws \Exception
+     */
+    public function getPluginLastActivationTime($pluginName)
```
tsteur, line 1671: not any important but could move this methods to the `Plugin` class so things would maybe feel a bit more natural like `$plugin->getLastActivationTime()`. Not sure that works though in our context maybe we need to have this information when the plugin is not loaded yet or unloaded already or so. Feel free to ignore

core/Plugin/Manager.php
tsteur: eg in `activatePlugin` we do:

```
        $plugins = $this->pluginList->getActivatedPlugins();
        if (in_array($pluginName, $plugins)) {
            // plugin is already activated
            return;
        }
```

and could use similar logic

core/Archive/ArchiveInvalidator.php
tsteur: I guess a user could always configure it differently above setting.

diosmosis: For this a plugin could manually invalidate, since rearchiveRepors is really just a smart wrapper for `markArchivesAsInvalidated`.

diosmosis:
@tsteur applied changes

core/Plugin.php
```diff
@@ -513,6 +513,36 @@ public function isTrackerPlugin()
         return false;
     }
 
+    /**
+     * @param string $pluginName
+     * @return Date|null
+     * @throws \Exception
+     */
+    public function getPluginLastActivationTime($pluginName)
```
tsteur, line 521: @diosmosis could rename it to `getLastActivationTime()` and remove the `$pluginName` and instead use `$this->pluginName`?

tsteur:
👍  looks good to me. Left one comment though and didn't really test it. @sgiehl maybe you could give it a test?

core/Plugin.php
diosmosis: 👍

core/Plugin/Manager.php
```diff
@@ -1654,5 +1668,17 @@ public function loadPluginTranslations()
...
+    {
+        $optionName = self::LAST_PLUGIN_ACTIVATION_TIME_OPTION_PREFIX . $pluginName;
+        Option::set($optionName, time());
+    }
+
+    private function savePluginDeactivationTime($pluginName)
+    {
+        $optionName = self::LAST_PLUGIN_DEACTIVATION_TIME_OPTION_PREFIX . $pluginName;
+        Option::set($optionName, time());
+    }
```
sgiehl, line 1682: Should we clean up those entries whenever a plugin gets completely uninstalled?

sgiehl (approved):
Storing those values works as expected.

core/Plugin/Manager.php
diosmosis: I think it might still be useful to know when the plugin was last there... might be more useful to know when it was installed/uninstalled. What do you think @tsteur?

tsteur: No preference here. Might be useful though to keep them and I can't think of some downside when keeping these entries.

core/Archive/ArchiveInvalidator.php
tsteur: 👍

PR #16685, opened 2020-11-09
Fix overflow handling of datatables

sgiehl:
### Description:

With introducing the datatable scroller in #14594, there actually shouldn't be any further need of checking if the table has enough space, as it always fits. Guess adding a overflow-y was needed to ensure the overlays were at least visible with scrolling. 
The current handling actually always displays a vertical scroll bar as soon as the table would be too wide (without the datatable scroller), causing all overlays to be displayed in the report instead of overlaying it:

Chrome:
image

Firefox & IE:
image

### Review

* [ ] Functional review done
* [ ] Security review done
* [ ] Code review done


tsteur:
@sgiehl I'm not sure what this fixes? Is there an issue for it?

sgiehl:
There is no issue yet afaik. Came across that issue when testing the page performance stuff. You can easily reproduce it with any dashboard widget that would be wider than the available space. You can use the page url widget for example in a small dashboard column. It will always show the vertical scrollbar, even though that isn't actually needed. When you click one of the footer icons, the overlays aren't showing correctly because of that.

diosmosis:
Can't reproduce the issue, but I didn't notice any problem w/ the change so I guess it's good to at least remove unneeded code. It's possible I didn't test the exact place where this happens though...

sgiehl:
@diosmosis Steps to reproduce:
Go to the dashboard. Add the pages widget. The widget should always show an vertical scrollbar. Click the limit selection in the footer, this should kind of break the layout (which should not happen with this changes)

diosmosis:
@sgiehl was able to reproduce, he fix works 👍

diosmosis (approved):

PR #16686, opened 2020-11-09
Adds nonce to one click update form

sgiehl:


tsteur:
@sgiehl this also needs to be added to `updateHttpsError.twig`?

Issue #16689, opened 2020-11-10
'Visits Overview' not updating while `Visits in Real-time` works fine (Matomo 4)


PepijnBoers:
Hi,

I'm having some issues with my matomo server since updating to the 4.0.0-rc1 beta. I'm using a crontab to automatically archive my matomo reports. However, the `Visits Overview` statistics are not updated anymore. The `Visits in Real-time` does work as intended and shows recent visits. All the system checks succeed and I'm not sure what to look at. I tried deleting the `tmp` directory but it didn't resolve anything.

Manually executing the archive command does not show new API requests. Are there some steps I can take to resolve this issue?

```
root@c3e43c5fd0ff:/var/www/html# ./console core:archive
INFO [2020-11-10 18:42:53] 335  ---------------------------
INFO [2020-11-10 18:42:53] 335  INIT
INFO [2020-11-10 18:42:53] 335  Running Matomo 4.0.0-rc1 as Super User
INFO [2020-11-10 18:42:53] 335  ---------------------------
INFO [2020-11-10 18:42:53] 335  NOTES
INFO [2020-11-10 18:42:53] 335  - Async process archiving supported, using CliMulti.
INFO [2020-11-10 18:42:53] 335  - Reports for today will be processed at most every 10 seconds. You can change this value in Matomo UI > Settings > General Settings.
INFO [2020-11-10 18:42:53] 335  - Archiving was last executed without error 8 min 27s ago
INFO [2020-11-10 18:42:53] 335  ---------------------------
INFO [2020-11-10 18:42:53] 335  START
INFO [2020-11-10 18:42:53] 335  Starting Matomo reports archiving...
INFO [2020-11-10 18:42:53] 335  Start processing archives for site 1.
INFO [2020-11-10 18:42:53] 335  Finished archiving for site 1, 0 API requests, Time elapsed: 0.002s [1 / 2 done]
INFO [2020-11-10 18:42:53] 335  Start processing archives for site 4.
INFO [2020-11-10 18:42:53] 335  Finished archiving for site 4, 0 API requests, Time elapsed: 0.002s [2 / 2 done]
INFO [2020-11-10 18:42:53] 335  Done archiving!
INFO [2020-11-10 18:42:53] 335  ---------------------------
INFO [2020-11-10 18:42:53] 335  SUMMARY
INFO [2020-11-10 18:42:53] 335  Processed 0 archives.
INFO [2020-11-10 18:42:53] 335  Total API requests: 0
[...]

PR #16701, opened 2020-11-11
Better detect failed archiving jobs.

diosmosis:
### Description:

Refs #16689

Better detection of in progress invalidations: if an invalidation's status is set to in progress, but the invalidation was created over a day ago. (We also update the ts_invalidated time to now() when starting an archive).

CC @tsteur 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/DataAccess/Model.php
+++ b/core/DataAccess/Model.php
@@ -632,7 +632,7 @@ class Model
         $table = Common::prefixTable('archive_invalidations');
 
         // set archive value to in progress if not set already
-        $statement = Db::query("UPDATE `$table` SET `status` = ? AND ts_invalidated = NOW() WHERE idinvalidation = ? AND status = ?", [
+        $statement = Db::query("UPDATE `$table` SET `status` = ? AND ts_started = NOW() WHERE idinvalidation = ? AND status = ?", [
             ArchiveInvalidator::INVALIDATION_STATUS_IN_PROGRESS,
             $invalidation['idinvalidation'],
             ArchiveInvalidator::INVALIDATION_STATUS_QUEUED,
@@ -871,7 +871,7 @@ class Model
     public function resetFailedArchivingJobs()
     {
         $table = Common::prefixTable('archive_invalidations');
-        $sql = "UPDATE $table SET status = ? WHERE status = ? AND ts_invalidated < ?";
+        $sql = "UPDATE $table SET status = ? WHERE status = ? AND ts_started IS NOT NULL AND ts_started < ?";
 
         $bind = [
             ArchiveInvalidator::INVALIDATION_STATUS_QUEUED,
--- a/core/DataAccess/Model.php
+++ b/core/DataAccess/Model.php
@@ -632,7 +632,7 @@ class Model
         $table = Common::prefixTable('archive_invalidations');
 
         // set archive value to in progress if not set already
-        $statement = Db::query("UPDATE `$table` SET `status` = ? WHERE idinvalidation = ? AND status = ?", [
+        $statement = Db::query("UPDATE `$table` SET `status` = ? AND ts_invalidated = NOW() WHERE idinvalidation = ? AND status = ?", [
             ArchiveInvalidator::INVALIDATION_STATUS_IN_PROGRESS,
             $invalidation['idinvalidation'],
             ArchiveInvalidator::INVALIDATION_STATUS_QUEUED,
@@ -866,5 +866,20 @@ class Model
         $table = Common::prefixTable('archive_invalidations');
         $sql = "UPDATE $table SET status = " . ArchiveInvalidator::INVALIDATION_STATUS_QUEUED . " WHERE idinvalidation = ?";
         Db::query($sql, [$idinvalidation]);
+    }
+
+    public function resetFailedArchivingJobs()
+    {
+        $table = Common::prefixTable('archive_invalidations');
+        $sql = "UPDATE $table SET status = ? WHERE status = ? AND ts_invalidated < ?";
+
+        $bind = [
+            ArchiveInvalidator::INVALIDATION_STATUS_QUEUED,
+            ArchiveInvalidator::INVALIDATION_STATUS_IN_PROGRESS,
+            Date::now()->subDay(1)->getDatetime(),
+        ];
+
+        $query = Db::query($sql, $bind);
+        return $query->rowCount();
     }
 }
--- a/core/CronArchive.php
+++ b/core/CronArchive.php
@@ -345,6 +345,13 @@ class CronArchive
             return;
         }
 
+        $failedJobs = $this->model->resetFailedArchivingJobs();
+        if ($failedJobs) {
+            $this->logger->info("Found {failed} failed jobs (ts_invalidated older than 1 day), resetings status to try them again.", [
+                'failed' => $failedJobs,
+            ]);
+        }
+
         $countOfProcesses = $this->getMaxConcurrentApiRequests();
 
         $queueConsumer = new QueueConsumer($this->logger, $this->websiteIdArchiveList, $countOfProcesses, $pid,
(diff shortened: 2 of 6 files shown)


diosmosis:
Note: working on a test

core/DataAccess/Model.php
```diff
@@ -786,10 +788,35 @@ public function isArchiveAlreadyInProgress($invalidatedArchive)
             $bind[] = $invalidatedArchive['report'];
         }
 
-        $sql = "SELECT MAX(idinvalidation) FROM `$table` WHERE idsite = ? AND date1 = ? AND date2 = ? AND `period` = ? AND `name` = ? AND status = 1 AND $reportClause";
+        $sql = "SELECT MAX(idinvalidation) as idinvalidation, ts_invalidated FROM `$table` WHERE idsite = ? AND date1 = ? AND date2 = ? AND `period` = ? AND `name` = ? AND status = 1 AND $reportClause";
```
tsteur, line 791: not sure it's possible or make sense but do we also need a max or min on the `ts_invalidated`?

core/DataAccess/Model.php
```diff
@@ -786,10 +788,35 @@ public function isArchiveAlreadyInProgress($invalidatedArchive)
...
+            return true;
+        }
+
+        StaticContainer::get(LoggerInterface::class)->info("Found invalidation over a day old (ts_invalidated = {inv}, idinvalidation = {idinv}), assuming it failed and deleting.", [
+            'inv' => $inProgressInvalidation['ts_invalidated'],
+            'idinv' => $inProgressInvalidation['idinvalidation'],
+        ]);
+
+        // delete the in progress invalidation since we assume it errored
+        $this->deleteInvalidations($inProgressInvalidation);
```
tsteur, line 817: random thought... might not make any sense. If we assume it failed, would we rather set it to `0` so it will be rearchived again in the hope it works and doesn't fail the next time? Although I suppose this can get complicated as we'd probably then also would maybe need to invalidate any parent period in case they were finished in the meantime?

core/DataAccess/Model.php
diosmosis: This was a mistake I fixed locally, should be a LIMIT 1 query.

core/DataAccess/Model.php
diosmosis: Parent periods shouldn't start if an intersecting lower period is there so this *should* be ok. The period might just keep failing though.

core/DataAccess/Model.php
tsteur: 👍

core/DataAccess/Model.php
tsteur: Not sure what to do best here. I'm tempted to say we set the status `0` and better maybe fail repeatedly than having potentially wrong data? Problem obviously being it wouldn't archive any other data until this is resolved although I think if newer invalidations are added, and they are for newer periods, then these ones would be maybe processed eventually because we prefer more recent dates AFAIK. Might not be thinking this fully through though.

diosmosis: @tsteur Actually, this is in the check for an existing in progress archive, so that would mean there already is another invalidation identical to this one.

I guess we can also check in CoreArchive.php for in progress invalidations that are over a day old and change the status on them before we actually start any archiving. What do you think about that?

tsteur: I see. I guess here in `isArchiveAlreadyInProgress ` if the same invalidation wants to start, and there is one in progress for 1 day already then we can delete it.

There would be still some that would never be detected. It could be in CronArchive.php or could be a task to check for these? Could in this case maybe only use this logic and undo the change in `isArchiveAlreadyInProgress ` as having a general check should help? Ideally we just don't execute the query looking for in progress archive invalidations older than 1 day too often. Maybe only once hourly or once on cronarchive start or so.

diosmosis: > There would be still some that would never be detected. It could be in CronArchive.php or could be a task to check for these? Could in this case maybe only use this logic and undo the change in isArchiveAlreadyInProgress as having a general check should help? Ideally we just don't execute the query looking for in progress archive invalidations older than 1 day too often. Maybe only once hourly or once on cronarchive start or so.

Yes, this was what I mentioned in the second paragraph. I'll go w/ this.

diosmosis:
@tsteur updated

core/DataAccess/Model.php
```diff
@@ -867,4 +867,19 @@ public function releaseInProgressInvalidation($idinvalidation)
         $sql = "UPDATE $table SET status = " . ArchiveInvalidator::INVALIDATION_STATUS_QUEUED . " WHERE idinvalidation = ?";
         Db::query($sql, [$idinvalidation]);
     }
+
+    public function resetFailedArchivingJobs()
+    {
+        $table = Common::prefixTable('archive_invalidations');
+        $sql = "UPDATE $table SET status = ? WHERE status = ? AND ts_invalidated < ?";
```
tsteur, line 874: @diosmosis could we add a test for this? and might be also good to document this in https://developer.matomo.org/guides/archiving ?

core/DataAccess/Model.php
```diff
@@ -632,7 +632,7 @@ public function startArchive($invalidation)
         $table = Common::prefixTable('archive_invalidations');
 
         // set archive value to in progress if not set already
-        $statement = Db::query("UPDATE `$table` SET `status` = ? WHERE idinvalidation = ? AND status = ?", [
+        $statement = Db::query("UPDATE `$table` SET `status` = ? AND ts_invalidated = NOW() WHERE idinvalidation = ? AND status = ?", [
```
tsteur, line 635: would almost rather need a different field for this like `ts_start`? although maybe not too important. Just thought it might make some troubleshooting bit more difficult but I think it's actually fine as it would only affect invalidations that are in progress

tsteur (commented):
looks good to me. The only thing I'm scared of is that an archive run that takes longer than a day would be basically always reset to `0` and always run again but suppose one day shouldn't happen. I wouldn't be surprised if it happens though eg when a monthly archive is triggered, and then it notices some days are outdated, and it also rearchives several days and then weeks etc. Even though we triggered month, I think it might still archive individual days when needed not sure...

Should we change it to maybe 30 hours instead? Still not perfect but I think that would make things bit safer and less likely to run into that issue.

Actually... after more thinking this might not be needed to set it to 30 hours. Because: If it was to try to start that archive again, and it is still in progress, and archiving through CLI is used then it would notice this one is already running. And next time the archive is triggered it should finish in less than 24 hours considering the individual days and weeks within that month would be eventually archived.

diosmosis:
> looks good to me. The only thing I'm scared of is that an archive run that takes longer than a day would be basically always reset to 0 and always run again but suppose one day shouldn't happen. I wouldn't be surprised if it happens though eg when a monthly archive is triggered, and then it notices some days are outdated, and it also rearchives several days and then weeks etc. Even though we triggered month, I think it might still archive individual days when needed not sure...

this shouldn't technically happen since we order the periods by day => month and don't allow simultaneously archiving intersecting periods. there might be some edge cases where it does however, like someone deletes an archive right as a month archive starts, but it would be rare I think. best way to avoid this would be to check the Lock before resetting an archive, not sure if that would be worth doing.

core/DataAccess/Model.php
diosmosis: this was a quick change, I can add an extra column if needed

tsteur: I have no big preference. Leaving this up to you. Not sure how much work it is. I reckon it may be useful for debugging purposes in the future to understand things better but may not be needed

tsteur:
> this shouldn't technically happen since we order the periods by day => month and don't allow simultaneously archiving intersecting periods

I'm thinking this might still happen. When requesting a month period then it has to archive within the same request a missing day archive etc.

diosmosis:
@tsteur updated

PR #16702, opened 2020-11-11
Delayed rearchiving

diosmosis:
### Description:

Refs #16396

Schedule rearchiving calls in a list and apply later in core:archive. CC @tsteur 

### Review

* [ ] Functional review done
* [ ] Code review done


core/Archive/ArchiveInvalidator.php
```diff
@@ -528,20 +535,48 @@ public function removeInvalidations($idSite, $plugin, $report = null)
...
+            } catch (\Throwable $ex) {
+                $this->logger->info("Failed to create invalidations for report re-archiving (idSites = {idSites}, pluginName = {pluginName}, report = {report}, startDate = {startDateTs}): {ex}", [
+                    'idSites' => json_encode($idSites),
+                    'pluginName' => $pluginName,
+                    'report' => $report,
+                    'startDateTs' => $startDateTs,
+                    'ex' => $ex,
+                ]);
+            } finally {
+                $reArchiveList->remove([$entry]);
```
sgiehl, line 578: Does it make sense to remove the record even if it failed to create the invalidation? Shouldn't we keep it and try to add it another time?

sgiehl (changes requested):
Guess the advantage of this changes is that it won't take longer to perform any action that created dozens of invalidations before, as those records are now created while archiving, right?

core/Archive/ArchiveInvalidator.php
```diff
@@ -528,20 +535,48 @@ public function removeInvalidations($idSite, $plugin, $report = null)
...
+    public function applyScheduledReArchiving()
+    {
+        $reArchiveList = new ReArchiveList($this->logger);
+        $items = $reArchiveList->getAll();
+
+        foreach ($items as $entry) {
+            try {
+                list($idSites, $pluginName, $report, $startDateTs) = $entry;
+
+                $this->reArchiveReport($idSites, $pluginName, $report, Date::factory($startDateTs));
```
tsteur, line 568: maybe we should only create date instance if startDateTs is not empty cause otherwise it might result in an error or creates a 'now' date or something else unexpected?

core/Archive/ArchiveInvalidator.php
```diff
@@ -528,20 +535,48 @@ public function removeInvalidations($idSite, $plugin, $report = null)
...
      * @param string|null $report
      * @param Date|null $startDate
      */
-    public function reArchiveReportSafely($idSites, string $pluginName, string $report = null, Date $startDate = null)
+    public function scheduleReArchiving($idSites, string $pluginName, string $report = null, Date $startDate = null)
     {
         try {
-            $this->reArchiveReport($idSites, $pluginName, $report, $startDate);
+            $reArchiveList = new ReArchiveList($this->logger);
+            $reArchiveList->add([$idSites, $pluginName, $report, $startDate ? $startDate->getTimestamp() : null]);
```
tsteur, line 550: Personally would prefer if we used an array like `['idsites' => $idSites, ...]` this would allow us to quite easily extend or change the structure in the future if/when needed. and makes it maybe also more clear what each array item represents? Not too important though. It be mostly an issue if we eg ever removed for example an argument and then would need to do things like `[$idSites, $pluginName, null, null, $newArgument]` or something

core/CronArchive/ReArchiveList.php
```diff
@@ -0,0 +1,23 @@
...
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ *
+ */
+
+namespace Piwik\CronArchive;
+
+use Piwik\Concurrency\DistributedList;
+use Psr\Log\LoggerInterface;
+
+class ReArchiveList extends DistributedList
```
tsteur, line 15: it's great this was so easy to do and that we have this distributedlist already!

core/Archive/ArchiveInvalidator.php
tsteur: No preference from my side here as I reckon usually it shouldn't fail? if there may be some reasons why it might happen could also keep the entry and retry later

core/CronArchive.php
```diff
@@ -751,6 +751,9 @@ public function invalidateArchivedReportsForSitesThatNeedToBeArchivedAgain($idSi
             return;
         }
 
+        $this->logger->debug("Applying queued rearchiving...");
+        $this->invalidator->applyScheduledReArchiving();
```
tsteur, line 755: @diosmosis could we maybe call this somewhere else so it's called less often? Eg only at the start of the cron archive which be typically every hour?

tsteur (commented):
@diosmosis ideally we also adjust `removeInvalidationsSafely` to additionally remove any entry from the list that matches the given `pluginName` (and maybe idsites)? just to prevent a race condition when quickly activating and deactivating would cause the invalidations to be added after the plugin was deactivated. although maybe these invalidations would be ignored anyway when the plugin is no longer active?

core/CronArchive.php
diosmosis: That should only be called once per core:archive.

core/Archive/ArchiveInvalidator.php
diosmosis: 👍

core/Archive/ArchiveInvalidator.php
diosmosis: 👍

core/Archive/ArchiveInvalidator.php
diosmosis: It should only fail if the input is broken for some reason, this is mostly to replicate the old behavior where if it failed for some reason, invalidations wouldn't be added.

diosmosis:
> ideally we also adjust removeInvalidationsSafely to additionally remove any entry from the list that matches the given pluginName (and maybe idsites)?

👍

core/CronArchive.php
tsteur: not sure I'm seeing it right. Isn't it called as part of `getNextArchivesToProcess()` and therefore as part of the while loop?

diosmosis: Oh you're right, nevermind, I had it at the start of run() initially but moved it for some reason. I'll move it bacck.

diosmosis:
@tsteur Applied pr fixes.

tsteur:
I'll be mostly in meetings tmrw. @sgiehl could you test this tomorrow? and then we adjust the plugins that use this in activate/deactivate plugin

core/Archive/ArchiveInvalidator.php
```diff
@@ -590,12 +603,47 @@ public function removeInvalidationsSafely($idSites, $pluginName)
...
 
+    public function removeInvalidationsFromDistributedList($idSites, $pluginName = null)
+    {
+        $list = new ReArchiveList();
+        $entries = $list->getAll();
+
+        foreach ($entries as $index => $entry) {
+            $entry = @json_decode($entry, true);
+            if (empty($entry)) {
+                continue;
```
sgiehl, line 621: wouldn't it make sense to always remove entries that can't be decoded? Otherwise they might kept in the list forever

core/Archive/ArchiveInvalidator.php
sgiehl: Now that the entries are json encoded, that needs to be done for the removal as well. Otherwise it won't be removed at all. Alternatively you could also use another variable for the decoded entry...

```suggestion
                $reArchiveList->remove([json_encode($entry)]);
```

core/Archive/ArchiveInvalidator.php
```diff
@@ -528,20 +535,61 @@ public function removeInvalidations($idSite, $plugin, $report = null)
...
+                $entry = @json_decode($entry, true);
+                if (empty($entry)) {
+                    continue;
+                }
+
+                $this->reArchiveReport(
+                    $entry['idSites'],
+                    $entry['pluginName'],
+                    $entry['report'],
+                    !empty($entry['startDate']) ? Date::factory($entry['startDate']) : null
```
sgiehl, line 580: should we maybe cast `$entry['startDate']` to integer before passing it to the factory. It should normally already be an integer, but if it would be a string (containing a number), the factory will throw an exception

sgiehl:
tested that locally with adjusted Cohorts PR. Besides the comments I've added and #16705 it seems to work as expected.

tsteur:
👍  @diosmosis so seems good to merge once tests pass?

core/Archive/ArchiveInvalidator.php
diosmosis: 👍

core/Archive/ArchiveInvalidator.php
diosmosis: 👍

sgiehl:
Guess you missed that comment: https://github.com/matomo-org/matomo/pull/16702#discussion_r522023501
Will create a PR to fix that

diosmosis:
> Guess you missed that comment: #16702 (comment)

yes, missed it, the tab that had this github page didn't update I guess

PR #16704, opened 2020-11-11
Append a suffix to temporary downloaded geoip files

sgiehl:
### Review

* [x] Functional review done
* [x] Code review done


diosmosis:
Seems like it might be a bit of work, but is it possible to build this into the Http class so anyone that downloads to a file will have the suffix?

sgiehl:
@diosmosis does it make sense to do that always? It would actually be a breaking change as downloaded files would suddenly have another name as expected - unless we make that optional, but then it would be kind of useless as the default would need to be to not append anything to have bc.

diosmosis:
@sgiehl good point, can't think of a way for it to just automatically work for existing uses.

diosmosis (approved):
tested locally + code looks good

PR #16711, opened 2020-11-13
Fix removal of rearchiving records

sgiehl:
### Description:

follow up to #16702

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -566,9 +566,9 @@ class ArchiveInvalidator
         $reArchiveList = new ReArchiveList($this->logger);
         $items = $reArchiveList->getAll();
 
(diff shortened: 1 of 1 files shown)


diosmosis:
👍 nice catch

core/Archive/ArchiveInvalidator.php
```diff
@@ -566,9 +566,9 @@ public function applyScheduledReArchiving()
         $reArchiveList = new ReArchiveList($this->logger);
         $items = $reArchiveList->getAll();
 
-        foreach ($items as $entry) {
+        foreach ($items as $item) {
```
tsteur, line 569: btw for such tricky fixes might be good to add a test case if possible as it otherwise might regress again easily

PR #16713, opened 2020-11-13
DB-IP URL is automatically updated in the options

### LINKED ISSUE #16618: Mention DB Lite DB will be automatically updated (filename etc)
Hi, I know MaxMind put their DB behind API / login, and now Matomo uses db-ip download. For at least a year I have noticed that cities are very wrong, they are not at all accurate as before (countries are still accurate, as I can tell so far).

Today I devoted some time to exploring this issue, and discovered that update db link is hard coded with a date from January 2020. But, Matomo correctly reports that latest db version is from October 2020.

Does this affect downloading latest version or not (see screenshot)? Do we have to manually update this link from time to time?

matomo-geoip-db

nina-py:
### Description:

Added a method called by the updater task that updates the DB-IP URL if needed to the latest downloaded one instead of the one set in the options when Matomo was initially installed. This removes much of the confusion at seeing a seemingly outdated URL being used on the GeoIP2 plugin page.

Would love to contribute a test as well but need to make sure I add the right test in the right place - I think a system test would be the right one for this method as I'd need to make sure an option was changed?

Tested the changes by manually setting the `geoip2.loc_db_url` option to `https://download.db-ip.com/free/dbip-city-lite-2020-05.mmdb.gz` in the database, then running the updater task with

```bash
./console core:run-scheduled-tasks "Piwik\Plugins\GeoIp2\GeoIP2AutoUpdater.update"
```
Result - option set to `https://download.db-ip.com/free/dbip-city-lite-2020-11.mmdb.gz` and the plugin page now displays the latest URL in the "Download URL" field.

Fixes #16618.

### Review

* [x] Functional review done
* [x] Code review done

### DIFF
--- a/plugins/GeoIp2/GeoIP2AutoUpdater.php
+++ b/plugins/GeoIp2/GeoIP2AutoUpdater.php
@@ -102,12 +102,13 @@ class GeoIP2AutoUpdater extends Task
             $locUrl = Option::get(self::LOC_URL_OPTION_NAME);
             if (!empty($locUrl)) {
                 $this->downloadFile('loc', $locUrl);
-                $this->updateDbIpUrlOption($locUrl);
+                $this->updateDbIpUrlOption(self::LOC_URL_OPTION_NAME, $locUrl);
             }
 
             $ispUrl = Option::get(self::ISP_URL_OPTION_NAME);
             if (!empty($ispUrl)) {
                 $this->downloadFile('isp', $ispUrl);
+                $this->updateDbIpUrlOption(self::ISP_URL_OPTION_NAME, $ispUrl);
             }
         } catch (Exception $ex) {
             // message will already be prefixed w/ 'GeoIP2AutoUpdater: '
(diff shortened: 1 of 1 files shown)


plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -798,4 +799,26 @@ protected function fetchUrl($url)
...
+
+    /**
+     * Updates the DB-IP URL option value so that users see
+     * the updated link in the "Download URL" field on the plugin page
+     * instead of the one that was set when Matomo was installed months
+     * or even years ago.
+     *
+     * @param  string  $url The URL retrieved from the options
+     */
+    protected function updateDbIpUrlOption(string $url): void
```
sgiehl, line 811: Could you maybe change the method, so it's also able to update the ISP database url as well. Someone could for example use the free asn database from DB-IP (https://download.db-ip.com/free/dbip-asn-lite-2020-11.mmdb.gz)

nina-py: Of course - I wasn't sure if the ISP DB was also sources from DB-IP so I limited this method to updating the location URL.

nina-py: Done - I set the ISP DB link to `https://download.db-ip.com/free/dbip-asn-lite-2020-04.mmdb.gz` in my options and backdated the location DB URL as well, and re-run the updater task - it updated both URLs to the November ones.

plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -798,4 +800,31 @@ protected function fetchUrl($url)
...
+            self::isDbIpUrl($url)
+            && ($option === self::LOC_URL_OPTION_NAME || $option === self::ISP_URL_OPTION_NAME)
+        ) {
+            $latestUrl = $this->getDbIpUrlWithLatestDate($url);
+
+            if($url !== $latestUrl) {
+                Option::set($option, $latestUrl);
+            }
+        }
+    }
```
sgiehl, line 829: Guess we could remove the second parameter and read the option directly in the method. Doesn't hurt to do that again, as the value will be cached once it was read. So maybe something like:

```php
    protected function updateDbIpUrlOption(string $option): void
    {
        if ($option !== self::LOC_URL_OPTION_NAME && $option !== self::ISP_URL_OPTION_NAME) {
            return;
        }
    
        $url = trim(Option::get($option));

        if (self::isDbIpUrl($url)) {
            $latestUrl = $this->getDbIpUrlWithLatestDate($url);

            if($url !== $latestUrl) {
                Option::set($option, $latestUrl);
            }
        }
    }
```

nina-py: Thanks @sgiehl, makes sense to get rid of the second parameter if option values are cached.

sgiehl (approved):
Tested the changes locally. Works as expected 👍

sgiehl:
@tsteur Guess should be fine to merge that for 4.0 already. Feel free to do that or to move it to another milestone.

tsteur:
Thanks @nina-py very appreciated

PR #16716, opened 2020-11-15
Fix ecommerce and goal tracking may not work

tsteur:
### Description:

Seems they rely on page view action in order to work and should therefore not send the ca parameter

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
build js

PR #16717, opened 2020-11-15
typo in deleting blob archive data

diosmosis:
### Description:

archive tables are determined by start date, but we're using end date in this one place (probably due to autocomplete error)

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Updates/4.0.0-rc4.php
+++ b/core/Updates/4.0.0-rc4.php
@@ -10,8 +10,10 @@
 namespace Piwik\Updates;
 
 use Piwik\Config;
+use Piwik\Container\StaticContainer;
 use Piwik\DataAccess\ArchiveTableCreator;
 use Piwik\Date;
+use Piwik\Plugin\ReleaseChannels;
 use Piwik\Updater;
 use Piwik\Updates as PiwikUpdates;
 use Piwik\Updater\Migration\Factory as MigrationFactory;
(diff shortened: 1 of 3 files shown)


tsteur:
LGTM @diosmosis I suppose you add the DB upgrade statement here?

diosmosis:
@tsteur added the new update

core/Updates/4.0.0-rc4.php
```diff
@@ -0,0 +1,56 @@
...
+    public function __construct(MigrationFactory $factory)
+    {
+        $this->migration = $factory;
+    }
+
+    public function getMigrations(Updater $updater)
+    {
+        $migrations = [];
+
+        $dates = ['2020-01-01', '2020-10-01', '2020-09-01'];
```
tsteur, line 38: @diosmosis does it need november in there too? I reckon we could actually remove September but no preference. We only released the beta on like 29th of September so it wouldn't have been much of an issue back then

diosmosis: oh yes, it should, will fix

diosmosis: @tsteur updated

core/Updates/4.0.0-rc4.php
```diff
@@ -0,0 +1,56 @@
...
+        foreach ($dates as $date) {
+            $date = Date::factory($date);
+            $numericTable = ArchiveTableCreator::getBlobTable($date);
+            $blobTable = ArchiveTableCreator::getNumericTable($date);
+
+            $migrations[] = $this->migration->db->sql("DELETE FROM `$blobTable` WHERE idarchive NOT IN (
+                SELECT idarchive FROM `$numericTable`)");
```
tsteur, line 45: probably won't make a difference but might be better to use `SELECT DISTINCT` in the subquery? I reckon this might speed up the deletion since it will need to check a lot less entries for each row whether id is included

PR #16724, opened 2020-11-16
SameSite cookie attribute can now be configured in JS tracker

fealXX:
### Description:

Re-implements the Changes made by @beger in [PR#16504](https://github.com/matomo-org/matomo/pull/16504), added the discussed functionality

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/js/piwik.js
+++ b/js/piwik.js
@@ -6890,7 +6890,7 @@ if (typeof window.Matomo !== 'object') {
          * Constructor
          ************************************************************/
 
-        var applyFirst = ['addTracker', 'forgetCookieConsentGiven', 'requireCookieConsent', 'disableCookies', 'setTrackerUrl', 'setAPIUrl', 'enableCrossDomainLinking', 'setCrossDomainLinkingTimeout', 'setSessionCookieTimeout', 'setVisitorCookieTimeout', 'setCookieNamePrefix', 'setSecureCookie', 'setCookiePath', 'setCookieDomain', 'setDomains', 'setUserId', 'setVisitorId', 'setSiteId', 'alwaysUseSendBeacon', 'enableLinkTracking', 'setCookieConsentGiven', 'requireConsent', 'setConsentGiven'];
+        var applyFirst = ['addTracker', 'forgetCookieConsentGiven', 'requireCookieConsent', 'disableCookies', 'setTrackerUrl', 'setAPIUrl', 'enableCrossDomainLinking', 'setCrossDomainLinkingTimeout', 'setSessionCookieTimeout', 'setVisitorCookieTimeout', 'setCookieNamePrefix', 'setCookieSameSite', 'setSecureCookie', 'setCookiePath', 'setCookieDomain', 'setDomains', 'setUserId', 'setVisitorId', 'setSiteId', 'alwaysUseSendBeacon', 'enableLinkTracking', 'setCookieConsentGiven', 'requireConsent', 'setConsentGiven'];
 
         function createFirstTracker(matomoUrl, siteId)
         {
--- a/js/piwik.js
+++ b/js/piwik.js
@@ -3337,7 +3337,7 @@ if (typeof window.Matomo !== 'object') {
                 var cookieValue = visitorIdCookieValues.uuid + '.' +
                     visitorIdCookieValues.createTs + '.';
 
-                setCookie(getCookieName('id'), cookieValue, getRemainingVisitorCookieTimeout(), configCookiePath, configCookieDomain, configCookieIsSecure);
+                setCookie(getCookieName('id'), cookieValue, getRemainingVisitorCookieTimeout(), configCookiePath, configCookieDomain, configCookieIsSecure, configCookieSameSite);
             }
 
             /*
--- a/js/piwik.js
+++ b/js/piwik.js
@@ -5759,11 +5759,15 @@ if (typeof window.Matomo !== 'object') {
              * This should be used when your website is only available under HTTPS
              * so that all tracking cookies are always sent over secure connection.
              *
+             * Warning: If your site is available under http and https,
+             * setting this might lead to duplicate or incomplete visits.
+             *
              * @param bool
              */
             this.setSecureCookie = function (enable) {
                 if(enable && location.protocol !== 'https:') {
                     logConsoleError("Error in setSecureCookie: You cannot use `Secure` on http.");
+                    return;
                 }
                 configCookieIsSecure = enable;
             };
@@ -5772,7 +5776,12 @@ if (typeof window.Matomo !== 'object') {
              * Set the SameSite attribute for cookies to a custom value.
              * You might want to use this if your site is running in an iframe since
              * then it will only be able to access the cookies if SameSite is set to 'None'.
+             *
+             *
+             * Warning: 
              * Sets CookieIsSecure to true on None, because None will only work with Secure; cookies
+             * If your site is available under http and https,
+             * using "None" might lead to duplicate or incomplete visits.
              *
              * @param string either Lax, None or Strict
              */
--- a/js/piwik.js
+++ b/js/piwik.js
@@ -5762,7 +5762,7 @@ if (typeof window.Matomo !== 'object') {
              * @param bool
              */
             this.setSecureCookie = function (enable) {
-                if(location.protocol !== 'https:') {
+                if(enable && location.protocol !== 'https:') {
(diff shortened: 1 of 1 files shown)


js/piwik.js
```diff
@@ -5759,9 +5762,30 @@ if (typeof window.Matomo !== 'object') {
              * @param bool
              */
             this.setSecureCookie = function (enable) {
+                if(location.protocol !== 'https:') {
+                    logConsoleError("Error in setSecureCookie: You cannot use `Secure` on http.");
```
tsteur, line 5766: @fealXX should we maybe return and not set `configCookieIsSecure =true` when it is not on HTTPS?

Also I think we need to change the if to 

```js
if(enable && location.protocol !== 'https:') {
    logConsoleError('...')
   return;
}
```

No need to check for https when `enable=false`

js/piwik.js
```diff
@@ -5759,9 +5762,30 @@ if (typeof window.Matomo !== 'object') {
...
+            /**
+             * Set the SameSite attribute for cookies to a custom value.
+             * You might want to use this if your site is running in an iframe since
+             * then it will only be able to access the cookies if SameSite is set to 'None'.
+             * Sets to Lax if invalid parameter is passed, sets CookieIsSecure to true on None.
+             *
+             * @param string either Lax, None or Strict
+             */
+            this.setCookieSameSite = function (sameSite) {
+                if (sameSite != 'None' && sameSite != 'Lax' && sameSite != 'Strict') {
```
tsteur, line 5780: btw not too important but be great to use `!==` and below `===` for exact comparison.

Also should we maybe compare lowercase and then convert the first letter to upper case when setting the value? Just mentioning this because some users will try to set `lax` or `strict`.

Meaning be like 

```js
this.setCookieSameSite = function (sameSite) {
sameSite = String(sameSite).toLowerCase();
 if (sameSite === 'none' || sameSite === 'lax' || sameSite === 'strict') {
    sameSite = sameSite.charAt(0).toUpperCase() + sameSite.slice(1)
} else {
   logConsoleError('ignored value for sameSite. Please use either lax, none, or strict')
}
```

see above example be also good to not silently convert value to `Lax`

js/piwik.js
```diff
@@ -5759,9 +5762,30 @@ if (typeof window.Matomo !== 'object') {
...
+                    logConsoleError("Error in setSecureCookie: You cannot use `Secure` on http.");
+                }
                 configCookieIsSecure = enable;
             };
 
+            /**
+             * Set the SameSite attribute for cookies to a custom value.
+             * You might want to use this if your site is running in an iframe since
+             * then it will only be able to access the cookies if SameSite is set to 'None'.
+             * Sets to Lax if invalid parameter is passed, sets CookieIsSecure to true on None.
```
tsteur, line 5775: could we mention that `None` will only work on HTTPS?

tsteur: btw thinking we maybe could also mention that `None` might cause issues if the user is sometimes on HTTPS and sometimes not (eg a site allows access through HTTP and HTTPS).

js/piwik.js
tsteur: If a user has a site available through HTTP and HTTPS, then setting `sameSite('None')` would only work on HTTPS. On HTTP it would log an error but still try to set `sameSite=None`. I wonder if we would need to additionally add something like below

```js
if (sameSite === 'none' && location.protocol !== 'https:') {
   logConsoleError('sameSite=none wont work because it is not using HTTPS')
   return;
}
```

this way we would keep `Lax` and a user can conveniently set `sameSite('None')` and know it will only be applied on HTTPS traffic.
We would still keep the additional ` if(location.protocol !== 'https:') {` check in `setSecureCookie` as this is useful too.

js/piwik.js
tsteur: also thinking we might need to mention that setting `None` will also directly set `secure cookie flag` and if users browse HTTP and HTTPS during same visits it would create two visits because two different cookies would be created AFAIK

tsteur (changes requested):
Thanks for this @fealXX very appreciated . Left a few comments 👍

js/piwik.js
fealXX: As Lax is the Default value, I was reverting to that, so that a non-valid combination can't happen.
Will change in next commit.

js/piwik.js
fealXX: Done!

js/piwik.js
fealXX: Along the lines of this?
setSecureCookie
"Warning: If your site is available under http and https, setting this might lead to duplicate or incomplete visits."

setCookieSameSite
"Warning: Sets CookieIsSecure to true on None, because None will only work with Secure; cookies If your site is available under http and https, using "None" might lead to duplicate or incomplete visits."

js/piwik.js
fealXX: I don't think it is necessary to revert to Lax on HTTP, as the usecase for Samesite=None will only work on https and the cookie will not be accepted by the browser in that case anyways.

js/piwik.js
```diff
@@ -5756,12 +5759,45 @@ if (typeof window.Matomo !== 'object') {
...
+             *
+             * Warning: 
+             * Sets CookieIsSecure to true on None, because None will only work with Secure; cookies
+             * If your site is available under http and https,
+             * using "None" might lead to duplicate or incomplete visits.
+             *
+             * @param string either Lax, None or Strict
+             */
+            this.setCookieSameSite = function (sameSite) {
+                if (sameSite === 'none' || sameSite === 'lax' || sameSite === 'strict') {
```
tsteur, line 5789: @fealXX could we do add a `sameSite = String(sameSite).toLowerCase()` as a first line? So it also works for `None` and `NONE` etc? then below `else if` could be changed to `else` like 

```js
this.setCookieSameSite = function (sameSite) {
    sameSite = String(sameSite).toLowerCase()
     if (sameSite === 'none' || sameSite === 'lax' || sameSite === 'strict') {
         sameSite = sameSite.charAt(0).toUpperCase() + sameSite.toLowerCase().slice(1);
    } else (sameSite !== 'None' && sameSite !== 'Lax' && sameSite !== 'Strict') {
                    logConsoleError('Ignored value for sameSite. Please use either Lax, None, or Strict.');
       return;
     }
    ....
```

(also added a `.toLowerCase()` where it converts the sameSite to upper case first then lower case

js/piwik.js
```diff
@@ -5756,12 +5759,45 @@ if (typeof window.Matomo !== 'object') {
...
+                    sameSite = sameSite.charAt(0).toUpperCase() + sameSite.slice(1);
+                } else if (sameSite !== 'None' && sameSite !== 'Lax' && sameSite !== 'Strict') {
+                    logConsoleError('Ignored value for sameSite. Please use either Lax, None, or Strict.');
+                    return;
+                }
+                if (sameSite === 'None') {
```
tsteur, line 5795: I reckon it be still great to not set `sameSite=None` on HTTP just to make sure the cookie still works on HTTP. So be great to convert it to `Lax` on HTTP protocol when it is otherwise none. I mean like 

```js
if (sameSite === 'None') {
   if (location.protocol === 'https:') {
       this.setSecureCookie(true);
   } else { sameSite = 'Lax' }
} 
```

I tested it with ` document.cookie = 'foo=bar;path=/;secure;SameSite=None'` on HTTP and it would otherwise show issues like this 
image

js/piwik.js
tsteur: actually I just realise now that the code below also supports already like lowercase and upper case first so could also ignore this change. It be only interesting if we wanted to support `NONE` etc but that's not needed.

tsteur (commented):
👍  thanks for the quick changes @fealXX. Left another minor comment. Once this is changed will do some functional tests

js/piwik.js
fealXX: I do think that is a more elegant solution though.

js/piwik.js
fealXX: The cookie will be set, but it will not be sent with subsequent requests, if you're in a situation where sameSite=None would be needed. Added the fallback and a warning for reverting to Lax nonetheless

tsteur:
Worked for me. Wasn't sure if it affects eg deleteCookie but all seems to work and seems safe to merge for 4.0 release.

We will next need to merge this into a different branch in our repo, build the minified JS and add a simple test to `API methods` that the method exists and update the documentation. Hoping to do this tomorrow.

PR #16727, opened 2020-11-17
Add convenience method for rescheduling archiving for a single plugin.

diosmosis:
### Description:

Add new method Plugin::schedulePluginReArchiving() to allow rescheduling archiving for a plugin.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Plugin.php
+++ b/core/Plugin.php
@@ -8,6 +8,8 @@
  */
 namespace Piwik;
 
+use Piwik\Archive\ArchiveInvalidator;
+use Piwik\Container\StaticContainer;
 use Piwik\Plugin\Dependency;
 use Piwik\Plugin\Manager;
 use Piwik\Plugin\MetadataLoader;
@@ -461,6 +463,38 @@ class Plugin
         ));
     }
 
+    /**
+     * Schedules re-archiving of this plugin's reports from when this plugin was last
+     * deactivated to now. If the last time core:archive was run is earlier than the
+     * plugin's last deactivation time, then we use that time instead.
+     *
+     * Note: this only works for CLI archiving setups.
+     *
+     * Note: the time frame is limited by the `[General] rearchive_reports_in_past_last_n_months`
+     * INI config value.
+     *
+     * @throws \DI\DependencyException
+     * @throws \DI\NotFoundException
+     */
+    public function schedulePluginReArchiving()
+    {
+        $lastDeactivationTime = $this->getPluginLastDeactivationTime();
+
+        $dateTime = null;
+
+        $lastCronArchiveTime = (int) Option::get(CronArchive::OPTION_ARCHIVING_FINISHED_TS);
+        if (empty($lastCronArchiveTime)) {
+            $dateTime = $lastDeactivationTime;
+        } else if (empty($lastDeactivationTime)) {
+            $dateTime = $lastCronArchiveTime;
+        } else {
+            $lastCronArchiveTime = Date::factory($lastCronArchiveTime);
+            $dateTime = $lastDeactivationTime->isEarlier($lastCronArchiveTime) ? $lastDeactivationTime : $lastCronArchiveTime;
+        }
+
+        $archiveInvalidator = StaticContainer::get(ArchiveInvalidator::class);
+        $archiveInvalidator->scheduleReArchiving('all', $this->getPluginName(), $report = null, $dateTime);
+    }
 
     /**
      * Extracts the plugin name from a backtrace array. Returns `false` if we can't find one.
--- a/core/Plugin.php
+++ b/core/Plugin.php
@@ -486,7 +486,7 @@ class Plugin
         if (empty($lastCronArchiveTime)) {
             $dateTime = $lastDeactivationTime;
         } else if (empty($lastDeactivationTime)) {
-            $dateTime = $lastCronArchiveTime;
+            $dateTime = Date::factory($lastCronArchiveTime);
         } else {
             $lastCronArchiveTime = Date::factory($lastCronArchiveTime);
             $dateTime = $lastDeactivationTime->isEarlier($lastCronArchiveTime) ? $lastDeactivationTime : $lastCronArchiveTime;
--- a/core/Plugin.php
+++ b/core/Plugin.php
@@ -492,6 +492,10 @@ class Plugin
             $dateTime = $lastDeactivationTime->isEarlier($lastCronArchiveTime) ? $lastDeactivationTime : $lastCronArchiveTime;
         }
 
+        if (empty($dateTime)) { // sanity check
+            $dateTime = null;
+        }
+
         $archiveInvalidator = StaticContainer::get(ArchiveInvalidator::class);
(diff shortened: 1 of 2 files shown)


core/Plugin.php
```diff
@@ -461,6 +463,38 @@ public function getMissingDependenciesAsString($piwikVersion = null)
...
+     */
+    public function schedulePluginReArchiving()
+    {
+        $lastDeactivationTime = $this->getPluginLastDeactivationTime();
+
+        $dateTime = null;
+
+        $lastCronArchiveTime = (int) Option::get(CronArchive::OPTION_ARCHIVING_FINISHED_TS);
+        if (empty($lastCronArchiveTime)) {
+            $dateTime = $lastDeactivationTime;
```
tsteur, line 487: in this case dateTime be an integer?

core/Plugin.php
```diff
@@ -461,6 +463,38 @@ public function getMissingDependenciesAsString($piwikVersion = null)
...
+        } else if (empty($lastDeactivationTime)) {
+            $dateTime = $lastCronArchiveTime;
+        } else {
+            $lastCronArchiveTime = Date::factory($lastCronArchiveTime);
+            $dateTime = $lastDeactivationTime->isEarlier($lastCronArchiveTime) ? $lastDeactivationTime : $lastCronArchiveTime;
+        }
+
+        $archiveInvalidator = StaticContainer::get(ArchiveInvalidator::class);
```
tsteur, line 495: maybe we also add an `if (empty($dateTime))  { dateTime = null} ` just in case a value is false or 0?

core/Plugin.php
```diff
@@ -461,6 +463,38 @@ public function getMissingDependenciesAsString($piwikVersion = null)
...
+    {
+        $lastDeactivationTime = $this->getPluginLastDeactivationTime();
+
+        $dateTime = null;
+
+        $lastCronArchiveTime = (int) Option::get(CronArchive::OPTION_ARCHIVING_FINISHED_TS);
+        if (empty($lastCronArchiveTime)) {
+            $dateTime = $lastDeactivationTime;
+        } else if (empty($lastDeactivationTime)) {
+            $dateTime = Date::factory($lastCronArchiveTime);
```
tsteur, line 489: @diosmosis if `Option::get(CronArchive::OPTION_ARCHIVING_FINISHED_TS); ` returns `false` then here it would end up being `Date::factory(0)` which may cause issues? It would be 

```
The date '1970-01-01' is a date before first website was online. Try date that's after Aug 6, 1991 (timestamp 681436800).: 0
```

core/Plugin.php
tsteur: be also good to add this check not that we're passing like `$archiveInvalidator->scheduleReArchiving('all', $this->getPluginName(), $report = null, 0);` or anything similar

core/Plugin.php
diosmosis: `$lastDeactivationTime` should be a Date

core/Plugin.php
diosmosis: in that case the first if branch should be hit (`empty($lastCronArchiveTime)`)

core/Plugin.php
diosmosis: 👍

core/Plugin.php
tsteur: 👍  didn't check but might be good to have a test for it if it's not the case yet

diosmosis: should be covered by the first test case in PluginTest

core/Plugin.php
```diff
@@ -461,6 +463,42 @@ public function getMissingDependenciesAsString($piwikVersion = null)
...
```
sgiehl, line 489: Shouldn't `$dateTime` be `null` if the plugin had not been deactivated before, so it triggers archiving for all dates in the past? 🤔

core/Plugin.php
```diff
@@ -461,6 +463,42 @@ public function getMissingDependenciesAsString($piwikVersion = null)
...
+        } else {
+            $lastCronArchiveTime = Date::factory($lastCronArchiveTime);
+            $dateTime = $lastDeactivationTime->isEarlier($lastCronArchiveTime) ? $lastDeactivationTime : $lastCronArchiveTime;
```
sgiehl, line 492: Assuming that the last archive run time is smaller than the deactivation date, wouldn't it make sense to simply skip calling `scheduleReArchiving`. As no archives should have been built for any newer dates yet, they should be build with the next archiving process nevertheless, shouldn't they?

sgiehl (changes requested):

core/Plugin.php
tsteur: last archive run time is not really so reliable. AFAIK it gets updated even if only few archives finish etc but doesn't mean they all finished. I reckon it's fine in worst case to rather rearchive a few days of reports

core/Plugin.php
tsteur: @diosmosis correct me if I'm wrong but the rearchive method should then limit it to say max 6 months (or whatever value is configured)

tsteur: it would be limited here: https://github.com/matomo-org/matomo/blob/4.0.0-rc4/core/Archive/ArchiveInvalidator.php#L470-L472

core/Plugin.php
tsteur: I double checked and it does 
image

tsteur:
@diosmosis looking through all the comments etc this one looks good to merge for me. Can you confirm?

core/Plugin.php
diosmosis: they wouldn't because I don't think core:archive will create invalidations for them, but that's another problem I think... @tsteur if core:archive hasn't run for a few days, should we automatically create invalidations for those days? We'd check the last run time and if it's more than just today/yesterday, add more invalidations.

diosmosis:
@tsteur will merge

core/Plugin.php
sgiehl: @diosmosis did you have a look at my comment? Still thinking using `$lastCronArchiveTime` in this case seems wrong.

tsteur: @sgiehl it should then be limited see link above or not? also if `lastCronArchiveTime` is empty then it would have went in the first `if`

sgiehl: But `$dateTime` is set to the last archiving time, so the reports would be built from that date on, or not?

tsteur: I see. Assuming it last ran  today it would only rearchive today? Is this what you mean? We could set it to `null` in that case?

tsteur: @sgiehl created https://github.com/matomo-org/matomo/pull/16748 good find 👍

diosmosis: I think it's not expected for the plugin deactivation time to be unset except in the case of installing a plugin (ie, the normal workflow is deactivate plugin, later activate, which triggers this code). In that case we can just set it to null?

PR #16728, opened 2020-11-17
take into account broken archive invalidations when resetting failed jobs

diosmosis:
### Description:

Refs #16689 

Take into account broken archive invalidations that do not have ts_started set (since they were added before ts_started was introduced).

cc @tsteur 

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (approved):
lgtm

PR #16730, opened 2020-11-17
Remove unneeded metric format in Goal overview

### LINKED ISSUE #16726: Goal conversions wrongly formatted when using German language
See https://demo.matomo.cloud/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=year&date=2020-11-15&segment=&category=Goals_Goals&subcategory=General_Overview when selecting `Deutsch` (German) in the language selector.

image

You can see eg conversions should be "232469" but it is shown as `232,47` in German which means `232.47` in English. So the shown value is 1000 times lower than expected. This also applies to the other goals on the same page

image

and it applies to the individual goal pages

image

Was wondering if it is maybe related to https://github.com/matomo-org/matomo/issues/15262 but don't think so.

@sgiehl maybe we already have an issue for this? Couldn't find anything. I suppose it's not an easy fix? (just asking in case you remember from previous fixes)

sgiehl:
### Description:

fixes #16726

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
Thank you @sgiehl this works for me. Also applied the patch on the org demo (will be overwritten on next update )https://demo.matomo.org/index.php?module=CoreHome&action=index&idSite=62&period=day&date=yesterday#?idSite=62&period=year&date=2020-11-10&segment=&category=Goals_Goals&subcategory=General_Overview

PR #16684, opened 2020-11-09
Set name and default value for certain dimensions.

diosmosis:
### Description:

As title. FYI @tsteur 

Refs #16609 

### Review

* [ ] Functional review done
* [ ] Usability review done (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] Security review done
* [ ] Code review done
* [ ] Tests were added if useful/possible
* [ ] Reviewed for breaking changes
* [ ] Developer changelog updated if needed
* [ ] Documentation added if needed
* [ ] Existing documentation updated if needed

PR #16731, opened 2020-11-17
Improve newly added dimensions/metrics

sgiehl:
### Description:

Due to #16684 some dimensions are not showing up in custom reports and have automatically configured metrics. Some of them are not useful:

* The Provider column shouldn't have a name, as it should not appear as dimension at all. It's already available through Provider plugin and will/should be moved there in the future.

* All browser plugin dimensions are currently having automatically configured metrics. But metrics such as `Avg. Plugin (Java) Per Visit` doesn't make sense. So only configure a useful metric for each plugin.

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
works 
image

PR #16733, opened 2020-11-17
SameSite cookie attribute can be configured for JS tracker

### LINKED ISSUE #16161: New visitor ID generated per action in iframes (SameSite cookie issue)

tsteur:
### Description:

follow up from https://github.com/matomo-org/matomo/pull/16724 https://github.com/matomo-org/matomo/pull/16724

fix https://github.com/matomo-org/matomo/issues/16161

Thanks to @beger and @fealXX

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
fyi updated the JS API reference and created https://matomo.org/faq/how-to/how-do-i-track-a-website-within-an-iframe/

tsteur:
build js

Issue #16741, opened 2020-11-18
Installing Matomo using ExtraTools plugin


nrenault:
I'm using ExtraTools plugin to install Matomo.

Everything is working with the last stable version of Matomo (3.14.1).

Today I try to install Matomo 4.0.0-rc4 but I have an error when I want to activate ExtraTools (before installing Matomo):

`[PDOException]  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'db_matomo.option' doesn't exist`

When I look the config file, activated plugins seems not here anymore.

If I understand, with Matomo 4, plugins (activation/desactivation) are now managed in database instead of the config file, is it true ?

There is any way to be able to use ExtraTools to install Matomo anymore ? (or something else to be able to make a full automatic installation of matomo)

Thanks !

sgiehl:
Hi @nrenault 

The de-/activation is still managed in the config file. But whenever a plugin is de-/activated we store a timestamp in the options table to know when the plugin was last de-/activated.

I haven't used the `ExtraTools` plugin yet, so can't say what needs to be changed to get it working again.

nrenault:
Ok thanks you for details.

I use ExtraTools to install Matomo, but for working, the plugin needs to be activated and after install Matomo. So if add/edit datas in the database is mandatory to activate/desactivate a plugin, it's not possible anymore to use it for install Matomo.

tsteur:
I created a PR in https://github.com/matomo-org/matomo/pull/16744/files not sure if these changes change anything though?

nrenault:
Hi @tsteur

I will wait to test it to confirm you, but it's sounds good and I think it will solve my issue with ExtraTools.

Thanks for you reactivity !

nrenault:
Hello,

I checked this morning with RC6 version, but I still have the same error.

```
matomo]# php console plugin:activate ExtraTools
ERROR [2020-11-20 05:17:44] 9902  Uncaught exception: /matomo/libs/Zend/Db/Statement/Pdo.php(236): SQLSTATE[42S02]: Base table or view not found: 1146 Table 'matomo.option' doesn't exist
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'matomo.option' doesn't exist
[Zend_Db_Statement_Exception] SQLSTATE[42S02]: Base table or view not found: 1146 Table 'db_matomo.option' doesn't exist
[...]

PR #16744, opened 2020-11-18
Ignore plugin timings when matomo is not installed yet

tsteur:
### Description:

refs https://github.com/matomo-org/matomo/issues/16741

Not sure if it will fix the issue as I don't know how ExtraTools works. Figured this might still be good to have though maybe just in case to prevent possible install errors. Could maybe even skip saving the timing while Matomo is not installed yet maybe but be good to have them if/when possible. Might be useful for cloud too haven't checked yet (edit: tested and cloud setup still works normal)

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Plugin/Manager.php
+++ b/core/Plugin/Manager.php
@@ -527,7 +527,8 @@ class Manager
 
         // execute deactivate() to let the plugin do cleanups
         $this->executePluginDeactivate($pluginName);
-        $this->savePluginDeactivationTime($pluginName);
+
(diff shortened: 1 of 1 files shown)


tsteur:
@diosmosis do we actually need the last activation time at all? If I see things right we're not using it so could maybe ignore it. Although can also keep it assuming it might be useful for troubleshooting in future

diosmosis:
@tsteur I don't think it's currently useful for anything, just seemed like it might be useful in the future. Will merge this.

PR #16747, opened 2020-11-18
Assume hidden keywords to be available by default

### LINKED ISSUE #16742: Use `hiddenkeyword` for Search engines by default
while this mostly affects https://github.com/matomo-org/searchengine-and-social-list/ I think this should be changed in Matomo itself
reported in https://forum.matomo.org/t/search-engines-in-the-list-of-websites/37192/12?u=lukas

Matomo doesn't detect a lot of search engines anymore as the detection assumes that the referrer URL contains the search query specified in the search engine list. But hardly any search engine sends it anymore, so it doesn't work.

As a workaround it seems like the `hiddenkeyword` keyword was added to the yaml file to specify the URL that is used to do the redirect over and which will be then the referrer URL.

But as referrer policy and such things are now widespread, no one opens links via redirects anymore and this is again pointless.

One could fix this issue by adding

```yaml
hiddenkeyword:
  - '/^$/'
  - '/'
```

to every search engine in the list, but at this point the setting becomes useless and instead I think Matomo should instead handle every referrer URL like `https://www.startpage.com/` as a search engine independent on the exact URL or query parameters.

sgiehl:
### Description:

fixes #16742

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
Would it be a good idea to add a visit w/ a hiddenkeyword referrer to a system test as well as the unit test used?

sgiehl:
@diosmosis added simple system test

PR #16751, opened 2020-11-19
Restrict tracking requests to max 10000 per tracker per page load

tsteur:
### Description:

We limit the tracking requests now to 10000 per page load by default. This is to prevent issues where there is a bug in the website and for example page views are being tracked in a loop by accident due to a bug. You might notice this bug and fix this bug but then there might be still visitor who have a previous version cached or haven't reloaded the page (eg they watch a video or listen to audio or just didn't close the browser window). This way, if a visitor has the page open for quite a long time this automatically stops tracking requests from being sent. 

The limit can be customised to your liking. 

The counter will not be reset on `trackPageView` but only on an actual page reload as the counter is kept in memory.

Logs an error message should limit be reached.

Putting this into 4.0 release since it could be seen as a breaking change if someone was to track a LOT of requests during one page load.

Not sure if there's a better name than "page load" but "page view" be confusing as we don't reset the counter on page view because `trackPageView` might be actually called in a loop and it would prevent the feature from working. 

Will update JS API reference once merged

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
build js

sgiehl:
Wondering if it maybe would make sense to automatically reset that limit after a day or something like this. There might be other tracking cases for single page applications that are actually almost never be reloaded, like order terminals for anything

tsteur:
@sgiehl that's what we're trying to prevent though that they keep sending requests still after 1 day etc when there's no pagereload etc. I think in this case they could maybe set a higher limit or disable this functionality with `-1` (added this feature).

I would say we could reset the counter on `setVisitorId` (this would maybe restore old behaviour in very few cases but not in SPA), or we could reset it when `trackPageView` is called and the URL is different compared to last time (but this would kind of prevent the feature from working where there is a bug so could almost just as well not merge the feature as it would only add complexity plus this behaviour be hard to document and explain).

I reckon 10K req even for a SPA is quite a bit and not that easily reached. Could also set it to 20K by default if that helps.

tsteur:
build js

sgiehl:
@tsteur that was just a random thought. Guess support for `-1` should be enough

tsteur:
Actually after thinking for 30 min about it I'm closing this PR for now. If we start resetting counter after certain time or actions things become less predictable and hard to explain/document/reproduce and it's not usable. The only nice way would be to have this in default tracking code where besides `trackPageView` we also have a `setMax...(10000)` but this bloats the embed tracking code for maybe a small problem. Generally it be hard to find for some users there is such a limit etc.

PR #16754, opened 2020-11-19
Update PHP extension requirements & deprecate Common::mb_* methods

sgiehl:
### Description:

Our System check currently marks some PHP extensions as required that we actually don't need to require.

* SPL
* Reflection

Those two are meanwhile built in PHP and can't be unavailable, so doesn't make sense to check them at all.

* mbstring

We indirectly already had `symphony/polyfill-mbstring` included in the composer requirements, so the methods are always available. Checking for the extension is therefor not needed anymore, as everything should work without it. (We could also consider adding this as recommended module)

* iconv 

I've added `symphony/polyfill-iconv` to ensure those methods are always available. We are actually not using that often, but the mbstring polyfill uses it and it allows us to remove the requirement. (We could also consider adding this as recommended module)

In addition I have marked or `Common::mb_*` methods as deprecated and replaced their usage with the `mb_*` functions directly, as those methods are actually doing nothing else. (We should do that for our plugins later as well)

refs #16650

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
btw just in case it was planned be good to not replace any further upgrade methods since we don't have any issues there currently and it doesn't really improve anything.

PR #16755, opened 2020-11-19
Introduce PHP CS to improve code quality

sgiehl:
### Description:

In order to improve our code quality a bit, it would imho be awesome by stating to introduce automatic PHPCS checks.
I've started with a very basic ruleset, that only disallows the usage of short open tags and forces using \n as line endings.

There is also a new workflow, that will automatically run for Pull Requests, so it can be directly seen if a PR breaks any coding rules. 

If that PR get's merged I would add some other rules step by step and directly provide fixes for code that doesn't comply. Maybe anyone else also has some suggestions for rules we should use.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -1,0 +1,17 @@
+name: PHPCS check
+
+on: pull_request
+
+jobs:
+  phpcs:
+    name: PHPCS
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: composer install --dev --prefer-dist --no-progress --no-suggest
+      - name: PHPCS check
+        uses: chekalsky/phpcs-action@v1
+        with:
+          phpcs_bin_path: './vendor/bin/phpcs'
+          enable_warnings: true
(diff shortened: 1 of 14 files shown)


sgiehl:
I've added another commit that adds a short open tag (should be removed before merge). The check fails now because of it. The error is also directly highlighted in the changed files tab.

tsteur:
sure could start with just this and see over time if/when we run into issues that could have been detected through this and then add it when needed (or when we notice something as part of a review we could think whether we can automatically detect certain thing in the future). Generally not so much coding style as it's not as important but probably more like `not using unserialize method directly` checks etc.

sgiehl:
Disallowing some methods like `eval` or `create_function` seems also be good. Will add a simply rule for that and for forcing to use `Common::safe_unserialize` instead of `unserialize`. Any other method we should disallow?

tsteur:
> Any other method we should disallow?

Not needed for now. Be good to keep it simple.

sgiehl:
@tsteur @diosmosis can we merge this one? Might be helpful to have some basic checks and being able to add new checks if something pops up in the future that would have been detectable with phpcs...

tsteur:
👍

PR #16760, opened 2020-11-20
multi_server_environment setting now disallows config edits from Web UI

### LINKED ISSUE #14390: multi_server_environment does not prevent config changes from WebUI
I installed a Matomo installation 3.9.1 on two servers, both running CentOS 7 with nginx and php 7.2.10.
If setting the setting multi_server_environment=1 in config.ini.php I would expect that I can not install plugins and change configurations from the web interface, according to https://matomo.org/faq/new-to-piwik/faq_134/

> this setting lets Matomo know you’re using multiple servers and for example it will then not allow the installation of a plugin via the UI (or the plugin would be only installed on one server), or it will not allow config file changes from the UI (as the config file would be updated on one server only).

The installation of plugins is not permitted. However, I can still change configurations from the web interface, for example "System" -> "General settings" -> "Trusted Matomo Hostname" which results in inconsistent configurations on the servers:

```
$server1:  cat config.ini.php 

...
[General]
multi_server_environment = 1
trusted_hosts[] = "f1-stats.de"
trusted_hosts[] = "f2-stats.de"
trusted_hosts[] = "stats.de"
trusted_hosts[] = "test.example.com"
...
$server1:  md5sum config.ini.php 
9721fb51375b215f1587148a7b0752bf  config.ini.php
```

```
$server2:  cat config.ini.php 

...
[General]
multi_server_environment = 1
trusted_hosts[] = "f1-stats.de"
trusted_hosts[] = "f2-stats.de"
trusted_hosts[] = "stats.de"
...
$server2:  md5sum config.ini.php 
4d655c14bd8493ac1f2d592d98c92d71  config.ini.php
```

"Diagnostic" -> "Config file" also shows, that multi_server_environment is set to 1.

nina-py:
### Description:

Turning the multi_server_environment setting on now hides the following
settings from the Administration -> Settings -> General settings page:

- Archiving settings
- Email server
- Trusted Matomo hostname
- Update settings

...so that users cannot set these separately on different instances of Matomo. Assuming I got these right, are there any other settings that should be disabled when `multi_server_environment` is turned on?

Fixes #14390.

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
Makes sense @nina-py 

Might need to add a note in `config/global.ini.php` for `multi_server_environment` that this flag doesn't need to be enabled when the config file is on a shared filesystem like NFS or EFS.

nina-py:
Thanks @tsteur, added a note to the global config file.

nina-py:
Hi @tsteur , is there anything else that needs to be done in this PR? Or are you guys just way too busy at the moment following the 4.0.x release?

tsteur:
@nina-py at the moment we're indeed quite busy with the Matomo 4 release. @sgiehl can you do another review maybe?

plugins/CoreAdminHome/templates/generalSettings.twig
```diff
@@ -10,7 +10,7 @@
     {{ ajax.errorDiv() }}
     {{ ajax.loadingDiv() }}
 
-{% if isGeneralSettingsAdminEnabled %}
+{% if isGeneralSettingsAdminEnabled and not isMultiServerEnvironment %}
```
sgiehl: I guess that is actually not needed to hide the archiving settings. They are stored in the database when configured through UI. See https://github.com/matomo-org/matomo/blob/f5e9420a987340b036fa342e876ab92e314f4ec7/core/ArchiveProcessor/Rules.php#L265-L272

Only if that values is not available the one in config should be used.

Hiding the email settings only makes more sense, as they are stored in config only.

plugins/CoreUpdater/SystemSettings.php
```diff
@@ -52,19 +53,20 @@ protected function init()
...
-
-        $isWritable = Piwik::hasUserSuperUserAccess() && CoreAdminController::isGeneralSettingsAdminEnabled();
         $dbSettings = new Settings();
         if ($isWritable && $dbSettings->getUsedCharset() !== 'utf8mb4' && DbHelper::getDefaultCharset() === 'utf8mb4') {
             $this->updateToUtf8mb4 = $this->createUpdateToUtf8mb4();
         }
+
+        $isWritable = $isWritable && PluginUpdateCommunication::canBeEnabled();
+        $this->sendPluginUpdateEmail = $this->createSendPluginUpdateEmail();
+        $this->sendPluginUpdateEmail->setIsWritableByCurrentUser($isWritable);
```
sgiehl: Actually only the release channel is a configuration that is managed in config. The other values are "normal" settings, stored in database. So for those it wouldn't hurt to have them writable.

sgiehl (changes requested):

plugins/CoreAdminHome/templates/generalSettings.twig
nina-py: Thanks @sgiehl, I'll tackle this over the weekend.

nina-py: @sgiehl, I updated the template to hide only the Email Server settings if in multi-server environment.

plugins/CoreUpdater/SystemSettings.php
nina-py: Thank you for that, I have updated the code to reflect that (and did a very minor refactor to remove duplicate code).

nina-py:
Hi @sgiehl, in addition to the changes you requested earlier, I have updated a UI screenshot (an extra comment for the setting about half-way through the screenshot): https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/44409/UIIntegrationTest_admin_diagnostics_configfile.png. CI build in progress.

As far as I can see, the other failures in those two jobs are not related to the changes introduced in this PR. I've also rebased the branch. Please review.

nina-py:
@sgiehl, just to follow up on the most recent CI build for this PR - the updated screenshot matches but there is one failure elsewhere with the following:

```bash
There was 1 failure:

1) Piwik\Tests\Integration\ReleaseCheckListTest::test_screenshotsStoredInLfs

Some Screenshots are not stored in LFS: tests/UI/expected-screenshots/UIIntegrationTest_admin_diagnostics_configfile.png

Failed asserting that an array is empty.

/home/travis/build/matomo-org/matomo/tests/PHPUnit/Integration/ReleaseCheckListTest.php:180
```

Do I need to do anything about this one?

sgiehl:
@nina-py  We are storing our screenshots in Git LFS to not waste space in this repo. You may need to install Git LFS to update the the srceenshots properly

nina-py:
Thank you @sgiehl, I think I got this right this time around - the commit diff shows

```
Git LFS file not shown 
```

but the build is yet to run.

sgiehl (approved):
Looks good now 👍  @nina-py

nina-py:
Thank you @sgiehl for all your help with this one!

PR #16765, opened 2020-11-22
core:archive add note message when continuing an existing queue

diosmosis:
### Description:

Noticed core:archive no longer says when it's continuing an existing archive run (if it ever did). Found it confusing for it to just pick up a different queue locally, so thought this message might be useful.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/CronArchive.php
+++ b/core/CronArchive.php
@@ -308,6 +308,10 @@ class CronArchive
         $this->allWebsites = $websitesIds;
         $this->websiteIdArchiveList = $this->makeWebsiteIdArchiveList($websitesIds);
 
+        if ($this->websiteIdArchiveList->isContinuingPreviousRun()) {
+            $this->logger->info("- Continuing ongoing archiving run by pulling from shared idSite queue.");
+        }
+
         if ($this->archiveFilter) {
             $this->archiveFilter->logFilterInfo($this->logger);
         }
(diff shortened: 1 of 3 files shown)


core/CronArchive.php
```diff
@@ -1001,6 +1000,11 @@ private function logArchiveTimeoutInfo()
             $this->logger->info("- Archiving was last executed without error "
                 . $this->formatter->getPrettyTimeFromSeconds($dateLast, true) . " ago");
         }
+
+
+        if ($this->websiteIdArchiveList->isContinuingPreviousRun()) {
```
tsteur, line 1005: @diosmosis not too important since we don't have it for any of the other methods in those classes but might make sense to add an interface to the sharedsiteids and fixedsiteids just to make sure those things are implemented (or could check if the method exists)

tsteur (changes requested):
@diosmosis 

getting this error because at this point it seems websiteIdArchiveList is not defined yet?

> Uncaught exception: Error: Call to a member function isContinuingPreviousRun() on null in core/CronArchive.php:1005
Stack trace:
#0 core/CronArchive.php(295): Piwik\CronArchive->logArchiveTimeoutInfo()
#1 core/CronArchive.php(266): Piwik\CronArchive->init()

diosmosis:
@tsteur updated

diosmosis:
@tsteur fyi, merged this one

PR #16766, opened 2020-11-22
Add regular expression support to list of user agents to exclude

### LINKED ISSUE #14186: List of user agents to exclude could support regular expressions
Currently it is not clear whether we support regular expressions for the setting "List of user agents to exclude".
In the list of IP addresses it says `Regular expressions such as /^sess.*|.*[dD]ate$/ are suported.`

Suggested:
* Check regular expression works when excluding user agents
* add `Regular expressions such as /^Good-example-of-regex-here$/ are supported.`
* rename `suported` to `supported` in the existing message (typo)

nina-py:
### Description:

- Added regex support to Administration -> Websites -> Settings ->
Global list of user agents to exclude
- Made sure old tests that use stripos() pass
- Added new tests
- Added a sentence to the inline help area about regex support
- Fixed a typo elsewhere as specified in the original issue.

Closes #14186.

### Review

* [ ] Functional review done
* [ ] Code review done


core/Tracker/VisitExcluded.php
```diff
@@ -360,8 +360,18 @@ protected function isUserAgentExcluded()
 
         if (!empty($excludedAgents)) {
             foreach ($excludedAgents as $excludedUserAgent) {
-                // if the excluded user agent string part is in this visit's user agent, this visit should be excluded
-                if (stripos($this->userAgent, $excludedUserAgent) !== false) {
+                // Build up a regex if the string supplied is not a regular expression.
+                // The resulting regex is case-insensitive to match the behaviour of stripos().
+                $userAgentRegex = $excludedUserAgent;
+                if (preg_match('~^[^/].+?[^/]i?$~', $excludedUserAgent)) {
+                    $userAgentRegex = '/' . $userAgentRegex . '/i';
```
sgiehl: If I see that correctly you are trying to handle regexes without surrounding `/` here. On the one side that might be a useful feature, but it might also breaks any definition that contains a special regex char on purpose. Let's assume someone wanted to exclude a certain useragent and maybe simply added `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.88 Safari/537.36` to that list. 
Your method would still match due to the stripos, but the generated regex would actually also match other useragents maybe. Maybe not the best example, but there might be definitions that might contain a `|` on purpose. Maybe we should not try building a regex but use a similar method as for checking the excluded url params instead:
https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/core/UrlHelper.php#L38-L52

You could maybe simply move the method to the `Common` class and reuse it for both....

sgiehl (changes requested):

core/Tracker/VisitExcluded.php
nina-py: Thanks for the review @sgiehl, I was trying to avoid a `@preg_match()` call with that trying to build a regex if supplied a plain string but on reflection it doesn't seem like a good idea. I could have used `preg_quote()` to escape special characters but it's probably more consistent to discard non-valid regular expressions and evaluate them with `stripos()` only as before.

I have had a look at the similar function you've highlighted and I don't think I can use it here without reworking it quite a bit. Instead, I've reworked the `isUserAgentExcluded()` method and updated the tests to include one for an invalid regex.

nina-py: Hi @sgiehl, is there anything else that needs to be done for this PR? I have noticed some UI tests are failing but it's unclear to me if that is anything to do with the changes proposed in this PR.

plugins/WebsiteMeasurable/MeasurableSettings.php
```diff
@@ -235,7 +235,9 @@ private function makeExcludedUserAgents()
                 . '<br /><br />'
                 . Piwik::translate('SitesManager_GlobalListExcludedUserAgents_Desc')
                 . '<br />'
-                . Piwik::translate('SitesManager_GlobalExcludedUserAgentHelp2');
+                . Piwik::translate('SitesManager_GlobalExcludedUserAgentHelp2')
+                . Piwik::translate('SitesManager_GlobalExcludedUserAgentHelp3', "/bot|spider|crawl|scanner/i")
```
sgiehl: Guess would make sense to insert a space between both strings.

```suggestion
                . Piwik::translate('SitesManager_GlobalExcludedUserAgentHelp2') . " "
                . Piwik::translate('SitesManager_GlobalExcludedUserAgentHelp3', "/bot|spider|crawl|scanner/i")
```

sgiehl (changes requested):
@nina-py  Sorry for not coming back earlier. Was a bit busy with various other stuff for the 4.0 release 🙈 

I've added another small note. Besides that there are a few tests failing. Some system test files need to be updated.
See https://travis-ci.com/github/matomo-org/matomo/jobs/448352173#L813  and  https://travis-ci.com/github/matomo-org/matomo/jobs/448352172#L805

Regarding the UI tests. There are a few failing due to your changes. You can view the changes on our build artifacts server:
https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/43882/

It also displays a command that allows you downloading the changed files. But only add/commit those files that had been changed through your PR. Some tests are sometimes failing randomly.

When commiting UI file you need to have git lfs installed, as those files are stored on git lfs.

If you have problems with updating those files let me know, and I can do that for you.

plugins/WebsiteMeasurable/MeasurableSettings.php
nina-py: Thank you for the review @sgiehl! I, too, am a bit busy at the moment but will be able to follow up this PR over the weekend.

sgiehl:
@nina-py could you maybe merge in the latest changed from `4.x-dev`? If tests are passing, guess this PR would be good to merge as well.

nina-py:
@sgiehl, there are some System tests that I need to update first, and go over the UI tests to see if any failures are relevant to changes in this PR. I'll get to it over the next couple of days!

nina-py:
Hi @sgiehl, I've rebased the branch once again, updated the failing System and UI tests and finally most checks are green! The remaining three mismatched UI screenshots are not related to the changes in this branch. Please review.

sgiehl (approved):
looks good now @nina-py 

@tsteur Feel free to merge this one if we shall include it into the upcoming release

nina-py:
Thank you @sgiehl!

diosmosis:
Thanks for another contribution @nina-py !

PR #16769, opened 2020-11-22
core:archive Better detection for end of sharedSiteId queue

tsteur:
### Description:

By default core:archive uses `SharedSiteIds` which is used by multiple workers. For example it has a queue of several idSites that need to be processed. As soon as this queue has been reset we want existing `core:archive` commands to finish next time `getNextSiteId` is called. This is to not have one core:archive command running forever potentially running into memory issues etc.

So far we have two detections when queue was reset

* When the queue is empty https://github.com/matomo-org/matomo/blob/4.0.0-rc6/core/CronArchive/SharedSiteIds.php#L171-L174
* When suddenly more sites are in the queue than before https://github.com/matomo-org/matomo/blob/4.0.0-rc6/core/CronArchive/SharedSiteIds.php#L176-L180 (this works eg when there are only 1 sites in the queue, then a different `core:archive` resets the list of all idSites and there are `50` sites in the queue now. This means there must have been a reset)
* Above check does not work though when eg `core:archive` is running hourly, and there is only one site in Matomo (so the queue has always only one site) and it takes > 1 hour to archive this site. What happens archiver A inits sharedSiteIds with `idSites=[1]` then it processed first site `so queue is []`, then other `core:archive` might set it to `sharedSiteIds=[1]` again. It's bit edge case and there probably would need to be some race conditions as the next archiver should technically empty the queue  as well.

Just to be safe thought it might be good to check that `getNextSiteId` doesn't return a siteId that it already returned before. Because this would mean we have already archived that site.

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
This wouldn't work if the next site, for some reason (maybe due to multiple archivers running), isn't in the already processed queue, correct? What about setting an option w/ the timestamp when the queue is created/reset, then checking that this hasn't changed getNextSiteId() is called?

tsteur:
@diosmosis changed the PR 👍

core/CronArchive/SharedSiteIds.php
```diff
@@ -162,6 +176,13 @@ public function getNextSiteId()
             return null;
         }
 
+        if ($this->initialResetQueueTime !== $this->getResetQueueTime()) {
+            // queue was reset/finished by some other process
+            $this->currentSiteId = null;
+            $this->done = true;
+            return null;
+        }
```
diosmosis, line 188: Should we maybe log here? Maybe a debug log or an info, since it seems like it might be useful to have in CLI output...

core/CronArchive/SharedSiteIds.php
```diff
@@ -49,11 +50,24 @@ public function __construct($websiteIds, $optionName = self::OPTION_DEFAULT)
...
             return $websiteIds;
         });
-        $this->numWebsitesLeftToProcess = $this->getNumSites();
+
+        $this->initialResetQueueTime = $this->getResetQueueTime();
+    }
+
+    public function setQueueWasReset()
+    {
+        Option::set($this->optionName . self::KEY_TIMESTAMP, floor(microtime(true) * 1000));
```
diosmosis, line 67: Not an issue, but curious why microtime is used instead of just time()? Something to do with server configurations?

diosmosis (approved):
Looks good, left a couple minor comments

core/CronArchive/SharedSiteIds.php
tsteur: because time would cause maybe concurrency issues not sure. It's a common practice that some users launch multiple archivers at the same second so just wanted to have it more precise just to be safe. Probably wouldn't cause an issue to use `time()`

core/CronArchive/SharedSiteIds.php
tsteur: Not sure it's needed as we would simply stop once iterated over all sites? Not sure how to add easily...

diosmosis: I mean a quick log, like, `The shared site ID queue was reset, stopping.` (or whatever actually happens if not stopping). Might be useful when looking through the logs?

core/CronArchive/SharedSiteIds.php
diosmosis: I see 👍

core/CronArchive/SharedSiteIds.php
tsteur: @diosmosis added a debug log message. It should usually not be that important since it's a regular/normal thing maybe

diosmosis: better to use LoggerInterface instead, but otherwise looks good

diosmosis:
@tsteur merged after this small tweak: https://github.com/matomo-org/matomo/pull/16769/commits/0bba9602a76765b67fc269b9c7c49f34891e7553

PR #16772, opened 2020-11-23
Remove existing invalidations before scheduling new ones

tsteur:
### Description:

@diosmosis is this what you meant regarding Custom Reports?

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -544,11 +544,11 @@ class ArchiveInvalidator
      * since adding invalidations can take a long time and delay UI response times.
      *
      * @param int|int[]|'all' $idSites
-     * @param string $pluginName
+     * @param string|int $pluginName
      * @param string|null $report
      * @param Date|null $startDate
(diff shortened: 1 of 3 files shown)


tsteur:
Tested it locally and looked good

diosmosis:
@tsteur am adding a test currently

tsteur:
@diosmosis just did add one now... committing

core/Archive/ArchiveInvalidator.php
```diff
@@ -544,12 +544,15 @@ public function removeInvalidations($idSite, $plugin, $report = null)
      * since adding invalidations can take a long time and delay UI response times.
      *
      * @param int|int[]|'all' $idSites
-     * @param string $pluginName
+     * @param string|int $pluginName
```
tsteur, line 547: allowed int since mysqli might return `int` in some cases. We can convert to `string` though if needed eg in Funnels etc

diosmosis (approved):

PR #16773, opened 2020-11-23
Disable features that are inaccurate if visit data is not profilable

### LINKED ISSUE #16363: Hide features that don't work when the website is configured in cookie-less tracking

diosmosis:
### Description:

Introduces a new core metric, `nb_profilable` calculated with VisitsSummary metrics as the sum of visits that are profilable. If less than 1% of visits for a period are not profilable features that depend on profilable data are hidden or disabled. This includes: 

* reports that rely on detecting returning visitors (everywhere in the UI)
* the nb_uniq_visitors metric (in the UI and API)
* segments for dimensions that rely on detecting returning visitors (hidden in the UI only)
* visitor log features that rely on detecting returning visitors

Other changes:
* nb_profilable is stored even if it's value is 0, this is to be able to differentiate between archives that are old and do not have the metric vs ones that are new and have 0 profilable visits. There is an alternative to check whether ts_archived is < a certain time, but getting this information w/ numeric data is a bit complicated. It might be do-able though.

Fixes #16363 

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
@tsteur / @sgiehl this is ready for an initial review

diosmosis:
@tsteur @sgiehl this review is ready for review. could use a general review of the strategy used first.

core/DataAccess/LogAggregator.php
```diff
@@ -423,6 +423,7 @@ protected function getVisitsMetricFields()
             Metrics::INDEX_BOUNCE_COUNT                   => "sum(case " . self::LOG_VISIT_TABLE . ".visit_total_actions when 1 then 1 when 0 then 1 else 0 end)",
             Metrics::INDEX_NB_VISITS_CONVERTED            => "sum(case " . self::LOG_VISIT_TABLE . ".visit_goal_converted when 1 then 1 else 0 end)",
             Metrics::INDEX_NB_USERS                       => "count(distinct " . self::LOG_VISIT_TABLE . ".user_id)",
+            Metrics::INDEX_NB_PROFILABLE                  => "sum(IF(" . self::LOG_VISIT_TABLE . ".profilable = 1 OR " . self::LOG_VISIT_TABLE . ".user_id IS NOT NULL, 1, 0))",
```
tsteur: the second OR is not needed or could even maybe cause some error if `enable_userid_overwrites_visitorid` is disabled. That's because profilable will be already set to `1` if a user_id is set when `enable_userid_overwrites_visitorid` is enabled (see https://github.com/matomo-org/matomo/blob/4.1.1/core/Tracker/Request.php#L719-L726)

core/API/Request.php
```diff
@@ -701,4 +702,47 @@ private function shouldDisablePostProcessing()
...
+        if ($idSite === false
+            || $period === false
+            || $date === false
+            || !is_numeric($idSite)
+            || Period::isMultiplePeriod($date, $period)
+        ) {
+            return true;
+        }
+
+        $isProfilable = Request::processRequest('VisitsSummary.isProfilable', [
```
tsteur: fyi wondering if it would be better to cache the result here in the transient cache instead of caching the `isProfilable` API. It's not much of a difference though but can boost performance a bit since we won't need to process the request and trigger various events every time.

core/Plugin/Report.php
```diff
@@ -231,11 +232,23 @@ protected function init()
...
      * @api
      */
     public function isEnabled()
     {
+        if ($this->isRequiresProfilableData()
+            || (isset($this->dimension) && $this->dimension->isRequiresProfilableData())
+        ) {
+            $isProfilable = Request::isCurrentPeriodProfilable();
+            if (!$isProfilable) {
+                return false;
```
tsteur: fyi haven't looked too much through the code yet and not tested yet but I would assume this will be responsible to remove certain reports from the reporting UI page when the data is not profilable? I suppose when changing the date the reports won't change since we wouldn't reload the available reports. to be checked. just thought I leave the comment here while I have it on my mind.

plugins/API/API.php
```diff
@@ -195,6 +197,14 @@ public function getSegmentsMetadata($idSites = array(), $_hideImplementationData
 
         $cache->save($cacheKey, $segments);
 
+        if ($_hideSegmentsIfDataNotProfilable
+            && !Request::isCurrentPeriodProfilable()
+        ) {
+            $segments = array_filter($segments, function ($data) {
```
tsteur: wondering the same as with the reports... when changing the date in the UI then the list of segments might not change if the list of segments was loaded earlier for a different date? It may be fine though. I reckon we might always want to show all segments though maybe as in different periods some data may be profilable and then we would want to be able to create segments for it? Bit hard to explain maybe what I mean

core/API/Request.php
```diff
@@ -701,4 +702,47 @@ private function shouldDisablePostProcessing()
...
     }
+
+    public static function isCurrentPeriodProfilable($idSite = null, $period = null, $date = null, $segment = null)
+    {
+        if (self::isProfilableCheckDisabledInTests()) {
+            return true;
+        }
+
+        $idSite = $idSite ?: Common::getRequestVar('idSite', $default = false);
+        $period = $period ?: Common::getRequestVar('period', $default = false);
```
tsteur, line 716: I'm thinking we'd want to convert a 'day' period to at least `week`. This way, if some days have profilables and some don't the experience is a bit more consistent and less unpredictable? This can happen eg when people can opt in to cookies or not. Some days maybe all visitors opt out vs some other days some opt in etc.

plugins/CoreHome/VisitorDetails.php
```diff
@@ -17,17 +17,24 @@ class VisitorDetails extends VisitorDetailsAbstract
     public function extendVisitorDetails(&$visitor)
     {
         $visitor['userId']                      = $this->getUserId();
-        $visitor['visitorType']                 = $this->getVisitorReturning();
-        $visitor['visitorTypeIcon']             = $this->getVisitorReturningIcon();
+        if ($this->isDataProfilable) {
+            $visitor['visitorType'] = $this->getVisitorReturning();
```
tsteur, line 21: does this maybe impact the logic in `plugins/Live/templates/_visitorDetails.twig`?

plugins/API/API.php
tsteur: I suppose it's less of an issue when a site always doesn't use visitorId (profilable) then the reports will be consistent and we do a reload when changing the site anyway. It's only an issue when changing to a date where the behaviour was different but this might be fine although it can cause an issue where it means that certain reports where profilables exist won't be shown maybe. Eg we removed for matomo org cookies a while ago and wondering if we view today, and then go back in date for say 1 year, we wouldn't see some data anymore. It would affect only few data though because the most important reports would still exist and it more impacts only unique visitors etc which would be still shown when going back in time without reloading the page I think.

tsteur (commented):
Left a first few comments @diosmosis might be also good for @mattab to go over it. 

I'm not sure yet how it impacts evolution graphs etc where some dates may be profilable and some aren't? Eg in matomo org when someone used to log in we would track a userId (not anymore). Meaning some days/weeks would be profilable while some others aren't. That might be more edge case though. 

wondering generally instead of removing some reports to maybe rather show a notice like "This report won't show any data because no profilable visitors" (would need to word it better). Might make it easier to find out why certain reports are missing and shows maybe that certain information could be gained by using cookies or userIds and might reduce support requests when people would ask why a certain report isn't shown.

diosmosis:
> wondering generally instead of removing some reports to maybe rather show a notice like "This report won't show any data because no profilable visitors" (would need to word it better).

:+1: this makes sense to me. It would be strange to just click a report and suddenly not be able to see a report. I'll make this change?

core/DataAccess/LogAggregator.php
diosmosis: 👍

core/API/Request.php
diosmosis: That makes sense. No one should call VisitsSummary.isProfilable directly.

core/Plugin/Report.php
diosmosis: I'll test this, it's been a while so I can't remember if this is something I tested or not.

core/API/Request.php
diosmosis: So only disable reports/etc. for a day if the entire week has enough profilable data? Sounds doable.

core/Plugin/Report.php
diosmosis: Actually if we don't remove reports/segments but disable them some other way, this changes a lot, so I'll wait for an answer there.

tsteur:
>  I'll make this change?
sounds great

core/API/Request.php
tsteur: 👍

tsteur: We'd basically simply replace day period with week

core/Plugin/Report.php
tsteur: @mattab any thoughts? For reports I would say we always show them and simply show a message why a report doesn't show any data (because not profilable).

For segments I think @mattab wanted segments to be removed (or someone else). But they may still be needed for historical data etc. Or if sometimes you have profilable data and sometimes not (we used to have this case on matomo org but might be bit edge case)

diosmosis: Segments are removed in the UI but nowhere else. We can also disable them and provide a tooltip saying why it doesn't make sense to use them on the current period. But when creating the segment, the period is arbitrary (eg, maybe they want it to be applied to a period w/ profilable data). Seems odd to see it greyed out and realize you have to change the period to be able to use it.

Perhaps it would be better to detect when a segment in use (ie, in the `segment=` query parameter) isn't going to provide correct data because the current period is not profilable, and display a notification.

diosmosis:
@mattab can you provide a UI review + look at the comments on the review and provide your thoughts?

core/Plugin/Report.php
diosmosis: @mattab can you provide an answer/opinion on ^?

plugins/API/API.php
diosmosis: @mattab can you take a look at ^? (that we should show all segments in API output but w/ eg `requiresProfilableData` property instead of just hiding them and filter in the UI)

plugins/CoreHome/VisitorDetails.php
diosmosis: should be resolved

diosmosis:
@tsteur looked over this pr again. In it I'm adding nb_profilable as a core metric, but it is saved even if the value is 0. This is so we can tell when it was never computed. An alternative is to not save if 0 like every other metric, but in that case also save `has_computed_profilable=1` as another metric. Do you think this is a better approach?

tsteur:
@diosmosis thought about it for a while and I reckon we can simply always store the column and no other column needed. Meaning we treat the values before it was not tracked as not profilable. It be worse if it was the other way around and by default it would assume a visitor was profilable.

tsteur:
@diosmosis is this ready for a review? I know it has the label but not sure when it was added.

diosmosis:
@tsteur it's been waiting for answers from @mattab, I'll remove the Needs Review tag.

mattab:
I'll review this later today

core/API/Request.php
```diff
@@ -701,4 +704,63 @@ private function shouldDisablePostProcessing()
...
+        if ($idSite === false
+            || $period === false
+            || $date === false
+            || !is_numeric($idSite)
+            || Period::isMultiplePeriod($date, $period)
+        ) {
+            return true;
+        }
+
+        if ($period == 'day') {
```
mattab, line 732: could we add a comment why?

plugins/API/API.php
mattab: makes sense to me that we would show all segments in API output but w/ eg requiresProfilableData property instead of just hiding them and filter in the UI

plugins/CoreHome/VisitorDetails.php
```diff
@@ -17,17 +17,24 @@ class VisitorDetails extends VisitorDetailsAbstract
...
+            $visitor['visitorTypeIcon'] = $this->getVisitorReturningIcon();
+        }
+
         $visitor['visitConverted']              = $this->isVisitorGoalConverted();
         $visitor['visitConvertedIcon']          = $this->getVisitorGoalConvertedIcon();
         $visitor['visitCount']                  = $this->getVisitCount();
```
mattab: maybe visitCount could be within a `if ($this->isDataProfilable) {` ?

plugins/CoreVisualizations/Visualizations/Sparklines/Config.php
```diff
@@ -160,6 +162,14 @@ public function replaceSparklineMetric($metricNames, $replacementColumns)
      */
     public function addSparklineMetric($metricName, $order = null, $graphParams = null)
     {
+        if (!Request::isCurrentPeriodProfilable()) {
+            if (is_array($metricName)) {
+                $metricName = array_filter($metricName, function ($value) { return $value != 'nb_uniq_visitors'; });
+            } else if ($metricName == 'nb_uniq_visitors') {
```
mattab, line 178: wondering why we filter out  `nb_uniq_visitors`here, but not other metrics? or is this done for other but nb_uniq_visitors needs special treatment maybe.

plugins/Goals/Reports/GetDaysToConversion.php
```diff
@@ -66,4 +66,9 @@ public function configureReportMetadata(&$availableReports, $infos)
             return $goal['name'] . ' - ' . $name;
         });
     }
+
+    public function isRequiresProfilableData()
+    {
+        return true;
```
mattab, line 72: would the dimensions `DaysToConversion` and `VisitsUntilConversion` also need to require profilable data?

core/Plugin/Report.php
mattab: > For reports I would say we always show them and simply show a message why a report doesn't show any data (because not profilable).

Yes, this would be the desired solution, it's less likely to break things and will also reduce any possible confusion.

mattab (commented):
Haven't finished my testing but wanted to post this already.

core/API/Request.php
mattab: Not sure if i'm seeing this correctly,
but I think currently there would be 7 transient caches entries created per week.
then maybe the cacheKey should not use $date but instead $week-$year or so?

mattab:
More feedback:    


* Nice work, this will be very valuable!
* Can we return `profilable` element in the response of `Live.getLastVisitsDetails` - it would be helpful to troubleshoot which visit is profilable or not, for example to validate value of `<nb_profilable>`, or to notice why the API output may not include some elements such as `visitorType`, `visitorTypeIcon`, `visitCount`, `daysSinceFirstVisit`, `daysSinceLastEcommerceOrder` etc.
* Getting error ` The following error just broke Matomo (v4.4.1): Call to a member function isRequiresProfilableData() on null
 on /home/matt/dev/matomo/core/Plugin/Report.php(1013)` at URL `index.php?forceView=1&viewDataTable=VisitorLog&module=Live&action=getLastVisitsDetails&small=1&idSite=1&period=day&date=2021-08-27&segment=&showtitle=1&random=6365` 
* i wasn't able to trigger the display of the text in plugins/CoreHome/templates/_nonProfilableDataWarning.twig -> were you able to get this to show? (it would be great to have  UI test for it if not done yet)                                                                       
* Currently when there is no profilable data at all (cookie-less enforced), it doesn't hide all of the reports in "Behavior > engagement". It only hides some of them (the ` Returning Visits Over Time` and the `Frequency Overview` are correctly hidden, but the ` Visits by Visit Number
` and ` Visits by Days Since Last Visit` are still displayed (expected them not to be displayed in this case?)

Screenshot from 2021-08-27 16-12-58

* also when these reports show in the report footer: `This report requires profilable data, but the current period of 2021-08-27 does not have any, so this report will not have useful information and should not be relied upon.` (see screenshot above)
  * Could we change it to `This report is inaccurate should not be relied upon. This report requires profilable data (tracking cookies enabled), but none of the visits did in the current period of 2021-08-27. ` 
  * Also, could we show the notice much more visible,  using a more visible "yellow notice" /warning style? just changing the CSS of another similar notice to warning it looks like this, which would be acceptable i think (at least visible and can't really miss it)
Screenshot from 2021-08-27 15-04-51

core/Plugin/Report.php
mattab: After doing more testing, i noticed sometimes they're removed and sometimes they have a footer message. I reckon that when >= 99% of all visits are not profilable, then we could directly hide all the reports. It makes the UI a lot cleaner. But if it causes difficulties, then it's fine to show all reports and have a  visible warning notification in the footer?

mattab:
> Segments are removed in the UI but nowhere else. We can also disable them and provide a tooltip saying why it doesn't make sense to use them on the current period. But when creating the segment, the period is arbitrary (eg, maybe they want it to be applied to a period w/ profilable data). Seems odd to see it greyed out and realize you have to change the period to be able to use it.

Could we leave the segments available, and not change the Segment Editor UI, but whenever they apply / use a segment that doesn't make in cookie-less (so when they are viewing the reports), then we could display a Warning notification explaining that the segment relies on data that is inaccurate because visits are not profilable?

plugins/API/API.php
diosmosis: showing all segments in API output and when the currently used segment is not profilable, a warning is displayed

core/Plugin/Report.php
diosmosis: showing all reports but when there's no profilable data for the selected period, there is a footer message saying so

core/API/Request.php
diosmosis: will add a comment

you're right, $date is not a good cache key, I'll switch to using the period's start date, except for range periods

plugins/CoreHome/VisitorDetails.php
diosmosis: :+1:

plugins/CoreVisualizations/Visualizations/Sparklines/Config.php
diosmosis: @mattab in the original issue only unique visitors is requested. I'm not sure if there are others we'd also want to remove. Fingerprints could be useful. Can user IDs be specified without a visitor ID?

plugins/Goals/Reports/GetDaysToConversion.php
diosmosis: :+1:

diosmosis:
@tsteur / @mattab this is ready for another review

mattab:
> in the original issue only unique visitors is requested. I'm not sure if there are others we'd also want to remove. Fingerprints could be useful. Can user IDs be specified without a visitor ID?

@diosmosis Here is the full scope we know so far, it is found with some more details in this FAQ: https://matomo.org/faq/general/faq_156/

Here is the full list of reports we need to disable or warn about inaccurate data:

* Unique Visitors and New VS Returning visitors metrics will be inaccurate
* Goals and Ecommerce conversions will be attributed to the channel used in the visit that converts
* Multi Attribution and Cohort reports won’t show data
* A small limited number of other reports will be inaccurate, these are:
  *  Days since last visit
  *  Visits by visit count
  *  Visits to Conversion
  *  Days to Conversion

diosmosis:
@mattab 

> Unique Visitors and New VS Returning visitors metrics will be inaccurate

The whole engagement sparklines widget is disabled so no need to specify _returning specifically.

> Goals and Ecommerce conversions will be attributed to the channel used in the visit that converts

Added a warning to goals & ecommerce pages.

> Multi Attribution and Cohort reports won’t show data

Created PRs for this.

Other reports are marked as requiring profile data.

tsteur:
@mattab was this one reviewed and good to merge once tests pass?

mattab:
Not yet, I will review it as soon as possible (hopefully this week)

tsteur:
what's the latest here @mattab ?

justinvelluppillai:
Any update here @mattab? Might be something we just close if not a priority anymore.

mattab:
@justinvelluppillai yes it is an important change for users, will remove lots of confusion and questions. Why it's important is because more and more people are using Matomo cookie-less (cookieless by default for all users, or if people don't opt-in to cookies in consent screens), then when this is the case, many reports in Matomo show no data. so the experience is quite confusing for people then, and Matomo looks buggy. let's discuss offline and review this work together maybe?

PR #16774, opened 2020-11-23
Update PHP requirement in README

sgiehl:
### Description:

Guess doesn't make sense to show PHP 5.5 as requirement, even though Matomo 3 will be in LTS for a while... 

We also should maybe update this page so the commands at least install php 7.2: https://matomo.org/docs/requirements/#recommended-configuration

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Matomo is released under the GPL v3 (or later) license, see [LICENSE](LICENSE).
 
 ## Requirements
 
-  * PHP 5.5.9 or greater
+  * PHP 7.2.5 or greater
   * MySQL version 5.5 or greater, or MariaDB 
   * PHP extension pdo and pdo_mysql, or the MySQLi extension
   * Matomo is OS / server independent


Findus23:
> We also should maybe update this page so the commands at least install php 7.2: https://matomo.org/docs/requirements/#recommended-configuration

I guess we can generalize them. At least on Debian e.g. php-fpm is always a metapackage that installs the proper version from the repo.

tsteur:
@Findus23 be great to adjust that guide too. Does it work to simply replace `7.0` with eg `8.0` or `7.4`?

Findus23:
@tsteur 

You can simply replace it with 
> $ sudo apt-get install php php-curl php-gd php-cli mysql-server php-mysql php-xml php-mbstring

as all of these packages exist in both Debian and Ubuntu and will install the proper version.

tsteur:
Thanks @Findus23 updated it and also prepared https://github.com/matomo-org/developer-documentation/pull/397

PR #16777, opened 2020-11-24
re-add missing condition for --skip-segments-today

### LINKED ISSUE #16775: --skip-segments-today no longer working in Matomo 4?
If I see in the code correctly this flag doesn't do anything anymore? Can you check @diosmosis ?

It should basically skip archiving a segment in today's timezone (I think unless it was created or changed recently see https://github.com/matomo-org/matomo/blob/3.14.1/core/CronArchive.php#L2000 ) see https://github.com/matomo-org/matomo/blob/3.14.1/plugins/CoreConsole/Commands/CoreArchiver.php#L105-L106

diosmosis:
### Description:

Fixes #16775 

The missing condition was the only difference in the code I could see.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Archive.php
+++ b/core/Archive.php
@@ -234,7 +234,6 @@ class Archive implements ArchiveQuery
             $isMultipleDate);
     }
 
-    // TODO: check period has today, not equals
     public static function shouldSkipArchiveIfSkippingSegmentArchiveForToday(Site $site, Period $period, Segment $segment)
     {
         $now = Date::factory('now', $site->getTimezone());
--- a/core/Archive.php
+++ b/core/Archive.php
@@ -234,6 +234,15 @@ class Archive implements ArchiveQuery
             $isMultipleDate);
     }
 
+    // TODO: check period has today, not equals
+    public static function shouldSkipArchiveIfSkippingSegmentArchiveForToday(Site $site, Period $period, Segment $segment)
+    {
+        $now = Date::factory('now', $site->getTimezone());
+        return $period->getLabel() === 'day'
+            && !$segment->isEmpty()
+            && $period->getDateStart()->toString() === $now->toString();
+    }
+
     /**
      * Queries and returns metric data in an array.
      *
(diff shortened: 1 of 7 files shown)


sgiehl (approved):

sgiehl:
@diosmosis two Integration tests are failing. Guess it's unrelated to this PR. But would be good to fix them....

tsteur:
@diosmosis  what happens if there is an `archive_invalidation` entry for this segment that is for `today`. Then it sends the request to archive. It thinks the segment was archived (but in reality it was skipped), and then it removes the archive_invalidation entry.

Would the segment still be archived the next day after midnight? Or maybe because there's no archive_invalidation entry it wold not be archived in that case? Could you double check this maybe?

diosmosis:
@tsteur I see, will add the check to QueueConsumer as well and test both cases.

tsteur:
@diosmosis let me know when it's pushed and we'll include it in 4.0.2.

diosmosis:
@tsteur finished the test

core/Archive.php
```diff
@@ -234,6 +234,15 @@ public static function factory(Segment $segment, array $periods, array $idSites,
             $isMultipleDate);
     }
 
+    // TODO: check period has today, not equals
```
tsteur: @diosmosis is this a todo that still needs work? Looking at code it works fine as we only look at today anyway.

tsteur:
Tested it quite a few times and worked for me 👍  left one comment regarding a comment in the code. otherwise feel free to merge @diosmosis 🚀

core/Archive.php
diosmosis: It's not needed, I must have been thinking of non-day periods, but for some reason we don't skip those archives... oh that's because before if we skipped day, the logic to launch higher periods wouldn't be executed. This will be a problem now correct? If so it's needed.

tsteur: Before we were still launching higher periods that include today @diosmosis

diosmosis: Ok, then not needed

PR #16779, opened 2020-11-24
Fix seo widget

### LINKED ISSUE #16778: SEO Plugin broken
After upgrading to 4.0.0 the SEO plugin appears to be broken quite badly, no extra php error log is given.

index.php?date=today&module=SEO&action=getRank&disableLink=0&widget=1&idSite=10101&period=day
Call to a member function getRows() on null
in core/DataTable/Renderer.php line 446

sgiehl:
### Description:

fixes #16778

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
@sgiehl is there any test that should have found this error?

sgiehl:
@tsteur the seo widget was actually excluded from UI tests. Guess that was due to the external requests with changing results. I've now added a UI test with fake responses for the requested providers. So hopefully we should see it the next time the widget breaks.

Btw. I've also updated the Alexa metric provider. It requested eg `https://data.alexa.com/data?cli=10&url=matomo.org` to fetch some metrics. But at least for me that URL now always returns `Okay`, so guess it's no longer available. (ping @Findus23)

PR #16780, opened 2020-11-24
Improves Geolocation diagnostics

### LINKED ISSUE #16427: Replace confusing GeoIP warnings
I looked into the GeoIP system check and I am completly confused by both the code and the warnings Matomo gives:

When I set up Matomo with the default geolocation provider, I get the following warning in the system check:
`
Geolocation works, but you are not using one of the recommended providers. If you have to import log files or do something else that requires setting IP addresses, use the PHP GeoIP 2 implementation</a> and install <a href="https://matomo.org/docs/geo-locate/" rel="noreferrer noopener" target="_blank">maxminddb extension.
`
The HTML is broken and it recommends users to install maxminddb, something nearly nobody should need to do.

Looking at the code doesn't really clear anything up:
https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/plugins/UserCountry/Diagnostic/GeolocationDiagnostic.php#L45-L65

The warning I expect most people to see there (`DefaultLocationProviderDesc1`) is only displayed if `!$isProviderInstalled` which only happens if the status of the currently selected geolocation provider is not INSTALLED, but the default provider is obviously always installed, so I think noone ever sees the warning that provides the actual help.

I also can't think of a combination of mixed up translation strings or conditions that would make sense.

@sgiehl, do you know more about how this is supposed to work?

sgiehl:
### Description:

* Introduced a new method `getUsageWarning` that allows triggering custom diagnostic warnings when a location provider is being used. (Currently used for Default and GeoIP2 Server provider)
* Diagnostic now check these states:
  * configured provider is not available   -> Error
  * configured provider is not working    -> Error (with provider error message if available)
  * configured provider is working but has usage warning -> Warning (with provider usage warning)
  * configured provider is working (no usage warning)   -> OK

In addition the errors will now also be shown in geolocation admin is notification.

fixes #16427

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis (approved):
Looks good and works locally. Only thing I noticed is that the server module warning you mentioned in the original ticket is not here. Does this still need to be added?

sgiehl:
@diosmosis switchich to the server module should show this in the system check:

image

PR #16782, opened 2020-11-24
Adds missing field for Campaign ID to campaign builder

### LINKED ISSUE #16233: Campaign URL Builder: Campaign ID field is missing in the app
Campaign ID filed can be found in [https://matomo.org/docs/tracking-campaigns-url-builder/](https://matomo.org/docs/tracking-campaigns-url-builder/) but it is not available when accessing from Matomo and this can be viewed from our [demo](https://demo.matomo.org/index.php?module=CoreHome&action=index&date=today&period=week&idSite=62#?idSite=62&period=week&date=today&segment=&category=Referrers_Referrers&subcategory=Referrers_URLCampaignBuilder) and same can be reproduced in Cloud and On-Premise.

sgiehl:
### Description:

fixes #16233

### Review

* [ ] Functional review done
* [ ] Code review done


plugins/Referrers/lang/en.json
```diff
@@ -18,6 +18,8 @@
         "CampaignGroupHelp": "This parameter is used to track a Campaign Group to identify a group or audience. Campaign groups are helpful for tracking the overall performance of multiple campaigns with similar goals.",
         "CampaignPlacement": "Campaign Placement",
         "CampaignPlacementHelp": "Used to describe the advertising placement or position, for example an entire website, specific pages from a website, an individual ad unit positioned on a single page, a video, a mobile phone app, and more.",
+        "CampaignId": "Campaign ID",
+        "CampaignIdHelp": "This parameter is used to track a Campaign ID: an identifier which can be a number or any string of data up to 100 characters.",
```
diosmosis: I think this might sound better as: "a Campaign ID: a unique identifier for the command consisting of at most 100 numbers, letters or symbols.", what do you think?

diosmosis:
tested locally, everything works. thought there would be a test failure in CampaignBuilder_spec.js, but not seeing one.

tsteur:
@diosmosis the fields only appear when MarketingCampaign plugin is installed, that's why.

plugins/Referrers/lang/en.json
tsteur: Updated 👍

sgiehl: @tsteur that was the text we are using here as well: https://matomo.org/docs/tracking-campaigns-url-builder/
Maybe we should update that one then, too?

tsteur: updated it, page might be cached though for a bit

PR #16783, opened 2020-11-24
Improve scrolling of dialogs

### LINKED ISSUE #14968: Screen to smoothly scroll back up to initial position after closing a popover
Feature request: when opening the Segmented Visit Log, for example, Behaviour → Pages → XXX → Segmented Visit Log, when you click out of it after scrolling down, you have to scroll all the way back up and it takes a while to find your place. Is there anyway for the popup Segmented Visit Log to scroll on its own, without the background scrolling down as well?

## Reproduce

* Open [Segmented Visitor log](https://demo.matomo.org/index.php?module=CoreHome&action=index&date=today&period=week&idSite=62&updated=1#?idSite=62&period=week&date=today&segment=&category=General_Actions&subcategory=General_Pages&popover=RowAction$3ASegmentVisitorLog$3AActions.getPageUrls$3ApageUrl$253D$253Dhttps$25253A$25252F$25252Fdivezone.net$25252Fjobs$3A$257B$2522date$2522$253A$2522today$2522$252C$2522period$2522$253A$2522week$2522$257D)
* Scroll down to bottom of popover
* click outside of popover to close it
* Expected: the screen smoothly scrolls back up to where the top of the popover over
* Got instead: the screen does not move, we are now at the bottom of the page and not clear where we clicked on initially (have to scroll back up sometimes quite a few screens)

Notes:

* When the URL is loaded in a new tab with the popover already opened (like in the reproduce URL above), the app doesn't know which row was clicked so in this case, it could maybe just scroll back to the top of the screen?

sgiehl:
### Description:

Page will be frozen at the current scroll position and only the dialog will scroll (if too high for window)

Currently only tested in Chrome & Firefox on Windows. We need to do some more cross browser testing before merging.

fixes #14968

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
cross browser testing via browserstack/locally:

* on windows, IE 11, scrolling in the dialog works, but outside the modal doesn't seem to scroll (not sure if it's related to this or just broken)

image

* works in windows, edge
* works in mac/chrome
* works in mac/firefox
* works in mac/safari
* works in ios/safari
* doesn't quite work in android/chrome, when I open the dialog it scrolls to near the top for some reason (the original scroll position changes)

diosmosis:
code looks good 👍, there are some screenshot test failures

sgiehl:
I've changed that to use `$(window).scrollTop()` instead of `window.scrollY`, which should fix the behavior for IE11

PR #16784, opened 2020-11-24
Ensure unescaped commas are used for periods in url param

### LINKED ISSUE #15238: going to 'All Websites' dashboard and selecting a date range and clicking a site results in error message
Preface
The system version is 3.13.0, and the update time is 2019.11.26-27. The system has been updated and the server has been migrated

After selecting a date range from all websites dashboard, clicking on a site in the list of sites will report a time period or an incoming time parameter error"error: date format must be: yyyy-mm-dd, or 'today' or 'yesterday' or any keyword supported by the strtotime function (see http://php.net/strtotime for more information):" , Data can only be displayed if the date range is re-selected;

图片1
图片2

sgiehl:
### Description:

Might be a bit hackish solution, but Matomo simply can't handle an encoded date range in the url at various places. So might be the simplest solution to avoid having an encoded value instead of searching for all places that might make problem...
 
fixes #15238

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
I did some digging and I think this can also be fixed by handling %2C or urldecoding in Range::parseDateRange. (The other alternative fix is to urldecode every time 'date' is accessed via Common::getRequestVar, but that seems excessive.)

sgiehl:
@diosmosis  I tried some other stuff as well, but couldn't get it working for all cases. If you have a better solution feel free to simply adjust the PR

diosmosis:
The problem with replacing the comma is that it will only be done client side in this one place, though we of course change the URL in many places, sometimes setting location.href directly, and it could cause problems w/ segments that contain commas given the extra confusing segment encoding logic.

I just checked multisites, but what are the other cases? I can try to fix those as well.

sgiehl:
@diosmosis When we want to "support" an escaped comma in the url parameter we need to be sure, that it works on all pages. IIrc it didn't work for various reports

diosmosis:
Replacing the comma in every part of the URL will have side effects. `%2C` in a URL is the same as `,`, it shouldn't need to be "supported", the real problem is that we don't urldecode query parameters server side implicitly (afaik we do in broadcast, but it's done on the whole URL so I guess might not work great in another portion). If this must be done client side, perhaps instead of replacing it in every parameter (like in the `segment` which could have strange side effects), we could just change it in the date parameter, which is what's causing problems for the referenced issue.

diosmosis:
Found the root cause, it wasn't server side. When the url contains an empty hash, the hash is created using the main url values, but it double encodes them. This was due to `reportingmenu.controller.js` in `makeUrl` using the date from `piwik.getSearchParam()`, which returns the value without decoding, to construct a URL via `$.param` which expects url decoded values. This created a double encoded value which couldn't be parsed.

PR #16797, opened 2020-11-25
use composer/ca-bundle for default CA bundle

Findus23:
As mentioned in https://github.com/matomo-org/matomo/pull/13743#issuecomment-441435508 a long time ago.

This should not change anything as composer/ca-bundle was already an indirect dependency.
### Review

* [ ] Functional review done
* [ ] Code review done


composer.lock
```diff
@@ -4025,5 +3796,5 @@
     "platform-overrides": {
         "php": "7.2.5"
     },
-    "plugin-api-version": "2.0.0"
+    "plugin-api-version": "1.1.0"
```
sgiehl: @tsteur @diosmosis @Findus23 At some point we maybe should decide to use the same composer version 🙈  We are constantly changing this, as well as adding/removing the support and funding sections.

Findus23: Indeed I just pushed the lock file for the tests as it will probably be out of date once this can be merged.

Composer 2 seems to be quite a major change so maybe we should look into it once the Matomo 4 updates have calmed down.

sgiehl:
I've rebased the branch and resolved the conflicts. Should imho be good to merge now

PR #16814, opened 2020-11-26
Update session ids on update

### LINKED ISSUE #16798: Error upgrading db while upgrading to 4.0.0-b1 (session update fails)
see https://forum.matomo.org/t/error-upgrading-db-while-upgrading-to-4-0-0-b1/39478

> * /opt/matomo/core/Updates/4.0.0-b1.php:
> Error trying to execute the migration 'ALTER TABLE `matomo_session` CHANGE `id` `id` VARCHAR(191);'.
> The error was: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'AQBqYXZhLnV0aWwuVHJlZU1h8AEBAW9yZy5hcGFjaGUuY29tbW9ucy5iZWFud...' for key 'PRIMARY'

sgiehl:
### Description:

Instead of updating every single session we could also try `UPDATE session SET id=SHA2(id, 512)`, but that method might be unavailable if mysql is not configured for SSL, causing the update to fail maybe. So guess it's better this way.

fixes #16798

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Updates/4.0.0-b1.php
+++ b/core/Updates/4.0.0-b1.php
@@ -11,6 +11,7 @@ namespace Piwik\Updates;
 
 use Piwik\DataAccess\TableMetadata;
 use Piwik\Date;
+use Piwik\Db;
 use Piwik\DbHelper;
 use Piwik\Plugin\Manager;
 use Piwik\Plugins\CoreHome\Columns\Profilable;
@@ -28,6 +29,7 @@ use Piwik\Common;
 use Piwik\Config;
 use Piwik\Plugins\UserCountry\LocationProvider;
 use Piwik\Plugins\VisitorInterest\Columns\VisitorSecondsSinceLast;
+use Piwik\SettingsPiwik;
 use Piwik\Updater;
 use Piwik\Updates as PiwikUpdates;
 use Piwik\Updater\Migration\Factory as MigrationFactory;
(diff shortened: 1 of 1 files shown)


core/Updates/4.0.0-b1.php
```diff
@@ -104,6 +106,19 @@ public function getMigrations(Updater $updater)
...
         $migrations[] = $this->migration->db->dropColumn('user', 'token_auth');
 
+        // remove outdated sessions to avoid useless updates
+        Db::exec('DELETE FROM ' . Common::prefixTable('session') . ' WHERE modified + lifetime < ' . time());
+
+        $salt = SettingsPiwik::getSalt();
+        $sessions = Db::fetchAll('SELECT id from ' . Common::prefixTable('session'));
+
+        foreach ($sessions as $session) {
+            $migrations[] = $this->migration->db->boundSql(
```
tsteur, line 116: this would actually print sessionId hashes on the screen... of course they are hashed but it may still not be ideal. Matomo may be in that state for a while and people could see them ... it's very low risk though that it 1) becomes visible and 2) someone either has the salt or tries to brute force them (eg a sys admin who is updating Matomo could have the salt but no access to the DB and now could technically brute force the original session id) . If someone was to downgrade later to Matomo 3 then it become an actual sessionId though. 

image

I wonder should we maybe only do this for session Ids that are > 190 characters? In all other cases we could simply log out users? and these we could even simply delete?

Or we simply execute the query in the background similar to the delete from session.

tsteur: in any case we should not do this for any 128 character session as this indicates the sessionId might be already hashed

core/Updates/4.0.0-b1.php
```diff
@@ -252,6 +257,16 @@ public function getMigrations(Updater $updater)
 
     public function doUpdate(Updater $updater)
     {
+        $salt = SettingsPiwik::getSalt();
```
tsteur, line 260: we only execute it as part of doUpdate... which should be fine

tsteur:
Note: this test fails but also does in 4.x https://travis-ci.com/github/matomo-org/matomo/jobs/449214044#L1438

PR #16816, opened 2020-11-26
Fix hash algo is missing during upgrade

tsteur:


core/Updates/4.0.0-b1.php
```diff
@@ -80,7 +80,8 @@ public function getMigrations(Updater $updater)
                     'login' => $user['login'],
                     'description' => 'Created by Matomo 4 migration',
                     'password' => $userModel->hashTokenAuth($user['token_auth']),
-                    'date_created' => Date::now()->getDatetime()
+                    'date_created' => Date::now()->getDatetime(),
+                    'hash_algo' => 'sha512'
```
tsteur, line 84: I'll also add an update for existing migrations

diosmosis (approved):

PR #16819, opened 2020-11-27
Merging 4.x-dev into master

mattab:
this is one of the steps in https://github.com/matomo-org/matomo/issues/14601


mattab:
Wrong PR, needs to be the other way around

PR #16822, opened 2020-11-27
Improve token auth migration

tsteur:
### Description:

as discussed. will basically look like this:

image

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Updates/4.0.0-b1.php
+++ b/core/Updates/4.0.0-b1.php
@@ -246,9 +246,6 @@ class Updates_4_0_0_b1 extends PiwikUpdates
 
         $updater->executeMigrations(__FILE__, $this->getMigrations($updater));
 
-        // just in case it was not executed manually by a user...
-        MigrateTokenAuths::migrate();
-
         if ($this->usesGeoIpLegacyLocationProvider()) {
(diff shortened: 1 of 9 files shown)


core/Updates/4.0.0-b1.php
```diff
@@ -269,6 +246,9 @@ public function doUpdate(Updater $updater)
 
         $updater->executeMigrations(__FILE__, $this->getMigrations($updater));
 
+        // just in case it was not executed manually by a user...
+        MigrateTokenAuths::migrate();
```
diosmosis: Is this necessary? Since the migration is added above, if the user clicks 'Update', it will definitely run right? If so, this would be redundant. Or is there some way for the user to opt out of that specific migration running?

tsteur: true, it's not needed .will remove it now

diosmosis:
Noticed the core:matomo4-migrate-token-auths command doesn't provide any output, maybe a 'Done.' message would be useful to have?

diosmosis (approved):
left one other comment, otherwise looks good

tsteur:
Updated it 👍

PR #16826, opened 2020-11-27
Adjust archive_invalidations index

### LINKED ISSUE #16811: Matomo 4 migration: Index column size too large
reported in https://forum.matomo.org/t/fehler-bei-installation-in-strato-hosting-paket/39480

> SQLSTATE[HY000]: General error: 1709 Index column size too large. The maximum column size is 767 bytes.

PHP-Version 7.2 and MySQL-Version 5.6.

sgiehl:
### Description:

The index actually differs from the one defined in the update script: 
https://github.com/matomo-org/matomo/blob/1fa10c628c751d43b346a909e68833d26d5c08ff/core/Updates/4.0.0-b1.php#L105

Not sure if the index on `name` was on purpose. If so we need to limit the field or the index

fixes #16811

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Db/Schema/Mysql.php
+++ b/core/Db/Schema/Mysql.php
@@ -322,7 +322,7 @@ class Mysql implements SchemaInterface
                                             status TINYINT(1) UNSIGNED DEFAULT 0,
                                             `report` VARCHAR(255) NULL,
                                             PRIMARY KEY(idinvalidation),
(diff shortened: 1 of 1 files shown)


tsteur:
@diosmosis which one is the correct index? the one in the update script or the schema?

diosmosis:
I don't think we want an index on name, not sure why that was in there.

PR #16830, opened 2020-11-27
Allow using last (week|month|year) as date param

### LINKED ISSUE #13818: Support date format last week, last month, last year
https://demo.matomo.org/index.php?module=MultiSites&action=index&idSite=3&period=day&date=last%20week

If you open this link, the datepicker is not present at all, but last week is working as intended.

On dashboard the datepicker is present.

sgiehl:
### Description:

Using one those values actually won't directly pass them to strtotime, as it might have imho unexpected results. `last week` passed to `strtotime` for example always returns the monday of the previous week. Instead I guess it would be expected to result in the day 7 days ago. Similar applies to last month and year. 

For last month there are still some edge cases on the last day of a month when the previous month didn't have that much days.

E.g. `last month` on 31th Dezember would return 1st Dezember

fixes #13818

### Review

* [ ] Functional review done
* [x] Code review done


tests/PHPUnit/Unit/DateTest.php
```diff
@@ -72,6 +72,39 @@ public function testYesterday()
...
+    /**
+     * create last year object check that timestamp is correct (midnight)
+     *
+     * @group Core
+     */
+    public function testLastYear()
+    {
+        $date = Date::lastYear();
+        $this->assertEquals(strtotime(date("Y-m-d", strtotime('-1 year')) . " 00:00:00"), $date->getTimestamp());
+    }
```
diosmosis, line 117: I think you could test these more effectively by using `Date::$now`. Right now it looks like you're mostly duplicating the function logic in the tests?

sgiehl: @diosmosis  I've improved the date tests (and also fixed `yesterdaySameTime`, that actually wasn't respecting the now timestamp for the returned time)

diosmosis:
@mattab / @tsteur would we need to update documentation anywhere for this?

diosmosis:
Should these be usable in the API? If so, would it be good to have a system test or two for them?

diosmosis:
Tested locally, works well, tested comparison + export links. Left one other test related comment, otherwise lgtm

tsteur:
Yes be great to document here if it's supported in the API https://developer.matomo.org/api-reference/reporting-api

sgiehl:
@diosmosis it actually wasn't possible to use those as api params. at least the results were incorrect.
I've added a new commit to make that possible. It also fixes the usage when period=range.
Also added some system tests.

tests/PHPUnit/System/SpecialDateParamsTest.php
```diff
@@ -0,0 +1,78 @@
...
+                ],
+            ],
+            [
+                ['VisitsSummary.getVisits', 'Actions.getPageTitles'],
+                [
+                    'idSite'       => self::$fixture->idSite,
+                    'date'         => 'last year,last week',
+                    'periods'      => ['range'],
+                    'testSuffix'   => '_range',
+                ],
```
diosmosis, line 67: Just for completeness, maybe we could change a couple of these dates to `last-week/lastYear`?

diosmosis (approved):
Left one more minor comment, otherwise LGTM

tests/PHPUnit/System/SpecialDateParamsTest.php
sgiehl: done

PR #16835, opened 2020-11-29
Change referrer url size to 1500

tsteur:
fix https://github.com/matomo-org/matomo/issues/16810

Avoids running into row size exceeded in some cases.

### Description:

Please include a description of this change and which issue it fixes. If no issue exists yet please include context and what problem it solves.

### Review

* [ ] Functional review done
* [ ] Usability review done (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] Security review done [see checklist](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] Code review done
* [ ] Tests were added if useful/possible
* [ ] Reviewed for breaking changes
* [ ] Developer changelog updated if needed
* [ ] Documentation added if needed
* [ ] Existing documentation updated if needed

PR #16836, opened 2020-11-29
Trim referrer url if needed

tsteur:
refs https://github.com/matomo-org/matomo/pull/16835#issuecomment-735466614

fyi @sgiehl

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Referrers/Columns/ReferrerUrl.php
+++ b/plugins/Referrers/Columns/ReferrerUrl.php
@@ -15,14 +15,19 @@ use Piwik\Tracker\Action;
 
 class ReferrerUrl extends Base
 {
+    const MAX_LEN = 1500;
     protected $columnName = 'referer_url';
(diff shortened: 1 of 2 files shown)


mattab:
fyi added to https://matomo.org/docs/data-limits/
Screenshot from 2020-11-30 11-53-34

sgiehl:
@mattab 1500 bytes is actually not correct I guess. It's a VARCHAR(1500) so there are 1500 characters allowed. Depending on the collation one character can have up to 3 or 4 bytes. Might be better to write characters instead maybe

mattab:
@sgiehl sounds good! updated to this:
Screenshot from 2020-12-07 09-47-05

tsteur:
@mattab as mentioned in slack by default the referrer page url is also limited to 1024 characters but can be increased to 1500 through the same setting.

PR #16837, opened 2020-11-30
More rigorous checks for non-existant site so archiving wont fail if …

diosmosis:
…a site is deleted + tweaks to some debug logs

### Description:

Deleted some sites while a canceled/failed archive had been run before, and the next core:archive run would keep failing since the sites in the list didn't exist. Added some more checks for whether the site exists or not and fixed the existing check method.

FYI @tsteur 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/CronArchive.php
+++ b/core/CronArchive.php
@@ -26,6 +26,7 @@ use Piwik\DataAccess\ArchiveSelector;
 use Piwik\DataAccess\ArchiveTableCreator;
 use Piwik\DataAccess\Model;
 use Piwik\DataAccess\RawLogDao;
+use Piwik\Exception\UnexpectedWebsiteFoundException;
 use Piwik\Metrics\Formatter;
 use Piwik\Period\Factory;
 use Piwik\Period\Factory as PeriodFactory;
@@ -417,6 +418,12 @@ class CronArchive
             }
 
             $idSite = $archive['idsite'];
+            if (!$this->siteExists($idSite)) {
+                $this->logger->debug("Site $idSite no longer exists, no longer launching archiving.");
+                $this->deleteInvalidatedArchives($archive);
+                continue;
+            }
+
             $dateStr = $archive['period'] == Range::PERIOD_ID ? ($archive['date1'] . ',' . $archive['date2']) : $archive['date1'];
             $period = PeriodFactory::build($this->periodIdsToLabels[$archive['period']], $dateStr);
             $params = new Parameters(new Site($idSite), $period, new Segment($segment, [$idSite], $period->getDateStart(), $period->getDateEnd()));
@@ -760,6 +767,12 @@ class CronArchive
 
     public function invalidateArchivedReportsForSitesThatNeedToBeArchivedAgain($idSiteToInvalidate)
     {
+        if (!$this->siteExists($idSiteToInvalidate)) {
+            $this->invalidator->forgetRememberedArchivedReportsToInvalidateForSite($idSiteToInvalidate);
+            $this->logger->info("Site $idSiteToInvalidate no longer exists, skipping invalidation.");
+            return;
+        }
+
         if ($this->model->isInvalidationsScheduledForSite($idSiteToInvalidate)) {
             $this->logger->debug("Invalidations currently exist for idSite $idSiteToInvalidate, skipping invalidating for now...");
             return;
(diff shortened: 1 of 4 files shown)


tests/PHPUnit/Integration/CronArchive/QueueConsumerTest.php
```diff
@@ -354,6 +355,70 @@ public function test_invalidateConsumeOrder()
...
+                // empty
+            ],
+        ];
+
+        try {
+            $this->assertEquals($expectedInvalidationsFound, $iteratedInvalidations);
+        } catch (\Exception $ex) {
+            print "\nInvalidations inserted:\n" . var_export($invalidations, true) . "\n";
+            throw $ex;
+        }
```
sgiehl: Why should assertEquals throw an exception? Shouldn't that be more like
```suggestion
        $this->assertEquals($expectedInvalidationsFound, $iteratedInvalidations, "Invalidations inserted:\n" . var_export($invalidations, true));
```

tests/PHPUnit/Integration/CronArchive/QueueConsumerTest.php
```diff
@@ -354,6 +355,70 @@ public function test_invalidateConsumeOrder()
...
+
+        $iteratedInvalidations = [];
+        while (true) {
+            $next = $queueConsumer->getNextArchivesToProcess();
+            if ($next === null) {
+                break;
+            }
+
+            foreach ($next as &$item) {
+                Db::query("UPDATE " . Common::prefixTable('archive_invalidations') . " SET status = 1 WHERE idinvalidation = ?", [$item['idinvalidation']]);
```
sgiehl: Not sure what that update is good for. Is it so `getNextArchivesToProcess` would not end up in a loop if any record is returned? 🤔

tests/PHPUnit/Integration/CronArchive/QueueConsumerTest.php
```diff
@@ -560,4 +625,9 @@ protected static function configureFixture($fixture)
         parent::configureFixture($fixture);
         $fixture->createSuperUser = true;
     }
+
+    private function getInvalidationCount()
+    {
+        return Db::fetchOne("SELECT COUNT(*) FROM " . Common::prefixTable('archive_invalidations'));
```
sgiehl: That method actually isn't used anywhere 🤔

sgiehl (changes requested):

tests/PHPUnit/Integration/CronArchive/QueueConsumerTest.php
diosmosis: As long as the message (which will be very long) is readable and only displayed on failure, that would work too.

tests/PHPUnit/Integration/CronArchive/QueueConsumerTest.php
diosmosis: QueueConsumer only iterates through invalidations, it does not start archives, CronArchive does this as it launches new processes. This is to simulate this for the test, since QueueConsumer is only part of the whole loop.

core/CronArchive.php
```diff
@@ -1263,4 +1276,14 @@ public function checkNoDanglingInvalidations(array $idInvalidations)
             ]);
         }
     }
+
+    private function siteExists($idSite)
+    {
+        try {
+            $site = APISitesManager::getInstance()->getSiteFromId($idSite);
```
tsteur: btw could use `new Site($idSite)` here directly as it would use cache but maybe that be actually the problem and we don't want to hit the cache.

tsteur: actually be great if this hit the cache maybe? or would it cause a problem if the site was suddenly removed?

tsteur:
Looks generally good once tests pass @diosmosis  👍  there are a few comments to look at and a merge conflict but nothing major.

core/CronArchive.php
diosmosis: The cache would exist for the entire core:archive run, so a site could be removed say 30 mins after archiving starts, but it's data could still be cached, and we'd try to process it. We could maybe clear the cache before starting a new site archive so it's cached a bit... I'll think about it.

diosmosis:
@tsteur @sgiehl found a simpler solution that will avoid extra queries

PR #16840, opened 2020-11-30
Show an error notification in UI when given date/period combination is invalid

### LINKED ISSUE #16342: Error Call to undefined method Piwik\\DataTable\\Simple::getComparisons()
On request `index.php?forceView=1&viewDataTable=sparklines&module=Goals&action=get&idGoal=10&allow_multiple=0&only_summary=1&idSite=2&period=day&date=2020-08-17,2020-08-23&segment=&showtitle=1&random=8740`

following error happened because a data table map was returned instead of a regular data table. And in this case the first row is a dataTable and not a row.

> Error: {"message":"Call to undefined method Piwik\\DataTable\\Simple::getComparisons()","file":"\/plugins\/CoreVisualizations\/Visualizations\/Sparklines.php","line":125,"request_id":"12345","backtrace":" on \/plugins\/CoreVisualizations\/Visualizations\/Sparklines.php(125)\n#0 \/plugins\/CoreVisualizations\/Visualizations\/Sparklines.php(94): Piwik\\Plugins\\CoreVisualizations\\Visualizations\\Sparklines->fetchConfiguredSparklines()\n#1 \/core\/Plugin\/Report.php(320): Piwik\\Plugins\\CoreVisualizations\\Visualizations\\Sparklines->render()\n#2 \/plugins\/CoreHome\/Controller.php(58): Piwik\\Plugin\\Report->render()\n#3 [internal function]: Piwik\\Plugins\\CoreHome\\Controller->renderReportWidget(Object(Piwik\\Plugins\\Goals\\Reports\\Get))\n#4 \/core\/FrontController.php(590): call_user_func_array(Array, Array)\n#5 \/core\/FrontController.php(165): Piwik\\FrontController->doDispatch('Goals', 'get', Array)\n#6 \/core\/dispatch.php(34): Piwik\\FrontController->dispatch()\n#7 \/index.php(27): require_once('\/c...')\n#8 {main}"}

sgiehl:
### Description:

fixes #16342

### Review

* [ ] Functional review done
* [ ] Code review done


plugins/CoreHome/javascripts/broadcast.js
```diff
@@ -784,10 +784,10 @@ var broadcast = {
 
             // try looking for multi value param
             lookFor = param + '[]=';
-            startStr = url.indexOf(lookFor);
+            startStr = url.lastIndexOf(lookFor);
             if (startStr >= 0) {
                 var result = [getSingleValue(startStr)];
-                while ((startStr = url.indexOf(lookFor, startStr + 1)) !== -1) {
+                while ((startStr = url.lastIndexOf(lookFor, startStr + 1)) !== -1) {
```
diosmosis: Would this be an issue if we have an array param like `&key[]=val1&key[]=val2&key[]=val3` where the new array value has reversed order? Eg, w/ lastIndexOf, I would expect it to be `['val3','val2','val1']`, but callers should expect it to be the other way around.

diosmosis (changes requested):
left one comment, otherwise code looks good and seems to work locally

I'm assuming the content screenshot changes are due to the getSearchParam change, though I couldn't figure out where exactly it would use the improper date before.

plugins/CoreHome/javascripts/broadcast.js
sgiehl: I'll revert that, when searching for multiple values

diosmosis (approved):

diosmosis:
@sgiehl some angularjs tests are failing

sgiehl:
@diosmosis there was a problem with `broadcast.getParamValue()`.
That method actually never worked correctly for some cases. It was searching the url for `param + '='` or `param + '[]='`, which was actually not correct in some cases.

As soon as the request param name was included in another param the method actually returned the wrong value.
Let's assume a url like `?myparam=1&param=2`. Calling `broadcast.getParamValue('param')` would actually have returned `1` instead of `2`. That behavior actually caused the test failures as `subcategory` contains `category` and so the wrong value was returned.

I've changed the method so it hopefully should work correct now. Let's see if all tests are still passing...

plugins/CoreHome/angularjs/comparisons/comparisons.service.spec.js
```diff
@@ -260,6 +261,7 @@
             it('should return false if comparison is disabled for the page', function () {
                 $location.search('category=MyModule2&subcategory=disabledPage2&date=2018-01-02&period=day&segment=&compareDates[]=2018-03-04&comparePeriods[]=week&compareSegments[]=comparedsegment');
                 $rootScope.$apply();
+                console.log($window.location.href);
```
diosmosis: fyi, debug code

diosmosis:
@sgiehl given the method was buggy, do you think it might help to have angular unit tests for this method in broadcast?

sgiehl:
@diosmosis added some simple tests

diosmosis:
@sgiehl one screenshot test failure needs to be looked at: https://builds-artifacts.matomo.org/matomo-org/matomo/invaliddate/44416/PeriodSelector_invalid.png then it's good to merge i'd say

sgiehl:
@diosmosis the test should pass now

diosmosis:
@sgiehl there appear to be two more comparison related test failures

sgiehl:
@diosmosis ok. hope this time everything works again 🙈

PR #16841, opened 2020-12-01
Fix unknown keyword is not shown in transitions report in seach engines section

tsteur:
### Description:

Transitions report currently doesn't show the Unknown keyword and therefore the percentages for search keywords are completely wrong. Basically if there are 700 searches, and 5 times the keywords were detected from the search, then it calculates the percentage based on the 5 detected keywords. If 5 keywords were used once, then each keyword gets shown 20% and to the user it looks like 20% of 700 searches.

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
I guess this isn't tested by an existing system test, should we add one?

tsteur:
@diosmosis no tests failed so it's likely not covered. We could add a test but am quite busy and don't really have time to add one and figured at least it'll be fixed. Better than not having it fixed. I reckon ideally there was a test but it's also not crazy important.

diosmosis:
@tsteur this is in the 4.1.0 milestone so I think someone could add it later on if there was time

PR #16844, opened 2020-12-01
only select invalidations that were added before a sites archiving began

diosmosis:
### Description:

To avoid potential delays in finishing archiving we only handle invalidations that were added before archiving for a specific site begins.

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl:
@diosmosis there's a test failing. Otherwise the codes looks imho good.

PR #16849, opened 2020-12-01
Don't accept files that are bigger than the upload limit when uploading plugins

### LINKED ISSUE #16847: The parameter ‘confirmPassword’ isn’t set in the Request
reported in https://forum.matomo.org/t/warning-message-on-new-installation/39557/3 (futher down)

> I do however have an issue trying to install the google analytics plugin as i get this server error either trying to access the marketplace or import a zip file. Any ideas on that one? From what I read in the other article, it seems the issues may have been related?
>
> ( 500 - Internal server error. ### There is a problem with the resource you are looking for, and it cannot be displayed.)

`ERROR Piwik\ExceptionHandler[2020-12-01 11:04:09 UTC] [ffda2] Uncaught exception: C:\inetpub\wwwroot\matomo\core\Common.php(535): The parameter ‘confirmPassword’ isn’t set in the Request, and a default value wasn’t provided.`
`ERROR Piwik\ExceptionHandler[2020-12-01 11:04:09 UTC] [ffda2] The parameter ‘confirmPassword’ isn’t set in the Request, and a default value wasn’t provided.`

https://github.com/matomo-org/matomo/blob/f4dd5624a1f56602ea449668a40e00b70f0e28f4/core/Common.php#L535
This is most likely called from 
https://github.com/matomo-org/matomo/blob/7585967cbd80d7869f57a9adc69c02e8d14cbc33/plugins/CorePluginsAdmin/Controller.php#L120

sgiehl:
### Description:

fixes #16847 

### Review

* [x] Functional review done
* [x] Code review done


diosmosis:
Noticed if a strange value is set in php.ini, it is used as is and shows up in data-max-size. Eg, if I use "8     M", data-max-size is `"8     "`. Setting it to `blahblah` or `8blahblahM` sets data-max-size to "" which I think is correct, so it's just if there's a bunch of spaces and no math is done I guess.

core/SettingsServer.php
```diff
@@ -219,6 +219,35 @@ public static function getMemoryLimitValue()
...
+                case 'K':
+                case 'k':
+                    return substr($memory, 0, -1) / 1024;
+            }
+            return $memory / 1048576;
+        }
+
+        // no max upload size
+        return false;
+    }
```
diosmosis, line 221: Would it be possible and a good idea to have some unit tests for this code?

diosmosis (approved):
Works and code looks good, just left a couple of minor comments

sgiehl:
@diosmosis I've improved the code a bit more and added a simple test

Issue #16865, opened 2020-12-02
Add new system check: Check all expected table & columns a plugin expects to exist actually exist


narfk:
with the default Tracking Code (Backend) i got this error with the new 4.0.0 Version.
In 3.14.1 this works fine.

```
<script type="text/javascript">
  var _paq = window._paq = window._paq || [];
  _paq.push(['trackPageView']);
  _paq.push(['enableLinkTracking']);
  (function() {
    var u="https://statistics.domain.dd/piwik/";
    _paq.push(['setTrackerUrl', u+'matomo.php']);
    _paq.push(['setSiteId', '2']);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.type='text/javascript'; g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
  })();
</script>
```

Header
Status400
Bad Request
VersionHTTP/1.1
Transferred1.13 KB (410 B size)
Referrer Policystrict-origin-when-cross-origin

While is was inspecting matomo.js i found setRequestMethod.
So i tried this.

```
<script type="text/javascript">
  var _paq = window._paq = window._paq || [];
  _paq.push(['setRequestMethod', 'GET']);
  _paq.push(['trackPageView']);
  _paq.push(['enableLinkTracking']);
  (function() {
    var u="https://statistics.domain.dd/piwik/";
    _paq.push(['setTrackerUrl', u+'matomo.php']);
    _paq.push(['setSiteId', '2']);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.type='text/javascript'; g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
  })();
</script>
```

without any results.
Its still a POST and i got a bad request.

The URL in the networkinspector looks weird for a POST
```https://statistics.domain.dd/piwik/matomo.phhp?action_name=title&idsite=2&rec=1&r=944278&h=14&m=29&s=34&url=...```

![example](https://user-images.githubusercontent.com/38341745/100882406-74e4b980-34af-11eb-8f70-faee6b131831.png)

Any advices?

**Comment by sgiehl** [comment]
I don't think that has something to do with the POST request method. Matomo tries to use `sendBeacon` if that is available, which uses POST by default. 

Maybe try enabling the debugging for tracking by setting
```
[Tracker]
debug = 1
```
in your config.ini.php

[...]

PR #16870, opened 2020-12-02
When forcing GET request method in JS tracker then disable send beacon

tsteur:
### Description:

refs https://github.com/matomo-org/matomo/issues/16865

Will update the docs once merged.

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
build js

js/piwik.js
```diff
@@ -4827,6 +4827,9 @@ if (typeof window.Matomo !== 'object') {
             this.getContent = function () {
                 return content;
             };
+            this.isUsingAlwaysUseSendBeacon = function () {
```
sgiehl, line 4830: Maybe `isAlwaysUsingSendBeacon` would be a better name.

sgiehl (approved):

js/piwik.js
tsteur: it's only an internal name and will be removed in the release

PR #16871, opened 2020-12-02
Use existing Request method w/ correct logic for limiting widgetize access.

### LINKED ISSUE #16869: Widget access always only works with view access not respecting new config setting
I think https://github.com/matomo-org/matomo/blob/4.0.3/plugins/Widgetize/Controller.php#L36-L39 is not respecting the `enable_framed_allow_write_admin_token_auth` setting @diosmosis ?

It should be maybe also using https://github.com/matomo-org/matomo/blob/4.0.3/core/API/Request.php#L461 when token_auth is not empty?

Technically, I think that code is not even needed in Widgetized controller as it's already done in Frontcontroller but it be generally maybe still good to simply call that existing method also (unless there's some reason not to call that method). Just to be sure it'll be executed in widgetized action no matter how it is executed.

I didn't actually test it. Only saw the code and it looks like this will be causing issues.

diosmosis:
### Description:

FYI @tsteur 

Fixes #16869 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Widgetize/Controller.php
+++ b/plugins/Widgetize/Controller.php
@@ -9,6 +9,7 @@
 namespace Piwik\Plugins\Widgetize;
 
 use Piwik\Access;
+use Piwik\API\Request;
 use Piwik\Common;
 use Piwik\Container\StaticContainer;
 use Piwik\FrontController;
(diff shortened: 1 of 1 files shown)


tsteur:
Works 👍

PR #16875, opened 2020-12-03
Do not force api session usage for widgetized views

### LINKED ISSUE #16867: Displaying widgets in iframes logs bruteforce attacks
I use iframes to render some of the widgets in a website located on another server but on the same local network. Matomo's host and this website's are also subdomains of the same domain.

I added the iframes like suggested on the `index.php?module=Widgetize&action=index` page, with some little adjustments and using the token of a user who has the right to view the website:

```
<!-- Real time visits-->
<div class="widgetIframe" id="realTimeVisits"><iframe width="450" height="320" src="https://stats.mydomain.com/index.php?module=Widgetize&action=iframe&disableLink=1&widget=1&moduleToWidgetize=Live&actionToWidgetize=widget&idSite={matomoSiteId}&period=month&date=today&token_auth={matomoToken}" scrolling="yes" frameborder="0" marginheight="0" marginwidth="0"></iframe></div>

<!-- Last visits graph-->
<div class="widgetIframe" id="lastVisitsGraph"><iframe width="450" height="260" src="https://stats.mydomain.com/index.php?module=Widgetize&action=iframe&disableLink=1&widget=1&moduleToWidgetize=VisitsSummary&actionToWidgetize=getEvolutionGraph&idSite={matomoSiteId}&period=day&date=today&token_auth={matomoToken}" scrolling="yes" frameborder="0" marginheight="0" marginwidth="0"></iframe></div>
```

The widgets do appear as expected, however everytime the second iframe is rendered (`getEvolutionGraph`), a new entry is added to the `matomo_brute_force_log` table.

I also had this very same problem with the widget `moduleToWidgetize=VisitsSummary&actionToWidgetize=get`, but only when the parameters `forceView=1&viewDataTable=VisitorLog` where added to the url; if those parameters are removed, the problem disappear. Of course, removing these parameters do not resolve the getEvolutionGraph case...

Because of  this problem, the user ip will be blocked for no reason if he visit the stats page more than the X times allowed by the brute force settings, making these widgets impossible to use in production.

sgiehl:
### Description:

refs #16867

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CoreHome/angularjs/common/services/piwik-api.js
+++ b/plugins/CoreHome/angularjs/common/services/piwik-api.js
@@ -50,7 +50,7 @@ var hasBlockedContent = false;
         {
             postParams['token_auth'] = piwik.token_auth;
             // When viewing a widgetized report there won't be any session that can be used, so don't force session usage
-            postParams['force_api_session'] = (piwik.broadcast.getValueFromUrl('module') !== 'Widgetize') ? 1 : 0;
+            postParams['force_api_session'] = piwik.broadcast.isWidgetizeRequestWithoutSession() ? 0 : 1;
         }
 
         function isRequestToApiMethod() {
@@ -194,7 +194,7 @@ var hasBlockedContent = false;
             if (isRequestToApiMethod() || piwik.shouldPropagateTokenAuth) {
                 params.token_auth = piwik.token_auth;
                 // When viewing a widgetized report there won't be any session that can be used, so don't force session usage
-                params.force_api_session = (piwik.broadcast.getValueFromUrl('module') !== 'Widgetize') ? 1 : 0;
+                params.force_api_session = piwik.broadcast.isWidgetizeRequestWithoutSession() ? 0 : 1;
             }
 
             return params;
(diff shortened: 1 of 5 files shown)


plugins/CoreHome/angularjs/common/services/piwik-api.js
```diff
@@ -49,7 +49,8 @@ var hasBlockedContent = false;
         function withTokenInUrl()
         {
             postParams['token_auth'] = piwik.token_auth;
-            postParams['force_api_session'] = '1';
+            // When viewing a widgetized report there won't be any session that can be used, so don't force session usage
+            postParams['force_api_session'] = (piwik.broadcast.getValueFromUrl('module') !== 'Widgetize') ? 1 : 0;
```
tsteur: @sgiehl wouldn't we need to look if `force_api_session` is present in the original URL or not or something? Like it would mean you can't use the widgetized api with the session API token. Not sure if we do use this somewhere or not in the UI or somewhere?

sgiehl: Guess makes sense to check that. will change that

sgiehl: @tsteur updated the code so it respects if force_api_session was already set. I'm not able to test the changes right now. Feel free to do that if needed otherwise I'll do that tomorrow

tsteur:
Worked for me generally. However, the visitor map UI tests are now failing. And when I'm on the widgetized page then I always get an error "failed to login"

image
image

I'll push a fix for this and will quickly look into visitor map issue.

Generally I think there might be still an issue though in general if someone was to embed any non widget action using a regular token auth?

tsteur:
I'm expecting the tests to pass and will likely merge to have it fixed for widgets. I'll add a comment to the original issue to next check it what happens when embedding other actions directly if there is any issue (such as Goals.manageGoals) or the entire UI etc

PR #16878, opened 2020-12-03
Adds missing report documentations

### LINKED ISSUE #16792: Have a report description for every report
We should click through the UI and should make sure to have a report description for every report shown in the UI. This is currently for example not the case for Device reports and some other reports
image

sgiehl:
### Description:

This also fixes an issue where the report documentation was not visible after switching to a related report.

fixes #16792

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl:
Would be good if someone else could look through the documentations I've added and maybe improve them further. Didn't spent too much time on thinking of good descriptions.

Also some of the dynamically created reports currently don't have a documentation, like the custom dimension reports. Not sure if we should add one there as well.

And we need to check all our plugins as well maybe

diosmosis:
@mattab maybe we could have Daniel, the person rewriting our technical guides, to look over these docs later? For now we could merge after checking for small errors, then they could look through the UI and make tweaks to the report documentation w/o having to look through code. What do you think?

plugins/Goals/lang/en.json
```diff
@@ -108,6 +109,7 @@
         "GoalsAdjective": "Goals %s",
         "VisitPageTitle": "Visit a given Page Title",
         "VisitsUntilConv": "Visits to Conversion",
+        "VisitsUntilConvReportDocumentation": "This reports show how many visits a visitor need to convert a goal.",
```
Findus23:
```suggestion
        "VisitsUntilConvReportDocumentation": "This reports show how many visits a visitor needs to convert a goal.",
```

Findus23 (commented):
just some simple spellchecking

diosmosis:
I think spending time making the report docs perfect here is probably time ill-spent, so I'll do some quick checks for typos and create a new issue to revise the report docs in the future. If that must be done in 4.1, it can be prioritized there.

tsteur:
Be great to do it as part of 4.1 👍

diosmosis:
@sgiehl I made some revisions to some of the docs, can you check if they sound better to you? Feel free to revert any changes I made.

sgiehl:
@diosmosis the changes look good. If noone else does one to take a look, it might be good to merge.

PR #16880, opened 2020-12-03
in cache buster only look at activated plugins, not loaded plugins

### LINKED ISSUE #16672: Manage Containers page not found
I host matomo 3.14.1 on my own server, following the [installation docs](https://matomo.org/docs/installation/) and everything works fine except the **Manage Containers** page in **Tag Manager** can not be load, as you can see below:

image

Matomo helps a lot and I've bought two plugins, Heatmaps and User flows, from the marketplace. And I also need to use tag manager, please help!

tsteur:
### Description:

This might fix https://github.com/matomo-org/matomo/issues/16672 . We could see if it then happens again. There's probably not much point in writing a test for it because it wouldn't really test the real scenario I suppose. Generally the thought is cache buster should only include active plugins, not loaded plugins. Maybe tag manager was loaded for this user but not activated.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/AssetManager/UIAssetCacheBuster.php
+++ b/core/AssetManager/UIAssetCacheBuster.php
@@ -33,14 +33,17 @@ class UIAssetCacheBuster extends Singleton
 
             $masterFile     = PIWIK_INCLUDE_PATH . '/.git/refs/heads/master';
             $currentGitHash = file_exists($masterFile) ? @file_get_contents($masterFile) : null;
+            $manager = Manager::getInstance();
 
-            $plugins = !$pluginNames ? Manager::getInstance()->getLoadedPluginsName() : $pluginNames;
+            $plugins = !$pluginNames ? $manager->getActivatedPlugins() : $pluginNames;
             sort($plugins);
 
             $pluginsInfo = '';
             foreach ($plugins as $pluginName) {
-                $plugin      = Manager::getInstance()->getLoadedPlugin($pluginName);
-                $pluginsInfo .= $plugin->getPluginName() . $plugin->getVersion() . ',';
+                if ($manager->isPluginLoaded($pluginName)) {
+                    $plugin      = $manager->getLoadedPlugin($pluginName);
+                    $pluginsInfo .= $plugin->getPluginName() . $plugin->getVersion() . ',';
+                }
             }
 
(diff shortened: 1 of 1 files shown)


diosmosis (approved):

core/AssetManager/UIAssetCacheBuster.php
```diff
@@ -33,14 +33,17 @@ public function piwikVersionBasedCacheBuster($pluginNames = false)
...
 
-            $plugins = !$pluginNames ? Manager::getInstance()->getLoadedPluginsName() : $pluginNames;
+            $plugins = !$pluginNames ? $manager->getActivatedPlugins() : $pluginNames;
             sort($plugins);
 
             $pluginsInfo = '';
             foreach ($plugins as $pluginName) {
-                $plugin      = Manager::getInstance()->getLoadedPlugin($pluginName);
-                $pluginsInfo .= $plugin->getPluginName() . $plugin->getVersion() . ',';
+                if ($manager->isPluginLoaded($pluginName)) {
```
diosmosis, line 43: Should we do this check here? If a plugin is in the activated plugins list, but for some reason isn't loaded, shouldn't the cache buster stay the same? I think we'd only change it if a plugin is activated or deactivated, not if a plugin isn't loaded for some reason.

tsteur: @diosmosis I'm not 100% sure but here's my thought. The asset manager posts an event to get all JS or CSS files. The event dispatcher in https://github.com/matomo-org/matomo/blob/4.3.1/core/EventDispatcher.php#L97 uses `getPluginsLoadedAndActivated()` and events will be only posted to plugins that are active and loaded. 

Meaning the hash would need to change if suddenly a plugin is loaded that wasn't loaded before because the JS/CSS would change too. I'm assuming a plugin may be activated but not loaded if it's missing from the filesystem (eg temporarily). Maybe there are also other reasons like when we unload a plugin in WhiteLabel etc.

I would probably keep the check but not 100% sure if it the thinking makes sense

diosmosis: I see, no your logic makes sense. The plugins used are the ones that are specifically used to generate the JS/CSS, and there can be a discrepancy with the list of activated plugins and list of loaded plugins.

PR #16886, opened 2020-12-05
If archive_invalidations is in inconsistent state, fix as getting next archive to process.

diosmosis:
### Description:

In case the archive_invalidations table ends up in an inconsistent state, make sure we process invalidations higher than what we find.

cc @tsteur 

### Review

* [ ] Functional review done
* [ ] Code review done


core/CronArchive/QueueConsumer.php
```diff
@@ -311,6 +315,66 @@ public function getNextArchivesToProcess()
...
+            $archiveToProcess['date2'],
+        ];
+
+        $reportClause = '';
+        if (!empty($archiveToProcess['report'])) {
+            $reportClause = " AND report = ?";
+            $bind[] = $archiveToProcess['report'];
+        }
+
+        $sql = "SELECT DISTINCT period FROM `$table` WHERE idsite = ? AND name = ? AND period > ? AND ? >= date1 AND date2 >= ? $reportClause";
```
tsteur: btw is the distinct actually needed?

tsteur:
@sgiehl be great if you could review this tomorrow. We'd probably need to merge this at the same time as https://github.com/matomo-org/matomo/pull/16843 to fully avoid too many duplicates.

core/CronArchive/QueueConsumer.php
diosmosis: could do an array_unique as well, was trying to reduce the amount of data selected since there could potentially be a lot of invalidations for a site

tsteur: all good should be fine to keep distinct for now

core/CronArchive/QueueConsumer.php
```diff
@@ -278,6 +280,8 @@ public function getNextArchivesToProcess()
 
             $this->logger->debug("Processing invalidation: $invalidationDesc.");
 
+            $this->repairInvalidationsIfNeeded($invalidatedArchive);
```
tsteur: @diosmosis any thought about doing this in `CronArchive::launchArchivingFor()` after it finished successfully? Technically we don't need to schedule the higher period if it's not successful and also it could prevent a race condition where the `day` archive takes many hours and by the time it's finished the higher period does not exist anymore if I see this right?

core/CronArchive/QueueConsumer.php
```diff
@@ -311,6 +315,66 @@ public function getNextArchivesToProcess()
...
+            if ($id <= $archiveToProcess['period'] || $label == 'range') {
+                continue;
+            }
+
+            if (isset($higherPeriods[$id])) { // period exists in table
+                continue;
+            }
+
+            // archive is for week that is over two months, we don't need to care about the month
+            if ($label == 'month'
```
tsteur: not sure why we don't need to care about this case? There might be still issues eg when someone imports data in the past etc?

tsteur (changes requested):

core/CronArchive/QueueConsumer.php
diosmosis: >  any thought about doing this in CronArchive::launchArchivingFor() after it finished successfully?

I figured it would be better to do it here, since when we invalidate through ArchiveInvalidator, we always add all higher periods. So at the start, we always have the whole list of invalidations.

> Technically we don't need to schedule the higher period if it's not successful and also it could prevent a race condition where the day archive takes many hours and by the time it's finished the higher period does not exist anymore if I see this right?

The only case where the higher period would not exist anymore is if another archiver was processing the same site and picked it up, but this could happen w/o this repairing logic... In that case we'd have to do the intersecting period logic w/ a query on archive_invalidations.

core/CronArchive/QueueConsumer.php
diosmosis: It's because the week will be on two months, so it won't actually be used in archiving either of the parent months.

There could be the case where there was a day that was archived, then for some reason the month/year disappear, and it's just the week. Then I guess we would need new month/year re-archives, but wouldn't be able to.

It seemed a waste to have to re-archive both months, but I can add it if necessary.

core/CronArchive/QueueConsumer.php
tsteur: >  So at the start, we always have the whole list of invalidations.

you mean then we can batch insert it? I reckon ideally we do it after finishing the archive just to make sure there can't be a race condition or any other issues I would say. It be fine if it's just one insert at a time I would say (we could always tweak when needed)

core/CronArchive/QueueConsumer.php
tsteur: Got it now 👍  We only need to archive the correct month

core/CronArchive/QueueConsumer.php
diosmosis: @tsteur no I mean, in the invalidations table, normally we have all the higher periods before we start archiving, so doing the repair logic here mimics that behavior.

We can still do it after finishing an archive (though there will be less test coverage then).

tsteur: Be great to do it after finishing the archive. Could keep the method and just call it from CronArchive instead of QueueConsumer @diosmosis ? It's really just to avoid random edge cases etc which will happen

diosmosis:
@tsteur updated, there are still tests for it, but some of the effects seen in the queue consumer tests are no longer there

core/CronArchive.php
```diff
@@ -938,6 +934,77 @@ public function isThereExistingValidPeriod(Parameters $params)
...
+        if (!empty($archiveToProcess['report'])) {
+            $reportClause = " AND report = ?";
+            $bind[] = $archiveToProcess['report'];
+        }
+
+        $sql = "SELECT DISTINCT period FROM `$table` WHERE idsite = ? AND name = ? AND period > ? AND ? >= date1 AND date2 >= ? $reportClause";
+
+        $higherPeriods = Db::fetchAll($sql, $bind);
+        $higherPeriods = array_column($higherPeriods, 'period');
+        $higherPeriods = array_flip($higherPeriods);
```
tsteur: not important but could have done instead below an `in_array($id, $higherPeriods)` . not too important though

core/CronArchive.php
```diff
@@ -938,6 +934,77 @@ public function isThereExistingValidPeriod(Parameters $params)
...
+
+            $invalidationToInsert = [
+                'idarchive' => null,
+                'name' => $archiveToProcess['name'],
+                'report' => $archiveToProcess['report'],
+                'idsite' => $archiveToProcess['idsite'],
+                'date1' => $period->getDateStart()->getDatetime(),
+                'date2' => $period->getDateEnd()->getDatetime(),
+                'period' => $id,
+                'ts_invalidated' => $archiveToProcess['ts_invalidated'],
```
tsteur, line 990: should this have maybe the current date? Just to be more clear when it was inserted. Or is it so it's included to be archived in this archive run maybe? (as we maybe do somewhere `and ts_invalidated < idSiteStartTime`)

core/CronArchive.php
```diff
@@ -938,6 +934,77 @@ public function isThereExistingValidPeriod(Parameters $params)
...
+            $archiveToProcess['date2'],
+        ];
+
+        $reportClause = '';
+        if (!empty($archiveToProcess['report'])) {
+            $reportClause = " AND report = ?";
+            $bind[] = $archiveToProcess['report'];
+        }
+
+        $sql = "SELECT DISTINCT period FROM `$table` WHERE idsite = ? AND name = ? AND period > ? AND ? >= date1 AND date2 >= ? $reportClause";
```
tsteur: wondering. Do we here also need to include `and status = 0`?

tsteur (commented):
@diosmosis worked for me. Only need to maybe have a look at https://github.com/matomo-org/matomo/pull/16886/files#r537995820 as maybe the query needs to have to check status

tsteur:
Otherwise good to merge once the other comment re the `where` clause is resolved and tests pass

core/CronArchive.php
diosmosis: Yes, it's done that way so we process it in this batch. (getNextInvalidatedArchive does the `ts_invalidated < idSiteStartTime`)

core/CronArchive.php
diosmosis: sure, thought this was a bit faster, but for an array w/ 3 items at most, probably not

core/CronArchive.php
diosmosis: 👍

PR #16893, opened 2020-12-06
Fix ArgumentCountError: mysqli_stmt::bind_param() does not accept unknown named parameters

### LINKED ISSUE #16885: named parameters in mysqli_stmt->bind_param
ERROR [2020-12-01 18:52:03] 22241  Uncaught exception: ArgumentCountError: mysqli_stmt::bind_param() does not accept unknown named parameters in libs/Zend/Db/Statement/Mysqli.php:203
Stack trace:
#0 libs/Zend/Db/Statement/Mysqli.php(203): mysqli_stmt->bind_param('ssssssss', idarchive: 403, name: 'done', report: NULL, idsite: 19, date1: '2020-12-01 00:0...', date2: '2020-12-01 00:0...', period: 1, ts_invalidated: '2020-12-01 18:5...')

PHP 8.0.0
Matomo v. 4.0.0 - current

file: libs/Zend/Db/Statement/Mysqli.php
line: 199
$stmtParams[$k] = &$value;
must be:
$stmtParams[] = &$params[$k];

tsteur:
fix https://github.com/matomo-org/matomo/issues/16885

It was basically an array like `array('idarchive'=> '', '', ...)` when it should be `[1,'name']`

Wondering though if the better fix would be in https://github.com/matomo-org/matomo/blob/4.0.4/core/DataAccess/Model.php#L228-L237 to make sure it's a regular array there?

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Db/BatchInsert.php
+++ b/core/Db/BatchInsert.php
@@ -35,6 +35,7 @@ class BatchInsert
         $ignore    = $ignoreWhenDuplicate ? 'IGNORE' : '';
 
         foreach ($values as $row) {
+            $row = array_values($row);
             $query = "INSERT $ignore INTO " . $tableName . "
 					  $fieldList
 					  VALUES (" . Common::getSqlStringFieldsArray($row) . ")";


diosmosis:
> Wondering though if the better fix would be in https://github.com/matomo-org/matomo/blob/4.0.4/core/DataAccess/Model.php#L228-L237 to make sure it's a regular array there?

No opinion on this personally, seems like if we ignore the keys anyway this fix is fine

PR #16896, opened 2020-12-07
Only invalidate in past for segments that belong to the site …

### LINKED ISSUE #16892: Check only invalidations created for correct sites
We have a customer that reports this:

> Piwik\Segment->__construct(dimension17==DE, Array ([0] => 1)) - that looks like Matomo is trying to load a segment from another idSite for Site 1, which will fail because idSite1 does not have a dimension17.

image

Here is the backtrace:

```
[32mINFO [2020-12-05 15:30:25] 8485 [39m Start processing archives for site 1.
#0  Piwik\Segment->getSegmentByName(dimension17) called at [core/Segment.php:258]
#1  Piwik\Segment->getExpressionsWithUnionsResolved(Array ([0] => Array ([0] => ,[1] => Array ([0] => dimension17,[1] => ==,[2] => DE)))) called at [core/Segment.php:236]
#2  Piwik\Segment->initializeSegment(dimension17==DE, Array ([0] => 1)) called at [/var/app/current/core/Segment.php:145]
#3  Piwik\Segment->__construct(dimension17==DE, Array ([0] => 1)) called at [core/CronArchive/SegmentArchiving.php:128]
#4  Piwik\CronArchive\SegmentArchiving->findSegmentForHash(90a5a511e1974bca37613b6daec137ba, 1) called at [core/CronArchive/QueueConsumer.php:472]
#5  Piwik\CronArchive\QueueConsumer->findSegmentForArchive(Array ([idinvalidation] => 210,[idarchive] => ,[idsite] => 1,[date1] => 2020-12-04,[date2] =>
```

@diosmosis can you check if it's possible that we invalidate entries for a segment for a wrong site somehow?

That's using 4.0.4

diosmosis:
### Description:

… and only process an invalidation if it is for a site + we can load the segment

Fixes #16892
FYI @tsteur 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/CronArchive/SegmentArchiving.php
+++ b/core/CronArchive/SegmentArchiving.php
@@ -20,7 +20,6 @@ use Piwik\Plugins\SegmentEditor\Model;
 use Piwik\Segment;
 use Piwik\Site;
 use Psr\Log\LoggerInterface;
-use function DI\object;
 
 /**
  * Provides URLs that initiate archiving during cron archiving for segments.
@@ -305,7 +304,7 @@ class SegmentArchiving
         return $segments;
     }
 
-    public function isSegmentForSite($segment, $idSite)
+    private function isSegmentForSite($segment, $idSite)
     {
         return $segment['enable_only_idsite'] == 0
             || $segment['enable_only_idsite'] == $idSite;
--- a/core/CronArchive/SegmentArchiving.php
+++ b/core/CronArchive/SegmentArchiving.php
@@ -20,6 +20,7 @@ use Piwik\Plugins\SegmentEditor\Model;
 use Piwik\Segment;
 use Piwik\Site;
 use Psr\Log\LoggerInterface;
+use function DI\object;
 
 /**
  * Provides URLs that initiate archiving during cron archiving for segments.
@@ -89,7 +90,9 @@ class SegmentArchiving
 
         $segmentsForSite = $this->getAllSegments();
         foreach ($segmentsForSite as $storedSegment) {
-            if (!$this->isAutoArchivingEnabledFor($storedSegment)) {
+            if (!$this->isAutoArchivingEnabledFor($storedSegment)
+                || !$this->isSegmentForSite($storedSegment, $idSite)
+            ) {
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


core/CronArchive/SegmentArchiving.php
```diff
@@ -20,6 +20,7 @@
 use Piwik\Segment;
 use Piwik\Site;
 use Psr\Log\LoggerInterface;
+use function DI\object;
```
sgiehl: guess that use statement isn't needed

core/CronArchive/SegmentArchiving.php
```diff
@@ -294,7 +305,7 @@ public function getAllSegmentsToArchive($idSite)
         return $segments;
     }
 
-    private function isSegmentForSite($segment, $idSite)
+    public function isSegmentForSite($segment, $idSite)
```
sgiehl: Any reason for having that on public now? Seems it isn't used outside the class. Not even in tests...

core/CronArchive/SegmentArchiving.php
```diff
@@ -121,11 +124,19 @@ public function getSegmentArchivesToInvalidate($idSite, $checkOnlyForNewSegments
...
+            if (!$this->isAutoArchivingEnabledFor($segment)
+                || !$this->isSegmentForSite($segment, $idSite)
+            ) {
+                continue;
+            }
+
+            try {
+                $segmentObj = new Segment($segment['definition'], [$idSite]);
+            } catch (\Exception $ex) {
+                $this->logger->debug("Could not process segment {$segment['definition']} for site {$idSite}. Segment should not exist for the site, but does.");
```
sgiehl, line 135: The debug message confuses me. Isn't an exception thrown when the segment does not exist?
Also this case shouldn't occur at all if the `isSegmentForSite` works correctly, right?

sgiehl (changes requested):

core/CronArchive/SegmentArchiving.php
diosmosis: It shouldn't occur, but if it does it would break archiving, and I didn't think that needed to happen. This is a sanity check.

diosmosis:
@tsteur @sgiehl merging this, any other changes I can make in a new PR

PR #16901, opened 2020-12-07
better Windows detection

Findus23:
Not related to any actual issue, but I just came across this constant which should be more reliable and now with PHP 7.2 available for everyone.

> **`PHP_OS_FAMILY`** (string)
>
> The operating system family PHP was built for. One of `'Windows'` , `'BSD'` , `'Darwin'` , `'Solaris'` , `'Linux'` or `'Unknown'` . Available as of PHP 7.2.0.
> https://www.php.net/manual/en/reserved.constants.php

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/SettingsServer.php
+++ b/core/SettingsServer.php
@@ -91,6 +91,9 @@ class SettingsServer
      */
     public static function isWindows()
     {
+        if (PHP_OS_FAMILY == "Unknown") {
+            return DIRECTORY_SEPARATOR === '\\';
+        }
         return PHP_OS_FAMILY === "Windows";
     }
 
--- a/core/SettingsServer.php
+++ b/core/SettingsServer.php
@@ -91,7 +91,7 @@ class SettingsServer
      */
     public static function isWindows()
     {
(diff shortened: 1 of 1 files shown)


core/SettingsServer.php
```diff
@@ -91,7 +91,7 @@ public static function isIIS()
      */
     public static function isWindows()
     {
-        return DIRECTORY_SEPARATOR === '\\';
+        return PHP_OS_FAMILY === "Windows";
```
sgiehl, line 97: Wondering if there might be any case where `Windows` could be reported as `Unknown`. Guess we could fall back to the old check in that case. But might be enough to add that as soon as someone reports an issue with it...

Findus23: I doubt many people use binaries apart from the official ones on Windows. But this might make sense for non-(Linux|BSD|MacOS) operating systems that Matomo probably should handle like Windows (don't assume unix tools exists).

diosmosis (approved):
didn't test, but looks good

PR #16910, opened 2020-12-08
Adds possibility to configure protocols for allowed outgoing connections

sgiehl:
### Description:

Matomo should normally not send any request using other protocols than http and https.
In some uncommon use cases it might be needed to send other request. This can be done using a config setting to define the list of allowed protocols.

### Review

* [ ] Functional review done
* [ ] Code review done


config/global.ini.php
```diff
@@ -719,6 +719,10 @@
 ; If set to 0 it also disables the "sent plugin update emails" feature in general and the related setting in the UI.
 enable_update_communication = 1
 
+; This option defines the protocols Matomo's Http class is allowed to open.
+; If you may need to download GeoIP updates or other stuff using other protocols like ftp you may need to extend this list.
+allowed_protocols = 'http,https'
```
diosmosis: Do you think it might be better to name this `allowed_outgoing_protocols` or similar? To avoid confusion about what it refers to.

diosmosis (commented):
Left a comment, generally looks good, think the UI tests will have to be updated

config/global.ini.php
sgiehl: updated the name

diosmosis (approved):

diosmosis:
@sgiehl seems like there are some test failures

sgiehl:
@diosmosis should be fixed now

PR #16913, opened 2020-12-08
More efficient page performance request processor

### LINKED ISSUE #16912: Performance plugin causes select query in request processor
Can we maybe first check if there are any request parameters in the request processor before executing a select on every non page view tracking request? This would reduce CPU and load on a DB server quite a bit and make tracking faster.

tsteur:
### Description:

fix https://github.com/matomo-org/matomo/issues/16912

@sgiehl could you look at this one? Should this cause any issues? If so, what would be an alternative fix?

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (approved):
Makes sense to avoid any queries if no params are given. Haven't tested but if tests pass should be good to merge

PR #16920, opened 2020-12-09
Ensure metadata is available if page url can't be found

### LINKED ISSUE #16918: Call to a member function getDateStart() on boolean DataTable/Renderer/Rss.php
Error seen on Cloud:

 Error: {"message":"Call to a member function getDateStart() on boolean","file":"\/var\/www\/html\/core\/DataTable\/Renderer\/Rss.php","line":62,"request_id":"1e86d","backtrace":" on \/var\/www\/html\/core\/DataTable\/Renderer\/Rss.php(62)\n#0 \/var\/www\/html\/core\/DataTable\/Renderer\/Rss.php(34): Piwik\\DataTable\\Renderer\\Rss->renderTable(Object(Piwik\\DataTable\\Map))\n#1 \/var\/www\/html\/plugins\/API\/Renderer\/Rss.php(40): Piwik\\DataTable\\Renderer\\Rss->render()\n#2 \/var\/www\/html\/core\/API\/ResponseBuilder.php(193): Piwik\\Plugins\\API\\Renderer\\Rss->renderDataTable(Object(Piwik\\DataTable\\Map))\n#3 \/var\/www\/html\/core\/API\/ResponseBuilder.php(104): Piwik\\API\\ResponseBuilder->handleDataTable(Object(Piwik\\DataTable\\Map))\n#4 \/var\/www\/html\/core\/API\/Request.php(272): Piwik\\API\\ResponseBuilder->getResponse(Object(Piwik\\DataTable\\Map), 'Actions', 'getPageUrl')\n#5 \/var\/www\/html\/core\/Context.php(28): Piwik\\API\\Request->Piwik\\API\\{closure}()\n#6 \/var\/www\/html\/core\/API\/Request.php(273): Piwik\\Context::executeWithQueryParameters(Array, Object(Closure))\n#7 \/var\/www\/html\/plugins\/API\/Controller.php(46): Piwik\\API\\Request->process()\n#8 [internal function]: Piwik\\Plugins\\API\\Controller->index()\n#9 \/var\/www\/html\/core\/FrontController.php(605): call_user_func_array(Array, Array)\n#10 \/var\/www\/html\/core\/FrontController.php(166): Piwik\\FrontController->doDispatch('API', false, Array)\n#11 \/var\/www\/html\/core\/dispatch.php(32): Piwik\\FrontController->dispatch()\n#12 \/var\/www\/html\/index.php(25): require_once('\/var\/www\/html\/c...')\n#13 {main}"}

URL: https://xyz.matomo.cloud//?module=API&format=rss&pageUrl=%xxxxxxxxxxxxxxxx&idSite=1&period=day&date=last7&token_auth=XYZANONYMIZED&method=Actions.getPageUrl

Referrer:

GET: {"module":"API","format":"rss","pageUrl":"\xxxxxxxxxxxx","idSite":"1","period":"day","date":"last7","token_auth":"XYZANONYMIZED","method":"Actions.getPageUrl","filter_limit":100}

Time: 2020-12-09 09:09:38

sgiehl:
### Description:

When sending a request like this:

{"module":"API","format":"rss","pageUrl":"\xxxxxxxxxxxx","idSite":"1","period":"day","date":"last7","token_auth":"XYZANONYMIZED","method":"Actions.getPageUrl","filter_limit":100}

the api method `Actions.getPageUrl` currently might return a completely empty datatable, if the url can't be found. But the RSS renderer tries to use the `period` metadata of the datatable, which does not exist, and fails.

Simply returning the first subtables metadata for the empty table response solves that issue.

fixes #16918 

### Review

* [ ] Functional review done
* [ ] Code review done


plugins/Actions/API.php
```diff
@@ -423,7 +423,14 @@ protected function doFilterPageDatatableSearch($callBackParameters, $table, $sea
...
             // nothing found in all sub tables
-            return new DataTable;
+            $result = new DataTable;
+            $subTables = $table->getDataTables();
+            if (count($subTables) > 0) {
+                // use the first subtable's metadata to ensure basic metadata like `period` is available in response
+                $subTable = reset($subTables);
+                $result->setAllTableMetadata($subTable->getAllTableMetadata());
+            }
+            return $result;
```
diosmosis, line 428: Will this actually set the correct period though? If we query with a multi period, then this will set it to the first subperiod, so in the last7 example, it would be 7 days ago. Is this what callers of the API would expect? I think maybe the Rss formatter doesn't support multiperiod requests, and some other changes are needed?

diosmosis (changes requested):

plugins/Actions/API.php
sgiehl: Actually no. That piece of code is executed for datatable maps inside a map. I'll remove some useless code so it gets more clear.

diosmosis:
It's a bit clearer, but I'm still unsure as to whether the period we pick in the list of tables will be what is expected by callers. I also noticed that this code doesn't appear to be tested (there don't seem to be any tests that call getPageUrl/getOutlink/etc.). Do you think it would be a good idea to add an integration or system test to cover this behavior and maybe check the period that is returned? If it's there isn't one already, perhaps there could be a rss test for this as well?

sgiehl:
@diosmosis I've just added some RSS tests

diosmosis:
@sgiehl is there a test that requests a page URL that cannot be found (to trigger the bug this PR fixes)?

sgiehl:
@diosmosis I've just added a test that should throw an exception on 4.x-dev

PR #16925, opened 2020-12-09
do not update token usage in tracking requests

### LINKED ISSUE #16924: Tons of "General error: 1205" in php error log while importing via the python log importer
It seems that there is a problem with the user_token_auth while importing log via the python logimporter. 
While looking at the SQL processlist I see several UPDATE queries for the matomo_user_token_auth table, they run the entire time until the timeout:

`[09-Dec-2020 09:28:08 UTC] Error in Matomo (tracker): Error query: SQLSTATE[HY000]: General error: 1205 Lock wait timeout exceeded; try restarting transaction In query: UPDATE 'matomo_user_token_auth' SET 'last_used' = ? WHERE 'idusertokenauth' = ? Parameters: array (   0 => '2020-12-09 09:27:17',   1 => '207', )`

Manually executing one of these queries works like a charm in some milliseconds. But not while the python script runs.

After the timeout the python script echoes this:

2020-12-09 09:28:08,641: [INFO] Error when connecting to Matomo: HTTP Error 500: Internal Server Error
2020-12-09 09:28:08,642: [INFO] Retrying request, attempt number 2

That's since the latest 4.0.4 update (I updated from v3) and I use Python 3.7 to execute the script.

sgiehl:
### Description:

Seems updating the usage timestamp of a token might cause issues when using log import.
Not sure if we need to update the usage at all in this case. If so we maybe could also update it every 10 minutes only or so, but that would at least require an additional read request to check that.

fixes #16924

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/UsersManager/Model.php
+++ b/plugins/UsersManager/Model.php
@@ -19,6 +19,7 @@ use Piwik\Plugins\SitesManager\SitesManager;
 use Piwik\Plugins\UsersManager\Sql\SiteAccessFilter;
 use Piwik\Plugins\UsersManager\Sql\UserTableFilter;
 use Piwik\SettingsPiwik;
+use Piwik\SettingsServer;
 use Piwik\Validators\BaseValidator;
 use Piwik\Validators\CharacterLength;
 use Piwik\Validators\NotEmpty;
@@ -386,6 +387,10 @@ class Model
 
     public function setTokenAuthWasUsed($tokenAuth, $dateLastUsed)
     {
+        if (SettingsServer::isTrackerApiRequest()) {
+            return; // do not update usage in tracking requests as this can cause trouble during log import
+        }
+
         $token = $this->getTokenByTokenAuth($tokenAuth);
         if (!empty($token)) {
             $this->updateTokenAuthTable($token['idusertokenauth'], array(
--- a/plugins/UsersManager/Model.php
+++ b/plugins/UsersManager/Model.php
@@ -387,12 +387,17 @@ class Model
 
     public function setTokenAuthWasUsed($tokenAuth, $dateLastUsed)
     {
-        if (SettingsServer::isTrackerApiRequest()) {
-            return; // do not update usage in tracking requests as this can cause trouble during log import
-        }
-
         $token = $this->getTokenByTokenAuth($tokenAuth);
         if (!empty($token)) {
+
+            $lastUsage = strtotime($token['last_used']);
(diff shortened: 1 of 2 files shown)


tsteur:
Seems this doesn't fix the issue? Also it shouldn't really be a problem I think as when using the API concurrently we would see these issues as well maybe? Generally could change the where though like 

```sql
UPDATE 'matomo_user_token_auth' SET 'last_used' = ? WHERE 'idusertokenauth' = ? and last_used < ?
```
It likely wouldn't make a difference though . We basically only want to update it when `time()` is higher. When value doesn't change then mysql should be smart enough though.

sgiehl:
Maybe that's only one part of the issue. But firing thousands of updates on the same row isn't good in any case.
Will have a closer look at the code again later. Maybe there's any easy way to update it like only every 10 minutes or so

tsteur:
Be great to indeed only update it like every 10 min (in tracker and in general). Probably can't use general tracker cache as it would invalidate the cache every 10min when we use a much longer cache time on cloud.(1hr+)

sgiehl:
@tsteur updated the PR. There will now always be an select to check if the last date for the token is older than 10 minutes. Not sure if that might make problems if multiple requests are executed at the same time

plugins/UsersManager/Model.php
```diff
@@ -388,6 +389,15 @@ public function setTokenAuthWasUsed($tokenAuth, $dateLastUsed)
     {
         $token = $this->getTokenByTokenAuth($tokenAuth);
         if (!empty($token)) {
+
+            $lastUsage = strtotime($token['last_used']);
```
tsteur: @sgiehl can last_used key be empty?

plugins/UsersManager/Model.php
```diff
@@ -388,6 +389,15 @@ public function setTokenAuthWasUsed($tokenAuth, $dateLastUsed)
...
+            $newUsage = strtotime($dateLastUsed);
+
+            // update token usage only every 10 minutes
```
tsteur: might be good to add a comment why this is eg link to https://github.com/matomo-org/matomo/issues/16924 ?

plugins/UsersManager/Model.php
sgiehl: guess it can be. will add a check

plugins/UsersManager/Model.php
sgiehl: done.

PR #16926, opened 2020-12-09
Rewrite Annotations tests to run correctly

sgiehl:
### Description:

The tests actually didn't work anymore and passed silently. Will add some comments to explain that

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Annotations/tests/System/AnnotationsTest.php
+++ b/plugins/Annotations/tests/System/AnnotationsTest.php
@@ -264,29 +264,32 @@ class AnnotationsTest extends SystemTestCase
      */
     public function testMethodPermissions($hasAdminAccess, $hasViewAccess, $request, $checkException, $failMessage)
     {
+        if (true === $checkException) {
+            self::expectException(Exception::class);
+        } else {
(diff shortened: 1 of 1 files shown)


plugins/Annotations/tests/System/AnnotationsTest.php
```diff
@@ -264,30 +264,33 @@ public function getPermissionsFailData()
...
-            } catch (Exception $ex) {
-                $this->assertTrue(true); // pass
-            }
-        } else {
-            $request = new Request($request);
-            $request->process();
-
-        }
+        $request = new Request($request.'&format=original');
+        $request->process();
```
sgiehl, line 280: The requests actually did not contain a format, so xml response was returned. In case of an error there was no exception thrown by the request. But with the meanwhile used PHPUnit version `$this->fail()` throws an `AssertionFailedError` exception, which was directly caught and `$this->assertTrue(true)` did make all tests pass.

Will look through the other tests as well and check where we use `$this->fail()` in a try/catch

Issue #16927, opened 2020-12-09
After updating to 4.0.5, plugin file VisitsByDaysSinceLastVisit.php is not deleted


nkfilis:
After upgrading from 3.5.x to 4.0.5 today, the `console queuedtracking:process` died with this error:
```
Starting to process request sets, this can take a while
Uncaught exception in /path/to/matomo/installation/plugins/VisitorInterest/Columns/VisitsByDaysSinceLastVisit.php line 38:
Call to undefined method Piwik\Tracker\Request::getDaysSinceLastVisit()
```
It seems, the plugin was updated, but an old file wasn't deleted:
```
plugins/VisitorInterest/Columns:
total 27
-rw-r--r--  1 www  www   366 Dec  8 22:07 PagesPerVisit.php
-rw-r--r--  1 www  www   412 Dec  8 22:07 VisitDuration.php
-rw-r--r--  1 www  www   671 Dec  8 22:07 VisitorDaysSinceLast.php
-rw-r--r--  1 www  www  1963 Dec  8 22:07 VisitorSecondsSinceLast.php
-rw-r--r--  1 www  www  1062 Sep 23 23:20 VisitsByDaysSinceLastVisit.php
-rw-r--r--  1 www  www   371 Dec  8 22:07 VisitsbyVisitNumber.php
```

At least, after moving the file elsewhere, everything went thru fine. Maybe an issue with the plugin update?

I'm not 100% sure, which version I was upgrading from, but I see this in the terminal buffer: `Matomo database will be upgraded from version 3.14.1 to the new version 4.0.5.`

sgiehl:
@tsteur Seems we are having similar issues with various removed dimensions. Do you think it would maybe make sense to have some kind of dimension integrity checker, that holds a list of dimensions that should have been removed through an update and would remove it from the list of used dimensions if it's still there, so it can't make any problems?

sgiehl:
@tsteur created a simple draft on this branch: https://github.com/matomo-org/matomo/compare/removedimensions. 
Guess might be useful as it seems to happen for wordpress plugin as well. See https://github.com/matomo-org/wp-matomo/issues/393
Other option might be to clean up the options table, so the columns are actually not listed there anymore. But that would trigger the updater to add them again if they are not deleted...

tsteur:
@sgiehl sounds good. We maybe also want to document somewhere (or have a test if possible) to add future removed dimensions to that list

sgiehl:
[...]

PR #16934, opened 2020-12-10
Ensure removed dimensions are not used even if they still exist

sgiehl:
### Description:

In some cases it might happen that a dimension file that should have been deleted with an update fails to be removed.
In that case Matomo might fail in various places, as it might try to use a database column that does not exist any longer.

By maintaining a list of removed dimensions, we can ensure those dimensions won't be used anymore, even if the still exist.

TODO: Need to check if any other dimension was removed in 4.0 and how to document that clearly or maybe a some kind of test to check if something was removed but isn't in the list

refs #16927

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl:
@tsteur The PR as is would actually be good to merge and would possibly solve some update issues where one of the dimension file couldn't be removed.

But I'm currently thinking on how to best check for removed dimensions, to avoid such issues in the future if the list isn't updated. But actually I don't see a good way for doing that. We could check the git history for removed files and check if one of those files was a dimension (using it's own database column). But actually I'm not sure if it's worth the effort.

Also I wonder if we actually shouldn't remove the dimension version in option table with an update that removes a dimension.
For people where the updater didn't remove the file, that might actually cause a update adding the column again. So might be not that good, but at least would avoid such issues 🤔

tsteur:
@sgiehl we wouldn't want to look through git history. 

Could we add something like https://github.com/matomo-org/matomo/blob/4.0.5/plugins/Widgetize/tests/System/WidgetTest.php#L57-L75 where we have a list of all available dimensionIds. When one dimension is removed, we'd expect the test to fail and would mention something like "if the dimension was removed on purpose you now need to add it to the list of removed dimensions" or similar. Basically we'd maintain a hard coded list of dimensions we have in the test. So we notice when one is added or removed.

Haven't thought this through but maybe that would work.

Re removing options for these columns: It be a problem indeed if it was to try to add the columns again and again. Maybe instead the above test could help in the future as well to not add the same column name again? I wouldn't spend too much time on it but might be worth a check like if we had a list of `getDeprecatedDimensions()` and `getDeprecatedColumns()` then we could have a test that makes sure we don't add the same ones again in the future so the option could still exist.

sgiehl:
@tsteur I've updated the PR. For now the test only checks the dimensions holding a column definition. Guess those dimensions might make the most trouble when removed. Or do you think we should maintain a list of all dimensions?

core/Columns/Dimension.php
```diff
@@ -758,13 +758,35 @@ public static function getDimensions(Plugin $plugin)
...
+        foreach ($columns as $column) {
+            $instances[] = new $column();
         }
 
+        /**
+         * Triggered to filter / restrict dimensions.
+         *
+         * @param Dimension[] $dimensions An array of dimensions
+         */
+        Piwik::postEvent('Dimension.filterDimensions', array(&$instances));
```
tsteur: @sgiehl There's already a `Dimension.filterDimension` further above. Will debugging noticed it might not be needed to have it here as well and it's enough in `getDimensions()`? If we keep it here then we should in  `removeOldDimensions ` check if `$instances` is empty and return quickly not to get the removed dimensions 100 times when most plugins don't define any. I haven't fully tested but maybe it could be even removed the event here?

Not that it's a big change but I'm thinking we could even have the check for removed dimensions directly in this class for simplicity and wouldn't need an event for it.

plugins/CoreHome/CoreHome.php
```diff
@@ -487,4 +489,19 @@ public function getClientSideTranslationKeys(&$translationKeys)
...
+     * Ensures dimensions that should have been removed are not used any longer even if the file failed to remove on update
+     *
+     * @param $instances
+     */
+    public function removeOldDimensions(&$instances)
+    {
+        $removedDimensions = Dimension::getRemovedDimensions();
+
+        $instances = array_filter($instances, function ($dimension) use ($removedDimensions) {
+            $className = get_class($dimension);
```
tsteur: btw I suppose there's no performance difference between `get_class` vs `$dimension::class`?  Might not be needed though this improvement anyway as it's only called like 100 times

core/Columns/Dimension.php
sgiehl: Actually we need both calls of `Dimension.filterDimensions`. If someone directly uses the `getDimensions` method the dimensions actually weren't filtered before. Which was imho no correct behaviour. Only doing it in `getDimensions` would make it impossible to filter out a dimension that was added by another plugin using `Dimension.addDimensions` event.

Removing such dimensions here directly sound fine as well. Will move it here

plugins/CoreHome/CoreHome.php
sgiehl: `::class` doesn't work on instances. Also get_class returns kind of the "true" names of classes, whereas `::class` returns the name used. See this example:
```
echo get_class(new DatETime()); // prints DateTime
echo DatETime::class; // prints DatETime
```

core/Columns/Dimension.php
sgiehl: done.

core/Columns/Dimension.php
```diff
@@ -758,13 +758,37 @@ public static function getDimensions(Plugin $plugin)
         $columns = $plugin->findMultipleComponents('Columns', '\\Piwik\\Columns\\Dimension');
         $instances  = array();
 
-        foreach ($columns as $colum) {
-            $instances[] = new $colum();
+        foreach ($columns as $column) {
+            if (!in_array($column, self::getRemovedDimensions())) {
```
tsteur: @sgiehl be great to avoid calling that method in every for loop maybe?

plugins/DevicePlugins/DevicePlugins.php
```diff
@@ -56,6 +57,14 @@ public static function getAllPluginColumns()
                 $instance = new $className();
                 $instances[] = $instance;
             }
+
+            /**
+             * Triggered to filter / restrict dimensions.
+             *
+             * @param Dimension[] $dimensions An array of dimensions
+             */
+            Piwik::postEvent('Dimension.filterDimensions', array(&$instances));
```
tsteur: @sgiehl is this event name expected? Or maybe it was meant to remove some dimensions but now it's done in core  and not in plugin? Could call a core method here like `Dimension::removeDisabledDimensions()`

BTW just seeing we support other plugins to add extra device columns? Is this used currently? If not could maybe even just hard code the columns.

core/Columns/Dimension.php
tsteur: @sgiehl ok. Can we remove the event below though? That would be great. It be fine for now if not all dimensions are 100% filtered in these cases I would say since most of the time "all dimensions" would be called. I'd want to avoid potentially triggering a lot of events here slowing Matomo down in some cases.

plugins/DevicePlugins/DevicePlugins.php
sgiehl: > BTW just seeing we support other plugins to add extra device columns? Is this used currently? If not could maybe even just hard code the columns.

Using that for at least this plugin: https://github.com/sgiehl/piwik-plugin-DeviceFeatureWebGL
Not sure if anyone else is using it in a plugin....

sgiehl:
@tsteur applied the feedback

PR #16941, opened 2020-12-12
Improve handling of current user information for Filechecks

aragon999:
### Description:
I had a PHP Notice:
```
PHP Notice:  Trying to access array offset on value of type bool in Filechecks.php:205
```
while the actual problem was, that I was inside a docker container and had the wrong group-id assigned.

I only had a notice, since the user permissions still worked, but I think that place needs better handling. This PR is a first try to improve the situation.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Filechecks.php
+++ b/core/Filechecks.php
@@ -148,9 +148,21 @@ class Filechecks
 
         $group = trim(shell_exec('groups '. $user .' | cut -f3 -d" "'));
 
+        if (empty($group) && function_exists('posix_getegid') && function_exists('posix_getgrgid')) {
+            $currentGroupId = posix_getegid();
+
+            $group = posix_getpwuid($currentGroupId);
+            if (!empty($group['name'])) {
+                $group = $group['name'];
+            } else {
+                $group = $currentGroupId;
+            }
+        }
+
         if (empty($group)) {
             $group = 'www-data';
         }
+
         return $user . ':' . $group;
     }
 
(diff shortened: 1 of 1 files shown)


core/Filechecks.php
```diff
@@ -148,9 +148,21 @@ public static function getUserAndGroup()
 
         $group = trim(shell_exec('groups '. $user .' | cut -f3 -d" "'));
 
+        if (empty($currentUser) && function_exists('posix_getegid') && function_exists('posix_getgrgid')) {
+            $currentGroupId = posix_getegid();
+
+            $group = posix_getpwuid($currentGroupId);
+            if (is_array($group)) {
```
diosmosis: Instead of is_array, it would be better to use `!empty($group['name'])` (same for the other similar if statements).

core/Filechecks.php
```diff
@@ -148,9 +148,21 @@ public static function getUserAndGroup()
 
         $group = trim(shell_exec('groups '. $user .' | cut -f3 -d" "'));
 
+        if (empty($currentUser) && function_exists('posix_getegid') && function_exists('posix_getgrgid')) {
```
diosmosis: I think `$currentUser` here is meant to reference `$user`, correct?

diosmosis (changes requested):
Thanks for the first contribution @aragon999 ! I requested some changes, but in general I think this change is fine to include given the new logic is only meant to be used if we cannot find a user/group through existing methods. 👍

core/Filechecks.php
aragon999: No it actually should be `$group` :-) But that was a typo, thank you!

core/Filechecks.php
aragon999: Okay, changed that.

diosmosis:
I merged this, @aragon999, thanks again for the contribution!

PR #16943, opened 2020-12-13
Prevent memory error when using PHP8 and a session error happens

tsteur:
### Description:

Noticed it exceeded 512MB, then 2GB, ...

It was running in an endless loop because 

> PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function Zend_Session_Exception::handleSessionStartError(), 4 passed and exactly 5 expected i

See https://www.php.net/manual/en/function.set-error-handler.php

> Since PHP 8: errcontext was removed, and will no longer be passed to user callbacks.

There might be also other places where we maybe use error handler

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/libs/Zend/Session/Exception.php
+++ b/libs/Zend/Session/Exception.php
@@ -54,7 +54,7 @@ class Zend_Session_Exception extends Zend_Exception
      * @param  string $errstr
      * @return void
      */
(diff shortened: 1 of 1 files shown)


sgiehl (approved):
already had a similar fix in #16897. can confirm that works

PR #16946, opened 2020-12-14
core:archive Fix intersecting period logic so proper archives will archive in parallel.

diosmosis:
### Description:

FYI @tsteur 

Allows segment archives w/ same period to archive together.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/CronArchive/QueueConsumer.php
+++ b/core/CronArchive/QueueConsumer.php
@@ -215,7 +215,7 @@ class QueueConsumer
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


tsteur:
I'll put it for now into 4.1 but we can always change it if/when needed

sgiehl (commented):
Code looks fine. Tried to test that locally by starting multiple archivers but not sure if there is something special to look at. Seems all archivers were working smoothly concurrent, seems nothing was done twice or skipped...

diosmosis:
@tsteur fyi merging this

Issue #16842, opened 2020-12-01
Segments not working on preprocessed files since Matomo 4.0.0


robles-design:
Hi, I'm using Matomo 4.0.3. Since version 4.0.0 segments doesn't work on preprocessed files (see screenshot). If I switch to Visitors Log they are displayed correctly. Maybe it's the same issue like [#16832](https://github.com/matomo-org/matomo/issues/16832). 

Thanks for the great work. I love the new performance metrics.

![Matomo-Screenshot_20201201_031414](https://user-images.githubusercontent.com/25033650/100689579-0572ab00-3385-11eb-9320-7ba45b5d25aa.png)

tsteur:
Thanks for the report @robles-design I think @diosmosis is already working on this one

gusevmaksim:
Matomo 4.0.4 Same problem. Some senments are still processed, but most of them are not.

diosmosis:
@gusevmaksim for one of the segments that isn't being archived, would you be able to post a screenshot of the segment editor w/ that segment loaded? (Or select the row of the segment in the segment table and post the result?)

gusevmaksim:
![Segments](https://user-images.githubusercontent.com/15378153/101292599-96340580-3821-11eb-9a25-340cf4641948.png)

gusevmaksim:
![Segments](https://user-images.githubusercontent.com/15378153/101292644-e1e6af00-3821-11eb-8737-b2ed6ba7028e.png)

robles-design:
For me it works now with 4.0.4 but I had some similar problems after upgrading. Not all segments were created and I got a lot of server errors like Error: SQLSTATE[HY000]: General error: 2006 MySQL server has gone away. After logging in via SSH and running the commands manually all segments were created and the server errors are gone. Maybe this helps.

gusevmaksim:
I updated the database via the console, there were no errors in the process.

diosmosis:
@gusevmaksim I think this might be fixed by one of the existing PRs, but just to be certain, can you run `php console core:archive -vvv` and provide the output? (you can post it here or if you'd like you can email benaka@innocraft.com)

gusevmaksim:
Similar logs for each segment
`DEBUG [2020-12-07 06:55:18] 2111952  No next invalidated archive.
[...]

PR #16947, opened 2020-12-14
Apparently the segment needs to be double encoded when sent in the climulti:request command.

diosmosis:
### Description:

FYI @tsteur 

Currently the segment that gets archived can have a different hash than what is requested in the browser. Encoding the segment before sending it to RequestCommand seems to fix this, but I'm not exactly sure why.

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
@diosmosis is there an issue or comment where there's more information about this? It be unexpected this is suddenly needed or was this an issue in 3.X as well?

diosmosis:
@tsteur it's to solve the problem reported by this user: https://github.com/matomo-org/matomo/issues/16842#issuecomment-742466061

there's a small discussion there. It wasn't required in 3.x, I guess in 3.x the segment was already encoded once in core:archive, then the single `urlencode()` double encoded it properly? It's strange, but the hash is only correct on both UI requests and climulti requests, if it is double encoded here.

tsteur:
👍  I suppose it's not really trivial to add a test here? Or a test would probably archive the data when browser archiving is enabled. This would also explain why some archives aren't showing any data (while some others do) for our web account.

Eg this segment is showing data

image

while such a segment isn't showing data since Matomo 4

image

Can we make sure it won't cause any problems with other segments like that it's not falsely double encoded when using ActionsInVisit>1 for example?

On our demo for matomo website we can reproduce this as well by the looks. I'll patch it there already as well so we can see if the data eventually comes alright. Good find!

diosmosis:
Might be possible to add a test... will see.

PR #16970, opened 2020-12-17
Use correct tracker filename in privacy settings

### LINKED ISSUE #16953: Privacy manager JS tracker writable check uses wrong file
See https://github.com/matomo-org/matomo/blob/4.0.5/plugins/PrivacyManager/Controller.php#L189 where it uses hard coded the path to `/matomo.js`. However, the path may be changed through DI like this:

```

    'Piwik\Plugins\CustomJsTracker\TrackerUpdater' => DI\decorate(function ($previous) {
        $to = PIWIK_DOCUMENT_ROOT . '/';
        $to .= \Piwik\SettingsPiwik::rewriteMiscUserPathWithInstanceId('misc/user/');
        $to .= 'matomo.js';

        $previous->setToFile($to);

        return $previous;
    }),
```

To get the correct path it might be better to use something like 

```
StaticContainer::get('Piwik\Plugins\CustomJsTracker\TrackerUpdater')->getToFile()->getPath();
```

sgiehl:
### Description:

fixes #16953

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/PrivacyManager/Controller.php
+++ b/plugins/PrivacyManager/Controller.php
@@ -19,6 +19,7 @@ use Piwik\Nonce;
 use Piwik\Notification;
 use Piwik\Option;
 use Piwik\Piwik;
+use Piwik\Plugin\Manager;
 use Piwik\Plugins\CustomJsTracker\File;
 use Piwik\Plugins\LanguagesManager\LanguagesManager;
 use Piwik\Scheduler\Scheduler;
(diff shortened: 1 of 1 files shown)


plugins/PrivacyManager/Controller.php
```diff
@@ -187,8 +188,14 @@ public function privacySettings()
         if (Piwik::hasUserSuperUserAccess()) {
             $jsCodeGenerator = new TrackerCodeGenerator();
             $file = new File(PIWIK_DOCUMENT_ROOT . '/' . $jsCodeGenerator->getJsTrackerEndpoint());
+            $filename = $jsCodeGenerator->getJsTrackerEndpoint();
 
-            $view->trackerFileName = $jsCodeGenerator->getJsTrackerEndpoint();
+            if (Manager::getInstance()->isPluginActivated('CustomJsTracker')) {
+                $file = StaticContainer::get('Piwik\Plugins\CustomJsTracker\TrackerUpdater')->getToFile();
```
tsteur, line 194: just seeing we should probably in https://github.com/matomo-org/matomo/blob/4.1.0-b1/plugins/CustomJsTracker/TrackerUpdater.php#L52  call `$jsCodeGenerator->getJsTrackerEndpoint()` as well to get the correct endpoint for historical instances. It would now check for `matomo.js` being writable when many maybe use still `piwik.js`. `TrackerUpdater` itself wouldn't need this change as it tries to update both files anyway. But might be useful here (maybe also in the system check where we check if the tracker file is writable). 

But then there are likely also systems by now where Matomo would check for `piwik.js` but they now use `matomo.js` 😕  I suppose we can also just leave it as it is and assume that if one of them is writable that both are. It might be making more sense to check for `matomo.js`. Old systems who have `piwik.js` writable but not `matomo.js` would then get notified basically that the file will need write permission. 

So all good we don't need to change anything just wanted to put my thinking in writing to figure things out :)

tsteur:
worked for me. A downside was that it said "matomo.js" is not writable when it should have maybe said eg `misc/user/foo/matomo.js` is not writable but then we might also not want to disclose this path + it's an edge case that someone overwrites the path like this (which is basically only us) so all good.

PR #16971, opened 2020-12-17
Limit max execution time of archiving queries to 2 hours by default

### LINKED ISSUE #15241: Limit max execution time for all SQL queries executed during report archiving
To prevent edge cases where archiving (via the `core:archive` crontab) may hang or run for a very long time (for example if a specific combination of [custom reports](https://matomo.org/docs/custom-reports/) dimensions and website size). It does happen regularly (especially for large Matomo instances) that some SQL queries may hang, and to prevent any of these "process hanging/archiving not completing" issues, we'd like to limit the max execution time for all SQL queries executed during report archiving.

This can be done with the MySQL SQL query hint `MAX_EXECUTION_TIME`.

By default we'd set the limit to 2 hours (configurable via INI setting).

The hint should only be added to archiving queries (it's already there in the Live real-time SQL  queries), and make sure to not apply it to for example Matomo upgrades/schema changes.

sgiehl:
### Description:

This will actually only catch those queries using the ArchivingDbAdapter, but guess that should be most of them.

fixes #15241

### Review

* [ ] Functional review done
* [ ] Code review done


core/DataAccess/ArchivingDbAdapter.php
```diff
@@ -31,11 +33,17 @@ class ArchivingDbAdapter
...
+     * @var int
+     */
+    private $maxExecutionTime;
+
     public function __construct($wrapped, Lock $archivingLock = null, LoggerInterface $logger = null)
     {
         $this->wrapped = $wrapped;
         $this->archivingLock = $archivingLock;
         $this->logger = $logger;
+        $this->maxExecutionTime = Config::getInstance()->General['archiving_query_max_execution_time'];
```
diosmosis: Do we need to check for malformed input at all?

config/global.ini.php
```diff
@@ -356,6 +356,11 @@
 ; If you need any other period, or want to ensure one of those is always archived, you can define them here
 archiving_custom_ranges[] =
 
+; If configured, archiving queries will be aborted after the configured amount of seconds. Set it to -1 if the query time
+; should not be limited. Note: This feature requires a recent MySQL version (5.7 or newer). Some MySQL forks like MariaDB
+; might not support this feature which uses the MAX_EXECUTION_TIME hint.
+archiving_query_max_execution_time = 7200
```
diosmosis, line 362: Would it be a good idea to add a test to `DbHelper::addMaxExecutionTimeHintToQuery()` w/ `max execution time == -1`? Currently it's only tested when the value is `0`.

core/DataAccess/ArchivingDbAdapter.php
sgiehl: Don't think we are doing that for other config values that much. But will add a casting to float to ensure it's at least a number.

PR #16972, opened 2020-12-17
Show site selector as text only when only one site is available

### LINKED ISSUE #16789: If only one website exists, remove the site selector from reporting page
In past we had kept it there so people would notice they can track different sites with it. However, the "all websites" screen achieves the same. 

Having less controls makes the UI easier, and also makes it easier when no data has been tracked yet.

We could keep it though if we add a link "Add new site" after "all websites". This link would however only appear to super users so for other users the site selector be still useless. I'd prefer to remove the site selector therefore when there is only one site. 

Here's a screenshot of our demo
image

sgiehl:
### Description:

When only one website is available, the site selector will no longer work as selector. Instead all highlight effects are removed and the dropdown arrow is hidden as well. (Let me know if the UI should be modified even more.)

The site selector will also be shown this way while the list of available sites is loaded.
The list of sites will now be loaded automatically when the site selector is initialized. Before that was done on open only.

fixes #16789

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Plugin/Controller.php
+++ b/core/Plugin/Controller.php
@@ -751,7 +751,11 @@ abstract class Controller
 
         $pluginManager = Plugin\Manager::getInstance();
         $view->relativePluginWebDirs = (object) $pluginManager->getWebRootDirectoriesForCustomPluginDirs();
-        $view->isMultiSitesEnabled = Manager::getInstance()->isPluginActivated('MultiSites');
+        $view->isMultiSitesEnabled = $pluginManager->isPluginActivated('MultiSites');
+        $view->isSingleSite = Access::doAsSuperUser(function() {
+            $allSites = Request::processRequest('SitesManager.getAllSitesId', [], []);
+            return count($allSites) === 1;
+        });
 
         if (isset($this->site) && is_object($this->site) && $this->site instanceof Site) {
             $view->siteName = $this->site->getName();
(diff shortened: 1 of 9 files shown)


tsteur:
Thanks for this @sgiehl it actually looks still like a button now and when I click on it, nothing happens and it feels like a bug. Also it doesn't really help making the UI easier etc.

I would say lets remove the border but it looks not quite right and isn't helping that much either 
image

Ideally we would remove the entire site selector which be fine for > 90% of users (and therefore would make UI better) but I think that's not wanted. So can we instead, when there's only one site, change the "All websites" button and make it a "create new website" button (when user is super user and keep "all websites" for all other users)? Only if it's easily doable though. Otherwise we can also postpone this issue.

mattab:
After all, maybe we could remove the selector completely when there's only one website, as making the UI simpler for most people is probably more valuable indeed..

sgiehl:
> After all, maybe we could remove the selector completely when there's only one website, as making the UI simpler for most people is probably more valuable indeed..

You mean one website at all or one website available for the user?

tsteur:
@sgiehl when only one site exists overall

sgiehl:
Ok. The site selector is now hidden in the top controls when only one website exists overall.
I've not reverted the other changes yet, as they imho improve the behavior when the site selector is used on admin pages.

PR #16974, opened 2020-12-17
Check for instanceIds when checking for currently running archive processes

### LINKED ISSUE #16973: core:archive Detection of archive process already in progress does not check for instance ID
@diosmosis I think there might be a bug which we noticed on cloud as in https://github.com/matomo-org/matomo/blob/4.x-dev/core/CliMulti/RequestParser.php#L63-L96 it’s maybe not checking the instanceId of a matomo anymore when checking if an archive is already running.

Basically, if the current instance defines an instanceId, then Matomo should only consider commands that include the instanceId.

We have at any time eg many “day” archives running but for different --matomo-domain . I wonder if it maybe thinks there’s an archive running but it’s for a different domain.
We used to check for this indirectly here (as the command includes the --matomo-domain) https://github.com/matomo-org/matomo/blob/3.14.1/core/CliMulti.php#L323-L331

tsteur:
### Description:

fix https://github.com/matomo-org/matomo/issues/16973
### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/CliMulti/RequestParser.php
+++ b/core/CliMulti/RequestParser.php
@@ -9,6 +9,8 @@
 
 namespace Piwik\CliMulti;
 
+use Piwik\SettingsPiwik;
+
 class RequestParser
 {
     private $supportsAsync;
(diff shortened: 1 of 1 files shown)


core/CliMulti/RequestParser.php
```diff
@@ -62,9 +64,13 @@ private function filterNonArchivingJobs($commands)
...
     private function getPsLinesWithCliMulti(string $psOutput)
     {
+        $instanceId = SettingsPiwik::getPiwikInstanceId();
         $lines = explode("\n", $psOutput);
         $lines = array_map('trim', $lines);
-        $lines = array_filter($lines, function ($line) {
+        $lines = array_filter($lines, function ($line) use ($instanceId) {
+            if (!empty($instanceId) && strpos($line, 'matomo-domain=' . $instanceId) === false) {
+                return false;
+            }
```
diosmosis, line 73: not sure if this is a problem for cloud, but should we also check if `empty($instanceId) && strpos($line, 'matomo-domain=') !== false`? Will merge regardless after tests run, just thought it might be important.

tsteur: it should be not needed. When an instanceId is configured in the config, then we expect it to be always called with matomo-domain. And if no instanceId is set we assume it doesn't matter.

PR #16992, opened 2020-12-21
core:archive when looking for other climulti processes, also make sure idSite is same

diosmosis:
### Description:

As title. This is in case there are multiple archivers running on the same machine, we can pick up climulti requests for other sites.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/CronArchive/QueueConsumer.php
+++ b/core/CronArchive/QueueConsumer.php
@@ -407,6 +407,7 @@ class QueueConsumer
             if (!empty($archiveBeingProcessed['segment'])
                 && !empty($archiveToProcess['segment'])
                 && $archiveBeingProcessed['segment'] != $archiveToProcess['segment']
+                && urldecode($archiveBeingProcessed['segment']) != $archiveToProcess['segment']
             ) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


diosmosis:
FYI @tsteur

core/CronArchive/QueueConsumer.php
```diff
@@ -397,18 +397,25 @@ private function shouldSkipArchiveBecauseLowerPeriodOrSegmentIsInProgress(array
                 continue;
             }
 
+            if (empty($archiveBeingProcessed['idSite'])
```
sgiehl, line 400: Haven't had a deeper look at the code, but should it actually happen that `idSite` isn't set for an archiving process? If so wouldn't we need to check if `idSite` of both aren't empty here?  🤔

diosmosis: It's a bit confusing, but here `$archiveBeingProcessed` are the parameters for a climulti:request, while `$archiveToProcess` is a row in archive_invalidations. So `idSite` might not be set in a `climulti:request` execution, but `idsite` will always be in `$archiveToProcess`.

tsteur:
Seems tests are passing @diosmosis . is it good to merge?

PR #16994, opened 2020-12-22
In core:invalidate-report-data match segments by ID, name or definition ...

diosmosis:
### Description:

.. and warn when a definition does not match any known segment. When invalidating segments, if it doesn't exactly match the encoding we want, the hash will be wrong and we won't invalidate what we want (and will trigger archiving of a segment we don't want). This change allows better matching and warns in case it doesn't match.

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (approved):

diosmosis:
Noticed this could potentially be a BC break for any automated uses of the command (since the user would have to add `--yes` in that case) so not merging just yet.

plugins/CoreAdminHome/Commands/InvalidateReportData.php
```diff
@@ -54,6 +62,7 @@ protected function configure()
         $this->addOption('dry-run', null, InputOption::VALUE_NONE, 'For tests. Runs the command w/o actually '
             . 'invalidating anything.');
         $this->addOption('plugin', null, InputOption::VALUE_REQUIRED, 'To invalidate data for a specific plugin only.');
+        $this->addOption('yes', null, InputOption::VALUE_NONE, 'Assume yes if an unrecognized segment is given.');
```
mattab: Could we maybe clarify this help text as I don't quite get it? could the parameter name be maybe more clear?

diosmosis: `--ignore-unrecognized-segment`? or `--quiet`/`--force`?

diosmosis: It'd probably be better to just remove the check or make it just print out a warning.

Issue #16955, opened 2020-12-15
Provide more documentation on some archiving behaviour


tsteur:
@diosmosis just looking at some DB performance insights where a lot of new options are showing up in the top queries.

Then looked where they are used and wasn't really understanding what they do and whether they even need to be cached in an option.

* What does the last invalidation time do in https://github.com/matomo-org/matomo/blob/4.x-dev/core/CronArchive.php#L1026-L1035 ? I see it's used in segment archiving in https://github.com/matomo-org/matomo/blob/4.x-dev/core/CronArchive/SegmentArchiving.php#L157-L166 but I don't understand why it compares those dates etc.
* Also maybe the transient cache here is not needed as maybe the Option class itself caches it already? https://github.com/matomo-org/matomo/blob/4.x-dev/core/CronArchive.php#L1026-L1032
* Seeing also this min visit time for site option in https://github.com/matomo-org/matomo/blob/4.x-dev/core/ArchiveProcessor/Loader.php#L427-L442 . Not sure why this one is needed in https://github.com/matomo-org/matomo/blob/4.x-dev/core/ArchiveProcessor/Loader.php#L413-L422 and how that logic works? 
  * Like we could just query the `$this->rawLogDao->hasSiteVisitsBetweenTimeframe` directly to see if there's been any visit during the archive date range?
  * And maybe we wouldn't need to use the lazy cache in https://github.com/matomo-org/matomo/blob/4.x-dev/core/ArchiveProcessor/Loader.php#L429-L433 I wonder? (it should be fast)
  * I understand we get the min recorded visit time for a site and if the end date was before the min visit time, then we don't execute the `hasSiteVisitsBetweenTimeframe` but not sure when that actually happens and whether we can't just execute `hasSiteVisitsBetweenTimeframe` directly? Eg because it's a lazy cache it causes even a lot of inserts in server file sync when caches need to be cleared across servers etc and overall it might cause more problems caching it in this case than fetching it from DB directly I wonder.

Might be good to see if we can make some tweaks or remove some code and also maybe add some inline comments explaining what happens as I tried to understand it and wasn't sure what's happening there.

diosmosis:
Added an answer to the first question here: https://github.com/matomo-org/developer-documentation/pull/402/files#diff-89a35a798a82b03c6fd7dbc734fb0494f586e285c2cb55e3785d50b350061d21R131

diosmosis:
[...]

PR #17005, opened 2020-12-23
Refactor segment re-archiving in past behavior to be on demand

diosmosis:
### Description:

Currently we check if a segment has been created or updated since the last time we invalidated in core:archive, and if so invalidate in the past. This is more complicated and error prone, it is far easier to just invalidate in the past in the SegmentEditor API.

Refs #16955  

### Review

* [ ] Functional review done
* [ ] Code review done


plugins/SegmentEditor/API.php
```diff
@@ -407,4 +433,16 @@ private function getMessageCannotEditSegmentCreatedBySuperUser()
 
         return $message;
     }
+
+    private function reArchiveSegment($segmentInfo)
+    {
+        $definition = $segmentInfo['definition'];
+        $idSite = $segmentInfo['enable_only_idsite'] ?? 'all';
+
+        $idSites = Site::getIdSitesFromIdSitesString($idSite);
```
sgiehl: shouldn't that be done as super user? Otherwise it might only rearchive those sites the current use has access for.

diosmosis: changed

sgiehl:
Wondering if updating that behavior could lead to egde cases on update. E.g. if someone has a cron running only once nightly. After the cron ran a segment is changed and afterwards Matomo is update to that new behavior. The one changed segment won't get invalidated anymore, right? Would we need to do that once in an update maybe? 🤔

diosmosis:
@sgiehl that makes sense, I'll add an update to check as well.

diosmosis:
@sgiehl added the update

core/CronArchive.php
```diff
@@ -228,8 +228,6 @@ public function __construct($processNewSegmentsFrom = null, LoggerInterface $log
         $this->logger = $logger ?: StaticContainer::get('Psr\Log\LoggerInterface');
         $this->formatter = new Formatter();
 
-        $this->processNewSegmentsFrom = $processNewSegmentsFrom ?: StaticContainer::get('ini.General.process_new_segments_from');
```
sgiehl, line 231: Seems that property is no longer in use. Guess it could be removed as property and maybe also as constructor parameter

core/CronArchive/SegmentArchiving.php
```diff
@@ -319,4 +230,18 @@ private function getShouldForceArchiveAllSegments()
     {
         return !Rules::isBrowserTriggerEnabled() && !Rules::isBrowserArchivingAvailableForSegments();
     }
+
+    public function reArchiveSegment($segmentInfo)
+    {
+        $definition = $segmentInfo['definition'];
```
sgiehl, line 230: Should we maybe directly return at this point if the definition is empty for some reason? Guess that shouldn't happen, but would maybe avoid weird issues if it happens nevertheless.

plugins/SegmentEditor/API.php
```diff
@@ -258,6 +274,10 @@ public function update($idSegment, $name, $definition, $idSite = false, $autoArc
 
         $this->getModel()->updateSegment($idSegment, $bind);
 
+        if ($autoArchive && !Rules::isBrowserTriggerEnabled()) {
```
sgiehl, line 277: If I understand that correctly the re-archiving of a segment is only triggered if browser archiving is disabled.
Couldn't that cause an edge case if someone has browser archiving enabled, changes a segment and disables browser archiving afterwards?

sgiehl (changes requested):
Left some comments. Also it seems the branch has unresolved conflicts and some tests are failing...

plugins/SegmentEditor/API.php
diosmosis: Possibly, but I don't think we have to worry about that. If this is an issue, the user could just change the segment again (once to something else, and then back to the original). I don't expect people would really change a segment, then disable browser archiving right after.

core/CronArchive/SegmentArchiving.php
diosmosis: 👍

sgiehl: seems you didn't change that yet

diosmosis:
@sgiehl test should be fixed w/ the next build. the last change increases the amount of invalidated dates for segments, which I think is a bug fix. Before we were checking if the segment was created/edited after the earliest date for rearchiving, but it doesn't make sense to do that when invalidating for any report and not invalidating because of a segment.

sgiehl:
@diosmosis seems there are still some test failures. Also one of my comments is still open. Besides that it would imho be good to merge. Not sure if maybe @tsteur would like to have another look on that.

tsteur:
All good from my side. 👍

diosmosis:
@sgiehl fyi, fixed the tests and in doing so found another bug in the ArchiveInvalidator class. Before we were removing queued invalidations from tracking when invalidating any non-range period. But this meant that if we invalidated just a week, the day wouldn't get re-archived, even if there was a new visit.

Will wait to merge in case you want to do a new review of the change.

core/Archive/ArchiveInvalidator.php
```diff
@@ -559,6 +559,7 @@ public function scheduleReArchiving($idSites, string $pluginName = null, $report
                 'segment' => $segment ? $segment->getString() : null,
             ]));
         } catch (\Throwable $ex) {
+            throw new $ex;
```
sgiehl: Was that added for debugging only? If not it should be moved after the logging I guess.
Also shouldn't that simply be `throw $ex`? The `new` shouldn't be needed

sgiehl (changes requested):

core/Archive/ArchiveInvalidator.php
diosmosis: Yes debugging, will remove.

diosmosis:
@sgiehl updated

sgiehl (approved):

PR #17016, opened 2020-12-24
Add tracking spam prevention plugin as a submodule

tsteur:
### Description:

See https://github.com/matomo-org/plugin-TrackingSpamPrevention for the features that were added.

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
@diosmosis @sgiehl I suppose it is normal that we run all the UI tests of a plugin with a submodule again? Like in https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/44659/ 
image

diosmosis:
@tsteur I think it's because we consider the plugins bundled w/ core to be tested this often (testing compatibility on every matomo change). Otherwise the tests would only run when the plugin changes, even though a change in core could cause problems.

sgiehl:
Remaining test failures seem unrelated. Btw @tsteur did we decide if this plugin will be shipped with Matomo or if it will be available through Marketplace? We might need to adjust the build script maybe 🤔

tsteur:
@sgiehl it should be not included in core but instead it is already on the marketplace. Do we need to adjust anything?

PR #17045, opened 2021-01-05
small archiving invalidation loop optimization

diosmosis:
### Description:

Select less invalidations when we know were archiving segments for a specific period + add total iterations counter to output.

When we're archiving an "All Visits" archive, we know we won't be adding segments to it, so every segment archive we query will be skipped. So it's pointless to query for them all. This PR avoids selecting them so we don't have to go through the loop for them. May provide a visible performance boost in cases where there are many segments for a site.

FYI @tsteur 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/CronArchive/QueueConsumer.php
+++ b/core/CronArchive/QueueConsumer.php
@@ -108,6 +108,13 @@ class QueueConsumer
      */
     private $currentSiteArchivingStartTime;
 
+    /**
+     * The total number of times we queried for another invalidation.
+     *
+     * @var int
+     */
+    private $totalIterations = 0;
+
     public function __construct(LoggerInterface $logger, $websiteIdArchiveList, $countOfProcesses, $pid, Model $model,
                                 SegmentArchiving $segmentArchiving, CronArchive $cronArchive, RequestParser $cliMultiRequestParser,
                                 ArchiveFilter $archiveFilter = null)
@@ -173,7 +180,7 @@ class QueueConsumer
         // get archives to process simultaneously
         $archivesToProcess = [];
         while (count($archivesToProcess) < $this->countOfProcesses) {
-            $invalidatedArchive = $this->getNextInvalidatedArchive($this->idSite, array_keys($invalidationsToExcludeInBatch));
+            $invalidatedArchive = $this->getNextInvalidatedArchive($this->idSite, array_keys($invalidationsToExcludeInBatch), $archivesToProcess);
             if (empty($invalidatedArchive)) {
                 $this->logger->debug("No next invalidated archive.");
                 break;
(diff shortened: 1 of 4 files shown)


core/DataAccess/Model.php
```diff
@@ -758,6 +759,11 @@ public function getNextInvalidatedArchive($idSite, $archivingStartTime, $idInval
             $archivingStartTime,
         ];
 
+        if ($onlySelectSegmentArchives) {
+            $sql .= " AND CHAR_LENGTH(name) > 32 AND date1 = ? AND date2 = ? AND period = ?";
```
sgiehl: Maybe a comment would be useful why the check for length > 32 is used here.

core/CronArchive/QueueConsumer.php
```diff
@@ -329,13 +336,23 @@ private function archiveArrayContainsArchive($archiveArray, $archive)
...
         $iterations = 0;
         while ($iterations < 100) {
             $invalidationsToExclude = array_merge($this->invalidationsToExclude, $extraInvalidationsToIgnore);
 
-            $nextArchive = $this->model->getNextInvalidatedArchive($idSite, $this->currentSiteArchivingStartTime, $invalidationsToExclude);
+            ++$this->totalIterations;
+
+            $isSegmentArchiveInArchivesToProcess = $this->isSegmentArchiveInArchivesToProcess($archivesToProcess);
+            if ($isSegmentArchiveInArchivesToProcess) {
+                // TODO
```
sgiehl: Is there any todo left, or can that simply be removed?

core/DataAccess/Model.php
```diff
@@ -746,7 +746,8 @@ public function isSimilarArchiveInProgress($invalidation)
      * @param int[]|null $idInvalidationsToExclude
```
sgiehl: the doc block doesn't contain the new params

sgiehl (changes requested):

diosmosis:
@sgiehl applied review comments

core/DataAccess/Model.php
```diff
@@ -758,6 +764,13 @@ public function getNextInvalidatedArchive($idSite, $archivingStartTime, $idInval
             $archivingStartTime,
         ];
 
+        if ($onlySelectSegmentArchives) {
+            // we select only segment archiving by looking for archive done flags that have a length > 32. this works since the done flag
+            // for a segment archive will contain its hash
+            $sql .= " AND CHAR_LENGTH(name) > 32 AND date1 = ? AND date2 = ? AND period = ?";
```
sgiehl: Regarding the `32` chars. Wouldn't that also select invalidations for plugins with long names. e.g. `SearchEngineKeywordsPerformance` actually already has 32 chars 🙈 
I'm not deep enough into the code here to be sure if that cause any issues or not.

diosmosis: Yeah this could be an issue then. I guess we also have to do a split or something...

PR #17062, opened 2021-01-08
Display help icons for categories/subcategories 

### LINKED ISSUE #13716: Show a message in each most valuable pages in Matomo to explain why it's important

diosmosis:
### Description:

Adds a new notification type (not made public or advertised) for category/subcategory documentation. Also changes the styling for the in-report help icon, but doesn't make it permanently visible. Might be a good idea to do that I guess, but I wasn't sure.

Fixes #13716 

image

### Review

* [ ] Functional review done
* [x] Code review done


core/Category/Subcategory.php
```diff
@@ -143,4 +150,22 @@ public function getOrder()
...
+    public function getHelp()
+    {
+        return $this->help;
+    }
+
+    /**
+     * Sets the help text for this category.
+     * @param null $help
+     */
+    public function setHelp($help)
```
sgiehl: Might it make sense to add the `setHelp` to the `Category` class as well - for consistency reasons?

diosmosis: It should be in both

diosmosis: Actually I had it in both at first, but the help string can be long so instead I just forced category/subcategory to override getHelp(). I forgot to change this file at the time.

diosmosis:
@mattab @tsteur @sgiehl this is ready for review (code + UI)

plugins/CoreHome/Categories/VisitorsCategory.php
```diff
@@ -9,10 +9,26 @@
...
     protected $order = 5;
     protected $icon = 'icon-reporting-visitors';
+
+    public function getHelp()
+    {
+        $visitsLogUrl = Url::getQueryStringFromParameters([
+            'category' => 'General_Visitors',
+            'subcategory' => 'Live_VisitorLog',
+        ]);
+        $visitsLogUrl = '<a href="?' . htmlspecialchars($visitsLogUrl) . '" rel="noreferrer noopener">';
```
sgiehl: That link doesn't work. Guess it should be more like this:
```
<a href="#" onclick="this.href=broadcast.buildReportingUrl('category=General_Visitors&subcategory=Live_VisitorLog')">
```

sgiehl (changes requested):
Left a minor comment. Everything else seems to work as expected. But might be good if @mattab or @tsteur would also have a look at the colors and the new help texts to ensure they fit

plugins/CoreHome/Categories/VisitorsCategory.php
diosmosis: 👍 thanks, modified

diosmosis:
@mattab / @tsteur can you provide a UI review?

sgiehl (approved):

tsteur:
@diosmosis currently, it shows the help icons for the currently selected ones 
image

The help icon on the main category talks about the reporting page but probably needs to talk about the "visitors report pages" 
image

I don't know if it's easily possible but be great to adjust the titles like below

* How does the "Visitors" reporting section help me? // replace Visitors and visits log with the category/page name..
* How does the "Visits Log" reporting page help me?

If it's too complicated can also not do it.

And generally I wonder if it was better to only show the help icons for the currently hovered reporting category/page. Because we then adjusted the text to not refer to the current reporting page this would be possible. Of course it might be confusing to see a report help for one page in the context of a different page so we could also additionally switch the page and show the help text at once. That's just a thought though because I found it visually bit cluttered when hovering another menu item and the other two question marks are still shown. Eg here I hover "real time map" and I wonder if the other two question marks should disappear and it should be instead only be shown for the real time map. 
image

A problem be though that when clicking on the reporting category there is no page to show so maybe it doesn't work. For sections we could otherwise just show it as part of the current page though maybe. Personally, I would maybe show the help icon only for reporting pages.

Waiting for the thoughts from @mattab

diosmosis:
> If it's too complicated can also not do it.

It's do-able.

> Of course it might be confusing to see a report help for one page in the context of a different page so we could also additionally switch the page and show the help text at once.

Switching the page is a possibility. I can try this and we can revert if it's weird. Though actually, seeing a help icon on hover for every category on the left might be weird too... I guess we can do it on hover of a single menu item. I'll try this.

> Personally, I would maybe show the help icon only for reporting pages.

Makes sense for a first version.

mattab:
The UI looks good overall :+1: (feedback below)

>  Personally, I would maybe show the help icon only for reporting pages.

Agreed :+1:  the help icon can be removed from the top menus and only show for the current reporting page (on hover).

Also a small note but in terms of UX, because the icon is tiny (good for visual effect), it's (very) hard to click on it. 
Could we make the clickable area bigger around icon, so it's easier to hover on it and click it? (if i had to guess, making clickable area 2x wider and 2x higher could work)

Also on hover on the icon, could we maybe show the colored icon? (to show some feedback and hint that clicking might do something). (similar to how it shows blue when hovering the icon next to a report's title)

diosmosis:
@mattab / @tsteur applied all feedback

sgiehl:
btw. should we at some point mention those category classes on https://developer.matomo.org/guides/menus ?

tsteur:
:+1: now only need to wait for the help texts. We could already merge though and them later to avoid more merge conflicts later

sgiehl:
Some tests are failing due to the changes. @diosmosis could you update them so we can get that merged?

diosmosis:
The help texts are almost done I think. We can wait until then and I'll set them in this PR so I don't have to update expected test files twice.

PR #17073, opened 2021-01-11
Make admin menu collapsible

### LINKED ISSUE #16793: Admin menu should be collapsible for admins/super users
For users with admin and super user access:
By default we would want all menus to be collapsed except for the menu that has an active item. Meaning if you are on the "General Settings" page then "System" section should be expanded, all other menus should be collapsed/closed.

On toggle you can open/close a menu. 

If a user has view or write access, then the menu behaves as currently where all submenus are expanded (visible) by default as there are only few menu items available anyway.

image

sgiehl:
### Description:

fixes #16793

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl:
Note: UI tests in submodules may need to be updated before merging

diosmosis (approved):
code looks good, worked locally, good to merge once UI tests pass

Findus23:
It's quite minor, but for me (in Firefox) the animations wobble a bit when the transition ends and it is replaced with the static style. Maybe there is some padding/margin too much.


https://user-images.githubusercontent.com/6266037/104309628-c0d54600-54d2-11eb-89f7-cc4e39c30240.mp4

sgiehl:
seems to be caused by the switch between `overflow:visible` and `overflow:hidden`. Will look for a fix...

PR #17084, opened 2021-01-13
Fix initializing anonymous session

### LINKED ISSUE #17077: Anonymous user settings not working after v4 upgrade
After upgrading a Matomo site from 3.14.1 to 4.1.0, Anonymous users are no longer able to access the dashboard or sites directly.

in v4, accessing Matomo results in a login page despite allowing view access to sites and setting a default anonymous dashboard in Anonymous user settings.

sgiehl:
### Description:

I've debug for quite a while now, but actually couldn't find the change that actually caused the problem.
Even going a few hundreds commits back in history didn't fix the issue locally, so it seems to be there for a while.
Which actually makes it even more strange that it still works on demo.

Nevertheless, for some reason the Authentication does never return a success state for `anonymous` user. So the Auth object holds no login, and the list of available sites stays empty

fixes #17077

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Login/Auth.php
+++ b/plugins/Login/Auth.php
@@ -114,10 +114,6 @@ class Auth implements \Piwik\Auth
     {
         $user = $this->userModel->getUserByTokenAuth($token);
 
-        if (empty($user) && $login === 'anonymous') {
-            $user = $this->userModel->getUser('anonymous');
-        }
-
         if (!empty($user['login']) && $user['login'] === $login) {
             $this->userModel->setTokenAuthWasUsed($token, Date::now()->getDatetime());
             return $this->authenticationSuccess($user);
--- a/plugins/Login/Auth.php
+++ b/plugins/Login/Auth.php
@@ -114,6 +114,10 @@ class Auth implements \Piwik\Auth
     {
         $user = $this->userModel->getUserByTokenAuth($token);
 
+        if (empty($user) && $login === 'anonymous') {
+            $user = $this->userModel->getUser('anonymous');
+        }
+
         if (!empty($user['login']) && $user['login'] === $login) {
             $this->userModel->setTokenAuthWasUsed($token, Date::now()->getDatetime());
             return $this->authenticationSuccess($user);
--- a/plugins/UsersManager/Model.php
+++ b/plugins/UsersManager/Model.php
@@ -427,6 +427,10 @@ class Model
 
     public function getUserByTokenAuth($tokenAuth)
     {
+        if ($tokenAuth === 'anonymous') {
+            return $this->getUser('anonymous');
+        }
+
         $token = $this->getTokenByTokenAuthIfNotExpired($tokenAuth);
         if (!empty($token)) {
             $db = $this->getDb();


sgiehl:
Looking at the test results indicates, that we actually never returned a success state for anonymous before. The tests for that didn't change since 2014. 
Not sure how that could be solved in another way. If there is no successful authentication the session won't be initialized and so the current session will always be counted as outdated.
@diosmosis @tsteur maybe one of you has an idea for that. Don't want to waste more time debugging the code...

plugins/Login/Auth.php
```diff
@@ -114,6 +114,10 @@ private function authenticateWithLoginAndToken($token, $login)
     {
         $user = $this->userModel->getUserByTokenAuth($token);
 
+        if (empty($user) && $login === 'anonymous') {
```
tsteur: that would have regressed with the App specific token auths and I suspect it works on demo because it was installed pre Matomo 4 where we migrated existing token auths. So it would have been broken for new Matomo 4 installs. In tests I think we automatically create such a token maybe. So the alternative could be creating an `anonymous` token auth but not sure if that would have any side effects etc (eg if user deletes that token or user or so)... Does it also need to be added to `authenticateWithToken` (maybe needed for when calling the API)?

tsteur: or maybe a different fix be to add token_auth for anonymous user on install

sgiehl: But my local install was a pre Matomo 4 as well. And there is a migrated token_auth for anonymous user in the database.

tsteur: yes sorry. I was meaning it wouldn't be there maybe for new installs?

diosmosis: the original user said it happened after an update I don't think it is due to being a new install?

I debugged a bit locally and noticed in Login/Auth.php, `$token` is set to `anonymous`. Since this is not a token that is in `user_token_auth` of course it wouldn't find the user. Can you confirm if this is the case for you @sgiehl ?

sgiehl: @diosmosis yes that's the case for me, too.

tsteur: I see. It's actually here for new installs: https://github.com/matomo-org/matomo/blob/4.1.1-b2/core/Db/Schema/Mysql.php#L579 so maybe there was just a migration issue? I've likely deleted my token auths table a few times which explains why I ran into the issue.

sgiehl: Ok. That makes sense. Guess I deleted my token table a few times as well. So what would be the preferred solution?
Readding the token with an update script? Or should we try doing that on each update, to ensure it doesn't get lost somehow? Guess we could also try adding it when someone changes the settings for anonymous user access maybe.

diosmosis: Is there an update file that has the addition of the anonymous user_token_auth entry? I can't seem to find it (by grepping for anonymous in core/Updates).

tsteur: it was done here: https://github.com/matomo-org/matomo/blob/4.0.0-rc4/core/Updates/4.0.0-b1.php#L76 and AFAIK an anonymous user should usually always exist.

No preference on how to fix it. There are a lot of ways... could be also a mix of different ways... easiest be though to ensure maybe in a regular task maybe and when enabling/disabling anonymous that the token exists or not exists etc.

sgiehl: I've slightly modified the PR so `getUserByTokenAuth` for the token `anonymous` will always return the anonymous user. This should fix the problem without the requirement to have an anonymous token at all. And it should also be possible to authenticate in API with token `anonymous`

diosmosis:
code looks ok to me, can't think of a way this could be abused. not sure if @tsteur wants to take a look before merging?

tsteur:
Looks good @diosmosis 👍

PR #17085, opened 2021-01-13
Ensure all file handles to GeoIP databases are closed before replacing the file

### LINKED ISSUE #17023: Starting with php 7.3 on windows OS the geoip2php Update procedure is not working anymore.
I enable geoip2php, download the file and save the settings, 

then after at least one week it disables automatically and when navigating to the geo configuration page it shows this error message in red:

`The configured geolocation provider geoip2php is not available anymore, please configure another one.`

I need then to download the file again and start from beginning.

sgiehl:
### Description:

fixes #17023

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis (approved):
code looks good, I would personally but the cleanup logic in a explicit method to close the file handles rather than rely on php calling __destruct (in case there ends up being a strange case where it doesn't get called), but I think it will work regardless.

mstenz:
let me test the code and give you feedback

mstenz:
it works for me.

sgiehl:
@mstenz thanks for confirming

PR #17097, opened 2021-01-15
Limit GeoIP downloads to certain hosts only

sgiehl:
### Description:

It's currently possible to configure any url for downloading GeoIP databases from. Even though this option can be only changed by a super user and only if the GeoIP admin is enabled, limiting those options to only trusted hosts makes Matomo a little more secure. If someone needs to use the autoupdater to download files from somewhere else he can simply adjust the config.

For now the check is only done when setting the downloader option. Not sure if we should check that directly before download as well. Didn't add that yet, to prevent failures after a Matomo update if someone already has configured something else.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/GeoIp2/GeoIP2AutoUpdater.php
+++ b/plugins/GeoIp2/GeoIP2AutoUpdater.php
@@ -462,7 +462,7 @@ class GeoIP2AutoUpdater extends Task
         $isValidHost = false;
 
         foreach ($validHosts as $validHost) {
-            if (substr($host, -strlen($validHost)) === $validHost) {
+            if (preg_match('/(^|\.)' . preg_quote($validHost) . '$/i', $host)) {
                 $isValidHost = true;
                 break;
             }
--- a/plugins/GeoIp2/GeoIP2AutoUpdater.php
+++ b/plugins/GeoIp2/GeoIP2AutoUpdater.php
@@ -11,6 +11,7 @@ namespace Piwik\Plugins\GeoIp2;
 use Exception;
 use GeoIp2\Database\Reader;
 use Piwik\Common;
+use Piwik\Config;
 use Piwik\Container\StaticContainer;
 use Piwik\Date;
 use Piwik\Http;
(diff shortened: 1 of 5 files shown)


Findus23:
It would be nice if the error message displayed to the user was translatable and would explain a bit more detailed what they need to do to fix it (or link to an FAQ entry).

sgiehl:
@Findus23 good point. I've just improved the messages and made them translatable

plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -443,6 +444,37 @@ public static function setUpdaterOptions($options)
...
+
+        if (empty($schema) || empty($host) || !in_array(Common::mb_strtolower($schema), ['http', 'https'])) {
+            throw new Exception(Piwik::translate('GeoIp2_MalFormedUpdateUrl', '<i>'.$url.'</i>'));
+        }
+
+        $validHosts = Config::getInstance()->General['geolocation_download_from_trusted_hosts'];
+        $isValidHost = false;
+
+        foreach ($validHosts as $validHost) {
+            if (substr($host, -strlen($validHost)) === $validHost) {
```
diosmosis: Would this also return true for `abcdefgdb-ip.com`?

plugins/GeoIp2/tests/Unit/GeoIP2AutoUpdaterTest.php
```diff
@@ -146,4 +151,39 @@ public function testFetchPaidUrlForFullJson()
 
         $this->assertEquals('https://download.db-ip.com/key/ad446bf4cb9a44e4fff3f215deabc710f12f3.mmdb', $determinedUrl);
     }
+
+    /**
+     * @dataProvider getUpdaterUrlOptions
+     */
+    public function testInvalidUpdateOptions($url, $valid) {
```
diosmosis: minor issue: `{` should be on a new line, correct?

plugins/GeoIp2/GeoIP2AutoUpdater.php
diosmosis: Tested, it does.

diosmosis (changes requested):
found one issue, otherwise code looks good

plugins/GeoIp2/GeoIP2AutoUpdater.php
sgiehl: changed that

PR #17101, opened 2021-01-15
allow multiple calls to enableLinkTracking

Findus23:
I am not sure if I understand the issue correctly, but I think it should be possible to call `enableLinkTracking` multiple times.

reported in https://forum.matomo.org/t/enablelinktracking-re-scan-dom-after-content-changes/40269

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
LGTM, would be good to have a test for it in tests/javascript/index.php

tsteur:
build js

tsteur:
👍 indeed it can be called multiple times

Findus23:
It would be great if someone else could create the test as I don't have a great overview over tests/javascript/index.php and not enough time to more closely look into it.

tsteur:
@Findus23 I reckon it's fine it wouldn't need a test and be quite hard to test actually

PR #17130, opened 2021-01-21
Use PerformanceNavigationTiming API if available

### LINKED ISSUE #17102: PerformanceTiming API used by Matomo is deprecated, use PerformanceNavigationTiming if available
I just noticed that the PerformanceTiming API used by Matomo is deprecated and might be one day removed from browsers or only supported for backwards compatibility (*"Be aware that this feature may cease to work at any time."*)

See https://developer.mozilla.org/en-US/docs/Web/API/Performance/timing

The suggested replacement is the [PerformanceNavigationTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming) API which should be used instead

I think `window.performance.getEntriesByType("navigation")[0].toJSON()` gives a quick overview over the data.

Unfortunatly not all modern browsers support it yet, so I guess the replacement has to wait or both have to be supported at the same time.
- https://caniuse.com/mdn-api_performancetiming
- https://caniuse.com/mdn-api_performancenavigationtiming

sgiehl:
### Description:

fixes #17102

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl:
build js

js/piwik.js
```diff
@@ -3466,64 +3466,60 @@ if (typeof window.Matomo !== 'object') {
                 // note: there might be negative values because of browser bugs see https://github.com/matomo-org/matomo/pull/16516 in this case we ignore the values
                 var timings = '';
 
-                if (performanceAlias && performanceAlias.timing && performanceAlias
-                    && performanceAlias.timing.connectEnd && performanceAlias.timing.fetchStart) {
+                var performanceData = (typeof performanceAlias.getEntriesByType === 'function') && performanceAlias.getEntriesByType('navigation') ? performanceAlias.getEntriesByType('navigation')[0] : performanceAlias.timing;
```
diosmosis, line 3470: In the previous code we check that `performanceAlias != null`, do we still need to do this?

sgiehl: makes sense. I'll update the code

sgiehl:
build js

PR #17139, opened 2021-01-24
Fix possible service worker error in Firefox

### LINKED ISSUE #17136: Error in console when trying to communicate with service worker in Firefox with persistant cookies disabled
As reported on the forum: https://forum.matomo.org/t/firefox-84-promiseerror-securityerror-the-operation-is-insecure/40299/6?u=lukas

This can be reproduced by creating a fresh firefox profile (84.0.2 in my case), and opening forum.matomo.org.

> Failed to get service worker registration(s): Storage access is restricted in this context due to user settings or private browsing mode. _application-581e2539c09b2579657fb9679aaeeb510ddf1d491cdaf95d59f45e52559d6843.js:71615:34
>
>Failed to register/update a ServiceWorker for scope ‘https://forum.matomo.org/’: Storage access is restricted in this context due to user settings or private browsing mode. _application-581e2539c09b2579657fb9679aaeeb510ddf1d491cdaf95d59f45e52559d6843.js:71633:34
>
> Failed to register Service Worker: SecurityError: The operation is insecure. _application-581e2539c09b2579657fb9679aaeeb510ddf1d491cdaf95d59f45e52559d6843.js:71635:20
>
> Uncaught (in promise) DOMException: The operation is insecure. _application-581e2539c09b2579657fb9679aaeeb510ddf1d491cdaf95d59f45e52559d6843.js:1

​It seems like even if no service worker (for offline tracking) is set up, just trying to communicate with it fails.

https://github.com/matomo-org/matomo/blob/cc25c84ac6a45255ed2cf68b6e0777ecba50fa26/js/piwik.js#L6969-L6977

I guess the solution is either to make sure Matomo doesn't try to communicate with it when offline tracking is not set up (maybe even add an option to the regular tracking code if one wants to use offline tracking) or catch the promise, ignore it and document it in the offline tracking docs.

tsteur:
### Description:

fix https://github.com/matomo-org/matomo/issues/17136

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
build js

js/piwik.js
```diff
@@ -6971,7 +6971,9 @@ if (typeof window.Matomo !== 'object') {
         addEventListener(windowAlias, 'online', function () {
             if (isDefined(navigatorAlias.serviceWorker) && isDefined(navigatorAlias.serviceWorker.ready)) {
                 navigatorAlias.serviceWorker.ready.then(function(swRegistration) {
-                    return swRegistration.sync.register('matomoSync');
+                    if (swRegistration && swRegistration.sync) {
+                        return swRegistration.sync.register('matomoSync');
+                    }
```
sgiehl, line 6976: I'm not very deep into that serviceWorker stuff, but would there be an alternative that could be used if `sync` is not available? Or is it expected that this simply does not work in that case?

tsteur: @sgiehl it is not 100% required for this to work as it's only triggered when status changes from offline to online and our service worker also has an ability to detect this status change eventually it might just take a bit longer.

sgiehl: ok. then I guess should be good to merge

sgiehl (approved):

brainfoolong:
Hi! Thanks for this fix. Should this be already in `4.2.1`? Currently we still get this errors with Firefox 87.0 and Matomo `4.2.1`. Greetz

brainfoolong:
Nevermind. I cant reproduce this error now, seems to be a cache from old matomo version in the browser of a client. Sorry for this comment :)

PR #17144, opened 2021-01-25
Fix timezone issue when using magic date keywords

### LINKED ISSUE #17095: Widget is showing wrong data from yesterday instead of today
In ref to #16999 

The fix for 4.11 works on URL string with date specified, however with "date=today" it still has the issue of showing yesterday's stats for some dashboard widget if you are in other timezone

**Not working**
index.php?module=CoreHome&action=index&idSite=1&period=day&date=today#?idSite=1&period=day&date=today&segment=&category=Dashboard_Dashboard&subcategory=1

**Working - By selecting today's date URL string -> &date=2021-01-15**
index.php?module=CoreHome&action=index&idSite=1&period=day&date=today#?idSite=1&period=day&date=2021-01-15&segment=&category=Dashboard_Dashboard&subcategory=1

sgiehl:
### Description:

fixes #17095

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Period/Factory.php
+++ b/core/Period/Factory.php
@@ -168,8 +168,11 @@ abstract class Factory
         } else {
             if (!($date instanceof Date)) {
                 if (preg_match('/^(now|today|yesterday|yesterdaySameTime|last[ -]?(?:week|month|year))$/i', $date)) {
-                    // calculate the date taking the current time into account
-                    $date = Date::factory(Date::factory($date)->setTime(date('H:i:s', Date::getNowTimestamp()))->getTimestamp(), $timezone)->getDatetime();
+                    // most magic keywords will automatically set the time to 00:00:00
+                    // we need to set the time manually, so the timezone can be applied correctly
+                    $timestampWithCurrentTime = Date::factory($date)->setTime(date('H:i:s', Date::getNowTimestamp()))->getTimestamp();
+                    // apply the timezone and convert it to a string date
+                    $date = Date::factory($timestampWithCurrentTime, $timezone)->toString('Y-m-d');
                 }
                 $date = Date::factory($date);
             }
--- a/core/Period/Factory.php
+++ b/core/Period/Factory.php
@@ -168,10 +168,10 @@ abstract class Factory
         } else {
             if (!($date instanceof Date)) {
                 if (preg_match('/^(now|today|yesterday|yesterdaySameTime|last[ -]?(?:week|month|year))$/i', $date)) {
-                    $date = Date::factory($date, $timezone);
-                } else {
-                    $date = Date::factory($date);
+                    // calculate the date taking the current time into account
+                    $date = Date::factory(Date::factory($date)->setTime(date('H:i:s', Date::getNowTimestamp()))->getTimestamp(), $timezone)->getDatetime();
                 }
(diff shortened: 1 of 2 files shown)


core/Period/Factory.php
```diff
@@ -168,10 +168,10 @@ public static function makePeriodFromQueryParams($timezone, $period, $date)
         } else {
             if (!($date instanceof Date)) {
                 if (preg_match('/^(now|today|yesterday|yesterdaySameTime|last[ -]?(?:week|month|year))$/i', $date)) {
-                    $date = Date::factory($date, $timezone);
-                } else {
-                    $date = Date::factory($date);
+                    // calculate the date taking the current time into account
+                    $date = Date::factory(Date::factory($date)->setTime(date('H:i:s', Date::getNowTimestamp()))->getTimestamp(), $timezone)->getDatetime();
```
diosmosis: Can we split this to multiple lines so it's clearer what's being done? Otherwise the change looks good.

core/Period/Factory.php
```diff
@@ -168,10 +168,10 @@ public static function makePeriodFromQueryParams($timezone, $period, $date)
...
                 }
+                $date = Date::factory($date);
```
diosmosis, line 173: also can we add a comment or something here saying if `$date` is a `Date` (which can be created above), the time of day is stripped (which is not something someone would know unless they knew Date did that weird thing).

core/Period/Factory.php
sgiehl: done

core/Period/Factory.php
sgiehl: I've changed the code above so it returns a date instead of a datetime. That makes it unneeded to know if date would cut off the time or not.

diosmosis: nvm, this was old behavior. I think you changed it actually.

core/Period/Factory.php
```diff
@@ -168,8 +168,11 @@ public static function makePeriodFromQueryParams($timezone, $period, $date)
         } else {
             if (!($date instanceof Date)) {
                 if (preg_match('/^(now|today|yesterday|yesterdaySameTime|last[ -]?(?:week|month|year))$/i', $date)) {
-                    // calculate the date taking the current time into account
-                    $date = Date::factory(Date::factory($date)->setTime(date('H:i:s', Date::getNowTimestamp()))->getTimestamp(), $timezone)->getDatetime();
+                    // most magic keywords will automatically set the time to 00:00:00
+                    // we need to set the time manually, so the timezone can be applied correctly
+                    $timestampWithCurrentTime = Date::factory($date)->setTime(date('H:i:s', Date::getNowTimestamp()))->getTimestamp();
```
diosmosis: Now that I think about it, I think this is what the `Date::factoryInTimezone` method is meant to achieve. Can you check if this can be used instead? It's meant to retain the time of day before converting to the timezone. If it works, then the code will be simpler and we won't need to do something like this everywhere. Otherwise, feel free to merge.

sgiehl: Good hint. Seems I missed that method for some reason. Using it does the trick as well. At least the tests are still passing.

PR #17148, opened 2021-01-26
use Rules method instead of getting stored segments directly to pre-archive extra segments

diosmosis:
### Description:

Currently we are getting segments from the segment table directly, when it's possible to also add segments to archive to the config and in events, so we want to use the Rules method instead.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/CronArchive.php
+++ b/core/CronArchive.php
@@ -910,12 +910,12 @@ class CronArchive
                     $_forceInvalidateNonexistant);
             }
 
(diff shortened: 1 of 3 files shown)


sgiehl (changes requested):
code looks good, but there are some integration tests failing

diosmosis:
@sgiehl tests should be fixed

sgiehl (approved):

Issue #12586, opened 2018-02-27
Matomo, Tracking URLs, http vs https, visitor log issues


ryrun:
We have changed serveral websites to https. All users will be redirected via 301, all urls on our site was changed to https, too. First problem is, that Matomo's visitor log still shows "http" urls, which cant be possible, because the users will be redirected.

It seems this problem is an old one, someone posted it on the forum, too:
https://forum.matomo.org/t/tracking-von-https-in-url-wird-nur-http-angezeigt/22990 (german post)

The second problem is, that we are currently not 100% sure, when we use the "URLs" configuration for websites, that "https" will be also consididerd like the "non-www"/"www", when we have just used "http://www.example.org". There are some missing informations about this. This need to be clarified more.

mattab:
Hi @ryrun 
Thanks for the feedback.

-> can you please set only the HTTPS URLs in your website (in Admin > websites > Manage > edit). Then are the URLs displayed in the visitor log shown for HTTPS links only?

I can't remember exactly the logic but if your website URLs are defined as HTTPS in Matomo, and your users are viewing https page, then for sure it's expected we show HTTPS links in the visitor log. 
-> Could you please confirm if it's working for you in Matomo 3.6.0 or not?

> The second problem is, that we are currently not 100% sure, when we use the "URLs" configuration for websites, that "https" will be also consididerd like the "non-www"/"www", when we have just used "http://www.example.org". There are some missing informations about this. This need to be clarified more.

Sounds good, we will clarify this in the UI.


### Proposed steps to solve the issue

* Confirm whether linking to HTTPS is partially not working as expected
* Improve the inline help for definining website URLs and clarify that HTTPS and HTTP don't need to be specified like you don't need to specify with and without www.

tsteur:
I can't reproduce it. Just debugged and the protocol is used under which is was tracked originally: https://github.com/matomo-org/matomo/blob/3.9.1/plugins/Actions/VisitorDetails.php#L141

It can happen though if an HTTP url was tracked in the past because of this: https://github.com/matomo-org/matomo/blob/3.9.1/core/Tracker/Model.php#L179

We store the action based on a hash of the `url` but the protocol is ignored in the hash. @mattab explained two months ago or so this was done to not create one `action` entry for each different protocol and to save storage and improve performance big times. If we were to change this, we would probably need to change a lot of reports too as we would likely need to group by more reports (which may lead in inaccurate results for unique metrics etc and lots of other problems).

[...]

Issue #5312, opened 2014-06-06
Piwiks tracks visited https URLs as http


anonymous-matomo-user:
Hi! I'm running Piwik 2.3.0 on Apache HTTP behind an Nginx SSL Proxy. This is working fine and without the issues reported in the forums etc.

I use Piwik to track visits on a few websites that are also secured via SSL. All non-secured requests are redirected by Varnish, which means that no`
http://` adresses should reach Piwik or the tracked website.

Nevertheless, if an SSL-secured website is visited, Piwik shows the URL as normal`
http://` in the visitors log and I can't find the cause for this.

I already enabled $GLOBALS['PIWIK_TRACKER_DEBUG'] to have a look at the sent data:

```
action_name:vanished-site-title
idsite:3
rec:1
r:167255
h:16
m:5
s:28
url:https://www.vanished.site/page/
urlref:https://www.vanished.site/
_id:fa1926e74503fbed
_idts:1402063420
_idvc:1
_idn:0
_refts:0
_viewts:1402063420
pdf:1
qt:1
realp:0
wma:1
dir:0
fla:1
java:1
gears:0
ag:0
cookie:1
res:1920x1080
gt_ms:1336
```

This clearly shows that`
https://` was used to visit the page, but the visitor log presents the URL as`
http://www.vanished.site/page/`.

Is this probably a bug or am I missing something here?
Do you need further details?

Thank you!
Keywords: 2.3.0

mattab:
This issue was also reported in the forums: http://forum.piwik.org/read.php?2,126198

> Piwik incorrectly displays http instead of https
> 
>  I was stumped about this for a few weeks.
> 
> We have re-direct enabled for our site. Anybody that comes into an http page will be sent to https.
Piwik was displaying visitors visiting http versions of pages. It listed http about 50% of the time along with https. We checked our code and our htaccess files and everything seemed fine.
> 
> Today, it so happened that an employee was browsing our site and we could see them being fed http versions of the site. Going to their actual browser and checking their browsing history showed NO http pages in their browser only https. Piwik incorrectly displayed http in the Live Visitors view and on the dashboard.

Harest:
Any update on this issue ? Until now i was using self-signed SSL cert and i couldn't force https. I've switched to a CA delivering free SSL certs and i now force https on 1 site.  
I'm using nginx to redirect permanently (301) http visitors as long as an HSTS header.  
```server {
        listen   80;
        server_name  mysite.com www.mysite.com;
[...]

PR #17151, opened 2021-01-27
Use https for urls in visitor details if host is defined with https in site

sgiehl:
### Description:

refs #5312, #12586

### Review

* [ ] Functional review done
* [ ] Code review done


plugins/Actions/VisitorDetails.php
```diff
@@ -160,6 +161,12 @@ public function extendActionDetails(&$action, $nextAction, $visitorDetails)
             unset($action['url_prefix']);
         }
 
+        if (array_key_exists('url', $action) && preg_match('/^http:\/\/([^\/]+)\//i', $action['url'], $host)) {
```
tsteur: haven't tested it but might be better using a simple `strpos($url, 'http://')===0` and then use `parse_url` to get the host? Regexes are always tricky to understand and hard to say how well it works etc. Could be also good to add a test if possible. Performance wise it's probably fine as it would be only for HTTP urls.

sgiehl: Thought about that solution before, but wanted to avoid two operations for that. But guess most people should have https anyway and so only a strpos would be executed, which should for sure be faster....

plugins/Actions/VisitorDetails.php
```diff
@@ -160,6 +161,14 @@ public function extendActionDetails(&$action, $nextAction, $visitorDetails)
             unset($action['url_prefix']);
         }
 
+        if (array_key_exists('url', $action) && strpos($action['url'], 'http://') === 0) {
+            $host = parse_url($action['url'], PHP_URL_HOST);
+
+            if ($host && $this->shouldUseHttpsHost($visitorDetails['idSite'], $host)) {
+                $action['url'] = 'https://' . substr($action['url'], 7);
```
tsteur: Might be good for better readability to not hard code `7` but instead use like `strlen('http://')`.

Also we might want to use Common::mb_substr here?

plugins/Actions/VisitorDetails.php
```diff
@@ -252,6 +261,29 @@ public function extendActionDetails(&$action, $nextAction, $visitorDetails)
         unset($action['idlink_va']);
     }
 
+    private function shouldUseHttpsHost($idSite, $host)
+    {
+        static $siteUrlCache = [];
```
tsteur: is there any chance to not use static here as we can't reset it in tests? We could use the transient cache or so instead? Or might be even fine to just use a regular property as there's likely only one instance per request anyway maybe?

tsteur (dismissed):
Worked. Left 2 comments. Can you also already prepare an FAQ for this? Was wondering if we need to do the same for Actions.getPageUrls report for example maybe?

plugins/Actions/VisitorDetails.php
sgiehl: Will add a comment, so it's clear

plugins/Actions/VisitorDetails.php
sgiehl: transient cache should work as well...

sgiehl:
Afaik `Actions.getPageUrls` and reports like that are generating the urls while archiving. We could change that there or manipulate the datatable before it is returned from API.

Regarding FAQ: Not sure which question it actually should answer.

tsteur:
Regarding the FAQ: I meant something like: `How do I fix links in visitor log going to HTTP instead of HTTPS?` or something like this. Or "How do I change the links to my pages to use HTTPS in Matomo?"

> We could change that there or manipulate the datatable before it is returned from API.

Sounds good. I think it be good to have consistent behaviour there if possible.

core/Tracker/PageUrl.php
```diff
@@ -335,6 +338,44 @@ public static function reconstructNormalizedUrl($url, $prefixId)
...
+     * @param $idSite
+     * @param $host
+     * @return false|mixed
+     * @throws \Exception
+     */
+    public static function shouldUseHttpsHost($idSite, $host)
+    {
+        $cache = Cache::getTransientCache();
+
+        $cacheKeySiteUrls = sprintf('siteurls-%s', $idSite);
```
diosmosis: This is minor and not strictly necessary but we could use `CacheId::siteAware` here for consistency.

plugins/Actions/DataTable/Filter/Actions.php
```diff
@@ -66,6 +67,14 @@ public function filter($table)
                     $label = $row->getColumn('label');
                     if ($url) {
                         $row->setMetadata('segmentValue', urlencode($url));
+
+                        if ($site && strpos($url, 'http://') === 0) {
+                            $host = parse_url($url, PHP_URL_HOST);
+
+                            if ($host && PageUrl::shouldUseHttpsHost($site->getId(), $host)) {
+                                $row->setMetadata('url', 'https://' . Common::mb_substr($url, 7 /* = strlen('http://') */));
```
diosmosis, line 75: Is there a reason for using `mb_substr` here but normal `strtolower` above?

plugins/Actions/VisitorDetails.php
```diff
@@ -160,6 +162,14 @@ public function extendActionDetails(&$action, $nextAction, $visitorDetails)
             unset($action['url_prefix']);
         }
 
+        if (array_key_exists('url', $action) && strpos($action['url'], 'http://') === 0) {
+            $host = parse_url($action['url'], PHP_URL_HOST);
+
+            if ($host && PageUrl::shouldUseHttpsHost($visitorDetails['idSite'], $host)) {
+                $action['url'] = 'https://' . Common::mb_substr($action['url'], 7 /* = strlen('http://') */);
+            }
+        }
```
diosmosis, line 171: Can we make this and the above change visible in API tests? Eg, setting the site URL of one host in an existing test to https so part of the output changes.

diosmosis (commented):
Left a couple comments. Tested locally and it works 👍

plugins/Actions/VisitorDetails.php
sgiehl: will add/change some tests

plugins/Actions/DataTable/Filter/Actions.php
sgiehl: will change that. even though it actually doesn't make a difference. See #16754

sgiehl:
@diosmosis  I've applied some changes. Let me know if it's good to merge now.

PR #17168, opened 2021-01-30
add visitfrequency archiver to archive dependent archives

### LINKED ISSUE #17123: Graph of returning visitors doesn't show returning visitors when having the segment pre-configured in config file
I've recognized this a few days after the Update to 4.1.1, but it must have started with the latest beta before the update to 4.1.1 (see screenshot).

Behaviour > Engagement > Returning visitors
matomo-engagement-returning-visitors

diosmosis:
### Description:

Fixes #17123 

@tsteur found the fix. The cause is since we call `CoreAdminHome.archiveReports` now instead of `API.get`, the dependent archive isn't implicitly created. Adding an archiver for VisitFrequency fixes this, though I'm not sure if we want to maintain the old behavior.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Updates/4.1.2-b1.php
+++ b/core/Updates/4.1.2-b1.php
@@ -1,0 +1,62 @@
+<?php
+/**
+ * Matomo - free/libre analytics platform
+ *
+ * @link https://matomo.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ *
+ */
+
+namespace Piwik\Updates;
+
+use Piwik\Archive\ArchiveInvalidator;
+use Piwik\ArchiveProcessor\Rules;
+use Piwik\Container\StaticContainer;
+use Piwik\Date;
+use Piwik\Updater;
+use Piwik\Updates as PiwikUpdates;
+use Piwik\Updater\Migration\Factory as MigrationFactory;
+
+class Updates_4_1_2_b1 extends PiwikUpdates
+{
+    /**
+     * @var MigrationFactory
+     */
+    private $migration;
+
+    public function __construct(MigrationFactory $factory)
+    {
+        $this->migration = $factory;
+    }
+
+    public function getMigrations(Updater $updater)
+    {
+        $migrations = [];
+
+        if (!Rules::isBrowserTriggerEnabled()) {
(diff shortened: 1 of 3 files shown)


tsteur:
sounds good for now 👍 . I don't think it affects any other `*.get` methods so should be fine IMO

diosmosis:
@tsteur will also add an update to automatically invalidate, unless you think that's a bad idea.

tsteur:
@diosmosis I'm assuming this would mean we rearchive all the data from the last few months not just this report? cc @mattab

diosmosis:
@tsteur 

> I'm assuming this would mean we rearchive all the data from the last few months not just this report?

I think we can just re-archive visitfrequency from start of matomo 4 upgrade to now. We just need to trigger the VisitFrequency archiver which will process dependent archives. So we shouldn't need to do every archive.

tsteur:
👍 sounds good

diosmosis:
added the update file

core/Updates/4.1.2-b1.php
```diff
@@ -0,0 +1,62 @@
...
+        $migrations = [];
+
+        if (!Rules::isBrowserTriggerEnabled()) {
+            $dateOfMatomo4Release = Date::factory('2020-11-23');
+
+            $cmdStr = $this->getInvalidateCommand($dateOfMatomo4Release);
+
+            $migrations[] = new Updater\Migration\Custom(function () use ($dateOfMatomo4Release) {
+                $invalidator = StaticContainer::get(ArchiveInvalidator::class);
+                $invalidator->scheduleReArchiving('all', 'VisitFrequency', null, $dateOfMatomo4Release);
```
sgiehl, line 43: Will that work when only browser archiving is used?

diosmosis: it shouldn't be necessary for browser archiving, since VisitFrequency.get will be requested and will trigger archiving in that case. this bug should only affect core:archive setups

sgiehl: Makes sense. Looks good to merge then

sgiehl (approved):

PR #17180, opened 2021-02-02
Send 204 response for bulk requests to avoid possible CORB issue

### LINKED ISSUE #17159: Tracking requests might not be loaded because of CORB
In the browser developer console I see a warning linking to https://www.chromestatus.com/feature/5629709824032768 anyone seen this before?
See https://www.chromium.org/Home/chromium-security/corb-for-developers for more information.

HTTP with 204 seems to work but 200 blocked. `send_image=0` is used so it should have sent an HTTP 204 header. Actually... this output happens because of bulk requests. The request with HTTP 200 was sent as a bulk request in this example.

image

response headers:

image

The console warning reads `Cross-Origin Read Blocking (CORB) blocked cross-origin response XYZ with MIME type application/json. See https://www.chromestatus.com/feature/5629709824032768 for more details.`

For a URL where this can be reproduced feel free to ping me privately. I'm not sure if the tracking request will be executed or if just the response is not loaded. 

Goal of this issue is to make sure such requests are tracked and the warning is no longer shown.

sgiehl:
### Description:

fixes #17159

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl:
build js

diosmosis:
given this could potentially be a bc break for some users, should we mention this change in the developer changelog?

sgiehl:
guess that makes sense. will add it to the changelog

diosmosis:
Actually I misread the PR, it doesn't need an entry in the developer changelog. I'll remove it and merge.

PR #17187, opened 2021-02-03
Creates issue templates

sgiehl:
### Description:

Might help to improve issue quality and help our users to use the correct space/repo for their requests

### Review

* [ ] Functional review done
* [ ] Code review done


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,49 @@
+---
+name: "Bug report"
+about: Something isn't working correctly with Matomo
+labels: Bug
+
+---
+
+<!-- Please provide a short summary of the issue in the *Title* above -->
+
+<!-- Important: Please contact the Matomo community forum for questions -->
```
diosmosis: Maybe a link to the forum here (and other types)?

diosmosis:
Generally looks good to me, maybe @tsteur wants to take a look?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,49 @@
+---
+name: "Bug report"
+about: Something isn't working correctly with Matomo
+labels: Bug
```
Findus23: As said, I'd use a different tag for a non-triaged bugreport and keep Bug for things that are confirmed to be broken.

Findus23 (changes requested):

.github/ISSUE_TEMPLATE/bug_report.md
sgiehl: How about `Potential Bug`? Once someone of us was able to reproduce we can change that to `bug`, or add other labels like `Can't reproduce (yet)`

tsteur: `Can't reproduce (yet)` we used slightly differently before. Like we might want to set this label for a `Potential bug`

tsteur:
Looks good to me but be good to indeed change comment around the `bug` template

.github/ISSUE_TEMPLATE/bug_report.md
sgiehl: So shall we add a label automatically for bugs? Imho that might be helpful so it's directly possible to filter our issues better.
We could maybe also rename the current `Bug` label to something like `Bug (accepted)` or `Bug (reproduces)`, which we later would add as soon someone was able to reproduce. And then we would have a new label `Bug` that would be automatically added for new Bugs. We could also consider adding that new label for all issues currently having a Bug related label....

Findus23: I originally intended `Can't reproduce (yet)` for the issues, that a lot of people are reporting all the time, but no one of the team could yet reproduce. Which reminds me that we should maybe also add descriptions to all labels to more precisely define them.

tsteur: what is the difference between accepted and reproduces? Also not sure what reproduces mean? Does it mean we can reproduce it? Can a bug be accept and reproduced? Sorry not quite clear. I reckon something like `Bug (not confirmed)` could work? And then we change it simply to `Bug` if it is a bug? Or if we can't reproduce it etc we could add `Can't reproduce yet`? 

Be great to have clear descriptions for all of them eventually @Findus23 👍

sgiehl: Maybe we should rethink all our current labels at some time and and make them a bit clearer. I currently can't see a big difference between `Potential Bug` and `Can't reproduce (yet).`
It would be awesome if you are able to see the current "state" of an issue based on the labels
For bugs we could for example have `Potential Bug`, which would be automatically added when a bug report is created. Once someone had a look, that could be changed to `Bug` and/or `Regression`, or a `Can't reproduce (yet)` could be added.
Adding descriptions for all labels would help as well...

tsteur: `Potential bug` be basically a newly created issue by a user. Adding `Can't reproduce yet` would basically mean we tried to reproduce it, but we can't and we need more users having this issue or more details or ...  If we are able to reproduce the bug we would change the label to `bug`. We can also have a label `Not reproducible bug` instead and from `Potential bug` it goes to either `bug` or `not reproducible bug`. As always we can also change things any time which is always good to keep in mind and could just go with something eventually and then see how it works and tweak over time.

sgiehl:
```suggestion
labels: Potential Bug
```

sgiehl:
@tsteur @mattab Shall we merge this one then? If you want to have another look how creating an issue would look like afterwards: https://github.com/sgiehl/matomo/issues/new/choose

tsteur:
👍  great idea @sgiehl and let's see over time how it goes and tweak if needed

mattab:
Just seen it today, it looks awesome!

PR #17203, opened 2021-02-08
Allow including host name in Actions URL reports instead of merging paths regardless of URL.

### LINKED ISSUE #14104: Subdomains paths in Pages / other reports
We just rolled out Matomo (which is awesome). We noticed the same thing as [this user](https://forum.matomo.org/t/sub-domains-in-reports-pages-enty-pages-exit-pages/19186):

> The problem: In the “pages” report (and also the “entry pages” and “exit pages” reports) all URLs are mixed up independent on which sub domain they are running. First this is confusing, and second it seems to create wrong data! All index pages of the root and the sub domains are listed together as page URL “/index”. When I open the page overlay, the index page of the sub domain with the highest number of pageviews is displayed. But the data shown left hand in the “main metrics” section is the sum of all sub domains (pageviews, visits etc.). When I change the page overlay’s URL manually to another sub domain’s index page, the “main metrics” data stays the same.

So for example, we have URLs like this:

`https://www.newgrounds.com/art/view/diives`

In Matomo > Behavior > Pages it shows art > /index which is actually this URL:

`https://diives.newgrounds.com/art/`

Which is incorrect and unexpected.

We're already running this in our tracking code:

`_paq.push(["setDocumentTitle", document.domain + "/" + document.title]);`

But that doesn't seem to apply to Pages and other reports.

I searched around and I don't think this user put in a ticket so I thought I'd create one.

diosmosis:
### Description:

Fixes #14104 

CC @tsteur, @mattab 

Seemed like a simple change so I created a poc. Note: after creating it, I thought it might be good to use a measurable setting instead of an INI config (though the INI config could set the default). Either way, a small change.

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
Will close this one as it's not high priority and issue is not scheduled.

uschindler:
This pull request looks like it would fix many complaints about not seeing the subdomain/domain in the pages report. Why was it closed? I would wish to have this in 5.0.

Issue #10881, opened 2016-11-18
Error page served with HTTP 200 status


strugee:
Upon visiting my Piwik homepage today I got a page that said "an error occurred" followed by "Piwik couldn't write to some directories (running as user 'www-data')." and a list of commands to try.

I have monitoring in place that checks whether HTTP services respond with 200 OK, aka a success status code. Obviously Piwik is not OK in this scenario, and should return 500 Internal Server Error instead.

mattab:
Thanks for the suggestion @strugee definitely agree! This is similar to https://github.com/piwik/piwik/pull/10623

andyjdavis:
I have put up a PR containing a possible solution for this issue. https://github.com/matomo-org/matomo/pull/17206

diosmosis:
Fixed by  #17206

strugee:
Thanks to you both for your hard work :-)

PR #17206, opened 2021-02-08
dieWithHtmlErrorPage() should default to http 500

andyjdavis:
### Description:

This is a possible solution for this issue. https://github.com/matomo-org/matomo/issues/10881
"Error page served with HTTP 200 status"

Prior to this PR dieWithHtmlErrorPage() defaults to returning a http 200 if the caught exception doesn't have a code set on it. This PR instead causes dieWithHtmlErrorPage() to default to returning a http 500 unless another code is set.

This is the URL I have been using. Note the invalid date parameter. the error page is identical however its http status code is now 500 instead of 200.
http://localhost/matomo/index.php?module=CoreHome&action=index&idSite=1&period=day&date=INVALID

I am still working through getting the tests working locally. I get a few failures even before I make any changes. Despite that I wanted to put this PR up now just to see if this approach is fundamentally a problem in some way :)

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/ExceptionHandler.php
+++ b/core/ExceptionHandler.php
@@ -75,6 +75,8 @@ class ExceptionHandler
             && $exception->getCode() > 0
         ) {
             http_response_code($exception->getCode());
+        } else {
+            http_response_code(500);
         }
 
         self::logException($exception);


diosmosis:
Hi @andyjdavis, I merged this, thanks for the great contribution!

PR #17215, opened 2021-02-12
Archiver: do not create done flags like "done."

diosmosis:
### Description:

Noticed calling `getDoneStringFlagFor` w/ `$plugin = false` when `pluginOnly=1` can result in done flags like `done.`.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/ArchiveProcessor/Rules.php
+++ b/core/ArchiveProcessor/Rules.php
@@ -51,7 +51,9 @@ class Rules
      */
     public static function getDoneStringFlagFor(array $idSites, $segment, $periodLabel, $plugin)
     {
-        if (!self::shouldProcessReportsAllPlugins($idSites, $segment, $periodLabel)) {
+        if (!empty($plugin)
+            && !self::shouldProcessReportsAllPlugins($idSites, $segment, $periodLabel)
+        ) {
             return self::getDoneFlagArchiveContainsOnePlugin($segment, $plugin);
         }
         return self::getDoneFlagArchiveContainsAllPlugins($segment);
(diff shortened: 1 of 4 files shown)


tsteur:
@diosmosis would this be causing reading archives that were created with this flag? or would that basically not have worked already anyway?

diosmosis:
@tsteur if there were already archives made w/ `done.`, it might be an issue... we could have an update to change those I guess (an update `where name = 'done.'`?

tsteur:
@diosmosis could be done. Just trying to understand: Would it be otherwise an issue to having flags with `done.`? But all good to add such an update script and fixing it

diosmosis:
> Would it be otherwise an issue to having flags with done.? 

For leaving them, I think the only issue is if they're no longer queried and `done` archives aren't present. Now it might be an issue in that we might not be querying them properly? It's hard to tell w/ this issue exactly where it causes problems.

diosmosis:
Note: I did see them created in prod when making plugin specific archives, maybe nowhere else.

tsteur:
Also quickly another question: Would it cause issues if there are afterwards maybe two archives with a `done` flag? And they might contain different reports? I know it's a bit hard to say what would happen. I could imagine that some archive may be deleted or reports would be only read from one archive?

diosmosis:
> Would it cause issues if there are afterwards maybe two archives with a done flag?

I'm not sure what you mean? After the update, having two archives w/ `done` should be fine since that's what we want (all archives would have the `done` flag or done w/ plugin/segment.

tsteur:
All good then 👍

diosmosis:
merging this, fyi @tsteur

PR #17216, opened 2021-02-12
fix a couple bugs in archive querying when browser archiving is disabled

diosmosis:
### Description:

fixes the following bugs:

* do not do authorized to archive check when browser archiving is disabled in `ArchiveSelector::getArchiveIds()`. that function is only ever called when browser archiving is disabled so the request will never be authorized to archive, and we want to look at the latest invalidated archive if present.
* when selecting archives we want to select all partial archives not just the latest for a plugin. I think those changes were put in to get a test to pass, but since this function didn't have tests didn't notice the problem. This is needed for report specific invalidations.
* in the browser archiving / trigger=archivephp archiving path, the partial archive detection was not pulling partial archives newer than the latest known all plugins archive. fixed and with a test change that shows the error.

CC @tsteur 

### Review

* [ ] Functional review done
* [ ] Code review done


tests/PHPUnit/Integration/DataAccess/ArchiveSelectorTest.php
```diff
@@ -30,24 +30,85 @@ protected static function configureFixture($fixture)
...
+                [1],
+                [
+                    ['day', '2020-03-01'],
+                    ['week', '2020-03-02'],
+                ],
+                '',
+                [],
+                [
+                    'done' => [
+                        '2020-03-01,2020-03-01' => [1],
```
sgiehl, line 75: Maybe I didn't understand the code correctly, but shouldn't that contain the biggest archive id available for that period. Looking at the data above, that seems to be `3` and not `1`

diosmosis: The value for `3` is `DONE_ERROR` so we don't use it.

sgiehl: ah ok. that makes sense

diosmosis:
@tsteur can you take a look at this PR?

tsteur:
looks good @diosmosis

PR #17225, opened 2021-02-16
recreate all webserver config on updates

Findus23:
It seems like the full `createFilesForSecurity` that also creates the files for IIS is only run on install and some updates and not on the normal update.
```bash
➜  ~/public_html/matomo git:(4.x-dev) ✗ rg "createFilesForSecurity"
core/Updates/4.0.0-b1.php
261:        ServerFilesGenerator::createFilesForSecurity();

core/Updates/3.13.4-b1.php
21:        ServerFilesGenerator::createFilesForSecurity();

core/Updates/3.0.0-b4.php
56:        ServerFilesGenerator::createFilesForSecurity();

core/Updates/2.16.3-b1.php
20:        ServerFilesGenerator::createFilesForSecurity();

core/Updates/3.0.1-b1.php
22:        ServerFilesGenerator::createFilesForSecurity();

plugins/Installation/Controller.php
313:        ServerFilesGenerator::createFilesForSecurity();

plugins/Installation/ServerFilesGenerator.php
17:    public static function createFilesForSecurity()
```

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Updater.php
+++ b/core/Updater.php
@@ -331,7 +331,7 @@ class Updater
         $this->markComponentSuccessfullyUpdated($componentName, $updatedVersion);
 
         $this->executeListenerHook('onComponentUpdateFinished', array($componentName, $updatedVersion, $warningMessages));
-        ServerFilesGenerator::createHtAccessFiles();
+        ServerFilesGenerator::createFilesForSecurity();
         return $warningMessages;
     }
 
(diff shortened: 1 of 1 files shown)


sgiehl:
@Findus23  When doing that in updater, guess we could remove it from the single update scripts. Wouldn't make sense to do it multiple times when updating from an older version.

Findus23:
I'm not sure if modifying historic update scripts is worth it if the worst case is creating the same file twice after each other.

sgiehl:
Don't have a strong opinion on that. That just came up to my mind when reviewing... It would only remove some unneeded operations. But should also be fine to keep them 🤷

tsteur:
Be fine to keep them as the operation is quickly done and not happening too often

diosmosis (approved):

Issue #16990, opened 2020-12-21
When showing the Export URL, clarify that users must use their app token (instead of displaying a token_auth that only works in their session)


mattab:
After Matomo 4 upgrade, we got many reports that the "Export URL" feature is now confusing and people can't figure it out. 
What's happening is that people expect to be able to copy paste the “Export URL”, and send it to others, but it doesn’t work anymore to use this URL. (since Matomo 4, the token_auth only works for that particular user session).

To improve the situation, here is an idea that we came up with after a short discussion: 

When clicking the “Show export URL”, it indicates that the user wants to get the URL and use it. So when people click on it, we could instead

1.  automatically remove the token_auth and replace it by eg. `&token_auth=ENTER_YOUR_TOKEN_AUTH_HERE`

2. show the information in the  tooltip as a visible warning instead (and tweak the message a bit such as below, or better):

> Note: To use the generated export URL, you will need to specify an app token auth. You can configure these tokens in [Admin -> Security -> Auths Tokens.](open the link in a new tab?). Replace `ENTER_YOUR_TOKEN_AUTH_HERE` in the Export URL by your Auth token.

This would help people achieve their goals of sharing the URL, and reduce confusion.

tsteur:
> Note: To use the generated export URL, you will need to specify an app token auth. You can configure these tokens in [Admin -> Security -> Auths Tokens.](open the link in a new tab?). Replace ENTER_YOUR_TOKEN_AUTH_HERE in the Export URL by your Auth token.

It should also say to never ever share this URL with the replaced token with anyone else.

sgiehl:
should have been fixed with #17227

PR #17227, opened 2021-02-17
16990 export url token

flamisz:
### Description:

fix for #16990 

### Review

* [ ] Functional review done
* [ ] Code review done


plugins/CoreHome/angularjs/report-export/reportexport.directive.js
```diff
@@ -185,7 +185,7 @@
                         }
                     }
 
-                    exportUrlParams.token_auth = piwik.token_auth;
+                    exportUrlParams.token_auth = _pk_translate('CoreHome_EnterYourToken');
```
tsteur: btw @flamisz when clicking on the export button then I think we'd still want the correct token auth to be used. 
image

Only in the textarea we would want to show the `ENTER_YOUR_TOKEN_AUTH_HERE`.

When clicking on the export button currently it says 
image

vs it should show the exported data.

Also additionally, the parameter `&force_api_session=1` should not be shown in the text area 
image

Sorry this wasn't mentioned in the issue probably. Otherwise it would compare the token against a "session" token but not against a proper token from the configured auth tokens in the security page AFAIK. So this parameter would need to be removed in the textarea but kept when clicking on the "export" button.

flamisz: Yeah sure, you're right, the button needs the token. I fix that and delete the `&force_api_session=1` parameter from the url in the textarea (only there).

tests/UI/specs/ViewDataTable_spec.js
```diff
@@ -200,6 +200,12 @@ describe("ViewDataTableTest", function () { // TODO: should remove Test suffix f
         expect(await dialog.screenshot()).to.matchImage('export_options');
     });
 
+    it("should display the ENTER_YOUR_TOKEN_AUTH_HERE text in the export url", async function () {
+        await page.click('.toggle-export-url');
+
+        expect(await dialog.screenshot()).to.matchImage('export_options_2');
```
sgiehl, line 211: might be good to have the `dialog = await page.$('.ui-dialog');`  within the new test as well.
It might not be needed as it's actually already set in the test before. But imho we should try to avoid using global variables here if possible.

plugins/CoreHome/lang/en.json
```diff
@@ -104,7 +104,9 @@
         "RowLimit": "Row limit",
         "CustomLimit": "Custom limit",
         "ExportFormat": "Export format",
-        "ExportTooltip": "Note: The generated export URL will only work in the current browser session. If you want to use it somewhere else you need to use an app specific token. You can configure these tokens in Admin -> Security -> Token Auths.",
+        "ExportTooltip": "Note: The generated export URL will only work in the current browser session. If you want to use it somewhere else you need to use an app specific token. You can configure these tokens in [Admin -> Security -> Token Auths]. Replace ENTER_YOUR_TOKEN_AUTH_HERE in the Export URL by your Auth token. Warning: never share the URL with the real token with anyone else.",
+        "ExportTooltipWithLink": "Note: To use the generated export URL, you will need to specify an app token auth. You can configure these tokens in %1$s[Admin -> Security -> Auths Tokens]%2$s. Replace ENTER_YOUR_TOKEN_AUTH_HERE in the Export URL by your Auth token. Warning: never share the URL with the real token with anyone else.",
+        "EnterYourToken": "ENTER_YOUR_TOKEN_AUTH_HERE",
```
sgiehl: Not sure if we should have that one translatable at all.

plugins/CoreHome/lang/en.json
```diff
@@ -104,7 +104,9 @@
         "RowLimit": "Row limit",
         "CustomLimit": "Custom limit",
         "ExportFormat": "Export format",
-        "ExportTooltip": "Note: The generated export URL will only work in the current browser session. If you want to use it somewhere else you need to use an app specific token. You can configure these tokens in Admin -> Security -> Token Auths.",
+        "ExportTooltip": "Note: The generated export URL will only work in the current browser session. If you want to use it somewhere else you need to use an app specific token. You can configure these tokens in [Admin -> Security -> Token Auths]. Replace ENTER_YOUR_TOKEN_AUTH_HERE in the Export URL by your Auth token. Warning: never share the URL with the real token with anyone else.",
+        "ExportTooltipWithLink": "Note: To use the generated export URL, you will need to specify an app token auth. You can configure these tokens in %1$s[Admin -> Security -> Auths Tokens]%2$s. Replace ENTER_YOUR_TOKEN_AUTH_HERE in the Export URL by your Auth token. Warning: never share the URL with the real token with anyone else.",
```
sgiehl:
```suggestion
        "ExportTooltipWithLink": "Note: To use the generated export URL, you will need to specify an app token auth. You can configure these tokens in %1$s[Admin -> Security -> Auths Tokens]%2$s. Replace %3$s in the Export URL by your Auth token. Warning: Never share the URL with the real token with anyone else.",
```

We shouldn't have `ENTER_YOUR_TOKEN_AUTH_HERE` in this string as well, but instead should use a placeholder to inject it. Otherwise it could happen that it is translated in different ways.

sgiehl (changes requested):

plugins/CoreHome/lang/en.json
flamisz: So it means it will be in English for everyone? 
I can just "hard code" it in the js file than?

and of course, the test would fail on different languages... 🤦‍♂️

tests/UI/specs/ViewDataTable_spec.js
flamisz: yeah I was surprised these tests are actually built on top of each other. I'll see how it works exactly.

plugins/CoreHome/lang/en.json
sgiehl: Yes. English for everyone whould be fine in this case, as it's a placeholder that should be replaced nevertheless.
Also tests are always running in english, unless a test sets another language on purpose.

tests/UI/specs/ViewDataTable_spec.js
sgiehl: We actually have a lot ui tests that kind of "depend" on each other. Or at least depend on the previous result. 
Having them more independent might be a nice to have in some cases, but actually the tests would need a lot more time to run, as each test needs to load a new url and redo several clicks, that actually might have been done in the test before already.
So for this test we wouldn't need to redo everything that was done in the test before.

plugins/CoreHome/angularjs/report-export/reportexport.popover.html
```diff
@@ -61,7 +61,8 @@
     </div>
 
     <div class="col l12" ng-show="showUrl">
-        <textarea piwik-select-on-focus readonly class="exportFullUrl" title="{{ 'CoreHome_ExportTooltip'|translate }}">{{ getExportLink() }}</textarea>
+        <textarea piwik-select-on-focus readonly class="exportFullUrl">{{ getExportLinkWithoutToken() }}</textarea>
+        <div class="tooltip" ng-bind-html="'CoreHome_ExportTooltipWithLink'|translate:'<a target=_blank href=\'/index.php?module=UsersManager&action=userSecurity\'>':'</a>':'ENTER_YOUR_TOKEN_AUTH_HERE'"></div>
```
diosmosis: This URL starting w/ `/index.php` will break on subdirectory installs, correct?

tests/UI/specs/ViewDataTable_spec.js
```diff
@@ -196,10 +196,21 @@ describe("ViewDataTableTest", function () { // TODO: should remove Test suffix f
         await page.click('.activateExportSelection');
         await page.waitFor('#reportExport .btn');
 
-        dialog = await page.$('.ui-dialog');
+        var dialog = await page.$('.ui-dialog');
```
diosmosis: just a note, but the UI tests are run in puppeteer so you can use es6+ in them.

diosmosis (changes requested):
found one possible issue

plugins/CoreHome/angularjs/report-export/reportexport.popover.html
flamisz: good catch, thanks. I'll check it.

flamisz: hi @diosmosis I fixed this, using directory-relative URL instead of origin-relative.

tests/UI/specs/ViewDataTable_spec.js
flamisz: @diosmosis I modified to `let`

plugins/CoreHome/angularjs/report-export/reportexport.popover.html
```diff
@@ -62,7 +62,7 @@
 
     <div class="col l12" ng-show="showUrl">
         <textarea piwik-select-on-focus readonly class="exportFullUrl">{{ getExportLinkWithoutToken() }}</textarea>
-        <div class="tooltip" ng-bind-html="'CoreHome_ExportTooltipWithLink'|translate:'<a target=_blank href=\'/index.php?module=UsersManager&action=userSecurity\'>':'</a>':'ENTER_YOUR_TOKEN_AUTH_HERE'"></div>
+        <div class="tooltip" ng-bind-html="'CoreHome_ExportTooltipWithLink'|translate:'<a target=_blank href=\'./index.php?module=UsersManager&action=userSecurity\'>':'</a>':'ENTER_YOUR_TOKEN_AUTH_HERE'"></div>
```
diosmosis: you could also have `href='index.php?...` or just `href='?module=...` (the `./` is unnecessary)

flamisz: thanks, done

plugins/CoreHome/angularjs/report-export/reportexport.directive.js
```diff
@@ -30,7 +30,11 @@
 
                 scope.showUrl = false;
 
-                scope.getExportLink = function() {
+                scope.getExportLinkWithoutToken = function() {
+                    return scope.getExportLink(false);
+                }
+
+                scope.getExportLink = function(withToken = true) {
```
diosmosis: can't use es6+ here unfortunately (this is also causing the angularjs tests to fail). after this is fixed, i'll merge (sorry i keep finding issues one at a time)

flamisz: yes, I figured yesterday this on the other PR, I will fix it quickly here as well

Issue #13976, opened 2019-01-18
Segments are not deleted when User is deleted


pfleckenbw:
When I delete a user with username "TestUsername" and later create a new user with the same username, the second user get all segments defined of the first user.
IMHO. the delete-user-action should delete also the users segments. At least his private segments. All public segments should be transfered to an Admin of the website.

### Reproduction:
1. Create an User named "TestUsername"
2. Give him Permissions to WebsiteA
3. Sign in with  "TestUsername"
4. Create any segment
5. Delete the User "TestUsername"
6. Create another User named "TestUsername" (with any eMail-address)
7. Give him Permissions to WebsiteA
8. Sign in with "TestUsername"
9. The previously created Segement is already there :-(

### Environment
* Matomo-Versuib: 3.7.0
* MySQL-Version: 5.1.73-log
* PHP-Version: 5.6.39 (Yes, it's outdated, but I can't believe, that the PHP-Version is the reason for this issue

mattab:
Thanks for the report @pfleckenbw! This bug is still active it seems. 

Liking your suggestions below :+1: 

> the delete-user-action should delete also the users segments. At least his private segments. All public segments should be transferred to a [Super User.]

mattab:
fixed in  #17231

PR #17231, opened 2021-02-18
13976 segments not deleted

flamisz:
### Description:

fix for #13976 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CoreAdminHome/CoreAdminHome.php
+++ b/plugins/CoreAdminHome/CoreAdminHome.php
@@ -8,11 +8,10 @@
  */
 namespace Piwik\Plugins\CoreAdminHome;
 
+use Piwik\API\Request;
 use Piwik\Piwik;
 use Piwik\ProxyHttp;
-use Piwik\API\Request;
 use Piwik\Plugins\CoreHome\SystemSummary;
-use Piwik\Plugins\SegmentEditor\SegmentEditor;
 use Piwik\Settings\Storage\Backend\PluginSettingsTable;
 
 /**
@@ -52,8 +51,6 @@ class CoreAdminHome extends \Piwik\Plugin
     public function cleanupUser($userLogin)
     {
         PluginSettingsTable::removeAllUserSettingsForUser($userLogin);
-
-        SegmentEditor::transferAllUserSegmentsToSuperUser($userLogin);
     }
 
     public function getStylesheetFiles(&$stylesheets)
--- a/plugins/CoreAdminHome/CoreAdminHome.php
+++ b/plugins/CoreAdminHome/CoreAdminHome.php
@@ -8,10 +8,11 @@
  */
 namespace Piwik\Plugins\CoreAdminHome;
 
-use Piwik\API\Request;
 use Piwik\Piwik;
 use Piwik\ProxyHttp;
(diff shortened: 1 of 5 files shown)


flamisz:
Hi all, this PR is not "finished" yet, I'd like to add an extra test, any suggestion is appreciated.
The test that is already included is testing if the `transferAllUserSegmentsToSuperUser` is working properly, but one extra test is needed to make sure we actually call this function when a user is deleted.

It's happening in `plugins/CoreAdminHome/CoreAdminHome.php` on a `UsersManager.deleteUser` event. What's the best way to test this?

Thanks

tsteur:
Hi @flamisz if you wanted to test that it's actually called you could basically first check there are segments for that user like `$this->assertNotEmpty($segments->getsegments...)` delete that user (eg using UsersManager API) and check that no segments are defined for that user anymore like `$this->assertEmpty($segment->...)`. It wouldn't need a detailed test or so since the method itself is already tested.

plugins/CoreAdminHome/CoreAdminHome.php
```diff
@@ -51,6 +52,8 @@ public function addSystemSummaryItems(&$systemSummary)
     public function cleanupUser($userLogin)
     {
         PluginSettingsTable::removeAllUserSettingsForUser($userLogin);
+
+        SegmentEditor::transferAllUserSegmentsToSuperUser($userLogin);
```
tsteur: btw usually we would call this method in `Piwik\Plugins\SegmentEditor\SegmentEditor` class by listening to the `UsersManager.deleteUser` event see https://developer.matomo.org/guides/events . There we would then call the method to delete all segments for the deleted user. Basically the same way it's done in this class. It's not needed to do it that way but usually best practice to contain all the logic that belongs to the plugin in the actual plugin.

flamisz: I like that, more clean solution. I will modify it.

diosmosis:
Code looks good, but it appears to be causing another integration test to fail.

flamisz:
> Code looks good, but it appears to be causing another integration test to fail.

@diosmosis the test had errors in the `tearDown()`. I fixed those. it should be fine, will see soon :)

(of course the error was caused by my pr)

diosmosis (approved):
looks good, will merge tomorrow after a bit of time's passed since the 4.2.0 release

plugins/SegmentEditor/SegmentEditor.php
```diff
@@ -358,4 +361,38 @@ public static function getAllSegmentsForSite($idSite)
         }
         return $segments;
     }
+
+    public function onDeleteUser($userLogin)
+    {
+        $this->transferAllUserSegmentsToSuperUser($userLogin);
+    }
+
+    public static function transferAllUserSegmentsToSuperUser($userLogin)
```
diosmosis: Sorry, noticed one other issue: is there a reason this method is `static`? If not let me know and I'll change it before merging.

flamisz: definitely NOT, I'm not sure why it is static, sorry. thanks for changing it @diosmosis

PR #17233, opened 2021-02-18
Prevent possible error when accessing an index

tsteur:
### Description:

see https://wordpress.org/support/topic/undefined-index-piwik_pro_ads_enabled/

I think this should fix the issue. It's technically only a workaround. Not sure why this would show up as an error. I will try to find out more to see if there's a general issue there.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/ProfessionalServices/Advertising.php
+++ b/core/ProfessionalServices/Advertising.php
@@ -117,7 +117,10 @@ class Advertising
      */
     public static function isAdsEnabledInConfig($configGeneralSection)
     {
-        $oldSettingValue = @$configGeneralSection['piwik_pro_ads_enabled'];
+        $oldSettingValue = false;
+        if (isset($configGeneralSection['piwik_pro_ads_enabled'])) {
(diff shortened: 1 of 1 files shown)


core/ProfessionalServices/Advertising.php
```diff
@@ -117,7 +117,10 @@ private function getCampaignParametersForPromoUrl($campaignName, $campaignMedium
      */
     public static function isAdsEnabledInConfig($configGeneralSection)
     {
-        $oldSettingValue = @$configGeneralSection['piwik_pro_ads_enabled'];
+        $oldSettingValue = false;
+        if (isset($configGeneralSection['piwik_pro_ads_enabled'])) {
+            $oldSettingValue = @$configGeneralSection['piwik_pro_ads_enabled'];
```
diosmosis, line 122: Do we still need the `@` here?

tsteur: I suppose technically not but might also not hurt?

diosmosis: seems redundant, but ok

PR #17234, opened 2021-02-19
Ask users to refer us to other friends

### LINKED ISSUE #16803: Ask users to refer us to other friends
Not sure into which plugin to put this best. However, it needs to be disabled by WhiteLabel (eg through some event).

We'd basically show a block in the bottom similar to on our demo.matomo.cloud (might need to go incognito to see it)

image
(be great to give the social icons some white background or something so they stand out)
"Refer Matomo" is basically a "mailto" link that add's following subject:

subject:
```
Why I no longer use Google Analytics
```

message body:
```
Did you know Google uses data for “own purposes”? That means Google owns your data and is using it to monetise their advertising platforms.

This is why I don’t use Google Analytics for analysing the performance of my website. 

Instead I choose Matomo, an ethical alternative to Google Analytics that gives me 100% data ownership and protects the data of my website visitors. 

I’m sharing this message in the hope that you too will take back the power from Google and get complete ownership of your own data.

[Check out Matomo](https://matomo.org/google-owns-your-data?pk_campaign=refer&pk_kwd=onpremise)
```

The user can then customise the email in their email program and enter email addresses etc.

For social sharing we would use this text if possible:
```
Did you know Google is using your data for “own purposes”? This means Google owns your data and uses it to monetise their advertising platforms.

If you’re using Google Analytics, stay in control by switching to an ethical alternative like Matomo now! 

https://matomo.org/google-owns-your-data?pk_campaign=share&pk_kwd=onpremise
```

can we have buttons for twitter, facebook, linkedin and mastodon? We would not actually load their JS etc but use the share links like https://www.facebook.com/sharer.php and https://twitter.com/intent/tweet and https://www.linkedin.com/shareArticle/ . Not sure it's possible for mastodon?

After clicking on any of the links and the desired action was initiated could we maybe show a thank you modal dialog saying like `Thank you for spreading the word and creating a safer web`?

* When someone clicks on the `X` icon on the right then we don't show it for another 6 months. 
* Let's maybe in the beginning also only show it to super users (end users are often not that passionate about this topic)
* Once someone clicked on "refer matomo" or on the social media links we don't show it again for now

flamisz:
### Description:

fixes #16803 

### Review

* [x] Functional review done
* [x] Code review done


flamisz:
I haven't find a way to share on `mastodon` with a link, so that button is hidden on the ui. 
The FB share currently not working, as `matomo.org` is on spam-list at FB.

I fixed a small bug with the feedback popup: it called the `remindMeLater` method even when the user clicked on the never remind me button, so actually the user was reminded again in 90 days.

Findus23:
@flamisz If you want to make it a bit more complex, you could add a input field for people to enter their Mastodon URL and then redirect them to `https://mastodon.example/share?text=the%20text`
grafik

Findus23 (changes requested):
One minor thing, before I forget:
All texts should probably be translatable.

flamisz:
> One minor thing, before I forget:
> All texts should probably be translatable.

@Findus23 yes, that's a good point. for sure it should be. I will do that

plugins/Feedback/Controller.php
```diff
@@ -37,12 +35,32 @@ function index()
...
+        $nextReminder = Common::getRequestVar('nextReminder');
+
+        if ($nextReminder !== Feedback::NEVER_REMIND_ME_AGAIN) {
+            $nextReminder = Date::now()->getStartOfDay()->addDay($nextReminder)->toString('Y-m-d');
+        }
+
+        $feedbackReminder = new FeedbackReminder();
+        $feedbackReminder->setUserOption($nextReminder);
+
+        return json_encode(['Next reminder date: ' . $nextReminder]);
```
diosmosis, line 50: You might need to also call `Piwik\DataTable\Renderer\Json::sendHeaderJSON()` here and below because I don't think controller methods do that by default.

plugins/Feedback/angularjs/refer-banner/refer-banner.controller.js
```diff
@@ -0,0 +1,87 @@
...
+                var paramString = new URLSearchParams(params);
+
+                return base + paramString.toString();
+            }
+
+            if (type === 'facebook') {
+                var base = 'https://www.facebook.com/sharer.php?';
+
+                var params = { 't': text, 'u': url};
+                var paramString = new URLSearchParams(params);
```
diosmosis: `URLSearchParams` has no support in IE: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams and currently we still support IE11. Can you do this another way?

plugins/Feedback/Feedback.php
```diff
@@ -64,17 +68,73 @@ public function getClientSideTranslationKeys(&$translationKeys)
...
+        }
+
+        if (!Piwik::hasUserSuperUserAccess()) {
+            return false;
+        }
+
+        if ($this->isDisabledInTestMode()) {
+            return false;
+        }
+
```
diosmosis, line 122: Here, can we add an `@internal` event like `Feedback.showReferBanner` that takes a bool reference:

```
$shouldShowReferBanner = true;

/**
 * @internal
 */
Piwik::postEvent('Feedback.showReferBanner', [&$shouldShowReferBanner]);

if (!$shouldShowReferBanner) {
    return false;
}
```

This way, the WhiteLabel plugin can hook into it and make sure it's not shown at all when the plugin is active.

diosmosis (changes requested):
Some potential UX issues (cc @mattab / @tsteur):
- I noticed if you click one link the bottom area disappears. So if a user clicks on twitter, it goes to twitter.com, then they go back and want to click on another link (say linkedin), they're no longer able to. It doesn't seem like that big of a deal given the scope of the change, but it's something I noticed.

@mattab / @tsteur can you look at and approve the ux and copy?

banner at the bottom of the page:

image

thank you popup after clicking refer:

image

tsteur:
> I noticed if you click one link the bottom area disappears. So if a user clicks on twitter, it goes to twitter.com, then they go back and want to click on another link (say linkedin), they're no longer able to. It doesn't seem like that big of a deal given the scope of the change, but it's something I noticed.

No big preference here. I guess it could be good indeed to keep it open and let the user close it manually 👍  No big preference though.

mattab:
Sounds slightly better to "keep it open and let the user close it manually" :+1:

flamisz:
Hi @diosmosis, I made the changes according to your comments. 
@mattab now the banner won't close until the user will close it.

Should I make a PR in the Whitelabel plugin, so we handle this new event and can disable showing this banner?

mattab:
> Should I make a PR in the Whitelabel plugin, so we handle this new event and can disable showing this banner?

@flamisz Yes, it should be always disabled in White Label plugin, good that you thought about this one :+1:

PR #17243, opened 2021-02-22
rename testcookie to make it clear it is created by Matomo

Findus23:
inspired by https://forum.matomo.org/t/is-cookie-test-with-value-testvalue-set-by-matomo/40775/2

The cookie's name doesn't start with _pk, so it isn't obvious it is created by Matomo.

I also named the cookie `_pk_testcookie_domain`, so it isn't confused with this testcookie:

https://github.com/matomo-org/matomo/blob/bd7e099e46de3e10afc069624b713de4bd8df162/js/piwik.js#L3043-L3044

### Review

* [x] Functional review done
* [x] Code review done


Findus23:
build js

flamisz (approved):
LGTM

js/piwik.js
```diff
@@ -3381,8 +3381,9 @@ if (typeof window.Matomo !== 'object') {
 
             function isPossibleToSetCookieOnDomain(domainToTest)
             {
+                var testCookieName = configCookieNamePrefix + 'testcookie_domain';
                 var valueToSet = 'testvalue';
-                setCookie('test', valueToSet, 10000, null, domainToTest, configCookieIsSecure, configCookieSameSite);
+                setCookie(testCookieName, valueToSet, 10000, null, domainToTest, configCookieIsSecure, configCookieSameSite);
 
                 if (getCookie('test') === valueToSet) {
                     deleteCookie('test', null, domainToTest);
```
sgiehl: shouldn't that use the new `testCookieName` variable as well? 🤔

sgiehl (changes requested):

js/piwik.js
Findus23: You are of course right :+1:

Findus23:
build js

sgiehl (approved):

PR #17244, opened 2021-02-22
possibly avoid infinite looping if group_concat cuts off result

diosmosis:
### Description:

As title, test forthcoming.

### Review

* [ ] Functional review done
* [ ] Code review done


core/DataAccess/Model.php
```diff
@@ -52,15 +52,18 @@ public function __construct(LoggerInterface $logger = null)
      *
      * @param string $archiveTable
      * @param array $idSites
+     * @param bool $setMaxLen for tests only
```
sgiehl:
```suggestion
     * @param bool $setGroupContentMaxLen for tests only
```

core/DataAccess/Model.php
```diff
@@ -82,7 +85,7 @@ public function getInvalidatedArchiveIdsSafeToDelete($archiveTable)
             while (!empty($duplicateArchives)) {
                 $pair = $duplicateArchives[0];
                 if (strpos($pair, '.') === false) {
-                    continue; // see below
+                    break;
```
sgiehl, line 88: Maybe you could add a comment here, that this actually only happens when group_concat cut off some parts of the result. 
Also it can actually happen that `$pair` ends with an `.`. That would actually also be a cut off value, and should be handled as such. That also applies for the check below that prints the info message for it.

sgiehl (changes requested):

diosmosis:
@sgiehl applied pr feedback

core/DataAccess/Model.php
```diff
@@ -97,7 +100,7 @@ public function getInvalidatedArchiveIdsSafeToDelete($archiveTable)
             // if there is more than one archive, the older invalidated ones can be deleted
             if (!empty($duplicateArchives)) {
```
sgiehl, line 101: Does this additional check/loop still makes sense? Assuming we have some values in `$duplicateArchives` the `while` above should remove all values from that variable, unless there is a broken one in the end. But even with the broken one, the loop below would only do some logging an break.

diosmosis: @sgiehl yes it's still needed, the first loop removes the latest usable archive entries from the array. This can be DONE_PARTIAL entries as well as the latest DONE_OK entry. The second loop just collects the remaining values for purging. There might be a way to merge them but I suspect it would get confusing.

sgiehl: Ah ok, makes sense. I think I got it now 🙈

sgiehl (approved):

tsteur:
Looks good @diosmosis is the test https://travis-ci.com/github/matomo-org/matomo/jobs/485720146#L778 failing because of this PR? Feel free to merge otherwise

diosmosis:
@tsteur was looking at this yesterday. Something changed w/ the tracking, now only one visit gets tracked. Looking into why.

diosmosis:
build failure was random. I was able to reproduce it locally yesterday, but not today...

PR #17249, opened 2021-02-22
Correct multisite image url

### LINKED ISSUE #17248: All websites dashboard images may show as 404
Matomo 4.2

"All websites" dashboard (module=MultiSites&action=index)

The image is not displayed
grafik

404 error received
grafik

Findus23:
fixes #17248

### DIFF
--- a/plugins/MultiSites/angularjs/site/site.controller.js
+++ b/plugins/MultiSites/angularjs/site/site.controller.js
@@ -40,6 +40,7 @@
                     metric = 'revenue';
                     break;
             }
+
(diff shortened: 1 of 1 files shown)


diosmosis:
Using `?module=` is actually done throughout matomo. This seems more like a server configuration issue to me? Given matomo requests should almost always go to index.php by default.

Findus23:
While I agree that this is a server configuration issue, I didn't see any other Matomo requests that don't use index.php in the URL.
Both URLs linked to like `index.php?module=CoreAdminHome&action=home` and API requests like `index.php?date=today&module=UserCountryMap` always seem to include the index.php part in the URL.

diosmosis:
Those, I'm guessing, go through a single code path (ie, piwik-api.js) so would be uniform. I'm sure there are links in matomo, however, that link to `?...`. I guess we can still change this one though.

PR #17252, opened 2021-02-23
Fix type mismatch in tracker

gtbl2012:
### Description:

In current piwik tracker, the 'setUserId' using 'isNumberOrHasLength' to check the 'userId'.
```javascript
this.setUserId = function (userId) {
    if (isNumberOrHasLength(userId)) {
        configUserId = userId;
    }
};
```
while the 'getRequest' function using '(configUserId && configUserId.length)' to check the 'userId'.
```javascript
((configUserId && configUserId.length) ? '&uid=' + encodeWrapper(configUserId) : '')
```

Which causing user can set a userId in number and verify it by 'getUserId' func, but the userId cannot be reported.

This PR is to solve this confusing thing.

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
Hi @gtbl2012, thanks for the great first contribution! This is indeed a problem, thanks for the fix!

PR #17255, opened 2021-02-23
set secure and samesite for cookie on delete

### LINKED ISSUE #16637: core/Cookie.php:Cookie->delete() don't set secure and SameSite.
Hi,

We use AjaxOptOut-Plugin instead of the OptOut-Iframe: https://plugins.matomo.org/AjaxOptOut?piwikversion=3.

Matomo-Host and Website-Host don't share the same Domainname and use https.

The `IgnoreCookie::setIgnoreCookie();` function works fine.

But `IgnoreCookie::getIgnoreCookie()->delete();` fails to delete the "piwik_ignore".

Browser (like Chrome 86) mention that SameSite isn't set.

After some investigation we found the cause in /core/Cookie.php

```
    public function delete()
    {
        $this->setP3PHeader();
        $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain);
    }
```

Here is no handling, if SSL or Samesite is needed. We quick-patched it by setting the "delete" the ignore-cookie two times.

```
    public function delete()
    {
        $this->setP3PHeader();
        $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain);
        $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain, TRUE, FALSE, 'None');
    }
```

Now the first cookie gets ignored by Chrome, but the second cookie is accepted.

I hope you have a better idea to solve this problem.

i don't know if any issue in the cookie-meta-issue #14904 already cover that (i haven't found that scenario there).

Greetings
René

flamisz:
### Description:

fix: #16637 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Cookie.php
+++ b/core/Cookie.php
@@ -176,7 +176,13 @@ class Cookie
     public function delete()
     {
         $this->setP3PHeader();
-        $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain);
+
+        if (ProxyHttp::isHttps()) {
+            $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain, TRUE, FALSE, 'None');
+        } else {
+            $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain);
+        }
+
         $this->clear();
     }
 
@@ -292,7 +298,7 @@ class Cookie
             } elseif (!is_numeric($value)) {
                 $value = base64_encode($value);
             }
-            $cookieStrArr[] = "$name=$value"; 
+            $cookieStrArr[] = "$name=$value";
         }
 
         return implode(self::VALUE_SEPARATOR, $cookieStrArr);
--- a/core/Cookie.php
+++ b/core/Cookie.php
@@ -177,11 +177,8 @@ class Cookie
     {
         $this->setP3PHeader();
 
-        if (ProxyHttp::isHttps()) {
-            $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain, TRUE, FALSE, 'None');
-        } else {
-            $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain);
-        }
+        $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain);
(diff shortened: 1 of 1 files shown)


flamisz:
I reproduced the bug and what I found is Firefox is happy to delete the cookie without the extra settings, but Chrome doesn't. 

I checked the [setIgnoreCookie()](https://github.com/matomo-org/matomo/blob/f8cdf6ad9738a05a450b4f7cfc863a532b88b9be/core/Tracker/IgnoreCookie.php#L71-L90) function, and replicated the same if-else during the delete.

The question is: what happens if the cookie is already there, but it was made without `https` and then they changed to `https`?

Other solution could be to send two setcookie headers, one without and one with the extra settings. This is what the original issue's description mentions as a solution.

tsteur:
@flamisz I reckon it be good to have the additional core there too if it doesn't cause any issue. This way it might be more likely that any older cookie will be deleted too (or if the original cookie was set maybe on http and then now it deletes it on https etc). If there's no harm, it would be great to have.

flamisz:
@tsteur something like this?
```php
if (ProxyHttp::isHttps()) {
    $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain, TRUE, FALSE, 'None');
    $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain);
} else {
    $this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain);
}
```

or just simple (without if)
```php
$this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain, TRUE, FALSE, 'None');
$this->setCookie($this->name, 'deleted', time() - 31536001, $this->path, $this->domain);
```

tsteur:
would probably simply do the second one without if but I'm not too much into the details. Just seems like it could maybe cover more cases where it would otherwise maybe not delete the cookie.

flamisz:
I tried to replicate the case, when we have a non secure cookie on a secure site. It looks like Chrome doesn't care and will delete anyway with the secure and samesite settings, and Firefox actually deleted by itself the non-secure cookie on secure domain immediately.

flamisz:
I like the simple solution, let's go with that one. Multiple set-cookie in the header is possible and it doesn't hurt anything.

diosmosis:
tested locally, seems to work. (also tested in safari & ie11)

PR #17258, opened 2021-02-24
not supported browser exception

### LINKED ISSUE #16776: When accessing UI with unsupported browser throw error and suggest to use other browser
When IE 10 is used, we may want to show an exception that it is not supported instead of seeing a broken screen. This can also have security benefits since some features like `noreferrer` might not work in IE10.

flamisz:
### Description:

fixes #16776 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -22,6 +22,7 @@ use Piwik\Http\Router;
 use Piwik\Plugins\CoreAdminHome\CustomLogo;
 use Piwik\Session\SessionAuth;
 use Piwik\Session\SessionInitializer;
+use Piwik\SupportedBrowser;
 use Psr\Log\LoggerInterface;
 
 /**
@@ -154,7 +155,7 @@ class FrontController extends Singleton
         if (self::$enableDispatch === false) {
             return;
         }
-        
+
         $filter = new Router();
         $redirection = $filter->filterUrl(Url::getCurrentUrl());
         if ($redirection !== null) {
@@ -305,6 +306,7 @@ class FrontController extends Singleton
         );
 
         Filechecks::dieIfDirectoriesNotWritable($directoriesToCheck);
+        SupportedBrowser::checkIfBrowserSupported();
 
         $this->handleMaintenanceMode();
         $this->handleProfiler();
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -419,7 +419,7 @@ class FrontController extends Singleton
             $this->makeAuthenticator($sessionAuth); // Piwik\Auth must be set to the correct Login plugin
         }
 
-        if (Piwik::isUserIsAnonymous() && Piwik::getModule() !== 'API') {
+        if ($this->isSupportedBrowserCheckNeeded()) {
             SupportedBrowser::checkIfBrowserSupported();
         }
 
@@ -747,5 +747,31 @@ class FrontController extends Singleton
     {
         $requestId = self::getUniqueRequestId();
         Common::sendHeader("X-Matomo-Request-Id: $requestId");
+    }
+
+    private function isSupportedBrowserCheckNeeded()
+    {
+        if (Piwik::getModule() === 'API' && (empty(Piwik::getAction()) || Piwik::getAction() === 'index' || Piwik::getAction() === 'glossary')) {
+            return false;
+        }
+
+        if (Piwik::getModule() === 'Widgetize') {
+            return true;
+        }
+
+        $generalConfig = Config::getInstance()->General;
+        if ($generalConfig['enable_framed_pages'] == '1' || $generalConfig['enable_framed_settings'] == '1') {
+            return true;
+        }
+
+        if (Common::getRequestVar('token_auth', '', 'string') !== '') {
+            return true;
+        }
+
+        if (Piwik::isUserIsAnonymous()) {
+            return true;
+        }
+
+        return false;
     }
 }
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -751,6 +751,20 @@ class FrontController extends Singleton
 
     private function isSupportedBrowserCheckNeeded()
     {
+        if (defined('PIWIK_ENABLE_DISPATCH') && !PIWIK_ENABLE_DISPATCH) {
+            return false;
+        }
+
+        $userAgent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
+        if ($userAgent === '') {
+            return false;
+        }
+
+        $isTestMode = defined('PIWIK_TEST_MODE') && PIWIK_TEST_MODE;
+        if (!$isTestMode && Common::isPhpCliMode() === true) {
+            return false;
+        }
+
         if (Piwik::getModule() === 'API' && (empty(Piwik::getAction()) || Piwik::getAction() === 'index' || Piwik::getAction() === 'glossary')) {
             return false;
         }
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -306,7 +306,6 @@ class FrontController extends Singleton
         );
 
         Filechecks::dieIfDirectoriesNotWritable($directoriesToCheck);
-        SupportedBrowser::checkIfBrowserSupported();
 
         $this->handleMaintenanceMode();
         $this->handleProfiler();
@@ -418,6 +417,10 @@ class FrontController extends Singleton
             }
         } else {
             $this->makeAuthenticator($sessionAuth); // Piwik\Auth must be set to the correct Login plugin
+        }
+
+        if (Piwik::isUserIsAnonymous() && Piwik::getModule() !== 'API') {
+            SupportedBrowser::checkIfBrowserSupported();
         }
 
         // Force the auth to use the token_auth if specified, so that embed dashboard
--- a/core/SupportedBrowser.php
+++ b/core/SupportedBrowser.php
@@ -38,7 +38,11 @@ class SupportedBrowser
 
     public static function checkIfBrowserSupported()
     {
-        $userAgent = Http::getUserAgent();
+        $userAgent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
+        if ($userAgent === '') {
+            return;
+        }
+
         $ddFactory = StaticContainer::get(DeviceDetectorFactory::class);
         /** @var \DeviceDetector\DeviceDetector */
         $deviceDetector = $ddFactory->makeInstance($userAgent);
--- a/core/SupportedBrowser.php
+++ b/core/SupportedBrowser.php
@@ -53,7 +53,7 @@ class SupportedBrowser
 
     private static function browserNotSupported($shortName, $version)
     {
-        return array_key_exists($shortName, self::$notSupportedBrowsers) && $version <= self::$notSupportedBrowsers[$shortName];
+        return array_key_exists($shortName, self::$notSupportedBrowsers) && $version > 0 && $version <= self::$notSupportedBrowsers[$shortName];
     }
 
     private static function throwException()
--- a/core/SupportedBrowser.php
+++ b/core/SupportedBrowser.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * Matomo - free/libre analytics platform
+ *
+ * @link https://matomo.org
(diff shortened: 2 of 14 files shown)


sgiehl:
Looks good so far. We could maybe add a UI test to make sure this will keep working. @flamisz  it's possible to manipulate the useragent used in UI tests (see https://github.com/matomo-org/matomo/blob/4.x-dev/tests/UI/specs/OptOutForm_spec.js)

Findus23:
Do I understand it correctly that this runs the users user-agent through device-detector on any single request?

I can imagine that there are quite a few slow servers out there (or ones without regex-caching) where this could add more than just a few milliseconds to every request.

Maybe just adding it to the login should be enough (unless that's what the code already does).

core/SupportedBrowser.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        $userAgent = Http::getUserAgent();
+        $ddFactory = StaticContainer::get(DeviceDetectorFactory::class);
+        /** @var \DeviceDetector\DeviceDetector */
+        $deviceDetector = $ddFactory->makeInstance($userAgent);
+
+        $deviceDetector->parse();
+        $client = $deviceDetector->getClient();
+
+        if ($client['type'] === 'browser' && self::browserNotSupported($client['short_name'], (int)$client['version'])) {
```
Findus23, line 53: Really random guess (as I don't know PHP well enough):
Is it possible that there is some version string in some minor browser that looks something like `"v3.4.5"`  and will then become `0`?

And what about user agents that are detected as a popular browser, but don't include a version? (maybe instead of freezing the user agent, the browser vendor decides to remove it completely some day in the future to urge people to use feature-detection instead)

UserAgents feel a bit too chaotic for me to decide based on them to completly lock users out of using Matomo

flamisz: @sgiehl probably knows better this, but what I see in the device detector, it tries to get the major version of the client, without any `v` or `.`. Is that correct @sgiehl ?

But maybe a check if the version is 0 would be useful, you are right.

And the other question is the caching. The device detector has some kind of cache, hasn't it @sgiehl ?

I will look into the UI test.

tsteur:
BTW only running it on login could work for regular requests but for widgets it would always need to run (eg when` Piwik::getModule()===Widgetize`), it would also need to always run for anonymous users, and for API module/plugin we wouldn't want it to run (for performance etc since it wouldn't be "rendered" in the browser)

flamisz:
@tsteur I modified the logic: only checks the browser if the user is `anonymous` and the module is not `API`. This way even the widgets won't be checked if you are in the same browser, and of course before you log in, it will be checked.

core/FrontController.php
```diff
@@ -418,6 +419,10 @@ public function init()
             $this->makeAuthenticator($sessionAuth); // Piwik\Auth must be set to the correct Login plugin
         }
 
+        if (Piwik::isUserIsAnonymous() && Piwik::getModule() !== 'API') {
```
tsteur: @flamisz We'd need to make sure to always execute the browser check when the `Widgetize` API is used. Ideally, for simplicity and to make sure it runs in all cases we would simply always run it when the `Widgetize` module is requested. You may be logged in when viewing the Widgets (either through a session or token).

Regarding the API we would need to add a check like `$skipBrowserCheck = Piwik::getModule() === 'API' && (empty(Piwik::getAction()) || Piwik::getAction() === 'index' || Piwik::getAction() === 'glossary'))`. When you call the API, then basically the `index` action (which is the default action) will be executed. The browser check should not be executed for the `glossary` either (which you can see btw here: https://glossary.matomo.org/). For all other methods in the API controller  (like `listAllAPI`) we would still want to check the browser. 

And there's one more thing :) If the `[General] enable_framed_pages=1` or `[General]enable_framed_settings=1` config is enabled, then we probably also always want to execute the browser check unless it is the API. So in these cases we also want to execute the check when user is not anonymous. This feature is tricky and we don't recommend enabling this feature which is documented in https://matomo.org/faq/troubleshooting/faq_147/ but it basically allows any action to be embedded like if it was Widgetized.  You can find some information about the config here: https://developer.matomo.org/guides/piwiks-ini-configuration . I'm thinking instead of checking the config there is a more simpler solution though that would cover even more cases.

So basically we want to ensure the check is executed when user is logged in when:

* module === Widgetize
*  `[General] enable_framed_pages=1` or `[General]enable_framed_settings=1` config is enabled
* A third new check: We should also always execute the browser check if `Common::getRequestVar('token_auth', '', 'string')` is not empty. I think this would make both the previous two points obsolete as it should cover all the cases with widgetize and the 2 settings but just to be safe we may want to have this as yet another check :) 

And we always skip it when the API.index or API.glossary action is executed (see the code above).

Sorry for this being so complicated :) Happy to go over this tomorrow in the call :)

flamisz: Thanks, @tsteur for the detailed description, clear and easily understandable.
I make the change 👨‍💻

flamisz: @tsteur I updated the `FrontController.php`. 
Do you have any recommendations for tests? I made 4 UI tests. Any integration tests? Or more UI?
Thanks

tsteur: I think it be good to add a simple test for `checkIfBrowserSupported`. In this case I would have likely actually only created integration tests for `checkIfBrowserSupported ()` and for calling the `dispatch` method as integration or system tests are a lot faster to execute than UI tests and would have likely kind of achieved the same. Or one UI test and the rest integration/system tests. There isn't a need though to change this as everyone might do it differently and both solutions are legit. Generally, if we can test some behaviour using integration/system test like in this case, this would have probably been the preferred way because there isn't really any custom UI to test and it only checks if the exception is triggered. I will try to document this. It's hard to find the right balance there though and nobody gets it always right as it's not really clear when to do what.

tsteur: I created https://github.com/matomo-org/developer-documentation/commit/473259904455c866f7d8b6a8d468567e7660ec33#diff-6f870fb2ed7d589ffe2667ca2c4a9d2d3479fa9443a8a573d750bab93be0059f which should alter appear on https://developer.matomo.org/guides/tests-ui

flamisz: @tsteur just one more question about the tests.

I added tests for `checkIfBrowserSupported()` but the logic in the `FrontController` is not tested yet. Should we do something about that?

tsteur: The logic in the FrontController would be tested in the UI test so all good

core/SupportedBrowser.php
Findus23: I am not entirely sure if that is true for every browser in the detection list, but it might be.

And to answer the question about caching: DeviceDetector does caching, but it caches the conversion from the yaml rule files into one giant regex. The giant regex would still need to be executed every time DD is run, which may or may not take a bit of time.

sgiehl: The version should be without any prefix. 
Maybe we could cache the result or at least some kind of compatibility state in the session. Should be unlikely someone switches the useragent within a session 🙈

core/SupportedBrowser.php
```diff
@@ -0,0 +1,69 @@
...
+        'IE' => 10,
+        'SF' => 4,
+        'CH' => 15,
+        'OP' => 12,
+        'PS' => 12,
+    ];
+
+    public static function checkIfBrowserSupported()
+    {
+        $userAgent = Http::getUserAgent();
```
tsteur: btw @flamisz  it may not be the best method to get the user agent actually as it's only meant for the HTTP class as it returns a default user agent if none is set. If I see this right there is no method for this actually yet so we could just use something like `
        $userAgent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';`  and then if the user agent is empty then we'd directly `return true`. Or we could check in `isSupportedBrowserCheckNeeded` method if a user agent is set and return true/false there?

core/FrontController.php
```diff
@@ -743,4 +748,30 @@ private static function setRequestIdHeader()
         $requestId = self::getUniqueRequestId();
         Common::sendHeader("X-Matomo-Request-Id: $requestId");
     }
+
+    private function isSupportedBrowserCheckNeeded()
+    {
+        if (Piwik::getModule() === 'API' && (empty(Piwik::getAction()) || Piwik::getAction() === 'index' || Piwik::getAction() === 'glossary')) {
```
tsteur, line 768: btw one more thing...  we could technically return also `false` if `Common::isPhpCliMode()===true` meaning Matomo is used on CLI.

Indirectly this would be covered by the `is user agent empty check` below but could be also added additionally. Basically, on CLI there is no browser and thus no need to execute anything. The problem you will then run into is that unit tests run on CLI and thus it won't be executed there anymore. You would then need to add yet another check for `defined('PIWIK_TEST_MODE') && PIWIK_TEST_MODE` to detect if it's running in tests and then ignore this `is cli mode` check :) So maybe we rather just not check for CLI and rely on the `is user agent empty` check for now :)

core/SupportedBrowser.php
tsteur: FYI we would be only able to store it in a session if a session was also started ideally. Might not need to add it for now as it might not make it too much slower? It's just that additional caches/sessions make things quite a bit more complicated again and sometimes we wouldn't even start a session maybe in some cases (I think eg when widgetized is executed using token_auth etc) so it's use might be limited. Personally would maybe not add caching by session for now

core/FrontController.php
```diff
@@ -418,6 +419,10 @@ public function init()
             $this->makeAuthenticator($sessionAuth); // Piwik\Auth must be set to the correct Login plugin
         }
 
+        if ($this->isSupportedBrowserCheckNeeded()) {
```
tsteur, line 422: @flamisz sorry I just realise we will need to move this somewhere else in the frontcontroller. Maybe in the `dispatch` method or maybe we only execute the check when `defined('PIWIK_ENABLE_DISPATCH') && PIWIK_ENABLE_DISPATCH === true` which I think would work too.

Problem: In our Matomo for WordPress plugin we call the `init()` method whenever we need to bootstrap Matomo "headless" see https://github.com/matomo-org/wp-matomo/blob/4.1.3/classes/WpMatomo/Bootstrap.php#L41-L78  In that case we disable the dispatch and session start see https://github.com/matomo-org/wp-matomo/blob/4.1.3/classes/WpMatomo/Bootstrap.php#L49-L56

If we were to leave the check, and Matomo is bootstrapped as part of a regular website page load of a regular Matomo user, then we'd break their websites on some requests for older browsers.

I'm pretty sure the other WP-Matomo plugin does the same and I'm pretty sure we recommended for quite a while (or we still do somewhere maybe) calling the frontcontroller init method to bootstrap Matomo "headless" and would therefore also break other parts. You can see how this used to be done quite often (and probably still many do) eg here: https://forum.matomo.org/t/show-the-piwik-data-in-a-php-page/20230/2

Not sure if it's better to move it into the dispatch method or checking for the constant. It should be fine to simply not execute the check if `PIWIK_ENABLE_DISPATCH===false` I think.

core/SupportedBrowser.php
flamisz: yes, you are right. `$_SERVER['HTTP_USER_AGENT']` is the good solution here.

core/FrontController.php
flamisz: ok, I check this and modify

core/FrontController.php
flamisz: I'm happy to add the test mode check. this way it's clear what and why is happening

flamisz:
Hey @tsteur I modified the PR according to your comments. Please view it again when you have some time.

Thanks

tsteur:
@flamisz overall looks good from my perspective 👍  but didn't test it

flamisz:
@tsteur I did my best to test on a couple of browsers and a couple of use cases with BrowserStack.
And of course, we have some test cases as well.

tsteur:
👍  feel free to merge if the tests pass (didn't check)

mpdude:
This leads to HTTP 500 errors that alert my server monitoring; additionally, 5xx is not the right kind of status code for such conditions. 

Maybe `400 Bad Request` would be more appropriate, although it's not a perfect match either.

sgiehl:
@mpdude thanks for your suggestion. Would you mind creating a new issue for that referencing this PR. We likely might forget about that if there is no separate issue we can plan to work on.

sgiehl:
@mpdude just saw your comment on #17738. guess that should be enough

PR #17260, opened 2021-02-24
Hide unique visitors email in VisitsSummary.get if not enabled in config.

### LINKED ISSUE #16222: Hide Unique Visitors from email reports when the metric is not activated for this period
When `enable_processing_unique_visitors_*=0` in the config file, Email reports will still show the Unique Visitors metric. 

The goal of this issue is to hide the metric from email reports (and UI) when processing unique visitors is disabled for the period of the email report (or selected in the UI). 

## from user feedback

> We would like to request a change to one of the scheduled email reports. We are sending monthly reports to all of our members. Overall they are great and will be a good addition to the service we provide. One issue though is that they always list 0 unique visitors. It looks like this statistic is not available for monthly reports, although it is still included in the report.

> Could we have the number of unique visitors removed from our reports to avoid confusing our users?

diosmosis:
### Description:

Fixes #16222 

The only instance of unique visitors I could find when processing was disabled was in VisitsSummary.get, and I noticed the metric is enabled everywhere in this case. This removes the metric everywhere if processing is disabled.

Note: this will also remove the metric even if there is existing data. So if a user computes it for months initially, then disables it for months, the old data won't be visible. Not sure if this is desired (cc @tsteur/@mattab).

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/VisitsSummary/Reports/Get.php
+++ b/plugins/VisitsSummary/Reports/Get.php
@@ -43,12 +43,17 @@ class Get extends \Piwik\Plugin\Report
             new AverageTimeOnSite()
         );
         $this->metrics = array(
-            'nb_uniq_visitors',
             'nb_visits',
             $this->usersColumn,
             'nb_actions',
             'max_actions'
         );
+
+        $period = Common::getRequestVar('period', 'day');
+        if (SettingsPiwik::isUniqueVisitorsEnabled($period)) {
+            $this->metrics = array_merge(['nb_uniq_visitors'], $this->metrics);
+        }
+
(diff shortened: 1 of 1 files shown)


sgiehl (approved):

diosmosis:
@mattab / @tsteur can you tell me if this is desired behavior:

> Note: this will also remove the metric even if there is existing data. So if a user computes it for months initially, then disables it for months, the old data won't be visible. Not sure if this is desired

tsteur:
@diosmosis I reckon that's fine 👍

Issue #15501, opened 2020-02-03
Tracker API: trim spaces in event names, category, actions


mattab:
Currently in the Tracker API we do trim content names, pieces, targets strings. To be consistent we could also trim event names (`e_n`), event category (`e_c`), event actions (`e_a`). 

We could even trim a couple of other inputs where it could make sense like Page title (`action_name`), possibly site search keyword (`search`, `search_cat`). 

fyi: creating this issue as a result of a user and developer getting confused about duplicate data in an Events report. `Problem solved by "invisible empty space".`

tsteur:
sounds good

flamisz:
Hi @mattab and @tsteur 
I was looking into this issue and this is what I found:

I believe the `search` and `search_cat` already trimmed here:
[ActionSiteSearch](https://github.com/matomo-org/matomo/blob/3141a74700044ea930bee0f53618906f6f8a1a4b/plugins/Actions/Actions/ActionSiteSearch.php#L258-L259)

I checked the `action_name`, found a place where it should be trimmed, I make a PR for that.

I check the other request parameters too.

EDITED:

I checked the other parameters and they all looking good to me: either not string, or goes through some kind of validation or trimmed.
Do I miss something here?

tsteur:
@mattab looks indeed like event names etc are already trimmed (by looking at code, not reproducing). Do we maybe otherwise still have the link to the original conversation?

mattab:
@tsteur sure, you can find it in conversation #641565

flamisz:
> @tsteur sure, you can find it in conversation #641565

@mattab what is this number? where can I find this conversation?
thanks

mattab:
Here is what user wrote:

> The empty space was in the "ACTION" I have sanded throw the _paq.push("trackEvent",…) function, so I thought if the developer has any mistake in his code, then you can also trim the text.

Before:

> I will ask to update our Matomo version.
> Problem solved by "invisible empty space".
> Maybe it is better to have a "trim" function when you receive the tracker.

Before (original report):

> I have a new problem using the goals and funnels,
> I build a new goal with few steps and a funnel.
> Every time the user gets a specific level, the funnels goes out and it never finish the funnel.
> As you can see, I'm sending in the network panel the required action.
> screenshots missing but not relevant anyway

flamisz:
Thanks @mattab, it's useful. In this case indeed the action name was the problem. The PR will fix that.

sgiehl:
[...]

PR #17266, opened 2021-02-25
test for trim action name

flamisz:
### Description:

ref #15501 

Turned out it's not fixing the original issue. For more details see the comments on that issue.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Tracker/ActionPageview.php
+++ b/core/Tracker/ActionPageview.php
@@ -58,7 +58,7 @@ class ActionPageview extends Action
         $actionCategoryDelimiter = $this->getActionCategoryDelimiter();
 
         if ($actionCategoryDelimiter === '') {
-            return trim($actionName);
+            return $actionName;
         }
 
         // create an array of the categories delimited by the delimiter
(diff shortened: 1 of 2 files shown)


sgiehl (changes requested):
@flamisz  Looks fine. Maybe you could add a new test for that case where the delimiter is set empty and an actions (url) with pre- or appending spaces is tracked. Guess this could be added here: https://github.com/matomo-org/matomo/blob/4.x-dev/tests/PHPUnit/Integration/Tracker/ActionTest.php

flamisz:
@tsteur and @sgiehl I found something interesting with this `action_name` while I was writing a test. 
I modified back my changes to see the test a made is failing, but it's not. So I checked and we already trim the action name, but not there where I modified the code. 

https://github.com/matomo-org/matomo/blob/b17616097d6b353dc65b0d77773d0c78977646e8/core/Tracker/ActionPageview.php#L28-L29

and then:

https://github.com/matomo-org/matomo/blob/b17616097d6b353dc65b0d77773d0c78977646e8/core/Tracker/Action.php#L212-L215

So this couldn't be the original issue. I can add the tests and not modifying the code, so we will see actually it is trimmed. 
But then what was the issue for the user?

sgiehl:
@flamisz Had a closer look at the code myself now and it seems the action name is trimmed in all cases already. I guess there are actually no changes needed. Will comment the original issue as well.

sgiehl:
Merging this one as an additional test can't hurt ;-)

Issue #17129, opened 2021-01-21
When creating a new segment "pre-processing" is selected by default


BigBerny:
When you create a new segment "SEGMENTED REPORTS ARE PRE-PROCESSED (FASTER, REQUIRES CRON)" is selected by default even though it says "Default" in the other option: "SEGMENTED REPORTS ARE PROCESSED IN REAL TIME (DEFAULT)"

IMHO "Real time" makes more sense usually but that might depend on the user. At least I'd make it consistent.

Thanks a lot!

tsteur:
Hi @BigBerny do you mind posting a screenshot of what you refer to? I'm only seeing this: 
![image](https://user-images.githubusercontent.com/273120/105412698-ed8b1980-5c99-11eb-8961-f00799f47585.png) I'm not sure what other option you refer to for example.

BigBerny:
Sure! On our setup the other option is selected by default when we create a new segment (SEGMENTED REPORTS ARE PRE-PROCESSED (FASTER, REQUIRES CRON)):
![image](https://user-images.githubusercontent.com/1528504/105417628-2eecdc00-5c3c-11eb-917e-e57e7c7d2dba.png)

tsteur:
OK thanks I got it. It should basically append the word "Default" to this by default selected option and remove it from the other option 👍  all clear now, thanks for this @BigBerny

BigBerny:
Exactly. Or the other way around: making real-time processing default.
It's a minor problem though. 😊

flamisz:
Fixed by #17277

PR #17277, opened 2021-02-26
Remove the text "(Default)" in the new segment screen

flamisz:
### Description:

ref: #17129 

The default value depends on the `enable_browser_archiving_triggering` settings. The easiest fix to delete the `default` in the name, so it won't be confusing if someone set that value to false.

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (approved):
Makes sense for me to completely remove the `Default` text.

PR #17278, opened 2021-02-26
Fix Marketplace falsely thinks a plugin is installed when it is not

tsteur:
### Description:

Notices this earlier on demo2. We had removed various plugins from the disk but the config still had the plugins marked as installed. Then I wanted to reinstall the plugins but the marketplace wouldn't let me install it because it was thinking the plugin is already installed. That's because it was only looking at `PluginsInstalled[]` entries in the config but not if the plugin actually still existed in the file system.

This change now makes plugins installable again in such a case. I noticed few other usages of `isPluginInstalled()` but not always would we actually want to check if the plugin is in the filesystem. Either from a performance point of view (eg in the component update checker) or because the logic doesn't need/want it.

For now just wanted to fix this specific case. Seems there are no tests so far for the plugin manager but can add one if needed.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Plugin/Manager.php
+++ b/core/Plugin/Manager.php
@@ -1600,13 +1600,22 @@ class Manager
     }
 
     /**
-     * @param $pluginName
+     * @param string $pluginName
+     * @param bool $checkPluginExistsInFilesystem if enabled, it won't rely on the information in the config file only
+     *                                            but also check the filesystem if the plugin really is installed.
+     *                                            For performance reasons this is not the case by default.
      * @return bool
      */
-    public function isPluginInstalled($pluginName)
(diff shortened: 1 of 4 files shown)


flamisz (approved):
👍

diosmosis:
@tsteur is this something we'd want covered by a test? (I can add one if needed)

tsteur:
@diosmosis added test for the plugins manager 👍  I reckon it might not easy to add a test for the marketplace where it thinks the plugin is installed but actually missing from the files since the marketplace would also need to show that plugin (meaning it would need to exist on the marketplace). I'm thinking a test wouldn't be needed

PR #17288, opened 2021-03-02
Track JS errors only once per page view

### LINKED ISSUE #17163: JS Error tracking should record the same error only once per page view
When using the `enableJSErrorTracking`, then we'd want the same error to be reported only once per page view.

Otherwise if there's a loop causing errors, then you could be potentially dealing with thousands of tracking requests per user concurrently.

flamisz:
### Description:

fixes #17163 

### Review

* [x] Functional review done
* [x] Code review done


flamisz:
build js

flamisz:
build js

sgiehl:
build js

sgiehl (approved):
Looks good so far. 
@tsteur Not sure if that is something we should mention in the changelog. Or maybe we should update https://matomo.org/faq/how-to/how-do-i-enable-basic-javascript-error-tracking-and-reporting-in-matomo-browser-console-error-messages/ after merging and mention, that each error will only be tracked once?

flamisz:
@tsteur I think it's a good idea to mention it somewhere, we only track once per page view. The original issue name will get into the changelog, won't it?

tsteur:
build js

tsteur:
@flamisz updated https://matomo.org/faq/how-to/how-do-i-enable-basic-javascript-error-tracking-and-reporting-in-matomo-browser-console-error-messages/ and could also add it to the developer changelog https://github.com/matomo-org/matomo/blob/4.x-dev/CHANGELOG.md . wouldn't need to mention the issue number

flamisz:
build js

diosmosis:
@tsteur @sgiehl is this something we'd want to write a quick test for in https://github.com/matomo-org/matomo/blob/4.x-dev/tests/javascript/index.php

tsteur:
Yes, it be great to have a quick test for it 👍

js/piwik.js
```diff
@@ -184,7 +184,9 @@ if (typeof window.Matomo !== 'object') {
 
             trackerIdCounter = 0,
 
-            isPageUnloading = false;
+            isPageUnloading = false,
+
+            javaScriptErrors = [];
```
flamisz: Hey @diosmosis, I added a test. 

But now I have a question about this line. First of all, to make the test work, I needed to add a new method, to get this `javaSriptErrors`. That's fine I think.

I'm not sure this `javaSrciptErrors` is actually in a good place. Shouldn't it be around line 2388, inside the Matomo Tracker Class?

What's the difference and which one is better?
Thanks, Zoltan

diosmosis: @flamisz if you add it to 2388, then there will be one array per tracker. It's possible for users to create multiple trackers on the same website/webpage that track to different matomo's or different idsites (for example). Right now the array is "global". I'm not sure there's a benefit to either way, since in either case we want to skip the same javascript errors. I guess, maybe if a new tracker is created later on in the user's visit (say for a single page app where the visit time can be long), and a similar javascript error comes up, maybe we still want to track that one? I'm not sure about the use case, maybe @tsteur has an opinion.

diosmosis: Actually I guess it makes sense that we'd want to send the javascript error if the tracker hasn't seen it before, not if piwik.js hasn't seen it before. I think it makes sense to move the array if you can. Good call.

flamisz: great, that's what I thought. I move it and check the tests again if need to be updated or not. Thanks

flamisz:
build js

sgiehl:
@flamisz seems the javascript tests are still failing

flamisz:
build js

flamisz:
Hi @sgiehl, I fixed the js tests, but now the Travis CI - Pull Request build failed. What could it be? Is there a way to rebuild it? I couldn't find any (obviously without pushing any changes).

diosmosis:
@flamisz there should be a "restart build" button: 
image

you may need to authorize your github account w/ travis if it's not there (it might be in settings after logging in).

EDIT: that said, we generally only need one build to pass.

flamisz:
build js

flamisz:
build js

PR #17292, opened 2021-03-02
Hide loading errors when user changes reporting page

### LINKED ISSUE #10578: Display the "oops" red error message, only when there was an error, do not show it when user clicks away to load another menu or page
### Reproduce
- In Piwik 3, open the dahsboard
- while the dashboard is loading, click on a top menu link
- the dashboard cancels loading the widget, as expected... However
  - Got: a red error notification message
  - Expected instead: no red notification message, as I just clicked away and this is not an error
### screenshot

oops not needed

Regression from 2.x

### LINKED ISSUE #17104: Error Message notification is not hidden when navigating to a new page
When a server error results in the message below being displayed, transitioning to another page via the menu does not remove it, it continues to be displayed at the top of each page until a hard refresh is done.

Ver 4.1.1

`Oops… there was a problem during the request. Maybe the server had a temporary issue, or maybe you requested a report with too much data. Please try again. If this error occurs repeatedly please contact your Matomo administrator for assistance.`

flamisz:
### Description:

fixes #17104 

fixes #10578

### Review

* [x] Functional review done
* [x] Code review done

### DIFF
--- a/plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
+++ b/plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
@@ -218,6 +218,8 @@
         });
 
         $rootScope.$on('piwikPageChange', function (event) {
+            globalAjaxQueue.abort();
+
             $scope.helpShownCategory = null;
             $scope.currentCategory = piwikUrl.getSearchParam('category');
             $scope.currentSubcategory = piwikUrl.getSearchParam('subcategory');
--- a/plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
+++ b/plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
@@ -226,6 +226,8 @@
                 $scope.showHelp(showSubcategoryHelpOnLoad.category, showSubcategoryHelpOnLoad.subcategory);
                 showSubcategoryHelpOnLoad = null;
             }
+
+            $('#loadingError').hide();
         });
     }
 })();
--- a/plugins/CoreHome/angularjs/widget-loader/widgetloader.directive.js
+++ b/plugins/CoreHome/angularjs/widget-loader/widgetloader.directive.js
@@ -170,20 +170,22 @@
                                     element: currentElement,
                                 });
                             });
-                        })['catch'](function () {
+                        })['catch'](function (response) {
                             if (thisChangeId !== changeCounter) {
                                 // another widget was requested meanwhile, ignore this response
                                 return;
                             }
+
+                            if (response.xhrStatus === 'abort') {
+                                return;
+                            };
 
                             httpCanceler = null;
 
                             cleanupLastWidgetContent();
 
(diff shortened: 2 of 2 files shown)


sgiehl (dismissed):
I guess this makes it at least a bit better. But there are still some cases where an error message might be shown, that is unrelated. e.g. when you load a report page that sends xhr requests to fetch additional data. But before those requests were answered you already switch to another report page. If an old request then fails, the error message will be show even if it actually was for the page before. Guess that won't be that easy to fix, as we need some mechanism to automatically cancel xhr requests that aren't needed anymore (or at least ignore possible errors). Maybe we could store the url that was related to an XHR when it is sent or something similar. But guess that's something for another PR. ping @tsteur

flamisz:
@sgiehl yes, you are right. I haven't thought about that. Let's wait what are @tsteur's thoughts on this.

flamisz:
@sgiehl @tsteur luckily our `ajaxHelper.js` collects all ajax requests into an array and we already have an `abort` method for it:
[https://github.com/matomo-org/matomo/blob/5e2c5dce5cd88ba3872a72ae3e7675514e3a6098/plugins/Morpheus/javascripts/ajaxHelper.js#L44-L58](https://github.com/matomo-org/matomo/blob/5e2c5dce5cd88ba3872a72ae3e7675514e3a6098/plugins/Morpheus/javascripts/ajaxHelper.js#L44-L58)

So I added one more line to call this method on every reporting page load. I tested it locally and it did the job.

Findus23:
This might also be related to https://github.com/matomo-org/matomo/issues/10578

flamisz:
> This might also be related to #10578

@Findus23 yes, there is a chance it solves that as well. I try to reproduce the other issue without the modifications and then with this PR.

Update: I see now, the other issue is different. In this PR we just abort the ajax requests when we change reporting page. The other issue happens with a full page reload: the new page starts loading but the browser still shows the original page, which has an error now, because the request are not loading anymore. 
One solution for that one could be to delay that notification message a little bit, that time should be enough to the browser to load the new page.

flamisz:
I updated the PR, because I find out where is that error on dashboard coming from #10578.
I added 2sec timeout for showing that error message. In this way, if the user stays on the page and an error occurs, will see it anyway, but if we clicking away from the page, the browser will have time to start loading the other page, before the error message shows up.

diosmosis:
I haven't tested this yet, but for the second issue, would it also work to cancel ajax requests in the window's `unload` or `beforeunload` events? If that would work, the solution wouldn't depend on a timeout which might avoid edge cases where it takes a long time for the next page to start loading.

flamisz:
Hi @diosmosis, I modified the PR. Instead of using a timeout, I'm checking if the `xhrStatus` is `aborted`. If it's `aborted`, the error is not shown.

sgiehl (approved):
Played around with failing requests locally, but it seems the error message now is only shown if the error really affects the current page. Guess this would be good to merge now 👍

PR #17294, opened 2021-03-02
Set the previous month date to 1 when compare months

### LINKED ISSUE #17080: Matomo comparing month report with same month when clicking on the last day of the month
When selecting a time period of "Month" and selecting to compare to "Previous period" for some months causes Matomo to compare to the same month. 

Have tested and so far it occurs when selecting the 31st of the month for December, October, July, May, March 2020 e.g.: 
image

Link where we can see an example: https://demo.matomo.cloud/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=month&date=2020-12-31&segment=&category=Dashboard_Dashboard&subcategory=1&comparePeriods%5B%5D=month&compareDates%5B%5D=2020-12-01

image

flamisz:
### Description:

fixes #17080 

### Review

* [x] Functional review done
* [x] Code review done

### DIFF
--- a/plugins/CoreHome/angularjs/common/services/periods.spec.js
+++ b/plugins/CoreHome/angularjs/common/services/periods.spec.js
@@ -99,13 +99,11 @@
             expect(result.endDate).to.eql(clearDate(last));
         });
 
-        function mockDateNow01() {
-            return clearDate('2021-03-31').getTime();
-        }
-
         it('should parse last month properly when date is 31th march', function() {
             originalDateNow = Date.now;
-            Date.now = mockDateNow01;
+            Date.now = function() {
+                return clearDate('2021-03-31').getTime();
+            }
 
             var result = piwikPeriods.parseDate('last month');
 
@@ -114,13 +112,11 @@
             Date.now = originalDateNow;
         });
 
-        function mockDateNow02() {
-            return clearDate('2021-03-10').getTime();
-        }
-
         it('should parse last month properly', function() {
             originalDateNow = Date.now;
-            Date.now = mockDateNow02;
+            Date.now = function() {
+                return clearDate('2021-03-10').getTime();
+            };
 
             var result = piwikPeriods.parseDate('last month');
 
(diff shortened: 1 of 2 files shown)


sgiehl:
I guess the change is fine. Seems we currently don't have any unit tests for the periods class in javascript. Might maybe be good to at least add some basic tests that covers some of the edge cases that were making trouble. 
@flamisz Not sure if you already had a look at the angularjs tests. They are located with the javascript files but are named `*.spec.js`

flamisz:
You were right @sgiehl, the tests are important. I found one other very similar issue this way 😄 .

plugins/CoreHome/angularjs/common/services/periods.spec.js
```diff
@@ -0,0 +1,132 @@
...
+            expect(result.endDate).to.eql(clearDate(last));
+        });
+
+        function mockDateNow01() {
+            return clearDate('2021-03-31').getTime();
+        }
+
+        it('should parse last month properly when date is 31th march', function() {
+            originalDateNow = Date.now;
+            Date.now = mockDateNow01;
```
sgiehl: The additional method actually shouldn't be needed here and below. You could simply use a anonymous function

```suggestion
        it('should parse last month properly when date is 31th march', function() {
            originalDateNow = Date.now;
            Date.now = function() {
                return clearDate('2021-03-31').getTime();
            };
```

sgiehl (changes requested):
Left a minor suggestion for improvements. Besides that we could think about adding some test cases for invalid dates maybe, so we know how they are handled, but that does not need to be part of this PR actually.
Otherwise looks good 👍

plugins/CoreHome/angularjs/common/services/periods.spec.js
flamisz: Nice, I like it. Will make this change.

sgiehl (approved):

PR #17295, opened 2021-03-02
Prevent possible warning during GeoIP2 update if providers aren't initialised

tsteur:
### Description:

refs https://wordpress.org/support/topic/repeated-php-warnings-on-cron-execution/

It should be better to use the method anyway instead of hard coding the variable.

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (approved):
makes sense 👍

PR #17296, opened 2021-03-03
Check usernames and emails against each other on user create and update

### LINKED ISSUE #17070: Prevent new username from using existing user email address
It seems that when creating a new user, it's possible to set the username to an existing user email address which will then cause the existing user to not be able to log in.

For example, a super user with username admin and email address of admin@example.com creates a new user with email user@example.com and accidentally sets the new username to admin@example.com

This prevents the super user from being able to log in to their account using their email address.

image

flamisz:
### Description:

fixes #17070 

### Review

* [x] Functional review done
* [x] Code review done

### DIFF
--- a/plugins/UsersManager/tests/Integration/APITest.php
+++ b/plugins/UsersManager/tests/Integration/APITest.php
@@ -395,7 +395,7 @@ class APITest extends IntegrationTestCase
         $user2 = 'existed@example.com';
         $this->api->addUser($user2, 'password', 'userlogin2@password.de');
 
-        $this->api->updateUser($this->login, $this->password,  'existed@example.com', false, $this->password);
+        $this->api->updateUser($this->login, $this->password,  $user2, false, $this->password);
     }
 
     public function test_update_can_update_user_email_to_own_username()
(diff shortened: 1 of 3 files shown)


plugins/UsersManager/tests/Integration/APITest.php
```diff
@@ -387,6 +387,49 @@ public function test_updateUser_failsIfPasswordTooLong()
...
 
+    public function test_update_user_fails_if_email_exists_as_other_user_username()
+    {
+        $this->expectException(\Exception::class);
+        $this->expectExceptionMessage('UsersManager_ExceptionEmailExistsAsLogin');
+
+        $user2 = 'existed@example.com';
+        $this->api->addUser($user2, 'password', 'userlogin2@password.de');
+
+        $this->api->updateUser($this->login, $this->password,  'existed@example.com', false, $this->password);
```
sgiehl: could reuse the variable so it's easy to see it's the same value
```suggestion
        $this->api->updateUser($this->login, $this->password, $user2, false, $this->password);
```

plugins/UsersManager/tests/Integration/APITest.php
```diff
@@ -387,6 +387,49 @@ public function test_updateUser_failsIfPasswordTooLong()
...
+    }
+
+    public function test_update_can_update_user_email_to_own_username()
+    {
+        $user2 = 'ownemail@example.com';
+        $password = 'password';
+        $this->api->addUser($user2, $password, 'ownemail_wrong@example.com');
+
+        FakeAccess::$identity = $user2;
+        $this->api->updateUser($user2, $password, 'ownemail@example.com', false, $password);
```
sgiehl:
```suggestion
        $this->api->updateUser($user2, $password, $user2, false, $password);
```

plugins/UsersManager/tests/Integration/APITest.php
```diff
@@ -387,6 +387,49 @@ public function test_updateUser_failsIfPasswordTooLong()
...
+
+        $user2Array = $this->api->getUser($user2);
+        $this->assertEquals($user2Array['email'], 'ownemail@example.com');
```
sgiehl:
```suggestion
        $this->assertEquals($user2Array['email'], $user2);
```

sgiehl (changes requested):
Left a few minor comments. Otherwise looks good so far.

sgiehl (approved):

PR #17297, opened 2021-03-03
Provide way to extend visitorProfilePopup

gtbl2012:
### Description:

We find it hard to add custom HTML related to visitor info(userId etc.) in visitorProfilePopup via plugins. Because it doesn't provide any postEvents with visitor info or plugin API to add custom item.

Supplying visitorData to Template.afterVisitorProfileOverview could solve this problem. Making visitorProfilePopup more customizable.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Live/templates/getVisitorProfilePopup.twig
+++ b/plugins/Live/templates/getVisitorProfilePopup.twig
@@ -60,7 +60,7 @@
 
                 {{ profileSummary|raw }}
 
-                {{ postEvent('Template.afterVisitorProfileOverview') }}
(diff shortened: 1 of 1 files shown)


sgiehl:
@gtbl2012  Thanks for you PR.
Depending on what you are trying to achieve, it might already be possible with a `VisitorDetails` class. See https://developer.matomo.org/guides/visitor-log-and-profile

gtbl2012:
Thanks a lot, implementing ProfileSummary is better. I just missed that part.

PR #17300, opened 2021-03-03
Development mode notifications when Matomo not installed through git

### LINKED ISSUE #16922: replace generate:plugin warning with pointer to guide
When someone wants to create a plugin and is using the Matomo version from matomo.zip, they probably get an error like this one:

>Uncaught exception: Error: Class 'Piwik\Plugins\ExamplePlugin\ExamplePlugin' not found

This is expected as Matomo doesn't ship with the Example* plugins anymore, but doesn't help the person struggling with this.

Instead, the `generate:plugin` should detect if those plugins are missing and if not point people towards https://developer.matomo.org/guides/getting-started-part-1

flamisz:
### Description:

fixes #16922 

### Review

* [ ] Functional review done
* [x] Code review done

### DIFF
--- a/core/Plugin/ControllerAdmin.php
+++ b/core/Plugin/ControllerAdmin.php
@@ -23,6 +23,7 @@ use Piwik\Url;
 use Piwik\Version;
 use Piwik\View;
 use Piwik\ProxyHttp;
+use Piwik\SettingsPiwik;
 
 /**
  * Base class of plugin controllers that provide administrative functionality.
@@ -170,6 +171,25 @@ abstract class ControllerAdmin extends Controller
         Notification\Manager::notify('ControllerAdmin_HttpIsUsed', $notification);
     }
 
+    private static function notifyIfDevelopmentModeOnButNotInstalledThroughGit()
+    {
+        if (!Piwik::hasUserSuperUserAccess()) {
+            return;
+        }
+
+        if (SettingsPiwik::isGitDeployment()) {
+            return;
+        }
+
+        $message = Piwik::translate('General_WarningDevelopmentModeOnButNotGitInstalled');
+
+        $notification = new Notification($message);
+        $notification->context = Notification::CONTEXT_WARNING;
+        $notification->raw = true;
+        $notification->flags = Notification::FLAG_CLEAR;
+        Notification\Manager::notify('ControllerAdmin_DevelopmentModeOn', $notification);
+    }
+
     /**
      * @ignore
      */
@@ -313,6 +333,7 @@ abstract class ControllerAdmin extends Controller
         self::notifyWhenTrackingStatisticsDisabled();
         self::notifyIfEAcceleratorIsUsed();
         self::notifyIfURLIsNotSecure();
+        self::notifyIfDevelopmentModeOnButNotInstalledThroughGit();
 
         $view->topMenu = MenuTop::getInstance()->getMenu();
 
@@ -336,7 +357,7 @@ abstract class ControllerAdmin extends Controller
         /**
          * Posted when rendering an admin page and notifications about any warnings or errors should be triggered.
          * You can use it for example when you have a plugin that needs to be configured in order to work and the
-         * plugin has not been configured yet. It can be also used to cancel / remove other notifications by calling 
+         * plugin has not been configured yet. It can be also used to cancel / remove other notifications by calling
          * eg `Notification\Manager::cancel($notificationId)`.
          *
          * **Example**
--- a/core/Plugin/ControllerAdmin.php
+++ b/core/Plugin/ControllerAdmin.php
@@ -177,6 +177,10 @@ abstract class ControllerAdmin extends Controller
             return;
         }
 
+        if (!Development::isEnabled()) {
+            return;
+        }
+
         if (SettingsPiwik::isGitDeployment()) {
             return;
         }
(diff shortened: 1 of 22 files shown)


plugins/CoreConsole/Commands/GeneratePluginBase.php
```diff
@@ -397,4 +399,10 @@ private function replaceContent($replace, $contentToReplace)
         return $contentToReplace;
     }
 
+    protected function throwErrorIfNotGitInstalled()
```
sgiehl, line 407: instead of calling this method in every class' execute, you could extend the `initialize` method in this base class to check that. It should be called automatically before `execute`

flamisz: I wanted to do that, but there are commands (or at least one, the `development:enable`) where we don't want to throw it. So either I create exception or call one-by-one. But TBH I haven't tried the `initalize`, I used `__construct`. I can give it a go.

flamisz: Thanks for the tip @sgiehl, I couldn't see that `initialize` method. It works, I updated the PR

plugins/CoreConsole/Commands/GeneratePluginBase.php
```diff
@@ -397,4 +404,10 @@ private function replaceContent($replace, $contentToReplace)
...
+    {
+        if (!SettingsPiwik::isGitDeployment()) {
+            throw new NotGitInstalledException('Using this feature Matomo has to be installed through git!');
```
sgiehl: Maybe we could improve this message a bit and also mention the url `https://developer.matomo.org/guides/getting-started-part-1``, which should give more details how to set up for development.

lang/en.json
```diff
@@ -487,6 +487,7 @@
         "WarningFileIntegrityNoMd5file": "File integrity check could not be completed due to missing md5_file() function.",
         "WarningPasswordStored": "%1$sWarning:%2$s This password will be stored in the config file visible to everybody who can access it.",
         "WarningDebugOnDemandEnabled": "Tracker %1$s mode is enabled. For security reasons this should be enabled only for a short time frame. To disable it set %2$s to %3$s in %4$s",
+        "WarningDevelopmentModeOnButNotGitInstalled": "You are currently using Matomo in development mode, but it was not installed through git. It's not recommended to use Matomo in development mode in production environment.",
```
sgiehl, line 490: Not sure if we need this one translatable. Should be a very uncommon case when this is displayed and might be fine to always show the english text.  @tsteur what do you think about that?
If we decide to keep it translatable, we should also add the exception message and the command comment.

flamisz: I just added this into the language file, because it shows on the UI. 
I checked other exceptions and messages in the commands and they are mostly English (at least what I've seen). Kind of tried to follow the actual convention.
But easy to modify to each way, happy to do that.

plugins/CoreConsole/Commands/GeneratePluginBase.php
flamisz: Good point, I'll do it.

lang/en.json
flamisz: @tsteur what your thoughts on this?

plugins/CoreConsole/Commands/GeneratePluginBase.php
flamisz: @sgiehl I added the url to the message, but not sure about how it could be improved. any recommendation?

Currently it is:
`Using this feature Matomo has to be installed through git. For more information please visit https://developer.matomo.org/guides/getting-started-part-1.`

lang/en.json
tsteur: Have no thoughts on this. Both is fine. :) And as both is fine I'd just leave it there with a translation as it's less work to simply keep it compared to removing it but really no preference

plugins/CoreConsole/Commands/GeneratePluginBase.php
sgiehl: I'm not sure if the `installed through git` is that helpful. Downloading the package from GitHub might actually include the require plugin files, but there won't be a `.git/HEAD`, so the command would still raise an error. So actually it would be required to do a full git checkout. Guess it doesn't hurt if the message is a bit more technical, as it actually should be used by developers only.

So maybe something like `This development feature requires Matomo to be checked out from git. For more information please visit https://developer.matomo.org/guides/getting-started-part-1.`

sgiehl (approved):
LGTM now

core/Plugin/ControllerAdmin.php
```diff
@@ -170,6 +171,25 @@ private static function notifyIfURLIsNotSecure()
...
 
+    private static function notifyIfDevelopmentModeOnButNotInstalledThroughGit()
+    {
+        if (!Piwik::hasUserSuperUserAccess()) {
+            return;
+        }
+
+        if (SettingsPiwik::isGitDeployment()) {
+            return;
+        }
```
diosmosis, line 186: I haven't tested locally, so I might be missing something, but shouldn't this also check `if (Development::isEnabled())`?

diosmosis (changes requested):
Saw a possible issue but haven't tested locally

core/Plugin/ControllerAdmin.php
flamisz: you were right @diosmosis, I added the extra validation

sgiehl:
guess should be good to merge now.

PR #17302, opened 2021-03-04
Check if filter limit is -1 when querying log visits

### LINKED ISSUE #17237: Unwanted date filter automatically added to Live.getLastVisitsDetails
I've been working on migrating a system from Piwik 3 to Matomo 4; as part of this, I've been investigating why an API call which worked on the old system isn't working on the new one.

In brief, our system is making the following call:
/index.php?module=API&method=Live.getLastVisitsDetails&idSite=1&token_auth=xxxxxx&filter_limit=-1&segment=visitId%3D%3D12345678

I.e. we're requesting the details of a specific visit from a specific site.

However, on the new system, this was returning zero results.

Having dug into the code, I can see what's triggering this.  In brief, it's because we've set the filter_limit to -1.

```
// plugins/Live/API.php
    private function loadLastVisitsDetailsFromDatabase($idSite, $period, $date, $segment = false, $offset = 0, $limit = 100, $minTimestamp = false, $filterSortOrder = false, $visitorId = false)
    {
        $model = new Model();
        [$data, $hasMoreVisits] = $model->queryLogVisits($idSite, $period, $date, $segment, $offset, $limit, $visitorId, $minTimestamp, $filterSortOrder, true);
        return $this->makeVisitorTableFromArray($data, $hasMoreVisits);
    }

// plugins/Live/Model.php
    public function queryLogVisits($idSite, $period, $date, $segment, $offset, $limit, $visitorId, $minTimestamp, $filterSortOrder, $checkforMoreEntries = false)
    {
        // to check for more entries increase the limit by one, but cut off the last entry before returning the result
        if ($checkforMoreEntries) {
            $limit++;
        }

        // If no other filter, only look at the last 24 hours of stats
        if (empty($visitorId)
            && empty($limit)
            && empty($offset)
            && empty($period)
            && empty($date)
        ) {
            $period = 'day';
            $date = 'yesterdaySameTime';
        }
```

I.e, loadLastVisitsDetailsFromDatabase() is always hardcoding $checkforMoreEntries to true when it calls queryLogVisits().  And that means that queryLogVisits() increments $limit from -1 to 0, which then triggers the "last 24 hours" rule.

For now, and for this specific scenario, I can fix things by setting filter_limit to 1 (which will then be converted to 2 within queryLogVisits), since we're looking for a specific visit and therefore will only ever get one result back.

However, this feels like a logic issue; the documentation (https://developer.matomo.org/api-reference/reporting-api) still states that setting filter_limit to -1 is valid, and if the caller has requested an unlimited number of results, there shouldn't be any additional filters being applied!

It feels like the fix should be to check what value is set for filter_limit before setting $checkforMoreEntries.

flamisz:
### Description:

fixes #17237 

### Review

* [x] Functional review done
* [x] Code review done

### DIFF
--- a/plugins/Live/Model.php
+++ b/plugins/Live/Model.php
@@ -66,7 +66,7 @@ class Model
 
         foreach ($queries as $queryRange) {
             $updatedLimit = $limit;
-            if (!empty($limit)) {
+            if (!empty($limit) && (int)$limit > -1) {
                 $updatedLimit = $limit - count($foundVisits);
             }
 
@@ -96,7 +96,7 @@ class Model
                 $foundVisits = array_merge($foundVisits, $visits);
             }
 
-            if ($limit && count($foundVisits) >= $limit) {
+            if ($limit > 0 && count($foundVisits) >= $limit) {
                 if (count($foundVisits) > $limit) {
                     $foundVisits = array_slice($foundVisits, 0, $limit);
                 }
--- a/plugins/Live/Model.php
+++ b/plugins/Live/Model.php
@@ -43,7 +43,7 @@ class Model
     public function queryLogVisits($idSite, $period, $date, $segment, $offset, $limit, $visitorId, $minTimestamp, $filterSortOrder, $checkforMoreEntries = false)
     {
         // to check for more entries increase the limit by one, but cut off the last entry before returning the result
-        if ($checkforMoreEntries) {
+        if ((int)$limit > -1 && $checkforMoreEntries) {
             $limit++;
         }
 
@@ -104,7 +104,7 @@ class Model
             }
         }
 
-        if ($checkforMoreEntries) {
+        if ((int)$limit > -1 && $checkforMoreEntries) {
             if (count($foundVisits) == $limit) {
                 array_pop($foundVisits);
                 return [$foundVisits, true];
--- a/plugins/Live/Model.php
+++ b/plugins/Live/Model.php
@@ -104,7 +104,7 @@ class Model
             }
         }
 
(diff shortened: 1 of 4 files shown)


flamisz:
I see some tests are failing now, I'll check those.

plugins/Live/Model.php
```diff
@@ -104,7 +104,7 @@ public function queryLogVisits($idSite, $period, $date, $segment, $offset, $limi
             }
         }
 
-        if ($checkforMoreEntries) {
+        if ((int)$limit > -1 && $checkforMoreEntries) {
```
sgiehl: I guess this shouldn't be done here. Let's assume we set `$checkforMoreEntries=true`, then we expect to receive an array like `[array, boolean]`. If `$limit=-1`, we now would return only the result array, which would be kind of unexpected behavior. Simply reverting the change here should do the trick. The number of found visits can't ever be `-1`, so it will in this case always return `[array, false]`, which is correct, as there can't be more visits when looking without limit.

sgiehl (changes requested):

plugins/Live/Model.php
flamisz: I see, thanks. Now I know why the tests are failing. I will look at it tomorrow.

sgiehl:
The changes are looking good I think, but maybe we could add or adjust a test case that queries `Live.getLastVisitsDetails` with `filter_limit=-1` to ensure it returns the correct values.

sgiehl (approved):

sgiehl:
@flamisz  Looks good now. Feel free to merge if you are already done with adding tests.

PR #17303, opened 2021-03-04
Add new wordtruncate twig filter and use in datatable headers

### LINKED ISSUE #17242: Performance report table not displayed properly
reported in https://forum.matomo.org/t/version-4-2-0-darstellungsfehler/40770

This might be a regression of 4.2.0.

The tables in the Behaviour -> Performance reports have a very wide first column when displayed in German (English seems to be fine) causing a horizontal scrollbar to be needed to see all columns.

grafik

flamisz:
### Description:

fixes #17242 

When we have very long words in the `datatable` header rows, it can cause wide columns, so the user has to scroll horizontally to see every data.
With a new twig filter, we can truncate the header string word by word. The whole word and more detailed explanation still viewable by hovering the mouse on the text.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Twig.php
+++ b/core/Twig.php
@@ -619,7 +619,7 @@ class Twig
             $shortenWords = [];
 
             foreach ($words as $word) {
-                $shortenWords[] = strlen($word) > $limit ? substr($word, 0, $limit) . $append : $word;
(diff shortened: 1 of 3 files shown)


flamisz:
Closing for now, as we changing the language file instead.

PR #17304, opened 2021-03-04
archiving: look for all usable done flag types when checking for existing child periods

diosmosis:
### Description:

When checking for child periods we can use when archiving there might be invalidated archives that could be used which we should also detect. This mostly affects sites that use the tracker and can have no visits, but day archives with data, and no other higher archives. To my knowledge this just includes sites imported from google.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/DataAccess/Model.php
+++ b/core/DataAccess/Model.php
@@ -847,7 +847,12 @@ class Model
         while ($date->isEarlier($period->getDateEnd()->addPeriod(1, 'month'))) {
             $archiveTable = ArchiveTableCreator::getNumericTable($date);
 
+            // we look for any archive that can be used to compute this one. this includes invalidated archives, since it is possible
+            // under certain circumstances for them to exist, when archiving a higher period that includes them. the main example being
+            // the GoogleAnalyticsImporter which disallows the recomputation of invalidated archives for imported data, since that would
+            // essentially get rid of the imported data.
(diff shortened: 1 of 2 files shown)


core/DataAccess/Model.php
```diff
@@ -847,9 +847,10 @@ public function hasChildArchivesInPeriod($idSite, Period $period)
         while ($date->isEarlier($period->getDateEnd()->addPeriod(1, 'month'))) {
             $archiveTable = ArchiveTableCreator::getNumericTable($date);
 
+            $usableDoneFlags = [ArchiveWriter::DONE_OK, ArchiveWriter::DONE_INVALIDATED, ArchiveWriter::DONE_PARTIAL, ArchiveWriter::DONE_OK_TEMPORARY];
```
sgiehl, line 854: Using partial or temporary archives kind of makes sense for me, but why are we also using invalidated archives here?
Shouldn't the archiving be triggered for invalidated archives, that can be used then?

diosmosis: @sgiehl they should normally be rearchived before this archive gets launched (ie, weeks before months), but in the specific case of the GoogleAnalyticsImporter they don't have to be, since the plugin does not allow rearchiving of days w/ imported data (to make sure the imported data doesn't disappear through user or programming error). So in this case the days that are invalidated get ignored. It might be a better solution in the future to disallow invalidating them entirely...

sgiehl: Might be good to add a small comment above the line then, so it's clear why we also use invalidated archives here. Just to prevent we remove it later by accident.

sgiehl:
@diosmosis besides adding a comment to explain the behavior a bit, this one looks good to merge

PR #17315, opened 2021-03-08
Only allow valid hosts for prefixUrl in API listing

sgiehl:
### Review

* [ ] Functional review done
* [ ] Code review done


plugins/API/Controller.php
```diff
@@ -64,7 +64,8 @@ public function listAllMethods()
 
         $ApiDocumentation = new DocumentationGenerator();
         $prefixUrls = Common::getRequestVar('prefixUrl', 'https://demo.matomo.org/', 'string');
-        if (!UrlHelper::isLookLikeUrl($prefixUrls) || strpos($prefixUrls, 'http') !== 0) {
+        $hostname = @parse_url($prefixUrls, PHP_URL_HOST);
```
flamisz: The Error Suppression operator works differently in PHP 8. There is a possibility it will cause some issue later.

https://www.php.net/manual/en/language.operators.errorcontrol.php
 
Is it necessary here?

sgiehl: Actually I guess it shouldn't be needed. Will remove it.

sgiehl:
```suggestion
        $hostname = parse_url($prefixUrls, PHP_URL_HOST);
```

flamisz (approved):
LGTM

PR #17320, opened 2021-03-09
Throw invalid dimension exception during rendering graph with wrong height/width values

### LINKED ISSUE #17092: fatal error: Modulo by zero: method=imageGraph while widht = 0
I called the API by this:
/index.php?module=API&method=ImageGraph.get&idSite=1&token_auth=xxxxxxxxxxxxx&period=month&date=previous24&apiModule=VisitsSummary&apiAction=get&graphType=evolution&width=0&height=2

and got this fatal error message

grafik

width = 0 AND height != 0

width = 0 and height = 0 => it works fine.

flamisz:
### Description:

fixes #17092 

The package we use for rendering graphs has a bug. To avoid run into that bug, we can check the values we send to the package and if the values would cause the modulo zero error we throw an exception instead. 
Our plugin throws the exceptions as an image by default, but for this error it's not that useful, because usually it happens when the height value is too small, so the user wouldn't be able to read the error message. 
For this specific exception I used the default way.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/ImageGraph/API.php
+++ b/plugins/ImageGraph/API.php
@@ -14,6 +14,7 @@ use Piwik\Archive\DataTableFactory;
 use Piwik\Common;
 use Piwik\Container\StaticContainer;
 use Piwik\DataTable\Map;
+use Piwik\Exception\InvalidDimensionException;
 use Piwik\Filesystem;
 use Piwik\Period;
 use Piwik\Piwik;
@@ -493,6 +494,8 @@ class API extends \Piwik\Plugin\API
 
             // render graph
             $graph->renderGraph();
+        } catch (InvalidDimensionException $e) {
+            throw new Exception('Error: the graph dimension is not valid. Please try larger width and height values or use 0 for default values.');
         } catch (\Exception $e) {
 
             $graph = new \Piwik\Plugins\ImageGraph\StaticGraph\Exception();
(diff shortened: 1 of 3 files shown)


sgiehl:
Wondering if it might make more sense to define lower limits for height & weight and automatically set those parameters to the lower limit if they are smaller. That way we could avoid any exception at all.

flamisz:
> Wondering if it might make more sense to define lower limits for height & weight and automatically set those parameters to the lower limit if they are smaller. That way we could avoid any exception at all.

Hey @sgiehl, me and @tsteur were talking about that, and I could calculate that, but if the user wants a lower number, and the graph will be higher, maybe she/he will think it is a bug. Hard to tell which is the better solution, but probably it is something that happens very rarely.

tsteur:
@sgiehl the thought was that if we automatically change a height or width then it will look like a bug and we're getting bug reports and emails about it vs this way it's clear what the problem is and they can change the parameters to a higher number.

sgiehl:
Thought we would have it that way in other places, but wasn't able to find one right now. But it's also fine this way.

plugins/ImageGraph/API.php
```diff
@@ -493,6 +494,8 @@ public function get(
 
             // render graph
             $graph->renderGraph();
+        } catch (InvalidDimensionException $e) {
+            throw new Exception('Error: the graph dimension is not valid. Please try larger width and height values or use 0 for default values.');
```
sgiehl: Any reason for not simply re-throwing the caught exception and setting the message where the exception actually occurs?

flamisz: You're right, that's better solution 👍 , I'll make it like that.

sgiehl (approved):

PR #17321, opened 2021-03-09
Make sure not to clear the tracker cache so often when invalidating in core:archive

diosmosis:
### Description:

There can potentially be a lot of invalidating of the tracker cache since we call ArchiveInvalidator functions one after the other. This PR introduces a utility function that collects all tracker cache clears and calls them just once after a callback finishes executing.

### Review

* [ ] Functional review done
* [ ] Code review done


core/Tracker/Cache.php
```diff
@@ -240,14 +258,63 @@ public static function regenerateCacheWebsiteAttributes($idSites = array())
...
+            self::$delegatingCacheClears = true;
+            self::$delegatedClears = [];
+
+            $result = $callback();
+
+            self::callAllDelegatedClears();
+
+            return $result;
+        } finally {
+            self::$delegatingCacheClears = false;
```
tsteur: would we maybe call `self::callAllDelegatedClears();` as part of finally in case there is an exception but some cache clear calls are needed?

sgiehl (approved):
Not sure if that new method would be worth to be mentioned in the developer changelog. Otherwise looks good

tsteur:
It's not public API so no need to mention it 👍

tsteur:
fyi @diosmosis 2 tests still failing but maybe they are not due to this PR? https://travis-ci.com/github/matomo-org/matomo/jobs/489891666#L806

PR #17326, opened 2021-03-10
Add --skip-cache-clear option to core:update command  …

diosmosis:
### Description:

…so we don't have to clear caches when updating on an empty ec2 instance (for example). This PR uses a static member var to stop cache clears for the entire process since it is done once in the command, once in the CoreUpdater, and possibly multiple times in the plugin manager (if plugins are activated/deactivated for instance).

FYI @tsteur 

### Review

* [ ] Functional review done
* [ ] Code review done


core/Filesystem.php
```diff
@@ -20,12 +20,22 @@
...
     /**
      * Called on Core install, update, plugin enable/disable
      * Will clear all cache that could be affected by the change in configuration being made
      */
     public static function deleteAllCacheOnUpdate($pluginName = false)
     {
+        if (self::$skipCacheClearOnUpdate) {
+            return;
+        }
+
```
sgiehl, line 38: Shouldn't we reload the plugin information nevertheless? Wondering if it might cause weird issues when a plugin is updated, but the information isn't reloaded.

diosmosis: :+1: that makes sense

diosmosis: skipping this change as discussed on slack

PR #17333, opened 2021-03-11
Add event tags to top and bottom of body

MHarmony:
### Description:

I am currently developing a plugin that appends html to the top and bottom of the body. These two tags would be helpful in my plugin development.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Morpheus/templates/layout.twig
+++ b/plugins/Morpheus/templates/layout.twig
@@ -31,6 +31,7 @@
         {% endblock %}
     </head>
     <body id="{{ bodyId|default('') }}" ng-app="app" class="{{ bodyClass|default('') }}">
+        {{ postEvent('Template.bodyTop' ) }}
 
     {% block body %}
 
@@ -53,5 +54,7 @@
         </div>
 
         {% include "@CoreHome/_adblockDetect.twig" %}
+
+        {{ postEvent('Template.bodyBottom' ) }}
     </body>
 </html>


sgiehl:
Hi @MHarmony  thanks for creating the PR. The changes are looking fine.

@tsteur could you maybe have a quick look as well. Not sure if should add those new events or not. In general we should maybe try to avoid events where possible, as they might make Matomo slower and slower if we add too many. But guess those global events might be helpful at some point maybe.

tsteur:
I think it be fine to add them 👍  What would always be helpful be to know how/for what they will be used @MHarmony . This will make it easier if/when there are changes/refactorings/... and what could break when making changes in the future there. If you can't tell us it's fine but it be good to know for us since for these twig events we don't consider API and we'd still want to make sure when/if there are any changes in the future to not break anything.

MHarmony:
It's for notices at top/bottom of pages.

tsteur:
fyi @MHarmony I just documented our notifications classes. Just wondering if this might maybe work too? https://github.com/matomo-org/developer-documentation/pull/459/files

MHarmony:
@tsteur Unfortunately not. The notices are meant to be permanent, not toasts. Like a banner of sorts.

tsteur:
The notifications can be permanent too if triggered on every page load but all good. Just thought I mention it 👍

MHarmony:
@tsteur Thanks for the mention. I didn't know about these so this gives me something to use in the future even if it doesn't quite fit what I need here.

PR #17334, opened 2021-03-11
Fix check for invalid image sizes

sgiehl:
### Description:

Seems I've merged #17320 too early. Afterwards some system tests generating image started to fail, as an exception was thrown even though the dimension should be valid.

Debugged a bit through the code, and I think the new condition should prevent any modulo error but still show an image if it is possible.

@flamisz  maybe you could have a quick look and merge if it's fine.

### Review

* [ ] Functional review done
* [ ] Code review done


flamisz (approved):
Thanks for the fix.

I did a quick test and it works fine for me.

PR #17345, opened 2021-03-15
Require password confirmation for more plugin operations.

diosmosis:
### Description:

Adding requirement for password confirmation for: activating a plugin, deactivating a plugin, uninstalling a plugin. And for CorePluginsAdmin.setSystemSettings.

### Review

* [ ] Functional review done
* [ ] Code review done


flamisz:
Is it common for asking password as a URL parameter when we already have the token validation? Is it necessary? I can't remember using API like this.

diosmosis:
@flamisz if you grep through API files for 'passwordConfirmation', you'll find similar methods. It is an extra security precaution for sensitive settings and changes.

Also just realized I forgot to the UI related changes, will move this back to a draft.

diosmosis:
Added the UI changes. Also noticed we were using the old `ready:` event for materializecss modals, should be using `onOpenEnd` now.

plugins/CorePluginsAdmin/Controller.php
```diff
@@ -425,9 +425,22 @@ public function safemode($lastError = array())
...
     {
-        $pluginName = $this->initPluginModification(static::ACTIVATE_NONCE);
         $this->dieIfPluginsAdminIsDisabled();
 
+        $params = [
+            'module' => 'CorePluginsAdmin',
+            'action' => 'activate',
+            'pluginName' => Common::getRequestVar('pluginName'),
+            'nonce' => Common::getRequestVar('nonce'),
+            'redirectTo' => Common::getRequestVar('redirectTo'),
```
sgiehl, line 435: That actually seems not to work as expected. `redirectTo` might be set to `referer`. In this case the method redirects to the referrer sent with the request. As the password verify now adds another redirect in between, the referrer isn't the original anymore, causing a loop on the confirmation page.

diosmosis: :+1: should be fixed

diosmosis:
@sgiehl this should be ready for another review

sgiehl (approved):
Seems to work as expected now. Some UI tests for TagManager and QueuedTracking are failing due to the new password confirmation. Might be good to fix them before merging

PR #17349, opened 2021-03-16
Ensure site is selected in siteselector if only one site is available

sgiehl:
### Description:

When only one site is available the site selector does no longer provide the possibility to choose one. In this case this site should be automatically set as the selected one.

regression from #16972

### Review

* [ ] Functional review done
* [ ] Code review done


flamisz (approved):
LGTM, I tested locally. First I didn't get it, how can it happen, but now I see: when we have multiple sites, but the active user has access only for one.

PR #17357, opened 2021-03-18
Ensure login allow list is not checked for authenticated tracking requests

### LINKED ISSUE #66: Tracking failure when using plugin in Matomo 4.1.1
When using the Visitor Generator plugin in Matomo 4.1.1 the following tracking error occurs:

Measurable | Problem | Solution | Date | Page URL | Tracking URL | Action
-- | -- | -- | -- | -- | -- | --

Hilpert, Flatley and Bashirian (Id 2) | Request was not authenticated but authentication was required. | Set or correct a "token_auth" in your tracking request. Learn more | Feb 22, 2021 00:26:24 | http://example.org/?utm_campaign=bsa&utm_term=buildinternet | action_name=ExampleCode+-+technology+with+passion&idsite=2&rec=1&r=299079&h=7&m=17&s=55&url=http%3A%...

image

No errors are returned when running the console command to generate the visit data
image

No data is generated for the reports in Matomo 
image

Using the latest 4.0.3 plugin version.

sgiehl:
### Description:

fixes https://github.com/matomo-org/plugin-VisitorGenerator/issues/66

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CoreHome/CoreHome.php
+++ b/plugins/CoreHome/CoreHome.php
@@ -45,7 +45,7 @@ class CoreHome extends \Piwik\Plugin
             'AssetManager.filterMergedJavaScripts'   => 'filterMergedJavaScripts',
             'Translate.getClientSideTranslationKeys' => 'getClientSideTranslationKeys',
             'Metric.addComputedMetrics'              => 'addComputedMetrics',
-            'Request.initAuthenticationObject' => 'initAuthenticationObject',
+            'Request.initAuthenticationObject' => 'checkAllowedIpsOnAuthentication',
             'AssetManager.addStylesheets' => 'addStylesheets',
             'Request.dispatchCoreAndPluginUpdatesScreen' => 'initAuthenticationObject',
             'Tracker.setTrackerCacheGeneral' => 'setTrackerCacheGeneral',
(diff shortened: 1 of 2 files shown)


tests/PHPUnit/System/TrackerDisallowedIpTest.php
```diff
@@ -0,0 +1,53 @@
...
+        Fixture::createWebsite('2014-02-04');
+        Fixture::createSuperUser(false);
+    }
+
+    public function test_authenticateSuperUserOrAdmin_ShouldWorkWithIpNotOnLoginWhitelist()
+    {
+        $tracker = Fixture::getTracker(1, '2021-02-02 16:00:00', $defaultInit = true, $useLocalTracker = false);
+        $tracker->setTokenAuth(Fixture::getTokenAuth());
+        Fixture::checkResponse($tracker->doTrackPageView('test'));
+    }
```
diosmosis, line 57: Would it be a good idea to also test that the ip is blocked when not authenticated?

And would it be good to check a visit was inserted, or do you think that's unnecessary?

plugins/CoreHome/CoreHome.php
```diff
@@ -75,9 +75,14 @@ public function addStylesheets(&$mergedContent)
 
     public function initAuthenticationObject()
```
diosmosis: This unrelated to this PR, but this function name confused me, combined w/ the `return;` confused me. It might be better to rename it to something like `checkAllowedIpsOnAuthentication` or something.

diosmosis (changes requested):

tests/PHPUnit/System/TrackerDisallowedIpTest.php
sgiehl: @diosmosis The ip should not be blocked even if not authenticated. I've added/improved the tests

PR #17360, opened 2021-03-19
Fix remaining regressions & ui tests

sgiehl:
### Description:

While checking the remaining failing UI tests I've detected some more regressions from the latest merged PRs

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
+++ b/plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
@@ -24,6 +24,8 @@
 
         var showSubcategoryHelpOnLoad = null;
 
+        var initialLoad = true;
+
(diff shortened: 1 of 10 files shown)


plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
```diff
@@ -218,7 +220,9 @@
         });
 
         $rootScope.$on('piwikPageChange', function (event) {
-            globalAjaxQueue.abort();
+            if (!initialLoad) {
```
sgiehl, line 223: This should ensure that when the page loads initially it doesn't abort any ajax requests. Otherwise this makes problems when directly loading a popover with url params. Some UI tests where failing due to aborted requests.

sgiehl:
Seems all tests are passing again 🎉

PR #17367, opened 2021-03-22
Improve no data tracked yet screen

### LINKED ISSUE #16787: Improve no data tracked yet please set up tracking code screen
The goal of this issue is to improve the "No data has been recorded yet, get set up below" screen which users see when they haven't tracked any data yet.

So far we're showing seven different ways of embedding the tracking code and there's a lot of text and it is quite tricky for people to figure out what to do.

## Step 1: detection of site

We currently have guides for WordPress | Squarespace | Wix | SharePoint | Joomla | Shopify. When a site is created, can we try to fetch the content of the site and try to detect what CMS is being used? Eg WordPress would typically have some URLs with `wp-content/` in the HTML. Important be to have a timeout of say 5 seconds as some URLs won't be reachable (eg internet is disabled or many other reasons) and we don't want to delay the site creation too much.

If any of these CMS was detected, then we could say something like "It seems you're wanting to track a WordPress website, check out our integration guide for WordPress websites to get started".

Additionally, if we detect they use Google Tag Manager then we should also offer a new guide

## Step 2: Use of tabs

If we detected a system that we have a guide or integration for, then we expand the Integrations tab by default. Otherwise we show "Tracking code" first

The Matomo Tag Manager embed instructions should include already the JS code to embed the MTM into their website.

In the intro text there is a distracting like to "tracking code" page which we remove and instead we have a link to customise the tracking code as part of the tracking code tab section. And we add a link to simply email developers (it's already there but we simplify this).

A "other" group would list mobile apps, log analytics, and other ones that are more rarely used. It should again mention Google Tag Manager as part of the other section.

Could look like screen below but needs to be optimised as those tabs are not quite readable or noticeable that it's tabs maybe.
image

Note: currently not visible there is the button to "hide this message for 1 hour" which should be still there.

flamisz:
### Description:

fixes #16787 

### Review

* [ ] Functional review done
* [ ] Code review done


flamisz:
Couple of example websites I used trying different CMSs:

type | site
-----|----
matomo.org - WP | http://matomo.org	
squarespace | https://est-bostoncopywriter.com
wix | https://www.thegrilledcheesefactory.fr
joomla | https://www.nintendo.se	
shopify + GTM | https://www.manitobah.com
sharepoint | https://incometaxindia.gov.in

flamisz:
Hi @tsteur, 
the current page checks the WhiteLabel plugin and hides URLs (not all tho) to `matomo.org`. How strictly we'd like to follow that in the new solution? All the links that go to `matomo.org`? That way we can't show any instructions/guides from the FAQ. And what about links from the development documentation?

These are the links I use now:
- CMS integration guides
- GTM guide
- the `/integrate` page from `matomo.org`
- tracking dev documentation
- tag manager guide
- and guides for the other pages (log analytics, SDKs, HTTP tracking guide)

Thanks, Zoltan

tsteur:
The links to matomo.org should be fine (they were basically there before already). Also links to the developer website should be no problem @flamisz

plugins/CoreHome/angularjs/content-block/content-block.directive.js
```diff
@@ -83,8 +83,13 @@
                         }
                     }
 
+                    var tabs = angular.element('.tabs');
```
diosmosis: Is this code meant to initialize every `.tabs` element on the page or every element within the content block? If every element on the page I think this might be better put somewhere else.

plugins/SitesManager/Controller.php
```diff
@@ -158,14 +159,68 @@ public function siteWithoutData()
...
+        }
+
+        $needle = 'X-Wix-Published-Version';
+        if (strpos($response['data'], $needle) !== false) {
+            return [SitesManager::SITE_TYPE_WIX, $gtmUsed];
+        }
+
+        if ($response['headers']['expires'] === 'Wed, 17 Aug 2005 00:00:00 GMT') {
+            return [SitesManager::SITE_TYPE_JOOMLA, $gtmUsed];
+        }
```
diosmosis: Is there a source or reference that shows why this works? If so would it help to link to it in a comment here?

plugins/SitesManager/Controller.php
```diff
@@ -158,14 +159,68 @@ public function siteWithoutData()
...
             'showMatomoLinks' => $showMatomoLinks,
+            'siteType' => $siteType,
+            'instructionUrl' => $instructionUrl,
+            'gtmUsed' => $gtmUsed,
             'googleAnalyticsImporterMessage' => $googleAnalyticsImporterMessage,
+            'tagManagerActive' => $tagManagerActive,
         ), $viewType = 'basic');
     }
+
+    private function guessSiteTypeAndGtm()
```
diosmosis: This is kind of a simple method, but would be a good idea to unit test it?

Also I'm wondering if this could be done via AJAX? I'm not sure if it can exactly, but if so, it wouldn't affect page load time which might be better?

diosmosis (changes requested):
Tested locally, works, left a couple comments

plugins/SitesManager/Controller.php
flamisz: yes, good idea, I will put the reference there.

plugins/CoreHome/angularjs/content-block/content-block.directive.js
flamisz: good question. I double check it. I wanted to put the `$('.tabs').tabs();` to the page in a script tag, but even if I put inside the `document ready` the html was empty, so could find the tabs... it was annoying...
but I check it again.

plugins/SitesManager/Controller.php
flamisz: I was thinking about the test. I'll do it.

I check the AJAX call.

plugins/CoreHome/angularjs/content-block/content-block.directive.js
diosmosis: You could use an angular.run block, I guess, but not all the page content will be filled out by then and it wouldn't help for ajax... It might require a new attribute directive that's used on elements like `<div piwik-tabs>`. Otherwise, could just do it for this directive by injecting `$element` and using that (ie, `$element.find('.tabs')`).

flamisz: can you please check if this is what you thought?

plugins/SitesManager/Controller.php
flamisz: Currently, the controller renders a twig template, without any angular controller and uses some angular components (like the `piwik-content-block`). 

We can use a script tag, to make that ajax call (we use script tag checking every second if we have data), but that's not very nice, and DOM is not ready, etc, or I guess we can make a small angular component inside this plugin.

Do you want me to make it and use AJAX instead of the back-end?

plugins/CoreHome/angularjs/content-block/content-block.directive.js
diosmosis: that works :+1:

plugins/SitesManager/Controller.php
diosmosis: That depends on how much work you think it will be... if it's super complicated probably better not to do it for this feature since there are probably better things to work on. In that case, though, it would be good to see if there's a way to avoid having the page load time become 5+ seconds if the site is unresponsive for some reason. I guess we could cache the result of this function... if you choose to do that, take a look at the `Lazy` cache class.

(I'm not sure if there are docs about this or you already know, but there are three of these cache classes in Matomo:

- `Transient`: stored in memory for the current request only.
- `Lazy`: stored in a backend of some kind, can be file or redis or something else. The backend is loaded on request, so we only read the file or connect to redis when the specific data is requested.
- `Eager`: everything in this cache is loaded at the start of each request, so the data is readily available.
).

flamisz: @tsteur showed me `piwik-widget-loader`, I modified the PR to use that, loading the whole tabs part

flamisz:
I'll add the test for that cms type guess method.

flamisz:
As you can see there is a linked PR, matomo-org/tag-manager#307. I needed that, because inside the tabs, when we show the `tagmanager` code, without that change the user has to change the site and then change back, to show the containers.

But now I've found another issue. When the tabs are loaded, using the `widget-loader` and calls the tagmanager code includes the siteselector, and that throws an error from here:
https://github.com/matomo-org/matomo/blob/eb1a6ba1ff8a0a486f963eb4087714e94a9a4d12/plugins/CoreHome/angularjs/siteselector/siteselector.controller.js#L16

I remembered @sgiehl modified that part a couple of weeks ago. If I modify back to the original code, it works. It looks like the issue is that...

Ok, nothing (but I comment it anyway 😄 ) I see there is a new commit from @sgiehl and it has been fixed...
Is it ok if I rebase my branch?

diosmosis:
> Is it ok if I rebase my branch?

This would never be an issue :+1:

plugins/SitesManager/Controller.php
```diff
@@ -146,6 +148,34 @@ public function siteWithoutData()
...
+    public function siteWithoutDataTabs() {
+        try {
+            $response = Http::sendHttpRequest($this->site->getMainUrl(), 5, null, null, 0, false, false, true);
+        } catch (Exception $e) {
+            $response = false;
+        }
+        $guesser = new GtmSiteTypeGuesser();
+        $siteType = $guesser->guessSiteTypeFromResponse($response);
+        $gtmUsed = $guesser->guessGtmFromResponse($response);
+        $instructionUrl = SitesManager::getInstructionUrlBySiteType($siteType);
```
diosmosis, line 191: Can we cache `$siteType` / `$gtmUsed` so we don't send an HTTP request every time this page is viewed? Here's one example of how the cache is used: https://github.com/matomo-org/matomo/blob/fd14a155269038f6fe12ce31db20f8ebffaf00c5/plugins/RssWidget/RssRenderer.php#L52

diosmosis (changes requested):

plugins/SitesManager/Controller.php
```diff
@@ -24,12 +24,23 @@
...
 class Controller extends \Piwik\Plugin\ControllerAdmin
 {
+    /** @var Lazy */
+    private $cache;
+
+    public function __construct() {
+        $this->cache = Cache::getLazyCache();
+
+        parent::__construct();
+    }
```
diosmosis, line 41: Is this constructor necessary? Would it work to inject it like:

```
/** @var Lazy */
private $cache;

public function __construct(Lazy $cache)
{
    $this->cache = $cache;
    parent::__construct();
}
```
?

flamisz: yes, you're right, it works

PR #17368, opened 2021-03-22
Allow rearchive_reports_in_past_last_n_months setting value to be an int.

### LINKED ISSUE #17366: Improve configuration of rearchive_reports_in_past_last_n_months setting
`rearchive_reports_in_past_last_n_months` already has `last_n` in the name but still expects to configure `last4` as value for example. Goal of this issue is that 

* If only a number is configured, we skip the `substr` https://github.com/matomo-org/matomo/blob/4.2.1/core/Archive/ArchiveInvalidator.php#L799
* We also no longer hard code the `4` (which comes from `strlen('last')`) but would instead do something like `str_replace('last', '', $lastNMonthsToInvalidate)` which be more readable.

Note same code seems to be also in 2 places in Custom Reports where it would needs to be fixed too

diosmosis:
### Description:

Fixes #17366

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -796,9 +796,11 @@ class ArchiveInvalidator
             return null;
         }
 
-        $lastNMonthsToInvalidate = (int) substr($lastNMonthsToInvalidate, 4);
-        if (empty($lastNMonthsToInvalidate)) {
-            return null;
+        if (!is_numeric($lastNMonthsToInvalidate)) {
+            $lastNMonthsToInvalidate = (int)str_replace('last', '', $lastNMonthsToInvalidate);
+            if (empty($lastNMonthsToInvalidate)) {
+                return null;
+            }
         }
 
         return Date::yesterday()->subMonth($lastNMonthsToInvalidate)->setDay(1);
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -786,7 +786,7 @@ enable_tracking_failures_notification = 1
 
 ; Controls how many months in the past reports are re-archived for plugins that support
 ; doing this (such as CustomReports). Set to 0 to disable the feature. Default is 6.
-rearchive_reports_in_past_last_n_months = last6
+rearchive_reports_in_past_last_n_months = 6
 
 ; If set to 1, when rearchiving reports in the past we do not rearchive segment data with those reports. Default is 0.
 rearchive_reports_in_past_exclude_segments = 0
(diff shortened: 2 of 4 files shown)


flamisz:
Do we want to change the default settings in the `global.ini.php`. Currently, it is `last6`.

What about the other files where we use this setting?  We still check for `substr` in `plugins/CustomReports/Controller.php` or `plugins/CustomReports/Widgets/GetManageReports.php`.

```php
if (isset($general['rearchive_reports_in_past_last_n_months'])) {
    $reArchiveLastN = (int) substr($general['rearchive_reports_in_past_last_n_months'], 4);
}
```

diosmosis:
> Do we want to change the default settings in the global.ini.php. Currently, it is last6.

this is up to @tsteur 

>  We still check for substr in plugins/CustomReports/Controller.php or plugins/CustomReports/Widgets/GetManageReports.php.

A PR exists for those now.

tsteur:
Either `last6` or `6` is fine regarding the default setting. No preference.

flamisz (approved):
LGTM

diosmosis:
thanks for the quick review @flamisz :+1:

core/Archive/ArchiveInvalidator.php
```diff
@@ -796,9 +796,11 @@ private function getEarliestDateToRearchive()
             return null;
         }
 
-        $lastNMonthsToInvalidate = (int) substr($lastNMonthsToInvalidate, 4);
-        if (empty($lastNMonthsToInvalidate)) {
-            return null;
+        if (!is_numeric($lastNMonthsToInvalidate)) {
+            $lastNMonthsToInvalidate = (int)str_replace('last', '', $lastNMonthsToInvalidate);
```
sgiehl, line 800: In theory this would also allow values like  `last2last1` and would result in `21`. Not sure if we wouldn't want to discard such incorrect values instead.

diosmosis: cc @tsteur

tsteur: I think it's not needed and could keep things simple. People will mostly just configure a number anyway.

nelhefni:
@diosmosis thanks for the fix! This should help to fix this issue: https://forum.matomo.org/t/new-funnel-triggers-past-report-generation/41152/4
Before your fix, `rearchive_reports_in_past_last_n_months = 0` would throw an exception, because `empty(0)` is true, meaning`getEarliestDateToRearchive()` was returning `null` instead of the current month, leading to an issue in the calling method.

That being said, there is still a problem. If I set `rearchive_reports_in_past_last_n_months = 0` I would expect no past report to be generated when I update a Funnel, but because of `setDay(1)` we will always re-generate report since the first day of the current month.
https://github.com/matomo-org/matomo/blob/4.x-dev/core/Archive/ArchiveInvalidator.php#L806

diosmosis:
good catch @nelhefni, i'll fix this today

PR #17369, opened 2021-03-22
allow forcing backtrace to print in exceptiontotextprocessor

diosmosis:
### Description:

Allow forcing the backtrace to print in ExceptionToTextProcessor + a fix: remove the redundant message that gets printed when including the backtrace. The processor printed the message + the stack trace, but the stack trace includes the messages of all exceptions, so it's redundant. We could also always use the result of `getMessageAndWholeBacktrace()`, but I didn't want to change that much code for a small change.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Monolog/Processor/ExceptionToTextProcessor.php
+++ b/plugins/Monolog/Processor/ExceptionToTextProcessor.php
@@ -100,12 +100,13 @@ class ExceptionToTextProcessor
         }
 
         if (is_array($exception)) {
+            $message = $exception['message'] ?? '';
             if ($shouldPrintBacktrace && isset($exception['backtrace'])) {
                 $trace = $exception['backtrace'];
                 $trace = self::replaceSensitiveValues($trace);
-                return $trace;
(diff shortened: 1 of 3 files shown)


sgiehl (commented):
The code itself looks good. 
@diosmosis could you have a look at the method `getMessageAndWholeBacktrace`? I'm wondering if it is on purpose that the message actually is not included if `$exception` is given as array. Not sure when that is the case, but imho it should also include the message in that case, right?

diosmosis:
I remember, that was to keep it backwards compatible w/ some uses, probably fatal error reporting. I'll take a look.

diosmosis:
Can't see an array ever being used w/ this method, so I'm just going to change it and see if any tests fail. Worst case, the message may get printed out twice in some rare circumstances, so I think it's ok.

sgiehl:
@diosmosis there are various log related tests failing

sgiehl (approved):

PR #17370, opened 2021-03-22
French country map: Add ISO 3166-2 new codes + update regions

### LINKED ISSUE #15203: France regions are not up-to-date
Hello,

France region in Matomo are
image

Whereas several regions have been merged since the 01/01/2016 (official source : http://www.prefectures-regions.gouv.fr )
image

Regards.

ysard:
### Description:

Hi I partially reverse-engineered the `/plugins/UserCountryMap/svg/FRA.svg` file to update it in accordance with the regrouping of French regions since January 1, 2016. namely:
- removal of old regions
- addition of ISO 3166-2 codes in the additional attribute (`data-p`) of  `<path>` objects, as it is the case for other countries without any problem of ISO code. It's a trick but it works.

See the [Wikipedia page](https://fr.wikipedia.org/wiki/ISO_3166-2:FR) for more info about this administrative update.
Finally, we have 13 metropolitan regions instead of 22.

Editing was made with Inkscape. The method is not reproducible due to the lack of a functional map generator,
but I suppose that it's better than nothing...
However, it should be noted that even Natural Earth Data is not up to date on this point.
See the issue [here](https://github.com/nvkelso/natural-earth-vector/issues/429)

This closes issue #15203 and is related to #11929.

Demonstration image: 
image

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (approved):
Hi @ysard 
Thanks for creating the PR. Very appreciated.
Tested it locally and it works nicely 👍

Issue #17371, opened 2021-03-22
Ensure performance timings are sent as integer values


MichaelRoosz:
I am currently testing Matomo 4.2.1.

The new Performance Report feature is not working at all:

- in matomo_log_link_visit_action the only field with non-NULL data is time_network and its always 0
- the other fields are always NULL
- -> Reports are empty

This is what the browsers are sending:

chrome
pf_net: 0
pf_srv: 1.150000000052387
pf_tfr: 1.2499999993451638

firefox
pf_net: 0
pf_srv: 2
pf_tfr: 1

- floats in chrome seem strange as the backend seems to expect integers
- the values 0/1/2 look like seconds but I guess we want milliseconds?

-> the performance feature seems completely broken at this time

sgiehl:
@MichaelRoosz thanks for creating the issue. Actually both values should be in milliseconds.
I'll prepare a PR to round the values, so we use integer in all cases. 
Not sure if the float values might cause problems when storing them in database..

diosmosis:
Might be related to https://github.com/matomo-org/matomo/issues/17259

sgiehl:
There had been some fixed aroud the Performance Timing tracking. I assume this one should have been fixed as well.
@MichaelRoosz If you still have an issue with this, please let us know, so we can reopen the issue. Thanks.

PR #17372, opened 2021-03-22
Ensure to send performance timings as integer values

sgiehl:
### Description:

refs #17371

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl:
build js

diosmosis:
Should we maybe do this server side as well to be safe in case people use the tracker API directly or use log importing?

diosmosis (changes requested):
If adding server side changes, can we also change a test so it uses a float value in tracking?

sgiehl:
@diosmosis looking at the code it seems the values should actually already be casted to integer server side. At least they are defined as those here:
https://github.com/matomo-org/matomo/blob/23d551750442d12975ee9c2af6f73ac9bbf75c1a/core/Tracker/Request.php#L415-L420

So the changes here actually won't have a big effect. But it's still useless to send floats if we use them as integers...

diosmosis:
makes sense :+1:

PR #17380, opened 2021-03-23
Show notification when there is raw data but no data

### LINKED ISSUE #16790: When a period has no data, but raw data, show a message
Conditions when to apply this:
* For now only apply when "no segment" is select (report shows all visits) 

When loading the reporting UI, or when selecting a different date, can we check in an API call if `nb_visits=0`? If that's the case, we also check if there's any visit during that period in log tables raw data. If that's the case, we could show an info message like this:

image

Where click on visits log would move the user to the "visits log". Could we also have a link that opens the date selector? Wording likely needs to be improved (didn't think too much about exact wording).

Maybe we would only do this extra API call if the date includes "today" as in other cases it should not be needed.

flamisz:
### Description:

fixes #16790 

### Review

* [x] Functional review done
* [x] Code review done

### DIFF
--- a/plugins/CoreHome/angularjs/common/services/periods.js
+++ b/plugins/CoreHome/angularjs/common/services/periods.js
@@ -26,6 +26,7 @@
  * - **getDateRange()**: returns an array w/ two elements, the first being the start
  *                       Date of the period, the second being the end Date. The dates
  *                       must be Date objects, not strings, and are inclusive.
+ * - **containsToday()**: returns true if the date period contains today. False if not.
  * - (_static_) **parse(strDate)**: creates a new instance of this period from the
  *                                  value of the 'date' query parameter.
  * - (_static_) **getDisplayText**: returns translated text for the period, eg, 'month',
@@ -64,6 +65,10 @@
 
         getDateRange: function () {
             return [new Date(this.dateInPeriod.getTime()), new Date(this.dateInPeriod.getTime())];
+        },
+
+        containsToday: function () {
+            return todayIsInRange(this.getDateRange());
         }
     };
 
@@ -99,6 +104,10 @@
             endWeek.setDate(startWeek.getDate() + 6);
 
             return [startWeek, endWeek];
+        },
+
+        containsToday: function () {
+            return todayIsInRange(this.getDateRange());
         }
     };
 
@@ -131,6 +140,10 @@
             endMonth.setDate(0);
 
             return [startMonth, endMonth];
+        },
+
+        containsToday: function () {
+            return todayIsInRange(this.getDateRange());
         }
     };
 
@@ -162,6 +175,10 @@
             endYear.setDate(0);
 
             return [startYear, endYear];
+        },
+
+        containsToday: function () {
+            return todayIsInRange(this.getDateRange());
         }
     };
 
@@ -261,6 +278,10 @@
 
         getDateRange: function () {
             return [this.startDate, this.endDate];
+        },
+
+        containsToday: function () {
+            return todayIsInRange(this.getDateRange());
         }
     };
 
--- a/plugins/CoreHome/angularjs/common/services/periods.js
+++ b/plugins/CoreHome/angularjs/common/services/periods.js
@@ -275,7 +275,8 @@
             parse: parse,
             parseDate: parseDate,
             format: format,
-            RangePeriod: RangePeriod
+            RangePeriod: RangePeriod,
+            todayIsInRange: todayIsInRange
         };
 
         function getAllLabels() {
@@ -383,5 +384,17 @@
         date.setSeconds(0);
         date.setMilliseconds(0);
         return date;
+    }
+
+    function todayIsInRange(dateRange) {
+        if (!dateRange.isArray && dateRange.length !== 2) {
+            return false;
+        }
+
+        if (getToday() >= dateRange[0] && getToday() <= dateRange[1]) {
+            return true;
+        }
+
+        return false;
     }
 })();
--- a/plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
+++ b/plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
@@ -17,9 +17,7 @@
             selected: [],
             fetchMenuItems: fetchMenuItems,
             reloadMenuItems: reloadMenuItems,
-            findSubcategory: findSubcategory,
-            fetchVisits: fetchVisits,
-            fetchLiveVisits: fetchLiveVisits,
+            findSubcategory: findSubcategory
         };
 
         return model;
@@ -162,39 +160,6 @@
 
                 return model.menu;
             });
-        }
-
-        function fetchVisits(date, period, callback) {
-            var ajaxRequest = new ajaxHelper();
-
-            ajaxRequest.addParams({
-                module: 'API',
-                method: 'VisitsSummary.getVisits',
-                format: 'json',
-                date: date,
-                period: period,
-            }, 'get');
-            ajaxRequest.setCallback(callback);
-            ajaxRequest.setFormat('json');
-            ajaxRequest.withTokenInUrl();
-
-            ajaxRequest.send();
-        }
-
-        function fetchLiveVisits(date, period, callback) {
-            var ajaxRequest = new ajaxHelper();
-            ajaxRequest.addParams({
-                module: 'API',
-                method: 'Live.getLastVisitsDetails',
-                format: 'json',
-                date: decodeURIComponent(date),
-                period: period,
-            }, 'get');
-            ajaxRequest.setCallback(callback);
-            ajaxRequest.setFormat('json');
-            ajaxRequest.withTokenInUrl();
-
-            ajaxRequest.send();
         }
     }
 })();
--- a/plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
+++ b/plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
@@ -17,7 +17,9 @@
             selected: [],
             fetchMenuItems: fetchMenuItems,
             reloadMenuItems: reloadMenuItems,
-            findSubcategory: findSubcategory
+            findSubcategory: findSubcategory,
+            fetchVisits: fetchVisits,
+            fetchLiveVisits: fetchLiveVisits,
         };
 
         return model;
@@ -160,6 +162,39 @@
 
                 return model.menu;
             });
+        }
+
+        function fetchVisits(date, period, callback) {
+            var ajaxRequest = new ajaxHelper();
+
+            ajaxRequest.addParams({
+                module: 'API',
+                method: 'VisitsSummary.getVisits',
+                format: 'json',
+                date: date,
+                period: period,
+            }, 'get');
+            ajaxRequest.setCallback(callback);
+            ajaxRequest.setFormat('json');
+            ajaxRequest.withTokenInUrl();
+
+            ajaxRequest.send();
+        }
+
+        function fetchLiveVisits(date, period, callback) {
+            var ajaxRequest = new ajaxHelper();
+            ajaxRequest.addParams({
+                module: 'API',
+                method: 'Live.getLastVisitsDetails',
+                format: 'json',
+                date: decodeURIComponent(date),
+                period: period,
+            }, 'get');
+            ajaxRequest.setCallback(callback);
+            ajaxRequest.setFormat('json');
+            ajaxRequest.withTokenInUrl();
+
+            ajaxRequest.send();
         }
     }
 })();
--- a/plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
+++ b/plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
@@ -7,9 +7,9 @@
 (function () {
     angular.module('piwikApp').controller('ReportingMenuController', ReportingMenuController);
 
-    ReportingMenuController.$inject = ['$scope', 'piwik', '$location', '$timeout', 'reportingMenuModel', '$rootScope', 'piwikUrl', 'piwikApi'];
+    ReportingMenuController.$inject = ['$scope', 'piwik', '$location', '$timeout', 'reportingMenuModel', '$rootScope', 'piwikUrl'];
 
-    function ReportingMenuController($scope, piwik, $location, $timeout, menuModel, $rootScope, piwikUrl, piwikApi) {
+    function ReportingMenuController($scope, piwik, $location, $timeout, menuModel, $rootScope, piwikUrl) {
 
         $scope.helpShownCategory = null;
 
@@ -25,10 +25,6 @@
         var showSubcategoryHelpOnLoad = null;
 
         var initialLoad = true;
-
-        var UI = require('piwik/UI');
-        var notification = new UI.Notification();
-        var notificationId = 'only_raw_data_notification';
 
         $scope.currentCategory = piwikUrl.getSearchParam('category');
         $scope.currentSubcategory = piwikUrl.getSearchParam('subcategory');
@@ -223,41 +219,6 @@
             enterSubcategory(found.category, found.subcategory, found.subsubcategory);
         });
 
-        function showOnlyRawDataMessageIfRequired() {
-            notification.remove(notificationId);
-
-            if (segment !== '') {
-                return;
-            }
-
-            var subcategoryExceptions = [
-                'Live_VisitorLog',
-                'General_RealTime',
-                'UserCountryMap_RealTimeMap',
-            ];
-
-            if (subcategoryExceptions.indexOf($scope.currentSubcategory) !== -1) {
-                return;
-            }
-
-            piwikApi.fetch({ method: 'VisitsSummary.getVisits' }).then(function (json) {
-                if (json.value > 0) {
-                    return;
-                }
-
-                piwikApi.fetch({ method: 'Live.getLastVisitsDetails', filter_limit: 1, doNotFetchActions: 1 }).then(function (json)  {
-                    if (json.length == 0) {
-                        return;
-                    }
-
-                    var url = broadcast.buildReportingUrl('category=General_Visitors&subcategory=Live_VisitorLog')
-
-                    var message = _pk_translate('CoreHome_PeriodHasOnlyRawData', ['<a href="' + url + '">', '</a>']);
-                    notification.show(message, {context: 'info', id: notificationId});
-                });
-            });
-        }
-
         $rootScope.$on('piwikPageChange', function (event) {
             if (!initialLoad) {
                 globalAjaxQueue.abort();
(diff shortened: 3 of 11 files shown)


flamisz:
The original issue says:

> For now only apply when "no segment" is select (report shows all visits)

This solution works with segments as well but easy to disable if we don't want to support yet. Is it ok if it works with segments @tsteur?

Update:

the `Live.getLastVisitsDetails` does not work properly with segments, so I disable it for now

plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
```diff
@@ -161,5 +163,38 @@
...
+        }
+
+        function fetchLiveVisits(date, period, callback) {
+            var ajaxRequest = new ajaxHelper();
+            ajaxRequest.addParams({
+                module: 'API',
+                method: 'Live.getLastVisitsDetails',
+                format: 'json',
+                date: decodeURIComponent(date),
+                period: period,
```
sgiehl: I would suggest to add a `limit: 1` to that query. It's actually only relevant if there is a visit. No need to query all of them. So maybe the method should then be called something like `doesPeriodHasLiveVisits`

plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
```diff
@@ -235,6 +262,12 @@
 
             $('#loadingError').hide();
 
+            if ($scope.currentSubcategory != 'Live_VisitorLog') {
```
sgiehl: Might make sense to also _not_ show it on the live map

plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
```diff
@@ -161,5 +163,38 @@
...
+
+        function fetchVisits(date, period, callback) {
+            var ajaxRequest = new ajaxHelper();
+
+            ajaxRequest.addParams({
+                module: 'API',
+                method: 'VisitsSummary.getVisits',
+                format: 'json',
+                date: date,
+                period: period,
```
sgiehl: if it should work with segments you may need to add the segment parameter to this and the query below, to ensure it fetches the segmented data.

plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
```diff
@@ -161,5 +163,38 @@
...
+            ajaxRequest.send();
+        }
+
+        function fetchLiveVisits(date, period, callback) {
+            var ajaxRequest = new ajaxHelper();
+            ajaxRequest.addParams({
+                module: 'API',
+                method: 'Live.getLastVisitsDetails',
+                format: 'json',
+                date: decodeURIComponent(date),
```
sgiehl: guess the `decodeURIComponent` shouldn't be needed here

plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
```diff
@@ -161,5 +163,38 @@
                 return model.menu;
             });
         }
+
+        function fetchVisits(date, period, callback) {
+            var ajaxRequest = new ajaxHelper();
+
+            ajaxRequest.addParams({
```
sgiehl: As this is an angular component you should use `piwikApi` in favor of the ajaxHelper I guess. See https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/PrivacyManager/angularjs/manage-gdpr/managegdpr.controller.js#L24  for an example

plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
```diff
@@ -219,6 +223,29 @@
...
+                }
+
+                menuModel.fetchLiveVisits(decodeURIComponent(date), period, function(json) {
+                    if (json.length == 0) {
+                        return;
+                    }
+
+                    var url = piwik.broadcast.getHashFromUrl();
+                    url = piwik.broadcast.updateParamValue('category=General_Visitors', url);
+                    url = piwik.broadcast.updateParamValue('subcategory=Live_VisitorLog', url);
```
sgiehl: You should replace this with
```suggestion
                    var url = broadcast.buildReportingUrl('category=General_Visitors&subcategory=Live_VisitorLog')
```
otherwise it might contain unwanted parameters.

sgiehl (changes requested):

plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
flamisz: thanks, make sense. I'll do it

plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
flamisz: 👍

plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
flamisz: interestingly it is there (I mean in the request, even without adding it here). but I'll double-check it, and even it's there I'll add it.

plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
flamisz: I was wondering why we have both, but I see now 😄 
thanks

plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
flamisz: much nicer 👍

plugins/CoreHome/angularjs/reporting-menu/reportingmenu-model.js
flamisz: I checked the `piwikApi` and it adds all the parameters by default (idSite, period, segment, date...)

plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
tsteur: also the real time widget page etc. Eventually would maybe need to let plugins hook into this

tsteur: also the real time widget page. eventually would need to maybe let plugins hook into this somehow

plugins/CoreHome/angularjs/reporting-menu/reportingmenu.controller.js
```diff
@@ -219,6 +223,29 @@
...
+
+                    var message = _pk_translate('CoreHome_PeriodHasOnlyRawData', ['<a href="' + url + '">', '</a>']);
+                    notification.show(message, {context: 'info', id: notificationId});
+                });
+            });
+        }
+
```
tsteur: @flamisz I'm still needing to think about it more but overall thinking as soon as we notice there's visits for the given segment/date then we wouldn't need to execute this request anymore on subsequent page changes maybe and could use some "cached result"? At least the live query to the get last visits wouldn't need to be executed anymore since once we know it had data once it will also have it for all subsequent page views (unless date or segment changes). 

Overall wondering if we'd maybe also want to execute this say max once every minute or so (when page is completely reloaded then it would also be executed again directly) and maybe we'd execute it say max 5 times per page view or something. Let's maybe brainstorm tomorrow

plugins/CoreHome/angularjs/reporting-page/reportingpage.controller.js
```diff
@@ -35,6 +41,73 @@
...
+            var categoryExceptions = [
+                'HeatmapSessionRecording_Heatmaps',
+                'HeatmapSessionRecording_SessionRecordings',
+                'Marketplace_Marketplace',
+            ];
+
+            if (subcategoryExceptions.indexOf($search.subcategory) !== -1 || categoryExceptions.indexOf($search.category) !== -1 || $search.subcategory.toLowerCase().indexOf('manage') !== -1) {
+                $scope.showOnlyRawDataAlert = false;
+                return;
+            }
```
sgiehl: I'm wondering if there would be a more elegant way, that would also allow plugins to define their pages that only contain live data or configuration pages.
Or we could give all reports a new possible flag 'live-data'. And if all reports rendered on a page have this flag it won't show the warning. For configuration pages like the `Add new goal`, we could check if any report is included in the page after loading finished.
Might be worth to discuss on a better solution than having a static list. ping @tsteur @diosmosis

plugins/CoreHome/angularjs/reporting-page/reportingpage.directive.html
```diff
@@ -4,6 +4,8 @@
 
     <div ng-show="hasNoPage">{{ 'CoreHome_NoSuchPage'|translate }}</div>
 
+    <div class="alert alert-info" ng-show="showOnlyRawDataAlert" ng-bind-html="rawDataMessage"></div>
```
sgiehl: Is there any reason for directly having the alert in the code and not using the notification api? See
https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/plugins/GeoIp2/angularjs/geoip2-updater/geoip2-updater.controller.js#L93-L100

sgiehl: I know it was suggested that way in the original issue, but imho the grey on grey doesn't look good:

image

plugins/CoreHome/angularjs/reporting-page/reportingpage.controller.js
```diff
@@ -35,6 +41,73 @@
...
+                $scope.showOnlyRawDataAlert = true;
+            }
+
+            var $search = $location.search();
+
+            if ($search.segment !== '') {
+                $scope.showOnlyRawDataAlert = false;
+                return;
+            }
+
```
sgiehl, line 70: The original issue says:

> Maybe we would only do this extra API call if the date includes "today" as in other cases it should not be needed.

So guess another check for this is missing here. Or is it supposed to be displayed for all dates/periods?

sgiehl (changes requested):

plugins/CoreHome/angularjs/reporting-page/reportingpage.controller.js
tsteur: @sgiehl ideally this was the case but I think the solution is good enough as it works like in most cases (or all cases for core + matomo plugins/features) and it be quite a bit more work to make it work for all the other edge cases plus developers would need to do more to make this work etc. At least for now. Be good to use the feature for a while and release MVP and over time we can still see and make changes to it.

plugins/CoreHome/angularjs/reporting-page/reportingpage.directive.html
flamisz: The notification is different than the alert, other css, divs, etc. @tsteur mentioned this alert, because this one has this gray version. 
And I know we have actually an alert angularjs component as well, but I had issues with the link inside the notification (it is sanitized), so I just used the native div+css.

flamisz: I just talked to @tsteur, will use the notification.

plugins/CoreHome/angularjs/reporting-page/reportingpage.controller.js
flamisz: yes, that's still missing, I will add.

plugins/CoreHome/lang/en.json
```diff
@@ -122,6 +122,7 @@
...
         "SoftwareSubcategoryHelp": "The Software section shows the operating systems, browsers and plugins that your visitors are using to access the site so that you can optimise your site to ensure it is fully compatible with the most popular configurations..",
         "EngagementSubcategoryHelp1": "The Engagement section provides reports that help to quantify how many new and returning visitors you get. You can also review reports that break down the average time and number of pages per visit, as well as the number of times a visitor has been to your site and the most common number of days between visits.",
-        "EngagementSubcategoryHelp2": "This can help you to optimise for frequency and high-interaction visits in addition to maximising your reach."
+        "EngagementSubcategoryHelp2": "This can help you to optimise for frequency and high-interaction visits in addition to maximising your reach.",
+        "PeriodHasOnlyRawData": "It looks like reports for this period have not been processed yet. Do you want to see what's happening now? Check out the %1$sVisits log%2$s or choose a different date period until the reports are generated."
```
mattab, line 126: optional feedback: when we refer to existing menu names, it may be a bit easier to write simply:

> Check out the %s or choose

and then replace the placeholder by the link and the menu name (using translation key so there is no duplicate of menu names.)

flamisz: good catch, sure, I will do it 👍

sgiehl: While it might be ok to do it in this case, we should not try to avoid duplication in translations too much. Often it will be better to have duplication as it is easier for the translator to know the context. Also in some languages names might be written in another way based on their usage and including the version that was translated without any context might actually be grammatically incorrect sometimes...

sgiehl: Thinking about it again, we even should change it back in this case. For example in German `the` can be translated with `der`, `die` or `das` depending on the word that follows after it. As a translator doesn't know what follows in this case, it actually can't be translated correctly.

plugins/CoreHome/angularjs/common/services/periods.js
```diff
@@ -384,4 +385,16 @@
         date.setMilliseconds(0);
         return date;
     }
+
+    function todayIsInRange(dateRange) {
```
sgiehl, line 410: Wondering if it might be more useful to change that method so it directly works with period and date param. I guess in most places we are working with those params and not already with a parsed period.
Or alternatively we could add a method to all period classes instead, so in the end we would be able to call something like `$piwikPeriods.parse(currentPeriod, currentDate).containsToday()`

sgiehl: Whatever this methods looks like in the end: It would be useful to add some simple tests for it to prevent any regressions.

flamisz: I thought about that and now that you mentioned it, I like the `$piwikPeriods.parse(currentPeriod, currentDate).containsToday()` call.
yeah, definitely a test is useful here.

plugins/CoreHome/lang/en.json
tsteur: +1 for that. Avoiding duplication we stopped doing in most cases many years ago as it leads to poor translations.

plugins/CoreHome/angularjs/common/services/periods.js
flamisz: If I add the `containsToday()` to all period classes, should we add this as a breaking change or mention it anyway in the changelog? If someone has a custom range, but that doesn't have this method yet, there will be an error. Or should I check if the method is there? But in the future probably will forget to add the check anyway.
what should be the good solution here?

plugins/CoreHome/lang/en.json
mattab: interesting, makes sense.. could we document this best practise in this relevant section maybe https://developer.matomo.org/guides/translations#reusing-translations ?

tsteur: @mattab that's about a different "reusing" regarding using translations from core basically. Created https://github.com/matomo-org/developer-documentation/pull/463 . We didn't document it earlier since someone would usually not do this anyway but also good to specifically mention to not do it

plugins/CoreHome/angularjs/common/services/periods.js
sgiehl: Actually I doubt anyone is using custom period js classes at all. Also I don't think this part is public API. Guess we don't need to mention it in changelog 🤔   ping @tsteur

sgiehl (approved):
Looks go to me now. We could theoretically try to add a new UI test that shows this notification. But that might not be too easy. UI tests by default run the archiving, so by default reports would be generated when you add a visit for today. Guess we could maybe use an empty fixture and send a tracking request for today in the UI test itself or something like that.

plugins/CoreHome/angularjs/common/services/periods.js
tsteur: we won't need to mention it in the changelog as it's not public API (almost nothing in our JS is public API) .Was thinking maybe cohorts did but doesn't look like it and it seems we're not using it in any of our plugins so nothing needed to be done 👍

flamisz:
I tried to add that test, but couldn't make it work. I tried this inside the `_spec.js` file: (I tried with form params as well instead of params in the URL)

```php
       const trackerUrl = config.piwikUrl + "matomo.php?action_name=&idsite=1&url=http://piwik.net/test/url";

        await request({
            method: 'POST',
            uri: trackerUrl,
        });

        await page.goto(pageUrl);
        expect(await page.screenshot({ fullPage: true })).to.matchImage('show_notification_when_only_raw_data_exists');
```

But there is no new log in the db... 
Any thoughts on this @sgiehl?
I found one example here:
https://github.com/matomo-org/matomo/blob/3ab0d4ee6af2eccfd3e1e7e0d192a0032f42edcb/tests/UI/specs/OneClickUpdate_spec.js#L85-L97

EDIT: I talked to @tsteur about this, he showed me a couple of things in the tests, so I try to make it work.

flamisz:
@tsteur and @sgiehl I figured what was the problem and what is the solution. Even when I called the tracker it didn't work, because when we load the page that triggers the archiving, so I need to add this to test, and I don't need any extra tracker call this way.:

```php
       testEnvironment.configOverride.General = {
            browser_archiving_disabled_enforce: '1',
            enable_browser_archiving_triggering: '0',
        };
        testEnvironment.optionsOverride = {
            enableBrowserTriggerArchiving: '0',
        };
```

Only one thing that came up now. We show this notification only if today's date is in the date range. I couldn't find a way to mock the today in the UI tests. Are we happy with it if the test uses today every time? It means we can have some false negatives, when the tests run around midnight.

tests/UI/specs/OnlyRawDataNotification_spec.js
```diff
@@ -0,0 +1,33 @@
...
+ * Only Raw Data Notification screenshot tests.
+ *
+ * @link https://matomo.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ */
+
+describe("OnlyRawDataNotification", function () {
+    this.fixture = "Piwik\\Tests\\Fixtures\\OneVisit";
+
+    const generalParams = 'idSite=1&period=week&date=today';
```
sgiehl: to ensure that might not fail randomly e.g. on monday around midnight it might be better to use a fix range like `period=range&date=2021-03-01,today`. As archiving is disabled it shouldn't matter that this range will contain more and more days...

flamisz: thanks, didn't think about this. I'll try it

sgiehl (approved):
Should be good to merge now

Issue #17381, opened 2021-03-23
Referrer full path missing for most websites


alimony:
In the Acquisition > Websites view, almost all referrers only show the "index" path. I know the rest is not being anonymised or anything, looking at settings, and the fact that a handful of the websites also show a full path. It looks like I've been on the front page on some _major_ websites, which I know not to be true. In some cases I've been able to track down the specific path from which I've been linked, but that path does not show up in Matomo.

## Expected Behavior

I expect to see the full path for all referrers.

## Current Behavior

I see only the index page for almost all referrers.

## Possible Solution

Not sure.

## Steps to Reproduce (for Bugs)

1. Go to Acquisition > Websites
2. Expand referrers to see full paths.

## Context

I want to know from what specific pages I've been linked, and the current data does not tell me that, making the feature not very useful at the moment.

## Your Environment

* Matomo Version: 4.2.1
* PHP Version: 7.4.3
* Server Operating System: Ubuntu 20.04.1 LTS
* Additionally installed plugins: GoogleAnalyticsImporter 4.1.2

Findus23:
Hi,

Which Referrer the browser forwards to a website when the visitor clicks on a link to it, depends on the ReferrerPolicy. 
You can simply try it out by clicking on a link to your website, opening your browsers developer tools (<kbd>F12</kbd>) and entering `document.referrer` in the console.

Matomo can't record any data that doesn't exist and therefore only stores what the browser outputs with `document.referrer`.

If for example your website doesn't use HTTPS, but the linked website does then no browser allows sending the full URL as a referrer (as it would make the secure URL path visible unencrypted).

Also a few days ago Firefox updated their browser to set the default referrer policy (if the site doesn't set another one) to `strict-origin-when-cross-origin`  (https://blog.mozilla.org/security/2021/03/22/firefox-87-trims-http-referrers-by-default-to-protect-user-privacy/) which means that different origins (so domains) never get to see the full referrer, but just the domain.
Chrome also changed this recently and I'm sure other browsers will follow soon.

I hope this answers the topic. If anything is still unclear just reopen the issue.

PR #17382, opened 2021-03-23
more private Referrer-Policy: use strict-origin-when-cross-origin 

Findus23:
related to https://github.com/matomo-org/matomo/issues/17381, https://github.com/matomo-org/matomo/pull/15673 and https://github.com/matomo-org/matomo-nginx/pull/61

At the moment we are sending a less private Referrer-Policy than the one that would be used if we didn't send one.

So now 
- the full referrer is sent to the same origin
- just the domain is sent to another origin (as long as it is using HTTPS)
- nothing is sent to HTTP sites
(as long as `$this->useStrictReferrerPolicy` is not set in which case nothing is sent to other origins)

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/View.php
+++ b/core/View.php
@@ -296,7 +296,7 @@ class View implements ViewInterface
                 Common::sendHeader('Referrer-Policy: same-origin');
             } else {
                 // always send explicit default header
-                Common::sendHeader('Referrer-Policy: no-referrer-when-downgrade');
+                Common::sendHeader('Referrer-Policy: strict-origin-when-cross-origin');
             }
         }
 


diosmosis:
Code looks good to me :+1:, not sure we'd ever want to send the full matomo url referrer to another domain.

Any thoughts @tsteur, @flamisz, @sgiehl?

tsteur:
Hard to say if it would break anything. I remember https://github.com/matomo-org/matomo/pull/14482 broke the overlay when we tried various different settings earlier and we now worked around it with https://github.com/matomo-org/matomo/pull/14766/files

Generally sounds good to do this 👍   FYI seems it's not supported by IE11 

And I suppose nothing changes when someone switches from HTTP to HTTPS? Eg when logging in then we don't look at the full referrer? not sure if something behaves different there maybe (didn't check in detail)

flamisz:
In the code we have the comment `always send explicit default header`, and looks like this is the new default header. I'm not sure about what could it break, but definitely the more secure and up-to-date solution.

Findus23:
Maybe we could merge this quite early after a release, so there is a longer time to notice if this breaks something in a subtil way.

diosmosis:
Sounds good to merge this early (cc @tsteur)

PR #17387, opened 2021-03-24
Apache 2.4+ compatibility for misc/cron/.htaccess

jorisdrenth:
### Description:

In Apache 2.4+, the syntax for access control has changed. This causes the cron script to return a 500 error when running on Apache 2.4+. This PR checks which version of Apache is used and uses the right syntax.

See also: [https://httpd.apache.org/docs/2.4/upgrading.html](https://httpd.apache.org/docs/2.4/upgrading.html)

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/misc/cron/.htaccess
+++ b/misc/cron/.htaccess
@@ -1,15 +1,27 @@
 # Allow direct web access to Web cron
-# Apache 2.2
-<IfModule !authz_core_module>
-        <Files "archive.php">
-                Order Allow,Deny
-                Allow from all
-        </Files>
+<IfModule mod_version.c>
+        <IfVersion < 2.4>
+                <Files "archive.php">
+                        Order Allow,Deny
+                        Allow from all
+                </Files>
+        </IfVersion>
+        <IfVersion >= 2.4>
+                <Files "archive.php">
+                        Require all granted
+                </Files>
+	</IfVersion>
 </IfModule>
-
-# Apache 2.4+
-<IfModule authz_core_module>
-        <Files "archive.php">
-                Require all granted
-        </Files>
+<IfModule !mod_version.c>
+        <IfModule !mod_authz_core.c>
+                <Files "archive.php">
+                        Order Allow,Deny
+                        Allow from all
+                </Files>
+        </IfModule>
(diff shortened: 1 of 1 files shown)


misc/cron/.htaccess
```diff
@@ -1,5 +1,15 @@
 # Allow direct web access to Web cron
-<Files "archive.php">
-	Order Allow,Deny
-	Allow from all
-</Files>
+# Apache 2.2
+<IfModule !authz_core_module>
```
diosmosis: Looks like we already do something similar in some other places, except instead of `<IfModule !authz_core_module>`, we use `<IfVersion`: https://github.com/matomo-org/matomo/blob/2b35b46a0988f8c3b6f848c59be8d7d918965a52/plugins/Installation/ServerFilesGenerator.php#L251-L259

Could you change this to use that syntax to keep things consistent?

diosmosis (changes requested):
Hi @jorisdrenth, thanks so much for the contribution! I requested one change, after it's made we'll be able to merge it.

jorisdrenth:
Thanks for your comment @diosmosis. I've made the requested changes, also adding the `<IfModule !mod_version.c>` part which I found in the file you linked. I think this is the most consistent code I can make of it.

I've tested this version of the .htaccess and it works at my client's Matomo installation on our Apache 2.4 servers.

diosmosis:
@jorisdrenth does the mod_version one alone not work for you?

jorisdrenth:
@diosmosis No, it causes an Internal Server Error. mod_version is not available on our servers, that's why we always check for mod_authz_core.

In my opinion this is the best solution to the widest range of users. Also see [the comment](https://github.com/matomo-org/matomo/blob/2b35b46a0988f8c3b6f848c59be8d7d918965a52/plugins/Installation/ServerFilesGenerator.php#L242-L249) in the file you linked.

diosmosis:
Hi @jorisdrenth, you're right, my glance at the file should have been more thorough. Thanks again for the contribution!

jorisdrenth:
Np, thanks for accepting the PR @diosmosis!

PR #17389, opened 2021-03-24
icon for informational system check

Findus23:
This is quite subjective, but I think it looks nicer if there is also an icon for informational entries

Before:
grafik

Now:
grafik

One could use icon-info instead as it looks a bit nicer, but it doesn't fit the style of the other icons.

### Review

* [x] Functional review done
* [x] Code review done

### DIFF
--- a/plugins/Diagnostics/tests/UI/expected-screenshots/Diagnostics_page.png
+++ b/plugins/Diagnostics/tests/UI/expected-screenshots/Diagnostics_page.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:537abc41c9f56eb9bf976d5105834a4d4eaa6fc4b9718cdd3a8be71d50ab0c93
-size 426942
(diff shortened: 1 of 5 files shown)


Findus23:
Also there don't seem to be a LESS variable for the error/warning/info/sucess colors, so I copied the ones from notifications over.

If you are interested I could create a PR unifying them over all of Matomo.

flamisz:
These are the before-after screenshots, to make it easier to decide which one is preferred:

Screen Shot 2021-03-25 at 8 43 41 AM
Screen Shot 2021-03-25 at 8 44 20 AM

Findus23:
If someone doesn't like the icons in the informational results we could maybe remove them just there

sgiehl (approved):
code looks fine

sgiehl:
Personally I like the change. Maybe @tsteur or @mattab could state their opinion as well.
@Findus23  Guess the expected UI files would need an update before we could merge it...

tsteur:
All good 👍

Findus23:
Just noticed while updating tests that there is a separate CSS file for the installer.

PR #17394, opened 2021-03-25
allow plugins to customize the system checks page

Findus23:
I'm not sure if this is the best solution, but at the moment it isn't possible for a plugin to change how the system checks are displayed. Ideally one could dynamically add groups like "Optional", but that sounds very complex for a minor thing.

So the easiest solution to differentiate checks from different plugins I could think of is adding icons to the label.

In addition, I added an event to the top of the page (e.g. to add a warning, notice, disclaimer, etc.). This slows things down a bit, but this is already the slowest page due to tons of I/O, so it shouldn't matter much in comparison.

If someone has a better way to do this, I'm open to suggestions.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Diagnostics/Diagnostic/DiagnosticResult.php
+++ b/plugins/Diagnostics/Diagnostic/DiagnosticResult.php
@@ -28,6 +28,11 @@ class DiagnosticResult
     /**
      * @var string
      */
+    private $icon;
+
+    /**
+     * @var string
+     */
     private $longErrorMessage = '';
 
     /**
@@ -35,9 +40,10 @@ class DiagnosticResult
      */
     private $items = array();
 
-    public function __construct($label)
+    public function __construct($label, $icon = '')
     {
         $this->label = $label;
+        $this->icon = $icon;
     }
 
     /**
@@ -77,6 +83,14 @@ class DiagnosticResult
     public function getLabel()
     {
         return $this->label;
+    }
+
+    /**
+     * @return string
+     */
+    public function getIcon()
+    {
+        return $this->icon;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


diosmosis:
Can't notices be done via notifications? Or is placing it at this point in the HTML better for some reason?

Findus23:
By notice I mean less of a one-time notification about some event, but more adding an additional explainatory paragraph on top of the page (e.g. mentioning that some of the tests are in development and failed tests doesn't have to mean something is seriously wrong, but rather to report an issue with the test)

diosmosis:
@Findus23 this could be done w/ a permanent notification (doesn't close or have a close button) that is added on every view of the page. If you're thinking about alternatives to adding an event, that could be one.

Findus23:
That might work (I would have to pay attention so it isn't still shown on other pages).

But I would still have the other issue of needing a way to differentiate the tests of the plugin.

diosmosis:
@Findus23 can you explain what you mean by "differentiate the tests of the plugin"? Do you mean "mark each test by which plugin it comes from"?

Findus23:
> Do you mean "mark each test by which plugin it comes from"?

Yes. Especially make it clear to the user, which ones are core Matomo checks and which ones are checks that added by my plugin (and therefore potentionally not as stable as the core ones).

diosmosis:
And you want to do this by adding a section at the top of the page like a sort of legend? Or something else? Have you thought about adding this info next to the actual validation names or their results themselves? (Apologies if I'm misunderstanding, it's hard to visualize what's not in my head but someone else's ;) )

Findus23:
For now the workaround of using Emoji in the system check title and a permanent notification for general information on top works well enough, so I'd close this unless it is needed for something else in the future.

PR #17396, opened 2021-03-26
Use ranking query in custom dimensions archiver

diosmosis:
### Description:

As title. There might be a random test failure since RankingQuery doesn't apply ordering (probably a good thing).

### Review

* [ ] Functional review done
* [ ] Code review done


core/API/DocumentationGenerator.php
```diff
@@ -314,6 +314,10 @@ public function getExampleUrl($class, $methodName, $parametersToSet = array())
             } // if there isn't a default value for a given parameter,
             // we need a 'know default value' or we can't generate the link
             elseif ($defaultValue instanceof NoDefaultValue) {
+                if (defined('PIWIK_TEST_MODE') && $nameVariable != 'idSubtable') {
+                    // print error message since it is annoying to have to debug to find this
+                    print "$nameVariable required in $class::$methodName but no value given (this usually means your API test config is missing some parameters)\n";
```
sgiehl: That is now printed quite often within various working tests.

core/DataAccess/LogAggregator.php
```diff
@@ -1059,9 +1065,11 @@ protected function getActionsMetricFields()
      * @param bool|string $where An optional SQL expression used in the SQL's **WHERE** clause.
      * @param array $additionalSelects Additional SELECT fields that are not included in the group by
      *                                 clause. These can be aggregate expressions, eg, `SUM(somecol)`.
+     * @param RankingQuery|bool $rankingQuery
      * @return \Zend_Db_Statement
      */
-    public function queryConversionsByDimension($dimensions = array(), $where = false, $additionalSelects = array(), $extraFrom = [])
+    public function queryConversionsByDimension($dimensions = array(), $where = false, $additionalSelects = array(), $extraFrom = [],
+                                                $rankingQuery = false, $rankingQueryGenerate = false)
```
sgiehl, line 1076: The last parameter isn't documented. And from looking at the code it wasn't directly visible what a difference it makes. But seems when the ranking query is executed it directly returns all records where as otherwise it returns a query pointer, right?

plugins/CustomDimensions/Archiver.php
```diff
@@ -195,6 +217,19 @@ public function queryCustomDimensionActions(DataArray $dataArray, $valueField, $
...
         $logAggregator = $this->getLogAggregator();
         $query     = $logAggregator->generateQuery($select, $from, $where, $groupBy, $orderBy);
+
+        if ($this->rankingQueryLimit > 0) {
+            $rankingQuery = new RankingQuery($this->rankingQueryLimit);
+            $rankingQuery->addLabelColumn(array($valueField, 'url'));
+            // TODO: combine below 4
+            $rankingQuery->addColumn(Metrics::INDEX_PAGE_SUM_TIME_SPENT, 'sum');
+            $rankingQuery->addColumn(Metrics::INDEX_BOUNCE_COUNT, 'sum');
+            $rankingQuery->addColumn(Metrics::INDEX_PAGE_EXIT_NB_VISITS, 'sum');
```
sgiehl: Guess those line can be removed, as they are actually contained in `$metricIds`?

core/DataAccess/LogAggregator.php
```diff
@@ -421,7 +422,7 @@ public function generateQuery($select, $from, $where, $groupBy, $orderBy, $limit
         return $query;
     }
 
-    protected function getVisitsMetricFields()
+    public function getVisitsMetricFields()
```
sgiehl: Any reason for making this public? Couldn't find any additionally usage that requires it 🤔

sgiehl (changes requested):
There are also some tests failing that don't look random

core/DataAccess/LogAggregator.php
diosmosis: Yes that is correct, I'll add docs.

core/DataAccess/LogAggregator.php
diosmosis: I was using it, forgot to change it back :+1:

core/API/DocumentationGenerator.php
diosmosis: I'll do this in another pr.

plugins/CustomDimensions/Archiver.php
diosmosis: :+1:

core/DataAccess/LogAggregator.php
```diff
@@ -1075,6 +1087,17 @@ public function queryConversionsByDimension($dimensions = array(), $where = fals
         $orderBy = false;
         $query   = $this->generateQuery($select, $from, $where, $groupBy, $orderBy);
 
+        if (!empty($rankingQuery)) {
+            $sumColumns = array_keys($availableMetrics);
+            $rankingQuery->addColumn($sumColumns, 'sum');
```
tsteur, line 1092: @diosmosis can we assume that it is always a `sum`? or is it maybe also sometimes min, max, etc?

plugins/CustomDimensions/Archiver.php
```diff
@@ -245,4 +276,15 @@ protected function cleanCustomDimensionValue($value)
         return self::LABEL_CUSTOM_VALUE_NOT_DEFINED;
     }
 
+    private function getRankingQueryLimit()
+    {
+        $configGeneral = Config::getInstance()->General;
+        $configLimit = $configGeneral['archiving_ranking_query_row_limit'];
+        $limit = $configLimit == 0 ? 0 : max(
```
tsteur, line 298: I suppose this looks good. Was wondering if the 50K ranking query limit could cause issues since we technically allow 1000 root tables and 1000 entries per subtable by default. I'm assuming this will limit the number of results shown quite a bit but that's kind of also the point of this ranking query? 

For the ranking query limit would maybe do something like 

```php
$this->rankingQueryLimit = max(configLimit, 10 * $this->maximumRowsInDataTableLevelZero);
```

just in case someone has a high `datatable_archiving_maximum_rows_custom_dimensions` set and so you can still get different results per subtable but I think also not really needed.

I'm not entirely sure what the report groups by though and how the subtables are created so could also just leave as is.

core/DataAccess/LogAggregator.php
diosmosis: missed this :+1:

plugins/CustomDimensions/Archiver.php
diosmosis: subtables are urls, so it makes sense :+1:

Issue #17352, opened 2021-03-17
Don't allow enforcing 2FA unless the superuser has set it up already


Findus23:
reported in https://forum.matomo.org/t/problem-with-the-two-factor-authentication-setting/41128

If a user who is unable to set up 2FA accidentally enables `Require two-factor authentication for everyone`, they are unable to disable it until they set up 2FA.
(or they update the setting in the db:
```sql
update matomo_plugin_setting
set setting_value=0
where setting_name = 'twoFactorAuthRequired'
```
)

I think this setting should only be allowed if at least one superuser has 2FA already set up (or maybe only if the current superuser has 2FA already set up)

tsteur:
Ideally we would get someone to activate 2FA first and then allow enabling it. That might be a bit more work though so we could maybe rather add a message `Enforcing 2FA will require all users to set up 2FA using an authenticator app. This means they need to have a device like a smartphone or computer where they can install an app.`. 

We could have otherwise checked if the current visitor has 2FA set up and then try if we can disable a setting in the UI super easily if 2FA is disabled for the current account and adjust the message saying please set up 2FA first but it's not needed if it takes say more than 10 minutes.

andyjdavis:
I have put up a PR that just expands the text on the setting UI to provide more information about the consequences of turning on require 2fa. https://github.com/matomo-org/matomo/pull/17400

I have some questions about how to determine if the current user has 2fa enabled themselves.

The directory containing plugins/TwoFactorAuth/SystemSettings.php contains Controller.php, Validator.php, and TwoFactorAuthentication.php. The controller contains an instance of validator which contains an instance of twoFa. There are functions available on each level of that structure to determine if the user has 2fa set up. For example
Validator::check2FaEnabled()
Validator::checkVerified2FA()

Is there a preferred approach for accessing these methods (or similar) from the context of the system settings? I haven't been able to find an example of a plugin SystemSettings.php loading an instance of an object to allow it to selectively disable a setting on a per-user basis.

diosmosis:
[...]

PR #17400, opened 2021-03-28
Disallow enable 2FA requirement when user doesn't have 2FA enabled

andyjdavis:
### Description:
Trying to avoid a situation where admin turns on require 2fa without having it set up themselves. https://github.com/matomo-org/matomo/issues/17352

This is just a text change in the admin UI. I will ask my questions about how to determine whether the current user has 2fa enabled in the orginal issue (#17352 ).
2fa

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/TwoFactorAuth/SystemSettings.php
+++ b/plugins/TwoFactorAuth/SystemSettings.php
@@ -8,10 +8,13 @@
 
 namespace Piwik\Plugins\TwoFactorAuth;
 
+use Piwik\Piwik;
 use Piwik\Plugin;
 use Piwik\Settings\Setting;
 use Piwik\Settings\FieldConfig;
 use Piwik\Url;
+use Piwik\Plugins\TwoFactorAuth\TwoFactorAuthentication;
+use Piwik\Container\StaticContainer;
 
 class SystemSettings extends \Piwik\Settings\Plugin\SystemSettings
 {
@@ -29,11 +32,16 @@ class SystemSettings extends \Piwik\Settings\Plugin\SystemSettings
 
     private function createRequire2FA()
     {
-        return $this->makeSetting('twoFactorAuthRequired', $default = false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
-            $field->title = 'Require two-factor authentication for everyone';
-            $field->description = 'When enabled, every user has to enable two factor authentication.';
+        $setting = $this->makeSetting('twoFactorAuthRequired', $default = false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
+            $field->title = Piwik::translate('TwoFactorAuth_RequireTwoFAForAll');
+            $field->description = Piwik::translate('TwoFactorAuth_RequireTwoFAForAllInformation');
             $field->uiControl = FieldConfig::UI_CONTROL_CHECKBOX;
         });
+
+        $isWritable = defined('PIWIK_TEST_MODE') || TwoFactorAuthentication::isUserUsingTwoFactorAuthentication(Piwik::getCurrentUserLogin());
+        $setting->setIsWritableByCurrentUser($isWritable);
+
+        return $setting;
     }
 
     private function create2FATitle()
--- a/plugins/TwoFactorAuth/SystemSettings.php
+++ b/plugins/TwoFactorAuth/SystemSettings.php
@@ -36,10 +36,12 @@ class SystemSettings extends \Piwik\Settings\Plugin\SystemSettings
             $field->title = Piwik::translate('TwoFactorAuth_RequireTwoFAForAll');
             $field->description = Piwik::translate('TwoFactorAuth_RequireTwoFAForAllInformation');
             $field->uiControl = FieldConfig::UI_CONTROL_CHECKBOX;
-        });
 
-        $isWritable = defined('PIWIK_TEST_MODE') || TwoFactorAuthentication::isUserUsingTwoFactorAuthentication(Piwik::getCurrentUserLogin());
-        $setting->setIsWritableByCurrentUser($isWritable);
+            $isWritable = defined('PIWIK_TEST_MODE') || TwoFactorAuthentication::isUserUsingTwoFactorAuthentication(Piwik::getCurrentUserLogin());
+            if (!$isWritable) {
+                $field->uiControlAttributes = ['disabled' => 'disabled'];
+            }
+        });
 
         return $setting;
     }
--- a/plugins/Tour/Engagement/ChallengeSetupTwoFa.php
+++ b/plugins/Tour/Engagement/ChallengeSetupTwoFa.php
@@ -8,7 +8,6 @@
  */
 namespace Piwik\Plugins\Tour\Engagement;
 
-use Piwik\Container\StaticContainer;
 use Piwik\Piwik;
 use Piwik\Plugins\TwoFactorAuth\TwoFactorAuthentication;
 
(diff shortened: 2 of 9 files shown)


plugins/TwoFactorAuth/SystemSettings.php
```diff
@@ -31,7 +31,7 @@ private function createRequire2FA()
     {
         return $this->makeSetting('twoFactorAuthRequired', $default = false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
             $field->title = 'Require two-factor authentication for everyone';
-            $field->description = 'When enabled, every user has to enable two factor authentication.';
+            $field->description = 'When enabled, every user has to enable two factor authentication. Enforcing 2FA will require all users to have access to an authenticator app. This means they need to have a device like a smartphone or computer where they can install an app.';
```
sgiehl: Interesting that the titles & descriptions in this file actually seems to be not translatable. Would you mind moving them to the translation file within your PR?

andyjdavis: I have now attempted to do that however it doesn't seem to be working as expected. Once I include a call to Piwik::translate() the general settings page still loads but only partially. The 2fa section and, it seems, everything after that is omitted.

andyjdavis:
I believe I currently have two independent problems in this PR. Quite impressive for such a small amount of code :| I'm new to this repo so likely making simple mistakes. Hopefully those mistakes are immediately apparent to someone who knows Matomo better.

1) This error is displayed.
Maximum function nesting level of '500' reached, aborting!

The 500 comes from my PHP settings.
xdebug.max_nesting_level=500

The error is caused by something in these 3 lines as it doesn't occur if I comment them out. The function to set up the settings appears to be being called over and over.
```
$twoFa = StaticContainer::get(TwoFactorAuthentication::class);
$isWritable = $twoFa->isUserUsingTwoFactorAuthentication(Piwik::getCurrentUserLogin());
$setting->setIsWritableByCurrentUser($isWritable);
```
Interestingly the error is displayed even when first trying to access matomo before even reaching a log in page.


2) As mentioned in another comment thread once I include a call to Piwik::translate() the general settings page only partially loads. This assumes I have the 3 lines of code above commented out to let the page load at all. Other than adding the strings to the plugin's en.json I'm unsure if there are other steps i need to perform.

Having Matomo in development mode doesn't seem to result in anything else being output. The error log just contains the message "Error in Matomo: Maximum function nesting level of &#039;500&#039; reached, aborting!"

plugins/TwoFactorAuth/SystemSettings.php
sgiehl: You need to add a `use Piwik\Piwik;` at the file head, where the other use statements are.

plugins/TwoFactorAuth/SystemSettings.php
```diff
@@ -29,11 +29,17 @@ protected function init()
...
-            $field->description = 'When enabled, every user has to enable two factor authentication.';
+        $setting = $this->makeSetting('twoFactorAuthRequired', $default = false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
+            $field->title = Piwik::translate('TwoFactorAuth_RequireTwoFAForAll');
+            $field->description = Piwik::translate('TwoFactorAuth_RequireTwoFAForAllInformation');
             $field->uiControl = FieldConfig::UI_CONTROL_CHECKBOX;
         });
+
+        $twoFa = StaticContainer::get(TwoFactorAuthentication::class);
+        $isWritable = $twoFa->isUserUsingTwoFactorAuthentication(Piwik::getCurrentUserLogin());
+        $setting->setIsWritableByCurrentUser($isWritable);
```
sgiehl: this requires some more use statements in the head:
```
use Piwik\Plugins\TwoFactorAuth\TwoFactorAuthentication;
```

diosmosis:
Hi @andyjdavis looks like there's some infinite recursion. From a quick look, it seems like it's caused by accessing `TwoFactorAuthentication` in the `SystemSettings` class. `TwoFactorAuthentication` requires `SystemSettings` to be injected in the constructor, and `SystemSettings` tries to access `TwoFactorAuthentication` during construction creating a cyclic dependency. You will probably have to refactor and move the logic in `TwoFactorAuthentication::isUserUsingTwoFactorAuthentication()` to a static method and access that in `SystemSettings`.

plugins/TwoFactorAuth/SystemSettings.php
andyjdavis: Done. Thanks :)

andyjdavis:
Hi @diosmosis I've had a go at avoiding the circular dependency and I think I've got it all working. It was fairly straightforward to make TwoFactorAuthentication::isUserUsingTwoFactorAuthentication() a static method. It calls a few private member functions which I have also made static. Luckily it didn't turn out to require anything that made having an instance of TwoFactorAuthentication necessary.

I believe I have found and updated every location calling TwoFactorAuthentication::isUserUsingTwoFactorAuthentication()

Let me know if the way I have done this isn't ok. As an alternative I could leave the existing method as non-static then add a new static method. The non-static method could just call the static method. That would reduce the size of the diff and mean that any other third party code (plugins etc) not in this repo that use the TwoFactorAuthentication class wouldn't need to change. I'm not sure if that is a risk here.

andyjdavis:
Looks like there are some problems I need to fix. I will get onto them as soon as I am able.

andyjdavis:
Update. There is a failing test for which I don't yet have a solution that I am happy with plus the last commit in this PR definitely shouldn't be merged. I am setting this PR to draft until I get it into a state where i believe it can be merged (pending review of course).

This problem is that the superuser created for system tests does not have 2fa set up. That means that don't have the ability to turn on the system setting requiring all users to use 2fa. One of the 2fa related system tests toggles on the require 2fa setting but this now results in an error.

diosmosis:
@andyjdavis I took a quick look and looks like the latest build is passing (except for BlobReportLimitingTest which is a random failure). Unless I'm missing something, I'm happy to review again tomorrow.

andyjdavis:
@dio The last commit currently in this PR makes the test pass but I'm pretty sure it largely defeats the point of the 2fa check. Apologies for leaving that commit up. I'm at work right now so can't remove it.

https://github.com/matomo-org/matomo/pull/17400/commits/4e6af4893795e69de78dbbf5f1a24c257a1ef304

diosmosis:
Ok @andyjdavis, taking a look at that code, a quick fix for the problem would be to replace `Piwik::hasSuperUserAccess()` with `defined('PIWIK_TEST_MODE')`. No need to go through too much trouble for this PR, though, we know it's not your job :+1:. You're free to finish it up in your own time or if you don't have the time to spare one of us can put the finishing touches in. (If you want to finish the PR yourself, then you're right about it being possible to simplify the static method change. If it would make the PR effectively smaller it would be a good thing to do.)

andyjdavis:
>defined('PIWIK_TEST_MODE')

@diosmosis this makes things so much simpler than what I was trying to do :) The tests should now pass.

When not in test mode $setting->setIsWritableByCurrentUser() relies on whether the current user is using 2fa.

diosmosis:
Hi @andyjdavis, thanks for making the change! The code looks good. I tested the change locally and I noticed the removal of the field could be a problem, since a user may never know the option exists. Instead of using `setIsWritableByTheCurrentUser`, can you instead use `uiControlAttributes` to set a `disabled="disabled"` property on the form field? Something like:

```
if (!$isUsing2FA) {
    $field->uiControlAttributes = ['disabled' => 'disabled'];
}
```

andyjdavis:
@diosmosis I had the same usability concern but wasn't sure how to go about it. I've now pushed a commit that disables the setting without removing it. I have also updated the string to give the user some information on what they need to do to get the setting to enable.

diosmosis:
Thanks @andyjdavis, the change looks good to merge now! Thanks again for the contribution and for following through with the requested changes.

andyjdavis:
@diosmosis thank you for your patience while I worked through this :)

PR #17402, opened 2021-03-28
fix ScheduledReports.sendReport API

Findus23:
It seems like no one (apart from @Chardonneaur) used that API in years as it always returns 
```json
{
  "result": "error",
  "message": "The period 'never' is not supported. Try any of the following instead: day, week, month, year, range"
}
```

even if the period is definitely not set to `'never'`. It seems like the wrong variable is used here, as `$report['period_param'] = $period;` is set above.

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (approved):

sgiehl:
Wondering why there actually is no test that failed because of that.

Findus23:
I assume everything (apart from the HTTP API which isn't used by anyone) uses 
https://github.com/matomo-org/matomo/blob/fc2e8a5ea1d37a8db37e9e2330ba67faacdc71a2/plugins/ScheduledReports/ScheduledReports.php#L318
directly

tsteur:
We're using this API in our app itself eg in the tasks to send every report etc. Also when clicking in the UI on "send report now" so would have assumed this worked before? Just wanting to make sure we won't regress anything

diosmosis:
@tsteur the task only supplies the report ID, not an override period. I'm writing a quick test for it now.

diosmosis:
Actually just seeing the angular controller uses the API directly w/ the period param...

diosmosis:
I see, it happens if the report's email schedule is set to a value that doesn't map to a matomo period type. (We use the correct value when generating the report, but pass the wrong value to the sendReport event; but I don't think it's used much there.)

Issue #17344, opened 2021-03-14
Automatically comment on PRs after inactivity


tsteur:
see https://github.com/matomo-org/matomo/issues/17318#issuecomment-793471155

In #17318 we're discussing about following up automatically when a PR has been waiting for a review for too long. This issue is about automatically following up when we're waiting for the PR author.

We could maybe follow up with (community) PRs automatically if they have been inactive for a while for example when we're waiting for changes for say 2 weeks or more.

After 6 weeks we could close the PR and remind the PR author that we'll reopen the PR again once work continues etc.

so far we have this message for when a PR is more than 1 month old: https://github.com/matomo-org/matomo/wiki/Triage-the-issue-trackers#pull-request-more-than-1-month-old

This logic would ideally also apply to our own PRs but we'd need to have a custom label that we can assign to ignore this logic as sometimes we want to keep a PR open for longer.

I wonder though how we'd know if a PR is inactive for too long? I suppose it would be if there PR was last marked as `Review: Changes requested` and there hasn't been a commit since then?

@flamisz recommended https://github.com/actions/stale https://github.com/marketplace/actions/close-stale-issues for closing the issues automatically. We'd also want to comment automatically though.

fyi @mattab

tsteur:
Any thoughts @mattab @sgiehl @flamisz @diosmosis ?

flamisz:
@tsteur the action I recommended can comment and add a label on the pr or issue first, and closing later (can setup how many days of inactivity triggers what). Lots of options. 
I added to one of my private repos and set it up for dry run. It means it doesn't add any comment, label, or closing the pr, only we can see in the run log, what would have happened in live mode.

sgiehl:
For this case we could even use an GitHub App instead of an action. e.g. https://probot.github.io/apps/stale/

tsteur:
@flamisz great that it could comment first say after 2 weeks and then close after 6 weeks. Does this time start from PR creation? Is there any chance to only start the time measurement for example from the time `Needs Review` label was added or once the PR is no longer `Draft` and then the time from the last commit or so?

So far it says `had no activity for a specified amount of time` but not sure how they define activity. It might work though as well if we just say if nothing happened for say 2 weeks in a PR then leave a comment.

michalkleiner:
We currently have three actions to mark stale issues and ping core reviewers on PRs that need review and didn't have activity for some time.

[...]

Issue #17318, opened 2021-03-08
Making sure we review all pull requests


tsteur:
Hi there,

currently we have a few PRs that are waiting for reviews for > 10 days like https://github.com/matomo-org/matomo/pull/17264 and https://github.com/matomo-org/matomo/pull/17266

I wonder if we could have a github action or some other reminder that pings all of us if there was an issue with "needs review" and there was no comment or review for 7 days? Or maybe even just 5 days? Of course it's not always that trivial since we might be waiting for a change from the author etc but maybe also https://github.com/matomo-org/matomo/issues/17317 would simplify the logic a bit.

We could maybe start with a simple reminder once there has been no comment from us for say 3 or 5 business days (or 5-7 regular days) and then see how this works over time and could then make adjustments to it if needed. The added comment could say something like `This issue is in "needs review" but there has been no comment for X days. @tsteur ...` we would then ping all of us maybe.

Or maybe the issue was simply that there was no milestone assigned in which case we'd want to make sure a `Needs Review` PR always also has a milestone?

Or are there better/other ways maybe? Any thoughts appreciated @diosmosis @Findus23 @flamisz @sgiehl @mattab 

Maybe also #17317  would make things better and a new github action wouldn't be needed. Not sure it's even possible to solve with a github action.

In https://github.com/matomo-org/matomo/issues/17344 there's also an issue for automatically following up when we're waiting for a PR author for too long.

diosmosis:
Lately the problem I've had is just not having time, w/ deploys and L3 issues. That's mostly died down though, so I should be able to get back to reviewing regularly. It could be prioritized higher I suppose (ie, pr reviews before investigating L3 issues).

sgiehl:
@tsteur such stuff should be possible with github actions for sure. Maybe there are already some matching actions like: https://github.com/marketplace/actions/pull-request-reviewer-reminder-action

flamisz:
This action is promising:

https://github.com/actions/stale
https://github.com/marketplace/actions/close-stale-issues

tsteur:
@sgiehl regarding https://github.com/marketplace/actions/pull-request-reviewer-reminder-action

do you know from when they start the timer to measure the turnaround time? Maybe it won't work for our case regarding the labels etc?

@flamisz for closing issues automatically created https://github.com/matomo-org/matomo/issues/17344

sgiehl:
@tsteur yes, that won't work depending on labels. guess the time is calculated based on the creation time.
[...]

PR #17404, opened 2021-03-29
add a new stale action

flamisz:
### Description:

refs #17318 #17344 

I checked the logs of the dry runs and actually did not find any when actually it would comment on any or. The current setup checks only PRs (not issues) and only PRs were created after 2021-03-01.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/.github/workflows/follow-up-reviews.yml
+++ b/.github/workflows/follow-up-reviews.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/stale@v3
         with:
-          debug-only: true
+          debug-only: false
           start-date: '2021-03-01T00:00:00Z'
           days-before-pr-stale: 7
           days-before-pr-close: -1
--- a/.github/workflows/follow-up-reviews.yml
+++ b/.github/workflows/follow-up-reviews.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/stale@v3
         with:
-          debug-only: false
+          debug-only: true
           start-date: 2021-03-01T00:00:00Z
           days-before-pr-stale: 7
           days-before-pr-close: -1
--- a/.github/workflows/follow-up-reviews.yml
+++ b/.github/workflows/follow-up-reviews.yml
@@ -10,11 +10,11 @@ jobs:
       - uses: actions/stale@v3
         with:
           debug-only: false
-          start-date: '2021-03-01T00:00:00Z'
+          start-date: 2021-03-01T00:00:00Z
           days-before-pr-stale: 7
           days-before-pr-close: -1
-          only-pr-labels: 'Needs Review'
+          only-pr-labels: Needs Review
           days-before-issue-stale: -1
           days-before-issue-close: -1
-          stale-pr-message: 'This issue is in "needs review" but there has been no activity for 7 days. ping @tsteur @sgiehl @diosmosis @flamisz'
-          stale-pr-label: 'Stale'
+          stale-pr-message: This issue is in "needs review" but there has been no activity for 7 days. ping @tsteur @sgiehl @diosmosis @flamisz
+          stale-pr-label: Stale
--- a/.github/workflows/inactive-prs-closing-message.yaml
+++ b/.github/workflows/inactive-prs-closing-message.yaml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/stale@v3
         with:
-          debug-only: false
+          debug-only: true
           start-date: 2021-03-01T00:00:00Z
           days-before-pr-stale: 42
           days-before-pr-close: -1
--- a/.github/workflows/inactive-prs-closing-message.yaml
+++ b/.github/workflows/inactive-prs-closing-message.yaml
@@ -1,4 +1,4 @@
-name: 'Handle inactive PRs'
+name: 'Handle inactive PRs - closing message'
 on:
   schedule:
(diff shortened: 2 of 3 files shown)


flamisz:
let me know if you happy with this @tsteur

tsteur:
thanks @flamisz that's great. Do you have maybe a link to a summary what actions it would do?

Can we maybe tweak `This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.` and add something like "If you don't want this PR to be closed automatically in ... days then you need to assign the label 'Do not close'."

@mattab @sgiehl @diosmosis @Findus23 be great to read over the config as well. There are two different actions one for `Not Review in 7 days` and one for stale PRs where there was no activity for a certain amount of time.

sgiehl:
Does that really work as expected. The action actually produces this warning: `Warning: Reached max number of operations to process. Exiting.`
Sounds like it would go through all issues and exit as there are too many?

flamisz:
> Sounds like it would go through all issues and exit as there are too many?

There is a limit on operations, you can read about it here: https://github.com/actions/stale#operations-per-run

flamisz:
@tsteur I modified the stale message. Please check if it is ok now. 

>Do you have maybe a link to a summary what actions it would do?

What do you mean? A summary about dry runs, so we know what would have happened?

tsteur:
> What do you mean? A summary about dry runs, so we know what would have happened?

@flamisz yes that be great. Just so we know it would do the right thing and that it'll work and won't the wrong PRs etc

flamisz:
Short summary for the `follow up reviews` action:

Yesterday it found this PR #17394 and wanted to stale it, which means add a `Stale` label and a comment message:
"This issue is in "needs review" but there has been no activity for 7 days. ping @tsteur @sgiehl @diosmosis @flamisz"

I checked the result for a run couple of days ago, and it didn't stale it, because it was updated less than 7 days ago (that time).

The same happened with this one #17387.

Skipped all the PR without the required label (Needs review).

It goes through the issues/PRs by the last updated date, descending. It runs 30 operations per run (by default). I'm not sure what is one operation when it's in production mode, so we will see how it works, how many PRs will be checked per run. It runs once daily.

flamisz:
Handle inactive PRs 

A couple of pr (like #17333 ) got stale during yesterday's run.

No PR has been closed, because we check PRs only from 1st March and the days before close is set to 42.

tsteur:
Thanks @flamisz 

Any more thoughts @mattab @sgiehl @diosmosis @Findus23 ? We'd otherwise activate it and would just need to keep an eye on it that we don't accidentally close a false PR. 

Regarding closing stale PRs. I wonder if we'd need to change to something like this if that's possible:

* If the contributor was last active (like pushed a commit) then we wouldn't want to close the PR automatically maybe because in this case we are late (we are the problem) and it's not the contributor we're waiting on. In that case we'd want to only comment again and ping us to review and follow up
* Vs if we commented last and the contributor hasn't reacted in a while then we'd want to close the PR.

Not sure if that makes sense or is clear what I'm thinking of? Fingers crossed it's possible to configure this.

flamisz:
> If the contributor was last active (like pushed a commit) then we wouldn't want to close the PR automatically maybe because in this case we are late (we are the problem) and it's not the contributor we're waiting on. In that case we'd want to only comment again and ping us to review and follow up
Vs if we commented last and the contributor hasn't reacted in a while then we'd want to close the PR.

Unfortunately, I don't think it's that flexible. I take a look, but I doubt it. The best we can do is checking the labels probably, but that has to be handled by us, and we forget it, etc. 
I double-check what are the options we have.

flamisz:
@tsteur I checked the action, what is the best we can do, regarding to checking who was the last who did something with the PR. 
Like I thought, there is no setting for that. The best option we can have if you use the `Waiting for user feedback` label whenever we comment on a PR and we really wait for the user. 
This way the action can test if the label is there, and only stale/close it if that's the case.

Do you want me to add this to the yaml file?

flamisz:
@tsteur we were talking about to do not close the PR automatically only add a comment, but this is not possible with this action. It closes the PR when adds the close comment and optional label.

tsteur:
@flamisz don't we also just add a comment with the other action after 7 days? Was hoping it could look similar to https://github.com/matomo-org/matomo/pull/17404/files#diff-487d40658fdc47419d39d147a502671a427a1df852c41b5555b382ae515111d7R12-R20 or maybe I forgot about something?

flamisz:
> @flamisz don't we also just add a comment with the other action after 7 days? 

@tsteur sure we can do that, what we can't do is add a comment after 7 days, and then a new one after 42 days, because the 2nd comment would close the PR. If we want to make 2 warnings (one at day 7, and one at day 42) we need a new action.

What do you think we should do here? We can add a comment with mentions to us at day 7, and than close it or not close it at day 42? We can  add a new comment with mentions on that day too, and we can reopen if it was a mistake.

tsteur:
We could use a new action and we wouldn't want to close it automatically. Only a comment be great 👍

tsteur:
we can then always change later again

flamisz:
@tsteur I added a new action. Do you think we should give it a try like this, but still just dry-run? I week maybe, and we can go live the week after.

tsteur:
sure sounds good @flamisz

flamisz:
I test with the new settings for one more week

PR #17408, opened 2021-03-30
store segment hash in DB

### LINKED ISSUE #17301: Store the segment hash in the DB table
refs https://github.com/matomo-org/matomo/issues/17050 which we're currently still discussing and looking into.

In any case I'm thinking to fix the segment hash problem hopefully once and for all I think we should introduce a new column to the segment DB table that stores the segment hash that will be used when fetching archives etc. This will also make it future proof if we ever change the segment definition and/or if we change the calculation for hashes etc without needing to update any archives etc.

We would basically need an update script that sets the correct hashes for any existing segment. And then adjust https://github.com/matomo-org/matomo/blob/4.2.1/core/Segment.php#L464 to try and find a matching hash from the segment DB table. We might need to encode / decode a segment definition to see if we find a match in the table.

@diosmosis do you think that would be useful? If so I reckon we could start working on this already as it's independent of a decision for #17050 which will probably still need a while (@mattab and I already quickly talked about it but need more time)

flamisz:
### Description:

fixes #17301  

### Review

* [ ] Functional review done
* [ ] Code review done


flamisz:
Hi @tsteur and @diosmosis,

this is very WIP, but I'd like to know if I go in the right direction.

A couple of questions I have:
- should I check if the hash column exists or the updater will solve it anyway (I mean it can not happen that the code updated but DB not?)
- when we create the update file? whenever we need a migration? so this PR should contain a new update file?
- how we fill the hash column for the existing segments in the db? does that go into the update file?
- @tsteur mentioned in the issue, ideally we can remove the `urlencodes` from this PR: https://github.com/matomo-org/matomo/pull/17029/files. can we @diosmosis?

sorry I have too many questions, I just'd like to understand more and do properly. thanks

diosmosis:
> should I check if the hash column exists or the updater will solve it anyway (I mean it can not happen that the code updated but DB not?)

It shouldn't happen that the code is updated but DB is not updated. If it does happen, then it's because of a bug in the update process, or because of user error. So you shouldn't have to check if the added column exists. (There are exceptions in this because sometimes during the one click update, some new code will execute after it is written to the hard disk, while the old database hasn't been updated, but this shouldn't be one of those cases.)

> when we create the update file? whenever we need a migration? so this PR should contain a new update file?

Yes :+1: and updates are tied to a version, so first you'd bump the version in `Version.php` (setting to a new beta or rc based on what's in there already).

> how we fill the hash column for the existing segments in the db? does that go into the update file?

Yes :+1: it should be initialized to `urlencode($segment['definition'])` in the update file.

> @tsteur mentioned in the issue, ideally we can remove the urlencodes from this PR: https://github.com/matomo-org/matomo/pull/17029/files. can we @diosmosis?

I think it depends on how this is implemented. To get rid of those urlencodes, in the `Segment` constructor you'd have to search through every stored segment in the db for a definition that matches, then if found, use the hash in the database.

> sorry I have too many questions, I just'd like to understand more and do properly. thanks

from what matt & thomas said, questions are to be encouraged :+1:

diosmosis:
I guess we don't have docs for writing updates/migrations, I'll put it on my list to try and write them (unless someone gets to them first).

flamisz:
@tsteur and @diosmosis I'm wondering what else we need in this PR?
Can we get rid of those `urlencode` calls from the other PR? I checked that part of the code, and most of the time it gets the hash in the end of the process from the `Segment`, and now the `getSegmentHash` method does its best to find it (without anything, with `urlencode` and with `urldecode` as well).

core/Updates/4.3.0-b2.php
```diff
@@ -0,0 +1,55 @@
...
+    public function getMigrations(Updater $updater)
+    {
+        $migrations = [];
+
+        $migrations[] = $this->migration->db->addColumn('segment', 'hash', 'CHAR(32) NULL', 'definition');
+
+        $segmentTable = Common::prefixTable('segment');
+        $segments = Db::fetchAll('SELECT idsegment, definition from ' . $segmentTable);
+        foreach ($segments as $segment) {
+            $hash = md5(urlencode($segment['definition']));
```
diosmosis: This should be `md5($segment['definition'])`. Apologies, I know I said it needs to be `urlencode`d, I was referring to Segment construction. The `Segment::getSegmentHash()` function decodes the definition before `md5`-ing, so:

```
$segment = new Segment(urlencode($segment['definition']));
$hash = $segment->getHash();
```

is the same as

```
md5(urldecode(urlencode($segment['definition'])))
```

So if not going through `Segment`, calling `md5` directly works (but only for definitions in the segment table).

flamisz: @diosmosis and @tsteur there is an issue with this one.
If we have a segment currently in a db like this `actions%3E%3D2`, current Matomo version will use the hash of the `urlencoded` version of it, so when we create the hash here:

```php
$hash = md5($segment['definition']);
```

it won't be the same as we have in the `archive_numeric` tables.
One solution would be using `$hash = md5(urldecode($segment['definition']));` in the update file. 
We don't support double encoded segments in the db, do we? Only in URL/API?

diosmosis: The segment condition value is encoded multiple times, not the segment condition operator. 

The hash is computed overall with the fully urlencoded segment, because that is what is sent in the `segment` parameter. The stored segments are created by sending a `definition=` parameter, which does not have special treatment and gets urldecoded once automatically. So there is a discrepancy with what is in the stored segment table and what is sent as a query parameter (`segment=urlencode(pageUrl==urlencode(urlencode(segmentValue)))`. The hash in the table has to match the hash computed w/ the fully encoded segment. The tests should cover this.

diosmosis: The hash that gets put in the table needs to be the same hash that is created in this test: https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/SegmentEditor/tests/System/ApiTest.php#L21 when run on 4.x-dev.

flamisz: thanks, @diosmosis, much more clear now.

core/Segment.php
```diff
@@ -460,6 +464,32 @@ public function getHash()
 
     public static function getSegmentHash($definition)
     {
+        $cache = Cache::getEagerCache();
+        $cacheKey = self::CACHE_KEY . md5($definition);
+
+        if ($cache->contains($cacheKey)) {
+            return $cache->fetch($cacheKey);
+        }
```
flamisz, line 473: I cache segments, can you please check @diosmosis if this is a good way to use the cache? I used eager cache, because every request calls this, even if we request all visits.

diosmosis: Some matomo users create a very large amount of segments, so loading these on every request might be an issue. I'm also not sure if the tracker uses the eager cache, but it seems possible to me that it could hit some code that uses it and loads all this data. If the issue is to avoid hitting the database, then using the Lazy cache should be ok? It would just hit the filesystem. 

Does this code also get executed if `$definition == ''`? There could be a special check for empty string if so.

Can you list some of the cases where this is called even when there is no segment supplied? (For example, when `$definition == ''`, etc.) Does it happen only on API requests to get data or on requests, eg, in the admin section and elsewhere?

cc @tsteur in case he has a different opinion.

tsteur: I'd probably use a lazy cache here simply because it shouldn't be needed in most requests (assuming the method is not called for an empty definition string or maybe for an empty definition string we could always return same hard coded `return md5('');`? ). Assuming this changes rarely we could cache them for quite a long time.

If the segment hash can ever change then we'd need to invalidate the cache on update. 

Initially I thought to cache only the list of definitions and their hashes and then iterate over them every time (like `[['definition' => '...', 'hash' => '...'], [...], [...]]`) to find the right hashes so we don't create too many cache files if there are heaps of segments (as it can slow down the filesystem access) but then of course the downside is needing to load a larger cache file as it includes the information for all caches. I'd probably rather go with this kind of catch and match a segment by looping over the arrays but either way might be fine. Both have their benefits and disadvantages and it's not clear which one is better.

Something to still look at is how to handle the idSites and when the same definition was to exist for different sites with different hashes. This one may be super tricky. It wouldn't be too important right now as they technically should have all the same hashes for the same definition but this might not always be the case due to bugs in the past and if the calculation in the future changes.

flamisz: We call it a LOT. On a page reload (eg on the devices report), without choosing any segment, there are 300 calls for that method. Then click on another report (ajax calls), we have more than 200 calls again.
Currently, it does not resource-intensive, because we don't hit the DB, but we will with the new version.
If we go with the lazy cache, @tsteur list idea is good, because that way we don't need multiple files.

@diosmosis we have a check for the empty definition, but it looks like some plugins load all the segments even if the definition is empty (we call the `getHash` even more, but if the `definition` is empty, it doesn't call the other method).

flamisz: I tested again, how many times we call the function. 
The tracker didn't call it.
The dashboard is the worst. This time it called it more than 2000 times when browser archiving was on, and around 1000 times when it was off.
Other pages call it less. 

But it's much more if I choose any segments. The numbers above were with all visits.

I checked the eager cache file for the demo site. Currently, it's 530K. When I added 100 segment hashes (by hand, locally) it went up to 539K. I believe that won't be an issue.

diosmosis: (Answering because I don't see a reply from @tsteur. If you've already discussed it w/ him, feel free to ignore my comment.)

You're right it doesn't seem like adding segment hashes will cause the eager cache to get significantly larger. Unless there's, say, a matomo instance w/ 1000 users and each one for some reason creates an average of 50 segments (which isn't entirely implausible since I've seen segments like `pageUrl==...,pageUrl==...,pageUrl==...,...` for like maybe 50 different page URLs). In this case we'd add potentially 50000 hashes to the eager cache.

I also believe you when you say the tracker didn't call it or use the eager cache. The tracker and the rest of matomo, however, aren't strictly separated, so it would be pretty easy for some change in the future to accidentally trigger a code path that uses the eager cache, even without ever using the `Segment` class. Maybe in the case of the hyopthetical user above, this happens, and tracking slows down a bit. Since tracking is meant to handle a high volume of concurrent requests, any performance degradation can be seen as a problem.

Both of these situations are hypothetical and would not be common if they happened at all, but edge cases always happen in some form, and this one would be particularly annoying, since the only sign of an issue we'd have is "tracking is a little slower". And all of these edge cases could be avoided by just using the Lazy cache. Which, to my knowledge, would not be noticeably slower than if we used the eager cache here.

I also believe you when you say the segment hashing method is called many times on every request. However, that's clearly a surprise, and not something we expect or want. So ideally we'd get some more context as to why that happens, where it happens, and what we can do about it, rather than optimize around something that we don't to happen in the first place.

All of that said, I am not the authority here, so @tsteur can of course give his opinion as well.

flamisz: thanks for the comment @diosmosis.
We had a chat with @tsteur this morning. I'm pretty sure lots of edge cases can happen, and I can't see through the codebase as much as you, so every comment and recommendation is very helpful.

I modified the caching and added comments to the code. Can you please both check it and let me know what you think about it?

Another idea I have: currently we use the hashing in a couple of places, so probably the `createHash` method in the `SegmentEditor` model should be public static, so we call that everywhere, this way when we modify it, we just modify in one place.

core/Segment.php
```diff
@@ -460,8 +465,42 @@ public function getHash()
...
+        // something like "visitorType==new"
+        if ($cache->contains(self::SEGMENT_HAS_BUILT_CACHE_KEY)) {
+            return $defaultHash;
+        }
+
+        // the segment hash is not built yet, let's do it
+        $cache->save(self::SEGMENT_HAS_BUILT_CACHE_KEY, true);
+
+        $model = new SegmentEditorModel();
+        $segments = $model->getAllSegmentsForAllUsers();
```
tsteur: @flamisz to remove the amount of data and speed things up a bit we could add a new query to only fetch `hash` and `definition`. It's not too super important though unless there were really heaps of entries. So maybe for now we skip this as it's likely not needed.

core/Segment.php
```diff
@@ -460,8 +465,42 @@ public function getHash()
...
+        $foundSegment = false;
+
+        foreach ($segments as $segment) {
+            $cacheKey = self::CACHE_KEY . md5($segment['definition']);
+
+            $cache->save($cacheKey, $segment['hash']);
+
+            if ($segment['definition'] === $definition || $segment['definition'] === urldecode($definition) || $segment['definition'] === urlencode($definition)) {
```
tsteur: not sure I understand this part? I think I would have maybe expected something like this:

```php
foreach ($segments as $segment) {
            $cacheKey = self::CACHE_KEY . md5($segment['definition']);
            $cache->save($cacheKey, $segment['hash']);
            
            $cacheKey = self::CACHE_KEY . md5(urldecode($segment['definition']));
            $cache->save($cacheKey, $segment['hash']);
            
            $cacheKey = self::CACHE_KEY . md5(urlencode($segment['definition']));
            $cache->save($cacheKey, $segment['hash']);
}
   if ($cache->contains($cacheKey)) {
            return $cache->fetch($cacheKey);
        }
        return $defaultHash;
```

Not sure if that makes any sense though? I'm just wondering if otherwise we wouldn't find the correct cache key depending if the segment definition was passed urlencoded or urldecoded etc. Not sure it's clear what I mean?

core/Segment.php
```diff
@@ -460,8 +465,42 @@ public function getHash()
...
+
+        // if the cache for segments already built, but this segment was not found,
+        // we return the default segment, this can be a segment from url or
+        // something like "visitorType==new"
+        if ($cache->contains(self::SEGMENT_HAS_BUILT_CACHE_KEY)) {
+            return $defaultHash;
+        }
+
+        // the segment hash is not built yet, let's do it
+        $cache->save(self::SEGMENT_HAS_BUILT_CACHE_KEY, true);
```
tsteur: just to be safe we would maybe want to set this one after the `for` loop and after all cache entries were set correctly. Just in case there was some exception in the for loop and we would maybe then write a partial cache only but assume we have written the full cache. not sure what would happen in case of an exception etc and if it's even possible but just meaning to be safe.

tsteur (changes requested):
@flamisz looks like that should work 👍  left some minor comments

core/Segment.php
flamisz: I see what you mean, we save all versions for every cache. But the `cache->contains` call after this will check all the caches, even that is not stored in the file yet, only saved? Is that correct?

core/Segment.php
flamisz: make sense 👍

core/Segment.php
tsteur: I'm not 100% sure what you mean but I think it's correct. 👍

flamisz:
Hi @diosmosis, can you please review it again (hopefully last time) and we can merge it then?

diosmosis:
When the tests pass, I'll take another look :+1:

plugins/SegmentEditor/tests/System/ApiTest.php
```diff
@@ -62,4 +62,50 @@ public function test_segmentHashWorkflow_whenSegmentIsCrazyEncoded()
...
+            'autoArchive' => 1,
+            'enabledAllUsers' => 1,
+            'format' => 'json',
+            'token_auth' => Fixture::getTokenAuth(),
+        ]);
+        self::assertStringContainsString(urlencode($segment), $url);
+
+        Http::sendHttpRequest($url, 10);
+
+        $segments = SegmentEditorApi::getInstance()->getAll();
```
tsteur: btw here you will want to fetch a specific segment using the `get($idSegment)` method and not all of them maybe. When you use above `$idSegment = SegmentEditorApi::getInstance()->add()` then you will also have the ID.

tsteur: just in case the segments returned... the new one might not be at the end (unless there is some sort)

tests/PHPUnit/Framework/TestCase/SystemTestCase.php
```diff
@@ -76,6 +76,8 @@ public static function setUpBeforeClass(): void
     {
         Log::debug("Setting up " . get_called_class());
 
+        gc_disable();
```
tsteur: this one we'll likely want to remove later again @flamisz

flamisz:
please note: this pr still contains the `testdox="true"` in the `phunit.xml`. Do we want to keep this? Sometimes it can be helpful to see the tests one by one, but we can always turn this on only when we need it.
@tsteur ?

tsteur:
sounds very useful to me @flamisz 🚀 💯  I'd set this in a new PR though so the team can review it separately in case someone has a good reason for not having it

core/Segment.php
```diff
@@ -460,8 +465,46 @@ public function getHash()
...
+
+        foreach ($segments as $segment) {
+            $cacheKeyTemp = self::CACHE_KEY . md5($segment['definition']);
+            $cache->save($cacheKeyTemp, $segment['hash']);
+
+            $cacheKeyTemp = self::CACHE_KEY . md5(urldecode($segment['definition']));
+            $cache->save($cacheKeyTemp, $segment['hash']);
+
+            $cacheKeyTemp = self::CACHE_KEY . md5(urlencode($segment['definition']));
+            $cache->save($cacheKeyTemp, $segment['hash']);
```
diosmosis, line 496: Is it possible to write a test to ensure this logic doesn't result in collisions between different segments? For example the segments:

```
pageUrl==http://abc.com/d+f
pageUrl==http://abc.com/d f
pageUrl==http://abc.com/d%20f
```

and maybe also including operator characters in segment values like:

```
pageUrl==http://abc.com/=/@/!/,/;
```

(segments not encoded correctly for readability)

flamisz: ok, I'm on it

flamisz: @diosmosis 
with the current logic, the hash for these definitions is the same, because we `urldecode` the definitions before we `md5` them when we store them in the DB. 

```php
    private function createHash($definition)
    {
        return md5(urldecode($definition));
    }
```

Is that something we don't want?

flamisz: I added new tests @diosmosis

tests/PHPUnit/Integration/Segment/SegmentsCacheTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $segment3 = $model->getSegment($idSegment3);
+        $key1 = Segment::CACHE_KEY . md5($segment1['definition']);
+        $key2 = Segment::CACHE_KEY . md5($segment2['definition']);
+        $key3 = Segment::CACHE_KEY . md5($segment3['definition']);
+
+        Segment::getSegmentHash('dummy');
+
+        $this->assertEquals($segment1['hash'], $cache->fetch($key1));
+        $this->assertEquals($segment2['hash'], $cache->fetch($key2));
+        $this->assertEquals($segment3['hash'], $cache->fetch($key3));
```
diosmosis: Can you also add assertions to check the contents of these cache keys (for each segment):

```
            $cacheKeyTemp = self::CACHE_KEY . md5(urldecode($segment['definition']));
            $cache->save($cacheKeyTemp, $segment['hash']);

            $cacheKeyTemp = self::CACHE_KEY . md5(urlencode($segment['definition']));
            $cache->save($cacheKeyTemp, $segment['hash']);
```
?

flamisz: I updated the tests, still green 😄  I wasn't sure TBH, but that's why we need tests 🚀

tsteur:
@flamisz @diosmosis was wondering if there's much left in this issue to do? Asking as we'd like to have this issue included in the next release

diosmosis:
@tsteur @flamisz merged it

PR #17410, opened 2021-03-30
Fixes possible notice

sgiehl:
### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/SupportedBrowser.php
+++ b/core/SupportedBrowser.php
@@ -50,7 +50,7 @@ class SupportedBrowser
         $deviceDetector->parse();
         $client = $deviceDetector->getClient();
 
-        if ($client['type'] === 'browser' && self::browserNotSupported($client['short_name'], (int)$client['version'])) {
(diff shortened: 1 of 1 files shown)


core/SupportedBrowser.php
```diff
@@ -50,7 +50,7 @@ public static function checkIfBrowserSupported()
         $deviceDetector->parse();
         $client = $deviceDetector->getClient();
 
-        if ($client['type'] === 'browser' && self::browserNotSupported($client['short_name'], (int)$client['version'])) {
+        if (!empty($client) && $client['type'] === 'browser' && self::browserNotSupported($client['short_name'], (int)$client['version'])) {
```
sgiehl, line 53: If device detector isn't able to detect any client the result of `getClient()` actually can be `null`, which then resulted in a notice here.

flamisz: good catch 👍

flamisz (approved):

PR #17412, opened 2021-03-30
track visits in JSTracker UI test to separate site so it does not affect most other tests screenshots

diosmosis:
### Description:

As title, should stop some random failures in the future.

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl:
@diosmosis the test is failing. Seems the site that's used now actually doesn't exist

diosmosis:
Noticed the UI test uses a custom fixture, so this wouldn't affect the other tests... Not sure why the others change when the order changes, then. Closing this one.

PR #17413, opened 2021-03-31
Use exceptions to provide more context to warning logs in Row.php

diosmosis:
### Description:

The warning logs in Row.php are pretty vague about what is actually going on to cause them, making them hard to reproduce and debug. Using exceptions, however, we can get stack traces in the right context.

Looks a bit odd, but it seems to work.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/DataTable/Row.php
+++ b/core/DataTable/Row.php
@@ -756,7 +756,7 @@ class Row extends \ArrayObject
                 implode(", ", $this->getColumns()),
                 $this->getIdSubDataTable()
             ));
-            StaticContainer::get(LoggerInterface::class)->warning("{exception}", ['ex' => $ex]);
+            StaticContainer::get(LoggerInterface::class)->warning("{exception}", ['exception' => $ex]);
         }
     }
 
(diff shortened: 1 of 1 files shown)


core/DataTable/Row.php
```diff
@@ -748,25 +750,27 @@ public static function isEqual(Row $row1, Row $row2)
...
-            Log::warning(
+            $ex = new \Exception(sprintf(
                 "Trying to add two strings in DataTable\Row::sumRowArray: %s + %s for column %s in row %s",
                 $thisColumnValue,
                 $columnToSumValue,
                 $columnName,
                 $this->__toString()
-            );
+            ));
+            Log::warning("{exception}", ['ex' => $ex]);
```
sgiehl: Any reason you changed that to using `StaticContainer` above, but not here? Might be good to have that consitent

sgiehl (changes requested):

core/DataTable/Row.php
diosmosis: no reason, I just missed that entirely

diosmosis:
Updated. PR was very incorrect before.

sgiehl (approved):
code looks fine now

PR #17416, opened 2021-03-31
better systemcheck error boxes

Findus23:
The system check `LongErrorMessage`-boxes were missing the bottom border making them look a bit weird:
image

There is unfortunately a pretty specific rule from another part of Matomo:
```css
table.entityTable tbody tr td {
     border-bottom: 1px solid #f2f2f2 !important;
}
```
so to make this look correct I also had to make the error rule more specific. 
Hopefully the .error class is used nowhere else in Matomo

But I think it should be fine as it is generally defined here:
https://github.com/matomo-org/matomo/blob/703ed9402e4aada544ec94e789021df86575ad15/plugins/Morpheus/stylesheets/simple_structure.css#L188-L193

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Installation/stylesheets/systemCheckPage.less
+++ b/plugins/Installation/stylesheets/systemCheckPage.less
@@ -2,20 +2,21 @@
     width: 40%;
 }
 
-.error {
-    color: red;
-    font-size: 100%;
-    font-weight: bold;
-    border: 2px solid red;
-    width: 550px;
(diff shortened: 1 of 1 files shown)


diosmosis:
The UI tests look ok and doesn't look like it's used from a quick grep, so should be fine to merge :+1:

PR #17426, opened 2021-04-06
Display French Guiana as separate country on map

### LINKED ISSUE #5056: French Guiana shows up as France in maps and reports
Please look at the south america map, Guyana->Suriname->[France]. 
France has to change to French Guiana
http://en.wikipedia.org/wiki/French_Guiana

sgiehl:
### Description:

before it was displayed as `France`. But as the Geo IP providers actually are not detecting it as region of France but as country instead, it makes more sense to display it that way...

fixes #5056

### Review

* [ ] Functional review done
* [ ] Code review done


flamisz (commented):
Unfortunately, when I'm clicking on the country to zoom in, the deepest zoom only shows an empty image instead of the region map of the country. Previously it zoomed to France.
Can we do something with that?

sgiehl:
> Can we do something with that?

We have the "problem" for various smaller countries, as the maps are missing there. See #321

PR #17430, opened 2021-04-07
allow skipping system check entries

Findus23:
suggested in https://forum.matomo.org/t/how-to-ignore-system-integrity-warnings/41368

I'm not entirely sure if this is useful to a lot of people and if this is the best way to implement it, but it seemed like a nice thing for people to be able to ignore specific warnings in their setup and not having to ignore them (and potentially miss others).

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -791,6 +791,9 @@ rearchive_reports_in_past_last_n_months = 6
 ; If set to 1, when rearchiving reports in the past we do not rearchive segment data with those reports. Default is 0.
 rearchive_reports_in_past_exclude_segments = 0
 
+; Add the class of a system check diagnostic here to skip it
+; system_check_ignored_reports[] = "Piwik\Plugins\Diagnostics\Diagnostic\CronArchivingCheck"
+
 [Tracker]
 
 ; When enabled and a userId is set, then the visitorId will be automatically set based on the userId. This allows to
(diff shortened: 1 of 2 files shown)


sgiehl:
@Findus23  that should already be possible by having a config.php like
```php
<?php

return [
    'diagnostics.disabled' => DI\add(['Piwik\Plugins\Diagnostics\Diagnostic\CronArchivingCheck'])
];
```

Findus23:
@sgiehl Many thanks for the info, I totally missed that. This indeed makes this change useless and is also more elegant.

Findus23:
How would someone do this without creating a custom plugin?
Can one place a `config.php` directly into the config directory? It doesn't seem to work for me.

Maybe this could be documented in a very short page on developer.matomo.org

sgiehl:
Haven't tested it. but it should work in a config.php in the config directory 🤔

sgiehl:
Seems it needs to be defined this way:
```
<?php

return [
    'diagnostics.disabled' => DI\add([DI\get('Piwik\Plugins\Diagnostics\Diagnostic\CronArchivingCheck')])
];
```

Findus23:
This works :+1:

MatomoForumNotifications:

https://forum.matomo.org/t/how-can-i-disable-these-alerts/53210/3

PR #17432, opened 2021-04-07
Run JavaScript tests also with node/puppeteer

### LINKED ISSUE #17424: Run Javascript tests with node/puppeteer as well as Phantomjs
followup to https://github.com/matomo-org/matomo/pull/14421

It seems like the JS tests are still using Phantomjs (which isn't maintained anymore).

https://github.com/matomo-org/travis-scripts/blob/e955829a58a5a6af00a40737c663ede98eef5dcb/travis.sh#L28-L33

It might be useful to also use Chromium/Pupeteer there to make sure Matomo JS still works in modern browsers.

sgiehl:
### Description:

The PR also includes some changes to JS tests, so they don't fail in a browser that supports sendBeacon

requires https://github.com/matomo-org/travis-scripts/pull/68
 
fixes #17424

~Before merging this on, the travis-scripts PR needs to be merged and the submodule reference in this PR needs an update.~

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
@sgiehl I'm almost thinking maybe we don't want this just yet as it may be more valuable to know that the tracking works in older browsers and then most likely it also works in newer browsers. Vs if we run it in a newer browser it's harder to ensure it works in an older browser. We may not want this change just yet maybe.

Findus23:
I think the Javascript test is one of the fastest (taking just a few minutes). So maybe it might be possible to run both.

diosmosis:
Just saw this after looking at the travis-scripts PR. I can revert the merge I did if needed.

sgiehl:
running both sounds fair enough. should be simple to do that. will update this PR

sgiehl:
Updated the PR so the travis job should first run the tests on node and afterwards on phantomjs

PR #17435, opened 2021-04-08
Add new validator class - alpha numeric spaces

flamisz:
### Description:

This new class will check if the validated field only contains alphanumeric characters and spaces. No special characters allowed.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/tests/PHPUnit/Unit/Validator/AlphaNumSpacesTest.php
+++ b/tests/PHPUnit/Unit/Validator/AlphaNumSpacesTest.php
@@ -18,12 +18,21 @@ use Piwik\Validators\AlphaNumSpaces;
  */
 class AlphaNumSpacesTest extends TestCase
 {
-    public function test_validate_success_strings()
+    public function test_validate_success_english_strings()
     {
         $this->validate('onlyalpha');
         $this->validate(123);
         $this->validate('abc 123');
-        $this->validate('abc 123 and 0');
+        $this->validate('ABC 123 and 0');
+
+        $this->assertTrue(true);
+    }
+
+    public function test_validate_success_special_chars()
+    {
+        $this->validate('Äpfel äpfel und stühle');
+        $this->validate('워드  信件 λέξη từ ngữ');
+        $this->validate('áâäæãåāëíûî');
 
         $this->assertTrue(true);
     }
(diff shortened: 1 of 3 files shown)


diosmosis:
I think we have unit tests for existing validators so we'd want to add one for this.

tests/PHPUnit/Unit/Validator/AlphaNumSpacesTest.php
```diff
@@ -0,0 +1,58 @@
...
+ * @group AlphaNumSpacesTest
+ */
+class AlphaNumSpacesTest extends TestCase
+{
+    public function test_validate_success_strings()
+    {
+        $this->validate('onlyalpha');
+        $this->validate(123);
+        $this->validate('abc 123');
+        $this->validate('abc 123 and 0');
```
sgiehl: might be good to also test edge cases like German umlauts, some Chinese characters and maybe even some utf8 emojis maybe

sgiehl (changes requested):

tests/PHPUnit/Unit/Validator/AlphaNumSpacesTest.php
flamisz: definitely, I will do that 👍

flamisz: I added more test cases with special characters. Emoji is not supported with this regex. Do you think we should let emojis as well?

tests/PHPUnit/Unit/Validator/AlphaNumSpacesTest.php
```diff
@@ -0,0 +1,67 @@
...
+class AlphaNumSpacesTest extends TestCase
+{
+    public function test_validate_success_english_strings()
+    {
+        $this->validate('onlyalpha');
+        $this->validate(123);
+        $this->validate('abc 123');
+        $this->validate('ABC 123 and 0');
+
+        $this->assertTrue(true);
```
sgiehl, line 28: might be better to use `$this->expectNotToPerformAssertions();` instead of doing an assert that can't fail.
Also you could also use one test for all strings with a dataprovider, like for the failing ones.

flamisz:
I close this now as we don't need it.

PR #17436, opened 2021-04-08
Escape URL in geolocation admin error message

sgiehl:
### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/GeoIp2/GeoIP2AutoUpdater.php
+++ b/plugins/GeoIp2/GeoIP2AutoUpdater.php
@@ -464,7 +464,7 @@ class GeoIP2AutoUpdater extends Task
         $host = $parsedUrl['host'] ?? '';
 
         if (empty($schema) || empty($host) || !in_array(Common::mb_strtolower($schema), ['http', 'https'])) {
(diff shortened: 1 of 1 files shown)


plugins/GeoIp2/GeoIP2AutoUpdater.php
```diff
@@ -464,7 +464,7 @@ protected static function checkGeoIPUpdateUrl($url)
         $host = $parsedUrl['host'] ?? '';
 
         if (empty($schema) || empty($host) || !in_array(Common::mb_strtolower($schema), ['http', 'https'])) {
-            throw new Exception(Piwik::translate('GeoIp2_MalFormedUpdateUrl', '<i>'.$url.'</i>'));
```
diosmosis, line 467: Looks good, but I'm wondering if we should use `Common::sanitizeString()` for consistency?

diosmosis (changes requested):
left a comment

plugins/GeoIp2/GeoIP2AutoUpdater.php
sgiehl: Makes sense. changed it to `Common::sanitizeInputValue()` as the other one has private access only

Issue #17428, opened 2021-04-07
Regression: some daily reports disappear from UI, and weekly reports are corrupted


nelhefni:
Hello,

We think PR https://github.com/matomo-org/matomo/pull/17216 has introduced some regressions to the archiving process since Matomo 4.2.0

FYI, browser archiving is disabled:
```
enable_browser_archiving_triggering = 0
browser_archiving_disabled_enforce = 1
archiving_range_force_on_browser_request = 0
```

For a given plugin, we are facing 2 issues since the update to Matomo 4.2.0, 
- Issue#1: Some daily reports disappear from the UI, even though `delete_reports_enable` is set to 0
- Issue#2: Some weekly reports have inaccurate data: gather the values of only some days of the week.

This seems to be related to the changes made with the handling of partial archives.
Indeed, we have a lot of partial archives in our DB, which doesn't contain the reports of our impacted plugins:

```
MariaDB [matomo]> select * from matomo_archive_numeric_2021_03 where idarchive = 46176;

| idarchive | name        | idsite   | date1      | date2      | period    | ts_archived         | value    |

|     46176 | myPlugin    |        1 | 2021-03-22 | 2021-03-22 |         1 | 2021-03-23 00:05:52 |       10 |
|     46176 | otherPlugin |        1 | 2021-03-22 | 2021-03-22 |         1 | 2021-03-23 00:05:52 |       20 |
|     46176 | done        |        1 | 2021-03-22 | 2021-03-22 |         1 | 2021-03-23 00:05:52 |        1 |
```

```
MariaDB [matomo]> select * from matomo_archive_numeric_2021_03 where idarchive = 56405;

| idarchive | name        | idsite   | date1      | date2      | period    | ts_archived         | value    |

|     56405 | otherPlugin |        1 | 2021-03-22 | 2021-03-22 |         1 | 2021-04-05 18:52:03 |       20 |
|     56405 | done        |        1 | 2021-03-22 | 2021-03-22 |         1 | 2021-04-05 18:52:03 |        5 |
```

As you can see, `myPlugin` report is not part of `partial archive` 56405 (`partial` because the value of `done` row is 5), but it is nevertheless selected since Matomo 4.2.0, as `ts_archived` is more recent.

Since https://github.com/matomo-org/matomo/pull/17216, `checkAuthorizedToArchive = false` is passed to `Rules::getSelectableDoneFlagValues` during the retrival of archive to display in the UI
[...]

PR #17437, opened 2021-04-09
Do not allow setting partial archives to DONE_INVALIDATED

diosmosis:
### Description:

Since DONE_INVALIDATED archives are treated the same as DONE_OK ones, we don't want partial archives to be set this way, as it changes the content.

I thought this was already in the code, but apparently it's not.

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
Refs #17428

diosmosis:
FYI @tsteur

sgiehl (approved):
Code looks good. Even though I might not be able to overlook to full impact of this change

diosmosis:
FYI @tsteur, merging this one

PR #17439, opened 2021-04-09
Fixes for specific case when partial archives have to initiate archiving for child archives

diosmosis:
### Description:

Refs #17428

Includes new test case and changes to make it pass. See inline comments for individual changes/fixes.

### Review

* [ ] Functional review done
* [ ] Code review done


core/ArchiveProcessor/Loader.php
```diff
@@ -105,6 +105,7 @@ public function prepareArchive($pluginName)
     private function prepareArchiveImpl($pluginName)
     {
         $this->params->setRequestedPlugin($pluginName);
+        $this->params->setIsPartialArchive(false);
```
diosmosis: Normally when configuring Params objects, we create new ones where partial archive is set to `false` to begin with. Here in Loader, however, we reuse Params objects to archive VisitsSummary by itself, and then the rest of the plugins. When reusing, we have to specifically reset this property, or other archives will think they are partial archives.

This is the only place this should be done, so I did not consider changing Params a good idea, since developers should only use the `setIsPartialArchive` method in Archiver instances for a specific use case.

core/ArchiveProcessor/Loader.php
```diff
@@ -117,8 +118,7 @@ private function prepareArchiveImpl($pluginName)
         // with a ts_archived >= the DONE_OK/DONE_INVALIDATED date.
         list($idArchives, $visits, $visitsConverted, $isAnyArchiveExists) = $this->loadExistingArchiveIdFromDb();
         if (!empty($idArchives)
-            && !$this->params->getArchiveOnlyReport()
```
diosmosis, line 130: Removed this since it seemed safer to reuse existing archives if we're not forcing single plugin archiving, and not depend on this state which comes from a query param.

core/ArchiveProcessor/Loader.php
```diff
@@ -184,13 +184,18 @@ protected function prepareCoreMetricsArchive($visits, $visitsConverted)
...
 
-            $pluginsArchiver = new PluginsArchiver($this->params);
-            $metrics = $pluginsArchiver->callAggregateCoreMetrics();
-            $pluginsArchiver->finalizeArchive();
+            $metrics = Context::executeWithQueryParameters(['requestedReport' => ''], function () {
+                $pluginsArchiver = new PluginsArchiver($this->params);
+                $metrics = $pluginsArchiver->callAggregateCoreMetrics();
+                $pluginsArchiver->finalizeArchive();
+                return $metrics;
+            });
```
diosmosis, line 197: Unset `requestedReport` so it's not picked up if we recurse here. If we're getting core metrics for a week, and there isn't a day archive already existing, we'll try to create the archive. But w/ requestedReport set, we may try to archive the other plugin as well as VisitsSummary. (Note: this can be seen in the new test if the code is removed.)

core/ArchiveProcessor/Rules.php
```diff
@@ -328,17 +328,20 @@ public static function getSelectableDoneFlagValues($includeInvalidated = true, P
         return $possibleValues;
     }
 
-    public static function isForceArchivingSinglePlugin()
+    public static function isRequestingToAndAbleToForceArchiveSinglePlugin()
```
diosmosis, line 337: Renamed to show it's true function more clearly.

core/ArchiveProcessor/Rules.php
```diff
@@ -328,17 +328,20 @@ public static function getSelectableDoneFlagValues($includeInvalidated = true, P
...
             return false;
         }
 
         return !empty($_GET['pluginOnly']) || !empty($_POST['pluginOnly']);
     }
 
+    public static function isActuallyForceArchivingSinglePlugin()
+    {
+        return Loader::getArchivingDepth() <= 1 && self::isRequestingToAndAbleToForceArchiveSinglePlugin();
+    }
```
diosmosis, line 349: We now have two methods to determine if we are forcing archiving of a single plugin:

- isRequestingToAndAbleToForceArchiveSinglePlugin() which returns true if the original request was for a pluginOnly archive, and if it came from archive.php (so we're able to in that case).
- and isActuallyForceArchivingSinglePlugin() which returns true if in this specific archiving request, we want to archive a pluginOnly archive. We only do this for the root request, for others we try to reuse existing archives. If no existing archives exist, archiving will launch for them so we don't need to do anything (unless, like in the Cloud plugin, we force disable it).

core/DataAccess/ArchiveSelector.php
```diff
@@ -205,10 +205,14 @@ public static function getArchiveIds($siteIds, $periods, $segment, $plugins, $in
...
-                    list($idarchive, $doneFlag) = explode('|', $pair);
+                    list($idarchive, $doneFlag, $value) = explode('|', $pair);
 
                     $result[$doneFlag][$dateStr][] = $idarchive;
-                    if (strpos($doneFlag, '.') === false) { // all plugins archive
+                    if (strpos($doneFlag, '.') === false // all plugins archive
+                        // sanity check: DONE_PARTIAL shouldn't be used w/ done archives, but in case we see one,
+                        // don't treat it like an all plugins archive
+                        && $value != ArchiveWriter::DONE_PARTIAL
+                    ) {
```
diosmosis, line 218: This and the above code is a safety check for if an archive has a `done` like flag but `DONE_PARTIAL` value. In case of problematic data like this, we just treat it like another partial archive.

core/DataAccess/ArchiveWriter.php
```diff
@@ -333,4 +337,17 @@ private function isRecordNumeric($value)
...
+        // if the done flag is not like done%.PluginName, then it shouldn't be a partial archive.
+        // log a warning.
+        if ($doneValue == self::DONE_PARTIAL && strpos($this->doneFlag, '.') == false) {
+            $ex = new \Exception(sprintf("Trying to create a partial archive w/ an all plugins done flag (done flag = %s). This should not happen.",
+                $this->doneFlag));
+            StaticContainer::get(LoggerInterface::class)->warning('{exception}', [
+                'exception' => $ex,
+            ]);
+        }
+    }
```
diosmosis, line 352: Adding a log just in case there is some other way to create a partial archive w/o a plugin done flag.

diosmosis:
FYI @tsteur

tsteur:
@diosmosis can you have a look at the failing test they might be related to this? If they fail because of this PR I will have another look after fixing it

diosmosis:
@tsteur relevant tests should pass now

core/ArchiveProcessor/Loader.php
tsteur: @diosmosis in line 113 should it set `$this->params->setArchiveOnlyReport($requestedReport)` or maybe `setArchiveOnlyReport()` should set `setIsPartialArchive` automatically?

tsteur: do we ever need to reset the original/previously set `IsPartialArchive` value maybe? haven't checked yet where it's reused

core/ArchiveProcessor/Loader.php
```diff
@@ -184,13 +184,18 @@ protected function prepareCoreMetricsArchive($visits, $visitsConverted)
 
             $this->params->setRequestedPlugin('VisitsSummary');
             $this->params->setArchiveOnlyReport(null);
+            $this->params->setIsPartialArchive(false);
```
tsteur, line 190: would we need to back up the previously set value and restore it below in line 198?

tsteur (commented):
haven't tested anything but left few comments just to check. Looked through the code and wasn't sure if it could maybe result in some wrong archiving like when we mark something as not partial but we actually only want to archive a plugin.

diosmosis:
> in line 113 should it set $this->params->setArchiveOnlyReport($requestedReport) or maybe setArchiveOnlyReport() should set setIsPartialArchive automatically?

I don't think it's technically necessary to set it, but it makes sense to be thorough. `setArchiveOnlyReport()` shouldn't call `setIsPartialArchive()`, that's called specifically by Archiver's when they handle the requestedReport parameter, letting ArchiveWriter know it only archived a single report.

I thought about unsetting it in that method, but we never re-use Parameters objects anywhere else (and shouldn't in normal uses), so it seemed like it could cause strange side effects.

> do we ever need to reset the original/previously set IsPartialArchive value maybe? haven't checked yet where it's reused

In that specific line of code, no, because that's the start of the prepareArchive() method. In fact in that line of code we don't need to call the method there. I'll remove that.

> would we need to back up the previously set value and restore it below in line 198?

Don't need to since we archive VisitsSummary before other plugins. We could do it, though, just in case something changes in the future.

tsteur:
> Don't need to since we archive VisitsSummary before other plugins. We could do it, though, just in case something changes in the future.

I reckon that might be useful just in case

diosmosis:
@tsteur applied pr feedback

tsteur:
fyi @diosmosis there are now some merge conflicts from merging the other PR

diosmosis:
@tsteur :+1: just fixed, we'll see if the build passes

tsteur:
btw is this one now no longer needed https://github.com/matomo-org/matomo/pull/17439/commits/706621b5476debe229f292ae511fa3cf1b4bf33d#diff-3ad7e38a2ff2d49f9ac9f268bf0ce70db9b1a924be38e5802816acc0ae343d32L108 ?

diosmosis:
@tsteur it was never really needed, since Loader is given a new Parameters instance just before using it. So it's already in a clean state.

PR #17440, opened 2021-04-09
speed up integration tests by not using DbHelper::getTablesInstalled()

diosmosis:
### Description:

Now in DbHelper::getTablesInstalled() we post an event and force load all activated plugins: https://github.com/matomo-org/matomo/blob/4.x-dev/core/Db/Schema/Mysql.php#L453-L458

This is quite a bit more work, it seems. Locally it slows each individual integration test case by about 14s. Removing it speeds the LoaderTest, for example, from 10mins to finish to 1min.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/tests/PHPUnit/Framework/TestCase/SystemTestCase.php
+++ b/tests/PHPUnit/Framework/TestCase/SystemTestCase.php
@@ -726,9 +726,13 @@ abstract class SystemTestCase extends TestCase
         DbHelper::truncateAllTables();
 
         // insert data
+        $archiveTables = Db::fetchAll("SHOW TABLES LIKE '%archive_%'");
+        if (!empty($archiveTables)) {
+            $archiveTables = array_column($archiveTables, key($archiveTables[0]));
+        }
         foreach ($tables as $table => $rows) {
             // create table if it's an archive table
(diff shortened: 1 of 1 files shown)


tests/PHPUnit/Framework/TestCase/SystemTestCase.php
```diff
@@ -726,10 +726,9 @@ protected static function restoreDbTables($tables)
         DbHelper::truncateAllTables();
 
         // insert data
-        $existingTables = DbHelper::getTablesInstalled();
         foreach ($tables as $table => $rows) {
             // create table if it's an archive table
-            if (strpos($table, 'archive_') !== false && !in_array($table, $existingTables)) {
+            if (strpos($table, 'archive_') !== false && !DbHelper::tableExists($table)) {
```
sgiehl: might it be more performant to manually do a `SHOW TABLES like 'archive_'` to fetch all existing archive tables at once, instead of firing the query for each table?

diosmosis:
@sgiehl updated.

tests/PHPUnit/Framework/TestCase/SystemTestCase.php
diosmosis: :+1:

sgiehl (approved):

PR #17445, opened 2021-04-12
Ensure correct no access message is shown

sgiehl:
### Description:

While testing I recognized, that when an access exception is thrown Matomo currently always shows a "session expired" error, even though the user might still be logged in, but simply does not have access for the current site.

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
Tested locally, works for me. Is it possible to get this tested?

sgiehl:
@diosmosis added some UI tests for it

diosmosis (approved):
I was thinking we could get away w/ an integration test, but UI works too.

PR #17457, opened 2021-04-14
Workaround error in Overlay when site has no URLs

diosmosis:
### Description:

When creating a site it is possible to not specify URLs. In this case, we send a plugin setting entry in the HTTP request as `['name' => 'urls']` w/ no `value` key. Currently this is ignored by Matomo, and the validation that happens does not occur. Which allows sites to have no URLs.

If this happens and Overlay is loaded, the JS will error when trying to find the host of the site URLs.

This is fixed in this PR w/ tests. The issue in Overlay is also worked around in case there are users who created sites w/ no URLs.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/SitesManager/API.php
+++ b/plugins/SitesManager/API.php
@@ -598,7 +598,7 @@ class API extends \Piwik\Plugin\API
         }
 
         $coreProperties = array();
-        $coreProperties = $this->setSettingValue('urls', $urls, $coreProperties, $settingValues, $isRequired = true);
+        $coreProperties = $this->setSettingValue('urls', $urls, $coreProperties, $settingValues);
         $coreProperties = $this->setSettingValue('ecommerce', $ecommerce, $coreProperties, $settingValues);
         $coreProperties = $this->setSettingValue('group', $group, $coreProperties, $settingValues);
         $coreProperties = $this->setSettingValue('sitesearch', $siteSearch, $coreProperties, $settingValues);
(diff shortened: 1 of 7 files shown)


plugins/SitesManager/API.php
```diff
@@ -598,7 +598,7 @@ public function addSite($siteName,
         }
 
         $coreProperties = array();
-        $coreProperties = $this->setSettingValue('urls', $urls, $coreProperties, $settingValues);
+        $coreProperties = $this->setSettingValue('urls', $urls, $coreProperties, $settingValues, $isRequired = true);
```
sgiehl: This makes it impossible to create any new roll ups, as for roll ups you can't define any urls.

sgiehl (changes requested):

plugins/SitesManager/API.php
diosmosis: Good point. I guess I can make it required if type == 'website'. (cc @tsteur)

tsteur:
@diosmosis this would be breaking API (even though in the code it should have already worked but from a user perspective things are potentially still breaking as it probably wasn't working for a while) and not sure if want to require URLs now or maybe wait to Matomo 5. Is there an issue for this PR?

diosmosis:
@tsteur this is to fix the root cause of L3-65

diosmosis:
@tsteur updated. Warning looks like:

image

tsteur:
👍  be great to have an issue to fix the addSite/updateSite API regarding requiring URLs as part of Matomo 5

diosmosis:
Created https://github.com/matomo-org/matomo/issues/17459

PR #17461, opened 2021-04-15
Add get-segment-sql development command for debugging

diosmosis:
### Description:

Prints out the query of a segment, can be used if trying to determine if a segment is query is being built correctly or not.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/SitesManager/Model.php
+++ b/plugins/SitesManager/Model.php
@@ -240,7 +240,7 @@ class Model
      * Returns the list of all the website IDs registered.
      * Caller must check access.
      *
-     * @return int[] The list of website IDs
+     * @return int[]|string[] The list of website IDs
      */
     public function getSitesId()
     {
--- a/plugins/SitesManager/Model.php
+++ b/plugins/SitesManager/Model.php
@@ -240,7 +240,7 @@ class Model
      * Returns the list of all the website IDs registered.
      * Caller must check access.
      *
-     * @return array The list of website IDs
+     * @return int[] The list of website IDs
      */
     public function getSitesId()
     {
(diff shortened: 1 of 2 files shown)


flamisz (commented):
I'm not 100% sure what is this for, but tested and got results that look ok to me 👍

flamisz (approved):

plugins/SitesManager/Model.php
```diff
@@ -240,7 +240,7 @@ public function getSiteFromId($idSite)
      * Returns the list of all the website IDs registered.
      * Caller must check access.
      *
-     * @return array The list of website IDs
+     * @return int[] The list of website IDs
```
sgiehl: That actually might not be correct. Depending on the mysql configuration it might be an array of strings. But guess we could add a cast to int in the foreach to ensure only ints are returned

diosmosis: I could do `int[]|string[]`. As long as it's not `array`, it'll be more useful.

sgiehl: Sure. no preference. Just wanted to mention that the return type can vary. Maybe casting to int would be even better, so we can rely on the return type?

diosmosis: Might be more useful, but given we don't do that for any other query in Matomo (that I know of), I'm not sure it's really needed (and might make things more inconsistent).

diosmosis:
@flamisz segments are converted into SQL to select visits in Matomo, but for many segments the SQL is complicated and hard to imagine in your head (or mine at least), so the command prints it out so you can see how it's converted from a segment string (ie, `pageUrl==whatever;browserCode=ff`) to SQL for a specific table. You can also put a print statement or log in Matomo, which is what I used to do, but I found it annoying to do that.

plugins/SitesManager/Model.php
sgiehl: Feel free to change the doc block only then. Having more strict return types is a more global thing for sure 🙈

Issue #9199, opened 2015-11-11
Bounce count / rate might not be accurate


tsteur:
Just seeing bounce count is calculated by `sum(case " . self::LOG_VISIT_TABLE . ".visit_total_actions when 1 then 1 when 0 then 1 else 0 end`

but it looks like `visit_total_actions` is also increased for events, searches etc.

This means when there is a pageview followed by an event, it is not counted as bounced but it should be I think

hpvd:
hmm why counting as bounced? If there's an event, the visitor does not disappear suddenly, do he?
Maybe you can give an example for the kind of event you have in mind?

tsteur:
Say you use JavaScript Error Tracking feature in Piwik which triggers an event when there is a JS error. It would never count any bounce. Or maybe there's a video player on a page that starts automatically and tracks an event etc. 

Some users might not want to have a bounce counted after a certain event as there was some kind of interaction but at list Piwik describes it as follows:

> The percentage of visits that only had a single pageview. This means, that the visitor left the website directly from the entrance page.

With the current implementation this is not the case

hpvd:
hmm, thanks for exampels and pointing to piwiks definition of a bounce.
We make the observation, that more and more visitors visit only one page.
Maybe that not always bad. E.g. if they find exactly what they were looking for that's pretty fine.
They could stay for 5minutes on the site: reading, open a image light box, watch a video etc.
That's not bad. They could also have seen the advertisements on the site, in Image gallery at the beginning of the video... They could perfectly keep in mind the brand of the website.
so is this a bounce?
What should they do more if google brings them directly  exactly to the place they were looking for?

tsteur:
A bounce doesn't have to be something bad. It can be good or bad. It always depends on the site and probably even on the page whether a bounce is good or bad. I'm not saying something is good or bad. Only saying that the current bounce calculation doesn't match the documentation in the UI and what users might expect.

hpvd:
sure, i get this :-)
Were only examples...

hpvd:
For many many people bounces are bad. Otherwise they would not count them...

mattab:
For this we need a new feature, the ability to define a given Event tracking request as "Non-Interactive". By default, events are "interactive" and assume that user did an action (clicked mouse, or typed on keyboard). When an event is marked as "non Interactive" then the event would not affect the bounce rate, nor would it affect "Time on website" or "Time on page". 

[...]

PR #17466, opened 2021-04-16
Allow goals being triggered when a visitor visits a certain number of pages

sgiehl:
### Description:

fixes #2029

### Review

* [ ] Functional review done
* [ ] Code review done


core/Tracker/GoalManager.php
```diff
@@ -220,6 +221,22 @@ private function detectNumericGoalMatch($goal, Action $action, VisitProperties $
...
+                if ($action instanceof ActionPageview) {
+                    $totalInteractions++; // current action won't be included in interaction count
+                }
+
+                if (empty($totalInteractions)) {
+                    return null;
+                }
+
+                // total interactions includes searches, so get number of searches and remove them
+                $totalSearches = (int) $visitProperties->getProperty('visit_total_searches');
```
sgiehl: Couldn't find a smarter way to get the number of page views without performing extra queries. Maybe someone else has a better idea?

diosmosis: Can you point out the extra query? I can't find it in the changes. (Note: VisitProperties->getProperty() just gets from an array, it doesn't perform a query AFAIK.)

sgiehl: That may have been a bit ambiguous. The current solution has no additional queries. However, I have not found any other possibility that would work without additional queries.

diosmosis: That makes this solution sound like the best one... unless there's a downside to it? I haven't really thought about it but I don't see anything obviously problematic.

diosmosis: @sgiehl can you answer ^? Is there anything that you know about this approach that isn't ideal? Just so I know why you think it's maybe not the best solution, even though it has no additional queries.

sgiehl: If we ever decide to count more actions into `visit_total_interactions`, or maybe a plugin manipulates that value to add custom interactions, the calculation might not longer be correct anymore.

diosmosis: I see, in that case it sounds like the best solution is a new dimension like `visit_total_pageviews` (if something like that doesn't exist). Since we can't add that until Matomo 5, this seems like the next best solution. We could create an issue to move to a new dimension, or just wait until 5.x to implement this.

cc @tsteur

tsteur: @sgiehl @diosmosis seems the only way to do it for now. Can we leave a comment in `shouldCountInteraction` in visit total interactions that adjusting the logic would break this feature? For example this will happen when we add features like https://github.com/matomo-org/matomo/issues/9199 where we can say a pageview or an event was interactive or non-interactive. This may for example cause some page views to not be counted as interaction or it may cause an event to be counted as interactive.

Personally, I would rather not work on this feature for now as it may cause more problems than it does good and not sure why the issue is marked as major as I don't think it be used that often. @mattab any thoughts if the feature could wait until Matomo 5?

mattab: Sounds good to me to wait until Matomo 5 and a new dimension/column can be added :+1:

sgiehl: I've updated the PR so it uses a new dimension and moved it to 5.0 milestone...

PR #17470, opened 2021-04-18
Avoid double encoding page titles

### LINKED ISSUE #14065: Avoid double HTML entities / special characters encoding
Hi, if the page's `<title>` (in `<head>` section) contains for e.g. already encoded **&** character:

`&#38;`

or:

`&amp;`

-----------------------------------------------

M will encode it again, resulting in:

`&amp;amp;`

being displayed in the tooltip (but tooltip just copies values from `` element)

-----------------------------------------------

which, again, will look like this in HTML:

`&amp;`

which is wrong.

-----------------------------------------------

It should simply be/look like this:

`&`

because it is already encoded HTML entity.

See screenshots for illustration.

matomo-html-char-encoding-bug-1
matomo-html-char-encoding-bug-2

flamisz:
### Description:

fixes: #14065 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Live/templates/getLastVisitsStart.twig
+++ b/plugins/Live/templates/getLastVisitsStart.twig
@@ -80,7 +80,7 @@
 {% set title %}
 {% if action.url|trim is not empty %}<span class='tooltip-action-url'>{{ action.url }}</span>{% endif %}
 
-{% if action.pageTitle is not empty %}<span class='tooltip-action-page-title'>{{ action.pageTitle }}</span>{% endif %}
+{% if action.pageTitle is not empty %}<span class='tooltip-action-page-title'>{{ action.pageTitle|raw }}</span>{% endif %}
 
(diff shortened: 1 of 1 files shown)


flamisz:
The problem what I found was that we use `htmlspecialchars` on the title before we save it to the db, and we double encode it in the twig file. We put into the DOM at line 83 (but this is hidden on the page), and than we use this already encoded version at line 88.

https://github.com/matomo-org/matomo/blob/84b9f9c33ce6402556008c8764a79747f24b5b0f/plugins/Live/templates/getLastVisitsStart.twig#L83-L88

diosmosis:
LGTM and works locally, merging

tsteur:
`|raw` would that be causing any security issue? Like when title maybe contains `{{` and `}}`? Would we need to use something like `|rawSafeDecoded`? Or if we ever tracked title not encoded in the past could this cause an issue?

PR #17475, opened 2021-04-20
Remove urlencode for segment definitions

### LINKED ISSUE #17301: Store the segment hash in the DB table

flamisz:
### Description:

fixes #17301 

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (commented):
Code looks good and tests are passing. But as I'm not that deep into the latest segment changes might be good if @diosmosis or @tsteur might have a quick look as well.

core/DataAccess/Model.php
```diff
@@ -619,7 +619,7 @@ public function getArchiveIdsForSegments($archiveTableName, array $segments, $ol
     private function getDeletedSegmentWhereClause(array $segment)
     {
         $idSite = (int)$segment['enable_only_idsite'];
-        $segmentHash = Segment::getSegmentHash(urlencode($segment['definition']));
+        $segmentHash = Segment::getSegmentHash($segment['definition']);
```
diosmosis: Here, maybe we should use the hash in the DB rather than computing from the definition?

tsteur: @diosmosis do you know why that would be needed here? The logic should maybe still work here too?

diosmosis: The definition is still encoded differently than the `segment` parameter? Or has how we need to encode them changed? Either way, wouldn't it be safer to use the hash in the DB (which should just be `$segment['hash']`)?

tsteur: Good point, here could probably simply use `$segment['hash']` directly 👍  the method should have otherwise tried the different decode,encode etc find the right hash too (I think)

flamisz: make sense, I'll do it

diosmosis:
There's an integration test failing now. Will merge once build passes.

flamisz:
@diosmosis tests are back to green.

PR #17487, opened 2021-04-22
Use empty nonce if not md5

flamisz:
### Description:

If the nonce parameter is valid, ignore it.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Nonce.php
+++ b/core/Nonce.php
@@ -99,22 +99,6 @@ class Nonce
         return true;
     }
 
-
-    /**
-     * Returns if a nonce is valid md5.
-     *
-     * @param string $md5
-     * @return bool
-     */
-    public static function isValidMd5($md5)
-    {
-        if (preg_match('/^[a-f0-9]{32}$/', $md5)) {
-            return true;
-        }
-
-        return false;
-    }
-
     // public for tests
     public static function isReferrerHostValid($referrer, $expectedReferrerHost)
     {
--- a/core/Nonce.php
+++ b/core/Nonce.php
@@ -99,6 +99,22 @@ class Nonce
         return true;
     }
 
+
+    /**
+     * Returns if a nonce is valid md5.
+     *
+     * @param string $md5
+     * @return bool
+     */
+    public static function isValidMd5($md5)
+    {
+        if (preg_match('/^[a-f0-9]{32}$/', $md5)) {
+            return true;
+        }
+
+        return false;
+    }
+
     // public for tests
     public static function isReferrerHostValid($referrer, $expectedReferrerHost)
     {
--- a/plugins/CoreAdminHome/OptOutManager.php
+++ b/plugins/CoreAdminHome/OptOutManager.php
@@ -172,10 +172,13 @@ class OptOutManager
 
         $setCookieInNewWindow = Common::getRequestVar('setCookieInNewWindow', false, 'int');
         if ($setCookieInNewWindow) {
+            $requestNonce = Common::getRequestVar('nonce', false);
+            $nonce = Nonce::isValidMd5($requestNonce) ? $requestNonce : '';
+
             $reloadUrl = Url::getCurrentQueryStringWithParametersModified(array(
                 'showConfirmOnly' => 1,
                 'setCookieInNewWindow' => 0,
-                'nonce' => Common::getRequestVar('nonce')
+                'nonce' => $nonce
             ));
         } else {
             $reloadUrl = false;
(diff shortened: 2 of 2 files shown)


core/Nonce.php
```diff
@@ -99,6 +99,22 @@ public static function verifyNonce($id, $cnonce, $expectedReferrerHost = null)
...
+
+    /**
+     * Returns if a nonce is valid md5.
+     *
+     * @param string $md5
+     * @return bool
+     */
+    public static function isValidMd5($md5)
+    {
+        if (preg_match('/^[a-f0-9]{32}$/', $md5)) {
```
diosmosis: Should we also check for capital `A-F`?

plugins/CoreAdminHome/OptOutManager.php
```diff
@@ -172,10 +172,13 @@ public function getOptOutView()
 
         $setCookieInNewWindow = Common::getRequestVar('setCookieInNewWindow', false, 'int');
         if ($setCookieInNewWindow) {
+            $requestNonce = Common::getRequestVar('nonce', false);
+            $nonce = Nonce::isValidMd5($requestNonce) ? $requestNonce : '';
```
tsteur: can we maybe just call `verifyNonce`?

core/Nonce.php
```diff
@@ -99,6 +99,22 @@ public static function verifyNonce($id, $cnonce, $expectedReferrerHost = null)
...
     }
 
+
+    /**
+     * Returns if a nonce is valid md5.
+     *
+     * @param string $md5
+     * @return bool
+     */
+    public static function isValidMd5($md5)
```
tsteur: btw if this method is in the Nonce class then a better name would be `isValidNonceFormat` or so. So we don't need to rename the method when it's no longer an md5 etc.

plugins/CoreAdminHome/OptOutManager.php
flamisz: `verifyNonce` checks more thinks and that would modify the business logic here. I'm not familiar as deeply with it, so I went with the safe solution. 
I haven't checked yet if we have tests or not for this part, but I can look into that and see if that passes that way or not.

core/Nonce.php
flamisz: You are right, capital letters are valid hex codes too. I can make the string lowercase before the `preg_match` check.

plugins/CoreAdminHome/OptOutManager.php
tsteur: I would say it should work to use verifyNonce so it only works when the correct nonce is used etc. Might be worth a test since then it also checks directly that actually the correct nonce is used (the verify method won't discard it which is why it likely should work).

core/Nonce.php
tsteur: probably ctype_xdigit would work too

plugins/CoreAdminHome/OptOutManager.php
diosmosis: Just thinking, if verifyNonce is used, would the optout form be a way to be able to guess the nonce? Since it's not removed after checking. And if so, and an attacker gets the nonce, can they do something w/ it?

tsteur: @diosmosis we can simply discard the nonce when it's not correct.

flamisz: I updated the pr using `verifyNonce` and discard it if it's not valid.

core/Nonce.php
flamisz: I don't use it anymore, so deleted

plugins/CoreAdminHome/OptOutManager.php
```diff
@@ -172,10 +172,18 @@ public function getOptOutView()
 
         $setCookieInNewWindow = Common::getRequestVar('setCookieInNewWindow', false, 'int');
         if ($setCookieInNewWindow) {
+            $requestNonce = Common::getRequestVar('nonce', false);
+            $nonce = $requestNonce !== false ? $requestNonce : '';
```
tsteur: btw is $requestNonce actually needed here? or could in line 186 have something like `'nonce' => $nonce ? $nonce : ''`? Tried to understand and was bit complicated.

flamisz: I'll check it, but probably it's not needed like this anymore. I'll update the pr.

flamisz: updated

flamisz:
Is there anything else to do here @tsteur?

tsteur:
I haven't tested it but looks good to me @flamisz . We'd only merge it after the 4.3 release though just to make sure we're not breaking anything

diosmosis (approved):
Tested locally, works. Code looks good too.

PR #17490, opened 2021-04-23
Add diagnostic that checks if server directories that should be private are accessible

diosmosis:
_###_ Description:

Adds a new diagnostic that checks that the tmp/ directory is not accessible, and checks that, if present, the .git/ directory is not accessible.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Diagnostics/Diagnostic/RequiredPrivateDirectories.php
+++ b/plugins/Diagnostics/Diagnostic/RequiredPrivateDirectories.php
@@ -11,13 +11,11 @@ namespace Piwik\Plugins\Diagnostics\Diagnostic;
 use Piwik\Common;
 use Piwik\Filesystem;
 use Piwik\Http;
-use Piwik\Piwik;
-use Piwik\Plugins\HeatmapSessionRecording\HeatmapSessionRecording;
 use Piwik\SettingsPiwik;
 use Piwik\Translation\Translator;
 
 /**
- * TODO: describe
+ * Checks whether certain directories in Matomo that should be private are accessible through the internet.
  */
 class RequiredPrivateDirectories implements Diagnostic
 {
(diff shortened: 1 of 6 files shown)


plugins/Diagnostics/Diagnostic/RequiredPrivateDirectories.php
```diff
@@ -0,0 +1,93 @@
...
+        $label = $this->translator->translate('Diagnostics_RequiredPrivateDirectories');
+
+        $privatePaths = [
+            ['tmp/', 'tmp/empty'], // created by this diagnostic
+            ['.git/', '.git/config'],
+        ];
+
+        // create test file to check if tmp/empty exists
+        Filesystem::mkdir(PIWIK_INCLUDE_PATH . '/tmp');
+        file_put_contents(PIWIK_INCLUDE_PATH . '/tmp/', 'test');
```
sgiehl: should this maybe be

```suggestion
        file_put_contents(PIWIK_INCLUDE_PATH . '/tmp/empty', 'test');
```

plugins/Diagnostics/Diagnostic/RequiredPrivateDirectories.php
```diff
@@ -0,0 +1,93 @@
...
+                    continue;
+                }
+
+                $testUrl = $baseUrl . $path;
+
+                if (!$isInternetEnabled) {
+                    $unknown = $this->translator->translate('Diagnostics_PrivateDirectoryInternetDisabled', $testUrl) . ' ' . $manualCheck;
+                    $results[] = DiagnosticResult::singleResult($label, DiagnosticResult::STATUS_WARNING, $unknown);
+                    continue;
+                }
```
sgiehl: Does it make sense to show all checks separately in this case? Maybe we could only show one warning and list all directories/files in it instead.

sgiehl (changes requested):

Findus23:
For reference: I also created a similar check in my plugin that also checks for `lang/en.json` (less severe, but indicates that the htaccess doesn't work), `cache/tracker/matomocache_general.php` and `config/config.ini.php` (also showing a warning if it only outputs `;`)

https://github.com/Findus23/matomo-DiagnosticsExtended/blob/main/Diagnostic/URLCheck.php

plugins/Diagnostics/Diagnostic/RequiredPrivateDirectories.php
diosmosis: This makes sense, since the setting will always be disabled in the loop

diosmosis:
@sgiehl updated the PR.

Also added the directories @Findus23 mentioned :+1: .

sgiehl (approved):
Seems to work as expected now, even though the check would be more accurate if the requests would be performed in the browser and not server side.

diosmosis:
Unfortunately I don't think there's an easy way to implement it so an actual browser separate from the server makes those requests. At least none that I can think of.

PR #17492, opened 2021-04-23
Improve method to determine date to archive segments from

sgiehl:
### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
:+1:

tsteur (commented):
doesn't this change behaviour should `segmentCreatedTime` be empty? We would have previously returned `null` but now we would treat it as `beginning_of_time` even if eg `from creation time` is configured? @sgiehl

PR #17503, opened 2021-04-26
detect brute force login attempts from different ip addresses

diosmosis:
### Description:

Detects brute force login attempts from different ip addresses, and notifies the user at most once every two weeks that someone is trying to break into their account this way.

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
This should be ready for review now

plugins/Login/Security/BruteForceDetection.php
```diff
@@ -144,4 +151,93 @@ private function getDateTimeSubMinutes($minutes)
     {
         return $this->getNow()->subPeriod($minutes, 'minute')->getDatetime();
     }
+
+    // TODO: need to make sure bruteforcedetection happens after any ip allowlists/blocklists have been used
+    public function isUserLoginBlocked($login)
```
tsteur, line 162: fyi there's still a todo

plugins/Login/Security/BruteForceDetection.php
```diff
@@ -144,4 +151,93 @@ private function getDateTimeSubMinutes($minutes)
...
+        }
+
+        if (!$this->hasNotifiedUserAboutSuspiciousLogins($login)) {
+            $this->sendSuspiciousLoginsEmailToUser($login, $count);
+        }
+
+        return true;
+    }
+
+    private function getTotalLoginAttemptsInLastHourForLogin($login)
```
tsteur: be good to move some of the private methods maybe in some other class and then have them public so they can be tested

plugins/Login/Security/BruteForceDetection.php
```diff
@@ -9,14 +9,21 @@
...
+use Piwik\Option;
+use Piwik\Plugins\Login\Emails\SuspiciousLoginAttemptsInLastHourEmail;
 use Piwik\Plugins\Login\SystemSettings;
 use Piwik\Updater;
 use Piwik\Version;
+use Psr\Log\LoggerInterface;
 
 class BruteForceDetection {
 
+    const OVERALL_LOGIN_LOCKOUT_THRESHOLD = 1000; // TODO: maybe make this come from DI config or INI config
```
tsteur: fyi another todo

tsteur: I think we could define this number as say 3 times the configured amount of login attempts by default. If usually 20 are allowed then we would allow 60 across IPs maybe? 
image

tsteur (commented):
@diosmosis had a quick look and looks good 👍  I didn't do a proper review though be good for @sgiehl or @flamisz to have a look too and test it works

plugins/Login/Security/BruteForceDetection.php
diosmosis: changed

plugins/Login/Security/BruteForceDetection.php
diosmosis: Moved to a new Model class. Couldn't think of a class that wouldn't require the calling method isUserLoginBlocked, so created a new Model class for Login. Integration tests were added too.

plugins/Login/Security/BruteForceDetection.php
diosmosis: :+1: dealt with

plugins/Login/Security/BruteForceDetection.php
```diff
@@ -144,4 +158,50 @@ private function getDateTimeSubMinutes($minutes)
...
+                    'exception' => $ex,
+                ]);
+            }
+        }
+    }
+
+    protected function getOverallLoginLockoutThreshold()
+    {
+        $settings = new SystemSettings();
+        $threshold = $settings->maxFailedLoginsPerMinutes->getValue() * 3;
```
flamisz, line 204: Why is this 3?

flamisz (changes requested):
It took some time to test it. The pr description was not enough for me as I didn't know how the brute force originally worked. But made it work in the end, but the sent email wasn't "templated" properly.

Screen Shot 2021-05-10 at 4 24 50 PM

plugins/Login/Security/BruteForceDetection.php
diosmosis: See https://github.com/matomo-org/matomo/pull/17503#discussion_r628997623

tsteur: it's just a randomly picked to allow a few more attempts across IPs before blocking it. It could be as well 2 or 5. Just thought 2 might be too low vs 5 is maybe too high. Could also set `2` though.

diosmosis:
> But made it work in the end, but the sent email wasn't "templated" properly.

Thanks for catching this. Fixed.

flamisz:
> > But made it work in the end, but the sent email wasn't "templated" properly.
> 
> Thanks for catching this. Fixed.

It looks good now 👍

flamisz (approved):

PR #17505, opened 2021-04-27
Fix determine segment rearchive date

sgiehl:
### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/CronArchive/SegmentArchiving.php
+++ b/core/CronArchive/SegmentArchiving.php
@@ -111,16 +111,25 @@ class SegmentArchiving
          */
         list($segmentCreatedTime, $segmentLastEditedTime) = $this->getCreatedTimeOfSegment($segmentInfo);
 
-        if ($this->processNewSegmentsFrom == SegmentArchiving::CREATION_TIME && !empty($segmentCreatedTime)) {
+        if ($this->processNewSegmentsFrom == SegmentArchiving::CREATION_TIME) {
+            if (empty($segmentCreatedTime)) {
+                return null;
+            }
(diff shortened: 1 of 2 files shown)


core/CronArchive/SegmentArchiving.php
```diff
@@ -111,16 +111,25 @@ public function getReArchiveSegmentStartDate($segmentInfo)
...
+                return null;
+            }
             $this->logger->debug("process_new_segments_from set to segment_last_edit_time, segment last edit time is {time}",
                 array('time' => $segmentLastEditedTime));
 
             return $segmentLastEditedTime;
-        } else if (preg_match("/^editLast([0-9]+)$/", $this->processNewSegmentsFrom, $matches) && !empty($segmentLastEditedTime)) {
+        } else if (preg_match("/^editLast([0-9]+)$/", $this->processNewSegmentsFrom, $matches)) {
+            if (empty($segmentLastEditedTime)) {
+                return null;
```
tsteur, line 131: was going to say here we could maybe fall back to `Date::now()` but I suppose it's not really needed anyway as we should always have an edited time.

tsteur (commented):
LGTM. Left one comment to look at optionally but not too important.

core/CronArchive/SegmentArchiving.php
sgiehl: An empty edit time shouldn't happen, indeed. Nevertheless assuming `Date::now()` when it wouldn't be set for any reason would only be correct if we assume this method is only called when the segment had been edited/created.

tsteur: 👍  all good, it's not needed. LGTM

PR #17508, opened 2021-04-29
Disable visitorId segment when feature disabled

### LINKED ISSUE #17488: Remove the "Visitor ID" segment when the "Visitor profile" feature has been disabled
In the context of CNIL submission of Matomo for exemption, it would be valuable if there was an easy way to disable the capability segment users based on the `Visitor ID` dimension.

We currently offer privacy feature to [disable "visitor profile" feature](https://matomo.org/faq/how-to/how-do-i-disable-the-visits-log-or-the-visitor-profile-feature/) in:

Screenshot from 2021-04-23 13-20-37

-> when either of these checkboxes is clicked (ie. the visitor profile is disabled), then we would also expect that it's not possible to [segment](https://matomo.org/docs/segmentation/) on the Visitor ID. But currently it is possible to create/use segments based on Visitor ID.

(What should happen to existing segments that use Visitor ID but now the dimension is not available? it would be useful if we could "hide" them from the segment list as long as the feature is disabled. And if the feature is enabled again then the segment become visible again?)

flamisz:
### Description:

fixes: #17488

### Review

* [ ] Functional review done
* [ ] Code review done


tsteur:
@flamisz could you maybe test below diff that might be an easier fix for it (dimensions usually create segments automatically)

```diff
diff --git a/plugins/CoreHome/Columns/VisitorId.php b/plugins/CoreHome/Columns/VisitorId.php
index 7c8fdec80e..fb4456c61e 100644
--- a/plugins/CoreHome/Columns/VisitorId.php
+++ b/plugins/CoreHome/Columns/VisitorId.php
@@ -9,11 +9,14 @@
 namespace Piwik\Plugins\CoreHome\Columns;
 
 use Piwik\Columns\DimensionMetricFactory;
+use Piwik\Columns\DimensionSegmentFactory;
 use Piwik\Columns\MetricsList;
 use Piwik\Metrics\Formatter;
 use Piwik\Piwik;
 use Piwik\Plugin\ArchivedMetric;
 use Piwik\Plugin\Dimension\VisitDimension;
+use Piwik\Plugins\Live\SystemSettings;
+use Piwik\Segment\SegmentsList;
 
 /**
  * Dimension for the log_visit.idvisitor column. This column is added in the CREATE TABLE
@@ -31,6 +34,17 @@ class VisitorId extends VisitDimension
     protected $sqlFilterValue = array('Piwik\Common', 'convertVisitorIdToBin');
     protected $type = self::TYPE_BINARY;
 
+    public function configureSegments(SegmentsList $segmentsList, DimensionSegmentFactory $dimensionSegmentFactory)
+    {
+        $systemSettings = new SystemSettings();
+        $visitorProfileEnabled = $systemSettings->disableVisitorProfile->getValue() === false
+            && $systemSettings->disableVisitorLog->getValue() === false;
+        if ($visitorProfileEnabled) {
+            parent::configureSegments($segmentsList, $dimensionSegmentFactory);
+        }
+
+    }
+
     public function configureMetrics(MetricsList $metricsList, DimensionMetricFactory $dimensionMetricFactory)
     {
         $metric = $dimensionMetricFactory->createMetric(ArchivedMetric::AGGREGATION_UNIQUE);
```

flamisz:
> @flamisz could you maybe test below diff that might be an easier fix for it (dimensions usually create segments automatically)

Hi @tsteur this solves a part of the issue, because this way `visitorId` is not included in the list segment list, but still need some other changes to hide the segments that include it, etc.

I refactor the code using this part as base, thanks for pointing this out.

flamisz:
@tsteur I have a question:
anyhow we solve it, we have to deal with cli commands. I'm not sure which commands will be involved, but the `core:archive` definitely will be. 
By default, if we not including `visitorId` in the segment list, the cli command will through an exception if there is any saved segments that has `visitorId` in the definition.
We still want to archive these segments, don't we? So for cli commands everything should work the same way whatever is that related live plugin setting?

tsteur:
@flamisz not sure with what you mean hide where it includes it? 

> By default, if we not including visitorId in the segment list, the cli command will through an exception if there is any saved segments that has visitorId in the definition.

That should be fine AFAIK as it's rarely used I would say and the LIVE setting is also new and rarely used yet so this combination should happen very rarely (or not at all) and we might want to rather go for the 80/20 solution. The alternative be to remove any such segments on update or when enabling this feature but I think it might not be needed.

flamisz:
@tsteur 
> @flamisz not sure with what you mean hide where it includes it?

In the saved segment list dropdown. Hide all segments that has any visitorId dimension, because if we don't hide those, choosing them will throw an error. The change you showed only hide the visitorid dimension from the action list when we create a new segment or editing one.

But we can keep them there and throw the error, as you said it's very rarely used settings, and we can expand the description for that setting to warn the admin.

So basically this issue about not using visitorid segment when this setting is turned on. What I did is:
1. hide the visitorId dimension from the action list when creating or editing saved segments - this is what your code does too
2. hide all saved segments from the dropdown that has visitorid dimension.
3. and I handled the API errors with specific error message
4. made the console commands working

Only the first point is enough? So after these modifications we will throw a `Error: Segment 'visitorId' is not a supported segment` when we want to use it but can't, like using the API or in a console command. This would be a very easy and small PR, only your code is enough for it.

Is this ok? Do we want to hide those saved segments? They can't use it anyway, we just throw the error then?

flamisz:
Just one more thing. If someone has this turned on already and has a cron job to run archiving after an update the job will fail if has any segment with visitorId. That's why I thought maybe we should handle that. But as you said, maybe there is no such a combination at all. I go with the bare minimum, update the pr, we can add more if we want.

tsteur:
> Just one more thing. If someone has this turned on already and has a cron job to run archiving after an update the job will fail if has any segment with visitorId. That's why I thought maybe we should handle that. But as you said, maybe there is no such a combination at all. I go with the bare minimum, update the pr, we can add more if we want.

Fully aware of this. Same problem is happening when you use segments defined by plugins and then disable the plugin. It's a problem for another day I would say :)

PR #17421, opened 2021-04-03
more markdown friendly system checks

Findus23:
While the new copyable system check output is really nice, it also takes up a lot of space in Github issues (e.g. https://github.com/matomo-org/matomo/issues/17420). (same on the forum)

I try to change three things to make the reports more readable when interpreted as Markdown (as on Github and the forum).

- smaller headings (as otherwise they use a lot of space)
- put output into a separate line (this way it isn't also formatted as the heading)
- wrap everything in a `<details>` (this way one doesn't have to scroll past everything when looking at a Github issue)

### Review

* [ ] Functional review done
* [ ] Potential edge cases thought about (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] Usability review done (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] Security review done [see checklist](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] Code review done
* [ ] Tests were added if useful/possible
* [ ] Reviewed for breaking changes
* [ ] Developer changelog updated if needed
* [ ] Documentation added if needed
* [ ] Existing documentation updated if needed

PR #17515, opened 2021-05-02
no links in copy-paste system check

Findus23:
related to #17421
inspired by https://forum.matomo.org/t/no-website-was-found-in-this-matomo-installation/41659/9?u=lukas

Some error messages in the system check contain `<a href="">`. In the plaintext/markdown version those are mostly useless and also trigger the forum limit of max two links per new user. 

Unless I am missing something, this should easily fix this.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Installation/templates/_systemCheckSection.twig
+++ b/plugins/Installation/templates/_systemCheckSection.twig
@@ -17,11 +17,11 @@
 
 #### {{ result.label }}:
 {% for item in result.items -%}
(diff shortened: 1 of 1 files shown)


tsteur:
Works for me. Could maybe still allow some elements (like `|striptags('<br><p><strong><code>')`) or so? Might not be too important though but could make our life easier to still have some formatting in there.

Findus23:
@tsteur Makes sense, I updated it.

PR #17517, opened 2021-05-03
Do not confuse normal rows with the label "-1" w/ the summary row

diosmosis:
### Description:

In several places, matomo will look for the summary row if the label == -1. This causes problems (sometimes fatal) if data is tracked w/ the value `-1`. In this case there will be a normal row w/ the same label as the summary row, causing strange things to happen.

This PR fixes this by relying less on using the label of a row to identify if it is a summary row. Changes include:
- do not index the summary row in the internal row label => row id index in `Row`
- when getting the row from the label, check the row index first, and if it is not found there, and the label is `-1`, then return the summary row (this is mainly for BC)
- when aggregating tables, aggregate the summary row specifically w/o searching for a row w/ a label in either table
- add a flag to `Row`, `isSummaryRow`, which is set by DataTable. this is now returned from `Row::isSummaryRow()` instead of checking the label

### Review

* [ ] Functional review done
* [ ] Code review done


core/DataTable/Filter/ReplaceSummaryRowLabel.php
```diff
@@ -55,11 +55,6 @@ public function filter($table)
         $row = $table->getRowFromId(DataTable::ID_SUMMARY_ROW);
         if ($row) {
             $row->setColumn('label', $this->newLabel);
-        } else {
-            $row = $table->getRowFromLabel(DataTable::LABEL_SUMMARY_ROW);
```
tsteur, line 59: not sure, just a random thought... would we need to iterate over all rows and call `isSummaryRow()` to find a summary row? Not sure when it would have gone into this code path. Or actually we could maybe expose a method 

```php
function getSummaryRow() { return $this->summaryRow; }
```

to the datatable additionally? Although probably not needed if dataTable makes sure it always has the right index

core/DataTable.php
```diff
@@ -1901,23 +1894,28 @@ protected function aggregateRowWithLabel(Row $row, $columnAggregationOps)
...
+        if (empty($thisRow)) {
+            if ($isSummaryRow) {
+                $this->addSummaryRow($otherRow);
             } else {
-                $this->addRow($row);
+                $this->addRow($otherRow);
             }
         } else {
-            $rowFound->sumRow($row, $copyMeta = true, $columnAggregationOps);
+            $thisRow->sumRow($otherRow, $copyMeta = true, $columnAggregationOps);
```
tsteur, line 1916: if $thisRow is not new and summary row will be aggregated based on the label or so, is it possible that we maybe would be aggregating a regular row with a summary row or the other way around and logic might not work 100% correct? Like would maybe need to call `setIsNotSummaryRow` or something and make sure it only aggregates summary rows together?

I'm thinking when it goes into `aggregateRowWithLabel()` and does `$rowFound = $this->getRowFromLabel($labelToLookFor);` then it might receive the summary row?

tsteur (changes requested):
@diosmosis haven't tested it yet but left few quick comments. I'm still seeing heaps of calls to `LABEL_SUMMARY_ROW` and getRowFromLabel. Not sure if this can cause various side effects?

core/DataTable/Filter/ReplaceSummaryRowLabel.php
diosmosis: No, we can do `getRowFromId(DataTable::ID_SUMMARY_ROW);`. The method checks for ID_SUMMARY_ROW and returns `$this->summaryRow`.

core/DataTable.php
diosmosis: > I'm thinking when it goes into aggregateRowWithLabel() and does $rowFound = $this->getRowFromLabel($labelToLookFor); then it might receive the summary row?

It might... we only do that for BC in getRowFromLabel(). Maybe we can introduce a getRowFromLabelWithoutSummaryRow() function or something?

diosmosis: @tsteur updated w/ the smallest fix I could find.

diosmosis:
@tsteur note: I didn't fix every place where the error could occur, since there would be a lot to look for, but I can if needed. Or I can create a new issue to make sure we don't use getRowFromLabel/getRowIdFromLabel w/ LABEL_SUMMARY_ROW.

tsteur:
@diosmosis seems a lot of uses of the label constant are in other plugins anyway so that might be fine for now. I'm thinking though it might be useful to call `setIsSummaryRow` in https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/Actions/ArchivingHelper.php#L573  just to be safe and so isSummaryRow will work correctly? 

Thinking https://github.com/matomo-org/matomo/blob/4.x-dev/core/DataTable.php#L1319-L1321 might need to be tweaked too to not break anything. Be good to check. Eg in truncate check there might be a `setIsSummaryRow` missing.

Also noticed there is a class `DataTableSummaryRow` could we set `isSummaryRow=true` there automatically?

I guess the other uses of the LABEL constant won't really cause any issues but be good to look over it.

I'm thinking a different issue might not be needed then.

diosmosis:
@tsteur 

> I'm thinking though it might be useful to call setIsSummaryRow in https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/Actions/ArchivingHelper.php#L573 just to be safe and so isSummaryRow will work correctly?

We could do this, though it's added as the summary row here: https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/Actions/ArchivingHelper.php#L388

If we do this for a row that isn't added with that method, however, it could cause other unforeseen problems, I think...

> Thinking https://github.com/matomo-org/matomo/blob/4.x-dev/core/DataTable.php#L1319-L1321 might need to be tweaked too to not break anything. Be good to check. Eg in truncate check there might be a setIsSummaryRow missing.

Already saw this. This is fine since the Truncate filter just sets the label of a *new* summary row which it adds. It doesn't search for a row by that label.

> Also noticed there is a class DataTableSummaryRow could we set isSummaryRow=true there automatically?

We could... however if the row is used as a normal row (for whatever reason) then it wouldn't be accurate and would lead to problems.

> I guess the other uses of the LABEL constant won't really cause any issues but be good to look over it.

I did look at those, BUT I didn't look at every use of getRowFromLabel. It's possible code, for example, loops through every row and uses getRowFromLabel (in fact I think I found one place looking quickly a bit before):

```
foreach ($dataTable1->getRows() as $row) {
    $dataTable2->getRowFromLabel($row->getColumn('label'));
}
```

`getRows()` includes the summary row so this would have the same issue, but we wouldn't know since the label isn't explicitly used here. We can look for these too, but it would take more time.

sgiehl:
Just tried local archiving with those changes, and the warning that was triggered before, isn't triggered anymore. But not yet sure if those changes might have other implications.

tests/PHPUnit/Integration/ArchiveProcessingTest.php
```diff
@@ -635,6 +635,79 @@ public function test_aggregateDataTableRecords_handlesPartialArchives()
...
+	<row>
+		<label>-1</label>
+		<nb_visits>12</nb_visits>
+		<nb_actions>6</nb_actions>
+	</row>
+	<row>
+		<label>-1</label>
+		<nb_visits>25</nb_visits>
+		<nb_actions>40</nb_actions>
+	</row>
```
sgiehl, line 704: Should we maybe have a system test for this case, so we see that only one of those rows' label would be replaced with `Others`?

diosmosis: added a test

diosmosis:
@tsteur @sgiehl I updated the PR, can you take another look and let me know if it's ok to merge or apply the fix on cloud?

sgiehl:
The changes look good for me now. Not sure if we might have any issues with archives that have been built before, with mixed up row with `-1` and summary row.

diosmosis:
@sgiehl I think the tables themselves will be handled fine, they'll just have been summed incorrectly (eg, adding the -1 row to the summary row or the summary row to the -1 row). The data would be off but we can't fix that w/o invalidating.

tsteur:
Looks good @diosmosis 

> We could do this, though it's added as the summary row here: https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/Actions/ArchivingHelper.php#L388

I would maybe make that change just in case it'll be used differently in the future. Otherwise once tests pass all good 👍

Issue #15575, opened 2020-02-16
Persistent notifications aren't removed. Make Error while creating/copying file to warning dismissable


Findus23:
quite low priority, but I wanted to write it down just in case I stumble across it again in the future.
(and maybe the same is true for other error messages and it confuses user)

When developing Matomo doesn't have write permission to `plugins/`. This doesn't matter as it isn't needed for things apart from the marketplace.

When one tries to download something from the marketplace this error message appears rightfully:

![grafik](https://user-images.githubusercontent.com/6266037/74612027-b0212200-5101-11ea-865e-6396dec7bf23.png)

The issue is that there is no way to get rid of the notification. Pressing the x only closes it with JS, but as it is rendered by Twig it will reappear on every page even when it doesn't matter anymore. 

Adding a `static::$session->unsetAll();` to 
https://github.com/matomo-org/matomo/blob/b16a791aa3650d85af829156129c2bd44c7cb075/core/Notification/Manager.php#L191-L202

removes it from the session and gets finally rid of the message (ending the session in another way would also work).

sgiehl:
maybe related to #15498

Findus23:
@sgiehl That sounds like the same general issue (notifications in the session don't disappear)

Findus23:
https://github.com/matomo-org/matomo/issues/12934 is also caused by the same issue.

tsteur:
To fix this problem we want to fix the generic problem that a persistent notification only seems to disappear when logging out and logging in.

You can reproduce this by applying this change

```diff
diff --git a/plugins/CorePluginsAdmin/PluginInstaller.php b/plugins/CorePluginsAdmin/PluginInstaller.php
index 4619ecca9f..11f10a1ac6 100644
--- a/plugins/CorePluginsAdmin/PluginInstaller.php
+++ b/plugins/CorePluginsAdmin/PluginInstaller.php
@@ -52,6 +52,7 @@ class PluginInstaller
 
     public function installOrUpdatePluginFromMarketplace($pluginName)
     {
+        throw new \Exception('foo');
         $this->checkMarketplaceIsEnabled();
 
         $this->pluginName = $pluginName;
```

Then go to marketplace, and try to install a plugin. Then go back for example to the marketplace page or the Platform -> API page in the Matomo admin.

[...]

PR #17518, opened 2021-05-04
Update mark notification as read method

flamisz:
### Description:

As in title.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CoreHome/angularjs/notification/notification.controller.js
+++ b/plugins/CoreHome/angularjs/notification/notification.controller.js
@@ -20,6 +20,7 @@
                 return;
             }
 
+            piwikApi.withTokenInUrl();
             piwikApi.post(
                 { // GET params
                     module: 'CoreHome',
--- a/plugins/CoreHome/angularjs/notification/notification.controller.js
+++ b/plugins/CoreHome/angularjs/notification/notification.controller.js
@@ -14,8 +14,7 @@
          * Marks a persistent notification as read so it will not reappear on the next page
          * load.
          */
-        this.markNotificationAsRead = function () {
-            var notificationId = this.notificationId;
+        this.markNotificationAsRead = function (notificationId) {
             if (!notificationId) {
                 return;
             }
--- a/plugins/CoreHome/Controller.php
+++ b/plugins/CoreHome/Controller.php
@@ -11,6 +11,7 @@ namespace Piwik\Plugins\CoreHome;
 use Exception;
 use Piwik\API\Request;
 use Piwik\Common;
+use Piwik\DataTable\Renderer\Json;
 use Piwik\Date;
 use Piwik\FrontController;
 use Piwik\Notification\Manager as NotificationManager;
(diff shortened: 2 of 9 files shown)


sgiehl (changes requested):
@flamisz  Did you test that locally? For me the XHR request is actually never sent, as the notificationId is always null. Guess it needs to be passed to `markNotificationAsRead` from the html template instead, as it isn't available in the controller.

tsteur:
btw @sgiehl @flamisz not sure what you exactly meant in the comment but if the request isn't sent currently that might be because of https://github.com/matomo-org/matomo/issues/15575#issuecomment-822897559 which may need to be fixed in order to test this issue :)

tsteur:
it might be totally unrelated though so feel free to ignore, only thought to mention it in case it's related

flamisz:
> @flamisz Did you test that locally? For me the XHR request is actually never sent, as the notificationId is always null. Guess it needs to be passed to `markNotificationAsRead` from the html template instead, as it isn't available in the controller.

@tsteur @sgiehl I realizes something is not ok with the notification, but I thought I just don't know how to use it properly. I was able to test it tho. I created permanent notification (hardcoded somewhere into a plugin) and used an API call to mark as readed. Before the PR I was able to mark as readed so it disappeared for the next request. With this modifications I needed a token to do that.

tsteur:
👍  we'd just need to double check that we also send the token in the actual request from the notification JS otherwise it won't work basically (but it's not working there currently anyway so could also ignore it and not test it)

plugins/CoreHome/angularjs/notification/notification.controller.js
```diff
@@ -20,6 +20,7 @@
                 return;
             }
 
+            piwikApi.withTokenInUrl();
```
flamisz, line 22: > 👍 we'd just need to double check that we also send the token in the actual request from the notification JS otherwise it won't work basically (but it's not working there currently anyway so could also ignore it and not test it)

@tsteur this pr contains this modification here ⬆️

sgiehl:
@flamisz while reviewing I actually saw why the notification xhr didn't work properly. I've pushed a commit to this branch to fix that. So maybe this then also fixes #15575

flamisz:
> @flamisz while reviewing I actually saw why the notification xhr didn't work properly. I've pushed a commit to this branch to fix that. So maybe this then also fixes #15575

thanks @sgiehl, this makes sense. I just wasn't sure if it's a bug or my lack of knowledge 😄

sgiehl:
@flamisz I've rebased the branch and added some new UI tests to check that closing the notifications work as expected. Feel free to have a look at the UI tests and merge the PR if they look good to you.

flamisz:
> @flamisz I've rebased the branch and added some new UI tests to check that closing the notifications work as expected. Feel free to have a look at the UI tests and merge the PR if they look good to you.

Thanks, @sgiehl, looking good 👍  
Probably wait with the merge until we have the new release.

sgiehl:
Might maybe be worth to include that in 4.3.0 maybe, as it fixes the uncloseable notification issue. ping @tsteur

PR #17519, opened 2021-05-04
Allow invalidating empty site

diosmosis:
### Description:

Some sites can legitimately have 0 visits, like sites imported from google. While debugging the GoogleAnalyticsImporter test failures I noticed nothing was being invalidated because of this check. Removing it in this PR.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -272,12 +272,6 @@ class ArchiveInvalidator
             list($plugin) = explode('.', $name);
         }
 
-        // remove sites w/ no visits
-        $trackerModel = new TrackerModel();
-        $idSites = array_filter($idSites, function ($idSite) use ($trackerModel) {
-            return !$trackerModel->isSiteEmpty($idSite);
-        });
-
         if ($plugin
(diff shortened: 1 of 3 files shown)


diosmosis:
Noticed this has other changes in it, will remove them...

diosmosis:
Updated.

tsteur:
@diosmosis do you remember why it was maybe added in https://github.com/matomo-org/matomo/commit/f5e9420a987340b036fa342e876ab92e314f4ec7#diff-88e28f71aefcd536256f874d55fa5d6eb2783391010861ed67461ea6bbda4ef2R261-R265 ? I suppose this also impacts roll ups. Looks good otherwise

diosmosis:
@tsteur it was an optimization. IIRC you suggested it in a review (I can't remember the exact reasoning), then decided it wasn't that important, but I added it anyway because "why not?". (I guess this is "why not" :) ). I guess we could still do the check, though, if we use the event in Loader.php (isSiteUsingTracker or something similar).

diosmosis:
@tsteur found another regression causing the importer tests to fail, added in the latest commit.

PR #17520, opened 2021-05-04
Use token auth when widgetize

### LINKED ISSUE #17335: Widgetize request with token_auth param fails if superuser session exists
As your documentation says I have created a view-only user with a token and when using that token to show a widget I'm getting an error saying "This user has superuser access". It works if I load the same url while logged out or in another browser. Seems odd if it shouldn't work since this should be a fairly common case.

Example url: http://localhost:8000/index.php?module=Widgetize&action=iframe&forceView=1&disableLink=1&token_auth=xxx&moduleToWidgetize=VisitsSummary&actionToWidgetize=getEvolutionGraph&viewDataTable=graphEvolution&idSite=1&date=2021-02-09,2021-03-10&period=range

## Expected Behavior
It should be possible to show the widget even if you are currently logged in as a super user in the same browser.

## Current Behavior
Error message saying
"This user has super user access. For embedding widgets super user token auths are not allowed. See our faq for more information."

## Possible Solution
I have tried to follow the code and it seems that when using the token_auth parameter there is a call to the method `Request::reloadAuthUsingTokenAuth` eventually landing in `Access::reloadAccess`, however it return without reloading since `$this->hasSuperUserAccess` is still true from the session auth. 

1. Create user with view permissions and generate a token
2. Create a widgetized url and add the token to the url
3. Load the url in a browser where you are logged in as a super user

## Context
We are trying to embed widgets in an external dashboard.

## Your Environment

I'm running your latest docker image.
* Matomo Version: 4.2.1

flamisz:
### Description:

fixes #17335 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -675,9 +675,9 @@ class FrontController extends Singleton
             return null;
         }
 
-        if (Common::getRequestVar('token_auth', '', 'string') !== '') {
-            return null;
-        }
+        if (Common::getRequestVar('token_auth', '', 'string') !== '' && !Common::getRequestVar('force_api_session', 0)) {
+             return null;
+         }
 
         $module = Common::getRequestVar('module', self::DEFAULT_MODULE, 'string');
         $action = Common::getRequestVar('action', false);
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -430,7 +430,9 @@ class FrontController extends Singleton
         ) {
             Request::reloadAuthUsingTokenAuth();
             Request::checkTokenAuthIsNotLimited($module, $action);
-            Session::close();
+            if ($module === 'Widgetize') {
+                Session::close();
+            }
         }
 
         SettingsServer::raiseMemoryLimitIfNecessary();
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -428,8 +428,13 @@ class FrontController extends Singleton
         if (Common::getRequestVar('token_auth', '', 'string') !== ''
             && Request::shouldReloadAuthUsingTokenAuth(null)
         ) {
+            if (!Common::isPhpCliMode()) {
+                Access::getInstance()->setSuperUserAccess(false);
+            }
+
             Request::reloadAuthUsingTokenAuth();
             Request::checkTokenAuthIsNotLimited($module, $action);
+            Session::close();
         }
 
         SettingsServer::raiseMemoryLimitIfNecessary();
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -430,9 +430,6 @@ class FrontController extends Singleton
         ) {
             Request::reloadAuthUsingTokenAuth();
             Request::checkTokenAuthIsNotLimited($module, $action);
-            if ($module === 'Widgetize') {
-                Session::close();
-            }
         }
 
         SettingsServer::raiseMemoryLimitIfNecessary();
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -428,10 +428,6 @@ class FrontController extends Singleton
         if (Common::getRequestVar('token_auth', '', 'string') !== ''
             && Request::shouldReloadAuthUsingTokenAuth(null)
         ) {
-            if (!Common::isPhpCliMode()) {
-                Access::getInstance()->setSuperUserAccess(false);
-            }
-
             Request::reloadAuthUsingTokenAuth();
             Request::checkTokenAuthIsNotLimited($module, $action);
             Session::close();
@@ -676,6 +672,10 @@ class FrontController extends Singleton
         if (Common::isPhpClimode()
             && !defined('PIWIK_TEST_MODE')
         ) { // don't use the session auth during CLI requests
+            return null;
+        }
+
+        if (Common::getRequestVar('token_auth', '', 'string') !== '') {
             return null;
         }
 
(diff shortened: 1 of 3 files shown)


core/FrontController.php
```diff
@@ -428,8 +428,13 @@ public function init()
         if (Common::getRequestVar('token_auth', '', 'string') !== ''
             && Request::shouldReloadAuthUsingTokenAuth(null)
         ) {
+            if (!Common::isPhpCliMode()) {
```
tsteur: @flamisz might be worth trying not to use the session authenticator when a token_auth is used in the query. Eg return `null` in `makeSessionAuthenticator()` if a session is set? Unsetting super user access etc may have side effects not sure

flamisz:
Hi @tsteur, I updated the PR, not resetting the super user access. During this update I've seen this comment in the code:

https://github.com/matomo-org/matomo/blob/104135aebddd112567f5ec17b8f11e757c7e57f2/core/FrontController.php#L426-L433

But I modified the `reloadAccess` method and now it works with widgets as well. Do you think it was delibaretly did this way? Could it be any side effect somewhere?
https://github.com/matomo-org/matomo/blob/104135aebddd112567f5ec17b8f11e757c7e57f2/core/Access.php#L137

tsteur:
@flamisz I think we can undo the changes in `Access.php` and apply this change:

```diff
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -675,7 +675,8 @@ class FrontController extends Singleton
             return null;
         }
 
-        if (Common::getRequestVar('token_auth', '', 'string') !== '') {
+        if (Common::getRequestVar('token_auth', '', 'string') !== ''
+            && !Common::getRequestVar('force_api_session')) {
             return null;
         }
 ```

It basically means if there's a `force_api_session=1` parameter then we still load the session and authenticate the session first which is needed for the session token to work. This way the dashboard etc will work and all other requests should be working too. It be only an issue if users falsely have `force_api_session=1` set in the request and don't use a session token but a real token auth. Then it won't work but then it generally shouldn't work anyway AFAIK.

It be still good for someone to review this though additionally.

**Comment by flamisz** [comment]
> @flamisz I think we can undo the changes in `Access.php` and apply this change:

I just realized the way it was is not working and tried to figure out how it should 😄. Thanks, I was going to check the `force_api_session` but wasn't sure where yet. 

I'll test it how it works this way.

File: core/FrontController.php [hunk 2]
```diff
@@ -430,6 +430,7 @@ public function init()
         ) {
             Request::reloadAuthUsingTokenAuth();
             Request::checkTokenAuthIsNotLimited($module, $action);
+            Session::close();
```
sgiehl: Any reason for closing the session at this point?

flamisz: During the process somewhere we start it, but actually we don't need it as we handle this request as stateless, so I thought we can close it (the original idea was coming from @tsteur).

sgiehl: In theory that might make sense, but not sure if that could cause any regressions if someone tries to perform a controller action with a token_auth, that actually tries to change the session. But maybe we can simply change it and wait if any regressions are reported.

tsteur: It should be fine to close the session if `force_api_session` is not present in the URL AFAIK. There could be a regression but it would be hard to know. In theory though in these cases the session shouldn't be started anyway.

diosmosis: Would this be an issue for the "embed all of matomo" use case (one big user uses this, can't remember their name starts w/ an 'M')? Ie, the `[General] enable_framed_pages` `[General] enable_framed_settings` config. I would think the session might be written to there.

diosmosis: Also, thought it might be useful to have test pages for these, so created: https://github.com/matomo-org/test-examples/pull/6

flamisz: I'm not familiar with those settings (framed). I tried with the new `test-example` and worked. How else should we test it? @diosmosis @tsteur

diosmosis: They allow users to embed whole matomo pages. One company I know of uses it to apply their own branding to all of Matomo and embed it as part of their website for their customers.

I think to test that it would work we'd need to write the session when a page is framed and see if it persists? Maybe sending a notification from PHP and seeing it in the embedded admin page?

flamisz: I tried something.

1. created a persistent notification in a controller
2. uncommented the line where I created it
3. reloaded Matomo page, the notification was there

I did the same when Matomo was running inside an iframe. This time at step 3, the notification was not there. It means the session was not persisted anyway.

What is the conclusion here? Should we keep that `Session::close()` line?
ping @tsteur @diosmosis @sgiehl

diosmosis: We could also just do the Session::close() if module==Widgetize. That might be enough?

flamisz: done 👍

tsteur: I realise using `new SessionNamespace` will actually lounge a session. Eg in `redirectDashboardToWelcomePage` method it might start a session and that might even happen when it's an API call or when the action is embedded as an iframe. Closing it now could break some things that were working before (whether they should have been working before or not is a different question). I didn't debug but likely the close shouldn't do anything because ideally it doesn't start the session in the first place so I'm tempted to simply remove the close call. And eg if `force_api_session=1` then the session might be still needed in order to write session changes

tsteur: If the request is an API call then we for sure don't want to use any session but widgetized requests may use a session I would say

diosmosis: @flamisz I think ^ means my idea wouldn't be enough

flamisz: Ok, I tried a couple of things, tested what I could, and looks like the `force_api_session` does the job here. If that's false, we don't have a session, so don't need to close. 

I couldn't find a case when we really need to close it.

PR #17521, opened 2021-05-04
Improve tracker performance when using custom dimensions (one less update query per tracking request) 

### LINKED ISSUE #17173: Improve tracker performance when using custom dimensions (one less update query per tracking request)
Now that custom dimensions is included in core I think we could improve the performance of the tracker and have less updates.

Currently, as soon as a custom dimension is configured for a site, we execute an additional query like `UPDATE log_visit SET last_idlink_va = '2222' WHERE idsite = '1' AND idvisit = '3333'`.

We could maybe get rid of this extra update query for existing visits if we knew we first insert the action and then update the visit. Then we could in the regular visit update also update the `last_idlink_va`. Not sure if there would be maybe a reason this wouldn't work and whether we have to update the visit before the action.

For new visits this would not work because we need to have the idvisit to insert the action. This likely makes this quite tricky to implement...

I was thinking we could maybe also get rid of the extra action update but this is not the case because it actually updates the action before this action.

**Note: the query is only executed when there are action custom dimensions defined so its impact is limited.**

tsteur:
## Description:

Performance improvement to have one less update on log_visit for existing visits if action dimensions are used: fix https://github.com/matomo-org/matomo/issues/17173

### Before this PR

* Insert new visit -> so get `idVisit`
* Insert log link action -> so we get `idlink_va`
* update visit to set value of the created `idlink_va` on the column `last_idlink_va`.

### After this PR

Same logic as before. We always have to create visit first before creating any log link action.

### Before this PR

* Update visit (we have idvisit already at this stage)
* Insert log link action -> so we get `idlink_va`
* update visit to set value of the created `idlink_va` on the column `last_idlink_va`.

### After this PR

* Insert log link action -> so we get `idlink_va`
* Update visit including setting the value of the created `idlink_va` on the column `last_idlink_va`.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Tracker/Visit.php
+++ b/core/Tracker/Visit.php
@@ -17,6 +17,7 @@ use Piwik\Date;
 use Piwik\Exception\UnexpectedWebsiteFoundException;
 use Matomo\Network\IPUtils;
 use Piwik\Plugin\Dimension\VisitDimension;
+use Piwik\Plugins\Actions\Tracker\ActionsRequestProcessor;
 use Piwik\Plugins\UserCountry\Columns\Base;
 use Piwik\Tracker;
 use Piwik\Tracker\Visit\VisitProperties;
@@ -194,6 +195,13 @@ class Visit implements VisitInterface
         // AND
         // - the last page view for this visitor was less than 30 minutes ago @see isLastActionInTheSameVisit()
         if (!$isNewVisit) {
+
+            foreach ($this->requestProcessors as $processor) {
+                if ($processor instanceof ActionsRequestProcessor) {
+                    Common::printDebug("Executing " . get_class($processor) . "::recordLogs()...");
+                    $processor->recordLogs($this->visitProperties, $this->request);
+                }
+            }
             try {
                 $this->handleExistingVisit($this->request->getMetadata('Goals', 'visitIsConverted'));
             } catch (VisitorNotFoundInDb $e) {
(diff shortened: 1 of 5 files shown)


core/Tracker/Visit.php
```diff
@@ -213,6 +214,10 @@ public function handle()
         $this->request->setThirdPartyCookie($this->request->getVisitorIdForThirdPartyCookie());
 
         foreach ($this->requestProcessors as $processor) {
+            if (!$isNewVisit && $processor instanceof ActionsRequestProcessor) {
+                // already processed earlier when handling exisitng visit see {@link self::handleExistingVisit()}
+                continue;
+            }
```
diosmosis, line 220: could this be put in ActionsRequestProcessor? ie, return if conditions are met?

diosmosis: oh i see, this has to be done before updating a visit... seems like it would all be clearer if this was all in ActionsRequestProcessor, though i'm not sure if that's possible

tsteur: I know it's not really clear which is why I'm not sure yet if it's worth merging it. Problem is for new visits we have to execute the actions processor later vs for existing visits we have to execute it before updating a visit...

diosmosis (approved):
not sure if there's anything else to do here, but this looks good to merge

PR #17522, opened 2021-05-05
Outlink tracking when DOM modified

### LINKED ISSUE #15780: Outlink and download tracking not working for links added after DOM ready (which many are these days eg angular, vue.js, react, ... SPA).

flamisz:
### Description:

fixes #15780

### Review

* [ ] Functional review done
* [ ] Code review done


flamisz:
build js

flamisz:
build js

diosmosis:
I see how this works, using the root element to always handle clicks (and it's definitely simpler), but I think there's potentially an issue if event bubbling is cancelled (ie, via stopPropagation()) somewhere along the way. In that case we might not track any outlinks. Is this something we want to handle @tsteur?

tsteur:
@diosmosis that's something we can't prevent but we set the third parameter in addEventListener to `true` so we should be receiving the event as they are usually not cancelled in the `useCapture` mode.

From https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

> A Boolean indicating whether events of this type will be dispatched to the registered listener before being dispatched to any EventTarget beneath it in the DOM tree

it basically starts in the html/body elements and then goes down before it would bubble up again AFAIK. And usually someone might stop propagation only at a later stage (they have default useCapture=false). It works quite well eg in TagManager etc and it shouldn't be a problem

js/piwik.js
```diff
@@ -4627,32 +4652,20 @@ if (typeof window.Matomo !== 'object') {
...
-                                // we make sure to setup link only once for each tracker
-                                linkElement.matomoTrackers.push(trackerInstance);
-                                addClickListener(linkElement, enable);
-                            }
-                        }
-                    }
+                if (enable) {
+                    addEventListener(documentAlias.body, 'mouseup', clickHandler(enable), true);
+                    addEventListener(documentAlias.body, 'mousedown', clickHandler(enable), true);
+                    addEventListener(documentAlias.body, 'contextmenu', clickHandler(enable), true);
```
diosmosis: This is basically the same code as `addClickListener()` except w/ useCapture = true, right? If so, adding a useCapture parameter to addClickListener, and then calling it here would mean less code, would that be better?

diosmosis:
Tested locally, works well. Created https://github.com/matomo-org/test-examples/pull/7 for testing.

Also left one small comment, otherwise looks good.

tsteur:
fyi because the 4.3 release is soon and the RC is already out we might want to merge that one only after the 4.3 release just in case anything is breaking etc.  And then we could directly update the SPA guide etc to mention the method no longer needs to be called etc

diosmosis:
@tsteur should we change the milestone?

tsteur:
Changed it for now 👍  Hoping it won't be later renamed to 4.5.0 after releasing 4.3. Then we would change the milestone again.

I could create a 4.3.1 milestone but there we'd mostly want fixes and not changes in there.

js/piwik.js
flamisz: yeah, I didn't realize it, after all the modifications basically they are the same. I'll do it

flamisz: updated

flamisz:
build js

PR #17523, opened 2021-05-05
Updates phpmailer/phpmailer to 6.4.1

sgiehl:
### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1440,16 +1440,16 @@
         },
         {
             "name": "phpmailer/phpmailer",
-            "version": "v6.1.8",
+            "version": "v6.4.1",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


flamisz:
Hi @sgiehl, how come the `composer.json` is not updated in this PR? Shouldn't we do that?

sgiehl:
@flamisz That's only needed when the composer.json actually prevents an update due to the version restriction.

diosmosis:
@sgiehl in the future, when we run composer update to update dependencies, will we lose this change?

sgiehl:
no, why should we? I simply ran a `composer update phpmailer/phpmailer` and added the changes of `composer.lock` as usual.

flamisz (approved):

PR #17526, opened 2021-05-06
Check if config loaded in getHostname()

flamisz:
### Description:

If the config already loaded when we use the `getHostname()` method, we can use the `$checkIfTrusted` as true when getting the host.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Config.php
+++ b/core/Config.php
@@ -225,8 +225,17 @@ class Config
      */
     public static function getHostname()
     {
+        $checkIfTrusted = true;
+        
+        try {
+            // this fails before the config was initialized
+            Config::getInstance()->existsLocalConfig();
+        } catch (\Exception $e) {
+            $checkIfTrusted = false;
+        }
+        
         // Check trusted requires config file which is not ready yet
-        $host = Url::getHost($checkIfTrusted = false);
+        $host = Url::getHost($checkIfTrusted);
 
         // Remove any port number to get actual hostname
         $host = Url::getHostSanitized($host);
--- a/core/Config.php
+++ b/core/Config.php
@@ -220,21 +220,13 @@ class Config
 
     /**
      * Returns the hostname of the current request (without port number)
+     * @param bool $checkIfTrusted Check trusted requires config which is maybe not ready yet,
+     *                             make sure the config is ready when you call with true
      *
      * @return string
      */
-    public static function getHostname()
+    public static function getHostname($checkIfTrusted = false)
     {
-        $checkIfTrusted = true;
-
-        try {
-            // this fails before the config was initialized
-            Config::getInstance();
-        } catch (\Exception $e) {
-            $checkIfTrusted = false;
-        }
-
-        // Check trusted requires config file which is not ready yet
         $host = Url::getHost($checkIfTrusted);
 
         // Remove any port number to get actual hostname
--- a/core/Config.php
+++ b/core/Config.php
@@ -139,7 +139,7 @@ class Config
         if (!empty($GLOBALS['CONFIG_INI_PATH_RESOLVER']) && is_callable($GLOBALS['CONFIG_INI_PATH_RESOLVER'])) {
             return call_user_func($GLOBALS['CONFIG_INI_PATH_RESOLVER']);
         }
-        
+
         $path = self::getByDomainConfigPath();
         if ($path) {
             return $path;
@@ -226,14 +226,14 @@ class Config
     public static function getHostname()
     {
         $checkIfTrusted = true;
-        
+
         try {
             // this fails before the config was initialized
-            Config::getInstance()->existsLocalConfig();
+            Config::getInstance();
         } catch (\Exception $e) {
             $checkIfTrusted = false;
         }
-        
+
         // Check trusted requires config file which is not ready yet
         $host = Url::getHost($checkIfTrusted);
 
@@ -318,7 +318,7 @@ class Config
     public function deleteLocalConfig()
     {
         $configLocal = $this->getLocalPath();
-        
+
         if(file_exists($configLocal)){
             @unlink($configLocal);
         }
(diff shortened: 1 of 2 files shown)


core/Config.php
```diff
@@ -225,8 +225,17 @@ public static function getByDomainConfigPath()
      */
     public static function getHostname()
     {
+        $checkIfTrusted = true;
+        
+        try {
+            // this fails before the config was initialized
+            Config::getInstance()->existsLocalConfig();
```
sgiehl: Not sure how often we are calling `getHostname` in the code, but this would actually add an additional file operation each time it's called. Wondering if we should check that another way if possible or maybe cache the result, so we don't check if the file exists each time it is called?

flamisz: I will rethink this and try to find a proper way to check if the config loaded instead of a `try...catch`

tsteur: @flamisz it's not about whether config file exists here it's more about checking if config has been loaded yet. Using Config::getInstance() here could actually result in a never ending loop as the getHostname method could be called while loading the config (might not be the case but it could be). 

I'm not sure it's possible to know if config has been loaded yet or not and in worst case we have to add a parameter `checkIfTrusted`

flamisz: the only way I find and we use somewhere else in the code is trying to get the instance itself. That throws an error in the `StaticContainer` if does not exists yet.

we don't need the `->existsLocalConfig();` tho, simple `Config::getInstance()` is enough but I haven't checked the never ending loop situation.

As you said, maybe adding the parameter is the solution.

flamisz:
@tsteur @sgiehl 

I updated the PR using parameter instead of the `try..catch`. Using this `getHostname()` method is very rare in our codebase.

We use it in the core updater in a couple of files and in `SettingsPiwik` but I didn't modify it there.

sgiehl (commented):
The code itself looks ok. But not sure if it fulfills the requirements. ping @tsteur 
When visiting the diagnostic page over a trusted host, the hostname is shown in the archiving command. Coming from an untrusted host for me it now always uses the IP address as `matomo-domain`, even though some trusted hosts are configured...

tsteur:
@sgiehl are you saying the `trusted host` check doesn't work? Could you maybe quickly debug why it's using the IP address for you?

In my case it worked without an issue 
image

sgiehl:
@tsteur Actually I guess the code does exactly what it should. If the current hostname used is untrusted the code falls back to `$_SERVER['SERVER_ADDR']`, which is the server IP:
https://github.com/matomo-org/matomo/blob/0d979018ce23e8e0187dd435cc8de99fcf3c3585/core/Url.php#L298-L312

Was just wondering if it would be more useful in this case to lookup the configured trusted hosts instead and use one of them instead of the IP. But might be unneeded, as viewing the diagnostics from an untrusted host might be uncommon.

tsteur:
Ideally if it's not a trusted host then we wouldn't append the matomo-domain option as it wouldn't work anyway. Not sure how easy it is but there should be methods for checking if we are on a trusted host or not.

sgiehl:
@tsteur guess only appending the `matomo-domain` part if `Url::isValidHost()` returns true should work

tsteur:
@flamisz could you look into this?

flamisz:
> @flamisz could you look into this?

yeah, sure

flamisz:
@tsteur @sgiehl where can I see that `matomo-domain` parameter? just to be able to test it.

flamisz:
@tsteur @sgiehl where can I see that `matomo-domain` parameter? just to be able to test it.

tsteur:
@flamisz it's right in the line below `        $domain = Config::getHostname($checkIfTrusted = true);`.

flamisz:
> @flamisz it's right in the line below ` $domain = Config::getHostname($checkIfTrusted = true);`.

@tsteur I meant in the UI or in any console command response. Like your screenshot, where is that from? So when I fix it I can see it works now.

tsteur:
@flamisz that's the system check page. You get to it by going to `Administration -> Diagnostics -> System Check`. It would be shown under `Last Successful Archiving Completion`. It's possible it won't be shown if archiving ran recently so you might need to temporarily adjust `CronArchivingLastRunCheck` class to force showing it.

flamisz:
> @flamisz that's the system check page. You get to it by going to `Administration -> Diagnostics -> System Check`. It would be shown under `Last Successful Archiving Completion`. It's possible it won't be shown if archiving ran recently so you might need to temporarily adjust `CronArchivingLastRunCheck` class to force showing it.

thanks, I found it, I'm on it 😄

flamisz:
@tsteur @sgiehl I updated the PR

plugins/Diagnostics/Diagnostic/CronArchivingLastRunCheck.php
```diff
@@ -90,8 +91,13 @@ public function execute()
 
     private function getArchivingCommand()
     {
-        $domain = Config::getHostname();
-        return PIWIK_INCLUDE_PATH . '/console --matomo-domain=' . $domain . ' core:archive';
+        if (Url::isValidHost()) {
+            $domain = Config::getHostname($checkIfTrusted = true);
```
sgiehl, line 95: doesn't `isValidHost()` actually automatically check if the host is trusted (if not disabled in config)? Wondering if we still need the changes in `Config::getHostname()`

tsteur: Probably good to still have it 👍

sgiehl: ok

sgiehl (approved):
Seems to work as expected now. I've rebased the branch, if all tests are still working it should be good to merge

PR #17527, opened 2021-05-06
Add link to new FAQ how to make the diagnostic “Managing processes via CLI” show Ok

mattab:
NOTE: Had to add the risky `|raw` to the item.comment, which may have security risks if some of the "Informational" diagnostics will contain random content that may be injectable by someone or the system. Maybe we should audit all informational diagnostics?

The new FAQ is: https://matomo.org/faq/troubleshooting/how-to-make-the-diagnostic-managing-processes-via-cli-to-display-ok/

Here is what it looks like:
Screenshot from 2021-05-06 17-01-20

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Diagnostics/Diagnostic/CronArchivingCheck.php
+++ b/plugins/Diagnostics/Diagnostic/CronArchivingCheck.php
@@ -78,7 +78,7 @@ class CronArchivingCheck implements Diagnostic
             $comment .= $this->translator->translate('Installation_NotSupported')
                 . ' ' . $this->translator->translate('Goals_Optional')
                 . ' (' . $this->translator->translate('General_Reasons') . ': ' . $reasonText . ')'
-                . $this->translator->translate('General_LearnMore', [' <a target="_blank" href="https://matomo.org/faq/troubleshooting/how-to-make-the-diagnostic-managing-processes-via-cli-to-display-ok/">', '</a>'])
+                . $this->translator->translate('General_LearnMore', [' <a target="_blank" href="https://matomo.org/faq/troubleshooting/how-to-make-the-diagnostic-managing-processes-via-cli-to-display-ok/">', '</a>']);
             $status = DiagnosticResult::STATUS_INFORMATIONAL;
         }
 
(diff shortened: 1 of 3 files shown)


sgiehl:
@mattab Actually it seems all other informational results are handled using this method:
https://github.com/matomo-org/matomo/blob/11f54ee6389bd8ac408385b0d1760936c496470c/plugins/Diagnostics/Diagnostic/DiagnosticResult.php#L62-L72
So maybe we could add some kind of html escaping there, to prevent any injection coming from server configuration variables or HTTP headers.

mattab:
Sounds good @sgiehl - unfortunately i won't have time to finish the work on this so maybe someone can do it :see_no_evil:

tsteur:
We can look into this eventually as part of 4.4 release

diosmosis:
If we escape the comment we won't be able to put the link in, correct? We could accomplish the same thing w/ another variable that's meant to not contain unsafe input (eg, `item.details` after `item.comment`).

sgiehl:
@diosmosis I've rebased the branch and updated it so comments for informational items are escaped by default. Even though it  currently seems unneeded, there is a new parameter to disable that.

plugins/Diagnostics/Diagnostic/CronArchivingCheck.php
```diff
@@ -77,7 +77,8 @@ public function execute()
             }
             $comment .= $this->translator->translate('Installation_NotSupported')
                 . ' ' . $this->translator->translate('Goals_Optional')
-                . ' (' . $this->translator->translate('General_Reasons') . ': ' . $reasonText . ')';
+                . ' (' . $this->translator->translate('General_Reasons') . ': ' . $reasonText . ')'
+                . $this->translator->translate('General_LearnMore', [' <a target="_blank" href="https://matomo.org/faq/troubleshooting/how-to-make-the-diagnostic-managing-processes-via-cli-to-display-ok/">', '</a>']);
             $status = DiagnosticResult::STATUS_INFORMATIONAL;
```
diosmosis, line 82: @sgiehl So, below, we still need to set `$escapeComment` to false, right?

sgiehl: actually not, as it doesn't use the `informationalResult` method, but created the `DiagnosticResultItem` item directly

diosmosis: :+1:

PR #17530, opened 2021-05-07
Allow invalidating non-day periods even if the period is for a date that had logs purged

diosmosis:
### Description:

When invalidating data, we check for dates that are before the configured log data deletion month count, and ignore them. This is a safety mechanism to make sure we do not invalidate day periods that have archive data, but for whom log data was deleted. This safety mechanism, however, breaks the GoogleAnalyticsImporter, when importing data that is before log data deletion threshold.

The importer imports day periods, then invalidates week periods and above to calculate the rest. If the data is before the log deletion threshold, the higher periods are never recomputed.

Fixed in this PR by only applying the check for day periods. If a week period is invalidated and there is no log data, the day periods should used.

TODO: check if week periods will trigger rearchiving of the day if the day is for some reason invalidated.

FYI @tsteur

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -701,14 +701,17 @@ class ArchiveInvalidator
      */
     private function removeDatesThatHaveBeenPurged($dates, $period, InvalidationResult $invalidationInfo)
     {
+        $period = $period ?: 'day';
+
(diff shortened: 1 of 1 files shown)


diosmosis:
closing in favor of a new option

Issue #14447, opened 2019-05-14
Feature request: option to disable tracking failure notifications


timodenissen:
Hello,

as asked in the forum [1]: please add an option to disable the automatic notifications about tracking failures.

Cheers,
Timo

[1] https://forum.matomo.org/t/how-to-disable-e-mails-about-tracking-failures/32854

fdellwing:
Is it enough to provide an option via `config.ini.php` or do we need to make this configurable via UI?

timodenissen:
An option in `config.ini.php` is fine for me, perhaps at a later time an option via UI?

fdellwing:
Did a PR for this, but others have to decide if that is enough ;)

You can test the change by applying this patch: https://patch-diff.githubusercontent.com/raw/matomo-org/matomo/pull/14448.patch

mattab:
documented in new FAQ: [How do I disable tracking failure notification emails?](https://matomo.org/faq/how-to/faq_35802/)

PR #17531, opened 2021-05-07
Send email notifications to a user whenever their settings are changed (2FA, token_auth, user is updated or deleted), and to super users when security settings are changed (brute force logic, 2FA, cors changes)

flamisz:
### Description:

Send email notifications when critical action happened.

Critical actions:
- making any kind of changes to the 2FA -> user 
- if a token auth was created or deleted -> user
- changes to the brute force logic, two FA and CORS were made -> superusers
- when users are created/deleted -> user

### Review

* [ ] Functional review done
* [ ] Code review done


plugins/CoreAdminHome/Emails/TwoFactorAuthDisabledEmail.php
```diff
@@ -0,0 +1,82 @@
...
+        $view = new View('@CoreAdminHome/_securityNotificationEmail.twig');
+        $view->login = $this->login;
+        $view->body = $this->getBody();
+
+        return $view;
+    }
+
+    private function getBody()
+    {
+        return Piwik::translate('CoreAdminHome_SecurityNotificationTwoFactorAuthDisabledBody');
```
sgiehl: If I saw that right, all Mail classes actually have almost the same content. Wouldn't it make sense to have a abstract parent class to avoid duplication? So all child classes would actually only need to define the `getBody()`

plugins/CoreAdminHome/lang/en.json
```diff
@@ -139,6 +139,17 @@
...
+        "SecurityNotificationTwoFactorAuthDisabledBody": "Someone disabled the two-factor authentication in your account. If it was you, carry on. If you don't recognize this activity, please reset your password and check your two-factor authentication app or device.",
+        "SecurityNotificationTokenAuthCreatedBody": "Someone created an auth token in your account. If it was you, carry on. If you don't recognize this activity, please reset your password.",
+        "SecurityNotificationTokenAuthDeletedBody": "Someone deleted an auth token in your account. If it was you, carry on. If you don't recognize this activity, please reset your password.",
+        "SecurityNotificationSettingsChangedByUserBody": "Someone modified the system settings. Affected settings: %s. If it was you, carry on. If you don't recognize this activity, please reset your password.",
+        "SecurityNotificationSettingsChangedByOtherSuperUserBody": "An other super user (%s) modified the system settings. Affected settings: %s.If you know about this activity, carry on. If you don't recognize this activity, please contact the other super users.",
```
sgiehl: Please always use numbered placeholder in translations eg. `%1$s`, `%1$s`,... Otherwise translator aren't able to change the order if that might be needed for proper translation.

plugins/TwoFactorAuth/Controller.php
```diff
@@ -148,6 +152,13 @@ public function disableTwoFactorAuth()
             $this->twoFa->disable2FAforUser(Piwik::getCurrentUserLogin());
             $this->passwordVerify->forgetVerifiedPassword();
 
+            $container = StaticContainer::getContainer();
+            $email = $container->make(TwoFactorAuthDisabledEmail::class, array(
```
sgiehl, line 156: Was there an specific reason for using a container->make instead of instantiating the class directly? Would it that way be possible to change the used mail class using DI?

plugins/TwoFactorAuth/Controller.php
```diff
@@ -148,6 +152,13 @@ public function disableTwoFactorAuth()
...
+                'login' => Piwik::getCurrentUserLogin(),
+                'emailAddress' => Piwik::getCurrentUserEmail()
+            ));
+            $email->send();
```
sgiehl: Just for completeness: If sending the mail fails for any reason this might throw an exception, which is currently uncaught.  It might then show an error message and the user might not be sure if the action was actually performed or not. Not sure if we should maybe better ignore possible exceptions in this case. Same applies for the other mails

plugins/UsersManager/API.php
```diff
@@ -988,6 +997,13 @@ public function deleteUser($userLogin)
         $this->model->deleteUserOptions($userLogin);
         $this->model->deleteUserAccess($userLogin);
 
+        $container = StaticContainer::getContainer();
+        $email = $container->make(UserDeletedEmail::class, array(
+            'login' => Piwik::getCurrentUserLogin(),
+            'emailAddress' => Piwik::getCurrentUserEmail()
+        ));
+        $email->send();
+
```
sgiehl, line 1008: Just out of curiosity. Does it really make sense to email the current user that he just created/deleted a user?
Image someone wants to set up a new Matomo for a bigger company and creates a large number of accounts using the API. He would then be spammed with emails...
Imho it would be more helpful if maybe the superuser, or any defined user would be informed if any user creates/removes a user

sgiehl (changes requested):
Left some comments that came to my mind when looking at the code changes. Haven't done any local tests yet...

Findus23:
One thing: To anticipate an issue like https://github.com/matomo-org/matomo/issues/14447, I guess it would be useful to have an option to disable those E-Mails (even though I find it hard to know where to draw the line)

tsteur:
> One thing: To anticipate an issue like #14447, I guess it would be useful to have an option to disable those E-Mails (even though I find it hard to know where to draw the line)

For MVP this might not be needed and we could wait until this comes up as a request and then someone could even develop a plugin for this etc.

plugins/CoreAdminHome/lang/en.json
flamisz: 👍

plugins/TwoFactorAuth/Controller.php
flamisz: I copied the example @diosmosis shared with me. That's the same reason I didn't use `try..catch` for sending the email.

plugins/TwoFactorAuth/Controller.php
flamisz: I will double check this.

plugins/UsersManager/API.php
flamisz: ping @tsteur

plugins/CoreAdminHome/Emails/TwoFactorAuthDisabledEmail.php
flamisz: One Mail class has some extra attributes but probably still worth it to make an abstract class for these security related mails. I'll see how it looks like.

plugins/UsersManager/API.php
tsteur: Yes it be great if the user got notified. That's because you want to know if someone else created a user using your user in case your login got hacked for example. A superuser might not be able to know if that was on purpose or not. And the emails you receive will be also be useful from an audit log perspective etc. 

Additionally maybe a superuser etc would want to be notified which I think we will be implementing using ActivityLog to be notified by email about certain events.

When many users are created regularly using the API then this might not be wanted but then we could eventually still have settings to disable certain emails (or a group of emails) but that be likely not needed for now and we could add it when needed.

plugins/TwoFactorAuth/Controller.php
flamisz: Now we have the abstract class, what if I add a method like this:
```php
public function safeSend()
{
    try {
        $this->send();
    } catch (\Exception $e) {
        // we do nothing if the email send was unsuccessful
    }
}
```

Any thoughts on this @sgiehl @diosmosis ?

plugins/TwoFactorAuth/Controller.php
diosmosis: This was done in this change: https://github.com/matomo-org/matomo/commit/fcd604b0bc1fdbb4f741a99a58abdc93dd25fd57

It was so plugins could change the contents of the email through DI (I guess we do this on cloud?). I'm not sure if it's needed for these emails (@tsteur would know).

plugins/TwoFactorAuth/Controller.php
diosmosis: It sounds like a good idea to not have exceptions from sending mail like these (which are side effects of other actions) break core functionality. I'm not sure we should just ignore the error though. Can we log it? eg,

```
catch (\Exception $e) {
    StaticContainer::get(LoggerInterface::class)->warning('Could not send whatever email: {exception}', ['exception' => $e]);
}
```

plugins/TwoFactorAuth/Controller.php
tsteur: it be good if this was possible to change emails generally on cloud 👍

plugins/CoreAdminHome/Emails/SecurityNotificationEmail.php
```diff
@@ -75,5 +77,15 @@ protected function getDefaultBodyView()
...
 
+    public function safeSend()
+    {
+        try {
+            $this->send();
+        } catch (\Exception $e) {
+            // we do nothing but log if the email send was unsuccessful
+            StaticContainer::get(LoggerInterface::class)->warning('Could not send {class} email: {exception}', ['class' => get_class($this), 'exception' => $e]);
+        }
+    }
```
diosmosis: I'm not sure if there was a reason for this being here, but personally I think this would definitely be useful in `Mail`.

plugins/CoreAdminHome/Emails/SettingsChangedEmail.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+
+    protected function getBody()
+    {
+        $pluginNameMap = [
+            'Login' => 'Brute Force',
+            'TwoFactorAuth' => 'Two-factor auth',
+            'CoreAdminHome' => 'CORS'
+        ];
```
diosmosis: Should these text files also be translated?

plugins/CoreAdminHome/lang/en.json
```diff
@@ -139,6 +139,17 @@
         "TrackingFailuresEmail1": "This is just to let you know that %s different kinds of tracking failures have occurred in the last days.",
         "TrackingFailuresEmail2": "To view all the failed tracking requests %1$sclick here%2$s.",
         "ImportFromGoogleAnalytics": "Import from Google Analytics",
-        "ImportFromGoogleAnalyticsDescription": "If you have a Google Analytics property and are switching to Matomo, you can import your existing data using the %1$sGoogleAnalyticsImporter%2$s plugin."
+        "ImportFromGoogleAnalyticsDescription": "If you have a Google Analytics property and are switching to Matomo, you can import your existing data using the %1$sGoogleAnalyticsImporter%2$s plugin.",
+        "SecurityNotificationEmailSubject": "Security Notification",
+        "SecurityNotificationRecoveryCodesShowedBody": "Someone viewed the two-factor authentication recovery codes of your account. If it was you, carry on. If you don't recognize this activity, please reset your password and check your two-factor authentication app or device.",
```
diosmosis: Is there a reason "If it was you, carry on. If you don't recognize this activity, please reset your password and check your two-factor authentication app or device." isn't in one translation but all of them? I think we only do this for single sentences where specific words can change, since languages can change the order of words. But to my knowledge sentence order wouldn't. Unless we want translators to be able to change copy?

plugins/CorePluginsAdmin/API.php
```diff
@@ -61,15 +63,47 @@ public function setSystemSettings($settingValues, $passwordConfirmation = false)
 
         $this->settingsMetadata->setPluginSettings($pluginsSettings, $settingValues);
 
+        $sendSettingsChangedNotificationEmail = false;
+        $notifyPluginList = ['Login', 'TwoFactorAuth', 'CoreAdminHome'];
```
diosmosis: Could this and the other map https://github.com/matomo-org/matomo/pull/17531/files#diff-bc00304f49a251eb19602465131174ad2b88956f5c2f2f9e316b08b90a065f78R38 be merged into a single static variable? It could be placed in the SecurityNotificationEmail class so these aren't defined buried in the code.

plugins/CorePluginsAdmin/API.php
```diff
@@ -61,15 +63,47 @@ public function setSystemSettings($settingValues, $passwordConfirmation = false)
...
+
         try {
             foreach ($pluginsSettings as $pluginSetting) {
                 if (!empty($settingValues[$pluginSetting->getPluginName()])) {
                     $pluginSetting->save();
+
+                    $pluginName = $pluginSetting->getPluginName();
+                    if (in_array($pluginSetting->getPluginName(), $notifyPluginList)) {
+                        $sendSettingsChangedNotificationEmail = true;
+                    }
```
diosmosis, line 77: If more than one plugin's settings are saved at a time, we still only send one email, correct? If so, does it make sense to only send one? The user won't be notified of the other changes in that case.

plugins/TwoFactorAuth/Controller.php
```diff
@@ -287,6 +312,14 @@ public function showRecoveryCodes()
             throw new Exception('You have to verify your password first.');
         }
 
+        if (!$regenerateSuccess && !$regenerateError) {
```
diosmosis, line 321: Is this `if` correct? At this point, `$regenerateError` is always false (it doesn't get set to true anywhere above), and it's correct we send the email if regenerating failed?

diosmosis: Looking more, `$regenerateError` is set below this, so I guess either it shouldn't be in this if, or this if should be below the next one?

plugins/CoreAdminHome/Emails/TokenAuthCreatedEmail.php
```diff
@@ -0,0 +1,21 @@
...
+use Piwik\Piwik;
+use Piwik\Plugins\CoreAdminHome\Emails\SecurityNotificationEmail;
+
+class TokenAuthCreatedEmail extends SecurityNotificationEmail
+{
+    protected function getBody()
+    {
+        return Piwik::translate('CoreAdminHome_SecurityNotificationTokenAuthCreatedBody');
+    }
+}
```
diosmosis, line 33: For this email that says:

```
Someone created an auth token in your account. If it was you, carry on. If you don't recognize this activity, please reset your password.
```

Maybe we should include the name of the token and suggest the user delete it also?

plugins/CoreAdminHome/Emails/TwoFactorAuthEnabledEmail.php
```diff
@@ -0,0 +1,21 @@
...
+use Piwik\Piwik;
+use Piwik\Plugins\CoreAdminHome\Emails\SecurityNotificationEmail;
+
+class TwoFactorAuthEnabledEmail extends SecurityNotificationEmail
+{
+    protected function getBody()
+    {
+        return Piwik::translate('CoreAdminHome_SecurityNotificationTwoFactorAuthEnabledBody');
+    }
+}
```
diosmosis, line 21: For this email:

```
Someone enabled the two-factor authentication in your account. If it was you, carry on. If you don't recognize this activity, please reset your password.
```

If someone else enables two factor auth, the user will not be able to login anymore, so resetting their password might not be possible. Perhaps it should be, "email your superuser at ... and request they disable two factor auth and reset your password immediately"?

plugins/CoreAdminHome/Emails/SettingsChangedEmail.php
```diff
@@ -0,0 +1,50 @@
...
+
+        if ($this->superuser) {
+            return Piwik::translate('CoreAdminHome_SecurityNotificationSettingsChangedByOtherSuperUserBody', [$this->superuser, $pluginNameMap[$this->pluginName]]);
+        }
+
+        return Piwik::translate('CoreAdminHome_SecurityNotificationSettingsChangedByUserBody', [$pluginNameMap[$this->pluginName]]);
+    }
+}
```
diosmosis, line 43: For these emails, would it be a good idea to also suggest the user take a look at the system settings in addition to resetting their password?

plugins/CoreAdminHome/Emails/TokenAuthDeletedEmail.php
```diff
@@ -0,0 +1,21 @@
...
+use Piwik\Piwik;
+use Piwik\Plugins\CoreAdminHome\Emails\SecurityNotificationEmail;
+
+class TokenAuthDeletedEmail extends SecurityNotificationEmail
+{
+    protected function getBody()
+    {
+        return Piwik::translate('CoreAdminHome_SecurityNotificationTokenAuthDeletedBody');
+    }
+}
```
diosmosis, line 43: Would it help to include the name of the deleted token auth here so the user knows which one they have to recreate?

plugins/UsersManager/API.php
```diff
@@ -988,6 +997,13 @@ public function deleteUser($userLogin)
         $this->model->deleteUserOptions($userLogin);
         $this->model->deleteUserAccess($userLogin);
 
+        $container = StaticContainer::getContainer();
+        $email = $container->make(UserDeletedEmail::class, array(
+            'login' => Piwik::getCurrentUserLogin(),
+            'emailAddress' => Piwik::getCurrentUserEmail()
+        ));
+        $email->safeSend();
```
diosmosis, line 1007: For these two emails, can we specify the login of the created user? So the superuser knows if which one to look at if the user is not recognized.

diosmosis (changes requested):
Tested locally, everything works. I left a couple code related comments, and a bunch more comments regarding the contents of the email (mostly making them more specific/informative).

plugins/CoreAdminHome/lang/en.json
flamisz: You mean it should have more translations and we link them in the `getBody()` method?

plugins/CorePluginsAdmin/API.php
flamisz: You're right, didn't think about that. I just checked the UI and there we save in groups, but the API actually can handle different groups at once. I'll modify it to send all of it. not just one.

plugins/CoreAdminHome/Emails/SecurityNotificationEmail.php
flamisz: sure, that way we can use it anytime

plugins/CoreAdminHome/Emails/SettingsChangedEmail.php
flamisz: 👍

plugins/TwoFactorAuth/Controller.php
flamisz: yeah, the logic is funny here, because if the `$regenerateSuccess` is false doesn't mean it was an error, we use this method when we show the codes, even when we don't regenerate.
but the `$regenerateError` part is true, probably have to move it after it. I recheck how this method works.

plugins/CoreAdminHome/Emails/TwoFactorAuthEnabledEmail.php
flamisz: should we use the first or all superuser email addresses for this?

diosmosis: Not sure, we could have all but depending on how many there are that could be a lot of emails. With just one, the one superuser might get all of the requests. Do you have an opinion @tsteur?

plugins/CoreAdminHome/lang/en.json
diosmosis: Something like:

```
'EmailBodyMessage' => 'Main message'
'IfItWasYouCarryOn' => 'If it was you, carry on. If you don't recognize this activity, please reset your password and check your two-factor authentication app or device.'
```
in the json files, and:
```
Piwik::translate('EmailBodyMessage') . Piwik::translate('IfItWasYouCarryOn')
```
in the php (or the equivalent in the twig files).

cc @sgiehl in case this is not preferred way.

plugins/CoreAdminHome/Emails/SecurityNotificationEmail.php
```diff
@@ -0,0 +1,85 @@
...
+use Piwik\Mail;
+use Piwik\View;
+use Piwik\Piwik;
+
+abstract class SecurityNotificationEmail extends Mail
+{
+    public static $notifyPluginList = [
+        'Login' => 'Brute Force',
+        'TwoFactorAuth' => 'Two-factor auth',
+        'CoreAdminHome' => 'CORS'
```
flamisz: I can't use the `translate` function here, because it's static. what's your recommendation @diosmosis?

plugins/CoreAdminHome/Emails/TwoFactorAuthEnabledEmail.php
```diff
@@ -0,0 +1,50 @@
...
+        return Piwik::translate('CoreAdminHome_SecurityNotificationTwoFactorAuthEnabledBody', [$email]);
+    }
+
+    private function getASuperUserEmail()
+    {
+        if (count($this->superuserEmails)) {
+            foreach ($this->superuserEmails as $login => $email) {
+                return $email;
+            }
+        }
```
flamisz: here it's possible we send the email to ourselves. should we find the first superuser email which is not us?
what if there is no other superuser? what is the recommendation in the email to the user?

plugins/CoreAdminHome/lang/en.json
```diff
@@ -139,6 +139,24 @@
...
+        "SecurityNotificationRecoveryCodesRegeneratedBody": "Someone generated new recovery codes for the two-factor authentication in your account. If it was you, carry on. If you don't recognize this activity, please reset your password and check your two-factor authentication app or device.",
+        "SecurityNotificationTwoFactorAuthEnabledBody": "Someone enabled the two-factor authentication in your account. If it was you, carry on. If you don't recognize this activity, email your superuser at %s and request they disable two factor auth and reset your password immediately.",
+        "SecurityNotificationTwoFactorAuthEnabledBodyNoEmail": "Someone enabled the two-factor authentication in your account. If it was you, carry on. If you don't recognize this activity...",
```
flamisz: if there is no other superuser, what is the recommendation in the email? what can the user do?

plugins/CorePluginsAdmin/API.php
```diff
@@ -61,15 +64,46 @@ public function setSystemSettings($settingValues, $passwordConfirmation = false)
...
+                    if (in_array($pluginName, array_keys(SecurityNotificationEmail::$notifyPluginList))) {
+                        $sendSettingsChangedNotificationEmailPlugins[] = $pluginName;
+                    }
                 }
             }
         } catch (Exception $e) {
             throw new Exception(Piwik::translate('CoreAdminHome_PluginSettingsSaveFailed'));
         }
+
+        foreach ($sendSettingsChangedNotificationEmailPlugins as $plugin) {
```
flamisz: in this loop, it's possible to send way too many emails, if there are more superusers in the system. do we still want to send on-the-fly? or use a job here?

flamisz (commented):
@diosmosis I updated my pr but added some questions as comments, can you please check those? 
and still waiting for some comments from the other, I will modify those parts after the comment.
thanks

plugins/CoreAdminHome/Emails/SecurityNotificationEmail.php
diosmosis: You could set the values to the translation keys:

```
'Login' => 'CoreAdminHome_BruteForce',
...
```

then call `Piwik::translate($notifyPluginList['Login'])`. It's done this way in quite a few places.

plugins/CoreAdminHome/Emails/TwoFactorAuthEnabledEmail.php
diosmosis: ping @tsteur / @mattab

plugins/CoreAdminHome/lang/en.json
diosmosis: Only way this would happen is if it happened for the only superuser. Not sure what they can do, I think tthey'd have to use the `twofactorauth:disable-2fa-for-user` console command for the user, then change the password. (cc @tsteur)

plugins/CorePluginsAdmin/API.php
diosmosis: We could reduce the number of emails by sending all the plugin names in a single email. (Ie, "There were changes made to the following system settings: X, Y, and Z.")

There could still be an edge case of too many superusers, I suppose. We could just send one email to multiple recipeints. Ie, using `Mail::addTo()` for each superuser. If that results in just one mail() call or SMTP request, I think it might be ok.

Setting up a queue and a task to process it for mails, however, seems like too much effort for this ticket.

(cc @tsteur)

plugins/CoreAdminHome/Emails/SecurityNotificationEmail.php
flamisz: what a great idea, I didn't think about it, thanks 😃

plugins/CorePluginsAdmin/API.php
flamisz: I reduced the number of emails by sending only one that contains all the setting groups that changed and sending one email to superusers using the `addTo()`.

diosmosis:
@tsteur can you look at the comments pinging you above? There are some questions I can't answer.

plugins/CoreAdminHome/Emails/TwoFactorAuthEnabledEmail.php
tsteur: Not sure I understand, it should be sent to the user that enabled the two factor and nobody else?

tsteur: Like you can't enable it for a different user AFAIK?

tsteur: Oh i see now it's for who we mention to contact, I'll have a think

tsteur: Let's keep it simple and do same thing as for password change etc.
 We could reuse the same translation key : `UsersManager_IfThisWasYouPasswordChange`. which says `If this was you, feel free to ignore this email. If this was not you, please contact your Matomo administrator immediately, as your account might have been compromised!` and this would maybe also make the other comments obsolete. 

I wonder if we can document this for future reference somewhere but wouldn't really know where right now. It's really good to have this `SecurityNotificationEmail` base class now maybe this could provide such a method or constant or something to be reused across emails?

plugins/CoreAdminHome/Emails/TwoFactorAuthEnabledEmail.php
flamisz: I modified the email messages, reusing from the password change email.

plugins/CoreAdminHome/Emails/TwoFactorAuthEnabledEmail.php
flamisz: I reused that translation key. I'll think about that method or constant.

plugins/CoreAdminHome/lang/en.json
flamisz: I keep it simple for now, and just use the same message.

bluikko:
I believe that this change breaks install via the ExtraTools. I really wish team Matomo would realize what the time is and #10257 got some attention.

tsteur:
@bluikko I've had a look through the PR and not sure where this could break ExtraTools. Have you created an issue for this problem on the [ExtraTools issue tracker](https://github.com/digitalist-se/extratools/issues)? It may be related to something else or maybe the plugin developer can work around it. Re #10257 we may put some focus on it in the near future but I can't promise anything.

PR #17532, opened 2021-05-07
Fix segment update scripts

sgiehl:
### Description:

Updating Matomo might currently fail, as the update script for 4.3.0-b4 tries to use the `hash` column, which is created in update for 4.3.0-b3. As the migrations are all gathered before they are executed, the hash column does not exist and the query fails.
This also happen for our updater UI test. See https://travis-ci.com/github/matomo-org/matomo/jobs/503575602#L1700

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Updates/4.3.0-b3.php
+++ b/core/Updates/4.3.0-b3.php
@@ -36,13 +36,6 @@ class Updates_4_3_0_b3 extends PiwikUpdates
         $migrations = [];
 
         $migrations[] = $this->migration->db->addColumn('segment', 'hash', 'CHAR(32) NULL', 'definition');
-
-        $segmentTable = Common::prefixTable('segment');
-        $segments = Db::fetchAll("SELECT idsegment, definition FROM $segmentTable WHERE deleted = ?", [0]);
-        foreach ($segments as $segment) {
-            $hash = md5(urldecode($segment['definition']));
-            $migrations[] = $this->migration->db->sql("UPDATE `$segmentTable` SET `hash` = '$hash' WHERE `idsegment` = '{$segment['idsegment']}'");
-        }
         return $migrations;
     }
 
--- a/core/Updates/4.3.0-b4.php
+++ b/core/Updates/4.3.0-b4.php
@@ -36,7 +36,7 @@ class Updates_4_3_0_b4 extends PiwikUpdates
         $migrations = [];
 
         $segmentTable = Common::prefixTable('segment');
-        $segments = Db::fetchAll("SELECT idsegment, hash, definition FROM $segmentTable");
+        $segments = Db::fetchAll("SELECT * FROM $segmentTable");
         foreach ($segments as $segment) {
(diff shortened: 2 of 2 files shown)


core/Updates/4.3.0-b3.php
```diff
@@ -36,13 +36,6 @@ public function getMigrations(Updater $updater)
         $migrations = [];
 
         $migrations[] = $this->migration->db->addColumn('segment', 'hash', 'CHAR(32) NULL', 'definition');
-
-        $segmentTable = Common::prefixTable('segment');
-        $segments = Db::fetchAll("SELECT idsegment, definition FROM $segmentTable WHERE deleted = ?", [0]);
-        foreach ($segments as $segment) {
-            $hash = md5(urldecode($segment['definition']));
-            $migrations[] = $this->migration->db->sql("UPDATE `$segmentTable` SET `hash` = '$hash' WHERE `idsegment` = '{$segment['idsegment']}'");
-        }
```
sgiehl, line 45: Removing this one, as otherwise all segments might be updated twice when updating from a version prior 4.3.0-b3

tsteur:
@sgiehl not fully sure I understand the problem. Updating it twice be fine I suppose as it's quick etc and this way people that copy/paste the queries and execute them manually could do this (although it should then be executed again anyway hopefully). Curious though how came it failed cause the hash column should have been added before the update query was executed. Was this not the case? Just checking in case there's a bug somewhere

diosmosis:
@tsteur the hash column was selected outside of a migration, so the migration that added the column would get executed after the query that used it was executed. This was the best fix I think.

tsteur:
This would explain this part  https://github.com/matomo-org/matomo/pull/17532/files#diff-c21583e394fbddcd7c201f880290f2b9756fbfbcce383765db917b728a06fde0L39  I think 👍 

but maybe not this part? https://github.com/matomo-org/matomo/pull/17532/files#diff-acfc32e315ad85ad5c07ba2653c366883113e0a91043c2095d380a18fafaf5a7L40-L45 would have expected this to work?

diosmosis:
I assume it was just a waste to have both sets of queries since they do the same thing.

sgiehl:
> I assume it was just a waste to have both sets of queries since they do the same thing.

Exactly

tsteur:
In that case it was to show the updates it would run and then make sure it will be executed just in case either step will be missed in which case it could lead to data deletion of reports that may not be possible to be recreated etc. I reckon it could be good to add it back just to be safe as it's a very critical update that if it goes wrong reports get deleted and then if raw data deletion is configured as well the reports are gone forever or they would need to invalidate data which isn't easy either and can take long time to rearchive data.

sgiehl:
Still don't get why it would be better to execute the same sql queries twice. I have removed the one in the earlier update script. So even if someone already was on `4.3.0-b3`, the updates will be performed again if needed 🤷

PR #17538, opened 2021-05-10
Add option to skip log purging check when invalidating archives in the rare cases it is needed.

diosmosis:
### Description:

For use in the GoogleAnalyticsImporter, if log purging is enabled, then we'll never be able to invalidate to dates before it. However, the importer can definitely import dates before it and uses archiving to fill up week, month and year periods. The flag would fix this.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -265,7 +265,7 @@ class ArchiveInvalidator
      * @throws \Exception
      */
     public function markArchivesAsInvalidated(array $idSites, array $dates, $period, Segment $segment = null, $cascadeDown = false,
-                                              $forceInvalidateNonexistantRanges = false, $name = null)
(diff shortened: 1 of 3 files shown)


core/CliMulti/Process.php
```diff
@@ -297,6 +297,9 @@ public static function getRunningProcesses()
      {
          $ids = explode("\n", trim(`ps x 2>/dev/null | awk '! /defunct/ {print $1}' 2>/dev/null`));
 
+         print getmypid() . "\n";
+         print `ps x 2>/dev/null | awk '! /defunct/ {print $1}' 2>/dev/null` . "\n";
+
```
sgiehl: Did you add that on purpose?

diosmosis: No, task switching w/ an L3 issue, will remove

PR #17539, opened 2021-05-10
Use Common::safe_unserialize for session data

sgiehl:
### Review

* [ ] Functional review done
* [ ] Code review done


flamisz (commented):
it's looking good to me, but should we add this modification into this file?
https://github.com/matomo-org/matomo/blob/4.x-dev/libs/README.md

sgiehl:
@flamisz Guess that file is actually already very outdated. I've updated it a bit. But guess we don't need to mention the change in this PR. The unserialize/serialize of session data was actually something we added to get notifications working, which is already mentioned...

flamisz (approved):

PR #17546, opened 2021-05-11
do not use login detection bruteforcedetection code during oneClickResults action

diosmosis:
### Description:

The one click update feature updates the code, then on the next page view the db upgrade screen is shown. The actions in between these two components use new code, but an old database schema. This PR makes sure we ignore unknown column errors in the bruteforce detection feature if we're in the one click update process.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Login/Security/BruteForceDetection.php
+++ b/plugins/Login/Security/BruteForceDetection.php
@@ -70,7 +70,11 @@ class BruteForceDetection {
     {
         $now = $this->getNow()->getDatetime();
         $db = Db::get();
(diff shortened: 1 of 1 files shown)


flamisz (approved):
I couldn't test it locally (didn't know how to get into this state) but the code looks ok to me, I see how it works and how is it useful. 👍

diosmosis:
@flamisz just fyi, so you're aware if you have to touch this code: this is for the oneclickupdate feature which is extremely hard to test locally. The feature downloads a package from builds.matomo.org, unzips it and replaces the current version of matomo with it. To test it locally w/o a build on builds.matomo.org you'd have to use the matomo-package repo to build a package from 4.x-dev code, then create a new release channel to get it from localhost, then initiate the process. Personally, I find using the test on travis to be a lot less trouble.

PR #17547, opened 2021-05-11
If no archives names are requested when querying archive data, do not initiate archiving.

diosmosis:
### Description:

This solves a very convoluted bug involving CustomReports:

- user creates a custom report with just unique visitors and places it in the visitors overview
- user views a large range (eg, 2021-01-01,2021-03-31)
- the first request on the page, for visits summary data succeeds creating plugin specific range archives with data
- the custom report request is triggered
- the custom report data is successfully requested
- since its for a range, CustomReports decides we're not allowed to archive unique visitors (because of an INI config setting) and removes it from the columns_to_display
- Visualization tries to get the site summary to build totals data
- this issues an API request to API.get for a range, using columns_to_display, which is empty
- API.get issues a request to VisitsSummary.get w/ 0 columns
- VisitsSummary.get creates an Archive instance and calls getDataTableFromNumeric() with an empty array for $names
- since it's a range, Archive.php tries to launch archiving. since no data is requested, there is no specific plugin to load, so we assume it's the all plugins archive. so we create a 'done' archive.
- PluginsArchiver does nothing, since there is nothing being requested
- then we finalize the 'done' archive w/ the DONE_OK flag
- now on the next request, all archive requests to got the new 'done' archive for the range, even though it contains nothing
- the user sees 0 for all metrics for the range

This is fixed in this PR by not launching archiving if nothing is requested in Archive.php.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Archive.php
+++ b/core/Archive.php
@@ -493,6 +493,9 @@ class Archive implements ArchiveQuery
 
         $result = new Archive\DataCollection(
             $dataNames, $archiveDataType, $this->params->getIdSites(), $this->params->getPeriods(), $this->params->getSegment(), $defaultRow = null);
+        if (empty($archiveNames)) {
+            return $result; // NOTE: note posting Archive.noArchivedData here, because there might be archive data, someone just requested nothing
+        }
 
         $archiveIds = $this->getArchiveIds($archiveNames);
         if (empty($archiveIds)) {
--- a/core/Archive.php
+++ b/core/Archive.php
@@ -476,6 +476,8 @@ class Archive implements ArchiveQuery
             $archiveNames = array($archiveNames);
         }
 
+        $archiveNames = array_filter($archiveNames);
+
         // apply idSubtable
         if ($idSubtable !== null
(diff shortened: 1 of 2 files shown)


core/Archive.php
```diff
@@ -493,6 +493,9 @@ protected function get($archiveNames, $archiveDataType, $idSubtable = null)
 
         $result = new Archive\DataCollection(
             $dataNames, $archiveDataType, $this->params->getIdSites(), $this->params->getPeriods(), $this->params->getSegment(), $defaultRow = null);
+        if (empty($archiveNames)) {
```
tsteur: should that be dataNames? if not, could maybe exit right in the top? Also what if `archiveNames` is eg `null` or `false` would then this still be detected? or maybe not if it is `[null]`?

diosmosis: It could be either. We need to return an empty DataCollection value so the result is properly created (eg, for DataTables we need correct metadata).

I'll add extra code for empty values.

PR #17554, opened 2021-05-13
Do not send idGoal=0 parameter for goals overview since 0 is for orders

### LINKED ISSUE #17502: Email report: Goals Overview is empty
When I configure an Email report with some goals to report, so I have a list of checkboxes.

If I now select the checkboxes "Goals overview" - marked red, see screenshot - , there are no results in the report. even though goals have been achieved.
grafik

This are the results in my report (downloaded report):
I have 14 conversions, but Goals overview is empty

grafik

## Environment

* Matomo Version: 4.2.1

diosmosis:
### Description:

Fixes #17502

In the past the UI controller method that outputted the goals overview used idGoal=0 to signify it was handling the overview. It would not pass that parameter on to the API, since the API treats idGoal=0 as IDGOAL_ORDER.

This means, currently, we are displaying order goal information in place of the goals overview for scheduled reports. To fix this, we can just not send the idGoal to the API.

The tests are failing, most seem like they are expected. The TSV test however changes the name of the report in the output (though this doesn't happen locally for me).

cc @tsteur for prioritization

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Goals/Reports/Base.php
+++ b/plugins/Goals/Reports/Base.php
@@ -45,7 +45,7 @@ abstract class Base extends \Piwik\Plugin\Report
         } else {
             $this->name = $goalNameFormatter(['name' => Piwik::translate('Goals_GoalsOverview')]);
         }
-        $this->parameters = ['idGoal' => 0];
+        $this->parameters = [];
         $this->order = $this->orderGoal;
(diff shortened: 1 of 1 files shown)


tsteur:
Will close this one as it's not high priority and issue is not scheduled.

PR #17566, opened 2021-05-16
use speak: never

### LINKED ISSUE #17565: CSS: "speak" property: "none" changed to "never"
grafik
link: https://demo.matomo.cloud/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=day&date=yesterday&segment=&category=Dashboard_Dashboard&subcategory=1

## Expected Behavior
`speak: never`

## Current Behavior
`speak: none`

## Possible Solution
`speak: never`
see https://www.w3.org/TR/css-speech-1/
grafik

see above

## Your Environment
* Matomo Version: 4.2.1 (demo.matomo.cloud)
* Browser: Firefox
* Operating System: windows 10

Findus23:
fixes #17565

seems like this was changed some time ago (`https://github.com/w3c/csswg-drafts/issues/510`)

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Live/stylesheets/live.less
+++ b/plugins/Live/stylesheets/live.less
@@ -226,7 +226,7 @@ ol.actionList > li:not(.pageviewActions) {
 
       // copied frim [class^=icon-] style
       font-family: 'matomo' !important;
-      speak: none;
+      speak: never;
       font-style: normal;
       font-weight: normal;
       font-variant: normal;
--- a/plugins/Morpheus/stylesheets/base/icons.css
+++ b/plugins/Morpheus/stylesheets/base/icons.css
@@ -13,7 +13,7 @@
 [class^="icon-"], [class*=" icon-"] {
   /* use !important to prevent issues with browser extensions that change fonts */
   font-family: 'matomo' !important;
-  speak: none;
+  speak: never;
   font-style: normal;
   font-weight: normal;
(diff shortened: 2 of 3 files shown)


diosmosis:
Is this something other browsers might still implement as `none`? I'm wondering if we should have both just in case.

Findus23:
@diosmosis I was wondering the same thing, but I can't find any information about this. But it seems like `none` was never in a finished standard, so no browser should have implemented it (at least in theory): `https://github.com/w3c/css-validator/issues/231`

sgiehl:
When I add `speak: never` to an element my Chrome reports it as invalid. The computed styles then show `speak: normal`.
`speak: none` would be valid in Chrome.
Firefox actually does not know the property `speak` and reports all values as invalid.

Findus23:
@sgiehl Indeed it seems to be the same for me. I'm not even sure if this is even implemented in a browser and does anything. Using `aria-hidden` on icons seems to be the better solution for accesibility anyway.

sgiehl:
So shall we simply close this one? Or should we add both (`none` and `never`), so the browser can decide which one to pick?

Findus23:
As `none` is the value chromium suggests and I think everyone else not supporting this at all, the best thing is probably to just close this.

PR #17569, opened 2021-05-16
Make sure summary row subtables are included in the serialized representaion of a datatable tree.

diosmosis:
### Description:

If a summary row has a subtable, it will not be serialized because we go over the `$rows` property, but the summary row is held in a different property. Fixed by including it.

This was causing the strange "Row with label '%s' (columns = %s) has already a subtable" errors. The summary row subtable couldn't be loaded from the DB, just the ID, so when summing them, we'd trigger this warning.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/DataTable.php
+++ b/core/DataTable.php
@@ -1317,10 +1317,10 @@ class DataTable implements DataTableInterface, \IteratorAggregate, \ArrayAccess
         $consecutiveSubtableIds = array();
         $forcedId = $subtableId;
 
-        // For each row, get the serialized row
+        // For each row (including the summary row), get the serialized row
         // If it is associated to a sub table, get the serialized table recursively ;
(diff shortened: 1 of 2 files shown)


flamisz (dismissed):
The code and test look good to me but didn't know where to test in the UI.

flamisz:
As I see in the tests there is an error.

diosmosis:
The test failure seems unrelated. This is for an L3 issue in JIRA so the description of the bug it fixes is hidden, and it would be pretty hard to manually test. You'd have to set the ranking query INI config really low, track a ton of data w/, say, events that creates top level reports w/ subtables (including in the Others row), then archive w/ ranking query. I couldn't actually get that to work, so just debugged on prod.

flamisz (approved):

Issue #17035, opened 2020-12-31
don't try to insert unrealistically high values for time_transfer and other performance metrics


Findus23:
see https://forum.matomo.org/t/512m-of-memory-limit-is-it-too-little/40047/5

No page takes 50 years to load, so I guess Matomo should not try to insert invalid values.

> [28-Dec-2020 07:09:09 UTC] Error in Matomo (tracker): Error query: SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'time_transfer' at row 1 In query: INSERT INTO matomo2_log_link_visit_action (idvisit, idsite, idvisitor, idaction_url, idaction_url_ref, idaction_name_ref, server_time, idpageview, pageview_position, time_spent_ref_action, time_dom_processing, time_network, time_server, time_transfer, idaction_name) VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?) Parameters: array (   0 => '620729',   1 => 23,   2 => '&� ��V�	',   3 => '122626',   4 => '122626',   5 => '122625',   6 => '2020-12-28 07:09:09',   7 => 'qDEDCT',   8 => 3,   9 => 185,   10 => 874,   11 => 2180,   12 => 283,   13 => 1608938301263,   14 => '122625', )

Then again it would be interesting to know where they come from.

poetaster:
I'm getting a similar error on field visitor_seconds_since_order.
`Error query: SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'visitor_seconds_since_order' at row 1 In query: INSERT INTO piwik_log_visit (idvisitor, config_id, location_ip, idsite, profilable, visit_first_action_time, visit_goal_buyer, visit_goal_converted, visit_last_action_time, visitor_returning, visitor_seconds_since_first, visitor_seconds_since_order, visitor_count_visits, visit_entry_idaction_name, visit_entry_idaction_url, visit_exit_idaction_name, visit_exit_idaction_url, visit_total_actions, visit_total_interactions, visit_total_searches, referer_keyword, referer_name, referer_type, referer_url, location_browser_lang, config_browser_engine, config_browser_name, config_browser_version, config_client_type, config_device_brand, config_device_model, config_device_type, config_os, config_os_version, visit_total_events, visitor_localtime, visitor_seconds_since_last, config_resolution, config_cookie, config_flash, config_ja...'`

[...]

PR #17574, opened 2021-05-17
Ignore performance metrics that are obviously too high

sgiehl:
### Description:

Added a check for one hour for now. Might be even fine to check for 10 minutes or so, as no website performance metric should contain a number higher that this. 

refs #17035

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/PagePerformance/Columns/Base.php
+++ b/plugins/PagePerformance/Columns/Base.php
@@ -37,6 +37,12 @@ abstract class Base extends ActionDimension
             throw new InvalidRequestParameterException(sprintf('Value for %1$s can\'t be negative.', $this->getRequestParam()));
         }
 
+        // ignore obviously incorrect values (nothing should take more than 1 hour to load)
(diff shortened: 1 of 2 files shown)


plugins/PagePerformance/Columns/Base.php
```diff
@@ -37,6 +37,12 @@ public function onNewAction(Request $request, Visitor $visitor, Action $action)
             throw new InvalidRequestParameterException(sprintf('Value for %1$s can\'t be negative.', $this->getRequestParam()));
         }
 
+        // ignore obviously incorrect values (nothing should take more than 1 hour to load)
+        // refs https://github.com/matomo-org/matomo/issues/17035
+        if ($time > 3600000) {
```
tsteur: @sgiehl since we don't know what people are tracking it be better to go with some max value like 16million since people might be tracking apps, or videos, or all kind of things using such parameters.

sgiehl: ok. changed hat.

PR #17594, opened 2021-05-19
Adds automatic release action

### LINKED ISSUE #119: Automate the packaging of Matomo beta/rc/stable releases via Github actions
To build a release we usually follow these steps:

* create new Github release (which automatically creates a tag)
* run `./scripts/build-package.sh 4.0.0-b3`
* the release is signed with my key so it asks for passphrase
* build finishes with success or error

Eventually, there's no way around it: we will need to automate the build of Matomo releases. It would be fantastic if this was completely automated for example via Github actions. I was thinking we could create a new repo for example `matomo-releases` or so, that only release managers would have access to (Matt, Thomas, Stefan, etc.). Then the Github actions would run in this repo and automate the steps above. We'd need to create a new public/private key pair to sign the build, that only release managers would have access to (and the Github action also). 

This would help a lot to let the team release versions without my involvement. And to save me some time. cc @tsteur @sgiehl

sgiehl:
This PR aims to set up a automatic build process.

## What this action does:

This action always needs to be triggered manually in the actions tab. This is possible with two purposes:

### New release in the current major version

When a new release has to be done this action simply needs to be triggered manually, without setting a version.
The action will then:

* Check if the user who triggered the action is in the team `matomo-org/release-team`
  --> If this is not the case the action will abort with an error
* The version contained in the file `core/Version.php` will be extracted and used
  --> If a tag with the determined version already exists the action will abort with an error
* A new tag for that version will be created
* A new build will be created using the script `.github/scripts/build-package.sh`
  --> This will create piwik and matomo release archives in `zip` and `tar.gz` as well as signature files ending on `.asc`
* In addition to this a new release will be created in the repository and the matomo release archives and signatures will be attached. Depending on the version this will be automatically marked as pre-release or not. The release message for each release type can be adjusted if needed. The result can be seen in my fork: [Pre-Release](https://github.com/sgiehl/matomo/releases/tag/4.3.0-b3) | [Normal release](https://github.com/sgiehl/matomo/releases/tag/4.3.0)

### Release in LTS version (or re-release of a broken one)

When releasing a LTS version we can't simply trigger the action. Instead we need to create the tag manually first. After this has been done, the actions needs to be triggered while providing the tagged version number in the according input:
image

The action will then:

* Check if the user who triggered the action is in the team `matomo-org/release-team`
  --> If this is not the case the action will abort with an error
* The action now checks if a tag for the given version exists
  --> If no tag exists the action will abort with an error
* A new build will be created using the script `.github/scripts/build-package.sh`
  --> This will create piwik and matomo release archives in `zip` and `tar.gz` as well as signature files ending on `.asc`
* As above a new release in the repo will be created for the tag

Additionally the action could also be used to re-release a "broken" release. Even though this should be an uncommon use case, it's possible to simply trigger the action with a version number that already exists. For pre-releases the script should simply run through and recreate the archives. They should then be replaced in the repo release. For stable releases this currently won't work as the build script aborts if the [version is already available on our build server](https://github.com/matomo-org/matomo/pull/17594/files#diff-7c93cffa01636081de4bcfa2ceb0ed68ef487e6fca1d29b4c63fe246ad683acdR169-R177)

## Requirements:

The team `matomo-org/release-team` needs to be filled with the persons who should be able to trigger this action. _This is currently a private group, so no one can directly see who actually has the permission_

The action requires the following repository tokens to be set up:

| secret name | secret value description |
| ---- | ----|
| GPG_CERTIFICATE | ASCII armored or Base64 encoded GPG certificate that is used to create the signatures for the archives |
| GPG_CERTIFICATE_PASS | Passphrase of the GPG key |
| RELEASE_PASSWORD | Password required to trigger the release action |

## Testing / Feedback

If someone wants to test the process feel free to get in touch on slack. It's set up on my fork and uses a slightly adjusted build script that uploads the releases to another account. Can give you access...

---------

fixes https://github.com/matomo-org/matomo-package/issues/119


mattab:
Great to see this progress! 

Btw another important step to complete at the end will be to update the instructions at: https://matomo.org/blog/2014/11/verify-signatures-piwik-packages/

flamisz:
Is it ready to review? I see the `needs review` label but it's still in draft.

sgiehl:
@flamisz That's because it requires some setup before it could be merged. It's also only a suggestion from my side how the process could work. So it has the needs review and rfc label actually to get some feedback.

sgiehl:
@mattab @tsteur would be awesome if you can find some time to look into this, so we maybe could set up the automatic release soon.

mattab:
From my end it looks fantastic @sgiehl - great work! Will save us all time and make our release process faster :rocket:

tsteur:
@sgiehl what's the status here? can we configure (probably requires @mattab for some keys) and merge and test it?

sgiehl:
@tsteur Not sure if someone wants to do a code review. Otherwise we can configure it

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+            Please use the attached archives for installing or updating Matomo.
+            The source code download is only meant for developers and will require extra work to install it. 
+             - Latest stable production release can be found at https://matomo.org/download/ ([learn more](https://matomo.org/docs/installation/)) (recommended)
+             - Beta and Release Candidates releases can be found at https://builds.matomo.org/ ([learn more](http://matomo.org/faq/how-to-update/faq_159/))"
+          else
+            echo ::set-output name=prerelease::false
+            body="## [Matomo ${version} Changelog](https://matomo.org/changelog/matomo-${version//./-}/)
+            
+            Please use the attached archives for installing or updating Matomo.
+            The source code download is only meant for developers and will require extra work to install it.
```
tsteur: could also mention "requires extra work to install it and is not secure to run in a production environment." 

Same above

tsteur: technically it should also not run in any staging environment or so. might need to word it differently.

tsteur:
👍  wasn't sure as it was set to draft. Generally had a look over it and looked ok-ish but don't know the details. I guess we'd mostly want to test it to ensure it works. Maybe in beginning we wouldn't test it with ssh piwik-builds but some other account.

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+          # DOKKU_SSH_KEY is the name of the repository secret
+          echo "${{ secrets.SSH_KEY }}" > /home/runner/.ssh/github_actions
+          chmod 600 /home/runner/.ssh/github_actions
+          ssh-agent -a $SSH_AUTH_SOCK > /dev/null   
+          ssh-add /home/runner/.ssh/github_actions
+    - name: Check preconditions, create tag, build and publish release
+      id: tag
+      env:
+        SSH_AUTH_SOCK: /tmp/ssh_agent.sock
+      run: |
```
Findus23: I think this script should really have a set -e as it should not rather fail than continue after a failing step.

sgiehl: This is done internally. See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+          body="${body//$'\n'/'%0A'}"
+          body="${body//$'\r'/'%0D'}"
+          echo ::set-output name=version::$version
+          echo ::set-output name=body::$body
+
+          git clone https://github.com/matomo-org/matomo-package.git matomo-package
+          cd matomo-package
+          ./scripts/build-package.sh $version
+      shell: bash 
+    - uses: ncipollo/release-action@v1
```
tsteur: can we use https://github.com/github/hub maybe instead of this action?

Like

`hub release create --message $VERSION --message "If…`

Not sure if we need "allowUpdates etc"

Otherwise should pin the action to a specific commit.

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+           echo "Action was not triggered by a member with release permission"
+           exit 1;
+          fi
+      shell: bash
+    - uses: actions/checkout@v2
+      with:
+        lfs: false
+    - name: Import GPG key
+      id: import_gpg
+      uses: crazy-max/ghaction-import-gpg@v3
```
tsteur: same here let's point to a specific commit SHA.

Or can we replace this with a regular command so we don't need to use this action?

The `with` does this mean their script gets our private key and passphrase?

.github/workflows/release.yml
```diff
@@ -0,0 +1,143 @@
...
+          body="${body//$'\n'/'%0A'}"
+          body="${body//$'\r'/'%0D'}"
+          echo ::set-output name=version::$version
+          echo ::set-output name=body::$body
+
+          git clone https://github.com/matomo-org/matomo-package.git matomo-package
+          cd matomo-package
+          ./scripts/build-package.sh $version
+      shell: bash 
+    - uses: ncipollo/release-action@880be3d0a71bc0fa98db60201d2cbdc27324f547
```
tsteur: can we use https://github.com/github/hub maybe instead of this action?

Like

`hub release create --message $VERSION --message "If…`

Not sure if we need "allowUpdates etc"

Otherwise should pin the action to a specific commit.

mattab:
it would be great to work on this soon, as it will help us to introduce another release manager, so we can release more often and regularly.

.github/workflows/release.yml
sgiehl: `hub release` would only create a tag & release. The action additionally uploads the zip/tar.gz file to github, so they will be attached as release assets. Seems that is not possible using hub.

.github/workflows/release.yml
sgiehl: Correct. Their script gets the private key and passphrase in order to set up GPG in a way so our script can use the key without the need to enter the password.

I'm pretty sure we could replace that with bash commands. But I don't know how to achieve that. Tried that when setting up this action for a while, but failed and then decided to look for an action that is able to achieve that.

If someone is keen to replace that with proper bash commands, feel free to give that a try.

tsteur: Looking at https://github.com/crazy-max/ghaction-import-gpg/blob/master/src/gpg.ts#L112 it executes `gpg --import --batch --yes $keyFile`?  (saw something similar in https://superuser.com/questions/1135812/gpg2-asking-for-passphrase-when-importing-secret-keys )

re the passphrase. It seems to be setting up an agent in https://github.com/crazy-max/ghaction-import-gpg/blob/master/src/main.ts#L41-L50 ? by executing https://github.com/crazy-max/ghaction-import-gpg/blob/master/src/gpg.ts#L35 eg if I see this right something along those lines:

```
echo "default-cache-ttl 7200
max-cache-ttl 31536000
allow-preset-passphrase" > $HOME/.gnupg/gpg-agent.conf
gpg-connect-agent "RELOADAGENT" /bye
gpg-connect-agent "PRESET_PASSPHRASE ${keygrip} -1 ${hexPassphrase}" /bye
gpg-connect-agent "KEYINFO ${keygrip}" /bye
```

Maybe this helps to find otherwise some further instructions/help for this?

tsteur: seeing also eg https://stackoverflow.com/questions/49491679/how-to-enter-gnupg-agent-key-passhprase-from-cli it seems important to allow preseting the passphrase and then use above commands to preset it

sgiehl: can give that a try if you think forcing the action to a specific commit isn't enough. But might take a while to figure out the exact commands to get that working fully

tsteur: I suppose the alternative would be doing a review of the code to make sure there's nothing suspicious done. I suppose technically someone could [brute force a specific commit hash](https://github.com/bradfitz/gitbrute) but it's unlikely. Could maybe time box to 60 minutes to make it work locally since above commands should pretty much do what they should and otherwise do a quick review of the other action and use a specific hash 👍

mattab:
@sgiehl What are the remaining steps to finish this project? hoping to introduce another release manager, so we can release more often and regularly :rocket:

sgiehl:
@mattab I need to find some time looking into or replacing the usage of the external actions we would currently use. Afterwards we need to configure the secrets in GitHub as defined in the PR description create the new GitHub group for release managers and merge this PR.
We could maybe first test it with a SSH key that actually uploads the packages somewhere else.

mattab:
@sgiehl Just curious (it's not urgent), when do you think you could work on this project (in how many weeks)? (after your current priorities of Hits + ongoing PRs reviews)

sgiehl:
I've started to resume the work on that one and got some questions/notes about the future process I would like to clarify.

- Currently we are using the `build_package.sh` in the other repo to build releases. To reduce the complexity a bit I have moved it into the `.github` folder in this repo. That way I can also adjust it a bit more to match what is really needed for the action
- Currently the build script uses a ssh key to connect to our build server in order to upload the releases. This works smoothly. Nevertheless I was wondering if we actually want to store a valid ssh key that allows connecting to our server as a github secret.
An alternative would be to create a script on our server that can be triggered via url, which would then download the release packages attached to the github release and store it on the server. That way the risk of misuse would be lowered, but it will for sure take some extra effort to develop it.
- Also if you have any wishes how to change / improve the build process (described in the PR description) please let me know.

@tsteur @mattab @justinvelluppillai your feedback would be welcome

tsteur:
> An alternative would be to create a script on our server that can be triggered via url, which would then download the release packages attached to the github release and store it on the server. That way the risk of misuse would be lowered, but it will for sure take some extra effort to develop it.

I guess it might be fairly easily developed? We would also want to have a very secure password there. It then gets quickly complicated as you'd want technically then brute force prevention etc but if it only downloads the script from Github then it may be fine and not needed as much. 

The risk is that someone could trigger a release of some random branch etc and then get the server to download it. But at least it wouldn't allow to change the latest version number and not any other packages assuming it's all written correctly. This risk exists though with either solution. Does the github action already enforce that we only allow a release RC and production releases from `next_release` and beta releases additionally from `4.x-dev`? I've now just set up a branch protection rule for `next_release` so there's no force push possible. I could also require "PRs" and approvals of at least one person for `next_release`. It's not done yet, but could do (not done yet) 
image


The zip is only one thing thought right? There's also the version number that needs to be updated on api.matomo.org?

Would it overall maybe easier to simply/better setting up a webhook? We can define a secret etc. I haven't thought too much about it though.I suppose we want to attach the release to the github release but maybe that's not needed? Just some thoughts... There are probably some downsides to it. Mostly that you then want some harder brute force protection maybe etc. Overall might be more complicated.

tsteur:
would have been nice if github allowed configuring creating releases only from specific branches. doesn't seem to be the case if I see this right

sgiehl:
I'll leave it as is for now, so the upload will be done using an ssh key. We can also start with that approach and improve/harden it when it's already running. Will now check for other improvements to finish the PR.

> Does the github action already enforce that we only allow a release RC and production releases from next_release and beta releases additionally from 4.x-dev?

No that is not enforced yet. It's possible to trigger the action for a any branch (where the action exists). On that branch a tag will be created. But I can add some checks, so this is only possible for `next_release` or `4.x-dev`. Btw. it will still be possible to create a tag manually on any branch and trigger the action providing that tag to build it. (This will be useful for building LTS version or if we ever need to release a fix for an older version or rebuilt one.)

> I've now just set up a branch protection rule for next_release so there's no force push possible. I could also require "PRs" and approvals of at least one person for next_release. It's not done yet, but could do (not done yet)

That would make sense to me.

> I suppose we want to attach the release to the github release but maybe that's not needed?

I guess we hadn't done that before, as it would have been extra effort to do that. But with a github action it can be done automatically, so I can't see any downside in doing that. Besides the fact that we no longer might have proper download stats (unless a github webhook is also triggered for downloads)

tsteur:
> No that is not enforced yet. It's possible to trigger the action for a any branch (where the action exists). On that branch a tag will be created. But I can add some checks, so this is only possible for next_release or 4.x-dev. Btw. it will still be possible to create a tag manually on any branch and trigger the action providing that tag to build it. (This will be useful for building LTS version or if we ever need to release a fix for an older version or rebuilt one.)

For security it would be great to only allow `\d.x-dev` and `next_release` for this for security as these branches are protected. Technically could even only allow `next_release` for even better security and require a comment or someone to confirm the PR.

Overall downloading the release from github might be then more secure then using SSH I reckon. As in worst case nobody can change historical releases and zips etc.

mattab:
+1 for `create a script on our server that can be triggered via url, which would then download the release packages attached to the github release and store it on the server. That way the risk of misuse would be lowered` and any other "safeguards" we can put in to protect releases

justinvelluppillai:
@sgiehl be good to prioritise this to see if we can use it to release 4.7.0 soon. Is there anything you're waiting on to proceed?

sgiehl:
Just to give an update on the current state:

The build script is currently working as described in the PR description. Theoretically it could be set up and it would already be possible to use it for building releases.

I've today managed to replace the use of the action `crazy-max/ghaction-import-gpg` with custom commands.

Last point would be to replace the mechanism to upload the releases to the build server with a server side script that is triggered from the action. Not yet sure how to solve that, will check tomorrow.

justinvelluppillai:
Thanks for the update @sgiehl. It would be good to do a test run of this, could we try with a beta release later this month?

sgiehl:
@mattab @tsteur Is it required to change the mechanism from uploading files to the server(s) to triggering a script on the server(s) that downloads the builds instead?

It's a way more complex than I initially thought. Currently we are uploading the releases as `matomo-X.zip` and `piwik-X.zip` to the server. Depending on the released version the version files like `LATEST_STABLE`, `LATEST_BETA`, ... are updated as well. That is not only done on the builds server, but the latter files are updated on API and WWW as well.
Looking at the code it seems at least the update on WWW can't work. It's done [here](https://github.com/matomo-org/matomo-package/blob/6d79db55b295453938e299706b3e66c663f93e49/scripts/build-package.sh#L655-L658), but actually `$WWW_PATH` is not defined and so it would try to upload to `/`, which I guess won't work.
Also I wonder if the certain updates to the API server are needed or if wouldn't simply create all those version related files on the build server and let the API server access those.
If not we would either need to have a separate script on the API server that is triggered while releasing or we would need to ssh from the build server to the api server to perform the updates that way.

Also I personally would only attach `matomo.zip` (and `.tar.gz`) to the github releases, as I don't see any advantage in also attaching a `piwik.zip` there. But if we download the release files from github, we would need to attach it or decide to not provide `piwik.zip` on the release server anymore (guess this might cause problems when updating from an older version).

Any thoughts?

tsteur:
@sgiehl I'm not sure I fully understand re the problem. Generally whether it is `piwik.zip` or `matomo.zip` shouldn't really matter as we can just create the needed files on demand or am I missing something? Whatever script downloads these files could just create the needed file names and copy it over? 

Re www path and API: we could use an API endpoint there as well to update these files or still use `ssh` there.

I'm thinking for `WWW_PATH`: We could refactor matomo.org to simply get the right data from builds.matomo.org. This should be a very trivial change of just changing the file path/ url (but didn't look). Or it could be a cron that syncs it with builds.matomo.org regularly like `wget https://builds.matomo.org/LATEST...`

Maybe on api.matomo.org we could have a cronjob that updates local `LATEST_` files as well and copies them over from builds.matomo.org? We would just need to make sure that in future we also have all the entries like `LATEST_2X` and `LATEST_2x_BETA` etc on builds.matomo.org as well.

Not sure if there be any issue with that? Maybe changing matomo.org and the api like this generally be a good idea to have only one source of truth and to not needing SSH access to 3 systems? Instead it be only one.

As for builds.matomo.org: We could use an ssh key + secure password or an HTTP endpoint that will get required information and download / create the needed files. Maybe SSH is more secure assuming that we can prevent brute force attacks etc with `fail2ban` or similar software vs on an endpoint we need to implement things like brute force protection  ourselves. Technically, even if someone did brute force the webhook endpoint for example, then they still wouldn't be able to do much as long as they can't also trigger a release. Vs if someone had the SSH keys for some reason, then it be an issue. So thinking some endpoint might be better?

sgiehl:
@tsteur the difference between piwik.zip and matomo.zip is not only the filename. It contains a folder with the same name. And iirc when Matomo is updating it looks for a folder named piwik or matomo in the downloaded release. Older versions of Piwik might though only look for a folder named piwik, which might then cause problems if that isn't the case. 

Creating all version-related info files only on the builds server sounds good. Guess that makes everything a bit more stable. So we would need to adjust API and WWW to use those files instead (or sync them somehow). I guess I don't have access to those repos (anymore). Can you give me access, I would then check where they are actually used, so we can maybe also clean that up a bit and avoid creating unused files in the release process.

Regarding SSH connection VS HTTP Endpoint:

The script currently already does a SSH connect to upload the build files and adjust the version files. Currently this is done without a ssh key password. Not sure if there is a big benefit in using a password, as we would need to provide the ssh key and the password as github secrets. So if someone gains access to such secrets somehow, there is no big difference. But it should be possible to use a password for sure.

Using an HTTP endpoint would be an alternative way. Writing a simple PHP script, that receives a version number that should be downloaded / updated is quite easy. But as you mentioned it requires some more work around that to set it up properly.

Personally I don't have a big preference, but the HTTP endpoint might be more secure. We could also decide to go with SSH for now and think about building and using an HTTP endpoint instead later.

tsteur:
For `WWW` I could just set up two cronjobs like `cd /home/... && curl -O https://builds.matomo.org/LATEST` (also for `LATEST_SIZE`).

For API: You have access now: https://github.com/matomo-org/api.matomo.org/settings/access?guidance_task= Maybe easiest be to have a small php script that we can run in a cronjob every 5 minutes and downloads all `LATEST*` files from build server. I could set up the cron.

Remains the issue with `piwik.zip` vs `matomo.zip`. The only solution would be to attach both release files in that case?

Re SSH it be better to have a password on it just to be more safe in case token gets leaked. I would maybe directly build the HTTP solution though as it might be not too much work. If we don't want to have a public API we could also just have a script that checks every 5 minutes for a new release? Then we won't need to worry about passwords and what not and no files be directly accessible on the server.

sgiehl:
> For `WWW` I could just set up two cronjobs like `cd /home/... && curl -O https://builds.matomo.org/LATEST` (also for `LATEST_SIZE`).

That sounds fine.

> For API: You have access now: https://github.com/matomo-org/api.matomo.org/settings/access?guidance_task= Maybe easiest be to have a small php script that we can run in a cronjob every 5 minutes and downloads all `LATEST*` files from build server. I could set up the cron.

I'll prepare a simple script to fetch the files from builds.matomo.org and store them locally. Will create a PR once it's done.

> Remains the issue with `piwik.zip` vs `matomo.zip`. The only solution would be to attach both release files in that case?

If we download the github assets that's the only solution I can think of. Otherwise we would need to kind of rebuilt the package on the build server after downloading it. 
Other solutions would be to either discontinue updating the piwik.* files (So they will stay on the current version. That would at least allow older versions to update to that version and from there to the latest version using matomo.*) or to simply symlink them to matomo.* files (that might break automatic updates from version before the rebrand, but maybe thats a tolerable thing, as manual update would still work though).
 
> Re SSH it be better to have a password on it just to be more safe in case token gets leaked. I would maybe directly build the HTTP solution though as it might be not too much work. If we don't want to have a public API we could also just have a script that checks every 5 minutes for a new release? Then we won't need to worry about passwords and what not and no files be directly accessible on the server.

I have started to build a script, that automatically detects if a new release is available on github and downloads / updates the required files. Not sure if we should run it every 5 minutes, as it might hit the github API limit without authentication.
Should I create a new private repo for that single file, or shall we include it somewhere else?

tsteur:
We can use a new repository for this 👍  We can also configure a token to not run into quota limits.

I wouldn't build the zip files again on the build server, there would be issues over time. I don't think it's a big issue to have `piwik.zip` and `matomo.zip` attached to the release. Most people will download it on the website or directly or through Matomo and only few people will download it from there and it shouldn't be a problem if both files are attached.

sgiehl:
I have cleaned up the action now, so it doesn't do any remote work and also attaches the piwik files.
The part here should be fully ready to review and for testing.
@justinvelluppillai @tsteur @mattab If someone of you wants to have a quick demo of that, let me know and we can setup a quick video call for that. I can easily demonstrate that on my fork.

I'm still working on finishing the scripts that will be running on the builds and api server. But they can later be checked and reviewed separately

justinvelluppillai:
Thanks @sgiehl I will see if I can set up a call for demo asap.

.github/workflows/release.yml
```diff
@@ -0,0 +1,158 @@
...
+  pull-requests: read
+  repository-projects: none
+  security-events: none
+  statuses: read
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to build'
```
justinvelluppillai:
```suggestion
        description: 'Or specify a tag to build from'
```

Maybe the action needs to check and update `core/Version.php` otherwise it would be wrong when releasing from a tag?

.github/scripts/clean-build.sh
```diff
@@ -0,0 +1,229 @@
+# This script removes all files that shouldn't be included in a release
```
Findus23, line 2:
```suggestion
#!/bin/bash
# This script removes all files that shouldn't be included in a release
```

I think we should add a shebang here to make sure it is always called with bash and not another shell.
Also am I missing something or is this file never called from the other scripts?

Findus23 (commented):
Maybe now that we are testing the scripts, we could introduce the fixes from https://www.shellcheck.net/. 
But none of them are severe enough that they should break something.

sgiehl:
@Findus23 applied some of the fixes

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
+
+name: Build release
+
+permissions:
+  actions: read
+  checks: read
+  contents: write
```
tsteur: can we add comments for what the different permissions are needed? That be great. eg it's not obvious why "contents" requires "write". This will make it easier in future to maintain it

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+              exit 1
+            fi
+
+            if [[ ${GITHUB_REF#refs/heads/} != "4.x-dev" && ${GITHUB_REF#refs/heads/} != "next_release" ]]
+            then
+              echo "A tag can only be created from branches '4.x-dev' and 'next_release'. Please create the tag manually if a release needs to be built from another branch."
+              exit 1
+            fi
+
+            echo "Creating a tag for $version"
```
tsteur, line 121: can we add a check if released from `*.x-dev` branch that this is only allowed for alpha/beta releases?

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+            git fetch --tags -q 2>/dev/null
+            tag_exists=$( git tag --list "$version" )
+
+            if [[ -n "$tag_exists" ]]
+            then
+              echo "A tag for $tag_exists already exists."
+              exit 1
+            fi
+
+            if [[ ${GITHUB_REF#refs/heads/} != "4.x-dev" && ${GITHUB_REF#refs/heads/} != "next_release" ]]
```
tsteur: Can we update the docs in https://developer.matomo.org/guides/release-management#when-starting-to-work-on-the-next-major-release that we need to update this release workflow when starting to work on a new major release? I assume we will need to change then `4.x-dev` -> `5.x-dev`

tsteur (commented):
Left a few minor comments. The password check we could add after it's been merged.

Otherwise looks all good to merge from my perspective once the comments have been addressed and once the process of using this is documented and how it works (eg how to allow someone to release etc)

.github/workflows/release.yml
sgiehl: changed the check. It can now be any branch that matches `[4-9]\.x-dev`. So that should work for quite a while...

.github/workflows/release.yml
sgiehl: Ok. Only beta release tags are now allowed to be created from `*.x-dev` branches within the action.

.github/workflows/release.yml
sgiehl: I've updated the permissions. Actually only `contents` is required for creating the tag and the release. added a comment there.

tsteur:
👍

sgiehl:
Guess we actually could merge this one now. Once the secrets have been added we can give it a first try...

PR #17598, opened 2021-05-19
Modiy refer banner

### LINKED ISSUE #17571: annoying banner (the 'refer matomo' banner at the bottom of the page)
No, seriously folks, I don't need this *reminder* in order to recommend something. Pls don't clutter already cluttered space with ridiculous popups and banners and what not. Pls don't make us use ad-blocker on our own self-hosted analytics ffs. Thx.

piwik-banner

Instead, it would be better to actually ad a changelog for the latest version 4.3.0 here https://matomo.org/changelog/

flamisz:
### Description:

fixes #17571 

### Review

* [ ] Functional review done
* [ ] Code review done


plugins/Feedback/Feedback.php
```diff
@@ -135,7 +139,12 @@ public function showReferBanner()
         $nextReminderDate = $referReminder->getUserOption();
 
         if ($nextReminderDate === false) {
-            return true;
+            $nextReminder = Date::now()->getStartOfDay()->addDay(6 * 30)->toString('Y-m-d');
```
tsteur: @flamisz I believe the "review reminder" is showing every 90 days so there's a risk both review and refer banners and shown at same point. 

Maybe after 6 months (180 days) we might show both reminders approx on a similar day? The 90 days in a review reminder is sent from `plugins/Feedback/angularjs/feedback-popup/feedback-popup.controller.js`. I now wonder if we maybe could add 135 days to the next "review reminder" date? This way it be 45 days before the popup would be shown. If no review reminder date is set yet, then we would maybe need to set the review reminder to be in 90 days and the refer reminder to be in 135 days? 

Bit complicated just wanting to avoid them being shown to close of each other :)

tsteur (changes requested):
looks good @flamisz  👍  Left a comment and wondering if we can directly close the popup once a share link was clicked and then we never show it again. Assuming most people would only share through one network anyway. 

I haven't looked yet is it already implemented to not show the banner again once clicked on a social network? Could maybe show in the thank you message otherwise "Thank you... we won't show this banner again" or something?

plugins/Feedback/Feedback.php
flamisz: Actually I check in the code, at line 111:

```php
if ($this->getShouldPromptForFeedback()) {
...
```
This way we can't show it at the same time. 
But to make sure they don't come up close to each other (next day, etc), make sense. I'll check your recommendation and add it.

flamisz:
> looks good @flamisz 👍 Left a comment and wondering if we can directly close the popup once a share link was clicked and then we never show it again. Assuming most people would only share through one network anyway.
> 
> I haven't looked yet is it already implemented to not show the banner again once clicked on a social network? Could maybe show in the thank you message otherwise "Thank you... we won't show this banner again" or something?

We can do that, yes. I think originally we tried something like this but modified thinking about someone who wants to share on multiple platforms. 
In this version, because we have the 2 buttons after the user shares it, I don't close it, just show the thanks message and the user can decide which button to use closing it.
But I like the idea to close it, don't show it again after a user shares even on one platform. They can share it without this modal if they want later too. 

So I will modify it to close the modal, save it to not show it again, and show the thanks message with the recommended text.

tsteur:
Cheers @flamisz 💯

Issue #17577, opened 2021-05-18
Split private directories system check into "Required" and "Recommended"


mddvul22:
Today, I upgraded to Matomo 4.3.0.  The Matomo system check complains about "Required Private Directories".  Specifically, it tells me:

"We found that the above URLs are accessible via the browser, but they should NOT be. Allowing them to be accessed can pose a potential security risk since the contents can provide information about your server and potentially your users. Please restrict access to them.

We also found that Matomo's config directory is publicly accessible. While attackers can't read the config now, if your webserver stops executing PHP files for some reason, your MySQL credentials and other information will be available to anyone. Please check your webserver config and deny access to this directory. "

I'm not positive, but something seems to be messed up here.  I have my config directory set to 700.  My config.ini.php is set to 600.  What should the permissions be?  Matomo doesn't say what it is looking for.  For me to lock it down further, I basically have to change take access away from the Apache user.  Maybe I'm missing something obvious, but this doesn't seem right.

## Your Environment
* Matomo Version: 4.3.0
* PHP Version: 7.2.24
* Server Operating System: Centos 8 Linux

sgiehl:
Hi @mddvul22. Thanks for creating this issue. Matomo actually tries to do a http request on the config file and checks if the response code to be between 400 and 500. So you may need to disallow access in webserver config.
See https://github.com/matomo-org/matomo/pull/17568

heurteph-ei:
Maybe some documentation somewhere should be updated... And a link to this doc should be provided in the warning message?

fireba11:
At least on my server 
config/config.ini.php returns a single ';'
tmp/cache/tracker/matomocache_general.php returns an empty file
causing the check to complain since something is returned.

Also to me it seems those files exist by default, how are they supposed to be protected? (The warning should tell you too ..)

Findus23:
I have explained this a bit more detailed here:
https://forum.matomo.org/t/nach-update-auf-4-3-fehlermeldung-required-private-directories/41903/4?u=lukas

Maybe the error message could be expanded with this help

fireba11:
Note:
I'm using centos7 and httpd, the .htaccess is there but doesn't seem to work for me.
I'm investigating possible reasons, the use of php-fpm being one candidate ...

Findus23:
One thing: With the current check Matomo doesn't accept it, if the config.ini.php returns a 304 e.g. to the homepage.

mddvul22:
[...]

PR #17604, opened 2021-05-21
Improve detection of private directories

tsteur:
### Description:

refs https://github.com/matomo-org/matomo/issues/17577

This is not a full solution yet for this issue but it may improve things for the next patch release by detecting redirects a bit better for the config file. 

Added FAQ with more information: https://matomo.org/faq/troubleshooting/how-do-i-fix-the-error-private-directories-are-accessible/  as users currently wouldn't know what to do. It's far from perfect but we can tweak it over time and provides more information for now.

Also now blocking `.git` directory automatically if it exists. Would be otherwise complicated to explain how to do it. Not sure why we didn't do it earlier.

Also added new command `core:create-security-files` to create these files automatically if Matomo does not have the permissions to do it automatically (see the FAQ). Initially, I had added the creation of these files to the `diagnostics:run` command (or when you open system report) but this would have caused issues as they would have potentially never noticed that they need to create these files after every update etc (because the security files would be created when viewing the system report but it would not show there was a problem between updating Matomo and viewing the report :-) ). Using this command it makes it easier for users to tell them in the guide how they can fix the issue. 

I will leave a comment about this in the issue what needs further tweaking.

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
@tsteur it looks like this includes some extra commits from another PR, is that right?

tsteur:
indeed @diosmosis I've recreated the PR in https://github.com/matomo-org/matomo/pull/17606

PR #17606, opened 2021-05-21
Improve required private directories check

tsteur:
### Description:

refs https://github.com/matomo-org/matomo/issues/17577

This is not a full solution yet for this issue but it may improve things for the next patch release by detecting redirects a bit better for the config file. 

Added FAQ with more information: https://matomo.org/faq/troubleshooting/how-do-i-fix-the-error-private-directories-are-accessible/  as users currently wouldn't know what to do. It's far from perfect but we can tweak it over time and provides more information for now.

Also now blocking `.git` directory automatically if it exists. Would be otherwise complicated to explain how to do it. Not sure why we didn't do it earlier.

Also added new command `core:create-security-files` to create these files automatically if Matomo does not have the permissions to do it automatically (see the FAQ). Initially, I had added the creation of these files to the `diagnostics:run` command (or when you open system report) but this would have caused issues as they would have potentially never noticed that they need to create these files after every update etc (because the security files would be created when viewing the system report but it would not show there was a problem between updating Matomo and viewing the report :-) ). Using this command it makes it easier for users to tell them in the guide how they can fix the issue. 

I will leave a comment about this in the issue what needs further tweaking.

replaces https://github.com/matomo-org/matomo/pull/17604

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CoreUpdater/Commands/SecurityFiles.php
+++ b/plugins/CoreUpdater/Commands/SecurityFiles.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * Matomo - free/libre analytics platform
+ *
+ * @link https://matomo.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ *
+ */
+namespace Piwik\Plugins\CoreUpdater\Commands;
+
+use Piwik\Plugins\Installation\ServerFilesGenerator;
+use Piwik\Plugin\ConsoleCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @package CoreUpdater
+ */
+class SecurityFiles extends ConsoleCommand
+{
+    protected function configure()
+    {
+        $this->setName('core:create-security-files');
+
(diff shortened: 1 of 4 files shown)


plugins/CoreUpdater/Commands/SecurityFiles.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        $this->setName('core:create-security-files');
+
+        $this->setDescription('Creates some web server security files if they haven\'t existed previously. Useful when using for example Apache or IIS web server and Matomo cannot create these files automatically because of missing write permissions.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        ServerFilesGenerator::createFilesForSecurity();
+        $output->writeln('Done. To check if this worked please open the system report or run `./console diagnostics:run` and look out for the private directories check. If it doesn\'t work you may need to execute this command using a user that has write permissions. Please note you may need to execut this command every time you update Matomo to a newer version.');
```
Findus23: Maybe also mention here that if it doesn't work, they should check if they are using another webserver than Apache or IIS

diosmosis (approved):
Looks good to me, tested locally and it works

plugins/CoreUpdater/Commands/SecurityFiles.php
tsteur: 👍

PR #17610, opened 2021-05-21
Use originalString when returning the Segment definition string

### LINKED ISSUE #17583: Creating/updating segments w/ values w/ encoded chars does not auto schedule rearchiving
## Expected Behavior

Creating or updating a segment w/ an encoded value, eg, `pageUrl==some%2Fpath` should automatically schedule rearchiving of past data. Invalidations should appear in the archive_invalidations table and should be handled by core:archive.

## Current Behavior

Creating or updating the aforementioned type of segment and running core:archive will add rows to archive_invalidations, but will not run them.

1. Open the segment editor and create a segment with an encoded value. For example, `Page URL is <url>`
2. Run core:archive. No jobs for the new segment will be created.
3. Look in the archive_invalidations table, invalidations will exist for the new segment.

## Your Environment

* Matomo Version:
* PHP Version:
* Server Operating System:

* Browser:
* Operating System:

diosmosis:
### Description:

Fixes #17583

The string definition passed into Segment will sometimes be decoded before being saved in Segment::$string. So if we reuse the string to create a new Segment, the new segment's hash will be different. This PR changes getString() to return the original string so we're always using the definition we started with.

EDIT: Modified this PR to not change Segment::getString(), since that resulted in a test failure that looked like a regression. Instead there's a new method. We can replace every occurrence if desired in 4.4.0.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Archive.php
+++ b/core/Archive.php
@@ -827,7 +827,7 @@ class Archive implements ArchiveQuery
             $this->initializeArchiveIdCache($doneFlag);
 
             $prepareResult = $coreAdminHomeApi->archiveReports(
-                $site->getId(), $period->getLabel(), $periodDateStr, $this->params->getSegment()->getString(),
+                $site->getId(), $period->getLabel(), $periodDateStr, $this->params->getSegment()->getOriginalString(),
                 $plugin, $requestedReport);
 
             if (!empty($prepareResult)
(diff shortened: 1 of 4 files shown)


diosmosis:
An alternative solution here is just to put the `urlencode()` back in.

PR #17617, opened 2021-05-25
Make sure that all segment archives are invalidated when invalidating all visits archives

diosmosis:
### Description:

When we invalidate an All Visits archive (`segment = ''`) we also invalidate all segment archives for the same site and period. This is because if the data for the main archive needs to be recomputed, then the segment archives may need to be recomputed as well.

Any archive that is invalidated needs to also show up in the archive_invalidations table. However, currently, we are not inserting the extra segment archives into that table. This is fixed by this PR.

### Review

* [ ] Functional review done
* [ ] Code review done


flamisz (approved):
I reviewed the code and tested it locally. It looks OK to me. 👍

PR #17618, opened 2021-05-25
Fix uncaught exception when running system tests for plugins

sgiehl:
### Description:

When running system tests for plugins there is currently an uncaught exception when the tests starts running.
See eg. https://travis-ci.com/github/matomo-org/matomo/jobs/507308739#L732

This is caused by a dataprovider that tries to fetch available segments before the database is actually set up. As the visitor id segment tries to check for a system settings this fails.

This actually caused the AutoSuggestAPI test not to run anymore

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
:+1:

PR #17620, opened 2021-05-25
clear EventDispatcher cache in tests

sgiehl:
### Description:

Was writing some tests that require the ActivityLog plugin but failed to get them working. The reason actually was, that the ActivityLog plugin dynamically tries to register hooks. To get that working in tests it's required to clear the registered hooks that might have been cached previously.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/tests/PHPUnit/Framework/TestCase/IntegrationTestCase.php
+++ b/tests/PHPUnit/Framework/TestCase/IntegrationTestCase.php
@@ -9,13 +9,11 @@
 namespace Piwik\Tests\Framework\TestCase;
 
 use Piwik\Access;
-use Piwik\Config;
+use Piwik\Cache as PiwikCache;
 use Piwik\Db;
-use Piwik\DbHelper;
-use Piwik\Menu\MenuAbstract;
+use Piwik\EventDispatcher;
 use Piwik\Option;
 use Piwik\Tests\Framework\Fixture;
-use Piwik\Cache as PiwikCache;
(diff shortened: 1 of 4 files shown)


tests/PHPUnit/Framework/TestCase/IntegrationTestCase.php
```diff
@@ -92,6 +93,7 @@ public function setUp(): void
 
         PiwikCache::getEagerCache()->flushAll();
         PiwikCache::getTransientCache()->flushAll();
+        EventDispatcher::getInstance()->clearCache();
```
diosmosis, line 95: Can we put this in `Fixture::clearInMemoryCaches()`?

diosmosis: I see we call that in tearDown only for some reason... seems like a mistake, and we should clear all of those caches here too...

sgiehl: @diosmosis updated the PR

sgiehl: @diosmosis seems clearing all caches at this point lets various integration tests start failing. Seems clearing all static instances is a problem at that point as some fixtures are manipulating those before.

sgiehl: Reverted it so it only adds the additional cache clear, but left a note so it's clear why not all caches are cleared

PR #17624, opened 2021-05-26
Fix possible notice when no user is set

sgiehl:
### Description:

Especially in tests it might be possible that a user is actually not set. If for any reason `Piwik::getCurrentUserEmail()` is then called a notice might be triggered causing a test to fail.

In theory there might also be other use cases where that could happen. Like using a custom login plugin, that actually does not use the users stored in database.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Piwik.php
+++ b/core/Piwik.php
@@ -176,7 +176,7 @@ class Piwik
     public static function getCurrentUserEmail()
     {
         $user = APIUsersManager::getInstance()->getUser(Piwik::getCurrentUserLogin());
-        return $user['email'];
+        return $user['email'] ?? '';
     }
 
     /**


core/Piwik.php
```diff
@@ -176,7 +176,7 @@ public static function getRandomTitle()
     public static function getCurrentUserEmail()
     {
         $user = APIUsersManager::getInstance()->getUser(Piwik::getCurrentUserLogin());
-        return $user['email'];
```
diosmosis, line 179: I think the phpdocs need to be changed here too. And do we need to mention in the developer changelog that the value can be empty now? I guess it might always have been possible if the user was `anonymous`...

sgiehl: The phpdocs now actually would fit. Before it could maybe have returned `null` if the user array was empty.
Don't think we need to mention it in the developer docs. The change actually only avoids a possible warning in this case.

PR #17636, opened 2021-05-28
Support for Tracker INI settings to be set for each website

diosmosis:
### Description:

This change allows INI config sections like `[Tracker_$idSite]` to be added which will override some Tracker config for individual sites. Config that should be applied overall (or in places where we don't know what the idSite is), cannot be overridden.

### Review

* [x] Functional review done
* [x] Code review done

### DIFF
--- a/core/Tracker/Response.php
+++ b/core/Tracker/Response.php
@@ -25,7 +25,7 @@ class Response
     {
         ob_start(); // we use ob_start only because of Common::printDebug, we should actually not really use ob_start
 
-        if ($tracker->isDebugModeEnabled() && TrackerConfig::getConfigValue('enable_sql_profiler')) {
+        if ($tracker->isDebugModeEnabled() && TrackerConfig::getConfigValue('enable_sql_profiler', $this->getIdSite())) {
             $this->timer = new Timer();
 
             TrackerDb::enableProfiling();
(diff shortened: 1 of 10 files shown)


core/Tracker/Response.php
```diff
@@ -25,7 +25,7 @@ public function init(Tracker $tracker)
     {
         ob_start(); // we use ob_start only because of Common::printDebug, we should actually not really use ob_start
 
-        if ($tracker->isDebugModeEnabled() && TrackerConfig::getConfigValue('enable_sql_profiler')) {
+        if ($tracker->isDebugModeEnabled() && TrackerConfig::getConfigValue('enable_sql_profiler', $this->getIdSite())) {
```
tsteur: fyi @diosmosis this method getIdSite doesn't seem to be defined?

tsteur: we could just remove idSite

diosmosis: :+1: this one doesn't need to be overridden

tsteur:
@diosmosis added https://matomo.org/faq/how-to/how-do-i-configure-a-config-setting-in-the-config-ini-php-differently-for-a-specific-site/ and also added the logic for 3 more settings where this may be quite useful. Once tests pass it should be good to merge.

tsteur:
@diosmosis a few tests seem to need fixing and then should be good to merge

PR #17654, opened 2021-06-08
Fix redirect link not working after plugin activation

Findus23:
reported in https://forum.matomo.org/t/i-get-an-error-when-i-click-to-activate-the-plugin/42173/2

It turns out unlike the regular "activate plugin" buttons, the ones shown after uploading a plugin zip don't have a `redirectTo` GET parameter set. 

https://github.com/matomo-org/matomo/blob/587cc39e0362719332d410b7a4d5ddcc68788eeb/plugins/CorePluginsAdmin/templates/uploadPlugin.twig#L29

This should not matter as an empty value means redirecting to the plugin or theme list (which is what we want.

https://github.com/matomo-org/matomo/blob/792bc0cfd5bc9365351097e07710f3148f3b3f32/plugins/CorePluginsAdmin/Controller.php#L469-L476

But https://github.com/matomo-org/matomo/pull/17345 added a section that reads the parameters, but forgot to set a default value for `redirectTo` in case it is missing, throwing the error reported on the forum

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CorePluginsAdmin/Controller.php
+++ b/plugins/CorePluginsAdmin/Controller.php
@@ -9,29 +9,27 @@
 namespace Piwik\Plugins\CorePluginsAdmin;
 
 use Exception;
-use Piwik\Access;
-use Piwik\API\Request;
+use Piwik\API\ResponseBuilder;
+use Piwik\ArchiveProcessor\Rules;
 use Piwik\Common;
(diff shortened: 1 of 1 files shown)


diosmosis:
Thanks for the fix @Findus23!

Issue #16942, opened 2020-12-13
Enable tracking performance and browser plugins without pageview (offer ability to force a specific page view ID in JS tracker)


JE4GLE:
Hello,

I use a combination of the HTTP Tracking API with PHP and the JavaScript Tracker. The PHP side tracks page views, ECommerce and other events, so that every page view and transaction is reported and not lost by users blocking the JavaScript tracker. In addition to the PHP side, the JavaScript tracker tracks the rest, like content tracking. This has the disadvantage that page views are not tracked using the JavaScript tracker and I lose important data like Performance tracking (Plugins are not so important, but share the principle).

My suggestion would be to allow sending performance and browser plugins without a page view. Maybe using an event or a separate method like `trackPerformance()`, which should not be used in combination with `trackPageView([customTitle])`.

Thank you!

tsteur:
Hi @JE4GLE 

for page performance metrics this might be already the case if you track for example an event on the page. Then Matomo would currently attribute this to the page view. However, this can only work if you track a page view ID server side. The problem is that you can't really know what page view ID to send because it's randomised in JavaScript tracker. So a method to force a specific page view ID to be used should be all that's needed here. Then any event or content tracking request would set the correct page performance metrics on the page view.

JE4GLE:
Thank you @tsteur

So I would call this method `setPageView` with my pageview ID in my PHP Tracker and the next event I send using the JavaScript tracker would automatically attach Performance Metrics data to the next event?
Wouldn't it be possible to just send a page view again using JS and because of the same pageview ID it wouldn't track 2 page views? This way, you would have a future-proof way of using the combined advantages of the JS page view and the PHP page view.

tsteur:
> So I would call this method setPageView with my pageview ID in my PHP Tracker and the next event I send using the JavaScript tracker would automatically attach Performance Metrics data to the next event?

Exactly.

> Wouldn't it be possible to just send a page view again using JS and because of the same pageview ID it wouldn't track 2 page views? This way, you would have a future-proof way of using the combined advantages of the JS page view and the PHP page view.

[...]

PR #17655, opened 2021-06-08
JS Tracker: new setPageViewId function to override the pageviewid and not have it auto generated 

JE4GLE:
### Description:

This Pull Request solves the issue mentioned in the following issue: https://github.com/matomo-org/matomo/issues/16942

It would allow for a combination of the PHP tracker (to track 100% of the PageViews) and the JavaScript tracker (to track screen size, plugins, …) by first generating a page view id on the PHP side and setting it after page generation using JavaScript.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/js/piwik.js
+++ b/js/piwik.js
@@ -63,7 +63,7 @@
     setCustomVariable, getCustomVariable, deleteCustomVariable, storeCustomVariablesInCookie, setCustomDimension, getCustomDimension,
     deleteCustomVariables, deleteCustomDimension, setDownloadExtensions, addDownloadExtensions, removeDownloadExtensions,
     setDomains, setIgnoreClasses, setRequestMethod, setRequestContentType, setGenerationTimeMs,
-    setReferrerUrl, setCustomUrl, setAPIUrl, setDocumentTitle, getPiwikUrl, getMatomoUrl, getCurrentUrl,
+    setReferrerUrl, setCustomUrl, setAPIUrl, setDocumentTitle, setPageViewId, getPiwikUrl, getMatomoUrl, getCurrentUrl,
     setDownloadClasses, setLinkClasses,
     setCampaignNameKey, setCampaignKeywordKey,
     getConsentRequestsQueue, requireConsent, getRememberedConsent, hasRememberedConsent, isConsentRequired,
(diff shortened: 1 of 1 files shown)


diosmosis:
build js

diosmosis:
Hi @JE4GLE, thanks for taking the time to create this pull request! Currently the jslint tests are failing, I believe because the new method is not currently in the list here: https://github.com/matomo-org/matomo/blob/4.x-dev/js/piwik.js#L66. Would you be able to add it?

diosmosis:
build js

diosmosis:
Hmm, looks like the tracker javascript building workflow isn't working on this PR. I'll build it manually before merging.

diosmosis:
@JE4GLE  Looks like this change would create a regression, actually. `logPageview()` generates a pageview id, and is supposed to generate a new one on every subsequent call so new pageviews have new IDs. Now, however, the pageview ID is only set on the first call, then since it's already set at the second call, nothing changes.

This could be fixed pretty easily by having another variable like `configIdPageViewSetManually` that is set to `true` in `setPageViewId()`. Then we check this in `logPageview()`. However I'm not sure if this causes problems for your usecase. Would you be able to change the pageview ID and still have it determined by the PHP tracker code you're using (assuming I understand the use case correctly)?

JE4GLE:
@diosmosis My use case would be:

1. Create a pageview request with a pageview ID in PHP and place it in the HTML code, before page generation
2. After the page has loaded, the JavaScript code will then read this ID in the HTML code and set it manually
3. Send the pageview again using JavaScript, but this time including additional data from JavaScript

Unfortunately I don't quite understand, why the pageview ID would change during one page visit. In my understanding, the user visits a page and a pageview ID is generated, which will stay the same, until the user reloads the page or goes to another one.

Anyways, here's your proposed change:

diosmosis:
@JE4GLE 

> Unfortunately I don't quite understand, why the pageview ID would change during one page visit. In my understanding, the user visits a page and a pageview ID is generated, which will stay the same, until the user reloads the page or goes to another one.

After `logPageview()` is called we expect the pageview to be over, and subsequent calls to use a different ID. In a normal website where you click links that load new pages, this wouldn't be needed, but for single page apps where the page never reloads, and the app calls logPageview() itself multiple times without the page reloading, this change would break things.

I guess if you're expecting only one pageview before a new page is loaded or the current page is reloaded, then I suppose you wouldn't be affected by this.

diosmosis:
The change looks good now and the build passes. Thanks again for taking the trouble to make the PR and apply fixes @JE4GLE!

PR #17657, opened 2021-06-09
Remove archive status locking since it is not needed anymore

diosmosis:
### Description:

The status and ts_started columns in the archive_invalidation table provide the same functionality now so the extra overhead of acquiring and re-acquiring locks is no longer needed.

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (commented):
Code looks fine. Some tests needs to be updated

PR #17660, opened 2021-06-10
Ignore invalid capabilities instead of throwing an exception

sgiehl:
### Description:

When the database contains an access entry with an capability that is no longer present (e.g. TagManager capability, but TagManager gets disabled), the user admin currently does not show any users, but also no error.

The missing error is caused by the incorrect Piwik::translate, which actually triggers a warning. The error itself is caused by the exception thrown when a capability does not exist.

fixes L3-108

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/UsersManager/API.php
+++ b/plugins/UsersManager/API.php
@@ -462,19 +462,39 @@ class API extends \Piwik\Plugin\API
         return $userSites;
     }
 
+    /**
+     * Throws an exception if one of the given access types does not exists.
+     *
+     * @param string|array $access
+     * @throws Exception
+     */
     private function checkAccessType($access)
     {
         $access = (array) $access;
 
-        $roles = $this->roleProvider->getAllRoleIds();
-        $capabilities = $this->capabilityProvider->getAllCapabilityIds();
-        $list = array_merge($roles, $capabilities);
-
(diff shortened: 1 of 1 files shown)


plugins/UsersManager/API.php
```diff
@@ -1448,8 +1448,13 @@ private function getRoleAndCapabilitiesFromAccess($access)
...
             } else {
-                $this->checkAccessType($entry);
-                $capabilities[] = $entry;
+                try {
+                    $this->checkAccessType($entry);
+                    $capabilities[] = $entry;
+                } catch (\Exception $e) {
+                    // if capability does not exist any longer (e.g. removed plugin) an exception might be thrown
+                    // we ignore that capability in that case
+                }
```
diosmosis, line 1473: Do we need to check specifically if this is a "capability does not exist" error? Is it possible that other errors might get thrown (if there's a bug, for instance, or if some code in checkAccessType() changes)? Maybe we need an `isValidCapability()` method...

sgiehl: changed that 👍

diosmosis:
The change looks good :+1: but there are some tests failing.

sgiehl:
@diosmosis should be fixed

PR #17661, opened 2021-06-10
Ensure redirects from logme method are only done to trusted hosts

sgiehl:
### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Login/Controller.php
+++ b/plugins/Login/Controller.php
@@ -315,7 +315,7 @@ class Controller extends \Piwik\Plugin\ControllerAdmin
 
         // only use redirect url if host is trusted
         if (!empty($parsedUrl['host']) && !Url::isValidHost($parsedUrl['host'])) {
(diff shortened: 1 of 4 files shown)


Findus23:
I think this could also be mentioned in the CHANGELOG as it might need changes from people who integrated Matomo with their online-platform.

plugins/Login/Controller.php
```diff
@@ -311,6 +311,15 @@ protected function authenticateAndRedirect($login, $password, $urlToRedirect = f
         // remove password reset entry if it exists
         $this->passwordResetter->removePasswordResetInfo($login);
 
+        $parsedUrl = parse_url($urlToRedirect);
+
+        // only use redirect url if host is trusted
+        if (!empty($parsedUrl['host']) && !Url::isValidHost($parsedUrl['host'])) {
+            $e = new \Piwik\Exception\Exception('The redirect URL host is not valid, it is not a trusted host. If this URL is trusted, you can allow this in your config.ini.php file by adding the line <i>trusted_hosts[] = "'.$parsedUrl['host'].'"</i> under <i>[General]</i>');
```
diosmosis: This is probably unnecessary, but just in case, can we sanitize `$parsedUrl['host']`?

sgiehl: changed that 👍

PR #17663, opened 2021-06-10
Updates some npm dependencies for screenshot tests

sgiehl:
### Description:

Some of the npm dependencies used for screenshot testing keep triggering security warnings.
Even though they aren't used in production it shouldn't hurt to update at least some of them.

Updating mocha might also be useful at some point as it is quite outdated, but that will require various adjustments in our code...

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/tests/lib/screenshot-testing/package-lock.json
+++ b/tests/lib/screenshot-testing/package-lock.json
@@ -29,11 +29,11 @@
       "integrity": "sha512-TMeqbNl2fMW0nMjTEPOwe3J/PRFP4vqeoNuQMG0HlMrtm5QxKqdvAkZ1pRBQ/ulIyDD5Yq0nJ7YbdD8ey0TO3g=="
     },
     "ajv": {
-      "version": "6.10.2",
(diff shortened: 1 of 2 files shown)


Findus23:
I guess updating puppeteer should also be useful as Chrome 80.0.3987.0 is quite old now. Maybe we could even automate this somehow so that we notice things that break in newer browsers.

sgiehl:
The problem with updating puppeteer is that in most cases all screenshots are changing slightly, so we actually can't automate it.
And doing the update takes some time, as you need to look through all changing screenshots in order to check if all changes are "valid". Might be useful to create an issue for that, so we can prioritize that.

diosmosis:
I guess we could have a github action that does it, say once a month, and just auto commits all the changed screenshots, but only creates a PR. Then someone can review each screenshot and make changes or merge. Would that work?

sgiehl:
@diosmosis guess that should work. Might be best to create an issue for it

diosmosis:
Created https://github.com/matomo-org/matomo/issues/17667

PR #17664, opened 2021-06-11
Add extra info when two non-summable columns are added together for easier debugging 

diosmosis:
### Description:

As title. Refs L3-104.

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (approved):
Guess having a test for that case doesn't make that much sense, right?
Looks good to merge otherwise

diosmosis:
Given this is a tweak to an error message, I think you're right that it doesn't need a test. I'll merge it.

PR #17665, opened 2021-06-11
Disable logme functionality by default

sgiehl:
### Description:

Logme functionality will be disabled for new installs by default now. For existing installs an update script should automatically enable it in config to keep bc.

Note: Once this has been merge we need to update the FAQ here https://matomo.org/faq/how-to/faq_30/ and maybe also some other places.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)*
 
 ### Breaking Changes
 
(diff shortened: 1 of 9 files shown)


diosmosis:
@sgiehl should the developer changelog be updated? I'm thinking it would be helpful to both let users know they need to set it explicitly, and that it's recommended for existing installs to disable it since most won't need this functionality.

sgiehl:
added a changelog entry

mattab:
@sgiehl Could you please document this in https://matomo.org/faq/how-to/faq_30/ as mentioned in the description? Thanks

sgiehl:
done

PR #17675, opened 2021-06-15
Make sure overlay API requests send force_api_session as POST param

diosmosis:
### Description:

In the JS tracker we try to detect whether we are in an Overlay iframe by checking the HTTP referrer. If the referrer is from a specific Overlay controller action, then we know we're in Overlay.

The regex for parsing the segment parameter in this detection logic is incorrect. In it, we end up matching the segment parameter, plus everything after it. For example, in `&segment=&token_auth=...&force_api_session=1` we match `&token_auth=...&force_api_session=1`, and assume this is the segment.

This is passed on to the Overlay/client.js code which appends it to every API request it makes, assuming that it's the segment. These parameters passed in the URL breaks the force_api_session code, since in the browser we make POST requests (so the token_auth is not saved in the browser cache), and server side we do not allow force_api_session to be used in $_GET if using a POST request.

Fixing this in two ways:

1. fixing the regex to not be greedy
2. making sure token_auth/force_api_session are not appended to the URL even if present in the API request

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
build js

diosmosis:
build js

sgiehl (commented):
The code changes look good. But didn't do any local testing yet

diosmosis:
The customer has replied it works, @sgiehl ok to merge this?

sgiehl:
Sure

PR #17677, opened 2021-06-16
Ensure login is correctly detected for brute force detection when using logme feature

sgiehl:
### Description:

When using logme feature the brute force detection on login level might not work correctly.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Login/Login.php
+++ b/plugins/Login/Login.php
@@ -221,11 +221,12 @@ class Login extends \Piwik\Plugin
         $login = StaticContainer::get(\Piwik\Auth::class)->getLogin();
         if (empty($login) || $login == 'anonymous') {
             $login = Common::getRequestVar('form_login', false);
+            $action = Common::getRequestVar('action', false);
+            if ($action == 'logme') {
+                $login = Common::getRequestVar('login', $login);
+            }
         }
-        $action = Common::getRequestVar('action', false);
-        if ((empty($login) || $login == 'anonymous') && $action == 'logme') {
-            $login = Common::getRequestVar('login', false);
-        }
+
         return $login;
     }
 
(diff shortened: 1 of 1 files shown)


plugins/Login/Login.php
```diff
@@ -219,11 +219,13 @@ protected static function isModuleIsAPI()
...
-        if (empty($login)
-            || $login == 'anonymous'
-        ) {
+        if (empty($login) || $login == 'anonymous') {
             $login = Common::getRequestVar('form_login', false);
         }
+        $action = Common::getRequestVar('action', false);
+        if ((empty($login) || $login == 'anonymous') && $action == 'logme') {
+            $login = Common::getRequestVar('login', false);
+        }
```
diosmosis: Doing `empty($login) || $login == 'anonymous'` seems a bit redundant. Would it be worth it to simplify that?

Also, do you think it makes sense to move this function to another class, for example, BruteForceDetection so it can be tested?

sgiehl: Will simplify the code. Moving it actually wouldn't make it much better I guess.

diosmosis: I was thinking moving might help so we could cover it in tests. Do you think testing it would be worth the trouble?

sgiehl:
As the parameter names are defined in Login plugin I don't think it would
make sense to move it. Also testing might only make sense by adding more UI
tests or maybe some kind of integration tests. Only testing that method
won't help if the parameter name can be changed somewhere else (eg in logme
method or login action form)

diosmosis:
I guess we could potentially add assertions to an existing login/logme test (for example, "assert that a new entry was added to the brute force log"). I'll merge this for now, if needed we can modify the test later.

PR #17683, opened 2021-06-17
fix regression in trackingspamprevention and warning in bound parameter checking code

diosmosis:
### Description:

Changes:

* allow option to disable allowed host check since trackingspamprevention makes a request to amazonaws
* bound parameters can be a single string rather than an array
* disable testdox and add build message output to enable it if the build hangs

### Review

* [ ] Functional review done
* [ ] Code review done


core/bootstrap.php
```diff
@@ -37,6 +37,7 @@
 @date_default_timezone_set('UTC');
 
 disableEaccelerator();
+printTestDoxHint();
```
sgiehl: Seems that currently printed in any case. Shouldn't that only be shown when running tests and if testdox isn't enabled?

sgiehl (changes requested):

core/bootstrap.php
diosmosis: @sgiehl I can add this, but given testdox will rarely be used, and will only be set manually, I didn't think it was worth the added complexity. Let me know if this is still an important change.

diosmosis:
Also note: after merging, TrackingSpamPrevention's minimum required matomo will need to be updated

sgiehl:
The problem is that most tests are failing. Guess it might make more sense
to move that to the console command that runs the tests maybe

diosmosis:
@sgiehl I see, nevermind, phpstorm gave me the wrong bootstrap file. It should be in the test's bootstrap.php.

Issue #17686, opened 2021-06-17
Make Matomo compatible with PHP8.1


Findus23:
The first alpha of php8.1 was just released:
https://www.php.net/archive/2021.php#2021-06-10-1

There are a few breaking changes, so it might be useful to already start fixing things to get them released in time.

Changes: https://github.com/php/php-src/blob/php-8.1.0alpha1/UPGRADING

The major breaking change is that built-in functions now (just like user-defined functions) no longer accept null in non-nullable arguments:
https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg

-------------------------------

<details>
<summary>Click to show fixed issues</summary>

- [x] monolog datetime
>  Deprecated: DateTime::__construct(): Passing null to parameter #1 ($datetime) of type string is deprecated in /home/lukas/public_html/matomophp8/vendor/monolog/monolog/src/Monolog/Logger.php on line 324
- fixed in `https://github.com/Seldaek/monolog/pull/1554`
- fix released in monolog/monolog `1.26.1`
- fixed by #17828

-----------------
- [x] glob in upgradephp
> WARNING [2021-06-17 21:51:48] 298624  /home/lukas/public_html/matomophp8/libs/upgradephp/upgrade.php(260): Deprecated - glob(): Passing null to parameter #2 ($flags) of type int is deprecated - Matomo 4.0.0-b3 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)
- fixed by https://github.com/matomo-org/matomo/pull/17687

------------------
- [x] db logging string instead of array
> WARNING [2021-06-17 22:11:46] 311625  /home/lukas/public_html/matomophp8/core/Db.php(829): Warning - foreach() argument must be of type array|object, string given - Matomo 4.4.0-b1 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)
- when running `./console core:update`
- fixed by #17688
-------------
- [x] PDO fetch default
> WARNING [2021-06-17 22:27:06] 321855  /home/lukas/public_html/matomophp8/libs/Zend/Db/Statement/Pdo.php(257): Deprecated - PDOStatement::fetch(): Passing null to parameter #2 ($cursorOrientation) of type int is deprecated - Matomo 4.4.0-b1 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already)
[...]

PR #17693, opened 2021-06-18
PHP 8.1 compatibility, fix "Deprecated - explode(): Passing null to parameter"

Findus23:
see https://github.com/matomo-org/matomo/issues/17686

`$ids` can be null because of

https://github.com/matomo-org/matomo/blob/1fcc10569378263775a1f8a3e2429406da9423ce/plugins/CoreConsole/Commands/CoreArchiver.php#L44

https://github.com/matomo-org/matomo/blob/1fcc10569378263775a1f8a3e2429406da9423ce/plugins/CoreConsole/Commands/CoreArchiver.php#L68-L71

when no command line arg is passed.

I'm not completely sure if this is the correct fix, but it seems reasonable to me.

Also changed a similar case with `$segmentIds`. 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Site.php
+++ b/core/Site.php
@@ -422,6 +422,10 @@ class Site
      */
     public static function getIdSitesFromIdSitesString($ids, $_restrictSitesToLogin = false)
     {
+        if (empty($ids)) {
+            return [];
+        }
+
         if ($ids === 'all') {
             return API::getInstance()->getSitesIdWithAtLeastViewAccess($_restrictSitesToLogin);
         }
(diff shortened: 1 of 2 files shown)


core/Site.php
```diff
@@ -422,6 +422,10 @@ public function getCreatorLogin()
      */
     public static function getIdSitesFromIdSitesString($ids, $_restrictSitesToLogin = false)
     {
+        if (empty($ids)) {
+            return [];
```
sgiehl, line 426: Didn't this method already always returns an array, even if `$ids` is empty?
If that isn't the case in PHP 8.1, would be good to know why that is the case.
Afaik `array_filter` and `array_unique` should always return an array...

sgiehl: Ah. I see. The explode would be the problem in PHP 8.1.

sgiehl (approved):
Looks good to me.

PR #17695, opened 2021-06-18
Stronger condition to check NFS

### LINKED ISSUE #12217: Bad filesystem detection (NFS)
Piwik 3.2.0

This is the system where Piwik run
` df -T`

shell

I have no NFS system, but Piwik think so

piwik_nfs_error

Seb35:
### Description:

Issue: #12217 - in the system check page, filesystems like btrfs subvolumes are wrongly recognised as NFS.

NFS filesystems are identified as 'nfs' or 'nfs4' by `df -T`, so the added condition tests the existence of `' nfs'`.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Filesystem.php
+++ b/core/Filesystem.php
@@ -152,6 +152,7 @@ class Filesystem
             // check if filesystem is NFS
             if ($returnCode == 0
                 && count($output) > 1
+                && preg_match('/\bnfs\d?\b/', implode("\n", $output))
             ) {
                 return true;
             }
@@ -161,9 +162,11 @@ class Filesystem
             $output = @shell_exec($command);
             if ($output) {
                 $commandFailed = (false !== strpos($output, "no file systems processed"));
-                $output = explode("\n", trim($output));
+                $output = trim($output);
+                $outputArray = explode("\n", $output);
                 if (!$commandFailed
-                    && count($output) > 1) {
+                    && count($outputArray) > 1
+                    && preg_match('/\bnfs\d?\b/', $output)) {
                     // check if filesystem is NFS
                     return true;
                 }


Seb35:
I tested on my Matomo installation where the directory `tmp` is a btrfs subvolume (and `tmp/sessions` is inside): the check now passes.

core/Filesystem.php
```diff
@@ -161,9 +162,11 @@ public static function checkIfFileSystemIsNFS()
             $output = @shell_exec($command);
             if ($output) {
                 $commandFailed = (false !== strpos($output, "no file systems processed"));
-                $output = explode("\n", trim($output));
+                $output = trim($output);
+                $outputArray = explode("\n", $output);
                 if (!$commandFailed
-                    && count($output) > 1) {
+                    && count($outputArray) > 1
+                    && strpos(' nfs', $output) !== false) {
```
diosmosis: Hi @Seb35, instead of looking for a space prefix, can we check for a word boundary using a regex? Just in case the output has tabs.

diosmosis (changes requested):
Hi @Seb35, thanks for going through the trouble of creating this PR! I've requested a single change, otherwise it looks good to merge!

core/Filesystem.php
Seb35: I changed the condition to `preg_match('/bnfs\d?\b/', $output)` to search "nfs", "nfs4" or any "nfs[0-9]" as a single word.

Seb35:
I did the requested change, it is now more robust.

diosmosis:
@Seb35 I tested locally and it works well, thanks again for creating this PR and contributing some code! We certainly appreciate the effort.

PR #17698, opened 2021-06-21
Add info about custom segments in archiving setting

### LINKED ISSUE #17479: UI: Disabling browser archiving is not clear it does not disable browser archiving for segment
image

We should at least mention that segments are still processed on demand when browser archiving is disabled (no need to mention it when the browser enforce setting is enabled in the config). This is important because it's not expected from the user perspective that this is the case.

Another option could be also to expose yet another setting in the UI to also disable it for segments. However, more settings aren't always the best and it be quite complicated for users to understand so I'm thinking it might be more like wondering why we don't enforce this directly for segments? I suppose this is because some users might have many segments and it would cause too much load to preprocess them all. However, maybe this problem is solved in the wrong way.

Currently, when segment reports are disabled for browser archiving, then all segments will be created as "auto archive/pre process using the cron". We could change this and by default have a report still be processed in the background but now still give them the option to select whether a report should be processed on demand. Then there would be another setting (might already exist) to not allow selecting segments to be processed in real time anymore (since this is needed by many users including us on the cloud where we don't want to process segments in real time).

This could of course break existing behaviour and would maybe need to be done as part of Matomo 5.

The least we could do would be to show the above message that segments will still be processed in real time under circumstances and this can behaviour can be changed in the ini. We don't want to show this message if the ini to force disable browser archiving for segments is disabled (don't show the message when `browser_archiving_disabled_enforce`).

fyi @mattab

justinvelluppillai:
… triggered archiving will not prevent processing of archives when Custom Segments are used, if config browser_archiving_disabled_enforce=0

### Description:

Fixes #17479 

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CoreAdminHome/Controller.php
+++ b/plugins/CoreAdminHome/Controller.php
@@ -299,6 +299,7 @@ class Controller extends ControllerAdmin
         $view->todayArchiveTimeToLive = $todayArchiveTimeToLive;
         $view->todayArchiveTimeToLiveDefault = Rules::getTodayArchiveTimeToLiveDefault();
         $view->enableBrowserTriggerArchiving = $enableBrowserTriggerArchiving;
+        $view->showSegmentArchiveTriggerInfo = !((bool) Config::getInstance()->General['browser_archiving_disabled_enforce']);
 
         $mail = Config::getInstance()->mail;
         $mail['noreply_email_address'] = Config::getInstance()->General['noreply_email_address'];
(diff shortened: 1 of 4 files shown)


plugins/CoreAdminHome/Controller.php
```diff
@@ -299,6 +299,7 @@ private function handleGeneralSettingsAdmin($view)
         $view->todayArchiveTimeToLive = $todayArchiveTimeToLive;
         $view->todayArchiveTimeToLiveDefault = Rules::getTodayArchiveTimeToLiveDefault();
         $view->enableBrowserTriggerArchiving = $enableBrowserTriggerArchiving;
+        $view->showSegmentArchiveTriggerInfo = !((bool) Config::getInstance()->General['browser_archiving_disabled_enforce']);
```
sgiehl: the typecast to bool is actually redundant as the `!` already forces a typecast to bool.

sgiehl: Might actually even be better to use `Rules::isBrowserArchivingAvailableForSegments()` here.

justinvelluppillai: Thanks - I agree and have updated accordingly.

PR #17700, opened 2021-06-22
Adds some more tests for log importer

sgiehl:
### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl:
Note: One of the new tests is currently failing due to https://github.com/matomo-org/matomo-log-analytics/pull/317

diosmosis:
Nice test addition :+1:! Would it also be good to have a replay log for the test? There we can also include strange encoding for the action_name.

Also think it might be a good idea to regenerate OmniFixture at some point after this so we can see how the UI tests respond. What do you think?

sgiehl:
@diosmosis 
replay log should be covered by this: https://github.com/matomo-org/matomo/blob/375dd9f6933dcf7eeb37df9d2d9f44d16f35a277/tests/PHPUnit/Fixtures/ManySitesImportedLogs.php#L278-L287

And the added tests won't change the OmniFixture, as the visits are imported within single tests and not within a Fixture.

diosmosis:
@sgiehl 

> replay log should be covered by this: 

I meant adding the encoding to the reply logs so an encoded value can be added to action_name for tests.

> And the added tests won't change the OmniFixture, as the visits are imported within single tests and not within a Fixture.

Would it be better if it was within a fixture so OmniFixture would pick it up? It might be useful to see if those types of encoded values breaks anything in the UI.

sgiehl:
@diosmosis I won't change anything here for now, as there is more important stuff to work on. Maybe you can have a quick final look and merge the additional tests if they are fine. We can consider changing them later as well.

diosmosis:
:+1: I'll make a new issue for adding encoded data to fixtures

PR #17703, opened 2021-06-23
Web cron archive - allow POST'd token_auth

### LINKED ISSUE #14190: Allow web cron via POST
I just noticed that the [Web Cron Docs](https://matomo.org/docs/setup-auto-archiving/#web-cron-when-your-web-host-does-not-support-cron-tasks) recommends accessing this URL (I just updated it to include https)
https://matomo.your-server.example/path/to/piwik/misc/cron/archive.php?token_auth=XYZ

Sending the admin token via GET isn't ideal, but it seems to be hardcoded:

https://github.com/matomo-org/matomo/blob/7edf461477aa2f732c3e1fda506d7a476d93b62d/misc/cron/archive.php#L60

Would it be possible to update the script to support POST (and mention it in the docs) or maybe even recommend people to directly call `CoreAdminHome.runCronArchiving`?

justinvelluppillai:
### Description:

Fixes #14190

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/misc/cron/archive.php
+++ b/misc/cron/archive.php
@@ -54,7 +54,7 @@ try 'php archive.php --url=http://your.piwik/path'
     $_GET['method'] = 'CoreAdminHome.runCronArchiving';
     $_GET['format'] = 'console'; // will use Content-type text/plain
 
-    if(!isset($_GET['token_auth'])) {
+    if(empty(Piwik\Common::getRequestVar('token_auth', '', 'string'))) {
         echo "
 <b>You must specify the Super User token_auth as a parameter to this script, eg. <code>?token_auth=XYZ</code> if you wish to run this script through the browser. </b><br>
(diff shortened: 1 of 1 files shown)


misc/cron/archive.php
```diff
@@ -54,7 +54,7 @@
     $_GET['method'] = 'CoreAdminHome.runCronArchiving';
     $_GET['format'] = 'console'; // will use Content-type text/plain
 
-    if(!isset($_GET['token_auth'])) {
+    if(empty(Piwik\Common::getRequestVar('token_auth', '', 'string'))) {
```
sgiehl:
```suggestion
    if('' === Piwik\Common::getRequestVar('token_auth', '', 'string')) {
```

As we know that the return value is an empty string if the parameter isn't set in the request, I think it would be more clean to check for that only. `empty` would also check for a couple of more things, which for sure won't hurt, but is simply unneeded.

sgiehl (approved):

PR #17718, opened 2021-06-28
for AWS Aurora users and when using a DB Reader: new INI setting to enable aurora read replica read committed for fixing purge lag performance issue

justinvelluppillai:
### Description:

Allows enabling aurora read replica read committed

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Db.php
+++ b/core/Db.php
@@ -184,6 +184,10 @@ class Db
 
         $db = @Adapter::factory($dbConfig['adapter'], $dbConfig);
 
+        if (!empty($dbConfig['aurora_readonly_read_committed'])) {
+            $db->exec('set session aurora_read_replica_read_committed = ON;set session transaction isolation level read committed;');
+        }
+
         self::$readerConnection = $db;
     }
 


core/Db.php
```diff
@@ -184,6 +184,10 @@ public static function createReaderDatabaseObject($dbConfig = null)
 
         $db = @Adapter::factory($dbConfig['adapter'], $dbConfig);
 
+        if (!empty($dbConfig['aurora_readonly_read_committed'])) {
```
tsteur, line 187: fyi we might still want to update `global.ini.php` here maybe @justinvelluppillai @diosmosis ?

diosmosis: Do we want to document the setting?

tsteur: I reckon it be good. Even if it's just useful mostly for us but then other devs will know what it's about.

justinvelluppillai: I'll document and create a new pull request to add the global.ini.php update

PR #17724, opened 2021-06-29
Added "aurora_readonly_read_committed" setting to  global ini

justinvelluppillai:
### Description:

refs #17718

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -57,10 +57,9 @@ password =
 dbname =
 port = 3306
 
-; If enabled, the reader will use an Aurora MySQL session-level configuration setting,
-; aurora_read_replica_read_committed, to use the READ COMMITTED isolation level on Aurora Replicas.
-; Using this setting can help reduce slowdowns and wasted space that can result from performing long-running
-; queries at the same time as transactions that modify your tables.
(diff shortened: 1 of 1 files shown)


config/global.ini.php
```diff
@@ -57,6 +57,12 @@
 dbname =
 port = 3306
 
+; If enabled, the reader will use an Aurora MySQL session-level configuration setting,
+; aurora_read_replica_read_committed, to use the READ COMMITTED isolation level on Aurora Replicas.
+; Using this setting can help reduce slowdowns and wasted space that can result from performing long-running
+; queries at the same time as transactions that modify your tables.
+aurora_readonly_read_committed =
```
diosmosis, line 63: This only applies to AWS correct? If so, I think we should mention that. And we should perhaps mention how it affects Matomo specifically (ie, long running queries being the archiving queries, and the transactions at the same time being tracking queries (if I understand correctly)).

justinvelluppillai: Thanks for the suggestion - I've updated it with a bit extra info and clarification about it only applying to AWS

diosmosis: It looks good :+1:. One of the UI tests will probably need an update, after that it's good to merge :+1:.

PR #17725, opened 2021-06-29
Multisite links to reports now pass token_auth

### LINKED ISSUE #17001: When embedding a widget with the "Multisites" report (All websites), the links are not working
I embed the Multisite Widget into an Ifram on my page. The iframe is showing nicely all websites with a view permission. So far so good. The problem begins if I click on a website link. I get redirected and have to login. It  would be much better if the auth_token is added to the dashboard deeplinks. This will make it possible to get on overview of all websites and switch between them.

I added the auth_token directly to the website dashboard deeplink and it worked aout of the box.

justinvelluppillai:
### Description:

Fixes #17001 

Pass token_auth parameter so embedded widget links log user in to their matomo dashboard when they click to view full reports.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/MultiSites/angularjs/site/site.controller.js
+++ b/plugins/MultiSites/angularjs/site/site.controller.js
@@ -13,6 +13,7 @@
 
         $scope.period = piwik.period;
         $scope.date   = piwik.broadcast.getValueFromUrl('date');
+        $scope.dashboardUrl = dashboardUrl;
         $scope.sparklineImage = sparklineImage;
         $scope.website.label  = piwik.helper.htmlDecode($scope.website.label);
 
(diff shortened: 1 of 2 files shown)


plugins/MultiSites/angularjs/site/site.controller.js
```diff
@@ -13,20 +13,24 @@
...
         };
 
-        function sparklineImage(website){
-            var append = '';
+        function tokenParam() {
             var token_auth = piwik.broadcast.getValueFromUrl('token_auth');
-            if (token_auth.length) {
-                append = '&token_auth=' + token_auth;
-            }
+            return token_auth.length ? '&token_auth=' + token_auth : '';
```
diosmosis, line 26: I think here we can just use `piwik.token_auth`. Does that work as well? It should always be set too, since when logged in we have to send it on every HTTP request.

diosmosis: I just noticed it was like that before... not sure if there was a reason for that.

diosmosis: I see nevermind, I understand the reasoning now.

diosmosis:
Tested locally, everything works :+1:

PR #17726, opened 2021-06-29
Add possiblity to open sites manager and directly edit a specific site using URL parameter

tsteur:
### Description:

Needed for some plugin. Adding `&editsiteid=\d+` should open the specified site directly. There may be edge cases where multiple sites match a searched number maybe but that be edge case as it likely would be only an issue if > 10 sites matched the specified number. Basically, the URL parameter triggers a search for that number / idsite and the site will be opened directly.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/SitesManager/angularjs/sites-manager/sites-manager-type-model.js
+++ b/plugins/SitesManager/angularjs/sites-manager/sites-manager-type-model.js
@@ -14,10 +14,23 @@
             typesById: {},
             fetchTypeById: fetchTypeById,
             fetchAvailableTypes: fetchAvailableTypes,
-            hasMultipleTypes: hasMultipleTypes
+            hasMultipleTypes: hasMultipleTypes,
+            removeEditSiteIdParameterFromHash: removeEditSiteIdParameterFromHash,
+            getEditSiteIdParameter: getEditSiteIdParameter
         };
 
         return model;
+
+        function getEditSiteIdParameter() {
+            var search = String(window.location.hash).substr('#/'.length);
+            var searchParams = piwik.helper.getArrayFromQueryString(search);
+            if (searchParams.editsiteid && $.isNumeric(searchParams.editsiteid)) {
+                return searchParams.editsiteid;
+            }
+        }
+        function removeEditSiteIdParameterFromHash() {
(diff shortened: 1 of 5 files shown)


diosmosis:
Works for me locally. I'm wondering if we should remove the `editsiteid` parameter after the user finished editing, otherwise the parameter might stick around and open the same site the next time they visit the manage measurables page.

tsteur:
good point @diosmosis 👍  changed it and also refactored slightly to have less duplicated code.

To be able to change the URL I had to move the id from a search query parameter to a hash so I can remove it without reloading the page. When saving or canceling then I remove the hash but not removing the search query as it's tricky to do and not really needed for now as this link will be used extremely rare. People can in worst case always click on left menu again to get all sites I reckon.

plugins/SitesManager/angularjs/sites-manager/sites-manager-type-model.js
```diff
@@ -14,11 +14,24 @@
...
 
+        function getEditSiteIdParameter() {
+            var search = String(window.location.hash).substr('#/'.length);
+            var searchParams = piwik.helper.getArrayFromQueryString(search);
+            if (searchParams.editsiteid && $.isNumeric(searchParams.editsiteid)) {
+                return searchParams.editsiteid;
+            }
+        }
+        function removeEditSiteIdParameterFromHash() {
+            window.location.hash = window.location.hash.replace(/editsiteid=\d+/g, '');
```
diosmosis, line 32: This could end up w/ an extra `&`, correct? Eg, `matomo.com/?date=abc&editsiteid=123` => `matomo.com/?date=abc&`. Added a suggestion to fix.
```suggestion
            window.location.hash = window.location.hash.replace(/[&?]editsiteid=\d+/g, '');
```

tsteur: The hash I'm using currently would be `...#/editsiteid=5` so it shouldn't be a problem?

diosmosis: Oh I see, it's the admin page so we don't have to worry about the hash having a query string, :+1:

PR #17727, opened 2021-06-29
Make anchor links on Admin General Settings page work

tsteur:
### Description:

Currently, links with anchors mostly don't work when linking to a different URL. That's because often the page is loaded async and it's unknown to the system when it's fully loaded. Therefore it is impossible to for example deep link to a specific system setting like https://mydomain/index.php?module=CoreAdminHome&action=generalSettings&editsiteid=5&period=day&date=yesterday&idSite=1#/HeatmapSessionRecording

This logic would make it work to make use of anchor links on the general settings page. There might be a generic solution to it but it's very tricky because the system can never really know whether it should/needs to wait for more requests to finish or whether some more might be coming etc. In many cases using again the `globalAjaxQueue` could work though instead as soon as there are 0 requests active after the load event then trigger this method. Such a solution may be quite tricky though as might then need to factor in whether user maybe already started scrolling etc. Could create a new issue to fix this properly maybe?

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CoreHome/angularjs/anchorLinkFix.js
+++ b/plugins/CoreHome/angularjs/anchorLinkFix.js
@@ -112,4 +112,12 @@
     handleScrollToAnchorAfterPageLoad();
     $(handleScrollToAnchorIfPresentOnPageLoad);
 
+    window.anchorLinkFix = {
+        scrollToAnchorInUrl: function () {
+            // may be called when page is only fully loaded after some additional requests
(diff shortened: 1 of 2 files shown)


diosmosis (approved):
Looks good to me. I don't think there's a proper solution w/o modifying angularjs or watching the dom. As long as it provides value I think it's fine to make sure it's done explicitly.

PR #17736, opened 2021-07-05
avoid large amounts of notifications being added to the session

### LINKED ISSUE #17732: Users can get logged out when many notifications are pending
You can reproduce it this way: 

* https://demo2.matomo.org/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=month&date=2021-06-10&segment=&category=Referrers_Referrers&subcategory=SEOWebVitals_SEOWebVitals
* Click on row evolution on a row
* You will see in the annotation view from the annotation request below message that you are logged out 
image

* If you don't see this error right away then close the popover again and open again the row evolution

It doesn't happen when I apply below change:

```diff
diff --git a/core/Notification/Manager.php b/core/Notification/Manager.php
index a621f27bfd..789e71a7c3 100644
--- a/core/Notification/Manager.php
+++ b/core/Notification/Manager.php
@@ -182,6 +182,7 @@ class Manager
 
     private static function isSessionEnabled()
     {
+        return false;
         return Session::isWritable() && Session::isReadable();
     }
```

Background: The requests on this page currently trigger notifications `Trying to add two strings in DataTable\Row::sumRowArray`... this is why this can't be reproduced on other cases.

I'm thinking this might be a regression from either https://github.com/matomo-org/matomo/pull/17413/files#diff-45b10ec69da2f6ebcecbfef283c2a608427a37f0f4790d0813fc4892894247d5R773 or maybe https://github.com/matomo-org/matomo/commit/ba6be4072538eaf54625ebfcead626107836c818#diff-943fba627b25c62a3fb10c11ee813fce6f225f90f85d1317f05ab07bbfba3e52R524 or some other change. Possible though that this has been an issue for longer and only happens when a big number of notifications are stored or so.

diosmosis:
### Description:

Fixes #17732

Changes:
* impose limit on notification message size when logging to notifications 
*  if in memory notification count exceeds max notification size in session, do not attempt to new ones it to the session
* Detect when session was too large to read and provide warning to user.

### Review

* [ ] Functional review done
* [ ] Code review done


core/Notification/Manager.php
```diff
@@ -111,10 +114,16 @@ private static function checkId($id)
     private static function addNotification($id, Notification $notification)
     {
         self::saveNotificationAcrossUiRequestsIfNeeded($id, $notification);
+        
+        if (count(self::$notifications) > self::MAX_NOTIFICATIONS_IN_SESSION) {
+            return false;
+        }
```
sgiehl: The method `removeOldestNotificationsIfThereAreTooMany` is used before to remove older notifications that exceed the limit. Is this check actually needed? Also it would prevent to add a new notification instead of discarding the oldest.

sgiehl (changes requested):
Might it make sense to add some tests for that?

core/Notification/Manager.php
diosmosis: Yes, the method removes the notifications from the session, not from the in memory array. So the current request can still insert 1000 notifications into the session. Then the next request will fail to read the session, long before this code gets invoked again.

diosmosis:
@sgiehl I added some relevant tests. I didn't add tests for all of DbTable.php or Notification\Manager.php, they were completely untested before. That might be good to do eventually.

core/Notification/Manager.php
sgiehl: But wouldn't it make more sense to discard the oldest notification instead of not adding new ones?

diosmosis: It could, but I think the added value is minimal there. Just because a notification was displayed earlier in the same request, doesn't mean it no longer needs to be displayed to the user. If it gets to the point that there are too many notifications to show to the user, then the user will simply not see all of them. Most notifications get triggered again anyway, so if some are dealt with, the others should also show up.

sgiehl:
> @sgiehl I added some relevant tests. I didn't add tests for all of DbTable.php or Notification\Manager.php, they were completely untested before. That might be good to do eventually.

guess we could create a new issue for creating more/better tests for that

core/Session/SaveHandler/DbTable.php
```diff
@@ -23,6 +23,9 @@
  */
 class DbTable implements Zend_Session_SaveHandler_Interface
 {
+    const SESSION_DATA_MAX_LEN = 65535;
```
sgiehl: Isn't the database field of type `MEDIUMTEXT` and would be able to store 16,777,215 chars?
Or is that limit coming from somewhere else?

diosmosis: @sgiehl this is the column I'm looking at: https://github.com/matomo-org/matomo/blob/4.x-dev/core/Db/Schema/Mysql.php#L278 . If I'm missing something, let me know.

diosmosis:
@sgiehl created https://github.com/matomo-org/matomo/issues/17743 re adding more tests

core/Session/SaveHandler/DbTable.php
tsteur: fyi on demo2 and locally it is mediumtext. not sure if that was changed at some point

diosmosis: @tsteur this is what i see on demo2:

```
MariaDB [74850_demo2]> show columns in `session`;
+----------+--------------+------+-----+---------+-------+
| Field    | Type         | Null | Key | Default | Extra |
+----------+--------------+------+-----+---------+-------+
| id       | varchar(191) | NO   | PRI | NULL    |       |
| modified | int(11)      | YES  |     | NULL    |       |
| lifetime | int(11)      | YES  |     | NULL    |       |
| data     | text         | YES  |     | NULL    |       |
+----------+--------------+------+-----+---------+-------+
4 rows in set (0.001 sec)

MariaDB [74850_demo2]> 
```

can you confirm? I think we're looking at two different things, but I'm not sure what.

tsteur: weird, seeing this now on demo2 the same but locally still seeing mediumtext. BTW maybe we also generally want to change the field to mediumtext?

diosmosis: We could, this would be another option. I don't know if we still want to be able to show more than 30 notifications at a time (I don't personally think there's value in this since no one will read through 30 notifications let alone 100).

diosmosis: Just a note: I'm also seeing `data TEXT` locally.

tsteur: Was more meaning it changing it additionally to prevent similar other issues.

diosmosis: @tsteur I know, that's what I meant by "this would be another option". I'll change it and create an update to change the column. Then we won't need the warning.

I'll also keep the 30 notification limit change that's currently in the PR. If you see anything wrong w/ that let me know.

diosmosis:
@sgiehl @tsteur changed the session column type

core/Version.php
```diff
@@ -20,7 +20,7 @@ final class Version
      * The current Matomo version.
      * @var string
      */
-    const VERSION = '4.4.0-rc1';
+    const VERSION = '4.4.0-rc2';
```
sgiehl: Guess this and the update script should be changed to `rc3` as `rc2` was already released.

diosmosis:
@sgiehl updated

sgiehl (approved):

core/Updates/4.4.0-rc3.php
```diff
@@ -0,0 +1,42 @@
+<?php
+/**
+ * Matomo - free/libre analytics platform
+ *
+ * @link https://matomo.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ *
+ */
+
+namespace Piwik\Plugins\core;
```
sgiehl: The namespace is incorrect. Causing some test failures

sgiehl (changes requested):

sgiehl:
Should be good to merge if tests are passing again now

tsteur:
As it's quite a big PR / change let's maybe merge it after the Matomo 4.4 release (as there have been already two RC releases and the release is due soon)

sgiehl (approved):

sgiehl:
@diosmosis I've merged in the latest changes from `4.x-dev` and updated the name of the update script as well as the version to `4.5.0-b1`. Guess this should be good to merge then...

tests/PHPUnit/Integration/Session/SaveHandler/DbTableTest.php
```diff
@@ -0,0 +1,37 @@
+<?php
+/**
+ * Matomo - free/libre analytics platform
+ *
+ * @link https://matomo.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ *
+ */
+
+namespace PHPUnit\Integration\Session\SaveHandler;
```
sgiehl: shouldn't that be `Piwik\Tests\Integration\...`? Same for the other file

diosmosis:
@sgiehl applied the changes. The UI tests seem to be due to the version bump I think, the "new update available" widget is being displayed (w/o content).

sgiehl:
@diosmosis we are already on `4.4.1`, so the version and update scripts needs to be changed to `4.4.2-b1` or `4.5.0-b1`

diosmosis:
I was sure I did that, maybe I didn't push... 😕

EDIT: Oh I see.

diosmosis:
@sgiehl build is passing (TwoFactorAuth failures unrelated)

sgiehl (approved):

PR #17744, opened 2021-07-08
17417 cookie domain

### LINKED ISSUE #17417: setting the cookie domain even when cookies are disabled for compatibility with "Require Cookie Consent" feature
### Description of issue:
When using a cookie domain such as .example.com and requiring cookie consent causes cookies for both www.example.com and .example.com to be created when cookie consent is given when using MTM
However, when cookie consent is withdrawn only one set of cookies are deleted (.example.com).
The www.example.com cookies remain including the mtm_cookie_consent cookie which causes .example.com cookies to be recreated when refreshing the page.

### Steps to reproduce:
1. Setup MTM container using the following settings:
Enable Link Tracking
Enable Cross Domain Linking
Require Cookie Consent
Cookie domain set to .example.com
Bundle Tracker
Register As Default Tracker

2. Visit the page and execute `_paq.push(['rememberCookieConsentGiven']);`
3. Refresh the page and check cookies created (There should be cookies for both www.example.com and .example.com when reproducing - in my testing I am using a website with a different subdomain of static.example.com)
It is expected here that cookies should only be set to .example.com
image

4. Execute `_paq.push(['forgetCookieConsentGiven']);` and refresh the page. Check cookies, there should be cookies that remained for www.example.com:
image

5. Refresh the page, it should have recreated the cookies for .example.com:
image

### Expected behaviour:
When not using a cookie domain and having Require Cookie Consent enabled, the Cookie Consent works as expected. It only creates cookies for the subdomain that is being viewed and the cookies are deleted when withdrawing cookie consent.

justinvelluppillai:
### Description:

Fixes #17417 by setting the cookie domain (configCookieDomain) even when cookies are disabled.

Also checks whether setCookie is able to successfully create a cookie and if not, log a console message for visibility.

Build js

### Review

* [ ] Functional review done
* [ ] Code review done


justinvelluppillai:
build js

diosmosis:
Tested locally via the test-examples repo, after adding the following code:

```
        _paq.push(['requireCookieConsent']);
        _paq.push(["setCookieDomain", ".matomotestexamples"]);
```

And viewing the page on the `www.matomotestexamples` URL (both mapped via /etc/hosts).

On 4.x-dev, `rememberConsentGiven()` creates cookies (only one set, for the www.matomotestexamples domain). On this branch, the cookies are not created and the message "There was an error setting the cookie" is in the console log. I'm not sure why it works in one but not the other...

justinvelluppillai:
@diosmosis I think that issue is because you are trying to set cookies on a top level domain (.matomotestexamples). Other than that your results are as I expect.

diosmosis:
@justinvelluppillai would you know why it works without your changes and sets the cookie correctly, but doesn't w/ the changes? I don't see anything in the code that could cause this, but it is reproducible.

justinvelluppillai:
@diosmosis it used to set the cookie on the wrong domain, ignoring the setCookieDomain directive. On the new it doesn't ignore the setCookieDomain directive and tries to set the cookie on the domain specified, but you have specified a top level domain on which it can't set a cookie, therefore it doesn't set the cookie on the new. Once you change the test domains to for example www.matomotestexamples.matomo and .matomotestexamples.matomo it will work

diosmosis:
@justinvelluppillai I see, thanks for the explanation!

diosmosis:
@justinvelluppillai fyi, there are still some JSLint errors it seems

justinvelluppillai:
build js

justinvelluppillai:
build js

justinvelluppillai:
build js

justinvelluppillai:
Passing js tests now @diosmosis - took me a few iterations of adding semicolons to get it through the jslint-ing.

Issue #17755, opened 2021-07-12
New console command "core:version" outputs the Matomo version information


ziegenberg:
## Summary

The Symfony built-in commands for the console provides an `--version (-V)` option to "_Display this application version_". It, however, is not just the version information, but it is prefixed with the string `Matomo version`, which makes it harder to parse and compare, e.g. with what `https://builds.matomo.org/LATEST` returns.

I'd like to have a `core:version` command that outputs the bare version information without any additional strings. This comes in handy when automating things using the console for headless updates.

## Your Environment

* Matomo Version: 4.3.1
* PHP Version: 8.0.8
* Server Operating System: RHEL
* Additionally installed plugins:

diosmosis:
Fixes by 17757

diosmosis:
Fixed by #17757

PR #17757, opened 2021-07-12
provide console command for version information (#17755)

ziegenberg:
### Description:

Adds a new console command core:version which outputs only the bare version information (#17755).

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
Thanks @ziegenberg for the contribution! The change looks good to me, and we could eventually add the ability to print out plugin versions and other component versions, so seems like a good change.

PR #17758, opened 2021-07-12
Fix possible warning that session is not active for anonymous user

### LINKED ISSUE #17737: Warning - session_regenerate_id(): Cannot regenerate session id - session is not active
Seeing this log thousands of times (every couple seconds or so) on our demo where anonymous access is enabled.

> 2021-07-06T12:02:45.942+12:00 WARNING Piwik\ErrorHandler[2021-07-06 00:02:45 UTC] [64aa5 demo-proxy.innocraft.cloud] /libs/Zend/Session.php(315): Warning - session_regenerate_id(): Cannot regenerate session id - session is not active - Matomo 4.3.0 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already) [internal function]: Piwik\ErrorHandler::errorHandler(),#1/libs/Zend/Session.php(315),#2/core/Session/SessionInitializer.php(96),#3/core/Session/SessionInitializer.php(30),#4/core/FrontController.php(416),#5/core/dispatch.php(31),#6/index.php(25) | WARNING Piwik\ErrorHandler[2021-07-06 00:02:45 UTC] [64aa5 demo-proxy.innocraft.cloud] /var/www/html/libs/Zend/Session.php(315): Warning - session_regenerate_id(): Cannot regenerate session id - session is not active - Matomo 4.3.0 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already) [internal function]: Piwik\ErrorHandler::errorHandler(),#1/libs/Zend/Session.php(315),#2/core/Session/SessionInitializer.php(96),#3/core/Session/SessionInitializer.php(30),#4/core/FrontController.php(416),#5/core/dispatch.php(31),#6/index.php(25)
-- | -- | --

I think this might be a regression of https://github.com/matomo-org/matomo/pull/17520/files#diff-36c19fe5134cf0a5d0edc32de9cc6583e3a08cc211255c63adc26c7071015618R677

We probably now start the session less often, and when then tries to authenticate anonymous because user is not logged in then it fails to regenerate the session because it wasn't started. So this might be for cases maybe where a token_auth is set but no force_api_session or so.

sgiehl:
### Description:

In case of an API request the session might be started but closed again right away. Therefor we need to check if it's writable to check if we need to initialize the session auth.

fixes #17737

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -409,7 +409,8 @@ class FrontController extends Singleton
                 && Piwik::isUserIsAnonymous()
                 && $authAdapter->getLogin() === 'anonymous' //double checking the login
                 && Piwik::isUserHasSomeViewAccess()
-                && Session::isSessionStarted()) { // only if session was started, don't do it eg for API
+                && Session::isSessionStarted()
+                && Session::isWritable()) { // only if session was started and writable, don't do it eg for API
                 // usually the session would be started when someone logs in using login controller. But in this
                 // case we need to init session here for anoynymous users
                 $init = StaticContainer::get(SessionInitializer::class);
--- a/tests/PHPUnit/Integration/FrontControllerTest.php
+++ b/tests/PHPUnit/Integration/FrontControllerTest.php
@@ -48,7 +48,7 @@ FORMAT;
         $this->assertEquals('error', $response['result']);
 
         $expectedFormat = <<<FORMAT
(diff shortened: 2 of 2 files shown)


diosmosis (approved):
There's a test failure, otherwise it looks good

Issue #9400, opened 2015-12-16
add syslog log_writer


Vebryn:
Hello,

log_writers currently supports screen, database and file.

syslog allows more flexibility in log management. Unix and Linux based distributions manage system logs and applications logs with rsyslog or syslog-ng component. Rsyslog manages and stores logs. It enhances security by centraling log management.

Pros :
- No more file rotation needed
- No more database log table purge
- No more logs into web application dir
- Centralized log management

Is it possible to add syslog log_writer ?

Configuration implementation example :
 [log]
 log_writers[] = syslog

Code implementation example :
 http://php.net/manual/fr/function.syslog.php

Thank you,

Best regards.

tsteur:
:+1: I was waiting for that since I started working with Piwik and am surprised that it is still not there :+1: :+1: :+1: definitely needed for a proper logging setup

mattab:
@Vebryn Thanks for the suggestion, a pull request would be very welcome :-)

mwithheld:
+1 on this please

Ref https://stackoverflow.com/questions/28591433/laravel-monolog-syslog-specify-individual-log-file for another example.
Ref https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/SyslogHandler.php for the 

I can code up a pull request, but I'm not familiar with Matomo code -- can someone point me at where the SyslogHandler should be added, or analogous code for the DB/file logging setup?

diosmosis:
Hi @mwithheld, a pull request would be appreciated, though technically this should be possible currently via dependency injection. You could add the handler by adding the following to the `config/config.php` file (create it if it's not there):

```
<?php

return [
    SyslogHandler::class => \DI\object()->constructor('ident...');

    'log.handlers' => \DI\decorate(function ($previous, ContainerInterface $container) {
        $previous[] = $container->get(SyslogHandler::class);
    }),
];
```

If you want to create a PR that allows setting it via INI config, you'd have to add an entry to `'log.handler.classes'` in `Monolog/config/config.php`.

diosmosis:
Fixed in https://github.com/matomo-org/matomo/pull/17764

PR #17745, opened 2021-07-08
Matomo application logs can now be written in syslog and errorlog

mwithheld:
Issue #9400

### Description:

Add log handlers syslog and errorlog

### Review

* [ ] Functional review done
* [ ] Potential edge cases thought about (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] Usability review done (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] Security review done [see checklist](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] Code review done
* [ ] Tests were added if useful/possible
* [ ] Reviewed for breaking changes
* [ ] Developer changelog updated if needed
* [ ] Documentation added if needed
* [ ] Existing documentation updated if needed

PR #17764, opened 2021-07-13
Matomo application logs can now be written in syslog and errorlog

mwithheld:
### Description:

Issue #9400 Add log handlers syslog and errorlog; Add config for syslog ident; Document options.

### Review

* [ ] Functional review done
* [ ] Code review done


mwithheld:
This is an update to PR #17745

plugins/Monolog/config/config.php
```diff
@@ -97,6 +99,13 @@
     'Piwik\Plugins\Monolog\Handler\FileHandler' => DI\create()
         ->constructor(DI\get('log.file.filename'), DI\get('log.level.file'))
         ->method('setFormatter', DI\get('log.lineMessageFormatter.file')),
+    
+    '\Monolog\Handler\ErrorLogHandler' => DI\create()
+        ->constructor(null, DI\get('log.level.errorlog'))
```
diosmosis: Looking at the ErrorHandler constructor, it seems like this shouldn't be set to null. If you want to only set the level parameter, and use defaults for the rest, you can use:

```
->constructorParameter('level', DI\get('log.level.errorlog'))
```

diosmosis:
@mwithheld Thanks for fixing the issues! I left one other comment. After it's addressed it should be good to merge.

plugins/Monolog/config/config.php
mwithheld: That is one nifty dependency injection system you've got going!  So would it be like this?
```
   '\Monolog\Handler\ErrorLogHandler' => DI\create()
   ->constructorParameter('level', DI\get('log.level.errorlog'))
   ->method('setFormatter', DI\get('log.lineMessageFormatter.file')),
```

Thanks so much for the guidance dizzy!

diosmosis: @mwithheld yes, that looks correct! And yes it is definitely a nifty DI system! The library is called php-di and the creator, @/mnapoli, worked w/ us for a while so our integration works rather well.

mwithheld:
Added commit 14992a8 ErrorLogHandler constructor: Avoid null

plugins/Monolog/config/config.php
```diff
@@ -97,6 +99,13 @@
     'Piwik\Plugins\Monolog\Handler\FileHandler' => DI\create()
         ->constructor(DI\get('log.file.filename'), DI\get('log.level.file'))
         ->method('setFormatter', DI\get('log.lineMessageFormatter.file')),
+    
+    '\Monolog\Handler\ErrorLogHandler' => DI\create()
+        ->constructorParameter('level', DI\get('log.level.errorlog'))
+        ->method('setFormatter', DI\get('log.lineMessageFormatter.file')),
+
+    '\Monolog\Handler\SyslogHandler' => DI\create()
+        ->constructor(DI\get('log.syslog.ident'), 'syslog', DI\get('log.level.syslog')),
```
diosmosis: Apologies for the back and forth, but one last question here: was it intentional not to use `->method('setFormatter', DI\get('log.lineMessageFormatter.file'))` here?

mwithheld:
Good catch - the discussion is welcome :)  That was not intended - I'm just
not used to using GitHub for editing PRs.

diosmosis:
@mwithheld I've merged the PR, thanks for going through the trouble of creating the PR and following up on it! The contribution is greatly appreciated.

mattab:
@diosmosis could you please document this new capability in https://matomo.org/faq/troubleshooting/faq_115/ ? Thanks

diosmosis:
documented

plugins/Monolog/config/config.php
```diff
@@ -97,6 +99,14 @@
     'Piwik\Plugins\Monolog\Handler\FileHandler' => DI\create()
         ->constructor(DI\get('log.file.filename'), DI\get('log.level.file'))
         ->method('setFormatter', DI\get('log.lineMessageFormatter.file')),
+    
+    '\Monolog\Handler\ErrorLogHandler' => DI\autowire()
+        ->constructorParameter('level', DI\get('log.level.errorlog'))
+        ->method('setFormatter', DI\get('log.lineMessageFormatter.file')),
+
+    '\Monolog\Handler\SyslogHandler' => DI\create()
+        ->constructor(DI\get('log.syslog.ident'), 'syslog', DI\get('log.level.syslog'))
```
orlitzky, line 108: Is this logging to the "syslog" facility? I would guess that "user" is most appropriate, but "syslog" is definitely not correct =)

(Discovered while trying to match on the facility and program name using syslog-ng.)

Lazy reference: https://en.wikipedia.org/wiki/Syslog#Facility

mwithheld: You are correct this should be 'user'.
Should probably use 
```
->constructorParameter('ident'=>DI\get('log.syslog.ident'))
->constructorParameter('level'=>DI\get('log.level.syslog'))
```

diosmosis:
Hi @mwithheld, thanks for creating the change so quickly! Would you be able to put it in a new pull request? This one has already been merged so it's no longer useful to push here.

orlitzky:
Ok, thanks! And thanks for adding syslog support in the first place. Obviously, I was waiting for it =)

Issue #17202, opened 2021-02-07
Matomo can be tricked to record spoofed X-Forwarded-For IPs


timdream:
## Problem

Considering the current implementation: https://github.com/matomo-org/matomo/blob/a31fd86b64c2c7d2deadcfca71540fb4b0152c10/core/IP.php#L99-L125

With the following setup:

```
[client] -> [reverse proxy 1] -> [reverse proxy 2] -> [matomo Nginx]
```

Since each proxy appends the IP of the incoming connection according to [RFC7239](https://tools.ietf.org/html/rfc7239), `X-Forwarded-For` will be this by the time the connection reaches Nginx (1):

```
X-Forwarded-For: (client), (proxy 1)
```

Matomo currently extracts the first IP from the list, since #10404. This is correct until you realized the "client" can pretended to be a proxy as well, by including its own `X-Forwarded-For` header, like this:

```shell
curl -i -H"X-Forwarded-For: 8.8.8.8" "http://example.com/matomo/matomo.php?..."
```

When this happens, proxy 1 will be faithfully preserve the header passed in and append the client IP after it (2):

```
X-Forwarded-For: 8.8.8.8, (client), (proxy 1)
```

And Matomo will extract `8.8.8.8` as the client IP.

### Solution 1: Trust the "transparent proxies"

If Matomo decided the first "client" IP can always be trusted, we can still extract the first IP address, and this issue can be closed as "works as intended." But it will in risk of having the analytics data being polluted. **Brutal force login detection can be tricked too.**

### Solution 2: Extract the last IP address

This would involve reverting #10404 and go back to extract the last IP address. Since the original `proxy_ips[]` configuration is still intact, when facing a header like (2), above, a correctly configured Matomo will be able to exclude the "proxy 1" IP address.

## Notes

* There can be a solution 1.1 where Mamoto will rely on proxy 1 (the first public-facing proxy) to remove the spoofed header. However this is not always possible (not possible for Apache AFAIK)
* It is unclear to me what other client IP headers (`HTTP_CF_CONNECTING_IP`, `HTTP_CLIENT_IP`, etc) behaves and if Matomo have to do things differently between these and the `X-Forwarded-For` header.
* #7060 is a staled discussion related to this, but the discussion talks about the header processing in the context before #10342.
* #16379 have since promoted `proxy_ips[]` pretty well in the documentation. We may consider doubling down on the documentation effort when we pick either solution for this issue.
[...]

PR #17765, opened 2021-07-13
add config to read the last IP address in the list of proxies rather than the first

diosmosis:
### Description:

Fixes L3-123.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -641,8 +641,10 @@ multi_server_environment = 0
 ; By enabling this flag the header HTTP_X_FORWARDED_URI will be considered for the current script name.
 proxy_uri_header = 0
 
-; If set to 1 we use the last IP in the list of proxy IPs when determining the client IP.
-proxy_ip_read_last_in_list = 1
+; If set to 1 we use the last IP in the list of proxy IPs when determining the client IP. Using the last IP can be more
+; secure when using proxy headers in combination with a load balancer. By default the first IP is read according to RFC7239
+; which is required when the client sends the IP through a proxy header as well as the load balancer.
+proxy_ip_read_last_in_list = 0
 
 ; Whether to enable trusted host checking. This can be disabled if you're running Matomo
 ; on several URLs and do not wish to constantly edit the trusted host list.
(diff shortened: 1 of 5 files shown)


tsteur:
Before merging once the config is final be good to document this security feature in https://matomo.org/faq/how-to-install/faq_98/ and maybe in https://matomo.org/docs/security-how-to/ we could as well mention that they might want to enable this config if they are using a load balancer. 

@mattab you will want to document this in the enterprise guide probably as well.

config/global.ini.php
```diff
@@ -641,6 +641,9 @@
 ; By enabling this flag the header HTTP_X_FORWARDED_URI will be considered for the current script name.
 proxy_uri_header = 0
 
+; If set to 1 we use the last IP in the list of proxy IPs when determining the client IP.
+proxy_ip_read_last_in_list = 1
```
sgiehl: Shouldn't we make `0` as default to keep bc. Or at least set this option automatically to `0` on update, to ensure it won't break any instances. Could maybe only do that if `proxy_client_headers` is set.

diosmosis: It was specified to set it to 1 in the issue (I assume it adds security). cc @tsteur

tsteur: By default we want to keep existing behaviour and follow the RFC so we'd set it to `0` as otherwise it would be a breaking change. 

We probably wouldn't leave the default at 1 and change it for existing installs as there are often random issues with it and it would require an additional step when setting up proxy headers. I reckon we could create an issue though to do this in Matomo 5 and put it directly into the Matomo 5 milestone.

tsteur: fyi could tweak the description a bit.Like

> ; If set to 1, the last IP in the list of proxy IPs will be used when determining the client IP. Using the last IP can increase the security when you are using proxy headers in combination with a load balancer. By default, the first IP is read according to RFC7239 which may be required in some cases when the client sends the IP through a proxy header as well as the load balancer.

diosmosis:
@sgiehl / @tsteur applied review feedback

config/global.ini.php
tsteur: FYI put https://github.com/matomo-org/matomo/issues/17202 in Matomo 5 milestone

sgiehl (approved):
looks good to me now

mattab:
> @mattab you will want to document this in the enterprise guide probably as well.

Sounds good, let me know when the user guide is updated and i'll copy the information in the PDF enterprise guide.

tsteur:
@mattab added a bit to the docs

PR #17768, opened 2021-07-14
if a failed login attempt comes from an API request, mark it as from the API in the brute force log

diosmosis:
### Description:

For debugging purposes it could be helpful to see what entries are from API requests (as opposed to login requests and tracker API requests).

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Login/Security/BruteForceDetection.php
+++ b/plugins/Login/Security/BruteForceDetection.php
@@ -8,6 +8,7 @@
  */
 namespace Piwik\Plugins\Login\Security;
 
+use Piwik\API\Request;
 use Piwik\Common;
 use Piwik\Container\StaticContainer;
 use Piwik\Date;
@@ -16,6 +17,7 @@ use Piwik\Option;
 use Piwik\Plugins\Login\Emails\SuspiciousLoginAttemptsInLastHourEmail;
 use Piwik\Plugins\Login\Model;
 use Piwik\Plugins\Login\SystemSettings;
+use Piwik\SettingsServer;
 use Piwik\Updater;
 use Piwik\Version;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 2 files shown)


sgiehl (dismissed):

plugins/Login/Security/BruteForceDetection.php
```diff
@@ -68,6 +71,10 @@ public function isEnabled()
 
     public function addFailedAttempt($ipAddress, $login = null)
     {
+        if (empty($login) && Request::isRootRequestApiRequest()) {
+            $login = self::API_LOGIN_PLACEHOLDER;
```
tsteur, line 75: would this cause any issue when there is an API call and it maybe searches for `isUserLoginBlocked(null)`? Could there be other side effects? 

Maybe generally when login is null it's likely API anyway?

diosmosis: > would this cause any issue when there is an API call and it maybe searches for isUserLoginBlocked(null)? Could there be other side effects?

I don't think it's expected to search for `isUserLoginBlocked(null)`. The query that searches for logins uses `login = ?` not `login is NULL` so I wouldn't expect that work anyway.

> Maybe generally when login is null it's likely API anyway?

I think it could also be a tracker request?

tsteur: Possible re tracking request 👍 . Generally, I think it be better to add a column `source` or something similar just to prevent random side effects or bugs. If I see this right indeed `isUserLoginBlocked` is only called when there is a login so far. It just could also cause maybe side effects in the future as it may not be expected behaviour to have a login set where there is none and generally it's maybe not too important of an issue to introduce possible side effects in the future?

diosmosis: A `source` column would probably be better. We could put the api method/module there as well. It wouldn't take too long to add, I'll look into it tomorrow.

tsteur: btw @diosmosis as the issue is not too high priority we could also just close the PR and work on it another day

diosmosis: It won't take very long, I'll get to it when I have some time.

tsteur:
not needed for now

PR #17771, opened 2021-07-14
If table prefix is not specified, the database abilities diagnostic can fail

### LINKED ISSUE #17762: Problem with SQL table after migrating from Matomo 3 to Matomo 4
Hello,
I don't know how this happened, but I see a problem after the upgrade from Matomo 3 to Matomo 4, the upgrade from Ubuntu 14.4 to Ubuntu 20.04 was performed as well.

## Current Behavior
The GUI reports that there is an error in the SQL syntax:

Erro: Mysqli prepare error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'option (option_name,option_value) VALUES (?,?)' at line 1

I don't see the same reports when I run diagnostics via './console diagnostics: run --all',  just from the GUI.  

Apparently I have no problem, everything works fine, but I don't like this report. I can see that there is a table called 'option' in the Matomo database, but when I list the table structure, I can see the same error shown by the GUI, the same error happened when I list the data in this table too, is this table important? I didn't find her in the documentation.

## Table OPTION
mysql> desc option;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'option' at line 1

mysql> select * from option;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'option' at line 1

I don't know if this table is the cause of the problem, but it seems to have some relationship, could someone help me with this error?

 Thank you so much.

diosmosis:
### Description:

If not table prefix is specified, the database abilities diagnostic can fail when it tries to batch insert into the option table. Fixed w/ backticks.

Fixes #17762

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Db/BatchInsert.php
+++ b/core/Db/BatchInsert.php
@@ -36,7 +36,7 @@ class BatchInsert
 
         foreach ($values as $row) {
             $row = array_values($row);
-            $query = "INSERT $ignore INTO " . $tableName . "
+            $query = "INSERT $ignore INTO `" . $tableName . "`
 					  $fieldList
 					  VALUES (" . Common::getSqlStringFieldsArray($row) . ")";
             Db::query($query, $row);
--- a/core/Db/BatchInsert.php
+++ b/core/Db/BatchInsert.php
@@ -31,6 +31,7 @@ class BatchInsert
      */
     public static function tableInsertBatchIterate($tableName, $fields, $values, $ignoreWhenDuplicate = true)
     {
+        $tableName = preg_replace('/[^a-zA-Z\d_-]/', '', $tableName);
         $fieldList = '(' . join(',', $fields) . ')';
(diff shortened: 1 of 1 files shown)


justinvelluppillai (commented):
I'm just getting started reviewing so I'll start participating by leaving comments without approvals so that someone else can also review.

It looks like this bug is because 'option' is a reserved word so the query needs the backticks to identify option as a table/column name. I guess that means tableInsertBatchIterate function will now work for table names with eg spaces or other reserved words but you have disallowed that with your preg_replace, which means spaces etc wouldn't work, which they already wouldn't have prior to this change.

Seems like a good fix to me.

sgiehl (approved):

PR #17772, opened 2021-07-14
check plugin exists in the filesystem on plugin activate command

### LINKED ISSUE #17770: activating non-existing plugin via console plugin:activate results in uncaught exception and plugin is listed as installed afterwards
## Expected Behavior

When trying to activate a non-existing plugin via ` console plugin:activate`, it should report the not existing plugin in a nice message and do not throw an uncaught exception. Also, it should not be listed in the list of installed plugins afterwards.

## Current Behavior

When trying to activate a non-existing plugin via ` console plugin:activate`, it throws an uncaught exception.
```
php console plugin:activate SecurityInfo2
ERROR [2021-07-14 15:45:55] 477937 Uncaught exception: /opt/matomo/core/Plugin/Manager.php(684): Plugin 'SecurityInfo2' cannot be found in the filesystem in plugins/ directory.
 
 [Exception] 
 Plugin 'SecurityInfo2' cannot be found in the filesystem in plugins/ directory. 
 
plugin:activate [plugin1] ... [pluginN]
```

Also, it's now listed in the list of installed plugins afterwards.

```
php console plugin:list 
+----------------------+-------------------+---------------+
| Plugin               | Core or optional? | Status        |
+----------------------+-------------------+---------------+
| Diagnostics          | Core              | Activated     |
| Login                | Core              | Activated     |
| CoreAdminHome        | Core              | Activated     |
| UsersManager         | Core              | Activated     |
| SecurityInfo         | Optional          | Activated     |
| SecurityInfo2        | Optional          | Not activated |
| LogViewer            | Optional          | Activated     |
+----------------------+-------------------+---------------+

```

1. call `php console plugin:activate pluginName` for a plugin not in `matomo/plugins`

## Context

Automating plugin installs via console for ansible playbooks.

## Your Environment

* Matomo Version: 4.3.1
* PHP Version: 8.0
* Server Operating System: CentOS Linux release 8.4.2105
* Additionally installed plugins: -

* Browser: -
* Operating System: -

diosmosis:
### Description:

If a plugin doesn't have a Plugin class, the plugin manager just uses `new Plugin($pluginName)`. Calling `getMissingDependencies()` immediately after triggers the plugin install code path, which adds the nonexistent plugin to the INI config. Now we check and provide a clear error message in the commands, and make sure we only try to load plugins that exist in the filesystem.

Fixes #17770

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Plugin/Manager.php
+++ b/core/Plugin/Manager.php
@@ -1185,10 +1185,6 @@ class Manager
             return $this->loadedPlugins[$pluginName];
         }
 
-        if (!$this->isPluginInFilesystem($pluginName)) {
-            throw new \Exception("Plugin $pluginName does not exist.");
-        }
-
         $newPlugin = $this->makePluginClass($pluginName);
 
         $this->addLoadedPlugin($pluginName, $newPlugin);
--- a/core/Plugin/Manager.php
+++ b/core/Plugin/Manager.php
@@ -1184,6 +1184,11 @@ class Manager
         if (isset($this->loadedPlugins[$pluginName])) {
             return $this->loadedPlugins[$pluginName];
         }
+
+        if (!$this->isPluginInFilesystem($pluginName)) {
+            throw new \Exception("Plugin $pluginName does not exist.");
+        }
+
         $newPlugin = $this->makePluginClass($pluginName);
 
         $this->addLoadedPlugin($pluginName, $newPlugin);
--- a/plugins/CorePluginsAdmin/Commands/DeactivatePlugin.php
+++ b/plugins/CorePluginsAdmin/Commands/DeactivatePlugin.php
@@ -38,6 +38,11 @@ class DeactivatePlugin extends ConsoleCommand
                 continue;
             }
 
+            if (!$pluginManager->isPluginInFilesystem($plugin)) {
+                $output->writeln("<error>Cannot find plugin files for $plugin.</error>");
+                continue;
+            }
+
             $pluginManager->deactivatePlugin($plugin);
 
             $output->writeln("Deactivated plugin <info>$plugin</info>");
--- a/plugins/CorePluginsAdmin/Commands/DeactivatePlugin.php
+++ b/plugins/CorePluginsAdmin/Commands/DeactivatePlugin.php
@@ -38,11 +38,6 @@ class DeactivatePlugin extends ConsoleCommand
                 continue;
             }
 
-            if (!$pluginManager->isPluginInFilesystem($plugin)) {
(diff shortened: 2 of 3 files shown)


core/Plugin/Manager.php
```diff
@@ -1184,6 +1184,11 @@ public function loadPlugin($pluginName)
         if (isset($this->loadedPlugins[$pluginName])) {
             return $this->loadedPlugins[$pluginName];
         }
+
+        if (!$this->isPluginInFilesystem($pluginName)) {
```
tsteur: @diosmosis I think it might be good if we undo this particular change just to not break any possible use cases or cause any side effects. We can always add it later if needed for some reason. I just want to prevent adding any possible side effects or causing any issues when there isn't really a problem (even the original issue itself is very minimal and very rarely any issue etc)

diosmosis: :+1: this is what I was thinking at first. However, I figured if there's no plugin folder, then we wouldn't want it to work. I can revert it though.

tsteur: 👍  I can't really think of much of an issue either but you never know. For starters it may slow down Matomo quite a bit as it would read the filesystem for every plugin (of which there can be many). This can make Matomo load quite slow and depending on the disk cause bottlenecks etc

diosmosis: @tsteur updated

plugins/CorePluginsAdmin/Commands/DeactivatePlugin.php
```diff
@@ -38,6 +38,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
                 continue;
             }
 
+            if (!$pluginManager->isPluginInFilesystem($plugin)) {
+                $output->writeln("<error>Cannot find plugin files for $plugin.</error>");
+                continue;
+            }
```
sgiehl: Does that check make sense here?
If a plugin name is provided that was never installed (is not included in plugin list in config.ini.php), the method will already fail with `already deactivated`. If it was installed, but only removed from file system it might make sense to allow a deactivation so it gets removed from the plugin list in config.ini.php and it's no longer tried to load it anymore.

sgiehl (changes requested):

plugins/CorePluginsAdmin/Commands/DeactivatePlugin.php
diosmosis: :+1: makes sense to me

diosmosis:
@sgiehl updated

sgiehl (approved):

PR #17789, opened 2021-07-18
Make error messages not leak potentially sensitive information when db connection fails

justinvelluppillai:
### Description:

Improve DB connection issue messages and provide framework for later improvements if any.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Db/Adapter.php
+++ b/core/Db/Adapter.php
@@ -9,6 +9,7 @@
 namespace Piwik\Db;
 
 use Zend_Db_Table;
+use Piwik\Piwik;
 
 /**
  */
@@ -145,8 +146,8 @@ class Adapter
     {
         $safeMessageMap = array(
             // add any exception search terms and their replacement message here
-            'MySQL server has gone away'    => 'Database server has gone away',
-            'Access denied'                 => 'Database access denied'
+            'MySQL server has gone away'    => Piwik::translate('General_ExceptionDatabaseUnavailable'),
+            'Access denied'                 => Piwik::translate('General_ExceptionDatabaseAccess')
         );
 
         foreach ($safeMessageMap as $search_term => $safeMessage) {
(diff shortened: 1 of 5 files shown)


core/Db/Adapter.php
```diff
@@ -45,14 +46,20 @@ public static function factory($adapterName, & $dbInfos, $connect = true)
...
+            try {
+                $adapter->getConnection();
+
+                Zend_Db_Table::setDefaultAdapter($adapter);
+                // we don't want the connection information to appear in the logs
+                $adapter->resetConfig();
+            } catch(\Exception $e) {
+                // we don't want the exception to leak information
+                $msg = self::safeExceptionMessage($e->getMessage());
+                throw new \Exception($msg);
```
diosmosis: Would it be better to rethrow the exception if nothing is matched in `safeExceptionMessage()`? This would preserve the original stack trace.

justinvelluppillai: Yes good idea, thanks

lang/en.json
```diff
@@ -177,6 +177,8 @@
         "ExceptionDatabaseVersion": "Your %1$s version is %2$s but Matomo requires at least %3$s.",
         "ExceptionDatabaseVersionNewerThanCodebase": "Your Matomo codebase is running the old version %1$s and we have detected that your Matomo Database has already been upgraded to the newer version %2$s.",
         "ExceptionDatabaseVersionNewerThanCodebaseWait": "Maybe your Matomo administrators are currently finishing the upgrade process. Please try again in a few minutes.",
+        "ExceptionDatabaseUnavailable": "Database server has gone away",
```
tsteur: @justinvelluppillai be great to still use `MySQL server has gone away` so when people search for a solution they will find our FAQ (it's the most viewed FAQ we have AFAIK). Also it helps people find other more relevant content when they search for the error.

diosmosis:
tested locally, looks good to me :+1:

PR #17798, opened 2021-07-20
Set CSP header to prevent XSS

### LINKED ISSUE #17773: Matomo should be setting content security policy to prevent some XSS
This is for the Matomo UI (not tracking). For better security we shouldn't leave it up to users to configure content security policy. We should set a CSP to only allow requests to the current domain. Note: When Matomo is behind a load balancer etc then it might not know the correct domain it is on so would need to use trusted hosts maybe (or in some cases).

Not sure if there are any external domains the UI requests. In any case plugins should allow to configure extra rules. These rules would probably be applied to all Matomo requests (as eg if a widget does external requests by the time the widget is requested then it cannot change the CSP). Additionally, for better security, we may give plugins the ability to enrich the CSP for a specific controller/action request. This way they would only change the CSP for the current request but not other pages. This can be useful if a plugin renders an entire page.  

We may also need to allow extra configurations through the config file.

We could also be making use of `nonce-source` which may be quite good but not sure it fully works or whether some other plugins may add script tags dynamically etc where this could get difficult.

justinvelluppillai:
### Description:

Fixes #17773 by setting a Content-Security-Policy which is fairly loose at this stage.

### Review

* [ ] Functional review done
* [ ] Code review done


Findus23:
I think we should only be using `Content-Security-Policy-Report-Only` for at least one or two releases as I'm sure there are a lot of people with weird setups where the CSP will break in fascinating ways. 
I still think it is worth creating a secure one for Matomo as it will make Matomo a lot more secure, but also provide an option to fall back to Report-Only or disable it, for setups where using one is not possible.

tsteur:
👍  be good indeed to have such an option to disable it, or using report-only. And could indeed start with report-only. Maybe in dev mode we would always from beginning use the real one so it's easier to make things compatible? might not make sense though just some thoughts. I suppose this would give some people more time to adjust their setup if needed. They might not notice these reports but at least they might notice it in the changelog maybe?

justinvelluppillai:
We could have a setting to choose between enabled, disabled, or report only mode, defaulting to disabled or report only.

justinvelluppillai:
I've added a couple of settings, one to enable/disable and one to specify a report-uri. CSP is disabled by default for now, specifying a report-uri will switch it to Content-Security-Policy-Report-Only

tsteur:
👍  @justinvelluppillai it seems report-uri is deprecated see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri ? Could maybe remove this feature and then have only one setting for the three different modes (disabled, report, enabled)?

Or seems it might be replaced by https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to ? I don't know how commonly used it is. If it's not commonly used that often someone could maybe add this feature as a plugin.

Findus23:
While report-uri seems to be deprecated, it is the only parameter that is actually supported by browsers:
https://caniuse.com/mdn-http_headers_csp_content-security-policy_report-uri

report-to is not supported by any non-chrome browser: https://caniuse.com/mdn-http_headers_report-to

That said, `Content-Security-Policy-Report-Only` makes also sense without a report-uri (especially during development) as it will log violations in the console.

justinvelluppillai:
Huh, I didn't notice that in my investigations. A couple of possible approaches:

1. Specify both report-uri and report-to as described here https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri
2. Join @Findus23 and @tsteur's solutions: have three modes only and exclude the report-uri/report-to fields and let it log to console.

Both trivial to implement, option 1 provides more possible configurations but I lean towards option 2 for simplicity, if everyone agrees?

Findus23:
I think we could either add two settings (one switching between on, report-only and off and one specifying the report-uri (or being empty)) or use three settings (like e.g. in https://github.com/Chocobozzz/PeerTube/blob/develop/config/default.yaml#L151-L154)

tsteur:
I suppose using two settings could work and let us extend it better in the future if we were to decide to add more features (which isn't currently needed).

So something like you mention

```
csp_enabled = 0/1
csp_report_only = 0/1 . # Only does something if csp is enabled
```

tsteur:
@justinvelluppillai it seems like the Page Overlay feature doesn't work yet when the URL is from a different page? I just tested it and it won't load the site.

justinvelluppillai:
@tsteur with the default settings we've chosen (report-only) it should still work as normal but report some issues. I can modify the Page Overlay feature to use a different (or no) CSP. I will take a look at getting the Page Overlay feature running locally then see what's best.

tsteur:
fyi I tested with in report mode (not report-only).

justinvelluppillai:
@sgiehl it'd be great if you could give this a check over and if all ok we can merge as it won't have too much impact until we enable it.

sgiehl:
@justinvelluppillai Sure. Feel free to simply use the github feature to request for reviews in the future if you need one ;-)

tests/PHPUnit/Unit/SecurityPolicyTest.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public function testDefaultSecurityPolicy() {
+        $expectedDefault = "Content-Security-Policy-Report-Only: " . $this->defaultPolicy;
+        $this->assertEquals($expectedDefault, $this->securityPolicy->createHeaderString());
+    }
+
+    public function testDefaultEnabledSecurityPolicy() {
+        $this->generalConfig['csp_report_only'] = 0;
+        $this->securityPolicy = new SecurityPolicy();
```
sgiehl: this line seems redundant, as it's already done in `setUp()`
```suggestion
```

tests/PHPUnit/Unit/SecurityPolicyTest.php
```diff
@@ -0,0 +1,84 @@
...
+
+        $expectedDefaultEnabled = "Content-Security-Policy: " . $this->defaultPolicy;
+        $this->assertEquals($expectedDefaultEnabled, $this->securityPolicy->createHeaderString());
+    }
+
+    public function testDisabledSecurityPolicy() {
+        $this->generalConfig['csp_enabled'] = 0;
+        $this->securityPolicy = new SecurityPolicy();
```
sgiehl: Same here
```suggestion
```

core/View/SecurityPolicy.php
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Constructor.
+     */
+    public function __construct() {
+        $this->policies['default-src'] = "'self' 'unsafe-inline' 'unsafe-eval'";
+
+        $generalConfig = Config::getInstance()->General;
+        $this->cspEnabled = $generalConfig['csp_enabled'];
+        $this->reportOnly = $generalConfig['csp_report_only'];
+    }
```
sgiehl: Not necessarily needed, but we maybe could use dependency injection were easily possible:

```suggestion
    public function __construct(Config $config) {
        $this->policies['default-src'] = "'self' 'unsafe-inline' 'unsafe-eval'";

        $generalConfig = $config->General;
        $this->cspEnabled = $generalConfig['csp_enabled'];
        $this->reportOnly = $generalConfig['csp_report_only'];
    }
```

sgiehl (changes requested):
Left a couple of minor comments. Btw. might also be good to merge in the latest changes from `4.x-dev` to see if tests are passing. Currently the UI tests are showing various failures, that actually look unrelated.

tests/PHPUnit/Unit/SecurityPolicyTest.php
justinvelluppillai: setup() does it but needs to be done again if changing the config to make sure the securityPolicy object picks up those changes.

tests/PHPUnit/Unit/SecurityPolicyTest.php
justinvelluppillai: same here

tests/PHPUnit/Unit/SecurityPolicyTest.php
sgiehl: Ah. I see. The values are picked up once in `__construct`...

tests/PHPUnit/Unit/SecurityPolicyTest.php
```diff
@@ -0,0 +1,84 @@
...
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        // set the default config explicitly so tests can change them without needing to reset them
+        $this->generalConfig =& Config::getInstance()->General;
+        $this->generalConfig['csp_enabled'] = 1;
+        $this->generalConfig['csp_report_only'] = 1;
+
+        $this->securityPolicy = new SecurityPolicy();
```
sgiehl: Due to the DI change this now needs to be `new SecurityPolicy(Config::getInstance())` or `StaticContainer::get(SecurityPolicy::class)`
same for the instances below.

justinvelluppillai: True thanks.

sgiehl (approved):
The code looks good now 👍

PR #17808, opened 2021-07-23
Ensure to use correct default value for segment in scheduled reports

### LINKED ISSUE #15335: email reports: segment doesn't save
When configuring an e-mail report, the selected segment (in my case "all visits") is not saved. When the report is opened again, the segment is still blank.

sgiehl:
### Description:

`report.idsegment` might be `null` when received from API. To ensure the default value is selected in that case, we need to convert that to an empty string.

fixes #15335

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/ScheduledReports/angularjs/manage-scheduled-report/manage-scheduled-report.controller.js
+++ b/plugins/ScheduledReports/angularjs/manage-scheduled-report/manage-scheduled-report.controller.js
@@ -63,6 +63,7 @@
                 'period': ReportPlugin.defaultPeriod,
                 'hour': ReportPlugin.defaultHour,
                 'reports': [],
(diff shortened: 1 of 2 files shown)


diosmosis:
I don't think this is specifically related, but testing locally when I tried to update a report I got: `Please specify a value for 'reports'.`. The HTTP request doesn't appear to have that parameter but ScheduledReports.updateReport requires it.

sgiehl:
@diosmosis That works for me. Did you maybe try to edit a report where no reports were chosen in the list? Maybe they became unavailable 🤔

diosmosis:
@sgiehl yep, that's what happened, not sure how though.

diosmosis (approved):

PR #17809, opened 2021-07-23
Set proper permissions for github actions

sgiehl:
### Description:

Actually I'm not totally sure if the changes might be to restrictive or if we could restrict some actions a bit more. Maybe write access to pull requests/issues would not be needed, if only commits are added and no comments. 

Note: Setting permissions to `none` would actually not be needed, as that's the default as soon as another permission is set. I left them in, so it's clear which permissions are available and set.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/.github/workflows/buildtrackerjs.yml
+++ b/.github/workflows/buildtrackerjs.yml
@@ -4,6 +4,18 @@ on:
   issue_comment:
     types: [created]
 
+permissions:
+  actions: read
+  checks: none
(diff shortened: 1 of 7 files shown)


diosmosis (approved):
makes sense

sgiehl:
I'll merge this now. If any action fails in the future we can check if the permissions are too restrictive later...

Issue #17738, opened 2021-07-06
Don't log an error when not supported browser is used


tsteur:
Seeing thousands of these logs on our cloud:

> ERROR Piwik\ExceptionHandler[2021-07-05 23:57:59 UTC] [52b29 demo-proxy.innocraft.cloud] /core/SupportedBrowser.php(68): <p><b>Your browser is not supported.</b></p><p>Due to security issues your browser is not supported. Please upgrade to a newer version.</p>  #0 /core/SupportedBrowser.php(54): Piwik\SupportedBrowser::throwException()  #1 /core/FrontController.php(423): Piwik\SupportedBrowser::checkIfBrowserSupported()  #2 /core/dispatch.php(31): Piwik\FrontController->init()  #3 /index.php(25): require_once('/c...')  #4 {main} | ERROR Piwik\ExceptionHandler[2021-07-05 23:57:59 UTC] [52b29 demo-proxy.innocraft.cloud] /core/SupportedBrowser.php(68): <p><b>Your browser is not supported.</b></p><p>Due to security issues your browser is not supported. Please upgrade to a newer version.</p> #0 /core/SupportedBrowser.php(54): Piwik\SupportedBrowser::throwException() #1 /core/FrontController.php(423): Piwik\SupportedBrowser::checkIfBrowserSupported() #2 /core/dispatch.php(31): Piwik\FrontController->init() #3 /index.php(25): require_once('/c...') #4 {main}

This may be caused for example by our monitoring tools or so. Could we adjust the logic to not log such an error when this exception happens?

I suppose we can catch `NotSupportedBrowserException` and make it not log. I'm thinking this would need to be done in `ExceptionHandler::dieWithHtmlErrorPage()`.

mpdude:
Also, it should not cause a `500` status code for the user, but rather something from the `4xx` range.

bisoldi:
We're now seeing this as well.  I'm confused as to how this error comes to be exactly.  If I can enumerate my questions regarding this....

- What is the "security issue exactly"?  
- Is this an exception that's being thrown within the browser, transmitted to the server and then logged out?  
- If the Matomo client-side doesn't support the browser, then how is it even able to transmit the error?

Some kind of call made its way from the browser to the server, in order to log this error.  Why does Matomo reject it in the form of a server-side error log rather than simply tracking it as a metric?

sgiehl:
[...]

PR #17812, opened 2021-07-25
Don't log an error when not supported browser is used

mwithheld:
Issue #17738 Don't log an error when not supported browser is used

### Description:

Issue #17738 Don't log an error when not supported browser is used

To reproduce the error (before the changes in this ticket):

- Monitor the server (e.g. Apache) error log, e.g. `tail -f error_log` 
- Do a GET on matomo with an unsupported browser version e.g. Firefox < version 32, e.g.
`wget -U "Mozilla/5.0 (Windows NT 5.1; rv:2.0) Gecko/20100101 Firefox/4.0" --no-check-certificate https://127.0.0.1/matomo/`

To test the fix:

- Repeat the steps to reproduce above - you should not see the message in the error log.
- In config.ini.php, set loglevel=DEBUG (the default is WARN):
```
[log]
log_level = DEBUG
```
- Repeat the steps to reproduce above - you should now see the message in the error log.

### Review

* [ ] Functional review done
* [ ] Code review done


mwithheld:
It might be better to instead change core/SupportedBrowser.php to not throw an exception but instead to set a 403 HTTP header, output the message, and exit within that file.

sgiehl:
@mwithheld to change the response code for this exception we could adjust the `NotSupportedBrowserException` to extend `HttpCodeException`

mwithheld:
@sgiehl  While true and a good idea, I'm proposing that this is not an Exception and so should not be handled like one -- it's normal program flow that should just [set a 403 HTTP header, output the message, and exit] (there's no code for this in core/SupportedBrowser.php).  It's more a design decision if you want to use the exception handling anyway - and that'd be fine IMO -- I just don't feel it's my choice to make.

sgiehl:
Guess using an exception makes it easy to automatically have the error screen rendered. Otherwise this would need to be triggered manually. But I agree that it's useless to log that exception. Personally I would keep the exception and maybe extend our exception with an additional flag that makes it possible to make the exception as "do not log". But I'm also fine with any other solution. Maybe @tsteur or @diosmosis have a stronger opinion on that?

tsteur:
While it's only one or two exceptions and it be good to do it like in the PR for simplicity 👍 . And when it's needed later down the road then we can still go for the more complex solution of having a flag or so.

diosmosis:
Random exits in the code makes tests unpredictable (eg, I just spent 2-3 days debugging a travis only test failure that was happening because our redirect logic just exits, and one test was triggering it but only on travis) and makes it much harder to debug user issues where the exit is triggered especially when we don't have access to the users install and/or they are non-technical while throwing an exception means we can get a stack trace. It would make sense to me to use HttpCodeException and in ExceptionHandler ignore those or log them at a debug level.

sgiehl:
@mwithheld  might it be possible that you pushed the latest changes to the wrong branch? Seems they are not included in this PR

mwithheld:
Those commits are in patch-4 but not coming through in the PR for some reason.  I've instead created PR #17915

mwithheld:
Please see #17915 awaiting review.

sgiehl:
Seems both PRs meanwhile have the same changes. Will close this on again in favor of #17915

PR #17817, opened 2021-07-27
When aggregating reports, inflate one expanded datatable at a time

diosmosis:
### Description:

Currently when aggregating reports over multi periods, we get the blob data for every child period, inflate the blob data for every period into a DataTable tree, then add them together. Afterwards we clean up the extra DataTables, but in the meantime a lot of memory is used, especially when the reports are large. This PR switches to inflating the DataTable tree for just one period at a time and should use less memory overall.

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
Transitions screenshot change is not correct, but it is due to an unrelated bug: https://github.com/matomo-org/matomo/issues/17875

tests/PHPUnit/System/ImportLogsTest.php
```diff
@@ -105,62 +105,6 @@ public function getApiForTesting()
         return $apis;
     }
 
-    /**
```
sgiehl: Any reason for removing those tests?

tests/PHPUnit/System/ImportLogsTest.php
```diff
@@ -78,7 +78,7 @@ public function getApiForTesting()
             // Mysqli rounds latitude/longitude
             $apiMethods = array('Live.getLastVisitsDetails');
         }
-        $apiMethods[] = 'Actions';
+        $apiMethods[] = 'Actions.getPageUrls';
```
sgiehl: shouldn't there be quite  few file removals for the API methods that aren't called anymore, or wasn't this change on purpose?

plugins/Contents/tests/System/expected/test_Contents__Actions.getPageUrls_month.xml
```diff
@@ -6,23 +6,23 @@
 		<nb_hits>2</nb_hits>
 		<sum_time_spent>540</sum_time_spent>
 		<nb_hits_with_time_network>2</nb_hits_with_time_network>
-		<min_time_network>0.0330</min_time_network>
-		<max_time_network>0.0330</max_time_network>
+		<min_time_network>0.033</min_time_network>
```
sgiehl, line 9: Is there a specific reason why the number formatting for those metrics changed?

sgiehl (changes requested):

tests/PHPUnit/System/ImportLogsTest.php
diosmosis: That was for debugging, will undo.

plugins/Contents/tests/System/expected/test_Contents__Actions.getPageUrls_month.xml
diosmosis: When querying, the data is a string '0.0330' in day reports. When aggregating, it is converted to a float, `0.033` and is stored that way in the datatable. This PR does not re-use rows in other tables when the row doesn't exist in the to be aggregated one, it uses `sumRow` on a `new Row()`, so the string value is converted to a float immediately.

diosmosis:
@sgiehl removed the debugging code and fixed some tests

plugins/Contents/tests/System/expected/test_Contents__Actions.getPageUrls_month.xml
sgiehl: that makes sense 👍

sgiehl (approved):
Looks good to me. Tests seem to be passing again 👍 
Not sure if maybe @tsteur wants to have another quick look before merging?

tsteur:
Looks good to me @diosmosis 👍  Great, this will reduce memory quite a bit 💯

PR #17818, opened 2021-07-27
get rid of stashed datatables before going into safe mode to free up memory

diosmosis:
### Description:

In out of memory fatal errors we try to trigger safemode and print out a partial stack trace. However, we'll generally not have the memory to actually do this, so we just trigger another out of memory error that is not handled. To alleviate this, this PR will remove every datatable in memory (as this is usually the cause of out of memory errors). Hopefully this will create a useful trace in the output.

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl (approved):
Guess that's worth a try.

PR #17823, opened 2021-07-27
limit the size of product reports

### LINKED ISSUE #17816: Limiting ecommerce product reports to 10,000 to avoid memory issues with new INI config setting to customise
We’ve migrated from GA to Matomo with end of May and are experiencing issues with the archiving of the “yearly” period for our two largest sites.

PHP has been configured to not apply memory limits. After processing daily, weekly and monthly periods, the archiving process starts with the yearly period and quickly - in a matter of seconds - eats up all the RAM of the processing host. In our case that is currently 15GB. Before we increase the RAM of the host, we’d like to understand the root cause (and if e.g. doubling it would actually help with the problem).

The issue has been discussed already via https://forum.matomo.org/t/memory-usage-issues-with-archiving-high-traffic-ecommerce-site/42334/13 but it looks like this is a dead end for us for now. Our archiving process is still regularly failing unfortunately.

It looks like the tracking of things like SKU data (and product name) per product page impression (see arrow towards the hover popup) seems to be a size issue as we have hundreds of thousands of those in a comparatively short amount of time of those.

https://forum.matomo.org/uploads/db1181/original/2X/5/5cd51893f2e864279473210c5c03ff6cfe2d8a50.jpeg

diosmosis:
### Description:

Applying row limits to product reports as we do to most other reports. This PR does not include the use of ranking query, just two new INI config options to limit the report sizes.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -567,6 +567,12 @@ datatable_archiving_maximum_rows_events = 500
 ; maximum number of rows for sub-tables of the Events tables (eg. for the subtables Categories>Actions or Categories>Names).
 datatable_archiving_maximum_rows_subtable_events = 500
 
+; maximum number of rows for sub-tables of the Products reports
+datatable_archiving_maximum_rows_products = 500
+
+; maximum number of rows for sub-tables of the Products reports
+datatable_archiving_maximum_rows_subtable_products = 500
+
 ; maximum number of rows for other tables (Providers, User settings configurations)
 datatable_archiving_maximum_rows_standard = 500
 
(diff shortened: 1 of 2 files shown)


justinvelluppillai (commented):
This looks pretty straightforward. Do we need to mention these new settings in any documentation (eg here https://matomo.org/faq/how-to/faq_54/)?

diosmosis:
:+1: makes sense to

sgiehl (commented):
Looks good. Do we need to mention that in any changelog? Before the report was actually unlimited and now it's truncated to 500 by default. The change might be relevant for those tracking more than 500 different products...

tsteur:
@sgiehl it'll be mentioned in the regular changelog later as part of the release

sgiehl (approved):

mattab:
Ideally we would set it to a higher value by default, as Ecommerce store owners have an expectation that the product report is accurate, and 500 products out of say 10,000 unique products would be quite inaccurate. For User ID we set a limit of 50,000. Maybe we could set 10,000 for products?

diosmosis:
Updated. Also removed the subtable setting since the products reports don't have subtab.es.

sgiehl (approved):

Issue #17750, opened 2021-07-10
Do we need to create a temporary segment table for unique week|month|year periods?


tsteur:
Seeing queries like these:

![image](https://user-images.githubusercontent.com/273120/125150575-ddf28500-e194-11eb-9bb3-d136a2897241.png)

As I think we mostly only calculate unique visitors for week|month|year|range I wonder if it's actually needed to create a temporary table for these metrics? Might be a lot faster to just get the unique numbers directly?

tsteur:
I didn't test at all, and not even sure it's the right place, and where it be fine to do it but meaning maybe we can do something like this (again, might be wrong place or we might not want it)

```diff
diff --git a/core/ArchiveProcessor.php b/core/ArchiveProcessor.php
index b231d57eb4..11e00c17da 100644
--- a/core/ArchiveProcessor.php
+++ b/core/ArchiveProcessor.php
@@ -527,6 +527,7 @@ class ArchiveProcessor
     protected function computeNbUniques($metrics, $sites)
     {
         $logAggregator = $this->getLogAggregator();
+        $previous = $logAggregator->disallowUsageSegmentCache();
         $sitesBackup = $logAggregator->getSites();
 
         $logAggregator->setSites($sites);
@@ -534,6 +535,9 @@ class ArchiveProcessor
             $query = $logAggregator->queryVisitsByDimension(array(), false, array(), $metrics);
         } finally {
             $logAggregator->setSites($sitesBackup);
+            if ($previous) {
+                $logAggregator->allowUsageSegmentCache();
+            }
         }
         $data = $query->fetch();
         return $data;
diff --git a/core/DataAccess/LogAggregator.php b/core/DataAccess/LogAggregator.php
index e283af2691..6e2d6b0b17 100644
--- a/core/DataAccess/LogAggregator.php
+++ b/core/DataAccess/LogAggregator.php
@@ -265,6 +265,14 @@ class LogAggregator
         $this->allowUsageSegmentCache = true;
     }
 
+
+    public function disallowUsageSegmentCache()
+    {
+        $previous = $this->allowUsageSegmentCache;
+        $this->allowUsageSegmentCache = false;
+        return $previous;
+    }
+

```

diosmosis:
Might merit testing first. From testing hyperloglog, it seemed as if the DISTINCT handling was the cause of slow performance. (The queries there were slow and had no segment. I guess we could compare a bare DISTINCT query vs. a DISTINCT query with a segment.)

[...]

PR #17827, opened 2021-07-28
Faster unique visitors/users query

tsteur:
refs https://github.com/matomo-org/matomo/issues/17750

### DIFF
--- a/core/ArchiveProcessor.php
+++ b/core/ArchiveProcessor.php
@@ -528,11 +528,14 @@ class ArchiveProcessor
     {
         $logAggregator = $this->getLogAggregator();
         $sitesBackup = $logAggregator->getSites();
-
+        $previous = $logAggregator->disallowUsageSegmentCache();
         $logAggregator->setSites($sites);
         try {
             $query = $logAggregator->queryVisitsByDimension(array(), false, array(), $metrics);
         } finally {
+            if ($previous) {
+                $logAggregator->allowUsageSegmentCache();
+            }
             $logAggregator->setSites($sitesBackup);
         }
         $data = $query->fetch();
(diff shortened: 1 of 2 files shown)


tsteur:
Potential patch we could apply to only use this logic when cohorts is not enabled but then the tests etc everything becomes less testable ...

```
diff --git a/core/ArchiveProcessor.php b/core/ArchiveProcessor.php
index 9365152532..510c418f08 100644
--- a/core/ArchiveProcessor.php
+++ b/core/ArchiveProcessor.php
@@ -528,12 +528,20 @@ class ArchiveProcessor
     {
         $logAggregator = $this->getLogAggregator();
         $sitesBackup = $logAggregator->getSites();
-        $previous = $logAggregator->disallowUsageSegmentCache();
+
+        $hasCohorts = \Piwik\Plugin\Manager::getInstance()->isPluginActivated('Cohorts');
+        if (!$hasCohorts) {
+            // when cohorts is enabled, then it's better to create the segment table as cohorts will also issue a
+            // unique query on log data for weeks etc. This way the segment query is only executed once and in most
+            // cases things are faster that way. Vs when cohorts is disabled, then we don't want to first create a
+            // temporary segment table for the segment query and then execute the count.
+            $previous = $logAggregator->disallowUsageSegmentCache();
+        }
         $logAggregator->setSites($sites);
         try {
             $query = $logAggregator->queryVisitsByDimension(array(), false, array(), $metrics);
         } finally {
-            if ($previous) {
+            if (!empty($previous)) {
                 $logAggregator->allowUsageSegmentCache();
             }
             $logAggregator->setSites($sitesBackup);
```

PR #17830, opened 2021-07-28
Don't check cookie value if cookie was deleted

### LINKED ISSUE #17829: Testcookie causes warning in browser console
related to https://github.com/matomo-org/matomo/pull/17744
reported in https://forum.matomo.org/t/javascript-error-trying-to-set-test-cookie-pk-testcookie-domain/42655

Since Matomo 4.4.0, setting the testcookie seems to cause this error to be logged in the browser console:

> There was an error setting cookie `_pk_testcookie_domain`. Please check domain and path.

sgiehl:
### Description:

The issue is reproducible calling  `_paq.push(['setCookieDomain', 'custom.matomo.dev']);`, which tries to set and remove a cookie on the given domain.

fixes #17829

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
is it possible to test this?

Findus23:
Does the test fail if any errors are logged during the execution?

diosmosis:
@Findus23 I'm not sure, but unless an exception is thrown, I doubt it.

js/piwik.js
```diff
@@ -2449,7 +2449,7 @@ if (typeof window.Matomo !== 'object') {
                     ';SameSite=' + sameSite;
 
                 // check the cookie was actually set
-                if (getCookie(cookieName) !== value) {
+                if ((!msToExpire || msToExpire >= 0) && getCookie(cookieName) !== value) {
```
tsteur, line 2452: should it be `msToExpire >= 1` or `>= 5` maybe as the value might be expired by the time it's set and read? I suppose we don't set such values so it doesn't really matter.

sgiehl: Guess no one should use such low values for expire time. If someone does nevertheless, might be good to have a warning, as it simply might not make any sense to set a cookie that expires before it can be used 🙈

sgiehl:
@diosmosis @Findus23 Within the javascript tests `setCookieDomain` is used, and it actually throws that error on console multiple times. But those errors are ignored and might be hard to test. Guess we would need to overwrite `console.error` and `console.log` in order to be able to get and compare the log message. Will check if that's fast to implement...

sgiehl:
Tried writing some tests for a while now, but for some reason the error message is not triggered when I would expect it (like setting an invalid cookie domain). It's only triggered when a cookie is deleted, which should be fixed with the changes here...

diosmosis:
@sgiehl if the test fails in 4.x-dev then it's still useful, right? If it's not possible though then I guess it's not vital.

sgiehl:
maybe we should consider adding some proper tests for the cookie methods
later. For some reasons they didn't trigger a console error when I would
have expected it. But maybe my expectations were wrong. Feel free to have a
look yourself Dizzy if you have some time today...

diosmosis:
Added a test that fails on 4.x-dev but passes on this branch. Will merge now.

PR #7577, opened 2015-03-30
Faster archiving of aggregated reports, also performance imprvovements in general

tsteur:
refs #7373
- Store subtables in chunks of 100 subtables per blob. Those 100 subtables are stored
  serialized as an array: `serialize(array($subtableID => subtableBlob))`. The first 100 subtables are
  stored in "chunk_0_99", the next 100 subtables are stored in "chunk_100_199", .... The `chunk_0_99` chunk contains only 99 subtables as we currently still store the root table separately.
- Subtable Ids are now consecutive from 1 to X
- We do no longer serialize the whole Row instance when archiving, instead we only
  serialize the Row's array which contains columns, metadata and datatable. This is not
  only more efficient but allows us to refactor the Row instance in the future (although
  we will always have to be BC)
- Faster `Row` implementation: Columns, Metadata and Subtables access is much faster now
- Faster `DataTable\Manager` based on `ArrayObject`
### Subtable Chunks
- The root table is still stored under the `recordName` and not within a chunk.
- Subtables are stored in chunks
- 100 subtables are stored per chunk
- They are identified by having the recordName `$recordName_chunk_0_99`, `$recordName_chunk_100_199` (this chunk stores the subtable 100-199).
- Subtables are stored serialized meaning `serialize(array([subtableId] => [blob of subtable]))`

Subtable blobs are merged into chunks in `ArchiveWriter` and split into single blobs in `ArchiveSelector`. The `ArchiveSelector` is also still able to load old subtable blobs (single blob for each subtable)

This results in much less entries in archive blob tables, the index is much smaller, and it will need less stored because of gzip. See https://github.com/piwik/piwik/issues/7373#issue-60019550 and https://github.com/piwik/piwik/issues/7373#issuecomment-84784884

To make this work I had to make subtableIds consecutive which is done in `DataTable::getSerialized`
### Serialize Array instead of Row instance

See https://github.com/piwik/piwik/issues/7373#issuecomment-84784884
### Faster Row implementation

See https://github.com/piwik/piwik/issues/7373#issuecomment-84784884, https://github.com/piwik/piwik/issues/7373#issuecomment-85229486
An even faster solution would be #7570 which we will not use for now as it is not as good debuggable.

Be aware: Once the subtable chunks are in production, there is no way back. We will have to be BC

PR #17831, opened 2021-07-28
make sure summary rows are serialized with their in-db subtable ID and not in memory one

diosmosis:
### Description:

As title, for summary rows we are currently serializing the row with the incorrect subtable ID, causing it not to load w/ the subtable.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/DataTable.php
+++ b/core/DataTable.php
@@ -1345,7 +1345,23 @@ class DataTable implements DataTableInterface, \IteratorAggregate, \ArrayAccess
 
         // we then serialize the rows and store them in the serialized dataTable
         $rows = array();
-        foreach ($this->getRows() as $id => $row) {
+        foreach ($this->rows as $id => $row) {
+            if (isset($consecutiveSubtableIds[$id])) {
+                $backup = $row->subtableId;
+                $row->subtableId = $consecutiveSubtableIds[$id];
+                $rows[$id] = $row->export();
+                $row->subtableId = $backup;
+            } else {
+                $rows[$id] = $row->export();
+            }
+        }
+
+        if (isset($this->summaryRow)) {
+            $id = self::ID_SUMMARY_ROW;
+            $row = $this->summaryRow;
+
+            // duplicating code above so we don't create a new array w/ getRows() above in this function which is
+            // used heavily in matomo.
             if (isset($consecutiveSubtableIds[$id])) {
                 $backup = $row->subtableId;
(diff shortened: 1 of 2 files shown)


tsteur (commented):
@diosmosis does this mean maybe it also has an impact on some other usages of `getRowsWithoutSummaryRow()` and `getSummaryRow()`?

We separated this at some point as `getRows()` was causing memory issues.

Do you know how this later detects the summary row in methods like `addRowsFromSerializedArray` where it relied on having the summary row ID?

I'm not sure I fully understand the issue yet.

diosmosis:
@tsteur from the blame it looks like we were never serializing the summary row. The specific line was last changed seven years ago: https://github.com/matomo-org/matomo/blame/b77e5ff594cdae59862dd4d863220baffc200bec/core/DataTable.php#L1323

And the summaryRow property has been there for nine years. So I think this has always been an issue.

> Do you know how this later detects the summary row in methods like addRowsFromSerializedArray where it relied on having the summary row ID?

We serialize the rows array which associates id w/ row data, so we simply unserialize it. The subtable ID after directly unserializing is used to get the subtable in the database. So it detects it because of the ID we serialize it with.

I'm not sure if this answers your questions...

tsteur (commented):
I understand now 👍  and the ID is afterwards still the same because the summary row has that ID anyway. This way we export also subtables of summary rows.

I do remember in https://github.com/matomo-org/matomo/pull/7577/files#diff-6542f24c55ecdc44d36de949a6c0be12f5a4e9ebcb064ae5a78e3a8cdb6e1b28R1269 years back making changes there for improving performance. 

Just to be safe we're not making things slower would it be too bad to duplicate the code there (as it's executed a lot with a lot of data) like 

```
        if (isset($this->summaryRow)) {
            if (isset($consecutiveSubtableIds[self::ID_SUMMARY_ROW])) {
                $backup = $row->subtableId;
                $row->subtableId = $consecutiveSubtableIds[self::ID_SUMMARY_ROW];
                $rows[self::ID_SUMMARY_ROW] =  $this->summaryRow->export();
                $row->subtableId = $backup;
            } else {
                $rows[self::ID_SUMMARY_ROW] =  $this->summaryRow->export();
            }
        }
```

diosmosis:
@tsteur :+1: that makes sense, to avoid the array operation

diosmosis:
applied review feedback

PR #17832, opened 2021-07-29
Add command to generate a system check

tsteur:
### Description:

Needed for https://github.com/matomo-org/developer-documentation/pull/498

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CoreConsole/Commands/GenerateSystemCheck.php
+++ b/plugins/CoreConsole/Commands/GenerateSystemCheck.php
@@ -48,7 +48,7 @@ class GenerateSystemCheck extends GeneratePluginBase
 
         $this->writeSuccessMessage($output, array(
             sprintf('plugins/%s/Diagnostic/%s.php for %s generated.', $pluginName, $className, $pluginName),
-            sprintf('You should now implement the method called <comment>execute</comment> in %.php', $className),
+            sprintf('You should now implement the method called <comment>execute</comment> in %s.php', $className),
             'You also need to make the diagnostic check known to Matomo in your "plugins/'.$pluginName.'/config/config.php".',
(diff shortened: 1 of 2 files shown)


tsteur:
Because the example plugin is neither in the release nor do we actually add the diagnostic check this should be fine to merge in 4.4.1 maybe. Otherwise can do in Matomo 4.5.0. Added some documentation in https://github.com/matomo-org/developer-documentation/pull/498

plugins/CoreConsole/Commands/GenerateSystemCheck.php
```diff
@@ -0,0 +1,106 @@
...
+                                'ExamplePlugin'  => $pluginName,
+        );
+
+        $whitelistFiles = array('/Diagnostic', '/Diagnostic/ExampleCheck.php');
+
+        $this->copyTemplateToPlugin($exampleFolder, $pluginName, $replace, $whitelistFiles);
+
+        $this->writeSuccessMessage($output, array(
+            sprintf('plugins/%s/Diagnostic/%s.php for %s generated.', $pluginName, $className, $pluginName),
+            'You should now implement the method called <comment>execute</comment> in API.php',
```
sgiehl:
```suggestion
            sprintf('You should now implement the method called <comment>execute</comment> in %.php', $className),
```

sgiehl (changes requested):

plugins/CoreConsole/Commands/GenerateSystemCheck.php
diosmosis: Should that be `%s`?

tsteur:
Thanks @sgiehl  👍  applied the patch

PR #17835, opened 2021-07-29
Fix sorting of Geolocation providers

sgiehl:
### Description:

Geolocation providers can be defined together with an `order`. This flag is used to sort the providers to generate the list of all providers. Currently when two providers have the same order flag, the second provider will replace the first on in the list. Currently this can cause problems when installing a new provider (like IP2Location), as the currently used geoip2php suddenly gets unavailable.

### Review

* [ ] Functional review done
* [ ] Code review done


diosmosis:
Could this benefit from a quick test?

sgiehl:
@diosmosis Added some simple tests. The second one would currently fail on 4.x-dev

PR #17842, opened 2021-07-30
Revert "more private Referrer-Policy: use strict-origin-when-cross-origin "

tsteur:
Reverts matomo-org/matomo#17382

Otherwise it may break the overlay if overlay is loaded cross origin (eg a different domain) as it would only receive the origin referrer but not path and subquery.


justinvelluppillai (approved):
👍

PR #17846, opened 2021-08-02
Remove checking for /proc existence

avkarenow:
### Description:

Currently the '-e' option in `ps` is not used and looks like /proc is no more needed to detect a running process.
The change can make CliMulti usable on some systems like FreeBSD, OpenBSD or macOS.

Tested on FreeBSD without /proc mounted.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/CliMulti/Process.php
+++ b/core/CliMulti/Process.php
@@ -236,10 +236,6 @@ class Process
             $reasons[] = 'awk is not available or did not run as we would expect it to';
         }
 
-        if (!self::isProcFSMounted() && !SettingsServer::isMac()) {
-            $reasons[] = 'procfs is not mounted';
-        }
-
         return $reasons;
     }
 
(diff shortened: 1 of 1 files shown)


diosmosis:
I'm not sure we can just remove this code, since we added it for a reason and can't test every possible system that might use Matomo. Instead we could provide a DI config option to disable the check, and check for more OSes (that are verified not to need the check), plus a faq that we link to. Thoughts @tsteur ?

tsteur:
I just checked it was added in https://github.com/matomo-org/matomo/commit/c076fdb4f27dc9a75191a16132e8b9bdb9cb810f  because of https://github.com/matomo-org/matomo/issues/5041

If I understand it right we used to use `ps -e` and `ps -ex` which required proc to be there. We are now only using `ps x` so I reckon we can likely indeed remove it like @avkarenow described as well.

@avkarenow can you confirm `ps x` works for you and when removing these lines then the archiving still works for you?

avkarenow:
Everything works fine on FreeBSD 12.2. I run it once every day without any problem with archiving on two systems.

tsteur:
I reckon it could be good merging this 👍  If there was any regression afterwards then we may need to change it again but wouldn't expect this to regress

PR #17852, opened 2021-08-03
Fixing some test failures

### LINKED ISSUE #17721: Tests are failing on PHP 7.4 and above
Currently our AllTests suite is silently failing on TravisCI. See https://travis-ci.com/github/matomo-org/matomo/jobs/520002593#L792
The test that silently aborts is this one: 
https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/plugins/CoreUpdater/tests/Integration/Commands/UpdateTest.php#L103-L116

Already saw the same behavior while trying to fix the tests for PHP 8 in #16897

When running the UpdateTests locally on PHP 7.2, 7.4 or 8.0 it works without problems for me.

@matomo-org/core-reviewers is anyone able to reproduce the test failure locally?

Target of this issue should be to
- [ ] Check why the build is actually green even if the test aborts
- [ ] Check why the test fails and fix it

diosmosis:
### Description:

Fixed several tests. The max execution time one is now failing though I'm not sure why. It's failing locally as well.

### Review

* [ ] Functional review done
* [ ] Code review done


sgiehl:
I tried to dig a bit deeper into the failure of the max execution time test. It is actually only failing for `MYSQLI`, with `PDO\MYSQL` it seems to work for all PHP version. Was not yet able to figure out why....

The TwoFactorAuth UI tests are still failing randomly it seems. If you don't want to spend more time trying to fix them right now, guess this PR would be good to merge.

diosmosis:
Thought I was able to reproduce the max_execution_time issue locally, but it's not working for me locally because i'm using mariadb and that syntax isn't supported by mariadb. Will look a bit further.

diosmosis:
I think the mysqli all tests build will need to be split into three parts.

diosmosis:
Setting the max_execution_time manually also did not work for mysqli. I thought maybe the comment was being stripped, but it's still there. No idea what's causing that.

sgiehl:
Looking at the query log in my local mysql server the only difference I can see is that with MYSQLI it always uses prepared statements, while mysql directly sends the queries. Maybe the query hints don't work properly with prepared statements. Will have a quick search and do some testing if I'm able to find out more.

sgiehl:
It seems to be a general issue with prepared statements and mysqli. Tried to test that directly by using the native mysqli methods, which Zend Framework should also do somewhere in the code.

#### MYSQLI Query

```php
$mysqli = new mysqli("localhost", "matomo", "matomo", "matomo");
$mysqli->query("SELECT /*+ MAX_EXECUTION_TIME(1) */ SLEEP(5) FROM matomo_log_visit");
```
This results as expected in 
```
PHP Warning:  mysqli::query(): (HY000/3024): Query execution was interrupted, maximum statement execution time exceeded in /srv/matomo/mysqli.php on line 6
```

#### MYSQLI Prepared Statement

```php
$mysqli = new mysqli("localhost", "matomo", "matomo", "matomo");
$stmt = $mysqli->prepare("SELECT /*+ MAX_EXECUTION_TIME(1) */ SLEEP(5) FROM matomo_log_visit");
$stmt->execute();
```
This runs through without any issues. Fetching the result of the query returns `1`, which is the result of the SLEEP method. So it seems the query hint is ignored. 

#### MYSQL Prepared Statement

Running a prepared statement directly in mysql console like this:
```sql
PREPARE stmt FROM 'SELECT /*+ MAX_EXECUTION_TIME(1) */ SLEEP(5) FROM matomo_log_visit';
EXECUTE stmt;
```
results in `SQL ERROR (3024): Query execution was interrupted, maximum statement execution time exceeded`

So query hints in prepared statements generally should work.

#### Conclusion

Guess query hints is something that can't work with the Zend MYSQLI adapter as it seems to use prepared statements always. Not sure how to go on with this.

ping @tsteur

tsteur:
Good find @sgiehl  couldn't find a bug report for this or any other information (but also only searched for 10min). What do you suggest here? Should we mention in some of the global config settings where you can limit the time that it may not work with mysqli?

diosmosis:
I'll merge this pr as is, the mysqli fix can go into a separate one.

sgiehl:
Created https://github.com/matomo-org/matomo/issues/17871 to keep track on the mysqli query hint issue.

sgiehl:
Will merge this now. Seems there are still some randomly failing UI tests, but guess that's something we can also check in further PRs.

PR #17855, opened 2021-08-03
Use Monolog:SyslogHandler syslog default facility

mwithheld:
### Description:

This is a further fix to #17764, fixing the Monolog:SyslogHandler create() call to use its constructor-default facility rather than manually setting it to 'syslog' incorrectly.


diosmosis:
@mwithheld I think you'll need to merge these changes w/ 4.x-dev to remove the conflicts, something like:

```
git checkout 4.x-dev
git pull upstream 4.x-dev # upstream should be the matomo-org/matomo repo
git checkout patch-2
git merge 4.x-dev
```

plugins/Monolog/config/config.php
```diff
@@ -105,7 +105,8 @@
         ->method('setFormatter', DI\get('log.lineMessageFormatter.file')),
 
     '\Monolog\Handler\SyslogHandler' => DI\create()
```
diosmosis:
```suggestion
    '\Monolog\Handler\SyslogHandler' => DI\autowire()
```

plugins/Monolog/config/config.php
```diff
@@ -105,7 +105,8 @@
...
-        ->constructor(DI\get('log.syslog.ident'), 'syslog', DI\get('log.level.syslog'))
+        ->constructorParameter('ident' => DI\get('log.syslog.ident'))
```
diosmosis:
```suggestion
        ->constructorParameter('ident', DI\get('log.syslog.ident'))
```

plugins/Monolog/config/config.php
```diff
@@ -105,7 +105,8 @@
...
+        ->constructorParameter('level' => DI\get('log.level.syslog'))
```
diosmosis:
```suggestion
        ->constructorParameter('level', DI\get('log.level.syslog'))
```

diosmosis:
@mwithheld left a couple suggestions, after those I should be able to merge it.

diosmosis:
Hi @mwithheld, sorry for the late reply, it seems like you added some commits for another PR into this one (the one dealing w/ NotSupportedBrowserException).

mwithheld:
Good catch - I've reset the branch to before the incorrect ones, force-pushed the branch, and merged updates from the 4.x-dev branch.  Hopefully that is clean now.

diosmosis:
Thanks for fixing up the PR @mwithheld! It's now merged.

PR #17865, opened 2021-08-09
Test Config::getBool() for clarity of configuration options

geekdenz:
### Description:

Clarifies when a configuration option is Boolean and checks whether it is.
Should introduce no functional changes, but since I had already done the effort think it might be worthwhile.

### Review

* [ ] Functional review done
* [ ] Code review done


tests/PHPUnit/Unit/Config/test_files/boolean_settings_test.ini.php
```diff
@@ -0,0 +1,8 @@
+[BoolSettings]
+one = 1
+onestr = '1'
+two = 2
+twostr = '2'
+invalid = 'invalid'
+truebool = true
```
sgiehl, line 13: you could also add a check for the values `ON` and `YES`, which should be converted to `true` as well as `OFF` and `NO`, which should be converted to `false` by the ini component. 
Also there might be edge cases like `1test`, `te1st`, which may be `1` when converted to int.

core/Config.php
```diff
@@ -331,6 +331,26 @@ public function &__get($name)
...
+     * @param string $name variable name
+     * @return bool whether it is considered set true (== 1)
+     *
+     * @api
+     */
+    public function getBool(string $section, string $name): bool
+    {
+        $section = $this->$section;
+        $value = $section[$name] ?? 0;
+        return $value == 1;
```
sgiehl: To be fully sure the value was configured as boolish value we could harden the check to something like:
```suggestion
        return 1 === $value || '1' === $value || true === $value;
```
Otherwise value like `1test` might be accepted, as it converts to `1` when casted to int.

tests/PHPUnit/testfile
```diff
@@ -0,0 +1,34 @@
+/* compile_me_once=first-cache-buster-stylesheet */
```
sgiehl: guess this file was added by accident

tests/testfile
```diff
@@ -0,0 +1,34 @@
+/* compile_me_once=first-cache-buster-stylesheet */
+/* Matomo CSS file is compiled with Less. You may be interested in writing a custom Theme for Matomo! */
```
sgiehl: this file seems unneeded as well.

sgiehl (commented):
Already left a couple of comments. In addition it might be good not to update submodules in PRs where it isn't needed. Especially the icons submodule might cause a lot changes in UI tests, so it's good to update it separately if needed.

core/Config.php
geekdenz: Left the check as is for now as it seems correct also for the cases you mentioned above, @sgiehl . Left if because of BC. I think we can implement stricter tests in the next version. I left a comment in that regard.

tests/PHPUnit/Unit/Config/ConfigTest.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Returns a boolean variable setting for convenience
+     * when calling e.g.
+     * Config::getInstance()->getBool('General', 'force_ssl')
+     * This also documents that a boolean is only true iff
+     * it is equal to 1.
+     * @see https://github.com/matomo-org/matomo/pull/17865
+     * return $value === 1 || $value === 'true' || $value === true; // was suggested for future use by @sgiehl
```
sgiehl:
```suggestion
     * Returns a boolean variable setting for convenience
     * when calling e.g. getBool('General', 'force_ssl')
     * This also documents that a boolean is only true if
     * it is equal to 1.
     * @see https://github.com/matomo-org/matomo/pull/17865
     * return 1 === $value || '1' === $value || true === $value; // was suggested for future use by @sgiehl
```

sgiehl (commented):
@geekdenz So what's left is actually only a test no one might ever look at again. Wondering if it makes more sense to clearly document the behavior and how we expect the config to be used here: https://developer.matomo.org/guides/piwiks-ini-configuration 🤔

tsteur:
@sgiehl we had discussed yesterday in a session to keep the tests to not throw away the tests as it's still good to have to ensure the ini reader converts the expected values to 1 and so we would notice any regression there. Technically the tests would actually need to go to the inireader or so but I suppose that's not too important.

tests/PHPUnit/Unit/Config/ConfigTest.php
```diff
@@ -0,0 +1,93 @@
...
+        $this->assertFalse($this->getBool('BoolSettings', 'twostr'));
+        $this->assertFalse($this->getBool('BoolSettings', 'invalid'));
+        $this->assertFalse($this->getBool('BoolSettings', 'oneinstr'));
+        $this->assertFalse($this->getBool('BoolSettings', 'twoinstr'));
+        $this->assertFalse($this->getBool('BoolSettings', 'isoff'));
+        $this->assertFalse($this->getBool('BoolSettings', 'isno'));
+        $this->assertFalse($this->getBool('BoolSettings', 'truestr'));
+    }
+
+    public function testGetBool_section_Exception()
```
tsteur: fyi @geekdenz because the  getBool is no production method anymore we wouldn't need these edge case tests  and also could simplify the `getBool` method to not throw any exception

geekdenz: removed the exception code and edge case

tests/PHPUnit/Unit/Config/ConfigTest.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        $this->assertFalse($this->getBool('BoolSettings', 'truestr'));
+        $this->assertFalse($this->getBool('BoolSettings', 'two'));
+        $this->assertFalse($this->getBool('BoolSettings', 'twostr'));
+        $this->assertFalse($this->getBool('BoolSettings', 'invalid'));
+        $this->assertFalse($this->getBool('BoolSettings', 'oneinstr'));
+        $this->assertFalse($this->getBool('BoolSettings', 'twoinstr'));
+        $this->assertFalse($this->getBool('BoolSettings', 'isoff'));
+        $this->assertFalse($this->getBool('BoolSettings', 'isno'));
+    }
```
sgiehl: In general I would suggest to use a dataprovider instead of having multiple test cases in one test. e.g. something like this:
```php
    /**
     * @dataProvider getTestCases
     */
    public function testGetBool(expected, $setting)
    {
        $this->assertSame($expected, $this->getBool('BoolSettings', $setting));
    }
    
    public function getTestCases()
    {
        return [
            [true, 'one'],
            [true, 'onestr'],
            [true, 'truebool'],
            [true, 'isyes'],
            [true, 'ison'],
            [false, 'truestr'],
            [false, 'two'],
            [false, 'twostr'],
            [false, 'invalid'],
            [false, 'oneinstr'],
            [false, 'twoinstr'],
            [false, 'isoff'],
            [false, 'isno'],
        ];
    }
```

tests/PHPUnit/Unit/Config/ConfigTest.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace Piwik\Tests\Unit\Config;
+
+use Piwik\Application\Kernel\GlobalSettingsProvider;
+use Piwik\Config;
+use Piwik\Config\ConfigNotFoundException;
```
sgiehl:
```suggestion
```

tests/PHPUnit/Unit/Config/ConfigTest.php
```diff
@@ -0,0 +1,66 @@
...
+use PHPUnit\Framework\TestCase;
+use Piwik\Container\StaticContainer;
```
sgiehl:
```suggestion
```

sgiehl (approved):

PR #17874, opened 2021-08-10
before trying to invalidate parent period, make sure period is enabled in INI config

### LINKED ISSUE #17872: If the period is not enabled in config "The period is not supported" message is shown in cron archiving
We don't use yearly archives because they take too long to process. Because of this, we have the following settings in our config:

[General]
enabled_periods_UI = "day,week,month,range"
enabled_periods_API = "day,week,month,range"

After upgrading from 3.14.0 to 4.4.1, I'm unable to run the cron archive. The output is below.

console core:archive --url=... --force-periods=day --disable-scheduled-tasks

INFO [2021-08-10 18:58:48] 85257  ---------------------------
INFO [2021-08-10 18:58:48] 85257  INIT
INFO [2021-08-10 18:58:48] 85257  Running Matomo 4.4.1 as Super User
INFO [2021-08-10 18:58:48] 85257  ---------------------------
INFO [2021-08-10 18:58:48] 85257  NOTES
INFO [2021-08-10 18:58:48] 85257  - Async process archiving supported, using CliMulti.
INFO [2021-08-10 18:58:48] 85257  - Reports for today will be processed at most every 10 seconds. You can change this value in Matomo UI > Settings > General Settings.
INFO [2021-08-10 18:58:48] 85257  - Archiving was last executed without error 7s ago
INFO [2021-08-10 18:58:48] 85257  - Will only process the following periods: day (--force-periods)
INFO [2021-08-10 18:58:48] 85257  ---------------------------
INFO [2021-08-10 18:58:48] 85257  START
INFO [2021-08-10 18:58:48] 85257  Starting Matomo reports archiving...
INFO [2021-08-10 18:58:48] 85257  Start processing archives for site 1.
INFO [2021-08-10 18:58:48] 85257    Will invalidate archived reports for today in site ID = 1's timezone (2021-08-10 00:00:00).
ERROR [2021-08-10 18:58:48] 85257  Uncaught exception: /home/aatrkco/public_html/stats/core/Period/Factory.php(131): The period 'year' is not supported. Try any of the following instead: day, week, month, range [Query: , CLI mode: 1]

  [Exception]
  The period 'year' is not supported. Try any of the following instead: day, week, month, range

diosmosis:
### Description:

Fixes #17872 

This has the consequence where if a week period is disabled, but the month period is enabled, the month will not archive. Which I think is expected since the month period depends on the week.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -211,6 +211,11 @@ enable_processing_unique_visitors_multiple_sites = 0
 ; Example use case: custom date range requests are processed in real time,
 ; so they may take a few minutes on very high traffic website: you may remove "range" below to disable this period
 enabled_periods_UI = "day,week,month,year,range"
+
+; The list of periods that are available in through the API. This also controls the list of periods that are allowed
+; to be archived. You can disable some of them if you have a high traffic website and archiving is too compute heavy.
+; NOTE: if you disable a period in the API, it's parent periods are effectively disabled as well. For example, if
+; month periods are disabled, then year periods can no longer be computed, so they are effectively disabled as well.
 enabled_periods_API = "day,week,month,year,range"
 
 ; whether to enable segment archiving cache
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -392,6 +392,7 @@ class ArchiveInvalidator
     {
         if ($period->getLabel() == 'year'
             || $period->getLabel() == 'range'
+            || !Period\Factory::isPeriodEnabledForAPI($period->getParentPeriodLabel())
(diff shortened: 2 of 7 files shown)


sgiehl (approved):

sgiehl:
> This has the consequence where if a week period is disabled, but the month period is enabled, the month will not archive. Which I think is expected since the month period depends on the week.

Not sure if that's expected by the user. In theory we could also calculate the month archives with the daily ones. Should we maybe add a hint in the config, that disabling a period might break the archiving for bigger periods?

sgiehl (approved):

PR #17877, opened 2021-08-11
Only set requests in tracking request set when correct value is given

### LINKED ISSUE #153: Second queue is growing forever
Hello, since few Matomo versions we have returning problem with the second queue which is growing forever (until I remove this plugin and reinstall it back, as I haven't found better method yet).

Matomo version: 4.4.1
MySQL version: 5.7.34-37-log
PHP version: 7.3.29-1+0~20210701.86+debian10~1.gbp7ad6eb
Backend: Redis

sudo -u www-data /usr/share/piwik/console queuedtracking:monitor
Queue is enabled
The command ./console queuedtracking:process has to be executed to process request sets within queue
Up to 8 workers will be used
Processor will start once there are at least 5 request sets in the queue
218717 (53+218338+68+60+54+60+40+44) request sets left in queue. 464.62M used memory (588.55M peak). 1 workers act...

How can I debug this?
Thanks :-)

tsteur:
refs https://github.com/matomo-org/plugin-QueuedTracking/issues/153#issuecomment-896856651

fix https://github.com/matomo-org/plugin-QueuedTracking/issues/153

Possible warning:

>  /usr/share/piwik/core/Tracker/RequestSet.php(36): Warning - Invalid argument supplied for foreach() -

Not sure how this would happen and it might hide an actual error maybe. I've been looking through code for a while and couldn't really see where this might happen.  The only one I could find was bulk tracking a wrong JSON being set. Maybe we should throw an invalid tracking parameter exception when `requests` is not an array? https://github.com/matomo-org/matomo/blob/4.4.1/plugins/BulkTracking/Tracker/Requests.php#L78

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/core/Tracker/RequestSet.php
+++ b/core/Tracker/RequestSet.php
@@ -33,6 +33,10 @@ class RequestSet
     {
         $this->requests = array();
 
+        if (empty($requests)|| !is_array($requests)) {
+            return;
+        }
+        
         foreach ($requests as $request) {
             if (empty($request) && !is_array($request)) {
                 continue;


tsteur:
User confirmed it works so far.

sgiehl (approved):

Issue #16552, opened 2020-10-09
Make requests faster by faster IO operations (eg file_exists is slower compared to is_file)


tsteur:
See https://bugs.php.net/bug.php?id=78285 

From https://thephp.cc/news/2020/10/caching-makes-everything-faster-right 
![image](https://user-images.githubusercontent.com/273120/95528863-ca639380-0a35-11eb-9296-3468e2fe44bc.png)

We’re probably having several hundred file_exists calls per request so replacing this calls with `is_file` could speed things up. Especially where the disk is slow or where there's heaps of traffic.

sgiehl:
This should nowadays not have any meaningful difference anymore.

PR #17879, opened 2021-08-12
Added config option for custom tracking image

### LINKED ISSUE #2672: Custom image to replace default 1x1 GIF image
Hello everyone,

I would like to change the default GIF image returned by piwik but I can't figure out where to change this.

I'am using the Image tracker code: (&lt;img src="http://127.0.0.1/piwik/piwik.php?idsite=1&rec=1" style="border:0" alt="" />

I would like to add a parameter to specify witch image I want or simply return an image that I specify the path somewhere.

A parameter can be a good solution to choose from multiple images.

Thank you.

bx80:
### Description:

fixes #2672 Custom image to replace default 1x1 GIF image

Added a new config option which may be used to specify a custom image to override the default 1x1 transparent gif. 
Either a image file location (with full path) or a base64 image string may be used.

`
[Tracker]
custom_image = /path/to/my_custom_image.png
`
or
`
[Tracker]
custom_image = "iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAIAAAACDbGyAAAACXBIWXMAAC4jAAAuIwF4pT92AAAAB3RJTUUH5QgLFiABlwQnpwAAABl0RVh0Q29tbWVudABDcmVhdGVkIHdpdGggR0lNUFeBDhcAAAAUSURBVAjXY/wjLMyABJgYUAGpfABbJQEsALGyNgAAAABJRU5ErkJggg=="
`

Two unit tests are included which set the config option for both an image file and base64 image string, then check the respective tracker responses against the original image data.

### Review

* [x] Functional review done
* [x] Code review done

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -996,6 +996,11 @@ innodb_lock_wait_timeout = 0
 ; For a list of tracking parameters you can use on the left side view https://developer.matomo.org/api-reference/tracking-api
 exclude_requests = ""
 
+; Custom image to return when tracker URL includes &image=1
+; Overrides the default 1x1 transparent gif
+; This should either be the full path to the image file or a base64 encoded image string wrapped in quotes
+custom_image =
+
 [Segments]
 ; Reports with segmentation in API requests are processed in real time.
 ; On high traffic websites it is recommended to pre-process the data
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -999,6 +999,7 @@ exclude_requests = ""
 ; Custom image to return when tracker URL includes &image=1
 ; Overrides the default 1x1 transparent gif
 ; This should either be the full path to the image file or a base64 encoded image string wrapped in quotes
+; For both image files and base64 encoded strings supported image types are gif, jpg and png
 custom_image =
 
 [Segments]
--- a/core/Tracker/Response.php
+++ b/core/Tracker/Response.php
@@ -198,26 +198,8 @@ class Response
 
         // Must have valid image data and a valid mime type to proceed
         if ($img && $size && isset($size['mime'])  && in_array($size['mime'], $supportedMimeTypes)) {
-
-            // Recreate the image to improve security
-            $newImg = imagecreatefromstring($img);
-            if ($newImg === false) {
-                return false;
-            }
-
             Common::sendHeader('Content-Type: '.$size['mime']);
-            switch ($size['mime']) {
-                case 'image/png':
-                    imagepng($newImg);
-                    break;
-                case 'image/jpeg':
-                    imagejpeg($newImg);
-                    break;
-                case 'image/gif':
-                    imagegif($newImg);
-                    break;
-            }
-
+            echo $img;
             return true;
         }
 
--- a/core/Tracker/Response.php
+++ b/core/Tracker/Response.php
@@ -176,31 +176,49 @@ class Response
      * Output a custom tracking image
      *
      * @param string $customImage The custom image setting specified in the config
+     *
      * @return bool True if the custom image was successfully output, else false
      */
     private function outputCustomImage(string $customImage): bool
     {
+        $supportedMimeTypes = ['image/png', 'image/gif', 'image/jpeg'];
+
+        $img = null;
+        $size = null;
 
         if (strlen($customImage) > 2 && substr($customImage, -2) == '==') {
-
-            $transGifBase64 = $customImage;
-            Common::sendHeader('Content-Type: image/gif');
-
-            echo base64_decode($transGifBase64);
-            return true;
+            // Base64 image string
+            $img = base64_decode($customImage);
+            $size = getimagesizefromstring($img);
+        } else if (is_file($customImage) && is_readable($customImage)) {
+            // Image file
+            $img = file_get_contents($customImage);
+            $size = getimagesize($customImage); // imagesize is used to get the mime type
         }
 
-        if (file_exists($customImage)) {
+        // Must have valid image data and a valid mime type to proceed
+        if ($img && $size && isset($size['mime'])  && in_array($size['mime'], $supportedMimeTypes)) {
 
-            $fp = fopen($customImage, 'rb');
-            $size = getimagesize($customImage); // imagesize is used to get the mime type
-            if ($fp && $size && isset($size['mime']))
-            {
-                Common::sendHeader('Content-Type: '.$size['mime']);
-                fpassthru($fp);
-                return true;
+            // Recreate the image to improve security
+            $newImg = imagecreatefromstring($img);
+            if ($newImg === false) {
+                return false;
             }
 
+            Common::sendHeader('Content-Type: '.$size['mime']);
+            switch ($size['mime']) {
+                case 'image/png':
+                    imagepng($newImg);
+                    break;
+                case 'image/jpeg':
+                    imagejpeg($newImg);
+                    break;
+                case 'image/gif':
+                    imagegif($newImg);
+                    break;
+            }
+
+            return true;
         }
 
         return false;
--- a/core/Tracker/Response.php
+++ b/core/Tracker/Response.php
@@ -153,9 +153,8 @@ class Response
 
         // Check for a custom tracking image
         $customImage = Config::getInstance()->Tracker['custom_image'];
-        if (!empty($customImage)) {
-            if ($this->outputCustomImage($customImage))
-                return;
+        if (!empty($customImage) && $this->outputCustomImage($customImage)) {
+            return;
         }
 
         // No custom image defined, so output the default 1x1 base64 transparent gif
--- a/core/Tracker/Response.php
+++ b/core/Tracker/Response.php
@@ -10,6 +10,7 @@ namespace Piwik\Tracker;
 
 use Exception;
 use Piwik\Common;
+use Piwik\Config;
 use Piwik\Profiler;
 use Piwik\Timer;
 use Piwik\Tracker;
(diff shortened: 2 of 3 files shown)


core/Tracker/Response.php
```diff
@@ -150,9 +151,20 @@ private function outputApiResponse(Tracker $tracker)
             return;
         }
 
+        // Check for a custom tracking image
+        $customImage = Config::getInstance()->Tracker['custom_image'];
+        if (!empty($customImage)) {
+            if ($this->outputCustomImage($customImage))
+                return;
```
sgiehl:
```suggestion
        if (!empty($customImage) && $this->outputCustomImage($customImage)) {
            return;
```

core/Tracker/Response.php
```diff
@@ -161,6 +173,40 @@ private function outputTransparentGif()
...
+    /**
+     * Output a custom tracking image
+     *
+     * @param string $customImage The custom image setting specified in the config
+     * @return bool True if the custom image was successfully output, else false
+     */
+    private function outputCustomImage(string $customImage): bool
+    {
+
+        if (strlen($customImage) > 2 && substr($customImage, strlen($customImage)-2, 2) == '==') {
```
sgiehl: This should do the same:

```suggestion
        if (strlen($customImage) > 2 && substr($customImage, -2) == '==') {
```

tests/PHPUnit/Unit/Tracker/ResponseTest.php
```diff
@@ -167,6 +168,55 @@ public function test_outputResponse_shouldOutputApiResponse_IfTrackerIsDisabled(
...
 
+    public function test_outputResponse_shouldOuputCustomImage_IfCustomBase64ImageSet()
+    {
+        // Base64 sample image string (4x4px red PNG made in GIMP)
+        $base64Image = 'iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAIAAAACDbGyAAAACXBIWXMAAC4jAAAuIwF4pT92AAAAB3RJTUUH5QgLFiABlwQnpwAAABl0RVh0Q29tbWVudABDcmVhdGVkIHdpdGggR0lNUFeBDhcAAAAUSURBVAjXY/wjLMyABJgYUAGpfABbJQEsALGyNgAAAABJRU5ErkJggg==';
+        
+        // Initialise the custom_image setting
+        $config = Config::getInstance();
+        $trackerSettings = $config->Tracker;
+        //$trackerSettings['custom_image'] = '/var/www/matomo/plugins/Morpheus/images/logo.png';
```
sgiehl: guess that line can be removed.
```suggestion
```

sgiehl (changes requested):
Hi @bx80,
thanks for your contribution!
I had a quick look at the code and left a couple of suggestions. Once PHPCS is fixed and tests are running through I will quickly test the functionality locally so we can merge that soon :-)

bx80:
Thanks for the suggestions, much appreciated.

config/global.ini.php
```diff
@@ -996,6 +996,11 @@
 ; For a list of tracking parameters you can use on the left side view https://developer.matomo.org/api-reference/tracking-api
 exclude_requests = ""
 
+; Custom image to return when tracker URL includes &image=1
+; Overrides the default 1x1 transparent gif
+; This should either be the full path to the image file or a base64 encoded image string wrapped in quotes
```
tsteur, line 1001: do I see this right that it has to be a gif when it's a string? Might need to mention this?

core/Tracker/Response.php
```diff
@@ -161,6 +172,40 @@ private function outputTransparentGif()
...
+        if (strlen($customImage) > 2 && substr($customImage, -2) == '==') {
+
+            $transGifBase64 = $customImage;
+            Common::sendHeader('Content-Type: image/gif');
+
+            echo base64_decode($transGifBase64);
+            return true;
+        }
+
+        if (file_exists($customImage)) {
```
tsteur: not too important but might want to use `is_file` as it's bit faster see https://github.com/matomo-org/matomo/issues/16552 . It usually wouldn't make a big difference unless someone tracks heaps of traffic. Technically would also need to check `is_readable`.

core/Tracker/Response.php
```diff
@@ -161,6 +172,40 @@ private function outputTransparentGif()
...
+
+            $fp = fopen($customImage, 'rb');
+            $size = getimagesize($customImage); // imagesize is used to get the mime type
+            if ($fp && $size && isset($size['mime']))
+            {
+                Common::sendHeader('Content-Type: '.$size['mime']);
```
tsteur: Was wondering if we maybe need to escape the mime type from a security perspective but probably not needed I think and could potentially maybe break behaviour.

core/Tracker/Response.php
```diff
@@ -161,6 +172,40 @@ private function outputTransparentGif()
...
+            Common::sendHeader('Content-Type: image/gif');
+
+            echo base64_decode($transGifBase64);
+            return true;
+        }
+
+        if (file_exists($customImage)) {
+
+            $fp = fopen($customImage, 'rb');
+            $size = getimagesize($customImage); // imagesize is used to get the mime type
```
tsteur, line 196: see https://www.php.net/manual/en/function.getimagesize.php

> This function expects filename to be a valid image file. If a non-image file is supplied, it may be incorrectly detected as an image and the function will return successfully, but the array may contain nonsensical values.
> Do not use getimagesize() to check that a given file is a valid image. Use a purpose-built solution such as the Fileinfo extension instead.

My security concern is that someone could configure to print a file that is made to look like an image when it's maybe not. A sys admin would likely not configure anything malicious there but maybe if a different security bug allowed to write certain config files then an attacker might try this way to get various different files printed to get information. 

Not sure what the best way is to prevent any potential security issue there. The content of that file would be printed to anyone so it could have quite some impact.

tsteur: If `allow_url_fopen` then it is possible even that the file points to internal intranet services etc which could result in Server-Side Request Forgery .

Also I believe `fopen` might be disabled on some installations and would maybe need to fallback to something else. So far we're actually always using `file_get_contents` (which would also fetch data from URLs)

maybe as a starter we should check that it's a local file and not a remote URL like `phar://`, `http://`. 

And we maybe need to recreate the image every time for improved security like using  https://www.php.net/manual/en/function.imagecreatefromgif.php etc

tsteur (commented):
Nice work, left a few comments to clarify

config/global.ini.php
bx80: Yes that's true, that should at least be covered in the config notes. I don't want to determine the mime type using getimagesize() as it will require writing the string to a file first which will incur disk i/o making the more performant string option pointless. Other options would be to add an additional config option to set the mime type when a string is used (increases the possibility of mis-configuration) or add a method that attempts to determine the mime type by looking for gif/png/jpg signature bytes at the start of the string.

What do you think?

core/Tracker/Response.php
bx80: Taking a quick look at the getimagesize php c source, it seems to be matching known byte headers and then returning a predefined mime type, so it doesn't look like there would be any way to usea carefully crafted image file to inject strings into the getimagesize return array. For extra safety we could check that the $size['mime'] string always matches a limited list of mime type we want to support, eg. image/gif, jpg or png.

config/global.ini.php
bx80: getimagesizefromstring() looks like a good option for automatically determining the mime type from the string without incurring disk overhead.

core/Tracker/Response.php
bx80: I'll switch to using file_get_contents and add validation to make sure the path only starts with a slash or a dot and not a protocol.

If we're working off a limited supported set of mime type then recreating the image with imagecreatefromgif/jpeg/png should be fairly straightforward. I can do some performance tests to see how much it slows things down.

bx80: I did a test using imagecreatefronstring() and imagepng() to recreate the image. Functionally it works fine, but performance is poor:

```
10,000x
image file 11kib PNG, regenerate   : 73.20s
 ... imagecreatefromstring()          8.09s
 ... imagepng()                      65.09s

image file 11kib PNG, pass through :  0.04s
image string 200b PNG, regenerate  :  0.17s
image string 200b PNG, pass through:  0.02s
```
All file loading is done outside the loop and there is no output being sent, so the time is all taken by the two GD functions.

11kib isn't an especially big image, so it seems that recreating the image on every response would create an unacceptable performance hit.

config/global.ini.php
bx80: Change committed to automatically determine the mime type for image strings

core/Tracker/Response.php
bx80: Changed to use is_file and is_readable. is_file is also filtering out remote files for file_get_contents

core/Tracker/Response.php
bx80: Mime types are now limited to restricted list

sgiehl:
@bx80 it seems the tests you have added are failing. See https://app.travis-ci.com/github/matomo-org/matomo/jobs/531164790#L735

core/Tracker/Response.php
bx80: I've reverted image recreation as I think it will be too slow.

bx80:
The tests were failing because of the image recreate commit I'd previous added, a recreated image wouldn't match the original source image. I've reverted image recreation change due to performance and so the tests should now pass.

core/Tracker/Response.php
sgiehl: @tsteur do you have any further concerns regarding the high traffic instances? I was wondering if it could make sense to somehow cache the mime type for some time, so the `getimagesize` wouldn't be called in each request.

tsteur: @sgiehl I would say we ignore this for now and look into it if/when it becomes a problem. eg 70s/10000 = is only 7ms per request for a 11KB image. It's not great but also not too bad maybe just to keep things simple for now.

sgiehl (approved):
From my point of view this should be ready to merge. Guess it's kind of expected that configuring a broken inline image (like `c4==`) also serves this broken image.
@tsteur shall we merge this for 4.5?

tsteur:
@sgiehl could you create an FAQ on matomo.org about how to change the tracking image before merging it? If we send a broken image when misconfigured then we should mention this also in the FAQ. We would need to mention how they can test the image.

sgiehl:
@tsteur created https://matomo.org/faq/how-to-change-the-tracking-image/

justinvelluppillai:
Thanks @bx80!

PR #17880, opened 2021-08-12
Use newer version of Puppeteer for UI tests

sgiehl:
### Description:

We are still using Puppeteer 2.1.1, which uses a quite "old" Chromium version.
Puppeteer 8.0.0 will update that to Chromium 90.

Note: I'm note yet updating to version 10.2.0, as with version 10 some methods that we are using were removed, also for some reason the tests are hanging after a fixture is set up. Don't want to waste time trying to figure out what exactly needs to be changed.

### Review

* [ ] Functional review done
* [ ] Code review done


mattab:
Is there any benefit to this project, besides maybe preventing some tech debt? Maybe this can wait for some months, or is there value in continuing this work now?

sgiehl:
This ensures our tests are running with a more modern Chrome. In theory the newer version would even support running the tests with Firefox as well. Also there are a lot more features, that can be quite useful for writing better tests. In addition I hope this brings some more stability into our tests and reduces the random failures.

Findus23:
Especially this makes sure that the UI tests (and other JS tests) represent the reality of people using modern versions of web browsers instead of Chrom(e|ium) 80 which was released in 2019 and might not trigger bugs in Matomo caused by changes in how browsers work since then.

sgiehl:
Seems I finally found a solution to "fix" the TwoFactorAuth UI tests, even though the solution is not optimal.
It seems due to the many calls for login + logout and the settings at some point the CoreJS javascript is served empty. I actually was not able to figure out where the exact problem is. Maybe it's a raise condition that two request want to create the asset file or something similar.
Disabling the merged assets for those tests seems to solve the test failures.

sgiehl:
I'll add the `needs review` label, but will still keep the draft status, as there are various updates in the submodules. So if this should be merged, we need to create PRs for the submodules first, merge them and update the submodules in this PR before merging it.

core/AssetManager/UIAsset/OnDiskUIAsset.php
```diff
@@ -125,7 +125,7 @@ public function exists()
      */
     private function assetIsReadable()
     {
-        return is_readable($this->getAbsoluteLocation());
+        return is_readable($this->getAbsoluteLocation()) && filesize($this->getAbsoluteLocation()) > 10;
```
tsteur: is there any chance we can remove this one or only use it in tests? Mentioning this regarding performance mostly. Also a short script like `$ = jQuery` could be skipped but not sure it would be out there. We have such a script in Matomo for WordPress for example but we're adding it using a string and it's not in a file.

sgiehl: oh. That actually was a test to check if that fixes the issue on travis, which it didn't. So I can simply revert it.

Issue #15322, opened 2019-12-29
Segments do not work with encoded URL's as advised. 


jmumby:
When creating a segment using encoding as advised in the help text, no results are shown. 

![image](https://user-images.githubusercontent.com/4339071/71562691-0b169f80-2ae9-11ea-96cf-17720a7751d1.png)

For example `https%3A%2F%2Fexample.com` yields no results in the visits log where as `https://example.com` does.

mattab:
Interesting, the inline text was designed for this page: 
https://developer.matomo.org/api-reference/reporting-api-segmentation
but indeed this same advice is wrong when within the UI, where the URLs should not be encoded.

PR #17882, opened 2021-08-13
hide segmented reports are pre-processed

### LINKED ISSUE #15036: Hide "AND segmented reports are pre-processed (faster, requires cron)" from the UI when enable_create_realtime_segments = 0
Currently the UI looks like this:
Screenshot from 2019-10-22 17-32-10

The last selector cannot be changed when `enable_create_realtime_segments = 0`

Because it provide no value in itself, it would be great to hide the text all together: "AND segmented reports are pre-processed (faster, requires cron)" from the UI when `enable_create_realtime_segments = 0`

geekdenz:
when enable_create_realtime_segments = 0
fixes #15036

### Description:

### Review

* [ ] Functional review done
* [ ] Code review done


plugins/SegmentEditor/SegmentEditor.php
```diff
@@ -316,6 +316,16 @@ public static function isAddingSegmentsForAllWebsitesEnabled()
...
 
+    /**
+     * Returns whether create realtime segments is enabled or not.
+     *
+     * @return bool
+     */
+    public static function isCreateRealtimeSegmentsEnabled(): bool
+    {
+        return Config::getInstance()->General['enable_create_realtime_segments'] == 1;
+    }
```
sgiehl, line 327: You could change this to use this new method as well:
https://github.com/matomo-org/matomo/blob/da44e3b3416f9012c774aa7abef19cb8cdbf58e1/plugins/SegmentEditor/API.php#L120-L121

geekdenz: Thanks @sgiehl .

Sorry but I don't know what you mean. The lines you sent are not a method. Do you mean not checking for `== 1`? I thought for BC that is what we wanted.

tsteur: @geekdenz what @sgiehl is meant replacing that logic to use the new method like 

```
$realTimeSegmentsEnabled = SegmentEditor:: isCreateRealtimeSegmentsEnabled();
```

Not the other way around.

sgiehl (approved):
Code looks fine and works as expected. We could maybe add another UI test that checks the UI when the option is disabled to ensure it doesn't regress in the future

justinvelluppillai:
@geekdenz This PR is linked to #15036 and #15322, is that correct?

sgiehl (approved):

PR #17884, opened 2021-08-13
Don't try to send security email when no email is provided

tsteur:
### Description:

Which happens mostly when eg a user is added on CLI.

We noticed this on cloud that it's causing issues and it's trying to add a user, tokens etc. and tries to send emails. The `safeSend` method is catching exceptions however it is logging data as warning. In CLI this causes the return value of an executed CLI command to not be `0` as our console treats any warning as an error.

I checked and generally on CLI the email should be an empty string.

There are also other approaches to fixing this. 

* We could not send the email  on `cli` but generally if there is a user it be better to send it.
* We could develop the check for each individual email but seems like a lot of duplicate code for everything

```
+        $email = Piwik::getCurrentUserEmail();
+        if (!empty($email)) {
+            $mail = $container->make(UserCreatedEmail::class, array(
+                'login' => Piwik::getCurrentUserLogin(),
+                'emailAddress' => Piwik::getCurrentUserEmail(),
+                'userLogin' => $userLogin
+            ));
+            $mail->safeSend();
+        }
```

* We could potentially listen to the "shouldSendMail" event and disable it there in the cloud plugin but this will later again cause issues in core.
* etc.

Generally we wouldn't want to send any such mail while creating a Matomo instance.

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/CoreAdminHome/Emails/SecurityNotificationEmail.php
+++ b/plugins/CoreAdminHome/Emails/SecurityNotificationEmail.php
@@ -81,5 +81,13 @@ abstract class SecurityNotificationEmail extends Mail
         return $view;
     }
 
+    public function send()
+    {
+        if (!empty($this->emailAddress)) {
+            return parent::send();
+        }
+        return false;
+    }
+
     abstract protected function getBody();
 }


sgiehl (commented):
Guess it should work this way. But wondering if we maybe shouldn't check for missing or invalid recipients in the transport class and maybe throw a more specific exception in that case, so we are able to catch and maybe ignore it on CLI or so

tsteur:
@sgiehl there is already an exception which gets caught in `safeSend()` which is used with the security notifications in most places. The problem is it's logging a warning in CLI which then makes the console command not return a success code. I could simply not log the warning on CLI but figured that might not be the best as sometimes there might be real issues where we'd want it logged even if email is invalid or so. I could do something like below (not tested though). 

It really depends though because some emails sent on CLI they actually represent an error if no email is provided so I wouldn't really want to do it that way.


```
diff --git a/core/Mail.php b/core/Mail.php
index 1986b5501e..e8851ac0d1 100644
--- a/core/Mail.php
+++ b/core/Mail.php
@@ -304,8 +304,11 @@ class Mail
         try {
             $this->send();
         } catch (\Exception $e) {
-            // we do nothing but log if the email send was unsuccessful
-            StaticContainer::get(LoggerInterface::class)->warning('Could not send {class} email: {exception}', ['class' => get_class($this), 'exception' => $e]);
+            $recipients  = array_filter($this->recipients);
+            if (!Common::isPhpCliMode() || !empty($recipients)) {
+                // we do nothing but log if the email send was unsuccessful
+                StaticContainer::get(LoggerInterface::class)->warning('Could not send {class} email: {exception}', ['class' => get_class($this), 'exception' => $e]);
+            }
         }
     }

```

sgiehl:
Sure, but it's a generic exception that is hard to differ from any other error that might occur. Throwing a custom exception in our code when no recipient is set, would allow us to handle that different.

tsteur:
@sgiehl I looked into this and started coding it. Because we don't want to have this behaviour for all emails it makes things only more complicated and results in bit of duplicated code and would need to change everywhere the call for security notification emails from `safeSend()` to `send()` so I could catch that exception and then handle things again. In the end the outcome is the very same. Or it would require a new method in mail to not fail when there is no recipient etc.

In many classes we do want to keep same existing behaviour eg when sending a custom alert and there is no recipient then it may be a bug and we wouldn't want to ignore this. 

FYI this is the exception that PHPMailer triggers 
image

sgiehl:
@tsteur What I don't like about this solution is that it's a special handling for certain mails and later someone might wonder why security mails simply don't fail to send when no recipient is set, but all other mails are failing. Or even worse someone might introduce a "bug" where no recipient is set accidentally and we might never notice as no error is triggered.
Personally I would prefer a more global solution, that can be used for all mails if needed. What about having a DI config like `ignore_mail_errors_for_empty_recipient` (or an event) and changing it for cloud in that special case were it's needed. If we later might need that somewhere else we can simply reuse it.

tsteur:
@sgiehl totally get you prefer a global solution. I can add some more complex code to achieve the same for the security notification mails only as a global solution simply won't work as much there. This is not cloud specific but overall also relevant for On-Premise especially once we add headless install. I will just fix this issue in cloud then and will close the PR.

PR #17887, opened 2021-08-14
JavaScript tracker exclude query parameters

### LINKED ISSUE #11627: Piwik JavaScript Tracker: Add possibility to set URL parameters to ignore
For a Piwik website or globally, it is already possible to exclude / ignore URL parameters. However, because the default request method is `GET`, ignored URL parameters would be still visible in the access log of the Piwik server.

Instead it would be great to be able to have a method in the JS tracker to ignore URL parameters like `_paq.push(['setIgnoreUrlParams', 'sid', 'cid']);`. Not sure what a good method name would be.

Because Piwik knows server side which parameters to ignore, we could also re-use this information in the JavaScript tracking code generator and automatically set this parameter.

In the tracker itself we already have a method `removeUrlParameter()` to remove a URL parameter so it should be not too hard to develop.  We would add this probably to the method `purify(url)` where we already modify the URL based on configuration.

bx80:
### Description:

Fixes #11627

Adds a new option to the JavaScript tracker to exclude query parameters from the tracked URL.

```js
_paq.push(["setExcludedQueryParams", ["uid", "sid"]]);
```
Existing purify() and removeUrlParameter() functions were reused.

The tracker code generator has been updated to automatically add this option to the generated javascript tracking code using the excluded query parameters configured under Websites > Settings.

Tracker code generator unit tests have also been updated to test for valid tracker code and string escaping with the option enabled.

(Initially used "setIgnoreUrlParams" as the option name as per issue report, but though it would be more consistent to use "setExcludedQueryParams" to match the existing config option and UI terminology.)

### Review

* [ ] Functional review done
* [ ] Code review done
* [X] Tests were added if useful/possible


tsteur:
Thanks for this @bx80  👍  It's not been reviewed yet but noticed that this PR may include changes from your other PR?

bx80:
Sorry about that, I unintentionally created a sub-branch of the other PR. I've now rebased the branch so it only has the relevant commits.

js/piwik.js
```diff
@@ -2468,6 +2471,12 @@ if (typeof window.Matomo !== 'object') {
                 // for outlinks or referrers
                 url = removeUrlParameter(url, configVisitorIdUrlParameter);
 
+                if (configExcludedQueryParams.length > 0) {
+                    for (var i = 0; i < configExcludedQueryParams.length; i++) {
```
sgiehl: Our JSLint complains about this, as variable declarations should be at the beginning. So simply remove the `var` here and add `i` at the beginning so its `var targetPattern, i;`

sgiehl (changes requested):
It would be awesome to also have some javascript tests for that. They could be added after this:
https://github.com/matomo-org/matomo/blob/da4df21e808fa983e32c6402c8f3de90145e0910/tests/javascript/index.php#L2510
Not sure how familiar you are with those tests, but it could be something like:
```
        tracker.setExcludedQueryParams(['sid', 'test']);

        equal( tracker.hook.test._purify('http://example.com/?sid=12345&test5=1'), 'http://example.com/?test5=1', 'http://example.com/?sid=12345&test5=1');
        equal( tracker.hook.test._purify('http://example.com/?asid=12345&test=1'), 'http://example.com/?asid=12345', 'http://example.com/?asid=12345&test=1');
        equal( tracker.hook.test._purify('http://example.com/?sid=test#hash'), 'http://example.com/', 'http://example.com/?sid=test#hash');

```

js/piwik.js
bx80: I'll get that variable declaration fixed up and take a look at adding those tests. Thanks for the hints..

justinvelluppillai (approved):
Looks good to me to merge. @sgiehl?

justinvelluppillai:
build js

sgiehl:
@justinvelluppillai building js doesn't work for pull requests coming from forks. I guess in theory it would be possible to enable that for our action (if the issuer allows edits by us), but not sure how much effort it would be.

Did you test the changes locally. I had only done a code review yet. If you can confirm that it works as expected, I guess it could be merged (once the build assets are updated)

btw. I have also added another missing JS test.

justinvelluppillai:
@sgiehl yeah I realised that after I attempted it hence my laughing emoji at my comment! How do we normally add the built assets, build them locally and commit to the PR?

sgiehl:
To do that, guess you would need to check out the fork, or at least add it as remote. Alternatively you could merge the PR into a new branch in our repo. Build the assets there and recreate the PR from that branch.

Btw. Feel free to create a new issue to adjust our action to also build assets for forks when possible

bx80:
Closed due to git branch merge issues and replaced by PR #18031

PR #17888, opened 2021-08-14
add piwikHelper and define piwik.helper

### LINKED ISSUE #17787: Javascript console error when clicking Install routine: Download System Check button
Install routine
Page 2

Button "Download System Check"

Does not work. JavaScript error.
grafik

## Expected Behavior
download works without an JavaScript error

## Current Behavior
download does not work

## Your Environment
* Matomo Version: 4.3.1
* Browser: Firefox 90

geekdenz:
fixes #17787

there may well be a better way of including JS requirements
such as piwikHelper.js but this was quick and it might not be
desirable to dig too deep for this

### Description:

### Review

* [ ] Functional review done
* [ ] Code review done

### DIFF
--- a/plugins/Installation/templates/_systemCheckSection.twig
+++ b/plugins/Installation/templates/_systemCheckSection.twig
@@ -1,4 +1,8 @@
 {% import _self as local %}
+<script type="text/javascript" src="plugins/Morpheus/javascripts/piwikHelper.js"></script>
+<script>
+    piwik.helper = piwikHelper;
+</script>
 <p>{{ 'Installation_CopyBelowInfoForSupport'|translate }}
     <br/> <br/>
     <a href="javascript:void(0);"
--- a/plugins/Installation/templates/_systemCheckSection.twig
+++ b/plugins/Installation/templates/_systemCheckSection.twig
@@ -1,15 +1,11 @@
 {% import _self as local %}
-<script type="text/javascript" src="plugins/Morpheus/javascripts/piwikHelper.js"></script>
-<script>
-    piwik.helper = piwikHelper;
-</script>
 <p>{{ 'Installation_CopyBelowInfoForSupport'|translate }}
     <br/> <br/>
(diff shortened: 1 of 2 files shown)


plugins/Installation/templates/_systemCheckSection.twig
```diff
@@ -1,4 +1,8 @@
 {% import _self as local %}
+<script type="text/javascript" src="plugins/Morpheus/javascripts/piwikHelper.js"></script>
```
tsteur: @geekdenz have you tried adding it to `Installation` controller in the `getInstallationJs` method? Does that work? That's the JS that is loaded.

plugins/Installation/templates/_systemCheckSection.twig
```diff
@@ -1,4 +1,8 @@
...
+<script>
+    piwik.helper = piwikHelper;
```
tsteur: Instead of defining here `piwik.helper` it would be better to change the `piwik.helper.sendContentAsDownload` to `piwikHelper.sendContentAsDownload`. That's because outside angular we always refer to it as `piwikHelper` vs in angular we usually load a dependency `piwik` which has `piwik.helper` defined. bit hard to explain but it should be basically `piwikHelper` in the download button.

tsteur (changes requested):

plugins/Installation/templates/_systemCheckSection.twig
geekdenz: OK, sorry, had a feeling about piwikHelper and then checked the issues and you did say.

Fixed now.

plugins/Installation/templates/_systemCheckSection.twig
geekdenz: Yeah that is better, thank you for the hint. At the time I wasn't quite sure where that code was, so created a draft anyway.

sgiehl (approved):
works 👍

PR #17896, opened 2021-08-17
log password reset email errors to log and do not reveal to user

### LINKED ISSUE #17091: Failing SMTP connection might result in information disclosure in password recovery
When the SMTP connection is not set up correctly, or failing due to other reasons like #17026, requesting for password recovery currently displays the full error message returned from the SMTP server. This could for example include the sender mail address or the login.
As the password recovery is public available we should consider not showing the full error message in this case.

geekdenz:
fixes #17091

### Description:

The lines changed may have already achieved this previously. I could not reproduce seeing any revealing details in the errors I provoked. However, it is now ensured that no detail of the exception is logged to the user and they just get told to contact the admin. The exception is logged to the system logs with the injected logger instance. Currently it is only using the implied `__toString()` method of the exception, which should provide adequate detail to admins. However, there might be better detail that could be logged here. Maybe we want to log the message with `$ex->getMessage()` instead like was previously fed back to the user.

Also, I noticed that `Piwik::translate` was used instead of using an injected `Translator` instance. @tsteur mentioned that we prefer to use DI, but there are a few instances of `Piwik::translate` in this file, so I left it for now, also because there it uses the injected class anyway and it might be easier to understand for a significant amount of users/developers.

### Review

### DIFF
--- a/plugins/Login/PasswordResetter.php
+++ b/plugins/Login/PasswordResetter.php
@@ -12,6 +12,7 @@ use Piwik\Access;
 use Piwik\Auth\Password;
 use Piwik\Common;
 use Piwik\Config;
+use Piwik\Container\StaticContainer;
 use Piwik\IP;
 use Piwik\Mail;
 use Piwik\Option;
@@ -104,6 +105,8 @@ class PasswordResetter
      */
     private $emailFromAddress;
 
+    private $logger;
+
     /**
      * Constructor.
      *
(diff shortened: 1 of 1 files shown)


plugins/Login/PasswordResetter.php
```diff
@@ -179,8 +191,8 @@ public function initiatePasswordResetProcess($loginOrEmail, $newPassword)
         } catch (Exception $ex) {
             // remove password reset info
             $this->removePasswordResetInfo($login);
-
-            throw new Exception($ex->getMessage() . Piwik::translate('Login_ContactAdmin'));
+            $this->logger->error($ex); // log exception details to log
```
tsteur, line 194: @geekdenz this would be potentially logged to the user screen. We don't have a functionality to only log to logs and in this case would likely also not want to log potential information to logs.

Could above information also be triggered for different errors where we maybe still want to show the original error message?

geekdenz: We could catch the specific Exception (find out which one - AuthenticationException or similar) and otherwise also log the message. But in case of an AuthenticationException, log:

> An unspecified error occured, please contact an admin.

or similar. Would this be desirable?

tsteur: we don't really need to log anything in particular here as the message will be shown to the user anyway.

geekdenz:
> When this error happens, please check your SMTP settings.

Also tell user where to change settings.

Catch the specific exception and show above error message.

geekdenz:
No need to merge as solution is documentation. See issue.

PR #17901, opened 2021-08-18
Add piwik.js public api method setPagePerformanceTiming.

### LINKED ISSUE #16859: Add JS tracker methods setPagePerformanceTiming and getCustomPagePerformanceTiming to set specific values for performance metrics
Similar to how we used to have `setGenerationTimeMs`.

We then need to document this in https://developer.matomo.org/guides/spa-tracking similar to how the other method used to be documented. Also of course need to document the JS Tracker reference etc.

We'll also need to adjust in JS tracker the logged console message in the `setGenerationTimeMs` method and mention to use this new method instead or maybe we can even somehow map setGenerationTimeMs to the new method but likely this wouldn't make any sense. Same in developer changelog need to adjust the text where we mention the `setGenerationTimeMs` removal.

@sgiehl feel free to close if we already have an issue for it. Couldn't find it.

It's to be seen how this makes sense because some performance metrics might not really make sense in SPA etc. Eg whether we have only one method and by default all metrics will be reset except the ones specified or whether we have individual metrics (but then the problem be that other performance values aren't reset and reports would become wrong).

We need a new JS tracker API method to set these [Performance Tracking HTTP API parameters](https://developer.matomo.org/api-reference/tracking-api#optional-page-performancehttpsmatomoorgfaqhow-tohow-do-i-see-page-performance-reports-info ). A name could be for example `setPagePerformanceTiming(networkTimeInMs, serverTimeInMs, ...)`  and then the related tracking API parameters should be appended in the next tracking request. Basically users should be able to call eg `_paq.push(['setPagePerformanceTiming', 35, 103, ...])`. As far as I know all the parameters should be optional and if no value is set, then we won't send the related tracking parameter. 

When someone sets the custom page performing timings, then we don't detect the timing ourselves.

The new method needs to be documented in https://developer.matomo.org/api-reference/tracking-javascript

Any set value will be sent along the next tracking request and then the configured values are reset so that the same values aren't sent in any following tracking request after that.

geekdenz:
fixes #16859

### Description:

Still need to add documentation.

Is it correct to sendRequest immediately?

Are there other parameters we need to take care of?

### Review


tsteur:
> Is it correct to sendRequest immediately?

@geekdenz it wouldn't be a separate request immediately but the information would be sent along the next tracking request. It's mentioned in the issue.

sgiehl:
@geekdenz I guess this might need to be implemented in another way. The `setPagePerformanceTiming` method might only need to set some internal variables for the performance metrics. And in `sendRequest` we check if custom values are provided and append them in that case and if not fall back to the auto detection. See
https://github.com/matomo-org/matomo/blob/da4df21e808fa983e32c6402c8f3de90145e0910/js/piwik.js#L3773-L3777

When appending custom values we also need to set `performanceTracked = true` to ensure the values aren't tracked twice.
Maybe we should also set `performanceTracked = false` in `setPagePerformanceTiming` to ensure the values are tracked again when some updates them.

geekdenz:
build js

geekdenz:
build js

geekdenz:
build js

geekdenz:
Only failure is now this:

https://app.travis-ci.com/github/matomo-org/matomo/jobs/533572041#L1120-L1122

Looks like it's not caused by my. So OK to merge?

js/piwik.js
```diff
@@ -5571,7 +5590,33 @@ if (typeof window.Matomo !== 'object') {
...
+                /*property pf_net, pf_srv, pf_tfr, pf_dm1, pf_dm2, pf_onl */
+                var data = {
+                    pf_net: networkTimeInMs,
+                    pf_srv: serverTimeInMs,
+                    pf_tfr: transferTimeInMs,
+                    pf_dm1: domProcessingTimeInMs,
+                    pf_dm2: domCompletionTimeInMs,
+                    pf_onl: onloadTimeInMs
+                };
+                configPagePerformanceTiming = queryStringify(data, isNumber);
```
sgiehl: here I guess `performanceTracked` should be set to `false`, to ensure the values are sent (again). This allows to overwrite them with a later request. In addition `performanceAvailable = true` might be needed so `appendAvailablePerformanceMetrics` will be called directly. Otherwise `performanceAvailable` will be set when the pageload finished, but there is no need to wait for that, when the values are set manually.

geekdenz: Thanks @sgiehl ! I was not aware of this. Added also a comment to the code for this as I don't think it is obvious to any developers other than you coming across this.

geekdenz:
build js

geekdenz:
Added some more tests here:

https://github.com/matomo-org/test-examples/commit/33c6d3cabcf5d14b6e55bfe1d1114c712a2dbfaa

geekdenz:
build js

geekdenz:
build js

geekdenz:
build js

geekdenz:
build js

geekdenz:
Please see the updated branch at https://github.com/matomo-org/test-examples/tree/trackingpageperformance to ease testing for this.

js/piwik.js
```diff
@@ -592,6 +592,19 @@ if (typeof window.Matomo !== 'object') {
 
             return matches ? matches[1] : url;
         }
+        function queryStringify(data, filterFn, throwErrorIfNotFilterFn) {
+            var queryString = '', k, filter = filterFn || function(value) { return value; }, isOk;
```
sgiehl: Seems the `isOk` variable isn't used anymore.

```suggestion
            var queryString = '', k, filter = filterFn || function(value) { return value; };
```

js/piwik.js
```diff
@@ -4904,6 +4926,9 @@ if (typeof window.Matomo !== 'object') {
             this.getConfigCookieSameSite = function () {
                 return configCookieSameSite;
             };
+            this.getConfigPagePerformanceTiming = function () {
+                return configPagePerformanceTiming;
```
sgiehl: I'm wondering if that method makes sense how it's currently implemented and if we should think about a better name for it. The page performance timings are actually no config that is set, as it's reset after the request was sent. So the result of this method might even be kind of unexpected in some cases. Better name might be `getCustomPagePerformanceTiming`

js/piwik.js
```diff
@@ -3477,6 +3493,12 @@ if (typeof window.Matomo !== 'object') {
             }
 
             function appendAvailablePerformanceMetrics(request) {
+                if (configPagePerformanceTiming !== '') {
+                    request += configPagePerformanceTiming;
+                    configPagePerformanceTiming = ''; // ensure not logged more than once
```
sgiehl: As mentioned in another comment unsetting this makes the method `getConfigPagePerformanceTiming` bit useless. Instead you could simply set `performanceTracked = true;`, which should avoid further calls to this method.

js/piwik.js
```diff
@@ -5571,7 +5596,55 @@ if (typeof window.Matomo !== 'object') {
...
+                    if (errorCaught) {
+                        message = 'Please ensure to only supply numbers for each parameter to ' +
+                            'setPagePerformanceTiming().';
+                    } else {
+                        message = 'setPagePerformanceTiming() called without parameters. It only makes sense to call ' +
+                            'this function with at least one performance parameter like networkTimeInMs. Also, ' +
+                            'please ensure to only supply numbers for each parameter.';
+                    }
+                    logConsoleError(message);
+                }
```
sgiehl: This code has a bit more complexity as needed. Wouldn't something like this be a bit simpler to read:

```js
var data = {
    pf_net: networkTimeInMs,
    pf_srv: serverTimeInMs,
    pf_tfr: transferTimeInMs,
    pf_dm1: domProcessingTimeInMs,
    pf_dm2: domCompletionTimeInMs,
    pf_onl: onloadTimeInMs
};

try {
    configPagePerformanceTiming = queryStringify(data, isNumber, true);
    if (configPagePerformanceTiming === '') {
        logConsoleError('setPagePerformanceTiming() called without parameters. It only makes sense to call ' +
            'this function with at least one performance parameter like networkTimeInMs. Also, ' +
            'please ensure to only supply numbers for each parameter.');
        return;
    }
                    
    performanceTracked = false; // to ensure the values are sent (again)
    performanceAvailable = true; // so appendAvailablePerformanceMetrics will be called directly
    // Otherwise performanceAvailable will be set when the pageload finished, but there is no need
    // to wait for that, when the values are set manually.
} catch (error) {
    logConsoleError(error.toString() + ' Please ensure to only supply numbers for each parameter to ' +
        'setPagePerformanceTiming().');
}
```

tests/javascript/index.php
```diff
@@ -4216,6 +4235,25 @@ function generateAnIframeInDocument() {
...
+            ok( ! /ShouldNotHave_pf_12345.*pf_net=12345/.test(results), 'setPagePerformanceTiming does NOT set "pf_net=12345" in subsequent request');
+            ok( /ShouldHave_pf_432_578_111.*pf_net=432&pf_srv=578&pf_tfr=111/.test(results), 'setPagePerformanceTiming sets "pf_net=432&pf_srv=578&pf_tfr=111" in request');
+            ok( ! /ShouldNotHave_pf_432_578_111.*pf_net=432&pf_srv=578&pf_tfr=111/.test(results), 'setPagePerformanceTiming does NOT set "pf_net=432&pf_srv=578&pf_tfr=111" in subsequent request');
+            ok( ! /ShouldHave_pf_NO.*pf_net=/.test(results), 'setPagePerformanceTiming sets nothing in request, but default does');
+            ok( ! /ShouldNotHave_pf_NO.*pf_net=/.test(results), 'setPagePerformanceTiming sets nothing in subsequent request');
+            ok( /ShouldHave_pf_1_2_3_4_5_6_7_8.*pf_net=1&pf_srv=2&pf_tfr=3&pf_dm1=4&pf_dm2=5&pf_onl=6/.test(results), 'setPagePerformanceTiming onle set 6 parameters in request');
+            ok( ! /ShouldNotHave_pf_1_2_3_4_5_6_7_8.*pf_net=1&pf_srv=2&pf_tfr=3&pf_dm1=4&pf_dm2=5&pf_onl=6/.test(results), 'setPagePerformanceTiming onle set 6 parameters in request');
```
sgiehl:
```suggestion
            ok( /ShouldHave_pf_1_2_3_4_5_6_7_8.*pf_net=1&pf_srv=2&pf_tfr=3&pf_dm1=4&pf_dm2=5&pf_onl=6/.test(results), 'setPagePerformanceTiming only sets 6 parameters in request');
            ok( ! /ShouldNotHave_pf_1_2_3_4_5_6_7_8.*pf_net=1&pf_srv=2&pf_tfr=3&pf_dm1=4&pf_dm2=5&pf_onl=6/.test(results), 'setPagePerformanceTiming only sets 6 parameters in request');
```

sgiehl (changes requested):
left a couple of comments

js/piwik.js
geekdenz: Set you as the author in the commit. Hope this is what you intended.

js/piwik.js
geekdenz: Agreed. Thanks!

js/piwik.js
```diff
@@ -592,6 +592,19 @@ if (typeof window.Matomo !== 'object') {
 
             return matches ? matches[1] : url;
         }
+        function queryStringify(data, filterFn, throwErrorIfNotFilterFn) {
```
tsteur: @geekdenz this method does two things. It validates data plus it builds a query. You'd probably want to split this. Also all uses of this method have `throwErrorIfNotFilterFn=true` meaning the parameter wouldn't even have been needed. Once you split the method things will get simpler as you no longer need to create a default filter function, no longer need this other third parameter etc.

js/piwik.js
```diff
@@ -592,6 +592,19 @@ if (typeof window.Matomo !== 'object') {
...
+            var queryString = '', k, filter = filterFn || function(value) { return value; };
+            for (k in data) {
+                if (data.hasOwnProperty(k)) {
+                    if (filter(data[k])) {
+                        queryString += '&' + encodeWrapper(k) + '=' + encodeWrapper(data[k]);
+                    } else if (throwErrorIfNotFilterFn && undefined !== data[k]) {
+                        throw new Error('Parameter value not allowed');
```
tsteur: BTW here users wouldn't know which parameter value is not allowed and which value they provided. In these cases we say something like `Parameter "pf_dm1" provided value "sde3" is not valid. Please provided a numeric value.` This will help them to know which parameter is wrong, what they had provided and how they can fix it.

js/piwik.js
```diff
@@ -5571,7 +5596,49 @@ if (typeof window.Matomo !== 'object') {
...
+                    pf_net: networkTimeInMs,
+                    pf_srv: serverTimeInMs,
+                    pf_tfr: transferTimeInMs,
+                    pf_dm1: domProcessingTimeInMs,
+                    pf_dm2: domCompletionTimeInMs,
+                    pf_onl: onloadTimeInMs
+                };
+
+                try {
+                    configPagePerformanceTiming = queryStringify(data, isNumber, true);
```
tsteur: I think here you would be wanting to check if the number is `numeric`. Otherwise a string like `'44.4'` is not accepted as only `44.4` would be allowed. There isn't much of a benefit of not allowing a number as a string (gets in the end converted to a string in the request anyway) but makes it a lot easier for users in case they have the number as a string.

js/piwik.js
```diff
@@ -5571,7 +5596,49 @@ if (typeof window.Matomo !== 'object') {
...
+             * @param transferTimeInMs
+             * @param domProcessingTimeInMs
+             * @param domCompletionTimeInMs
+             * @param onloadTimeInMs
+             */
+            this.setPagePerformanceTiming = function(
+                networkTimeInMs, serverTimeInMs, transferTimeInMs,
+                domProcessingTimeInMs, domCompletionTimeInMs, onloadTimeInMs
+            ) {
+                var data = {
```
tsteur, line 5641: @geekdenz fyi we should similar to the regular performance values https://github.com/matomo-org/matomo/blob/4.4.1/js/piwik.js#L3503 round it to the nearest integer. The values are in milliseconds meaning server side we don't store something like 44.4ms. By rounding it can prevent issues where the URL gets too long if the number was like `44.343493943943943`.

tsteur (changes requested):

geekdenz:
build js

geekdenz:
Before reviewing, maybe give me another chance to check some details tomorrow (2021-09-08). I'll comment once I've rechecked some old manual checks I have not done yet. Might save us all some time.

geekdenz:
build js

geekdenz:
build js

geekdenz:
Not 100% sure but I think this test failure is probably not related to this issue:
https://app.travis-ci.com/github/matomo-org/matomo/jobs/536091064#L1112-L1128
Maybe something a reviewer would want to consider before approving.

js/piwik.js
```diff
@@ -592,6 +592,43 @@ if (typeof window.Matomo !== 'object') {
...
+        function isPositiveNumberString(str) {
+            // !isNaN(str) could be used but does not cover '03' (octal) and '0xA' (hex)
+            // nor negative numbers
+            return (/^[1-9][0-9]*(\.[0-9]+)?$/).test(str);
+        }
+        function onlyNumbers(data) {
+            var result = {}, k;
+            for (k in data) {
+                if (data.hasOwnProperty(k)) {
+                    if (isNumber(data[k]) || isPositiveNumberString(data[k])) {
```
tsteur: just fyi likely you would only need to check for `isPositiveNumberString(data[k])` as it would also work for numbers. It's otherwise you apply different standards where `0` is allowed as a number but not as a string `'0'` for example.

tsteur: also you would probably need to define that `onlyPositiveNumbers` or something as it's bit misleading. 
Might be easier to simply call the method `onlyIntegers` and then convert the value directly to an integer here and remove the method `roundNumbers`. 

Simply because next time we need to use `roundNumbers` who knows what values we would accept. Maybe it would accept 0 or negative value and then things start to get complicated. It's almost too "abstract" now where it was doing too many things before :)

js/piwik.js
```diff
@@ -592,6 +592,43 @@ if (typeof window.Matomo !== 'object') {
...
+                    }
+                }
+            }
+            return result;
+        }
+        function roundNumbers(data) {
+            var result = {}, k;
+            for (k in data) {
+                if (data.hasOwnProperty(k)) {
+                    result[k] = Math.round(Number(data[k]));
```
tsteur: just fyi the `Number()` call wouldn't be needed here. No need to change it though.

js/piwik.js
tsteur: I would usually say it's not needed to change it because the method is used rarely etc. But there's some danger it will be later reused by others and then it becomes tricky that we accept different kind of values. This we should fix and only consistently accept the same kind of values.

js/piwik.js
```diff
@@ -592,6 +592,43 @@ if (typeof window.Matomo !== 'object') {
...
+            // nor negative numbers
+            return (/^[1-9][0-9]*(\.[0-9]+)?$/).test(str);
+        }
+        function onlyNumbers(data) {
+            var result = {}, k;
+            for (k in data) {
+                if (data.hasOwnProperty(k)) {
+                    if (isNumber(data[k]) || isPositiveNumberString(data[k])) {
+                        result[k] = data[k];
+                    } else if (isDefined(data[k])) {
```
tsteur: just fyi because you iterate over `k` you could simplify this to `else {}` and things would still work even if `data[k] === undefined`. We might want to change it as otherwise maybe an error wouldn't be triggered when undefined is past? Haven't tested it though

tsteur (changes requested):
Getting there 🎉  Left a comment re different behaviour of values.

geekdenz:
build js

js/piwik.js
geekdenz: I believe this is all covered with below commits now. Thanks for the pointers. Will also be useful for the future. :-)

js/piwik.js
geekdenz: I know what you mean. However, when a parameter is not passed it is undefined by default and will end up in the object. At some point one has to check for undefined and doing that for every parameter that is passed is not elegant and results in more code with the same outcome. Please consider the attached screenshot and if still in doubt test the code. The only reason it is there is because it was a problem that only came up in my testing.

image

js/piwik.js
geekdenz: I think it'll be a good change and reduces the amount of code. I didn't know that although it seems obvious now since js is loosely typed.

geekdenz:
build js

geekdenz:
Considerations that went into this:

- could have an object parameter where only desired parameters are set - but this would add complexity that has minimal positive impact on users
- could throw an error for each undefined - but this would make code and usage more complicated, need to assume users only pass `undefined` if intended

geekdenz:
I think there are still some improvements that should be made here before merging.

geekdenz:
Documented here:
https://github.com/matomo-org/developer-documentation/pull/549

geekdenz:
build js

geekdenz:
Saw that these seem fine to download:

https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/48972/

Not sure which change made these fail, but it seems not related to this PR.

sgiehl:
> Not sure which change made these fail, but it seems not related to this PR.

@geekdenz Those look like some random failures. Might be good to merge in the latest changes from 4.x-dev into your branch as on 4.x-dev all tests are currently passing.

tsteur:
Tested it locally and IE8 etc and seems to work 👍

sgiehl:
@tsteur Your last commit actually would have required another build of the javascript files. Will fix that directly on 4.x-dev

PR #17910, opened 2021-08-21
Expose methods to disable content security policy or loosen it for embedded iframes

justinvelluppillai:
### Description:

See dev-2271, this PR provides methods for SessionRecording to be able to embed a page in an iframe.

### Review


core/View/SecurityPolicy.php
```diff
@@ -16,6 +16,12 @@
  */
 class SecurityPolicy
 {
+    /*
+     * Commonly used rules
+     */
+    const RULE_DEFAULT = "'self' 'unsafe-inline' 'unsafe-eval' data:";
```
tsteur: @justinvelluppillai is the `data` needed? I read eg 

> data: Allows data: URIs to be used as a content source. This is insecure; an attacker can also inject arbitrary data: URIs. Use this sparingly and definitely not for scripts.

justinvelluppillai: @tsteur Session Recordings need it (TreeMirror js puts a data: image in) which made me expect other places may also. Session Recordings already have it in their rule so I could remove it in RULE_DEFAULT if you think that's unnecessary or better.

tsteur: yeah it be good to have it as much restricted as possible by default. Do we need the `data:` for style attributes? Not sure what it does and fonts?

core/View/SecurityPolicy.php
```diff
@@ -16,6 +16,12 @@
...
+    const RULE_EMBEDDED_FRAME = "'self' 'unsafe-inline' 'unsafe-eval' data: https:";
```
tsteur: Does the `https` mean we allow any https URL? I suppose heatmaps/session recording could also be used with `http`? would we ned to check something like `if (ProxyHttp::isHttps()) then allow also http`? Not sure if that makes sense.

BTW did you check what happens when a recorded page includes a youtube iframe for example? Would that still work to show the youtube iframe?

justinvelluppillai: Perhaps could set both http and https to basically allow anything for the iframe?

I have added a frame-src rule to allow iframes which allows youtube embeds. At this point it's fairly unrestricted in that page except no scripts.

core/View/SecurityPolicy.php
```diff
@@ -89,4 +103,16 @@ public function createHeaderString() {
...
+    /**
+     * A less restrictive CSP which will allow embedding other sites with iframes
+     * (useful for heatmaps and session recordings)
+     *
+     */
+    public function allowEmbedPage() {
+        $this->addPolicy('img-src', self::RULE_EMBEDDED_FRAME . ' data:');
+        $this->addPolicy('style-src', self::RULE_EMBEDDED_FRAME);
+        $this->addPolicy('font-src', self::RULE_EMBEDDED_FRAME);
+        $this->addPolicy('frame-src', self::RULE_EMBEDDED_FRAME);
```
tsteur: Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy do we maybe also need to allow `media-src` and `object-src`? not sure.

Also not sure maybe re `style-src-elem` which we may need to allow to load src from other domains? I'm not so much into it.

Otherwise looks good (haven't tested it yet though)

justinvelluppillai: @tsteur I have a little better understanding of what we're trying to achieve now and probably the best way to do it is to make the `default-src` looser and then have a more restrictive `script-src` to disallow scripts, and maybe an img-src that allows also `data:` scheme.

1. The embedPage uses a different default-src allowing `http:` `https:`
2. The embedPage has a script-src restricting to the same as the rest of matomo
3. The embed Page has a image-src allowing the same as 1. plus also `data:`

sgiehl:
@justinvelluppillai might be unrelated to this issue, but could you check if opening a row evolution shows csp warnings for you?
I'm seeing this for the sparklines:

`
[Report Only] Refused to load the image 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAMgAAAAyCAYAAAAZUZThAAAACXBIWXMAAA7EAAAOxAGVKw4bAAAEdklEQVR4nO3dMWwbZRjG8f+1FUhUCkaCQwJEg6oOrY7GGwMSsYVYbmiDUMxGgpgQQxqJbBWnk7JlSBo2liRdIsVLFswYu0wMiFB5AHVJKySqEwJjqaxmuAtcHMe+JOc7n+/5bc298X2p/Nx3733nM4jIiYy0ByCwtHe3CLhACZgAHgCrK+Xl3TTHJQpI6pb27s4DGydsvr9SXp5LcDjSRQFJUTBz/DSgbHGlvLyWxHjkuAtpDyDn3JhqZEgupT2AnCtFqJn4YOmz+Z+/+/1gQN2+16y1zj0iOUIBSddElKI/f/tnERj05i+alj3o9drAfoRd7kfY3189XqvlNWtRXj8zFJB0PQDeG1T04/b2VBw7My27ABQjlBaBlwbUvAV82PWzgmnZLcDxmrX66Uc4ehSQdK0yO...vTaRDhkUt6aENI0G808E+pFI7xNjAcoOdi/SfUb0yp38iFNhGebJn7gHStb0xrfSM36sCtATXtXJ9ihZ9P5TVrcwpHrjhRanIbEPUb+VatGPvAp31K7lcrxlour2JpfUMOBSvox77+rloxdiFnK+m6n0rkBKZlF03L/tu07Dtpj0WyY2xmkI7juPifWb4JPMG/SrFguG5L91PJWWU+IB3HOTxtutljc7vy/a/1hteeRP2GnME4XMXaonc4ACa+eedq6cvrrykcciaZDkjHcYoMWOwpPHdpYunG6zMJDUnGTM+VdNOyM/HVw9XHfxRnr7wcpbQMaK1DTi3TM8jzFy8UIpZGrRM5oucM4jVrUZbhU3frI6cE7EUorQ93JDKuMj2DGK5bx//kVz9tYGP4o5FxlOmABOY4/gWYYQuG6x4kNBYZM5kPiOG6+8AU/j00YU+AsuG6m4kPSsZG5hcKw4JFwyJwoFlD4vAvjOS9V2w4O8EAAAAASUVORK5CYII=' because it violates the following Content Security Policy directive: "default-src 'self' 'unsafe-inline' 'unsafe-eval'". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.
`

justinvelluppillai:
@sgiehl yes this is relevant to this issue - @tsteur this means we need the `data:` policy most places so I will add it also to the `default-src` directive for all matomo pages.

PR #17917, opened 2021-08-24
Update intl cldr 39

Findus23:
followup to #15607

While looking into the translations, I noticed that this hasn't been updated in a while and the URLs were pointing to the old archived repositories.
I also updated the URLs to not point to the master branch (containing beta versions), but at a stable release.

I replaced VEF with VES (see https://en.wikipedia.org/wiki/Venezuelan_bol%C3%ADvar)

Lastly I ran the script once with the latest 39 release and checking the English, German and French strings the changes are quite minor.

### Review


Findus23:
The largest change I noticed while scrolling through the files (as far as I can see without understanding most languages) is that Spanish now doesn't have any dots anymore after short day/week/etc. names (so `lun` instead of `lun.`)

Also a large fraction of the Georgian file has changed as words like `აზია` have been replaced with `Აზია` with the first letter now being a `GEORGIAN MTAVRULI CAPITAL LETTER AN`. Mtavruli seems to be a concept similar to uppercasing, is used in titles and a rather recent addtion to unicode 11 in 2018.

tsteur:
@Findus23 @sgiehl do we have a command to generate this? Should this be executed periodically say every 6 months? If that's the case be great to know then we can document this internally and create a process for it.

Findus23:
@tsteur Apart from the first two commits this is just a run of `php console translations:generate-intl-data` (and the next time updating `$CLDRVersion`)

tsteur:
how often should this happen?

Findus23:
I added a short description on how to update to https://github.com/matomo-org/developer-documentation/pull/523

I'd say at most this needs to be updated ever major release of CLDR (which is every few months), but in this PR we are skipping three major releases and the changes are still pretty minor. 
So I'd say to do this every 1-2 years or when a country/language/currency in the world changes its name (like with North Macedonia and Eswatini in the last PR).

And when a new Language is added to Matomo, I think this also needs to be done to fetch the strings for this language.

tsteur:
Thanks I've documented this process also internally and set up a reminder every 6 months for us to look into this.

sgiehl (approved):

PR #17918, opened 2021-08-25
If a usable archive exists and the invalidation is for a specific plugin, delete the invalidation

### LINKED ISSUE #17752: Automatically delete plugin specific/report specific invalidations that do not need to run
Sometimes both a plugin specific/report specific archive invalidation will exist alongside an all plugins archive invalidation. When processing these, the all plugins archive is processed first. Then the plugin specific/report specific ones are ignored, since we have a valid archive that was just created. This means those plugin/report specific invalidations never get deleted from the table.

So in this case, if there is a valid archive and the invalidation in question is plugin/report specific, we should just delete the invalidation instead.

justinvelluppillai:
### Description:

Delete plugin invalidations that are ignored because a newer archive exists, otherwise invalidations get stuck there indefinitely.

### Review

### DIFF
--- a/core/CronArchive/QueueConsumer.php
+++ b/core/CronArchive/QueueConsumer.php
@@ -238,7 +238,7 @@ class QueueConsumer
             if ($isUsableExists) {
                 $now = Date::now()->getDatetime();
                 $this->addInvalidationToExclude($invalidatedArchive);
-                if (is_null($invalidatedArchive['plugin'])) {
+                if (empty($invalidatedArchive['plugin'])) {
                     $this->logger->debug("Found invalidation with usable archive (not yet outdated, ts_archived of existing = $archivedTime, now = $now) skipping until archive is out of date: $invalidationDesc");
                 } else {
                     $this->logger->debug("Found invalidation with usable archive (not yet outdated, ts_archived of existing = $archivedTime, now = $now) ignoring and deleting: $invalidationDesc");
--- a/core/CronArchive/QueueConsumer.php
+++ b/core/CronArchive/QueueConsumer.php
@@ -237,8 +237,13 @@ class QueueConsumer
             list($isUsableExists, $archivedTime) = $this->usableArchiveExists($invalidatedArchive);
             if ($isUsableExists) {
                 $now = Date::now()->getDatetime();
-                $this->logger->debug("Found invalidation with usable archive (not yet outdated, ts_archived of existing = $archivedTime, now = $now) skipping until archive is out of date: $invalidationDesc");
                 $this->addInvalidationToExclude($invalidatedArchive);
+                if (is_null($invalidatedArchive['plugin'])) {
+                    $this->logger->debug("Found invalidation with usable archive (not yet outdated, ts_archived of existing = $archivedTime, now = $now) skipping until archive is out of date: $invalidationDesc");
+                } else {
+                    $this->logger->debug("Found invalidation with usable archive (not yet outdated, ts_archived of existing = $archivedTime, now = $now) ignoring: $invalidationDesc");
+                    $this->model->deleteInvalidations([$invalidatedArchive]);
+                }
                 continue;
             } else {
                 $now = Date::now()->getDatetime();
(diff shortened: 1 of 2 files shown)


justinvelluppillai:
Fixes #17752

core/CronArchive/QueueConsumer.php
```diff
@@ -237,8 +237,13 @@ public function getNextArchivesToProcess()
             list($isUsableExists, $archivedTime) = $this->usableArchiveExists($invalidatedArchive);
             if ($isUsableExists) {
                 $now = Date::now()->getDatetime();
-                $this->logger->debug("Found invalidation with usable archive (not yet outdated, ts_archived of existing = $archivedTime, now = $now) skipping until archive is out of date: $invalidationDesc");
                 $this->addInvalidationToExclude($invalidatedArchive);
+                if (is_null($invalidatedArchive['plugin'])) {
+                    $this->logger->debug("Found invalidation with usable archive (not yet outdated, ts_archived of existing = $archivedTime, now = $now) skipping until archive is out of date: $invalidationDesc");
+                } else {
+                    $this->logger->debug("Found invalidation with usable archive (not yet outdated, ts_archived of existing = $archivedTime, now = $now) ignoring: $invalidationDesc");
```
diosmosis: This message could be a bit more specific, letting us know we're explicitly deleting it.

diosmosis:
The code looks good. A test for this could be added to QueueConsumerTest (or an existing test modified).

Most of those tests work the same way, they:

1. setup the QueueConsumer instance
2. insert some archive_invalidations entries
3. then select everything through the QueueConsumer instance, collecting the invalidations in an array
4. and check that the result of that array equals a hardcoded expected value

You could add some invalidations to an existing test (ie, add a plugin invalidation for date/period, then add a recent archive 'done' row so it thinks there's a usable archive (note Date::$now can be set to make it seem recent), and check that it is both not selected and deleted afterwards). You can also do this in a new test by copy-pasting `test_invalidateConsumeOrder()` and changing `$invalidations` and `$expectedInvalidationsFound`, and adding the extra archive row insertion.

core/CronArchive/QueueConsumer.php
justinvelluppillai: Cool ok, I copied the same terminology used in the other deleted invalidations originally but will update to specify that it's deleted

diosmosis: It was just a suggestion, feel free to override. I just thought it would be good to have more info in the logs.

tests/PHPUnit/Integration/CronArchive/QueueConsumerTest.php
```diff
@@ -504,6 +504,82 @@ public function test_invalidateConsumeOrder()
...
+        $expectedInvalidationsFound = [
+            array(
+                ['idarchive' => '1', 'idsite' => '1', 'date1' => '2018-03-04', 'date2' => '2018-03-04', 'period' => '3', 'name' => 'done', 'report' => null, 'plugin' => null, 'segment' => '', 'ts_started' => null, 'status' => '0']
+            ),
+            array()
+        ];
+
+        $this->assertEquals($expectedInvalidationsFound, $iteratedInvalidations, "Invalidations inserted:\n" . var_export($invalidations, true));
+
+        // check that segment hash 2 is no longer in the invalidations table
```
diosmosis: I think the 'segment hash 2' comment needs to be changed/removed? Otherwise the PR looks good to me :+1:

core/CronArchive/QueueConsumer.php
```diff
@@ -237,8 +237,13 @@ public function getNextArchivesToProcess()
             list($isUsableExists, $archivedTime) = $this->usableArchiveExists($invalidatedArchive);
             if ($isUsableExists) {
                 $now = Date::now()->getDatetime();
-                $this->logger->debug("Found invalidation with usable archive (not yet outdated, ts_archived of existing = $archivedTime, now = $now) skipping until archive is out of date: $invalidationDesc");
                 $this->addInvalidationToExclude($invalidatedArchive);
+                if (is_null($invalidatedArchive['plugin'])) {
```
diosmosis: Oh, one small thing, we could use `empty()` instead of `is_null()` to be a bit more thorough.

diosmosis:
Sorry, github notifications said my review was requested, but i think that might have been the original request.

Issue #17907, opened 2021-08-20
Migrate translations to Weblate


sgiehl:
We used to use Transifex for our translation management, but decided to move on to Weblate in the future.
This issue is about to gather and keep track on the tasks we need to do in order to finish the Migration

- [X] Setup a project on Weblate   --> https://hosted.weblate.org/projects/matomo/
- [x] Look through & improve available configurations, filters & validations on Weblate and document the final settings in Confluence
- [x] Setup all resources on Weblate
  - [x] Matomo Core  https://github.com/matomo-org/matomo  (exclude Intl & maybe Example plugins)

  **Matomo open source plugins**
  - [x] Bandwidth Plugin  https://github.com/matomo-org/plugin-Bandwidth
  - [x] CustomAlerts Plugin  https://github.com/matomo-org/plugin-CustomAlerts
  - [x] CustomVariables Plugin  https://github.com/matomo-org/plugin-CustomVariables
  - [x] DeviceDetectorCache Plugin  https://github.com/matomo-org/plugin-DeviceDetectorCache
  - [x] GoogleAnalyticsImporter Plugin  https://github.com/matomo-org/plugin-GoogleAnalyticsImporter
  - [x] LoginLdap Plugin  https://github.com/matomo-org/plugin-LoginLdap
  - [x] MarketingCampaignsReporting Plugin  https://github.com/matomo-org/plugin-MarketingCampaignsReporting
  - [x] Provider Plugin  https://github.com/matomo-org/plugin-Provider
  - [x] QueuedTracking Plugin  https://github.com/matomo-org/plugin-QueuedTracking
  - [x] SecurityInfo Plugin  https://github.com/matomo-org/plugin-SecurityInfo
  - [x] TagManager Plugin  https://github.com/matomo-org/plugin-TagManager
  - [x] TasksTimetable Plugin  https://github.com/matomo-org/plugin-TasksTimetable
  - [x] TrackingSpamPrevention Plugin  https://github.com/matomo-org/plugin-TrackingSpamPrevention
  - [x] VisitorGenerator Plugin  https://github.com/matomo-org/plugin-VisitorGenerator

  **Third Party Plugins**
  - [x] InvalidateReports Plugin  https://github.com/innocraft/plugin-InvalidateReports
  - [x] BotTracker Plugin  https://github.com/Thomas--F/BotTracker https://github.com/Thomas--F/BotTracker/issues/77
[...]

PR #17927, opened 2021-08-27
Prepares Weblate migration

sgiehl:
Due to the migration to weblate some Transifex related stuff can be removed.
Also there might be some changes we need to add before the migration can take place.


lang/README.md
```diff
@@ -1,7 +1,7 @@
 ## Contribute
 
-If you want to improve an existing Matomo translation or contribute a new translation, please have a look at our [Transifex project](https://translations.piwik.org).
+If you want to improve an existing Matomo translation or contribute a new translation, please have a look at our [Weblate project](https://translations.piwik.org).
```
Findus23:
```suggestion
If you want to improve an existing Matomo translation or contribute a new translation, please have a look at our [Weblate project](https://hosted.weblate.org/engage/matomo/).
```
Or we could link to https://matomo.org/translations

sgiehl: or we use https://translations.matomo.org/, which should redirect as well.

Findus23: Indeed, I added a point to https://github.com/matomo-org/matomo/issues/17907 to update the redirect.

Findus23 (approved):

PR #17928, opened 2021-08-29
Improve check for correctly set cookie by casting to string

### LINKED ISSUE #17926: Console error "There was an error setting cookie `mtm_cookie_consent`. Please check domain and path." when calling 'rememberCookieConsentGiven'
I get the error "There was an error setting cookie `mtm_cookie_consent`. Please check domain and path." on every page load in my Chrome browser. 
I was able to narrow it down to this location in the minified matomo.js:
```
                if ((!dt || dt >= 0) && aD(dp) !== dq) {
                    var dk = "There was an error setting cookie `" + dp + "`. Please check domain and path.";
                    ak(dk)
                }
```
The problem is the second part of the condition. `aD(dp) !== dq` is indeed unequal, but only because the type is not correct. The first part returns `1630087076907` as a string, while the second part returns the same `1630087076907` as an integer.

## Expected Behavior
No console error should be thrown.

## Possible Solution
A typecasting would have to take place before (or a `!=` would have to be set instead of a `!==`).

I don't know why others don't see the error message. I can reproduce the error locally and on prod.

## Your Environment

* Matomo Version: 4.4.1
* PHP Version: 7.2.23
* Browser: Chrome 92
* Operating System: Windows 10

justinvelluppillai:
### Description:

Fixes #17926 

Cookies are string values so if we set a cookie with an int value we need to check the string value of that int against the set cookie to make sure it was correctly set (otherwise eg '12345' !== 12345 and we log an unnecessary console error)

### Review


justinvelluppillai:
build js

sgiehl (approved):

PR #17937, opened 2021-08-30
Load sparkline images only once they are near  view port

### LINKED ISSUE #17774: Goal Overview Report & widget on dashboard can cause a very large spike in requests
## Current Behavior

Currently when adding the "Goal Overview" widget to the dashboard or loading the Goal Overview report it will make at least 2 GET requests for each goal present in the Overview (At least one request for each sparkline per goal)
A Matomo server with just 7 goals for example makes around 25 requests just for sparklines on the Goals Overview.

On websites that have a large number of goals (For example >50) this can cause a spike in the number of PHP processes that can result in the maximum number of PHP child processes to be created, this results in the following warning:
`"WARNING: [pool www] server reached pm.max_children setting (50), consider raising it"`

This in turn results in unexpected performance issues that may be difficult to resolve (Other than simply removing the Goal Overview widget from the dashboard, but this doesn't solve the issue when loading the Goal Overview report)

## Possible Solution

Not sure if there is a quick fix for this sort of thing, other than somehow batching the requests for the sparklines so that it doesn't cause the child processes to be maxed?

## Context

For users running servers with a single endpoint (Tracking and UI on the same server) this can potentially result in the entire Matomo server being slow to respond to requests (Since PHP can't handle new requests coming in this could potentially lead to slow tracker performance or other issues)

tsteur:
### Description: 

fix https://github.com/matomo-org/matomo/issues/17774

Test locally and worked nicely. Had 11 goals and on my screen it loaded the first 9 goals right away and the other ones while scrolling. This way, if someone has say 25 goals, then we won't issue like 80 requests at the same time on goals overview or funnels overview page which can cause issues in the web server as well as database.

This won't impact reports like Visitors overview etc. where the sparklines are visible right away.

Works in modern browsers. 

FYI @mattab 

### Review

### DIFF
--- a/core/Twig.php
+++ b/core/Twig.php
@@ -116,7 +116,7 @@ class PiwikTwigFilterExtension extends \Twig\Extension\AbstractExtension
  */
 class Twig
 {
-    const SPARKLINE_TEMPLATE = '<img alt="" data-src="%s" width="%d" height="%d" />
+    const SPARKLINE_TEMPLATE = '<img loading="lazy" alt="" data-src="%s" width="%d" height="%d" />
     <script type="text/javascript">$(function() { piwik.initSparklines(); });</script>';
 
     /**


justinvelluppillai (approved):
This looks/works fine here.

Unrelated, but it feels like the piwik.initSparklines function below should just act on one element rather than all sparkline elements as it is called once for each sparkline anyway.

PR #17942, opened 2021-08-31
Delay call to setVisitorIdCookie to prevent cookie creation when cookies disabled

### LINKED ISSUE #17892: Don't create test cookies when cookies are disabled

justinvelluppillai:
### Description:

Fixes #17892 

### Review


justinvelluppillai:
build js

justinvelluppillai:
build js

justinvelluppillai:
Recreating to fix submodule issues

PR #17943, opened 2021-08-31
When processing a dependent archive for a range, then only process the requested plugin, not all plugins

tsteur:
### Description:

Say you are viewing the goals overview page for a range period. What happens is 

* because it is range it will only archive goals archive see https://github.com/matomo-org/matomo/blob/4.5.0-b1/core/ArchiveProcessor/Rules.php#L65-L73
* then it will call `processDependentArchive` with a segment in https://github.com/matomo-org/matomo/blob/4.5.0-b1/plugins/Goals/Archiver.php#L505
* then it will next time think it needs to process all reports because in https://github.com/matomo-org/matomo/blob/4.5.0-b1/core/ArchiveProcessor/Rules.php#L69-L71 the segment is no longer empty .
* and when then the customer then has eg a segment defined `visitorType==returning` then it would suddenly process all archives because `isSegmentPreProcessed($segment)===true`. Or when requesting a segment that is being pre-processed.

Generally, for ranges when triggered from the browser (not archive), we always only want to archive the specific plugin. 

### Review

### DIFF
--- a/core/ArchiveProcessor/Rules.php
+++ b/core/ArchiveProcessor/Rules.php
@@ -66,6 +66,9 @@ class Rules
             return false;
         }
 
+        if ($periodLabel === 'range' && !Rules::isBrowserArchivingAvailableForSegments() && !SettingsServer::isArchivePhpTriggered()) {
+            return false;
+        }
         if ($segment->isEmpty() && ($periodLabel != 'range' || SettingsServer::isArchivePhpTriggered())) {
             return true;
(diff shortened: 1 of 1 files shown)


tsteur:
Debugged and profiled this and now no longer aggregates multiple archives for all plugins. Only the requested ones. Hoping there are no side effects.

I would add a test but don't have the time plus the outcome stays the same anyway.

PR #17956, opened 2021-09-03
Add CLI config:get and config:delete

### LINKED ISSUE #16576: console (cli): Add config:delete to delete a key

mwithheld:
### Description:

Issue #16576 Add logic to CLI for config:delete and config:get

### Review


plugins/CoreAdminHome/Commands/ConfigDelete.php
```diff
@@ -0,0 +1,319 @@
...
+You can delete config values per the two sections below, where:
+- [Section] is the name of the section, e.g. database or General.
+- [config_setting_name] the name of the setting, e.g. username.
+- [array_value] For arrays, the specific array value to delete.
+
+(1) By settings options --section=[Section] and --key=[config_setting_name], and optionally --value=[array_value].  Examples:
+#Delete this setting.
+$ ./console %command.name% --section=database --key=username
+#Delete one value in an array:
+$ ./console %command.name% --section=database --key=username --value=DevicesDetection
```
sgiehl: for consistency:
```suggestion
$ ./console %command.name% --section=PluginsInstalled --key=PluginsInstalled --value=DevicesDetection
```

plugins/CoreAdminHome/Commands/ConfigDelete.php
```diff
@@ -0,0 +1,319 @@
...
+                throw new \InvalidArgumentException('When using options, --section and --key must be set');
+            case ! empty($argument):
+                $settingStr = $argument;
+                break;
+            case ! empty($options):
+                $settingStr = implode('.', $options);
+                break;
+            default:
+                // We should not get here, but just in case.
+                throw new Exception('Some unexpected error occurred parsing input values');
```
sgiehl:
```suggestion
                throw new \Exception('Some unexpected error occurred parsing input values');
```

plugins/CoreAdminHome/Commands/ConfigDelete.php
```diff
@@ -0,0 +1,319 @@
...
+        //Optionally could set this at runtime with: $output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE.
+        $debug = false;
+
+        // Gather options, then discard ones that are empty so we do not need to check for empty later.
+        $options = array_filter([
+            'section' => $input->getOption('section'),
+            'key' => $input->getOption('key'),
+            'value' => $input->getOption('value'),
+        ]);
+        $debug && fwrite(STDERR, PHP_EOL . __FUNCTION__ . '::Started with $options=' . (empty($options) ? '' : serialize($options)));
```
sgiehl: Personally I would vote for removing all those debug code. It actually makes the code bit harder to read. And imho debugging the code has more benefits than using plenty of debug output.
Not sure if maybe @tsteur has another opinion on that.

plugins/CoreAdminHome/Commands/ConfigDelete.php
```diff
@@ -0,0 +1,319 @@
...
+        if ( ! ($settingWrapped->setting instanceof SystemConfigSetting)) {
+            throw new \InvalidArgumentException('This function expects $settingWrapped->setting to be a SystemConfigSetting instance');
+        }
+
+        $config = Config::getInstance();
+
+        // Check the setting exists and user has permissions. If so, put it in the wrapper.
+        switch (true) {
+            case ! $settingWrapped->setting->isWritableByCurrentUser():
+                throw new \Exception('No write permissions to this setting');
```
sgiehl: Console commands are always executed as super user, so there is actually no need to check for permission.

plugins/CoreAdminHome/Commands/ConfigDelete.php
```diff
@@ -0,0 +1,319 @@
...
+            $debug && fwrite(STDERR, PHP_EOL . __FUNCTION__ . "::About to delete config for \$settingStr={$settingStr}");
+            $result = $this->deleteConfigSetting($settingWrapped);
+            $debug && fwrite(STDERR, PHP_EOL . __FUNCTION__ . '::Got $result=' . serialize($result));
+
+            if ($result) {
+                $output->writeln($this->wrapInTag('info', self::MSG_SUCCESS));
+            }
+        }
+
+        //Many matomo script output Done when they're done.  IMO it's not needed: $output->writeln(self::wrapInTag('info', 'Done.'));
```
sgiehl: The command prints a success or error message already. So no need to print anything else.
```suggestion
```

plugins/CoreAdminHome/Commands/ConfigDelete.php
```diff
@@ -0,0 +1,319 @@
...
+        ]);
+        $debug && fwrite(STDERR, PHP_EOL . __FUNCTION__ . '::Started with $options=' . (empty($options) ? '' : serialize($options)));
+
+        $argument = trim($input->getArgument('argument'));
+        $debug && fwrite(STDERR, PHP_EOL . __FUNCTION__ . '::Started with $argument=' . (empty($argument) ? '' : serialize($argument)));
+
+        // Sanity check inputs.
+        switch (true) {
+            case empty($argument) && empty($options):
+                throw new \InvalidArgumentException('You must set either an argument or set options --section and optional --key');
```
sgiehl: Guess it should be this:
```suggestion
                throw new \InvalidArgumentException('You must set either an argument or set options --section, --key and optional --value');
```

plugins/CoreAdminHome/Commands/ConfigDelete.php
```diff
@@ -0,0 +1,319 @@
...
+        $settingWrappedNew->isArray = is_array($section->$settingName);
+        $debug && fwrite(STDERR, PHP_EOL . __FUNCTION__ . "Set \$settingWrappedNew->isArray={$settingWrappedNew->isArray}");
+
+        if ( ! $settingWrappedNew->isArray && ! empty($settingWrappedNew->arrayVal)) {
+            throw new \InvalidArgumentException('This config setting is not an array');
+        }
+        if ($settingWrappedNew->isArray) {
+            $debug && fwrite(STDERR, PHP_EOL . __FUNCTION__ . "This config setting is an array");
+            if (empty($settingWrappedNew->arrayVal)) {
+                throw new \InvalidArgumentException('This config setting an array, and no array value was specified for deletion');
```
sgiehl:
```suggestion
                throw new \InvalidArgumentException('This config setting is an array, but no array value was specified for deletion');
```

plugins/CoreAdminHome/Commands/ConfigDelete.php
```diff
@@ -0,0 +1,319 @@
...
+    private const MSG_SUCCESS = 'Success: The setting has been deleted';
+
+    protected function configure()
+    {
+        $this->setName('config:delete');
+        $this->setDescription('Delete a config setting');
+        $this->addArgument(
+            'argument',
+            InputArgument::OPTIONAL,
+            "A config setting in the format Section.key or Section.array_key[], e.g. 'Database.username' or 'PluginsInstalled.PluginsInstalled[CustomDimensions]"
```
sgiehl:
```suggestion
            "A config setting in the format Section.key or Section.array_key[], e.g. 'Database.username' or 'PluginsInstalled.PluginsInstalled.CustomDimensions'"
```

plugins/CoreAdminHome/Commands/ConfigGet.php
```diff
@@ -0,0 +1,343 @@
...
+     * Throws an error on json_encode failure.
+     *
+     * @param mixed $var The variable to convert.
+     * @return string The JSON-formatted string.
+     */
+    private function toJson($var): string
+    {
+        $result = json_encode($var, JSON_UNESCAPED_SLASHES);
+        if ($result === false) {
+            throw new Exception('Failed to json_encode');
```
sgiehl:
```suggestion
            throw new \Exception('Failed to json_encode');
```

plugins/CoreAdminHome/Commands/ConfigGet.php
```diff
@@ -0,0 +1,343 @@
...
+
+use Piwik\Config;
+use Piwik\Plugin\ConsoleCommand;
+use Piwik\Settings\FieldConfig;
+use Piwik\Settings\Plugin\SystemConfigSetting;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Yaml\Yaml;
```
sgiehl: This might actually only be a dev dependency, so it might not be available in releases. But you might use `\Spyc` instead, which is included through the device detector dependency.

plugins/CoreAdminHome/Commands/ConfigGet.php
```diff
@@ -0,0 +1,343 @@
...
+            'format' => $input->getOption('format'),
+        ]);
+        $debug && fwrite(STDERR, PHP_EOL . __FUNCTION__ . '::Started with $options=' . (empty($options) ? '' : serialize($options)));
+
+        // If none specified, set default format.
+        if ( ! isset($options['format']) || empty($format = $options['format']) || ! in_array($format, self::OUTPUT_FORMATS, true)) {
+            $format = self::OUTPUT_FORMAT_DEFAULT;
+        }
+        // Since it is config and not input data, remove format from the options so we can just work with the section and format.
+        unset($options['format']);
```
sgiehl: Maybe handling the format independently makes the code a bit cleaner
```suggestion
        // Gather options, then discard ones with an empty value so we do not need to check for empty later.
        $options = array_filter([
            'section' => $input->getOption('section'),
            'key' => $input->getOption('key'),
        ]);

        // If none specified, set default format.
        $format = $input->getOption('format');
        if ( empty($format) || ! in_array($format, self::OUTPUT_FORMATS, true)) {
            $format = self::OUTPUT_FORMAT_DEFAULT;
        }
```

plugins/CoreAdminHome/tests/Integration/ConfigDeleteTest.php
```diff
@@ -0,0 +1,425 @@
...
+    public function setUp(): void
+    {
+        self::removeTestConfigFile();
+        parent::setUp();
+    }
+
+    public function tearDown(): void
+    {
+        parent::tearDown();
+    }
```
sgiehl: could be removed, as not needed.
```suggestion
```

plugins/CoreAdminHome/tests/Integration/ConfigDeleteTest.php
```diff
@@ -0,0 +1,425 @@
...
+
+    private function runCommandWithArguments(string $sectionName, string $settingName = '', string $value = ''): object
+    {
+        $debug = false;
+        $debug && fwrite(STDERR, PHP_EOL . str_repeat('-', 80) . PHP_EOL . __FUNCTION__ . "::Started with \$sectionName=$sectionName; \$settingName=$settingName; \$value=$value") . PHP_EOL;
+
+        $inputArr = [
+            'command' => self::COMMAND,
+            '-vvv' => false,
+            'argument' => $sectionName . $settingName . (empty($value) ? '' : ".$value"),
```
sgiehl: This would actually need to be this in order to work:
```suggestion
            'argument' => $sectionName . '.' . $settingName . (empty($value) ? '' : ".$value"),
```

plugins/CoreAdminHome/tests/Integration/ConfigDeleteTest.php
```diff
@@ -0,0 +1,425 @@
...
+
+        $this->assertStringContainsString('InvalidArgumentException', $output);
+    }
+
+    public function testEmptyArgsShouldYieldError()
+    {
+        $debug = false;
+
+        // Pass empty section name.
+        $resultObj = $this->runCommandWithArguments('');
```
sgiehl: Was it on purpose to only have one test running with arguments, that actually only passes an empty one? Maybe might be useful to also have some working tests and maybe some edge cases. (Same applies for the ConfigGetTest)

sgiehl (changes requested):
@mwithheld Thanks for creating this pull request. I did a review and left a couple of comments. Would be nice if you have some time to apply that

plugins/CoreAdminHome/Commands/ConfigDelete.php
mwithheld: Good point - done.

plugins/CoreAdminHome/Commands/ConfigGet.php
mwithheld: Thanks - Symfony Yaml is not in the pre-packaged download, anyway.  Changed to use Spyc.

plugins/CoreAdminHome/tests/Integration/ConfigDeleteTest.php
mwithheld: Good catch.  Fixed.

plugins/CoreAdminHome/tests/Integration/ConfigDeleteTest.php
mwithheld: They just fell off my radar.  Added tests with args to ConfigGetTest and ConfigDeleteTest classes.

plugins/CoreAdminHome/Commands/ConfigDelete.php
mwithheld: I have removed the debug code.

sgiehl (approved):
Thanks again @mwithheld for applying the changes. I've added some last fixes and now it should be good to merge 🎉

PR #17962, opened 2021-09-03
delete old log_* table data also without idvisit

### LINKED ISSUE #16529: Old data in log tables without idvisit column are not purged (affects plugins)
According to customers in https://forum.matomo.org/t/form-analytics-doesnt-cleanup-its-old-form-field-form-page-log-data/38567

> I’m using Matomo (3.14.0) with Form Analytics (3.1.28). Both “delete old raw data” and “delete old aggregated report data” policies are in place (with core:archive cron job). Still, in practice, I see that Form Analytics purges records from the log_form db table only, leaving log_form_field and log_form_page tables filled with all the historical data from the beginning of time.
> I can confirm that old records are correctly purged from log_form but not from log_form_page/log_form_field tables, matching them by idvisit/idlogform. How can I clean that old data and prevent its accumulation in the future?

We would expect that log_form_page and log_form_field are purged indeed.

from @tsteur this is supposed to be done by core but it looks like it’s not fully implemented there for log tables without idvisit column.

geekdenz:
fixes #16529

### Description:

I know and realise I have been reinventing the wheel here a bit and there is redundant code here. Also, I know already there are better ways of doing things. But I thought I'd spent a bit of time and some of it might still be useful at some point.

I guess we could find dependent tables in the sort function I wrote somehow and then join on `log_visit`'s idcolumn.

### Review

### DIFF
--- a/core/Common.php
+++ b/core/Common.php
@@ -1225,4 +1225,40 @@ class Common
         }
         return $validLanguages;
     }
+
+    /**
+     * compare 2 strings by strcmp but putting longer strings first
+     * @param string $a compare subject left
+     * @param string $b compare subject right
+     * @return int
+     */
+    public static function longerStrCmp(string $a, string $b)
+    {
+        $alen = strlen($a);
+        $blen = strlen($b);
+        $len = $alen - $blen;
+        if ($len === 0) {
+            return strcmp($a, $b);
+        }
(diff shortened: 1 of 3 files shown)


geekdenz:
not using this as it was only for communication and possible ideas

PR #17964, opened 2021-09-04
delete old log_* table data also without idvisit

### LINKED ISSUE #16529: Old data in log tables without idvisit column are not purged (affects plugins)

geekdenz:
fixes #16529

### Description:

Reuse logic from DataSubject and refactor into core.

To test this manually, please see https://github.com/matomo-org/test-examples/blob/trackingformlogs/logdeleter/general/index.php . You can configure it with https://github.com/matomo-org/test-examples/blob/trackingformlogs/config/config.example.php .

### Review

### DIFF
--- a/core/DataAccess/RawLogDao.php
+++ b/core/DataAccess/RawLogDao.php
@@ -145,22 +145,6 @@ class RawLogDao
     }
 
     /**
-     * Deletes conversions for the supplied visit IDs from log_conversion. This method does not cascade, so
-     * conversion items will not be deleted.
-     *
-     * @param int[] $visitIds
-     * @return int The number of deleted rows.
-     */
-    public function deleteFromLogTable($tableName, $visitIds)
-    {
-        $sql = "DELETE FROM `" . Common::prefixTable($tableName) . "` WHERE idvisit IN "
-             . $this->getInFieldExpressionWithInts($visitIds);
-
-        $statement = Db::query($sql);
-        return $statement->rowCount();
-    }
-
-    /**
      * Deletes conversion items for the supplied visit IDs from log_conversion_item.
      *
      * @param int[] $visitIds
--- a/core/LogDeleter.php
+++ b/core/LogDeleter.php
@@ -8,8 +8,10 @@
 
 namespace Piwik;
 
+use Piwik\Container\StaticContainer;
 use Piwik\DataAccess\RawLogDao;
 use Piwik\Plugin\LogTablesProvider;
+use Piwik\Plugins\PrivacyManager\Model\DataSubjects;
 use Piwik\Plugins\SitesManager\Model;
 
 /**
@@ -98,8 +100,13 @@ class LogDeleter
         $this->rawLogDao->forAllLogs('log_visit', $fields, $conditions, $iterationStep, function ($logs) use ($logPurger, &$logsDeleted, $afterChunkDeleted) {
             $ids = array_map(function ($row) { return (int) (reset($row)); }, $logs);
             sort($ids);
-            $logsDeleted += $logPurger->deleteVisits($ids);
+//            $logsDeleted += $logPurger->deleteVisits($ids);
 
+            $deleteCounts = StaticContainer::get(DataSubjects::class)->deleteDataSubjectsWithoutInvalidatingArchives(array_map(function($visitid) {
+                return ['idvisit' => $visitid];
+            }, $ids)); // [1,2] => [['idvisit'=>1],['idvisit'=>2]]
+            $currentCount = array_sum(array_values($deleteCounts)); // ['a' => 3, 'b' => 5] => 8
+            $logsDeleted += $currentCount;
             if (!empty($afterChunkDeleted)) {
                 $afterChunkDeleted($logsDeleted);
             }
--- a/core/LogDeleter.php
+++ b/core/LogDeleter.php
@@ -45,18 +45,10 @@ class LogDeleter
      */
     public function deleteVisits($visitIds)
     {
-        $numDeletedVisits = 0;
-
-        foreach ($this->logTablesProvider->getAllLogTables() as $logTable) {
-            if ($logTable->getColumnToJoinOnIdVisit()) {
-                $numVisits = $this->rawLogDao->deleteFromLogTable($logTable->getName(), $visitIds);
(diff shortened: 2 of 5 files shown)


geekdenz:
You'll need to apply this patch, otherwise it won't get into the code.

See https://github.com/matomo-org/test-examples/blob/trackingformlogs/logdeleter/general/index.php .

```patch
diff --git a/plugins/PrivacyManager/PrivacyManager.php b/plugins/PrivacyManager/PrivacyManager.php
index f11ff26b43..98055c9150 100644
--- a/plugins/PrivacyManager/PrivacyManager.php
+++ b/plugins/PrivacyManager/PrivacyManager.php
@@ -672,6 +672,7 @@ class PrivacyManager extends Plugin
         $deleteIntervalDays = $settings[$setting];
         $deleteIntervalSeconds = $this->getDeleteIntervalInSeconds($deleteIntervalDays);
 
+        return true;
         if ($lastDelete === false ||
             $lastDelete === '' ||
             ((int)$lastDelete + $deleteIntervalSeconds) <= time()

```

sgiehl:
@geekdenz Wouldn't it make sense to add/change some tests in [LogDeleterTest](https://github.com/matomo-org/matomo/blob/4.x-dev/tests/PHPUnit/Integration/LogDeleterTest.php) to prove that the new logic works as expected. Ideally the new/changed tests would fail on 4.x-dev but work with the changes of this PR.

tsteur:
@sgiehl it's not really needed since the logic we use itself is already tested.

sgiehl (dismissed):
Just tested it locally, seems to work as expected 👍

tsteur (changes requested):
@geekdenz it seems like maybe some other methods can now be removed as well like `Rawlogdao::deleteFromLogTable`?

The only other thought I have was that before we were doing `idvisit IN(1,2,3,4,4)` and now we use `idvisit=1 or idvisit = 2 or idvisit = 3`. I believe this is though the same for the query optimiser. Might be something to quickly double check / look into.

Also before we were doing `idvisit in (1,2,3,4)` and now we use placeholders using `?`. This is a regression and we should change the below query in `DataSubjects` to not use placeholders. When hundreds or thousands of placeholders are used then MySQL can get very very slow. We learned this over the years. We don't need to bind the parameters there if we cast the value to an integer.

geekdenz:
> `idvisit IN(1,2,3,4,4)`

will be more efficient, though just by a constant factor (`strlen('idvisit = 1 OR ') - strlen('1,')`), even just because the text is shorter. The query optimiser might optimise this automatically though, but it adds another step for it. It'll add up with lots of IDs. I'll do it as it is not difficult.

geekdenz:
OK. Double checked manually and got expected results:

```sql
mysql> SELECT distinct b.idvisit, c.visit_last_action_time, a.idlogform, a.idformview, a.field_name FROM matomo_log_form_field as a join matomo_log_form as b on a.idlogform=b.idlogform JOIN matomo_log_visit AS c ON c.idvisit=b.idvisit;
+---------+------------------------+-----------+------------+------------+
| idvisit | visit_last_action_time | idlogform | idformview | field_name |
+---------+------------------------+-----------+------------+------------+
|   10444 | 2020-08-01 00:00:00    |        24 | zaODNJ     | input1     |
|   10444 | 2020-08-01 00:00:00    |        24 | zaODNJ     | input2     |
|   10444 | 2020-08-01 00:00:00    |        24 | zaODNJ     | input3     |
|   10449 | 2020-08-01 00:00:00    |        28 | 9GlRNX     | input1     |
|   10449 | 2020-08-01 00:00:00    |        28 | 9GlRNX     | input2     |
|   10449 | 2020-08-01 00:00:00    |        28 | 9GlRNX     | input3     |
|   10443 | 2021-09-07 03:05:54    |        23 | ZDa9lV     | input1     |
|   10443 | 2021-09-07 03:05:54    |        23 | ZDa9lV     | input2     |
|   10443 | 2021-09-07 03:05:54    |        23 | ZDa9lV     | input3     |
+---------+------------------------+-----------+------------+------------+
9 rows in set (0.01 sec)

mysql> SELECT distinct b.idvisit, c.visit_last_action_time, a.idlogform, a.idformview, a.field_name FROM matomo_log_form_field as a join matomo_log_form as b on a.idlogform=b.idlogform JOIN matomo_log_visit AS c ON c.idvisit=b.idvisit;
+---------+------------------------+-----------+------------+------------+
| idvisit | visit_last_action_time | idlogform | idformview | field_name |
+---------+------------------------+-----------+------------+------------+
|   10443 | 2021-09-07 03:05:54    |        23 | ZDa9lV     | input1     |
|   10443 | 2021-09-07 03:05:54    |        23 | ZDa9lV     | input2     |
|   10443 | 2021-09-07 03:05:54    |        23 | ZDa9lV     | input3     |
+---------+------------------------+-----------+------------+------------+
3 rows in set (0.01 sec)
```

tsteur:
Fyi  tweaked the test cases to make sure it works fine when idvisit and idvisit+idsite is mixed and also to know it works when multiple deletions are requested. Was the easiest way for me to test the generated queries work.

Screenshot 2021-09-13 at 3 04 19 PM

Screenshot 2021-09-13 at 3 20 57 PM

tsteur:
Generally tested for quite a while and looks all good to me. Waiting for automated tests to finish before merging 🎉

geekdenz:
Thanks @tsteur for the added test and info!

PR #17970, opened 2021-09-06
Fix failing UI tests

sgiehl:
### Description:

Due to various code changes (like #17937, #17879) UI tests are currently failing. This PR aims to fix them again.

### Review


sgiehl:
Note: I've introduced a new method `page.waitForLazyImages()`, which removes the loading attribute from all images and then waits for all images to load. Otherwise we can't ensure in some UI tests which images would be loaded or not.

tsteur (commented):
Does it make sense to call this automatically as part of

* `this.waitForNetworkIdle() `
* `this.webpage.on('load', () => {`  (I know those ones are only loaded later but if it might still be good just in case we use it more in the future)
* maybe also in `this.webpage.on('requestfinished'`

Then developers maybe won't need to think about this as it can be tricky to find out etc

justinvelluppillai (approved):
Nice one.

justinvelluppillai:
@sgiehl sorry maybe jumped the gun merging there didn't notice @tsteur 's suggestion. Can create a new PR if we want to implement those suggestions also.

sgiehl:
@tsteur will check if that works and create a new PR if so

PR #18002, opened 2021-09-15
add mysql row format dynamic option

### LINKED ISSUE #16834: Use row_format=dynamic by default when creating a table
see https://dev.mysql.com/doc/refman/8.0/en/innodb-row-format.html

when creating a table can we specify the dynamic format like `CREATE TABLE t1 (c1 INT) ROW_FORMAT=DYNAMIC;`?

refs https://github.com/matomo-org/matomo/issues/16810 to avoid row size too large error.

We'd need to check that this works on all MySQL or not sure if we need to check whether `dynamic` row format is supported or whether some MySQL maybe don't support it.

peterhashair:
### Description:

Add row format equals dynamic as default, regarding the large text field for creating table failure. 
### Review


core/Db/Settings.php
```diff
@@ -32,9 +32,14 @@ public function getUsedCharset()
         return strtolower($this->getDbSetting('charset'));
     }
 
+    public function getRowFormat()
+    {
+       return $this->getDbSetting('disable_dynamic_row_format')?'':'ROW_FORMAT=DYNAMIC';
```
sgiehl: The setting `disable_dynamic_row_format` doesn't exist in database configuration and would need to be added here: https://github.com/matomo-org/matomo/blob/7e0b0e602068a558a84f1e16fde5e7e6cb0b7fa4/config/global.ini.php#L43-L47

We can't simply use the dynamic row format for all instances, as some might break. So we need to disable that feature by default in `global.ini.php` and overwrite the value automatically if that feature is supported. This needs to be done while installing Matomo like here:
https://github.com/matomo-org/matomo/blob/09c012345d6913364af475598872105ff2fc854a/plugins/Installation/Controller.php#L608
and within a new updates script like this one https://github.com/matomo-org/matomo/blob/03410d62c6885010bc0968dca9ebe5e23b7c8e87/core/Updates/4.4.0-b1.php#L31-L36
To check if the feature is supported guess we could simply check if `DbHelper::getDefaultCharset()` returns `utf8mb4` or not.

Btw. Imho using some more spaces makes the code easier to read. e.g.
```php
return $this->getDbSetting('disable_dynamic_row_format') ? '' : 'ROW_FORMAT=DYNAMIC';
```

sgiehl (changes requested):

config/global.ini.php
```diff
@@ -45,6 +45,9 @@
 ; Matomo should work correctly without this setting but we recommend to have a charset set.
 charset = utf8
 
+; disable dynamic row by default
+disable_dynamic_row_format = true
```
tsteur: fyi here we usually use `1` for enabled and `0` for disabled.

Also might be more readable by naming it `enable_dynamic_row_format = 0` but not too important

core/Updates/4.5.0-b1.php
```diff
@@ -29,11 +29,16 @@ public function __construct(MigrationFactory $factory)
 
     public function getMigrations(Updater $updater)
     {
-        $migration1 = $this->migration->db->changeColumnType('session', 'data', 'MEDIUMTEXT');
+        $migrations = [];
+        $migrations[] = $this->migration->db->changeColumnType('session', 'data', 'MEDIUMTEXT');
+
+        //database using utf8mb4 set row format dynamic
+        if (DbHelper::getDefaultCharset() == "utf8mb4") {
+            $migrations[] = $this->migration->config->set('disable_dynamic_row_format', false);
```
tsteur: fyi similar to above comment we usually would use here `0` instead of `false` for these config values.

config/global.ini.php
```diff
@@ -45,6 +45,9 @@
 ; Matomo should work correctly without this setting but we recommend to have a charset set.
 charset = utf8
 
+; disable dynamic row by default
```
tsteur: could we maybe explain a bit more why someone would want dynamic rows? eg mention it avoids "row size too large errors" for new tables. or maybe we could also mention that "we recommend enabling this setting if the default charset is set to utf8mb4".

cc @sgiehl  does that make sense?

core/Db/Settings.php
peterhashair: thx sgiehL, updated it

config/global.ini.php
```diff
@@ -45,6 +45,10 @@
 ; Matomo should work correctly without this setting but we recommend to have a charset set.
 charset = utf8
 
+; Matomo recommend enabling this setting if the default charset is set to utf8mb4
```
tsteur: can we slightly tweak this to `Matomo recommends enabling this setting if the default charset is set to "utf8mb4". It can avoid the MySQL error "row size too large" for new tables.`?

core/Updates/4.5.0-b1.php
```diff
@@ -29,11 +29,16 @@ public function __construct(MigrationFactory $factory)
 
     public function getMigrations(Updater $updater)
     {
-        $migration1 = $this->migration->db->changeColumnType('session', 'data', 'MEDIUMTEXT');
+        $migrations = [];
+        $migrations[] = $this->migration->db->changeColumnType('session', 'data', 'MEDIUMTEXT');
+
+        //database using utf8mb4 set row format dynamic
+        if (DbHelper::getDefaultCharset() == "utf8mb4") {
```
tsteur: @peterhashair because we have already release 4.5.0-b1 we would usually need to put this migration into a new file `4.5.0-b2`. It's not too important maybe for this one so we can leave it for now and don't need to change it.

tsteur (commented):
@sgiehl I'm not quite sure what we do during the installation. Are we using utf8mb4 by default when possible? Would we need to do anything during the installation so users benefit from this?

core/Updates/4.5.0-b1.php
peterhashair: @tsteur Removed.

tsteur: @peterhashair maybe you forgot to commit the new update fail? Not seeing the migration anymore anywhere

tsteur:
@sgiehl be great to let us know if you know maybe whether we need to do anything on installation re this setting

sgiehl:
Yes. We need to use a default that works for all users and change the value on installation. Them same way it's done for utf8mb4

peterhashair:
@tsteur Updated the Installation script, but not sure about the **ms sql** option, I set it `false` for now.

core/Db/Settings.php
```diff
@@ -32,9 +33,14 @@ public function getUsedCharset()
         return strtolower($this->getDbSetting('charset'));
     }
 
+    public function getRowFormat()
+    {
+        return $this->getDbSetting('enable_dynamic_row_format') ? 'ROW_FORMAT=DYNAMIC' : '';
```
tsteur: @peterhashair I'm now actually wondering if we actually need this config setting? If I understand it now right, then we can apply that row format whenever `$this->getUsedCharset() === 'utf8mb4'`?  Is that correct @sgiehl  or would we still need a setting in case some DB doesn't support this row format?

plugins/Installation/Controller.php
```diff
@@ -149,6 +149,9 @@ function databaseSetup()
 
                 DbHelper::checkDatabaseVersion();
 
+                // check if database is UTF8MB4 
+                $dbInfos['enable_dynamic_row_format'] = Db::get()->isConnectionUTF8MB4() ? 1 : 0;
```
tsteur: fyi I believe here we can simplify this using 
```php
        $config->database['enable_dynamic_row_format'] = $config->database['charset'] === 'utf8mb4' ? 1 : 0;
```

in the `createConfigFile` method. Then you won't need any of the methods using `isConnectionUTF8MB4` I suppose

core/Db/Settings.php
peterhashair: @tsteur  That's right, do you want me to remove the configuration, apply it to default if ut8mb4.

plugins/Installation/Controller.php
peterhashair: @tsteur that's right, let me tide this up.

core/Db/Settings.php
sgiehl: In theory there might be cases where `ROW_FORMAT=DYNAMIC` might be supported, but `utf8mb4` collation is disabled manually. Not sure if anyone would do that or if that makes sense at all 🤷 
But guess for simplicity we could for now go without the config option and if there are coming too many complains that it breaks something we can introduce the config option later maybe?

core/Db/AdapterInterface.php
```diff
@@ -70,4 +70,10 @@ public function isErrNo($e, $errno);
      * @return bool
      */
     public function isConnectionUTF8();
+
+    /**
```
tsteur: fyi this can be removed now

core/Db/Schema/Mysql.php
```diff
@@ -514,12 +514,14 @@ public function createTable($nameWithoutPrefix, $createDefinition)
     {
         $dbSettings   = new Db\Settings();
         $charset      = $dbSettings->getUsedCharset();
+        $dynamic = $charset==='utf8mb4' ? 'ROW_FORMAT=DYNAMIC' : '';
```
tsteur: technically it could be nice to still have the row format logic in `$dbSettings->getRowFormat()` then it is more reusable but not too important.
the variable you maybe want to name `$rowformat` instead of `$dynamic`

core/DbHelper.php
```diff
@@ -160,6 +160,18 @@ public static function isDatabaseConnectionUTF8()
         return Db::get()->isConnectionUTF8();
     }
 
+    /**
+     * Check database connection character set is utf8.
+     *
+     * @return bool  True if it is (or doesn't matter); false otherwise
+     */
+    public static function isDatabaseConnectionUTF8MB()
```
tsteur: this one we'll no longer need 👍

core/Updates/4.5.0-b2.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+namespace Piwik\Updates;
+
+use Piwik\DbHelper;
+use Piwik\Updater;
+use Piwik\Updates as PiwikUpdates;
+use Piwik\Updater\Migration\Factory as MigrationFactory;
+
+class Updates_4_5_0_b2 extends PiwikUpdates
```
tsteur: this update script we can now remove as well now that we removed the config setting

tsteur (changes requested):
Nice, we're getting there. Looks a lot cleaner without the config 👍  sorry for having to remove it but it'll simplify things in the future 🚀

peterhashair:
@tsteur Sorry about the issue being dragging too long, I should scope it a little bit better at the beginning.

tsteur:
@peterhashair all good, the problem was we didn't scope the issue well enough in the actual issue. We'll do better :) And good we found that we can simplify it all that's great.

core/Db/Schema/Mysql.php
```diff
@@ -515,11 +515,12 @@ public function createTable($nameWithoutPrefix, $createDefinition)
         $dbSettings   = new Db\Settings();
         $charset      = $dbSettings->getUsedCharset();
 
-        $statement = sprintf("CREATE TABLE IF NOT EXISTS `%s` ( %s ) ENGINE=%s DEFAULT CHARSET=%s ;",
+        $statement = rtrim(sprintf("CREATE TABLE IF NOT EXISTS `%s` ( %s ) ENGINE=%s DEFAULT CHARSET=%s %s;",
```
sgiehl: guess the `rtrim` is actually obsolete, as the query always ends with a `;`

sgiehl (commented):
left a minor comment. Otherwise looks good to me

core/Db/Schema/Mysql.php
peterhashair: @sgiehl Thanks for pointing that out. :)

sgiehl (approved):
Looks good to me know

PR #18014, opened 2021-09-15
disable http requests on config

### LINKED ISSUE #17589: Add config option to disable http requests in System Checks to prevent server crashes with mod_security
When opening the Admin area since 4.3 it shows a few more issues like "Required Private Directories" and a new database issue (can't see it right now, since server is down). When I open System Check to find out more about it, the page doesn't load and the whole server is not available anymore (ERR_CONNECTION_REFUSED). I can't even access Plesk.

It's reproducible and didn't happen in 4.2. Which log might be useful to find out what's going wrong?

## Expected Behavior
Opening System Check should open it

## Current Behavior
Opening System check crashes server.

Open System check

## Your Environment
Can't access it right now but what I know:
* Matomo Version: 4.3
* PHP Version: 8.0
* Server Operating System:

geekdenz:
### Description:

fixes #17589

### Review


config/global.ini.php
```diff
@@ -833,6 +833,9 @@
 ; If set to 1, when rearchiving reports in the past we do not rearchive segment data with those reports. Default is 0.
 rearchive_reports_in_past_exclude_segments = 0
 
+; Disable HTTP checks on install or Diagnostics
+disable_http_diagnostics = false
```
tsteur: @geekdenz could we use a different name for this as it only disables the private directory checks? And could we mention when someone would disable this check? 

Also  we usually use 0/1 instead of true/false.

And it might be more readable to use `enable_... = 1`.

FYI to disable external HTTP requests we already have a config: https://matomo.org/faq/troubleshooting/faq_16646/

This one won't disable all HTTP checks in diagnostics (there are few others) but only the private directory checks

plugins/Diagnostics/Diagnostic/AbstractPrivateDirectories.php
```diff
@@ -88,7 +89,22 @@ private function getUrlList(array $testUrls)
 
     protected function isAccessible(DiagnosticResult $result, $testUrl, $publicIfResponseEquals, $publicIfResponseContains)
     {
+        static $addedDisableHttpDiagnosticsWarning = false;
```
tsteur: Maybe put this logic in the `execute` method right after the check whether internet is disabled? Then you won't need to use `static` (which can cause issues in tests etc) and it will be executed only one for each private directory check. You can then say `This check was skipped as this check is disabled in the config. To enable this check set [...]... in the "config/config.ini.php" file.`

tsteur (changes requested):

config/global.ini.php
geekdenz: OK, thanks for letting me know. My commit below resolves this, I believe.

plugins/Diagnostics/Diagnostic/AbstractPrivateDirectories.php
geekdenz: Not 100% sure I implemented what you mean here. Please check below.

tsteur: @geekdenz I mentioned to put it into the `execute` method right after the check whether internet is disabled. That's where it needs to go so you don't need to check if it has been triggered before etc and this way others will know what's happening etc. Could you check if that makes sense?

geekdenz: Ah! Sorry. I think I understood now what you mean. Will do.

geekdenz: Still running Travis tests so not taken out of draft, but if you have time and patience, you could check if i implemented what you meant, @tsteur . Locally it works.

tsteur: 👍  haven't tested it but that looks better. You won't need `$this->addedDisableHttpDiagnosticsWarning` anymore now and fyi you're mentioning in the ini description that it runs the checks on "install" but in the beginning of the execute method we're making sure that we don't run these tests on install as they wouldn't make quite sense yet as some of these files wouldn't exist just yet etc. Meaning you will want to change eg `Enable HTTP checks on install or Diagnostics` to `Enable HTTP checks in the diagnostic system check`

geekdenz:
The SEO test fails consistently here, but this seems to have nothing to do with this PR.

plugins/Diagnostics/Diagnostic/AbstractPrivateDirectories.php
```diff
@@ -18,6 +19,7 @@ abstract class AbstractPrivateDirectories implements Diagnostic
 {
     protected $privatePaths = [];
     protected $accessiblePaths = []; // used like a set, but hashtable used underneath anyway, so map simpler php way
+    protected $addedDisableHttpDiagnosticsWarning = false;
```
tsteur: @geekdenz not too important but you shouldn't need this property anymore now that it is in the execute method.

config/global.ini.php
```diff
@@ -835,6 +835,12 @@
 ; If set to 1, when rearchiving reports in the past we do not rearchive segment data with those reports. Default is 0.
 rearchive_reports_in_past_exclude_segments = 0
 
+; Enable HTTP checks on install or Diagnostics for required and recommended private directories.
+; Set this to 0 if you need to skip it because your hosting provider makes your site inaccessible
+; for some times if it is run, e.g. on install or when running diagnostics.
+; Default is 1.
+enable_required_directories_diagnostic = 1
```
tsteur, line 839: see https://github.com/matomo-org/matomo/pull/18014#discussion_r710674258 this is not executed on "install"

tsteur (changes requested):
@geekdenz Left two minor comments @geekdenz be also good to put the PR in current milestone.

config/global.ini.php
tsteur: @geekdenz can we add back the other removed valuable information from the previous comment like 

`http checks ...for required and recommended private directories.`
 and 
`; Set this to 0 if you need to skip it because your hosting provider makes your site inaccessible`

The config doesn't disable all HTTP checks, that's what the internet enabled config is there for. And we usually give users an idea when the setting may need to be disabled and how they can do this. So be good if we add back this information.

plugins/Diagnostics/Diagnostic/AbstractPrivateDirectories.php
```diff
@@ -65,6 +66,18 @@ public function execute()
         }
 
         $result = new DiagnosticResult($label);
+        if (Config::getInstance()->General['enable_required_directories_diagnostic'] == 0) {
+            if (!$this->addedDisableHttpDiagnosticsWarning) {
```
tsteur: @geekdenz the `addedDisableHttpDiagnosticsWarning` property was removed but here it's still in the code

geekdenz: Oh man! Sorry! Stupid mistake. Thought it was a trivial change, did it in vim quickly and was planning to wait for the ci tests and forgot to put it in draft.

sgiehl:
Just out of curiosity: Why do we need an option to disable certain diagnostic checks? If someone wants to remove the private directory checks he can simply achieve that by creating a `config\config.php` with this content:
```php
return [
    'diagnostics.disabled' => DI\add([
        DI\get(\Piwik\Plugins\Diagnostics\Diagnostic\RequiredPrivateDirectories::class),
        DI\get(\Piwik\Plugins\Diagnostics\Diagnostic\RecommendedPrivateDirectories::class),
    ])
];
```

tsteur:
@sgiehl we usually don't ask users to edit `config.php` if it can be avoided. The ini file is a lot easier to change for people, especially non-technical people. And once they have more entries in `config.php` it may be easy for them to break things. Also this way it's independent of code. Like recently we changed class names for these and then it wouldn't work anymore or even the entire Matomo could break if we removed these classes maybe as the files would no longer exist.

geekdenz:
I think this might be unrelated: https://builds-artifacts.matomo.org/matomo-org/matomo/m-17589-4.x-dev/49276/UsersManager_permissions_bulk_access_set.png

config/global.ini.php
tsteur: @geekdenz this comment was resolved but we still need to tweak the description see previous comments. There's a needs review label on it but PR is also in status `Draft` so not sure if you still work on it or if it's ready for review.

geekdenz: Meant to review this earlier today. But only got around to it now. Will try to keep in mind to remove the label also.

plugins/Diagnostics/lang/en.json
```diff
@@ -6,6 +6,7 @@
         "MysqlMaxPacketSize": "Max Packet Size",
         "MysqlMaxPacketSizeWarning": "It is recommended to configure a 'max_allowed_packet' size in your MySQL database of at least %1$s. Configured is currently %2$s.",
         "ConfigFileIntroduction": "Here you can view the Matomo configuration. If you are running Matomo in a load balanced environment, the page might be different depending on from which server this page is loaded. Rows with a different background color are changed config values that are specified for example in the %1$s file.",
+        "EnableRequiredDirectoriesDiagnostic": "This check was skipped as this check is disabled in the config. To enable this check set [General] enable_required_directories_diagnostic in the \"config/config.ini.php\" file.",
```
tsteur: Just tested it and worked generally 👍  The last tweak I would make is to change the part `[General] enable_required_directories_diagnostic in the` to `[General] enable_required_directories_diagnostic = 1 in the` so it's more clear what needs to happen as otherwise users may be confused. Then it should be good to merge.

tsteur:
@geekdenz this looks good to merge. 

For some reason thought the PR build always fails see eg https://app.travis-ci.com/github/matomo-org/matomo/jobs/539047123#L1189

and also in the previous run see https://app.travis-ci.com/github/matomo-org/matomo/builds/238279592

Not sure if there went something wrong with .gitmodules etc. It says:

image

@bx80 experienced this recently as well and fixed it with https://developer.matomo.org/guides/git#fixing-the-error-fatal-remote-error-upload-pack-not-our-ref

Issue #17965, opened 2021-09-05
In Acquisition > Overview report, the Row selector is missing


mattab:
In the Acquisition > Overview report, the Row selector is missing below the evolution at the top.

Whereas you can still see the row selector under the  evolution graph in "Visitors > Overview" (other evolution graphs show it as well).

Expected the same behavior as in "Visitors > Overview", but in Acquisition > Overview it is missing.

it is problematic because often you need to see the evolution over multiple more than the default number of days or weeks or months, so it means users can't get more context/historical data.

Reproduced in Matomo Cloud  in https://demo.matomo.cloud/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=day&date=yesterday&segment=&category=Referrers_Referrers&subcategory=General_Overview and also in 4.5.0-b1

(this is a regression as it used to work)

tsteur:
@mattab it's not clear what you are meaning. 

maybe you mean https://github.com/matomo-org/matomo/issues/17499 where it works until you select more than 2 records to plot?

heurteph-ei:
@mattab @tsteur , maybe this:
![image](https://user-images.githubusercontent.com/79993552/132182280-a6361e1d-ff69-418a-ab50-bf8715957cd5.png)
![image](https://user-images.githubusercontent.com/79993552/132182407-0599f06c-0261-4d3b-bdf4-77382788f415.png)
(note: screenshots taken from version 4.3.1)

mattab:
@heurteph-ei thanks for posting the screenshot :+1:

sgiehl:
#17987 didn't solve that properly. See https://github.com/matomo-org/matomo/pull/17987#issuecomment-920095127

PR #17987, opened 2021-09-10
if not comparing, set show_limit_control to true for Evolution Report

geekdenz:
fixes #17965

### Description:

Assumes we want the report to show_limit_control in all cases but when comparing.

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #18017, opened 2021-09-16
only show_limit_control on Acquisition/Overview

geekdenz:
### Description:

Not super beautiful fix, but it works.

### Review


tsteur:
@geekdenz did you have a look at @sgiehl comment in https://github.com/matomo-org/matomo/pull/17987#issuecomment-920095127 ? Did changing this one to `true` maybe not fix the problem?

geekdenz:
@tsteur I already checked his comment and fixed it here:
https://github.com/matomo-org/matomo/commit/24a4cfe9c1b506aa7bcf8ac3bc61366e8bfac24c
Did you see that or is there something wrong with the fix?

geekdenz:
build js

geekdenz:
@sgiehl I believe this means it is fixed:

https://builds-artifacts.matomo.org/matomo-org/matomo/m-17965/49074/Comparison_row_evolution.png

sgiehl (changes requested):
@geekdenz please don't build JS files in pull request that actually didn't change anything in the piwik.js. This was actually forgotten when merging another PR and should either be directly fixed on 4.x-dev or with a separate PR. Here it will lead to merge conflicts.

The solution you proposed might fix the problem described in the issue, but actually might not be the simplest / best way.
The default value of `show_limit_control` is true. So it is enabled by default. See 
https://github.com/matomo-org/matomo/blob/be724571d2f477b4f553411d41df3fb22408c0a6/core/ViewDataTable/Config.php#L329-L335
or specifically for the evolution charts here:
https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/plugins/CoreVisualizations/Visualizations/JqplotGraph/Evolution/Config.php#L41-L45
So the reason why the limit control is not shown for the referrer evolution chart is, that it's disabled in the according report class:
https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/plugins/Referrers/Reports/GetReferrerType.php#L105
Simply removing this one line should fix the issue. Overwriting that again doesn't make sense at all.

geekdenz:
Sorry, @sgiehl and thanks for the suggestion.

tsteur:
I'll close https://github.com/matomo-org/matomo/issues/17965

fyi @geekdenz if you put `fix #17965` in the PR description then it will close the related issue automatically when merging.

PR #18023, opened 2021-09-16
Allow social & search definitions to be sourced locally

samjf:
### Description:

Issue: DEV-2088

**Introduced functionality:**
This change allows Matomo to source [search + social definitions yaml files](https://github.com/matomo-org/searchengine-and-social-list) locally, instead of reading them from remotely sourced updates.

If the introduced configuration variable is changed to 0 (false) then it will only ever load social and search definitions from the local YML files packaged with Matomo. 

 **Existing behavior**
How it works at the moment is that a weekly task will fetch the latest definition files from [Matomo github](https://github.com/matomo-org/searchengine-and-social-list). It will store this in the the Option table and retrieve it from the option table when definitions are requested and there is a cache miss.

### Review


config/global.ini.php
```diff
@@ -833,6 +833,10 @@
 ; If set to 1, when rearchiving reports in the past we do not rearchive segment data with those reports. Default is 0.
 rearchive_reports_in_past_exclude_segments = 0
 
+; If set to 1, then social and search engine definitions files will be downloaded from Matomo remotely to update
```
tsteur: These are actually downloaded from github see https://github.com/matomo-org/matomo/blob/4.4.1/plugins/Referrers/Tasks.php#L33-L49 and not from Matomo. This is something we need to improve in the future to "proxy" or serve them through Matomo servers.

tsteur (commented):
Left a quick comment @samjf I haven't tested yet but code looks good. Still needs to be tested (by me or some other reviewer)

config/global.ini.php
samjf: @tsteur Thanks for that. I've made a change based on your suggestion :)

config/global.ini.php
```diff
@@ -835,6 +835,10 @@
 ; If set to 1, when rearchiving reports in the past we do not rearchive segment data with those reports. Default is 0.
 rearchive_reports_in_past_exclude_segments = 0
 
+; If set to 1, then social and search engine definitions files will be will be synchronised using the internet.
```
sgiehl:
```suggestion
; If set to 1, then social and search engine definitions files will be synchronised using the internet.
```

config/global.ini.php
```diff
@@ -835,6 +835,10 @@
...
+; When set to 0 the definitions will be loaded from the local definitions (updated with core).
+enable_referrer_definition_syncs = 1
```
sgiehl, line 840: Should we mention here that this setting actually requires `enable_internet_features` to be enabled?

tsteur: Can mention it 👍

PR #18024, opened 2021-09-16
Check config file was written correctly

JasonMortonNZ:
### Description:

Issue: dev-2272

Whenever we save the config file, it be great to double check if the config file was written correctly. This PR adds a couple of small sanity checks on the config file upon writing its content and writes a log entry should the sanity checks not pass.

### Review


core/Config.php
```diff
@@ -410,13 +410,12 @@ protected function writeConfig()
                 throw $this->getConfigNotWritableException();
             }
 
+            if (!$this->sanityCheck($localPath)) {
```
tsteur: @JasonMortonNZ should we maybe try to save the file content once more if it failed and only then log if it's still a problem? Maybe this way we would prevent a potential issue that no config file exists.

core/Config.php
```diff
@@ -455,4 +454,24 @@ public static function setSetting($sectionName, $name, $value)
...
+
+    /**
+     * Sanity check a config file by checking its size and content
+     *
+     * @param string $localPath
+     * @return bool
+     */
+    public function sanityCheck(string $localPath): bool
+    {
+        $content = @file_get_contents($localPath);
```
tsteur, line 478: @JasonMortonNZ  I'm thinking before reading the content we would maybe want to call this method  
 to make sure it's read again from file system https://www.php.net/manual/en/function.clearstatcache.php

core/Config.php
```diff
@@ -455,4 +454,24 @@ public static function setSetting($sectionName, $name, $value)
...
+        $filesize = filesize($localPath);
+
+        if ($filesize < 1500 || strpos($content, '[database]') === false) {
```
tsteur: fyi I'm thinking we might actually only want to check the `[database]` if also `$output` above contained this section in the string to write.

Reason being that I just noticed eg in Matomo for WordPress we don't save a `[database]` section as it reads the WordPress DB credentials. Also some might actually define the DB using this plugin so there wouldn't be a DB section https://github.com/matomo-org/plugin-EnvironmentVariables

Maybe the sanity check would purely need to check if `trim($output)=== trim($content)`?

tsteur (changes requested):
Nice work 👍 , left a few comments as realised the check we use on Cloud might not work for On-Premise. Sorry about that.

core/Config.php
```diff
@@ -455,4 +457,25 @@ public static function setSetting($sectionName, $name, $value)
...
+
+    /**
+     * Sanity check a config file by checking contents
+     *
+     * @param string $localPath
+     * @return bool
+     */
+    public function sanityCheck(string $localPath, $expectedContent): bool
+    {
+        clearstatcache();
```
tsteur: @JasonMortonNZ be good to only invalidate the stat cache for the specific file ideally (second parameter)

JasonMortonNZ: @tsteur narrowed this down to the specific file 👍

core/Config.php
```diff
@@ -410,13 +410,15 @@ protected function writeConfig()
                 throw $this->getConfigNotWritableException();
             }
 
+            if (!$this->sanityCheck($localPath, $output)) {
+                // If sanity check fails, try to write the contents once more before logging the issue.
+                if (@file_put_contents($localPath, $output, LOCK_EX === false)) {
```
tsteur: @JasonMortonNZ I think there's a typo re the `=== false` which should be outside the function call?

Was also wondering if we should maybe check again the `if` to also do the sanity check again but not 100% sure. It may be useful I suppose but not sure if there could be any issues with that. 

```php
if (@file_put_contents($localPath, $output, LOCK_EX) === false
                    || !$this->sanityCheck($localPath, $output)) {
```

core/Config.php
```diff
@@ -410,13 +410,15 @@ protected function writeConfig()
...
+                    StaticContainer::get(LoggerInterface::class)->info("The configuration file {$localPath} did not write correctly.");
```
tsteur, line 416: I think we will want to post the event here before the log call in case we adjust the above `if` statement to also check for the `sanityCheck`. Otherwise we would post the event twice. I suppose putting the event here would make it harder to test if the event was triggered and it be no longer possible to properly test it I suppose.

core/Config.php
```diff
@@ -410,13 +410,15 @@ protected function writeConfig()
...
+                }
+            }
+
             $this->settings->getIniFileChain()->deleteConfigCache();
 
-            /**
```
tsteur, line 415: did you remove the description on purpose @JasonMortonNZ ?

core/Config.php
```diff
@@ -455,4 +457,25 @@ public static function setSetting($sectionName, $name, $value)
...
+     * @return bool
+     */
+    public function sanityCheck(string $localPath, $expectedContent): bool
+    {
+        clearstatcache(true, $localPath);
+
+        $content = @file_get_contents($localPath);
+
+        if (trim($content) !== trim($expectedContent)) {
+            Piwik::postEvent('Core.configFileSanityCheckFailed', [$localPath]);
```
tsteur: For the event it would be great to also add a little bit of text through a comment. The event will then be documented automatically on https://developer.matomo.org/api-reference/events

tsteur (changes requested):

core/Config.php
JasonMortonNZ: @tsteur yes, I was thinking about sanity check in this above `if` too and moving the event to be posted within, but it was tricky for the tests. I have added an optional 3rd argument `$notify` to the `sanityCheck()` function which allows us to set this to true on the second try, and this will trigger the event to be posted. This should prevent the event being posted twice, and also keep testability.

core/Config.php
JasonMortonNZ: @tsteur does the comment have to follow a specific format for it to be automatically documented?

tsteur: @JasonMortonNZ something similar to https://github.com/matomo-org/matomo/pull/18024/files#diff-8edb7d933e422f060b96277680eaada01d6ed1a7139e20963363b9973d625d12L415 where the comment is for the configFileChanged event. Basically, each variable in the array should be documented using `@param` and otherwise just regular comment AFAIK

PR #18028, opened 2021-09-17
add heartbeat ping tab tracking

### LINKED ISSUE #16459: Heartbeat timer ping request is not always sent (focusing away from the tab should trigger the ping request)
Heartbeat timer ping request is not always sent (focusing away from the tab should trigger the ping request). For example, can’t see a ping request when doing:

*    Open test page
*    Scroll for a bit. 
*    Wait 30 seconds.
*    Click on another tab.
*    Come back to the original tab.
*    got: no ping request, but expected one in this case?

We discussed internally and a comment was: focusing another browser tab works for me as well in incognito mode. It doesn’t work though in my regular browser window either because of some extension or because there are so many tabs open or so. I’d suggest you test this in incognito as well and see if it works there. Then also make sure to close all other tabs and eg have only two tabs open and no extension. Generally the code is adding the blur event to the window and if the browser doesn’t trigger it there’s not too much we can do. 

we can look to maybe additionally use the visibility API if it helps maybe:
```
document.addEventListener( 'visibilitychange' , function() {
if (document.hidden) {…
}, false );
```

peterhashair:
### Description:

When the user comes back to the original tab, will send a ping request again.

### Review


peterhashair:
build js

js/piwik.js
```diff
@@ -6251,7 +6260,7 @@ if (typeof window.Matomo !== 'object') {
             this.enableHeartBeatTimer = function (heartBeatDelayInSeconds) {
                 heartBeatDelayInSeconds = Math.max(heartBeatDelayInSeconds || 15, 5);
                 configHeartBeatDelay = heartBeatDelayInSeconds * 1000;
-
+ console.log("here");
```
tsteur: haven't tested it but fyi there is a console.log call here

peterhashair: @tsteur ah, sorry about that. let me remove this.

js/piwik.js
```diff
@@ -2894,6 +2894,14 @@ if (typeof window.Matomo !== 'object') {
                 }
             }
 
+            function heartBeatOnVisible() {
+                //come back to original tab
+                if (document.visibilityState === 'visible' && hadWindowMinimalFocusToConsiderViewed()) {
```
tsteur: I haven't tested it, but usually we send the `ping` event on "blur" meaning when it becomes hidden. I wonder if we need to test here for 

```js
if (documentAlias.visibilityState === 'hidden' && hadWindowMinimalFocusToConsiderViewed()) {
    heartBeatPingIfActivityAlias();
} else if (documentAlias.visibilityState === 'visible') {
    timeWindowLastFocused = new Date().getTime();
}
```

seems something like this would maybe replicate the focus/blur event behaviour? Again, I haven't tested though. Just wanted to give you some quick feedback. 

Also we usually don't use `document` directly but `documentAlias`. This way the JS minifier can optimise these calls using a shorter variable.

peterhashair: @tsteur you are right, they replicate the blur behavior.

peterhashair:
build js

peterhashair:
build js

peterhashair:
build js

tsteur:
Nice, worked 👍  also tested some IEs etc.

js/piwik.js
```diff
@@ -6267,9 +6276,11 @@ if (typeof window.Matomo !== 'object') {
                     if (windowAlias.removeEventListener) {
                         windowAlias.removeEventListener('focus', heartBeatOnFocus);
                         windowAlias.removeEventListener('blur', heartBeatOnBlur);
+                        windowAlias.removeEventListener('visibilitychange', heartBeatOnVisible);
                     } else if  (windowAlias.detachEvent) {
                         windowAlias.detachEvent('onfocus', heartBeatOnFocus);
                         windowAlias.detachEvent('onblur', heartBeatOnBlur);
+                        windowAlias.removeEventListener('visibilitychange', heartBeatOnVisible);
```
paladox, line 6283: Shouldn't this be detachEvent?

peterhashair:
@paladox thanks for that, you are absolutely right. Let me do a quick fix for that.

PR #18031, opened 2021-09-19
JavaScript tracker exclude query parameters

### LINKED ISSUE #11627: Piwik JavaScript Tracker: Add possibility to set URL parameters to ignore

bx80:
### Description:

Fixes #11627

(This is a re-submission of PR 17887 to work around merge failure)

Adds a new option to the JavaScript tracker to exclude query parameters from the tracked URL.

_paq.push(["setExcludedQueryParams", ["uid", "sid"]]);

Existing purify() and removeUrlParameter() functions were reused.

The tracker code generator has been updated to automatically add this option to the generated javascript tracking code using the excluded query parameters configured under Websites > Settings.

Tracker code generator unit tests have also been updated to test for valid tracker code and string escaping with the option enabled.

(Initially used "setIgnoreUrlParams" as the option name as per issue report, but though it would be more consistent to use "setExcludedQueryParams" to match the existing config option and UI terminology.)

### Review

* [x ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)


bx80:
build js

justinvelluppillai (commented):
Have just been looking at this and I can't see existing excluded params from Administration > Websites > Manage appearing in the generated tracking code. Is there something I'm missing?

edit: I see, the API endpoint only pulls the global excluded query params, not those applicable to the specific website I was expecting.

justinvelluppillai (approved):
Works here.

sgiehl:
If I see that correctly the Tracking Code Generator currently automatically adds a `setExcludedQueryParams` to the tracking code if any excluded params are defined globally or for the website. I'm wondering if it might make sense to make it possible to adjust that value (only within the code generator). @tsteur do you think we should add that?

justinvelluppillai:
@sgiehl I don't see where it gets the excluded params for the website - looks like it only gets the global ones currently?

tsteur:
@sgiehl not sure. I'm kind of thinking we wouldn't need to expose this to the tracker code generator UI at all as it's rarely needed maybe. I'm not sure if we should expose the global list of excluded URL parameters. Not sure if there could be any risk with this or not. I suppose it might not be. Generally, there is also the list of excluded parameters from the config `url_query_parameter_to_exclude_from_url=gclid,fbclid,fb_xd_fragment,fb_comment_id,phpsessid,jsessionid,sessionid,aspsessionid,doing_wp_cron,sid,pk_vid` which maybe would need to be added automatically too. However, excluding some parameters like `gclid,fbclid` can break functionality. 

@mattab what are your thoughts? The above logic implements that global excluded config parameter gets exposed in the tracking code generator. Say you have configured this: 
image

then it adds below code `_paq.push(["setExcludedQueryParams", ["foo","bar"]]);`.

image

Having the excludedParameters API parameter in the tracking code generator is 👍  but I would maybe not add any global excluded parameter by default. I would probably also not add a UI to enter extra parameters to exclude as it's simply needed so rarely. On the other side someone could argue when they copy/paste the tracking code they can still remove the line for removing URL parameters if they didn't want to expose this information but often users wouldn't even know what it is.

Note that the global excluded parameters list supports REGEXP which I think the JS logic isn't. 

To have a bit more security by default I would exclude these parameters by default `phpsessid,jsessionid,sessionid,aspsessionid`. There's also `sid|zenid` etc but these may be used for other purposes in an application meaning we cannot exclude them by default.

js/piwik.js
```diff
@@ -2502,12 +2505,16 @@ if (typeof window.Matomo !== 'object') {
...
             function purify(url) {
-                var targetPattern;
+                var targetPattern, i;
 
                 // we need to remove this parameter here, they wouldn't be removed in Matomo tracker otherwise eg
                 // for outlinks or referrers
                 url = removeUrlParameter(url, configVisitorIdUrlParameter);
 
+                for (i = 0; i < configExcludedQueryParams.length; i++) {
+                    url = removeUrlParameter(url, configExcludedQueryParams[i]);
```
tsteur, line 2515: btw does `removeUrlParameter` remove parameters case insensitive? If someone was to configure `phpsessionid` for example I wonder if it should also remove `PHPSESSIONID`

bx80:
If we're not wanting to expose the excluded parameters in the generated tracking code then maybe we could hash the values when generating the tracking code and then have purify() check those hashes against the hashed value of the parameters? That way it wouldn't be possible for users to just read off a list of interesting parameters from the tracking code.

js/piwik.js
bx80: At the moment it's case sensitive, easy enough to change removeUrlParameter() to be case insensitive. It's an edge case, but that could produce unwanted effects, eg. if a site had multiple query parameters with the same name in a different case used for different things. ?pid=123&pID=abc Then both would be removed.

tsteur:
I reckon with rainbow tables etc the original value could be often found out quite quickly. See internal slack we would for now go only with exposing the website specific excluded parameters. 
That means we probably can't exclude any other parameters like `phpsessid,jsessionid,sessionid,aspsessionid` by default which is fine.

If someone needs to exclude parameters then they can still exclude them manually by adjusting the tracking code.

js/piwik.js
tsteur: let's keep it for now as is and not change anything 👍

sgiehl:
Is there actually any need to include the server side excluded parameters into the tracking code? Imho we could simply not include them there at all. The server side exclusion will still work the same...
But having the option in the tracker nevertheless might make sense, as it allows to remove parameters from being sent along with the tracking request.

tsteur:
@sgiehl we don't need it there. In the issue back then I mentioned we could do it. I reckon now that it's already it can be a good benefit that we show the site specific excluded parameters automatically. It's good for privacy and security to remove these already in the site.

bx80:
build js

sgiehl:
build js

plugins/CoreAdminHome/angularjs/trackingcode/jstrackingcode.controller.js
```diff
@@ -66,12 +67,23 @@
...
+                module: 'API',
+                method: 'Overlay.getExcludedQueryParameters',
+                idSite: idSite,
+                filter_limit: '-1'
+            }).then(function (data) {
+                self.siteExcludedQueryParams[idSite] = data || [];
+                self.isLoading = false;
                 callback();
             });
+
```
sgiehl: This would actually call `callback` twice. But it should be loaded when everything is finished, right? So guess the simplest solution would be to execute the queries not in parallel by moving the second request into the first one's `then` part, like:

```suggestion
                // Load site excludedQueryParams
                piwikApi.fetch({
                    module: 'API',
                    method: 'Overlay.getExcludedQueryParameters',
                    idSite: idSite,
                    filter_limit: '-1'
                }).then(function (data) {
                    self.siteExcludedQueryParams[idSite] = data || [];
                    self.isLoading = false;
                    callback();
                });
            });

```

otherwise the `isLoading` wouldn't be set correctly all the time, as one request might still be running...

sgiehl:
It could make sense to also reflect those changes in the UI tests. We already have a test that shows the tracking code generator here: https://github.com/matomo-org/matomo/blob/03dbff2aaff77b93cef5158e8509f3c370947748/tests/UI/specs/UIIntegration_spec.js#L793-L803
So maybe we could simply add some globally excluded site parameters, so they are displayed within the tests. Guess we could add that code somewhere around here: https://github.com/matomo-org/matomo/blob/8c3f495c64a5209f0110f49588c4840ff2cd07fe/tests/PHPUnit/Fixtures/UITestFixture.php#L107-L110
To ensure some "bad parameters" won't hurt, it might be good to use some xss parameters like:
```php
        \Piwik\Plugins\SitesManager\API::getInstance()->setGlobalExcludedQueryParameters(implode(',', [
            $this->xssTesting->forTwig('excludedParameter'),
            $this->xssTesting->forAngular('excludedParameter'),
        ]));
```

tests/PHPUnit/Fixtures/UITestFixture.php
```diff
@@ -322,7 +322,11 @@ public static function getLocalTestSiteUrl()
     private function addNewSitesForSiteSelector()
     {
         for ($i = 0; $i != 8; ++$i) {
-            self::createWebsite("2011-01-01 00:00:00", $ecommerce = 1, $siteName = "Site #$i", $siteUrl = "http://site$i.com");
+            self::createWebsite("2011-01-01 00:00:00", $ecommerce = 1, $siteName = "Site #$i",
+                $siteUrl = "http://site$i.com", 1, null, null,
+                null, null, 0,
+                implode(',', [$this->xssTesting->forTwig('excludedParameter'),
+                              $this->xssTesting->forAngular('excludedParameter'),'sid']));
```
sgiehl: Not sure if it would be needed to add them for each site. Having the globally would be enough for the tests. But besides of that, the new XSS tests need to be added to https://github.com/matomo-org/matomo/blob/3565da78bff718f201247fc2f692c0be29220fde/tests/PHPUnit/Framework/XssTesting.php#L104-L108

bx80: The specification for this issue changed to only use site excluded parameters and ignore global excluded parameters, that's why I've added them to each test site. I've added the missing XSS expected entries, thanks.

sgiehl:
Tried updating the remaining UI tests, but the artifacts upload is currently not working, so I'm not able to check if the failing tests are related. Once the related UI tests are fixed guess this should be ready to merge

PR #18032, opened 2021-09-20
Optimise tables separately

### LINKED ISSUE #14719: Run OPTIMIZE TABLE on each table in a separate SQL query to work better with replication
Tonight One of our server started to lag due to matomo starting a 
OPTIMIZE TABLE archive_numeric_* and archive_blob_*
Of all the table (in a single query), that took on main DB almost 90min (and would have taken a similar amount on the slave).

A quick solution is to iterate over and perform 1 query per table, It will still hang on the bigger one, but is a much more reasonable delay.

Another solution can be to also set a different 
SET SESSION gtid_domain_id=1
https://mariadb.com/kb/en/library/parallel-replication/
https://dev.mysql.com/doc/refman/5.7/en/replication-gtids.html

So those maintenance job will run on a side thread, and real time ingestion will keep running.

Code wise is probably just a few lines.

This should be an option that will be enabled (if needed).
In case is a big installation (the one that suffer of that problem) the DBA / Admin will probably know what to do and what we are talking about (we just have to leave a few link to documentation).

If you are interested In this second proposal I can try to write some code example.
But again code wise is just a matter of 

SET SESSION gtid_domain_id=XYZ;

bx80:
### Description:

Fixes #14719

Rather than optimizing all tables in a single query, the optimize tables task will run each table optimization separately to allow breathing space for replicating servers.

### Review

### DIFF
--- a/core/Db.php
+++ b/core/Db.php
@@ -420,7 +420,7 @@ class Db
      * @param string|array $tables The name of the table to optimize or an array of tables to optimize.
      *                             Table names must be prefixed (see {@link Piwik\Common::prefixTable()}).
      * @param bool $force If true, the `OPTIMIZE TABLE` query will be run even if InnoDB tables are being used.
-     * @return bool|\Zend_Db_Statement
+     * @return bool
      */
     public static function optimizeTables($tables, $force = false)
     {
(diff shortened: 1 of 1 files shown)


core/Db.php
```diff
@@ -420,7 +420,7 @@ public static function deleteAllRows($table, $where, $orderBy, $maxRowsPerQuery
      * @param string|array $tables The name of the table to optimize or an array of tables to optimize.
      *                             Table names must be prefixed (see {@link Piwik\Common::prefixTable()}).
      * @param bool $force If true, the `OPTIMIZE TABLE` query will be run even if InnoDB tables are being used.
-     * @return \Zend_Db_Statement
+     * @return bool|\Zend_Db_Statement
```
sgiehl: Actually this method now always returns a bool
```suggestion
     * @return bool
```

PR #18040, opened 2021-09-21
Fix escaping of html attribute

sgiehl:
### Review

### DIFF
--- a/plugins/Diagnostics/Diagnostic/ForceSSLCheck.php
+++ b/plugins/Diagnostics/Diagnostic/ForceSSLCheck.php
@@ -11,6 +11,7 @@ use Piwik\Config;
 use Piwik\ProxyHttp;
 use Piwik\Translation\Translator;
 use Piwik\Url;
+use Piwik\View;
 
 /**
  * Check that Matomo is configured to force SSL.
@@ -38,10 +39,9 @@ class ForceSSLCheck implements Diagnostic
                 return [];
             }
 
-            $message = $this->translator->translate('General_UseSSLInstall', [
-                '<a href="' . htmlspecialchars('https://' . Url::getCurrentHost() . Url::getCurrentScriptName(false) . Url::getCurrentQueryString()) .'">',
-                '</a>'
-            ]);
+            $view = new View('@Diagnostics/force_ssl_link');
+            $view->link = 'https://' . Url::getCurrentHost() . Url::getCurrentScriptName(false) . Url::getCurrentQueryString();
+            $message = $view->render();
             return [DiagnosticResult::singleResult($label, DiagnosticResult::STATUS_WARNING, $message)];
         }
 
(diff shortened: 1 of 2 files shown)


plugins/Diagnostics/Diagnostic/ForceSSLCheck.php
```diff
@@ -39,7 +39,7 @@ public function execute()
             }
 
             $message = $this->translator->translate('General_UseSSLInstall', [
-                '<a href="https://'. Url::getCurrentHost() . Url::getCurrentScriptName(false) . Url::getCurrentQueryString() .'">',
+                '<a href="' . htmlspecialchars('https://' . Url::getCurrentHost() . Url::getCurrentScriptName(false) . Url::getCurrentQueryString()) .'">',
```
tsteur: @sgiehl does it make sense maybe to use twig here to render this as it might do a few more things? Looking at twig_escape_filter there are quite a few other things happening. Or maybe it also would work to do like below? (not tested though just random thought)

```php
http_build_query(Url::getArrayFromCurrentQueryString())
```

sgiehl: we can use this instead, if you would prefer
```suggestion
                '<a href="' . twig_escape_filter(\Piwik\Container\StaticContainer::get(\Piwik\Twig::class)->getTwigEnvironment(), 'https://' . Url::getCurrentHost() . Url::getCurrentScriptName(false) . Url::getCurrentQueryString(), 'html_attr') .'">',
```

tsteur: does twig not work there @sgiehl ? Using twig potentially our own filter etc would run as well.

sgiehl: It's the result of a Diagnostic check. Guess it's inserted with twig somewhere in a later step, but don't think we can use twig syntax here so it's evaluated later 🤔

tsteur: @sgiehl just tested below and it worked

```php
$view = new View('@Diagnostics/force_ssl_link');
            $link = $view->render();
            $message = $this->translator->translate('General_UseSSLInstall', [
                $link,
                '</a>'
            ]);
```

Might be easiest to use that or otherwise we need to use `piwik_escape_filter`. Just meaning as this code path my not be reached in tests if we refactoring things so it may be bit more "safe" to use just a twig template. In the end I suppose both have pros/cons. No big preference. We would at least need to use `piwik_escape_filter` though to fix lbrace

sgiehl: Ah sure. That way it should work. Hadn't thought about simply using a template....
Updated the code.

PR #18046, opened 2021-09-22
For login allow list add support for IPv6 

tsteur:
### Description:

Not sure why we didn't add this directly. see also faq https://matomo.org/faq/how-to/faq_25543/

### Review

### DIFF
--- a/config/global.php
+++ b/config/global.php
@@ -162,12 +162,15 @@ return array(
         $ipsResolved = array();
 
         foreach ($ips as $ip) {
+            $ip = trim($ip);
             if (filter_var($ip, FILTER_VALIDATE_IP)) {
                 $ipsResolved[] = $ip;
             } else {
                 $ipFromHost = @gethostbyname($ip);
-                if (!empty($ipFromHost)
-                    && filter_var($ipFromHost, FILTER_VALIDATE_IP)) {
+                if (!empty($ipFromHost)) {
+                    // we don't check using filter_var if it's an IP as "gethostbyname" will return the $ip if it's not a hostname
+                    // and we then assume it is an IP range. Otherwise IP ranges would not be added. Ideally would above check if it is an
+                    // IP range before trying to get host by name.
                     $ipsResolved[] = $ipFromHost;
                 } 
                 
--- a/config/global.php
+++ b/config/global.php
@@ -169,7 +169,9 @@ return array(
                 if (!empty($ipFromHost)
                     && filter_var($ipFromHost, FILTER_VALIDATE_IP)) {
                     $ipsResolved[] = $ipFromHost;
-                } elseif (function_exists('dns_get_record')) {
+                } 
+                
+                if (function_exists('dns_get_record')) {
                     $entry = @dns_get_record($ip, DNS_AAAA);
                     if (!empty($entry['0']['ipv6'])
                         && filter_var($entry['0']['ipv6'], FILTER_VALIDATE_IP)) {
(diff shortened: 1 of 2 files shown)


config/global.php
```diff
@@ -166,8 +166,15 @@
                 $ipsResolved[] = $ip;
             } else {
                 $ipFromHost = @gethostbyname($ip);
-                if (!empty($ipFromHost)) {
+                if (!empty($ipFromHost)
```
tsteur: the extra filter var check is needed as ipFromHost will be the hostname if it couldn't resolve it.

image

plugins/CoreHome/tests/Integration/LoginAllowlistTest.php
```diff
@@ -135,6 +135,12 @@ public function test_getAllowlistedLoginIps_shouldResolveIp()
         $this->assertSame(['192.168.33.1', '185.31.40.177', '127.0.0.1'], $this->allowlist->getAllowlistedLoginIps());
     }
 
+    public function test_getAllowlistedLoginIps_shouldResolveIpv6()
+    {
+        $this->setGeneralConfig('login_allowlist_ip', ['192.168.33.1', 'integration-test.matomo.org', '127.0.0.1']);
```
tsteur, line 140: added this DNS entry specifically for this test

tsteur: @Findus23 @sgiehl would it be an issue that `https://integration-test.matomo.org/` resolves `aaaa` to `::1`?

config/global.php
```diff
@@ -166,8 +166,15 @@
...
+                    && filter_var($ipFromHost, FILTER_VALIDATE_IP)) {
                     $ipsResolved[] = $ipFromHost;
+                } elseif (function_exists('dns_get_record')) {
```
sgiehl: wouldn't it generally make sense to do always both. A lookup for IPv4 and IPv6. I guess a hostname could actually resolve both, right?

plugins/CoreHome/tests/Integration/LoginAllowlistTest.php
```diff
@@ -132,7 +132,13 @@ public function test_getAllowlistedLoginIps_shouldReturnIpsAndTrimIfNeeded()
...
     {
         $this->setGeneralConfig('login_allowlist_ip', ['192.168.33.1', 'matomo.org', '127.0.0.1']);
-        $this->assertSame(['192.168.33.1', '185.31.40.177', '127.0.0.1'], $this->allowlist->getAllowlistedLoginIps());
+        $this->assertSame(['192.168.33.1', '185.31.40.177', '2a00:b6e0:1:200:177::1', '127.0.0.1'], $this->allowlist->getAllowlistedLoginIps());
+    }
+
+    public function test_getAllowlistedLoginIps_shouldResolveIpv6Only()
+    {
+        $this->setGeneralConfig('login_allowlist_ip', ['192.168.33.1', 'integration-test.matomo.org', '127.0.0.1']);
+        $this->assertSame(['192.168.33.1', 'integration-test.matomo.org', '::1', '127.0.0.1'], $this->allowlist->getAllowlistedLoginIps());
```
tsteur, line 141: it may not be expected that it returns the integration test hostname here, but it was already doing it before so it shouldn't be a problem. Only fixing the IPv6 part for now

plugins/CoreHome/tests/Integration/LoginAllowlistTest.php
sgiehl: can't think of any reason why that could be an issue

sgiehl (approved):
The test failures seem random, so guess this should be good to merge.

PR #18050, opened 2021-09-23
update admin level, user manage text

### LINKED ISSUE #17996: Manage Users: When user is admin, mention other users might exist
Only when a user has admin permission but is not a super user then we want to show few more sentences in the top of the users management:

>  Please note that you are logged in as an administrator. This means you only see and manage users that have access to the specific websites where you have "admin" permissions. Only a user with super user access can see and manage all users across all websites.

See screenshot image

We had issues with our premium features where someone thought that there are only 2 users on Matomo when in fact there were more. Because the user had only admin permission, all other users were not visible to that user. This caused licensing issues.

We can't mention how many users exist overall as this could cause issues for various Matomo installations and expose information that is not supposed to be exposed. However, we can say there may or may not be other users on this Matomo.

cc @mattab

peterhashair:
### Description:

update admin level, user manager hints text

### Review


plugins/UsersManager/lang/en.json
```diff
@@ -123,6 +123,7 @@
         "AtLeastView": "At least view",
         "ManageUsers": "Manage Users",
         "ManageUsersDesc": "Create new users or update the existing users. You can then set their permissions here too.",
+        "ManageUsersAdminDesc": "Create new users or update the existing users. You can then set their permissions here too. This means you only see and manage users that have access to the specific websites where you have \"admin\" permissions. Only a user with super user access can see and manage all users across all websites.",
```
sgiehl: Seems there is a part missing:
```suggestion
        "ManageUsersAdminDesc": "Create new users or update the existing users. You can then set their permissions here too. Please note that you are logged in as an administrator. This means you only see and manage users that have access to the specific websites where you have \"admin\" permissions. Only a user with super user access can see and manage all users across all websites.",
```

sgiehl:
Guess the UI screenshots will need another update. Besides that it should be good to merge I guess.

peterhashair:
@sgiehl updated it, but it seems like there is a test that failed is not relevant to that one request

sgiehl (approved):

PR #18051, opened 2021-09-23
More verbose error message when login nonce check fails

### LINKED ISSUE #16607: More verbose error message when login nonce check fails
At the moment Matomo only shows the following error:
> Fehler : Sicherheitschecks fehlgeschlagen. Bitte laden Sie das Formular erneut und prüfen Sie, ob Ihr Browser Cookies zulässt. Wenn Sie einen Proxy Server verwenden, müssen Sie Matomo so einrichten, dass es Proxy Header akzeptiert.

https://github.com/matomo-org/matomo/blob/da456513866ea0e276c51b046af5139244968a23/plugins/Login/lang/en.json#L8

But when the user has Cookies enabled (which is pretty likely) and is sure that they don't use a reverse proxy (shouldn't it say reverse proxy instead of proxy in the message?), there is no way for them to troubleshoot this issue further, and they will most likely just give up on using Matomo.

Maybe all checks that could fail in `verifyNonce()` and `isLocalUrl()` should be logged or even help display a more helpful error message.

https://github.com/matomo-org/matomo/blob/679e73f1236969db0c2d767655cb84456a727d24/core/Nonce.php#L70
https://github.com/matomo-org/matomo/blob/06d43857c48ada2fa7f1ad18a8309e8826c0e413/core/Url.php#L547

peterhashair:
### Description:

Fixes: #16607 
split nonce verifies into 3 parts. By extending the verifyNonce function pass 4th params with true or false, the default won't show details.  I am not sure about the security policy around this one, I know some errors required generic error messages to prevent getting too many details, but for this case, I don't think will applied. 

-   Token mismatch (InvalidNonceToken)
-   Referrer (InvalidNonceReferrer)
-   Origin (InvalidNonceOrigin)

- deprecated getMessageExceptionNoAccess Login/controller
- deprecated InvalidNonceOrHeadersOrReferrer in lang
- move getMessageExceptionNoAccessWhenInsecureConnectionMayBeUsed from Login/controller to core/Nonce
- fix missing 2 variables in translation.

### Review


Findus23:
Honestly I meant in #16607 that we should log or output more precisely which one of the many checks in https://github.com/matomo-org/matomo/blob/679e73f1236969db0c2d767655cb84456a727d24/core/Nonce.php#L70 and https://github.com/matomo-org/matomo/blob/06d43857c48ada2fa7f1ad18a8309e8826c0e413/core/Url.php#L547 fails instead of leaving the user guessing and debugging Matomo code. 

Adding the sentence by @Starker3 is also fine, but doesn't affect the core issue I meant.

@peterhashair unrelated, but it would be great if you could link issues in the PR description, so one can directly jump back to them instead of searching for the linked issue in the sidebar. And if you use something like `fixes #1234`, you don't need to manually link them either.

peterhashair:
@Findus23 Sorry about the fixes issue number, add them back. If I understand correctly, we want to split the verifyNonce into details with their own error messages, instead of combining them into one error log?

core/Nonce.php
```diff
@@ -65,26 +65,27 @@ public static function getNonce($id, $ttl = 600)
      *
      * @param string $id The nonce's unique ID. See {@link getNonce()}.
      * @param string $cnonce Nonce sent from client.
-     * @param string $expectedReferrerHost The expected referrer host for the HTTP referrer URL.
+     * @param null $expectedReferrerHost The expected referrer host for the HTTP referrer URL.
+     * @param bool $withErrorMessage The expected true of false to return with error message
      * @return bool `true` if valid; `false` otherwise.
      */
-    public static function verifyNonce($id, $cnonce, $expectedReferrerHost = null)
+    public static function verifyNonce($id, $cnonce, $expectedReferrerHost = null, bool $withErrorMessage = false)
```
tsteur: @peterhashair could we maybe rather have two different methods `verifyNonceWithErrorMessage` (there's likely better wording for it) and `verifyNonce`? That will reduce the complexity of the method a bit and often boolean parameters indicate that it should be maybe two methods. You can still reuse the code like 


```
public static function verifyNonce($id, $cnonce, $expectedReferrerHost = null){
    $error = self::verifyNonceWith...($id, $cnonce, $expectedReferrerHost);
    return empty($error);
}
```

core/Nonce.php
```diff
@@ -65,38 +65,59 @@ public static function getNonce($id, $ttl = 600)
...
         if (empty($cnonce) || $cnonce !== $nonce) {
-            return false;
+            return Piwik::translate('Login_InvalidNonceToken');
         }
 
         // validate referrer
         $referrer = Url::getReferrer();
         if (empty($expectedReferrerHost) && !empty($referrer) && !Url::isLocalUrl($referrer)) {
-            return false;
+            return Piwik::translate('Login_InvalidNonceReferrer');
```
tsteur: @peterhashair is there any chance we can slightly tweak the two same `InvalidNonceReferrer` messages to have them bit different? when users report us that they have a problem we'll then better know what the problem is.

peterhashair:
@tsteur split the error message,  but I don't think $expectedReferrerHost is used anywhere.  Should we set a variable in config.ini.php that reflected that value?

plugins/Login/lang/en.json
```diff
@@ -5,7 +5,11 @@
...
         "ExceptionInvalidSuperUserAccessAuthenticationMethod": "A user with Super User access cannot be authenticated using the '%s' mechanism.",
         "ExceptionPasswordMD5HashExpected": "The password parameter is expected to be a MD5 hash of the password.",
-        "InvalidNonceOrHeadersOrReferrer": "Form security failed. Please reload the form and check that your cookies are enabled. If you use a proxy server, you must %1$s configure Matomo to accept the proxy header%2$s that forwards the Host header. Also, check that your Referrer header is sent correctly.",
+        "InvalidNonceToken": "Form security failed, token miss match. Please reload the form and check that your cookies are enabled",
+        "InvalidNonceReferrer": "Form security failed, invalid referrer header, If you use a proxy server, you must %1$s configure Matomo to accept the proxy header%2$s that forwards the Host header. Also, check that your Referrer header is sent correctly and  If you previously connected using https, please ensure you are connecting over a secure (SSL/TLS) connection and try again.",
```
tsteur: There are placeholders used but no value is actually passed to them in the translate method. When this error happens it looks like this:

image
There should be also a dot (`.`) after `invalid referrer header` instead of a comma (`,`).

plugins/Login/lang/en.json
```diff
@@ -5,7 +5,11 @@
...
+        "InvalidNonceUnexpectedReferrer" : "Form security failed, the referrer header is different from the expected referrer header, check that your Referrer header is sent correctly.",
```
tsteur, line 10: I wonder how someone would check if the referrer header is sent correctly. But let's ignore this for now as I suppose we can ignore it a bit for now as `$expectedReferrerHost` isn't used there so far. Nothing to do for now I reckon. Seems to be indeed not used in the code base.

plugins/Login/lang/en.json
```diff
@@ -5,7 +5,11 @@
...
+        "InvalidNonceOrigin": "Form security failed, invalid origin, If you previously connected using https, please ensure you are connecting over a secure (SSL/TLS) connection and try again",
```
tsteur: Can we use a dot after `invalid origin` instead of a comma?

tsteur (changes requested):
@peterhashair left a comment as there's an issue with a translation.

Also the method `getMessageExceptionNoAccess` method in the login controller seems to be no longer used and we could remove it now?

The translation key `Login_InvalidNonceOrHeadersOrReferrer` seems to be still used in the `TwoFactorAuth/Controller`. Maybe in that controller we could similarly use the new method in the `loginTwoFactorAuth` to get the reason why the nonce check fails? Then we could remove the translation key `Login_InvalidNonceOrHeadersOrReferrer`.
image

peterhashair:
- deprecated getMessageExceptionNoAccess Login/controller
- deprecated  InvalidNonceOrHeadersOrReferrer in lang
- move getMessageExceptionNoAccessWhenInsecureConnectionMayBeUsed from Login/controller to core/Nonce
- fix missing 2 variables in translation.

tsteur:
@peterhashair just tested it when `Url::isSecureConnectionAssumedByPiwikButNotForcedYet== true`

It now shows this 
image

it doesn't look quite correct? I think previously it was appended to any error message.

peterhashair:
@tsteur sorry, I thought that was a generic error. Changed it back, now should only appear when the error appears.

plugins/Login/lang/en.json
```diff
@@ -5,7 +5,10 @@
         "ContactAdmin": "Possible reason: your host may have disabled the mail() function. <br \/>Please contact your Matomo administrator.",
         "ExceptionInvalidSuperUserAccessAuthenticationMethod": "A user with Super User access cannot be authenticated using the '%s' mechanism.",
         "ExceptionPasswordMD5HashExpected": "The password parameter is expected to be a MD5 hash of the password.",
-        "InvalidNonceOrHeadersOrReferrer": "Form security failed. Please reload the form and check that your cookies are enabled. If you use a proxy server, you must %1$s configure Matomo to accept the proxy header%2$s that forwards the Host header. Also, check that your Referrer header is sent correctly.",
+        "InvalidNonceToken": "Form security failed, token miss match. Please reload the form and check that your cookies are enabled",
```
tsteur: here we need to add a dot at the end
image

plugins/Login/lang/en.json
```diff
@@ -5,7 +5,10 @@
...
+        "InvalidNonceReferrer": "Form security failed, invalid referrer header. If you use a proxy server, you must %1$s configure Matomo to accept the proxy header %2$s that forwards the Host header. Also, check that your Referrer header is sent correctly and  If you previously connected using https, please ensure you are connecting over a secure (SSL/TLS) connection and try again.",
+        "InvalidNonceUnexpectedReferrer" : "Form security failed, the referrer header is different from the expected referrer header, check that your Referrer header is sent correctly.",
+        "InvalidNonceOrigin": "Form security failed, invalid origin. If you previously connected using https, please ensure you are connecting over a secure (SSL/TLS) connection and try again",
```
tsteur: same here we need to add a dot here

core/Nonce.php
```diff
@@ -65,38 +65,80 @@ public static function getNonce($id, $ttl = 600)
      *
      * @param string $id The nonce's unique ID. See {@link getNonce()}.
      * @param string $cnonce Nonce sent from client.
-     * @param string $expectedReferrerHost The expected referrer host for the HTTP referrer URL.
+     * @param null $expectedReferrerHost The expected referrer host for the HTTP referrer URL.
```
tsteur: not too important, could change this to `null|string`

tsteur (changes requested):
@peterhashair last 2 comments to add a `dot` at the end of sentences. Then it should be good to merge.

PR #18057, opened 2021-09-23
Remove travis secure variables

sgiehl:


tsteur:
Merged for now and we can add back if needed

Issue #18019, opened 2021-09-16
Make the opt out work when a PHP 32 bit is used


samuel-lobby:
Hello all,

The iframe I added for the opt out works fine, at least as far as tracking is concerned. If I click on the checkbox it is unchecked and the text of the iframe changes after a reload of the page. If I reload the page again, the "old" text is displayed plus the checkbox is checked again.
However, the cookie for the opt out is set correctly and also works correctly.
But I would like the checkbox and the text to remain after an opt out. Do I have to set anything extra for this or is this a bug?

Thank's a lot and best regards,
Samuel

Findus23:
Just a wild guess:
Is your iFrame embedded into a different domain as the one Matomo uses and do you have a browser with third party cookies disabled (like Chrom(e|ium) by default?

samuel-lobby:
The Matomo is used on a subdomain of the main domain.
I first thought it was the cookie banner, but that doesn't seem to be the problem. I set these two necessary cookies:
- MATOMO_SESSID
- mtm_consent_removed

In Matomo I deactivated tracking with cookies maybe that's also an important information.

tsteur:
Is there any chance you could post a URL to the page where we can reproduce this? It be otherwise hard to reproduce or guess what the problem might be @samuel-lobby

samuel-lobby:
@tsteur sure, here is the link to the privacy policy: https://www.hs-doepfer.de/datenschutz
Here I've implemented the opt out iframe. I hope this can be clarified as soon as possible.

tsteur:
Tested eg https://analytics.hs-doepfer.de/index.php?module=CoreAdminHome&action=optOut&language=de&setCookieInNewWindow=1&nonce=NONCE&time=1632171100302


It was trying to set an ignore cookie 
![image](https://user-images.githubusercontent.com/273120/134074289-484cf56d-ac39-4e34-b9cd-7aad8c0e9146.png)

but it wasn't set afterwards somehow. Nothing related in the logs, no issues.

Not sure what the problem might be.

@samuel-lobby I think you might not be using the latest version of Matomo? Is there any chance you could upgrade to see if that fixes it maybe?

samuel-lobby:
@tsteur I use Matomo 4.4.1. I made the update to this version after I recognized that this opt out iframe isn't working as it should. So maybe the cookie matomo_ignore is blocked from the cookie consent mananger?

Findus23:
[...]

PR #18060, opened 2021-09-24
add php binary format check

### LINKED ISSUE #18058: Add system check for 32 bit PHP / OS
refs https://github.com/matomo-org/matomo/issues/18019

Let's add a system check if a 32 bit PHP is used and mention we recommend upgrading to a 64 bit to prevent various issues that may come up in the future.

Not sure how many possible integer overflows there are so far in Matomo. We need to do a quick check as part of this issue to know if there might be already issues with 32 bit in PHP by looking through the code (like finding places where we do `time() + ...`

Does anyone know how common 32 bit installs are these days?

peterhashair:
### Description:
Fixes: #18058 
add os format check on installation, recommended 64 bit

### Review


plugins/Installation/lang/en.json
```diff
@@ -137,6 +137,9 @@
         "SystemCheckPhpSetting": "To prevent some critical issue, you must set the following in your php.ini file: %s",
         "SystemCheckUpdateHttps": "Update over HTTPS",
         "SystemCheckUpdateHttpsNotSupported": "Matomo cannot use HTTPS to update, it will fall back to the insecure HTTP update. Check that CURL or allow_url_fopen is supported and that the openssl PHP extension is installed: https:\/\/matomo.org\/faq\/troubleshooting\/faq_177\/.",
+        "SystemCheckProcessorFormat" : "OS Format",
+        "SystemCheckProcessorFormat64": "64-bit",
+        "SystemCheckProcessorFormat32": "32-bit.<p> We highly recommend upgrading to a 64-bit to prevent various issues that may cause inconvenient in the future.</p>",
```
sgiehl: I'm no native English speaker, so bare with me if I'm wrong, but shouldn't that be `inconvenience` instead of `inconvenient`? 🤔

plugins/Diagnostics/Diagnostic/SystemFormatCheck.php
```diff
@@ -0,0 +1,47 @@
...
+     */
+
+    public function __construct(Translator $translator)
+    {
+        $this->translator = $translator;
+    }
+
+    public function execute()
+    {
+        $label = $this->translator->translate('Installation_SystemCheckProcessorFormat');
```
sgiehl: Actually I don't think that `PHP_INT_SIZE` allows to draw conclusions about the system or the processor.
At least on windows it should be possible to install a x86 PHP version on a Windows version that actually is x64. Not sure if it would be possible to compile a x86 version on a linux system that is x64, actually never tried that...
Also I had something in mind that the BCMatch extension might also allow higher numbers even on a x86 system, but nt fully sure about that right now.
So maybe we should only mention that the PHP version is doesn't support 64bit numbers or something like that? 🤔

plugins/Installation/lang/en.json
Findus23: Maybe also write `to a 64-bit PHP binary`

plugins/Installation/lang/en.json
```diff
@@ -137,6 +137,9 @@
         "SystemCheckPhpSetting": "To prevent some critical issue, you must set the following in your php.ini file: %s",
         "SystemCheckUpdateHttps": "Update over HTTPS",
         "SystemCheckUpdateHttpsNotSupported": "Matomo cannot use HTTPS to update, it will fall back to the insecure HTTP update. Check that CURL or allow_url_fopen is supported and that the openssl PHP extension is installed: https:\/\/matomo.org\/faq\/troubleshooting\/faq_177\/.",
+        "SystemCheckProcessorFormat" : "OS Format",
```
Findus23: I'm not sure if OS Format is a well known term (a search only shows results on how to format disks).
As I don't know any better term myself, we could maybe instead use "PHP Binary" or something like this.

peterhashair:
@sgiehl  Yeah, sorry for my english. Iit's really hard to figure out ,if a webserver x86 runining on WOW64bit by php, or it could be 64 bit of website server but 32 bit version of php. I guess the OS format check is not a propery word. maybe like @Findus23 suggest **64-bit PHP Binary** is a better term.  What's your suggest?

sgiehl:
`64-bit PHP Binary` should be fine

plugins/Installation/lang/en.json
```diff
@@ -137,6 +137,8 @@
         "SystemCheckPhpSetting": "To prevent some critical issue, you must set the following in your php.ini file: %s",
         "SystemCheckUpdateHttps": "Update over HTTPS",
         "SystemCheckUpdateHttpsNotSupported": "Matomo cannot use HTTPS to update, it will fall back to the insecure HTTP update. Check that CURL or allow_url_fopen is supported and that the openssl PHP extension is installed: https:\/\/matomo.org\/faq\/troubleshooting\/faq_177\/.",
+        "SystemCheckProcessorFormat" : "64-bit PHP Binary",
+        "SystemCheckProcessorFormat32": "32-bit<p> We highly recommend upgrading to a 64-bit to prevent various issues that may cause inconvenience in the future.</p>",
```
justinvelluppillai: perhaps this line should also say "..We highly recommend upgrading to a 64-bit PHP binary to prevent..." to make more sense

plugins/Installation/lang/en.json
```diff
@@ -137,6 +137,8 @@
         "SystemCheckPhpSetting": "To prevent some critical issue, you must set the following in your php.ini file: %s",
         "SystemCheckUpdateHttps": "Update over HTTPS",
         "SystemCheckUpdateHttpsNotSupported": "Matomo cannot use HTTPS to update, it will fall back to the insecure HTTP update. Check that CURL or allow_url_fopen is supported and that the openssl PHP extension is installed: https:\/\/matomo.org\/faq\/troubleshooting\/faq_177\/.",
+        "SystemCheckProcessorFormat" : "64-bit PHP Binary",
+        "SystemCheckProcessorFormat32": "32-bit<p> We highly recommend upgrading to a 64-bit PHP binary to prevent various issues that may cause inconvenience in the future.</p>",
```
sgiehl: The translation key should be renamed, so they either describe what they contain or where they are used. Those keys are visible to the translators and should (if possible) kind of help them to understand the context.

plugins/Diagnostics/Diagnostic/SystemFormatCheck.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Piwik\Plugins\Diagnostics\Diagnostic;
+
+use Piwik\Config;
+use Piwik\SettingsServer;
+use Piwik\Translation\Translator;
+
+/**
+ * Check that the memory limit is enough.
+ */
+class SystemFormatCheck implements Diagnostic
```
sgiehl: Might be good to rename this class and file to something more meaningful. What about `PHPBinaryCheck`? Also the comment above is incorrect.

sgiehl (changes requested):

sgiehl (approved):
Failing tests seem to be unrelated, so should be good to merge now.

plugins/Installation/lang/en.json
```diff
@@ -137,6 +137,8 @@
         "SystemCheckPhpSetting": "To prevent some critical issue, you must set the following in your php.ini file: %s",
         "SystemCheckUpdateHttps": "Update over HTTPS",
         "SystemCheckUpdateHttpsNotSupported": "Matomo cannot use HTTPS to update, it will fall back to the insecure HTTP update. Check that CURL or allow_url_fopen is supported and that the openssl PHP extension is installed: https:\/\/matomo.org\/faq\/troubleshooting\/faq_177\/.",
+        "PhpBinaryCheck" : "64-bit PHP Binary",
+        "PhpBinaryCheckHelp": "32-bit <p> We highly recommend upgrading to a 64-bit PHP binary to prevent various issues that may cause inconvenience in the future.</p>",
```
tsteur: FYI I've done a 3 minute check to see if there are already any issues re integer overflows already and couldn't find any except for the ignore cookie (see issue https://github.com/matomo-org/matomo/issues/18019). Searched for eg `time() +` and `86400` and `365` in the code base.  It actually doesn't look like there will be any issues for the next 5-10 years but this can obviously change any time. 

Maybe could say 

```
We recommend upgrading to a 64-bit PHP binary to prevent bugs in the future.
```

I don't think there's a big urgency just yet which is why I would maybe remove the word "highly" recommended. and instead of "inconvenience" would say maybe something like "bugs" to make it more clear.
Or wondering if we maybe even need to show this yet at all assuming there won't be an issue for the next few years maybe? Meaning we could maybe better say `We recommend upgrading to a 64-bit PHP binary at the latest by January 2026 to prevent bugs in the future.`

When we say "in the future" then it might not be clear for the user when there will be a problem and they might panic and stop using Matomo if it's not easy to switch to a 64 bit right now.

Unless there are maybe some other integer overflows in there already? I'm assuming they might be hard to find?

peterhashair:
@tsteur can I merge that one?

tsteur:
Yes @peterhashair 👍

PR #18062, opened 2021-09-26
Transition data exporting

bx80:
### Description:

Fixes #6032

Visualisations and the standard report exporter expect a single datatable structure, so instead of wrapping the transitions in a visualisation, this change adds a new transition exporter directive which provides an export action on both the main transitions map view and on transitions shown by row actions. This export action then calls the API to return the data in the chosen export format.

### Review


plugins/Transitions/templates/renderPopover.twig
```diff
@@ -48,11 +48,28 @@
...
+</div>
+
 <script type="text/javascript">
     var Piwik_Transitions_Translations = {
         {% for internalKey, translation in translations %}
             "{{ internalKey }}": "{{ translation }}",
         {% endfor %}
         "": ""
     };
+    let tea = $('#transition-export-actions');
```
sgiehl: Guess we maybe should use `var` here, to ensure it would work in all browsers. Or can we maybe use `let` meanwhile, as support for IE will be dropped? ping @tsteur 

```suggestion
    var tea = $('#transition-export-actions');
```

plugins/Transitions/angularjs/transitionexporter/transitionexporter.directive.js
```diff
@@ -0,0 +1,78 @@
...
+
+                element.on('click', function () {
+
+                    scope.exportFormat = 'JSON';
+                    scope.exportFormatOptions = [
+                        {key: 'JSON', value: 'JSON'},
+                        {key: 'XML', value: 'XML'}
+                    ];
+
+                    scope.reportTitle = $rootScope.transitionExportParams['actionName'] + ' ' + _pk_translate('Transitions_Transitions');
```
sgiehl, line 64: This throws an error when you click the export icon, but the report has no data available. Maybe we should hide the export icon in this case?

plugins/Transitions/Transitions.php
```diff
@@ -40,6 +40,7 @@ public function getJsFiles(&$jsFiles)
     {
         $jsFiles[] = 'plugins/Transitions/javascripts/transitions.js';
         $jsFiles[] = 'plugins/Transitions/angularjs/transitionswitcher/transitionswitcher.controller.js';
+        $jsFiles[] = 'plugins/Transitions/angularjs/transitionexporter/transitionexporter.directive.js';
```
sgiehl, line 44: Guess you missed to add the new `less` file in the method above. Therefore, the layout currently looks broken.

sgiehl (changes requested):
Generally seems to work.
Btw. did you try reusing the report export we already have?
See https://github.com/matomo-org/matomo/blob/b66635e4e16f5af26171211b6196f1dff319e1c5/plugins/CoreHome/templates/_dataTableActions.twig#L57-L62
Guess it would need various adjustments, as it currently depends on a datatable, which we don't have for transitions. Also there are some configuration that would need to be hidden, as they aren't relevant here. But it basically does the same and would also allow to see the export url.

plugins/Transitions/templates/renderPopover.twig
bx80: Force of habit to use let, probably better to use var for consistency.

plugins/Transitions/angularjs/transitionexporter/transitionexporter.directive.js
bx80: I've added a null check for that error and the export icon should now be hidden if the transition has no data available.

bx80: Good pointer about the report export directive, I did spent some time looking at reusing it before taking this approach, but I couldn't find a clean way to make it work with both the transition switcher and the row action transitions without adding a lot of special case adjustments to it's code. So it seemed more effective to add a simplified directive just for transitions. There might also be other transition specific export formats later on such as images or some sort of flat csv format, which might be better kept within the scope of the transitions plugin. Hopefully that makes sense.

plugins/Transitions/angularjs/transitionexporter/transitionexporter.directive.js
```diff
@@ -0,0 +1,82 @@
...
+                        return;
+                    }
+
+                    scope.reportTitle = $rootScope.transitionExportParams['actionName'] + ' ' + _pk_translate('Transitions_Transitions');
+
+                    this.popover = Piwik_Popover.showLoading(_pk_translate('General_Export'), self.actionName, 200);
+
+                    var elem = $document.find('#transitionExport').eq(0);
+                    if (!elem.length) {
+                        elem = angular.element('<span ng-include="\'plugins/Transitions/angularjs/transitionexporter/transitionexporter.popover.html?cb=' + piwik.cacheBuster + '\'" id="transitionExport"></span>');
```
tsteur: I wonder if we should `encodeURIComponent` the `piwik.cacheBuster` here just in case a script can somehow manipulate the cache buster to avoid injections... or even better could use something like this:

```js
elem = angular.element('<span + '\'" id="transitionExport"></span>');
elem.setAttr(' ng-include',"\'plugins/Transitions/angularjs/transitionexporter/transitionexporter.popover.html?cb=' + encodeURIComponent(piwik.cacheBuster))
```

tsteur (changes requested):
Nice work @bx80  worked nicely. Just left one comment to look at just to be safe re security. Otherwise looks good to merge after addressing that comment 👍 🎉

tsteur (approved):
Looks good to merge @bx80  I would merge directly but tests are still running. Be good to double check tests are still passing (ignoring any random/unrelated test failures) and then merge 🎉

PR #18064, opened 2021-09-26
add mod security check on installation

### LINKED ISSUE #3371: New system check to warn that Piwik is not compatible with mod_security
Reported in: #2997, some work was done in the early days in #1460
- Hostgator users need to contact their host to disable mod_security.  Hundreds of piwik users have had to [contact their hosts](http://forum.piwik.org/read.php?2,44561,page=1#msg-44871) to disable mod_security.
- There are several known issues with Piwik and mod_security! 
  - [this thread](http://sourceforge.net/p/mod-security/mailman/message/32173108/) and [here](https://lists.owasp.org/pipermail/owasp-modsecurity-core-rule-set/2011-September/000864.html)

```

I have found that almost all rules in modsecurity_crs_41_sql_injection_attacks.conf need 

!REQUEST_COOKIES:/^_pk_ref.*/|!REQUEST_COOKIES:/^__utmz$/|!ARGS:gclid  
for google adwords, google analytics  and piwik to work ok with mod_security.
```
- [this one](http://forum.piwik.org/read.php?2,114386,page=1#msg-116260)

```
Pattern match "\\%(?![0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" at ARGS:pattern. [file "/usr/local/apache/conf/modsec2.user.conf"] [line "20"] [id "1234123440"] [msg "URL Encoding Abuse Attack Attempt"] [severity "WARNING"]
  U4n-qUIt@YIADWbyzkUAAAB1 86.112.15.155 35095 66.45.249.132 80
--74545369-B-- 
```

As a proposed solution to inform users of potential issues early:
- Write a system check entry to check for mod_security and issue Warning if it is detected.
  - Suggest to user that it is OK to disable mod_security for Piwik app. 
- Maybe in this system check message we could also link [to FAQ](http://piwik.org/faq/troubleshooting/faq_100/) and this FAQ could list the rules to disable in the mod_security config. if some users reading here may contribute them?

See similar #5081

peterhashair:
### Description:
Fixes #3371 

Add mod security check on the installation
  
- curl a standard SQL injection in URL that normally blocks by ModSecurity and return 403 to detected if ModSecurity is enabled, but it could block by other WAF.  

### Review


peterhashair:
Please be aware this mod security check only works with apache, and only hosting providers allow that action.

plugins/Diagnostics/Diagnostic/ModSecurityCheck.php
```diff
@@ -0,0 +1,44 @@
...
+
+    public function execute()
+    {
+        $label = $this->translator->translate('Installation_SystemCheckModSecurity');
+
+        $modSecurity = in_array('mod_rewrite', get_loaded_extensions()) !== false;
+
+        $status = $modSecurity ? DiagnosticResult::STATUS_WARNING : DiagnosticResult::STATUS_OK;
+        $comment = sprintf($this->translator->translate('Installation_SystemCheckModSecurityHelp'),
+          $modSecurity ? "On" : "Off / (Not Detected)",
```
sgiehl: This should use translations

plugins/Diagnostics/Diagnostic/ModSecurityCheck.php
```diff
@@ -0,0 +1,44 @@
...
+    public function __construct(Translator $translator)
+    {
+        $this->translator = $translator;
+    }
+
+    public function execute()
+    {
+        $label = $this->translator->translate('Installation_SystemCheckModSecurity');
+
+        $modSecurity = in_array('mod_rewrite', get_loaded_extensions()) !== false;
```
sgiehl: This doesn't look right. Shouldn't that be `mod_security` or something similar? Haven't checked that, but is mod_security even available in the loaded extensions? Or wouldn't we need to look for apache extensions here?

sgiehl (changes requested):

plugins/Diagnostics/Diagnostic/ModSecurityCheck.php
Findus23: I'm pretty sure get_loaded_extensions() returns the PHP extensions (like php-gd), so it would have nothing to do with mod_security.

plugins/Diagnostics/Diagnostic/ModSecurityCheck.php
```diff
@@ -0,0 +1,53 @@
...
+        $label = $this->translator->translate('Installation_SystemCheckModSecurity');
+
+        $status = DiagnosticResult::STATUS_OK;
+
+        $comment = sprintf($this->translator->translate('Installation_SystemCheckModSecurityHelp'),
+          $this->translator->translate('Installation_SystemCheckModSecurityOff'),
+          "<a href='https://matomo.org/faq/troubleshooting/faq_100/' target='_blank'>FAQ</a>");
+
+        // mod security is detected
+        if (function_exists('apache_get_modules') && in_array('mod_security', apache_get_modules())) {
```
sgiehl: Is anyone able to confirm that `mod_security` will appear in the apache modules when it's installed?

Findus23: It won't when using php-fpm as then PHP has no direct "connection" to Apache. I can't say for another mode as I never used apache.

peterhashair:
@sgiehl  that not always worked, I tested on Plesk and Cpanel, that doesn't work, but on standard install, it works. Maybe we should use curl to test specify ModSecurity URL rules that will cause Matomo tracking problems or penitently cause problems instead.

sgiehl:
I actually don't know the problems that may occur when using mod_security so not sure what the best approach would be

peterhashair:
@tsteur any suggestion on that one, ModSecurity is deprecated by 2024. By research the common rule ModSecurity has is a SQL inject URL, but that could be blocked by WAF as well. Not sure the best approach to that one. 🤔

tsteur:
@peterhashair I suggest we close the PR and I could close the issue for now as a wontfix. We scheduled the issue thinking it was easy to do but turns out it's real complicated and we don't have a reliable way to check for it easily so I would say it's not needed for now.

PR #18065, opened 2021-09-27
Show a summary of new features

### LINKED ISSUE #16805: Show a summary of what's new in the app
Could be similar to cloud (the third icon):
image

when clicking on it 
image
image

Ideally plugins can contribute such notifications as well.

We wouldn't mention every feature here but the most important ones. We could also regularly mention new FAQ's, new Guides etc.

To reuse existing code see `Cloud/templates/whatsnew.twig` and the `Menu.php`.

Should we also show this to anonymous when view access? I'd say lets have it disabled for now.

I had added this to the milestone a while back. As it’s important to let people know what’s new so they discover new features and improvements that they otherwise wouldn’t know about and help them utilise the tool better and users stay with us for longer.

* The core would define entries that are shown in the what's new section. 
  * I wonder if we will translate these because by the time they might be translated the entries might be removed again as we would maybe only keep entries for the last 2-3 releases in the code? Might not be worth it translating these entries.
* Plugins should be able to add entries to do this what's new section too
* In the cloud we would want to be able to completely change the shown entries. Meaning we probably need to use DI for these entries. This way cloud can change it and plugins can add entries too.

Structure for such content could be like
* Title
* Description (allowing HTML)
* Learn more link and link text

bx80:
### Description:

Fixes #16805 

Adds a "What's new" notification icon to the admin top menu (new material icon added to the matomo font)
Defines a new event CoreAdminHome.getWhatIsNew which plugins can register for and use to provide an array of items representing their significant changes. These will then be displayed on the "What's new" popup.

This approach leaves individual plugins responsible for updating or removing their "What's new" entries. Perhaps there should be some sort of expiry option? If a plugin adds a "What's new" entry but then isn't updated for 5 years then we wouldn't really want that 5 year old "new" entry to still be shown. 

### Review


plugins/CoreAdminHome/Menu.php
```diff
@@ -37,7 +37,7 @@ public function configureAdminMenu(MenuAdmin $menu)
                 $order = 12);
         }
 
-        if (Piwik::isUserHasSomeAdminAccess()) {
+        if (!Piwik::isUserIsAnonymous()) {
```
sgiehl: Was that change on purpose? Can't see any benefits if normal users are able to see tracking failures.

plugins/CoreAdminHome/Controller.php
```diff
@@ -307,4 +308,50 @@ private function handleGeneralSettingsAdmin($view)
...
+         * }
+         *
+         */
+        $newThings = [];
+        Piwik::postEvent('CoreAdminHome.getWhatIsNew', [&$newThings]);
+
+        if (count($newThings) == 0) {
+
+            $newThings[] = [
+                'title' => Piwik::translate('CoreAdminHome_WhatIsNewNoChangesTitle'),
```
sgiehl: Wondering if it might make more sense to simply hide the icon if there are no changes. This would make the UI bit more clean in this case and it even wouldn't have any benefit for the user to see or click on the icon.

sgiehl (changes requested):
For new features it would be good to add new UI tests covering the new feature. 

I had some thoughts regarding the expiry option, but actually I'm not sure if that would "solve" the possible problems. Guess that depends how accurate the feature should be. Not sure if showing features as new that had been added a couple of versions ago would be that helpful if someone already had a look at it before.

So letting plugins add items using an event might maybe be a bit inaccurate. We can't assume every user is updating Matomo and/or every plugin to each new version. So if each version only reports the new features that were added with the latest version (in the event) and someone skips a version he won't see any new features that were added in one of the skipped versions.

We could maybe always return all new features in the event, that were added in the last 2 or 3 versions, but make it possible to discard those messages in the overlay, so they wouldn't be shown when new messages would be added. Guess the messages then would need to have some kind of unique id, and we would need to store for each user which messages were discarded.

Maybe @tsteur can say something about the required accuracy or if the overlay should always show everything from the last versions... Also I wonder if we should have multiple sections in the overlay, so each plugin can have his own section.

plugins/CoreAdminHome/Menu.php
bx80: Sorry unintentional edit, I've committed a revert

plugins/CoreAdminHome/Controller.php
bx80: Yes, that makes sense. I've moved the gathering event into a method on CoreHome\Controller and it now checks that there are actually changes before adding the menu icon.

bx80:
After discussions with @tsteur and input from @sgiehl, I plan to make the following changes:

- Load the new features/changes from a JSON file in each plugin rather than have plugins pass changes via events.
  - When any update or plugin activation/deactivation happens then check each plugin for an optional changes.json file and merge them all together into an single combined changes.json file. 
  - Have each change require a date in additional to the name, description and link, so that changes can be shown in date order.
  - The "What's new" overlay should then be populated from the combined changes.json file on demand.
  - For expiry of old changes, implement a rolling window where a minimum quantity of the most recent (20?) updates are always shown no matter how old they are, after which any updates older than (180?) days are not shown.
- Store the date that each user last viewed the "What's new" overlay.
- Change the menu bar icon to material icon "bell" same as cloud (the current icon-new_releases looks too similar to help).
- Show an indicator on the "What's new" menu icon if the date of the combined changes.json file is newer than the date the user last viewed "What's new" (ie. there are changes they haven't seen yet). Using material icon "bell ringing" as the indicator in place of the normal "bell" icon.
- Add a UI test for the "What's new" overlay (will need to mock in a consistent set of changes).

sgiehl:
@bx80 haven't had a closer look yet. But there might be a disadvantage with this implementation. If the changes are sorted by the date defined in the json file, that wouldn't reflect when those changes got available in the actual Matomo instance. If you for example install/update a plugin that was already available for quite a while the changes might not even be visible when there are newer changes in other components, right?

bx80:
If the changes in a newly installed plugin were older than the 180 day cut off then, yes they might not be shown, if they are within 180 days then they will always be shown no matter how many other changes there are (might have to scroll though!)  I'm not sure there is a perfect solution here without tracking which user has seen which individual change, which is probably an overkill. 

We could start tracking the plugin install/update date and then also show the last x changes for plugins that were installed or updated in the last 30 days no matter how old those changes are, they'd appear at the bottom of the list if the dates were a while ago, but they at least be there. There would need to be further checks for fresh installs to prevent every plugin getting on the list because they were all just installed. Maybe the installed/updated date would need to be more recent that the Matomo setup date. This adds a fair bit of complexity,

I don't think we could use solely the plugin install/update date to determine the change list order as eventually a single plugin could have hundreds of changes defined going back years and we wouldn't want all of them to be top of the list just because there was a minor update last week.

plugins/CoreHome/Controller.php
```diff
@@ -319,4 +321,61 @@ public function saveViewDataTableParameters()
 
         ViewDataTableManager::saveViewDataTableParameters($login, $reportId, $parameters, $containerId);
     }
+
+    /**
+     * Return an array of changes provided by plugins
+     *
+     * @return array
+     */
+    public static function getChanges()
```
tsteur: fyi @bx80 this you might want to move into a new class  so it's not in a controller? Not sure if otherwise someone could call it directly on the controller otherwise and there's probably no need to have it in a controller if it shouldn't be accessible through the web anyway.

Could ideally also add a basic test if none is there yet

plugins/CoreAdminHome/Menu.php
```diff
@@ -47,9 +49,22 @@ public function configureAdminMenu(MenuAdmin $menu)
     public function configureTopMenu(MenuTop $menu)
     {
         $url = $this->urlForModuleAction('CoreAdminHome', 'home');
-
         $menu->registerMenuIcon('CoreAdminHome_Administration', 'icon-settings');
         $menu->addItem('CoreAdminHome_Administration', null, $url, 980, Piwik::translate('CoreAdminHome_Administration'));
+
+        $changes = \Piwik\Plugins\CoreHome\Controller::getChanges();
+        if (!Piwik::isUserIsAnonymous() && count($changes['changes']) > 0) {
```
tsteur: just to be save could maybe also check additionally if user has some view access?

plugins/CoreHome/Controller.php
bx80: I've moved the method to a new helper class and added a unit test to check that the method is loading values from a fixture generated file and sorting them correctly.

tsteur:
@bx80  are you still working on this issue? I haven't been reviewing as I vaguely remember we talked about potentially somehow slightly changing it to mark at what time the what's new was seen and then potentially remember that. If there is an efficient solution for it without storing heaps of information. Or maybe we just need to push all "what's new" entries into a table after all? Meaning eg when installing a plugin we look for new "what's new" entries and push them into the database and show them that way in the right order or something?

I'll remove the "needs review" label for now. Feel free to add it again if it's ready to review.

bx80:
@tsteur Just pickling this up again. The current implementation of this issue gets all the changes.json items for all plugins, sorts them by the date of the change and hides any items older than six months, always leaving at least x items no matter how old. Whenever a user views the changes list we store the last viewed date. If there are changes dated newer than the last time the list was viewed then the menu icon is changed to a ringing bell, otherwise it's a non-ringing bell and they can still see any recent changes based on change date.

As pointed out by @sgiehl automatically hiding items by change date causes a problem if for example, a plugin is installed 7 months after it's last release then the 'new' changes won't be shown at all.

We could stop hiding change items on based date and only hide/remove them when the user has viewed them but then this causes other problems - on a new install there could be hundreds unread of changes across all plugins. Also currently we're just counting the user opening the changes list dialog as having viewed the items - there is the possibility that they just clicked the changes list by mistake, close it and now the all change items have been removed without actually being read.

How about the following approach:

- Don't hide any changes based on the change date.
- Have two tabs on the dialog, 'New' and 'Read'(?), the dialog always opens to the 'New' tab showing a list of unread changes sorted by change date.
- Each change item that hasn't been read has a 'dismiss'/'ok' button which will mark it as read and move it to the 'Read' list for that user. (each change item's read status to be stored somewhere in user settings or a new table if necessary)
- Have a dismiss-all/mark-all-as-read button which will mark all unread 'New' changes as 'Read'
- If there are any unread change items then the menu icon will be a ringing bell and open to the 'New' list, otherwise it will be a non-ringing bell and open to the 'Read' tab.
- Documentation for plugin authors to recommend that old, obsolete changes be pruned from their changes.json when providing an update, so we don't end up with an entire plugin version history appearing in the changes list when a plugin is newly installed.

Variant B: As above but no dismiss buttons, just opening the dialog marks all 'New' changes items as 'Read', less clicking and they can always see the items on the 'Read' list later on.

Variant C: Don't have an unread tab at all, dismissed/viewed items are permanently hidden. Easier to implement. If there are no new items then we show a 'nothing new message' or hide the icon entirely. Probably not good to combine with B

What do you think?

tsteur:
I'm getting bit afraid it might be getting too complicated. I just checked some changelog/whatsnew services and there are usually not really two different tabs.

What about something similar to our [update logic](https://developer.matomo.org/guides/extending-database#defining-database-updates).

We could a directory say `whatsnew` and files for versions in there. Say

```
plugins/QueuedTracking/whatsnew/4.0.5.json
plugins/QueuedTracking/whatsnew/4.0.6.json
plugins/QueuedTracking/whatsnew/4.0.10.json
// and in core maybe
core/whatsnew/4.0.4.json
```

or even simpler might be to have all in one file like `plugins/QueuedTracking/whatsnew.json` and indexed by version number or have a key version number similar to your example had a date in there. Or of course it could come from an event etc.

Core knows which core/plugin version we are on, and which version are updating to. Then we could look for files in "whatsnew" directory similar to https://github.com/matomo-org/matomo/blob/4.x-dev/core/Updater.php#L343-L390 when someone is updating and push each change into a database table. That way we know which are the newest entries and which ones not.

a table like below
```
id | time | plugin_name | version | message
```

The problem might be finding the right place to check for these without trying to write these values to the DB too often. It might work in the same method where we also look for updates.

Would something like this be maybe a bit simpler/quicker to implement?

The benefit would be simpler UI, not needing to store what which user has seen, or read etc.

bx80:
@tsteur That's fairly similar to what we're doing now, just really a case of replacing the date in whatsnew.json with a version number and pushing records into a new table. Might be cleaner to just read from a single file rather than having to check for multiple files by version name if the update jumps several versions.

From a UI perspective we'd still have a growing list of update records in the table, are you suggesting that we always show the entire table sorted by when the most recent rows were added? So if the user wanted to they could keep scrolling back to see older changes?

If we want to be able to show the 'ringing bell' when there are changes the user hasn't seen then we could store the latest changes table record id against each user instead of the last viewed timestamp, most of the work for that is already done. 

So to summarize:
- Index items in whatsnew.json by version number instead of date
- Hook the plugin update logic to check for new whatsnew.json items by version and push to a new table 'whats_new'.
- The What's New dialog should simply show a list of everything in the table (with pagination) sorted by most recent records
- Store the whats_new table last record id against the user (instead of a timestamp) whenever they view the 'What's new' dialog and use that to determine whether to show a 'ringing bell' icon for new changes or just the non-ringing one if not.

Sounds good to me, are you happy with this spec?

tsteur:
> From a UI perspective we'd still have a growing list of update records in the table, are you suggesting that we always show the entire table sorted by when the most recent rows were added? So if the user wanted to they could keep scrolling back to see older changes?

We could show say the entries from last 3 months. If there are less than 10 entries, then we could show at least 10 entries.

> If we want to be able to show the 'ringing bell' when there are changes the user hasn't seen then we could store the latest changes table record id against each user instead of the last viewed timestamp, most of the work for that is already done.

Sounds great 👍 

Overall sounds also all good 👍

bx80:
This branch has now been updated and this issue is ready for review.
 
There is a new event which fires whenever a plugin is updated or installed. It is caught by CoreHome which then checks the affected plugin folder for a `changes.json` file. If this is found then any changes contained therein will be loaded into a new 'changes' database table. A unique index will prevent changes being loaded multiple times so version number checking can be avoided, this reduces complexity and potential processing problems with unknown plugin version formats.

The 'What's new' icon on the menu bar will be hidden entirely if there are no changes at all in the table. The popup will show changes in the order they were added to the table, latest first. Changes added more than 90 days ago will not be shown unless there would be less than 10 changes shown, in which case the most recent 10 changes are shown.

When a user views the 'What's new' popup then the maximum change table id shown to them is immediately stored against their user record. If the changes table contains records with higher sequence id then the 'What's new' icon will show as a ringing bell to indicate that there are unviewed changes, otherwise it will be a plain bell.

Although the version number and plugin name are not used for anything at the moment, it seems sensible to have them in the table in case there is a need to display this data or add additional filters later on.

To manually test plugin updates I did the following:
1) Added a changes.json file to a random plugin (RerUserDates)
2) Commented out the plugins/CorePluginsAdmin/PluginInstaller.php line 331 `$this->removeFolderIfExists($pluginTargetPath);` to prevent the plugin upgrade wiping the directory
3) Edited the plugin.json and changed the version number to something lower than the current marketplace version
4) Upgraded the plugin from the marketplace
This will cause the changes to be detected and loaded into the changes table.

This is a sample changes file:

```
[
  {"title":       "New feature x added",
   "description": "Now you can do a with b like this",
   "linkName":    "For more information go here",
   "link":        "https://www.matomo.org",
   "version":     "4.0.2"
  },
  {"title":       "New feature y added",
   "description": "Now you can do c with d like this",
   "linkName":    "For more information go here",
   "link":        "https://www.matomo.org",
   "version":     "4.0.1"
  },
  {"title":       "New feature z added",
   "description": "Now you can do e with f like this",
   "linkName":    "For more information go here",
   "link":        "https://www.matomo.org",
   "version":     "4.0.0"
  }
]
```
UI and integration tests have been updated.

bx80:
Just thinking about translations. Currently whatever text is in the changes.json will be shown in 'What's new' regardless of the user's chosen language. There should be a way for plugin authors to provide a list of changes in multiple languages.

 Options could be:

1. Don't do anything about this now. 
2. Look for changes.en.json, changes.de.json, etc and load them all into the table with a new column to denote language, then 'What's New' can filter the changes table by the current user's language setting. Easy to do but bypasses the current translations structure and if there are no translations for the chosen language then no changes would be shown. 
3. Still have one changes.json but all the text values are translation keys, store the translation keys in the changes table and then translate them in the template at run time. If there are no translations available then the user would be shown broken layout,  maybe fall back to English?

Or maybe there is a better way? Thoughts welcome! :slightly_smiling_face:

core/Db/Schema/Mysql.php
```diff
@@ -52,6 +52,7 @@ public function getTablesCreateSql()
                           superuser_access TINYINT(2) unsigned NOT NULL DEFAULT '0',
                           date_registered TIMESTAMP NULL,
                           ts_password_modified TIMESTAMP NULL,
+                          idchange_last_viewed TIMESTAMP NULL,
```
sgiehl, line 55: We used to use Options at lot for storing various user related information, to minimized the required schema changes I guess. @tsteur should an option be used in that case as well, or is an additional database column fine here?

plugins/CoreAdminHome/CoreAdminHome.php
```diff
@@ -32,6 +35,7 @@ public function registerEvents()
             'Template.jsGlobalVariables' => 'addJsGlobalVariables',
             'Translate.getClientSideTranslationKeys' => 'getClientSideTranslationKeys',
             'System.addSystemSummaryItems' => 'addSystemSummaryItems',
+            'Marketplace.pluginUpdated' => 'addPluginChanges',
```
sgiehl: So the changes shall only be visible when a plugin us updated? Not when it is installed? Also I was wondering if the changes of a certain plugin should be removed when the plugin is uninstalled (not disabled)?

sgiehl: Also it would be better to use existing events like `PluginManager.pluginInstalled` and `Updater.componentUpdated`

plugins/CorePluginsAdmin/PluginInstaller.php
```diff
@@ -78,6 +78,29 @@ public function installOrUpdatePluginFromMarketplace($pluginName)
...
+             *
+             * public function pluginUpdated(string $pluginName)
+             * {
+             *    ...
+             * }
+             *
+             */
+            if ($pluginName) {
+                Piwik::postEvent('Marketplace.pluginUpdated', [$pluginName]);
+            }
```
sgiehl: As mentioned in the other comments. This new event shouldn't be needed.

plugins/CoreAdminHome/templates/whatisnewblock.twig
```diff
@@ -0,0 +1,10 @@
+<div class="card">
+    <div class="card-content">
```
sgiehl: Any reason why this template isn't integrated into the other one? It doesn't have much complexity or is very big, so can't see much advantages in extracting it.

plugins/CoreHome/ChangesHelper.php
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Find any new version changes for a plugin
+     *
+     * @param $pluginName
+     * @param $version
+     */
+    public static function loadChangesForPlugin($pluginName, $version)
+    {
+        $file = Manager::getPluginDirectory($pluginName).'/changes.json';
```
sgiehl: I guess it works fine that way. But thinking about it a bit more I wondered if we should design that a bit more flexibel.
The method to provide changes could be located in the `Piwik\Plugin` class. e.g. something like `getChanges`. The base method would read the json file of the plugin then. But it would be possible to overwrite that method in each plugin and provide the changes another way, like directly returning an array or parsing a changelog.md.

plugins/CoreHome/ChangesHelper.php
```diff
@@ -0,0 +1,142 @@
...
+use Piwik\Date;
+use Piwik\Db;
+use Piwik\Piwik;
+use Piwik\Plugin\Manager;
+use Piwik\Plugins\UsersManager\Model AS UsersModel;
+
+/**
+ * CoreHome changes helper class
+ */
+class ChangesHelper
```
sgiehl: Any reason for having a class with static methods only? I would suggest to have non static classes and instead of using the class staticly get the instance using `DI` where it's needed. That way you could also use dependency injection in the constructor to inject used classes like plugin manager, db or the UsersModel

plugins/CoreAdminHome/angularjs/whatisnew/whatisnew.directive.js
```diff
@@ -0,0 +1,43 @@
+/*!
+ * Matomo - free/libre analytics platform
+ *
+ * @link https://matomo.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ */
+
+/**
```
sgiehl: If I see that right, this new angular directive and all the changes to the menu classes are actually there to dynamically create a popup loading the whatsnew url when clicking the menu item, right? I think you can achieve that much more easier by simply giving the menu entry a onclick event to `Piwik_Popover.createPopupAndLoadUrl('module=CoreAdminHome&action=whatIsNew', 'popovername')`

plugins/CoreHome/ChangesHelper.php
```diff
@@ -0,0 +1,142 @@
+<?php
+/**
+ * Matomo - free/libre analytics platform
+ *
+ * @link https://matomo.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ *
+ */
+namespace Piwik\Plugins\CoreHome;
```
sgiehl: As this is more or less something global and not related to the `CoreHome` plugin, I would expect this to be located in `core\Plugin\` e.g. `Piwik\Plugin` namespace

plugins/CoreHome/ChangesHelper.php
```diff
@@ -0,0 +1,142 @@
...
+    {
+        $file = Manager::getPluginDirectory($pluginName).'/changes.json';
+        if (file_exists($file)) {
+            $changes = file_get_contents($file);
+            if ($changes) {
+                $json = json_decode($changes, true);
+                if ($json && is_array($json)) {
+                    $json = array_reverse($json);
+                    foreach ($json as $change) {
+                        if (isset($change['version']) == $version) {
```
sgiehl: That means that changes for versions that were skipped when updating aren't displayed at all, right?

plugins/CoreHome/ChangesHelper.php
sgiehl: Thinking more about that it might even be better to separate things a bit more. So there is one model/dao class that handles the changes entries, having methods like `addChanges`, `getChanges`, `getChangesAfter`, ... and that class is than used by some state class that provides methods like `getChangesForUser` ,`loadPluginChanges`,... If both classes are using dependency injection it will also be possible to write tests for those methods easily, by using mock objects.

sgiehl (changes requested):
Left some comments how to make the code a bit cleaner and simpler.

plugins/CoreAdminHome/CoreAdminHome.php
bx80: Changed to use existing events, so changes will checked on both update and install.

plugins/CorePluginsAdmin/PluginInstaller.php
bx80: Removed

plugins/CoreAdminHome/templates/whatisnewblock.twig
bx80: I've merged the block into the main template.

plugins/CoreHome/ChangesHelper.php
bx80: There is now a getChanges() method on the base plugin class which will load changes from the changes.json file (if it exists), it no longer loads the changes into the database and simply returns an array of changes for the plugin, so this should be easy for plugins to override.

plugins/CoreHome/ChangesHelper.php
bx80: I've reworked the class structure to use DI rather than static methods. There are now two classes `\Piwik\Changes\Model` which handles all the change data operations independent from any user context and `\Piwik\Changes\UserChanges` which retrieves changes and updates last viewed settings in the context of a supplied user model.

plugins/CoreAdminHome/angularjs/whatisnew/whatisnew.directive.js
bx80: I tried something like that initially but the top menu builder doesn't support the onclick parameter. I can't see any reason why top menu items shouldn't be able to use onclick events so I've modified the menu builder and top menu template to allow this. The unnecessary angular directive has been removed.

plugins/CoreHome/ChangesHelper.php
bx80: I've added a new `Piwik\Changes` namespace to contain the changes model and user changes classes. The plugin installed / updated event handlers has been moved to `CorePluginAdmin` which seems more appropriate than `CoreHome` and merely passes the event to the `Piwik\Changes\Model` methods.

plugins/CoreHome/ChangesHelper.php
bx80: Good spot. I've removed the version check, so an attempt will be made to insert all changes loaded from the plugin into the table, but since there is a unique index on plugin_name, version and title and `INSERT IGNORE` is being used then this will prevent duplicates. I'm trying to avoid getting into comparing plugin version numbers which could be in any format.

plugins/CoreAdminHome/CoreAdminHome.php
bx80: Also now catching the `PluginManager.pluginUninstalled` event and removing any entries for that plugin from the changes tables.

justinvelluppillai:
Hey @bx80 if you've resolved the issues above can you please mark them resolved? Then maybe be good to request review directly on this one to get it moving along.

bx80:
All issues should now be resolved, except for whether to keep the user 'last changes viewed' value stored as a new column on the  user table or store as a row in the options table. If a decision can be made on this then I can make the change if required.

core/Db/Schema/Mysql.php
bx80: I checked with @tsteur about using a database column for this and it's ok.

bx80:
After discussion with @tsteur we won't implement translations for the 'What's new' feature at the moment.

This PR is now be ready for final review.

plugins/CoreAdminHome/Menu.php
```diff
@@ -47,9 +47,23 @@ public function configureAdminMenu(MenuAdmin $menu)
     public function configureTopMenu(MenuTop $menu)
     {
         $url = $this->urlForModuleAction('CoreAdminHome', 'home');
-
         $menu->registerMenuIcon('CoreAdminHome_Administration', 'icon-settings');
         $menu->addItem('CoreAdminHome_Administration', null, $url, 980, Piwik::translate('CoreAdminHome_Administration'));
+
+        $model = new UsersModel();
+        $userChanges = new UserChanges($model->getUser(Piwik::getCurrentUserLogin()));
```
sgiehl: In case the user can't be found, this will currently throw an PHP error, as the constructor of `UserChanges` expects an array, but `UsersModel::getUser` might return false then.

_Note: This can happen e.g. in some special cases where we execute some code within `Access::doAsSuperUser`, and the login is set to `super user was set`_

plugins/CoreAdminHome/Menu.php
```diff
@@ -47,9 +47,23 @@ public function configureAdminMenu(MenuAdmin $menu)
...
+
+        $newChangesStatus = $userChanges->getNewChangesStatus();
```
sgiehl: This might throw a database exception during the update. After the code has been updated, the update page will be loaded in order to show the required database updates. This might already trigger the code here, causing the exception, as the `changes` table does not yet exist. You can either add a try/catch here, or in the method itself.

plugins/CoreAdminHome/templates/whatisnew.twig
```diff
@@ -0,0 +1,24 @@
...
+            <div class="card-content">
+                <span style="float: left; font-size:32px; color:#3450A3; margin-right:10px" class="icon-new_releases"></span>
+                <h2 class="card-title">{{ change.title }}</h2>
+                {{ change.description | raw }}
+                {% if not change.link is empty %}
+                <p>
+                    <br>
+                    <a href="{{ change.link }}" target="_blank">{{ change.link_name }}</a>
+                </p>
+                {% endif %}
```
sgiehl: Is there any reason why those changes don't display the plugin name somewhere? I think in a lot cases it will be useful to know that. Imagine some changes like `Added/dropped support for XY`. Not displaying the according plugin name with that might then look like all applies to Matomo itself.

core/Changes/Model.php
```diff
@@ -0,0 +1,114 @@
...
+     * @param string $pluginName
+     * @param array  $change
+     */
+    public function addChange(string $pluginName, array $change)
+    {
+        $table = Common::prefixTable('changes');
+        $insertSql = "INSERT IGNORE INTO " . $table . ' (created_time, plugin_name, version, title, description, link_name, link) 
+                      VALUES (NOW(), ?, ?, ?, ?, ?, ?)';
+
+        $this->db->query($insertSql, [$pluginName, $change['version'], $change['title'], $change['description'], $change['link_name'], $change['link']]);
```
sgiehl: Are all attributes required for a change? If e.g. `link` or `link_name` are not required it might throw a notice here if they were not defined in the changes file.

plugins/CorePluginsAdmin/CorePluginsAdmin.php
```diff
@@ -26,10 +29,35 @@ public function registerEvents()
...
+    /**
+     * Remove any changes from a plugin that has been uninstalled
+     *
+     * @param string $pluginName The name of the plugin that was uninstalled
+     */
+    public function removePluginChanges(string $pluginName)
+    {
+        $changes = new ChangesModel(Db::get(), PluginManager::getInstance());
+        $changes->removeChanges($pluginName);
+    }
```
sgiehl, line 59: We maybe should add a try/catch around adding/removing the changes. I'm actually not sure if that could otherwise cause issues when someone updates from an older version to a version that already contains some plugins with changes files. Depending on the order the updates are processed maybe the plugins are updated before the database migration runs and adding plugin changes might fail, as the table not yet exists.

sgiehl (changes requested):
Not yet fully finished with the review, but thought I leave the comments already.

plugins/CoreAdminHome/Menu.php
bx80: I've added a check to avoid trying to load the user changes if `UsersModel::getUser` returns false.

plugins/CoreAdminHome/Menu.php
bx80: Added a try/catch for the mysql 'table not found' exception in `UserChanges::getNewChangesStatus()` which will then return `NO_CHANGES_EXIST` preventing any attempt to load changes from the table.

plugins/CoreAdminHome/templates/whatisnew.twig
bx80: I've added the plugin name as a prefix to the title, so it will show something like `MyPlugin - Added/dropped support for XY`

core/Changes/Model.php
bx80: `version`, `title` and `description` are required. There is now a check in `core/Changes/Model::addChange()` to only add a change item to the table if it has these fields, if they are missing then a warning is logged and the change item is ignored.

`link_name` and `link` are optional and will now only be added to the insert if they are set.

Added an extra assertion to the `ChangesTest` integration test to check that changes items can be added without a `link` or `link_name` value.

plugins/CorePluginsAdmin/CorePluginsAdmin.php
bx80: I've added a try/catch to both `core/Change/Model::removeChanges()` and `core/Change/Model::addChange()` to handle missing table exceptions.

core/Changes/Model.php
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * Add any new changes for a plugin to the changes table
+     *
+     * @param string $pluginName
+     *
+     * @throws \Exception
+     */
+    public function addChanges(string $pluginName)
```
sgiehl: Not necessarily needed, but if we start using type hints more, we should also use return type hints, even if its
`: void`

core/Changes/Model.php
```diff
@@ -0,0 +1,151 @@
...
+    {
+        if ($this->pluginManager->isValidPluginName($pluginName) && $this->pluginManager->isPluginInFilesystem($pluginName)) {
+            if ($this->pluginManager->isPluginLoaded($pluginName)) {
+                $plugin = $this->pluginManager->getLoadedPlugin($pluginName);
+                if (!empty($plugin)) {
+                    $plugin->reloadPluginInformation();
+                }
+            } else {
+                $plugin = $this->pluginManager->loadPlugin($pluginName);
+            }
```
sgiehl: Guess that could be simply
```
$plugin = $this->pluginManager->loadPlugin($pluginName);
```
It will load the plugin if needed and return it. Can't see any reason for reloading the plugin information.

core/Changes/Model.php
```diff
@@ -0,0 +1,151 @@
...
+
+        if (isset($change['link_name']) && isset($change['link'])) {
+            $fields[] = 'link_name';
+            $fields[] = 'link';
+            $params[] = $change['link_name'];
+            $params[] = $change['link'];
+        }
+
+        $insertSql = 'INSERT IGNORE INTO ' . $table . ' (created_time,'.implode(',', $fields).') 
+                      VALUES (NOW(),'.implode(',', array_fill(0, count($params), "?")).')';
```
sgiehl: `Common::getSqlStringFieldsArray` could be used to generate the `?, ?, ...` list

core/Changes/Model.php
```diff
@@ -0,0 +1,151 @@
...
+
+        try {
+            $this->db->query($insertSql, $params);
```
sgiehl, line 124: Can't `db->insert` be used here?

core/Changes/Model.php
```diff
@@ -79,11 +89,38 @@ public function removeChanges(string $pluginName)
...
+        } catch (\Exception $e) {
+            // Ignore table not found
+            if ($e->getCode() === 42) {
```
sgiehl: Is that the error code for table not found? I think we elsewhere use this to check that:
```
Db::get()->isErrNo($e, Migration\Db::ERROR_CODE_TABLE_NOT_EXISTS)
```

core/Changes/UserChanges.php
```diff
@@ -0,0 +1,102 @@
...
+        } else {
+            $selectSql = "SELECT COUNT(*) AS a, COUNT(*) AS n FROM ".Common::prefixTable('changes');
+            $params = [];
+        }
+
+        try {
+            $res = $this->db->fetchRow($selectSql, $params);
+        } catch (\Exception $e) {
+            // Ignore table not found
+            if ($e->getCode() === 42) {
```
sgiehl: same as mentioned in the other comment...

plugins/CoreAdminHome/Controller.php
```diff
@@ -311,4 +313,21 @@ private function handleGeneralSettingsAdmin($view)
...
+    /**
+     * Show the what is new changes list
+     */
+    public function whatIsNew() {
+
+        Piwik::checkUserHasSomeViewAccess();
+        Piwik::checkUserIsNotAnonymous();
+
+        $model = new UsersModel();
+        $userChanges = new UserChanges($model->getUser(Piwik::getCurrentUserLogin()));
```
sgiehl: `$model->getUser` does not necessarily return an array, but the `UserChanges` constructor expects that. As this is typehinted we should ensure the correct type is given, even though it should normally not happen.

plugins/CoreAdminHome/Menu.php
```diff
@@ -47,9 +47,26 @@ public function configureAdminMenu(MenuAdmin $menu)
...
         $menu->registerMenuIcon('CoreAdminHome_Administration', 'icon-settings');
         $menu->addItem('CoreAdminHome_Administration', null, $url, 980, Piwik::translate('CoreAdminHome_Administration'));
+
+        $model = new UsersModel();
+        $user = $model->getUser(Piwik::getCurrentUserLogin());
+        if ($user) {
+            $userChanges = new UserChanges($user);
+
+            $newChangesStatus = $userChanges->getNewChangesStatus();
+            if (!Piwik::isUserIsAnonymous() && Piwik::isUserHasSomeViewAccess() && $newChangesStatus !== UserChanges::NO_CHANGES_EXIST) {
```
sgiehl: I guess it would make sense to move the check for `!Piwik::isUserIsAnonymous() && Piwik::isUserHasSomeViewAccess()` around the new code, so fetching the changes isn't even tried if it's not needed.

plugins/CoreAdminHome/templates/whatIsNew.twig
```diff
@@ -0,0 +1,24 @@
...
+</div>
+<div class="whatisnew">
+
+    {% for change in changes %}
+
+        <div class="card">
+            <div class="card-content">
+                <span style="float: left; font-size:32px; color:#3450A3; margin-right:10px" class="icon-new_releases"></span>
+                <h2 class="card-title">{{ change.plugin_name }} - {{ change.title }}</h2>
+                {{ change.description | raw }}
```
sgiehl, line 16: is it on purpose to have that printed with `|raw`? In theory someone could place some eval code into the changes and it wouldn't be filtered. I'm not sure if we should allow any HTML within those change messages at all. Imho we don't need that. Or is that required @tsteur

plugins/CoreAdminHome/templates/whatIsNew.twig
```diff
@@ -0,0 +1,24 @@
...
+                {% if not change.link is empty %}
+                <p>
+                    <br>
+                    <a href="{{ change.link }}" target="_blank">{{ change.link_name }}</a>
```
sgiehl: It's important to escape html attributes to ensure it doesn't break the html. Also for potential external links we should still use `rel="noopener"`
```suggestion
                    <a href="{{ change.link|e('html_attr') }}" target="_blank" rel="noopener">{{ change.link_name }}</a>
```

core/Changes/UserChanges.php
```diff
@@ -0,0 +1,102 @@
...
+        if ($idchangeLastViewed !== null) {
+            $selectSql = "
+                SELECT COUNT(*) AS a,
+                  (SELECT COUNT(*) FROM " . Common::prefixTable('changes') . " WHERE idchange > ?) AS n
+                FROM ".Common::prefixTable('changes');
+            $params = [$idchangeLastViewed];
+        } else {
+            $selectSql = "SELECT COUNT(*) AS a, COUNT(*) AS n FROM ".Common::prefixTable('changes');
+            $params = [];
+        }
```
sgiehl: Shouldn't any database usage be placed in the model?

sgiehl (changes requested):
Additionally to the comments:
* It would be nice to have an example changes file, so developers can see how they should look like. Might be good to place it within the `ExamplePlugin`
* It seems you have commited some UI test files, that are now fully broken: e.g.: https://github.com/matomo-org/matomo/blob/6b76c04f7f54331d458614ce159699b6cef00064/plugins/Login/tests/UI/expected-screenshots/Login_bruteforcelog_noentries.png

core/Changes/Model.php
bx80: Good point, I've added return hints and will start adding them to new methods going forwards.

core/Changes/Model.php
bx80: Yes, much better idea. Must. Stop. Reinventing. Wheel. :slightly_smiling_face:

core/Changes/Model.php
bx80: I've replaced the various error number checks with this code instead, definitely easier to read.

plugins/CoreAdminHome/Controller.php
bx80: Added a check for non-array values returned from `$model->getUser()`, in which case an exception will be thrown.

plugins/CoreAdminHome/templates/whatIsNew.twig
bx80: It's on purpose, the original issue screenshots show bullet points in the description and said `Description (allowing HTML)`. Could markdown be an option instead?

plugins/CoreAdminHome/templates/whatIsNew.twig
bx80: Noted and fixed.

core/Changes/UserChanges.php
bx80: I've adjusted the UserChanges model to call a new method in Changes\Model to do the database work.

bx80:
Sorry about the accidental UI screenshots :facepalm:, I've done a commit to revert them.

I've added a sample `changes.json` file to the example plugin, one with a link and one without to show it's okay not to have a link. These changes are present in integration testing which is why the changes test was tweaked slightly.

plugins/CoreAdminHome/templates/whatIsNew.twig
sgiehl: Not sure here. We could also use `strip_tags` and only allow some tags like `<ul><li><br><b><strong><i><em>` and maybe ``, as they would be blocked anyways, if they aren't local

plugins/ExamplePlugin/changes.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "title":       "New feature x added",
+    "description": "Now you can do a with b like this",
+    "linkName":    "For more information go here",
```
sgiehl: Should that be `linkName` or `link_name`? In PHP code you are looking for `link_name`, therefor this change would be added without a link...

bx80: Good spot, fix committed.

sgiehl:
@bx80 there are still some tests failing. One last feedback:

* The links in the what's new popover don't have a special styling. Might be good to have it directly visible that it's a link.

Otherwise this would be good to merge from my point of view. But not sure if maybe @tsteur wants to have another quick look to check if it fulfills the requirements.

Btw. Are we planning to also post some changes in core with each version? If so we maybe should also have a separate changes file for core maybe, as not every feature can be mapped to a specific plugin though.

bx80:
@sgiehl The tests related to this change should all be passing now. There are three UI tests still failing but I don't think they are related.

Good point about link visibility. I've added some styling to match what we typically have elsewhere for 'more info' links.

For core/general changes I was thinking we could put them in the changes.json for the CoreHome plugin which would avoid the need to check for another file. I've tweaked the whatIsNew template to not prefix any CoreHome changes with the plugin name, so they will look like general Matomo changes.

tsteur:
The only thing I would maybe change is a slightly different background color similar to what we have in the cloud.

It looks like we're actually using a Modal there and not a popover. And it has the background `background-color: #fafafa;`
image

it could still be a popover if we could ideally change the background color.


Also noticed in Mobile maybe the width is too wide?  not sure if that's maybe a general Popover issue or if we specify the width somewhere.
image

The modal adjusts automatically by the looks 
image

Maybe it's easier to use Modal? Sorry about bringing this up so late in the process.

Also some events would be good so that Cloud can filter out certain messages for specific plugins or so.

tsteur:
BTW what I find helpful too would be already documenting it on https://github.com/matomo-org/developer-documentation/ 
Sometimes when documenting something then it helps finding things that are maybe complicated to explain and then I sometimes realise need to simplify something. Not saying that's the case here or that it will be. Sometimes that's not the case. I'm just meaning in past it helped me find issues if there are any as then often the docs are hard to explain.

I wonder if we should maybe rename this page https://developer.matomo.org/guides/extending-database to something more generic "Providing Updates" related and mention it there? We could then later also document there how to perform config migrations etc.

bx80:
@tsteur I couldn't find a straightforward way to switch to a modal instead of a popover, we don't seem to have a modal based equivalent of `Piwik_Popover.createPopupAndLoadUrl()` and it seemed an overkill to create a vue component just for this. Instead I added some styling to the popover to set the background colour and adjust to the width dynamically based on screen size.

I've added a new `Changes.filterChanges` event which provides an array of changes that are about to be shown, so Cloud can catch this event and filter out any changes that should be hidden. To allow this filtering to work consistently I modified the `doChangesExist()` code to call `getChangesItems()` rather than using a separate count query, otherwise changes could be found by the count query causing the notification icon to be shown, but then when showing the popup those changes could have been removed by the filter event.

PR 596 has been submitted for review to the developer docs repository with an explanation of how the `changes.json` works. I didn't rename the `Extending the database` section as it seems to be very much about using the database in code rather than updates. Maybe a separate `Providing updates` section for change notifications and migrations could work?

tsteur:
👍  sounds good, we can have a separate `Providing updates` section 

All good re popover vs modal

PR #18076, opened 2021-09-29
[Vue] add test to make sure umd module files are production builds

diosmosis:
### Description:

To catch when someone pushes a development build of a UMD module (as I did in the alert PR).

### Review

### DIFF
--- a/plugins/CoreConsole/Commands/ComputeJsAssetSize.php
+++ b/plugins/CoreConsole/Commands/ComputeJsAssetSize.php
@@ -11,6 +11,7 @@ namespace Piwik\Plugins\CoreConsole\Commands;
 use Piwik\AssetManager;
 use Piwik\Common;
 use Piwik\Development;
+use Piwik\Filesystem;
 use Piwik\Metrics\Formatter;
 use Piwik\Piwik;
 use Piwik\Plugin;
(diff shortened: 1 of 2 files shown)


sgiehl:
@diosmosis would it make sense to also have a test that checks that the built vue files are up to date? Wondering if it otherwise could happen that someone changes a vue component but forgets to update or push the umd files. 🤔

diosmosis:
@sgiehl That sounds like a good idea... I guess it could be done by checking if the diff w/ 4.x-dev has changes to `vue/src` files, and making sure there are also changes to `vue/dist`. But that wouldn't be foolproof since a dev could forget to compile after making an additional source code change.

Alternatively we could modify the travis-scripts to run the build command (after doing an npm install) and checking that nothing was modified. What do you think?

sgiehl:
Not sure if adding that to travis is a good option. Personally I would prefer moving such stuff to a github action. Similar to the PHPCS the action could simply run building the vue js files and check if there would be any changes available. That way we could even enforce the check and a PR couldn't be merged anymore if the files aren't up to date.

diosmosis:
@sgiehl ok, i'll look into using a github action.

tsteur:
@diosmosis can we merge this already or do we need to wait for the github action?

diosmosis:
@tsteur the github action will prevent more issues so I'll still look into that.

diosmosis:
Closing in favor of https://github.com/matomo-org/matomo/pull/18080 which will rebuild vue files automatically.

PR #18080, opened 2021-09-29
[Vue] Adding a "build vue" github action to build vue files

diosmosis:
### Description:

I haven't used github actions before so this is mostly a guess. @sgiehl can you take a look and see if there's anything off? I'm also not entirely sure how to test this. Unless I can invoke the action in this PR w/o merging into 4.x-dev?

### Review


sgiehl:
It can only be executed/tested when it's in the base branch. But you could create a fork where you add the action in the base branch to test it on the fork.

diosmosis:
@sgiehl / @tsteur this is ready for review. The action will now always run for every PR, build vue files, and if there are changes, will push to the PR so the files are always up to date.

tsteur:
@diosmosis looks overall good to me. Might wait with merging until @sgiehl is back from holidays.

Sometimes there might be merge conflicts when different PR changed files in a same plugin etc. In that case we simply need to rebase 4.x-dev into the branch, possibly resolve merge conflicts manually (probably would need to call the build command ourselves) and commit again? 

Not sure if I fully understand what the action does.

diosmosis:
> Not sure if I fully understand what the action does.

It just runs `vue:build` every time a PR is updated, and if there are changes, commits them (in case a developer forgets to re-compile).

> Sometimes there might be merge conflicts when different PR changed files in a same plugin etc. In that case we simply need to rebase 4.x-dev into the branch, possibly resolve merge conflicts manually (probably would need to call the build command ourselves) and commit again?

Mostly, yes. The UMD files don't need to be looked at, other merge conflicts have to be handled, then either run build manually, or push and this command would build again. It doesn't solve the case where the merge is fine *except* for the UMD files which we just want to build again. Not sure there's a way around that.

.github/workflows/buildvue.yml
```diff
@@ -0,0 +1,120 @@
+name: Build Vue files
+
+on: pull_request
```
sgiehl: Should we restrict the types here? Currently it would run for any change of a pull request (including the change of labels,  assignee, review state or when closing a pull request). See https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request

.github/workflows/buildvue.yml
```diff
@@ -0,0 +1,120 @@
...
+          fi
+
+          echo "$PR - $PR_INFO - ${{ github.event.pull_request.number }}"
+
+          echo ::set-output name=branch::$REF
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ steps.vars.outputs.branch }}
+          lfs: false
+        if: steps.vars.outputs.branch != ''
```
sgiehl, line 56: As this command is executed automatically when a pull request is changed, I wonder if we should add an additional check if a vue file was actually changed in this pull request and abort the execution otherwise.

sgiehl (changes requested):

.github/workflows/buildvue.yml
diosmosis: :+1: only done on synchronize now

.github/workflows/buildvue.yml
diosmosis: :+1: done

diosmosis:
@sgiehl updated the pr

.github/workflows/buildvue.yml
```diff
@@ -0,0 +1,128 @@
...
+            exit 0;
+          fi
+
+          if [[ $BASE != $GITHUB_REPOSITORY ]]
+          then
+            echo "It's only possible to update local branches"
+            exit 1;
+          fi
+
+          VUE_FILES_MODIFIED=$(git diff HEAD^ --name-only plugins/*/vue/**/* | wc -l)
```
sgiehl: wouldn't this only look at the last commit? So if someone pushes multiple commits including some vue files, but the last commit doesn't contain any it would abort here?

diosmosis: Ok, checking the base branch now. So it will always rebuild even if the last set of commits doesn't modify vue file, but somewhere the PR does.

diosmosis: ping @sgiehl

diosmosis: actually, nvm, it seems non-trivial to be able to reference commits/branches before the checkout command is run

diosmosis:
@sgiehl ok, it's working now

.github/workflows/buildvue.yml
```diff
@@ -0,0 +1,137 @@
+name: Build Vue files
+
+on:
```
sgiehl, line 3: Might make sense to make it possible to manually trigger that action for a branch maybe:

```suggestion
on:
  workflow_dispatch:
```

sgiehl (commented):
Left another comment. Otherwise I guess it looks fine. Hard to test though. So guess we should merge and adjust it later if anything doesn't work as expected.

.github/workflows/buildvue.yml
diosmosis: I don't think this makes sense, since right now it will always run. I think it's better not to have to think about when it should be run? Especially since this is a safeguard for when people have already forgotten. If the command doesn't work for some reason, we can always build manually until its fixed.

diosmosis:
@sgiehl left a reply above. @tsteur do you want to take a look at this as well?

tsteur:
@diosmosis had a very quick look and looks all good to me. Looks good to merge to me and we can always change things as needed IMO

PR #18086, opened 2021-09-30
Visual indication of incomplete periods on charts

### LINKED ISSUE #10291: make a difference in graphs for data of "unfinished periods" and "complete periods"
Graphs show data of of "unfinished periods" and "complete periods" in same way.*

Since imho comparison of data of "unfinished periods" with "complete periods" does not make big sense, ther should be visialization for the differnce to trigger people to be aware of this problem
One idea is not connecting the data point of "unfinished period" to the others before even this period is finished.
This may look like this:
2016-07-08_10h38_37

* why are there unfinished periods?
When comparing data e.g. number of visits, entry pages etc
with data of day before there is a problem:
when checking earlier than 23:59 
there is always a difference because
the period taken into account is shorter 
than the complete day before
and with this not all "things" of this day has already happened.

An example:
Looking at data at 9:00 in the morning:
actual data is taken from 0:00 to 9:00 = 9h
the data for comparison form 0:00 to 0:00 = 24h

bx80:
### Description:

Fixes #10291 

For charts where the last point on the horizontal axis has a date that is either today or in the future, then that segment of the line will be plotted in a dashed style to indicate that data for that period will be incomplete.

image

### Review


plugins/CoreVisualizations/javascripts/jqplot.js
```diff
@@ -377,6 +377,23 @@ function rowEvolutionGetMetricNameFromRow(tr)
...
+                this.jqplotParams['incompleteDataPoints'] = 0;
+                if (this.param.date && this.param.period) {
+
+                    var endDate = null;
+                    if (this.param.date.indexOf(',')) {
+                        var s = this.param.date.split(',')
+                        endDate = s[s.length-1];
+                    } else {
+                        endDate = this.param.date;
+                    }
```
sgiehl: This won't work for periods that end today or in the future. The `date` param is often not the end of the period range. I guess the easiest solution would be to use the [piwikPeriods service](https://github.com/matomo-org/matomo/blob/2c30dc8d231af58cdf2516a413da2d86bf7d62c6/plugins/CoreHome/angularjs/common/services/periods.js) to parse the period/date and then use the end date for comparing

sgiehl (changes requested):
That already looks nice 👍 
Left a comment as the date check isn't proper. Besides that I'm not sure if that change should also be reflected in the scheduled reports. For those the graphs are rendered using the ImageGraph plugin. That one currently won't use a dotted line for incomplete periods. Not sure if that should be the case for consistency reasons. ping @tsteur @mattab

tsteur:
`ImageGraph` we could have a separate issue for it maybe as most important be the UI (re 80/20 rule). cc @mattab

mattab:
Sounds good to have a separate issue for it, as the most pain is felt in the UI :+1:

mattab:
Looks great. Feedback:
* As people will wonder what the dashed line means, maybe we could edit the tooltip on the last period and mention `(incomplete period.)` under the date in the tooltip? or some other better wording?

plugins/CoreVisualizations/javascripts/jqplot.js
bx80: Thanks for the tip, I didn't know about piwikPeriods, much easier!

tsteur:
@bx80 looks like there might be changes from another PR in there from exclude parameters?

bx80:
@tsteur Because the periods service had been converted to vue while I was working on this issue I merged 4.x-dev into this branch to use those changes. The 4.x-dev branch now includes the exclude parameters change which I think is why those changes are appearing in this branch. I should perhaps have used rebase instead of merge?

justinvelluppillai:
@bx80 merge should normally be fine. I don't see any changes from your other PR here, has it resolved itself meantime or did I miss something? It does have some conflicts now though.

tsteur:
Looks all good to me and works @bx80 . Could we create a UI test though showing the dotted line that this works?

bx80:
It might be tricky to do a UI test as the dotted line will only show if the current period contains today, the check for whether it is today or not is being done in the periods service. I've tried overriding the Date.now() function prior to loading the page in a UI test, which is how the periods service tests work. but those tests directly call the periods calls, and that approach doesn't seem to work here, as I'm guessing a new javascript vm is being spawned by the test system when loading the page. Is there a usual way to fake today's date in javascript for a UI test?

bx80:
@justinvelluppillai The outstanding conflicts are all from the vue.js typescript build output, I suppose I'll need to merge 4.x-dev into this branch again to pickup whatever has changed CoreHome typescript in 4.x-dev and then do another ./console build:vue, then commit an updated set of UMDs, then resolve this branch conflict by preferring them over the current 4.x-dev versions (assuming 4.x-dev hasn't changed again in the meantime) Seems like a 'build vue' github action like we have for 'build js' could save a lot of time going forward.

tsteur:
@bx80 I was hoping that it might work similarly to https://github.com/matomo-org/matomo/blob/4.5.0/tests/UI/specs/OnlyRawDataNotification_spec.js#L11-L13

I'm thinking if we could use `idSite=1&period=year&date=today` as URL and in the fixture track a wee bit of data for this year, and some data for the past year, then this might work? We would probably just need to update the image once a year but otherwise the screenshot might not change.

Be great to check if that could work. It might be even easiest to create a new `spec` file just for this and a dedicated fixture where we track bit of data for this and for the previous year. Since it's something that could easily regress it be very valuable to have a screenshot test for it.

bx80:
@tsteur That approach worked well, thank you. I've committed a new UI test and dedicated fixture to create a few visits in the current and previous year and compare the visitors overview chart.

PR #18099, opened 2021-10-05
Proportional evolution comparison for incomplete periods

### LINKED ISSUE #6784: All websites dashboard: Calculate evolution, making days interpolation and not absolute values
Calculate evolution, making days interpolation and not absolute values as now. So, at 1st of the 30 days month All Websites dashboard looks like this - http://screencast.com/t/KmFkryLR29f, while there is no reason to make everything scary red.
In my opinion 1st of the month data evolution should be calculated as value \* [total days in the month] / [day in a month] , so October 1st data should be equal value \* 1/31, October 10 value \* 10/31 etc.

### LINKED ISSUE #16067: Dashboard percentage incorrect on massive increase
My website increased traffic from about 100 visits to a few thousands. The percentage dropped down to a low number (3%) where instead it should show be above 1000%. Interestingly when sorting by evolution the order is correct, therefore I guess it is probably a display issue.

CleanShot 2020-06-15 at 10 53 10@2x

CleanShot 2020-06-15 at 10 58 50@2x

bx80:
### Description:

Fixes #6784 

This PR changes multi site evolution metrics to calculate proportionally to the percent of the current period which is complete.

For example, if there were 1,000 site visits last week, the current time is midnight on Monday and there have been 150 visits so far this week, then the All Websites visits evolution percent would have shown as a red -85% because 150 visits is a big drop from 1,000. (This Week Visits - Last Week Visits) / (Last Week Visits)

However this in an unfair comparison as the current week is incomplete. If the remaining days of the week continue to at 150 visits then the current week will be an improvement over the previous week.

Since only ~14% of the current week has elapsed a better comparison would be 14% of the previous week total visits vs the current visits in this week, this results in a green 5% evolution percent. (This Week Visits - (Last Week Visits * Percent of Period Complete) / (Last Week Visits * Percent of Period Complete)

### Review


plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
```diff
@@ -170,4 +179,40 @@ private function getPastDataTable()
...
+     * @param DataTable $pastData
+     * @return float|int
+     * @throws \Exception
+     */
+    public static function getRatio(DataTable $currentData, DataTable $pastData)
+    {
+        $ratio = 1;
+
+        $p = $pastData->getMetadata('period');
+        $pStart = new \DateTime($p->getDateStart().' 00:00:00');
```
tsteur: @bx80 is it maybe possible to use our `Piwik\Date` class for this? I suppose something like 

`$pStart = $p->getDateStart()->setTime('00:00:00')` should work as it should create a new object.. It then also has a `getTimestamp()` method so I imagine the code to be quite similar.

I suppose generally there is nothing wrong with using `\DateTime` directly we just haven't done it in the past AFAIK. This might be something to think about in the future. I suppose the benefit of reusing the class is in case there are some kind of quirks or issues then they can be fixed in one class instead of in all the different places or so.

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
```diff
@@ -170,4 +179,40 @@ private function getPastDataTable()
...
+        $pEnd = new \DateTime($p->getDateEnd().' 23:59:59');
+
+        $c = $currentData->getMetadata('period');
+        $cStart = new \DateTime($c->getDateStart().' 00:00:00');
+        $cEnd = new \DateTime($c->getDateEnd().' 23:59:59');
+
+        $now = new \DateTime("now");
```
tsteur: This may not be 100% accurate but we can possibly ignore. In some other data tables we have the information when the report was last generated as a metadata in `DataTable::ARCHIVED_DATE_METADATA_NAME`. That might be interesting because certain periods we might only archive every 6 hours or once per day meaning the math would be a bit off. Eg yearly, monthly we often only calculate once a day.

I wonder if somewhere in the MultiSite report generation this metadata gets lost or whether it's never there in the first place. If we could use the date when the report was generated that could make quite a difference.
image

tsteur (changes requested):
@bx80 nice, I did a very basic check locally and this seems to have worked 🎉 

Could we also add some kind of system tests for this to ensure this works (and keeps working in the future)?

Not sure if easy to do. Be good to add a tooltip on the evolution percentage when the currently viewed report includes today to then say something like that this percentage is based on comparing the time that has been past in the current period.
image

Or if I understand the code directly, that's not just for periods that include `today` but eg would apply when viewing a `month` and the previous `month` has more or less days.

Otherwise people will look at the previous month, and calculate the percentage themselves and be confused why they come up with different percentage.

@mattab can you confirm we also want to adjust the percentage when eg one month has 30 days and the other 31 and then adjust the percentage accordingly? Or should this happen only when the currently viewed period includes today?

mattab:
> @mattab can you confirm we also want to adjust the percentage when eg one month has 30 days and the other 31 and then adjust the percentage accordingly? Or should this happen only when the currently viewed period includes today?

This "proportional-adjustement" logic should happen only when currently viewed period includes today.

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
bx80: I've added a check for the archived date metadata, if it is there then it will be used as the current date/time for comparison purposes, if not then the current date/time will be used instead.

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
bx80: I've switched it to use the Piwik\Date class, makes sense to be consistent.

bx80:
I've added a tooltip to explain when the comparison is proportional.
image

If the current period doesn't contain the current date/time then the proportional adjustment will not be applied (ratio = 1, do nothing)

Also added a unit test which feeds some fixed date empty datatables to the ratio calculation code to check that the expected ratio is returned.

mattab:
Q: would make the tooltip maybe more descriptive if possible to remove any possible confusion in how we came up the with number, for example this below (or a better text):

> The currently selected time period is 53% complete. 
> When the previous period was also 53% complete, there would have been an estimated 5,300 visits (out of a total of 10,000 visits in the previous period). 
> 12,925 visits in today|this week|this month compared to 5,300 visits in the previous period (2021-09-07.) Evolution: +x%

(note: the last line is from the existing tooltip used in Acquisition > Overview report)

tsteur:
@mattab yes it would be, and thought about it, but considering 80/20 it be likely good enough to just have something generic

bx80:
It was quick to add some more detail to the tool tip, so I went ahead and did it, hope that's okay!

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
```diff
@@ -170,4 +186,49 @@ private function getPastDataTable()
...
+     * current incomplete period has passed
+     *
+     * If the current period end is in the past then the ratio will always be 1, since the current period is complete.
+     *
+     * @param DataTable $currentData
+     * @param DataTable $pastData
+     * @return float|int
+     * @throws \Exception
+     */
+    public static function getRatio(DataTable $currentData, DataTable $pastData)
```
tsteur: @bx80 can we add some tests for this method?

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
```diff
@@ -101,6 +108,15 @@ public function compute(Row $row)
         $currentValue = $this->getMetric($row, $columnName);
         $pastValue = $pastRow ? $this->getMetric($pastRow, $columnName) : 0;
 
+        // Reduce past value proportionally to match the percent of the current period which is complete, if applicable
+        $ratio = $this->getRatio($this->currentData, $this->pastData);
+        $period = $this->pastData->getMetadata('period');
```
tsteur: fyi here you could use `DataTableFactory::TABLE_METADATA_PERIOD_INDEX`

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
```diff
@@ -170,4 +186,49 @@ private function getPastDataTable()
...
+        $c = $currentData->getMetadata('period');
+        $cStart = $c->getDateStart()->setTime('00:00:00');
+        $cEnd = $c->getDateEnd()->setTime('00:00:00');
+
+        $nowTS = Date::getNowTimestamp();
+
+        $metadata = $currentData->getAllTableMetadata();
+
+        // If we know the date the the datatable data was generated then use that instead of now
+        if (isset($metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
```
tsteur: it seems this property is not actually set? I wonder if we could request a different a different data table that has this information and then pass it as another argument in `plugins/MultiSites/API.php`

something like `$archive->getDataTable(['Actions_actions_url']);` (makes it bit slower though of course, maybe there's a different way to get the data directly from the DB archive table or so not sure). Not sure if that would work but maybe this way we could actually get the correct archive date and have this bit more accurate as some reports would be archived less often.

tsteur: If it's too complex could also ignore it for now

tsteur (changes requested):
@bx80 nice, that works already. Left few comments and seems there are some tests that need adjusting https://app.travis-ci.com/github/matomo-org/matomo/jobs/542051052 and https://app.travis-ci.com/github/matomo-org/matomo/jobs/542051049

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
bx80: There are now unit tests in plugins/CoreHome/tests/Unit/EvolutionMetricTest that feed in fixed values to the method and check that the ratio is only applied for incomplete date ranges.

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
bx80: I've added an optional parameter to the Row->set() method to allow metadata to be included when creating a row, also added some code to preserve any existing row metadata if the datatable is merged with indexes. This means the Archive->getDataTableFromNumericAndMergeChildren() method will now return row level ts_archived metadata set by default and no additional data table request should be needed. Seems like it could be useful to have this value available by default for other purposes in the future too.

tsteur (changes requested):
Nice. I see you got the `ts_archived` working 👍 

There are some tests failing in https://app.travis-ci.com/github/matomo-org/matomo/builds/239598140

Can we maybe remove this metadata from each row in the final API output? This way we reduce chances of tests failing anymore randomly as ts_archived might be sometimes bit different and reduce a lot of the output in the API response.

We can still use the `ts_archived` in our other filters.

bx80:
@tsteur @sgiehl I've removed metadata from the API output and fixed the broken tests, this PR should now be ready for review again.

tsteur:
Awesome 👍  @sgiehl can you maybe have a last look?

core/Archive/DataCollection.php
```diff
@@ -147,11 +147,17 @@ public function &get($idSite, $period)
      * @param string $period eg, '2012-01-01,2012-01-31'
      * @param string $name eg 'nb_visits'
      * @param string $value eg 5
+     * @param array  $meta Optional metadata to add to the row
      */
-    public function set($idSite, $period, $name, $value)
+    public function set($idSite, $period, $name, $value, $meta = null)
```
sgiehl: we actually can use type hints a bit more to ensure the correct type is given to a method
```suggestion
    public function set($idSite, $period, $name, $value, array $meta = null)
```

We could actually do that for every new parameter or method we introduce, so at some point we don't need all those additional type checks anymore. Feel free to go through the other code in this PR as well and add some type hints where useful.

core/DataTable/Map.php
```diff
@@ -513,6 +513,19 @@ public function getMetadataIntersectArray($name)
         return array_values($data);
     }
 
+    /**
+     * Delete row metadata by name in every row.
+     *
+     * @param       $name
+     * @param false $deleteRecursiveInSubtables
```
sgiehl:
```suggestion
     * @param bool $deleteRecursiveInSubtables
```

core/DataTable.php
```diff
@@ -1001,6 +1001,31 @@ public function getRowsMetadata($name)
         return $metadataValues;
     }
 
+    /**
+     * Delete row metadata by name in every row.
+     *
+     * @param       $name
+     * @param false $deleteRecursiveInSubtables
```
sgiehl:
```suggestion
     * @param bool $deleteRecursiveInSubtables
```

plugins/MultiSites/lang/en.json
```diff
@@ -5,6 +5,11 @@
         "PluginDescription": "View and compare all your websites and apps in this useful 'All Websites' dashboard.",
         "TopLinkTooltip": "Compare Web Analytics stats for all of your Websites.",
         "SingleWebsitesDashboardDocumentation": "This report gives you an informational overview for a specific website, containing the most general metrics about your visitors.",
-        "AllWebsitesDashboardDocumentation": "This report gives you an informational overview for each of your websites, containing the most general metrics about your visitors."
+        "AllWebsitesDashboardDocumentation": "This report gives you an informational overview for each of your websites, containing the most general metrics about your visitors.",
+        "EvolutionComparisonIncomplete": "The currently selected time period is %1$s complete.",
+        "EvolutionComparisonProportional": "When the previous period was also %1$s complete, there would have been an estimated %2$s %3$s\n (out of a total of %4$s %3$s in the previous period).",
+        "EvolutionComparison": "%1$s %2$s this %3$s compared to %4$s %2$s in the previous period (%5$s) Evolution: %6$s"
```
sgiehl: This one might actually be very hard to translate without any further context. Also in some languages it might not even work as `%3$s` is filled with a period name, but the `this`, might need to be translated depending on the filled in word.
In German it would for example be `an diesem Tag` (this day), `in dieser Woche` (this week), `in diesem Monat` (this month), `in diesem Jahr` (this year). Other languages might have similar issues.
Maybe we need to have one translation per period type.

plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js
```diff
@@ -66,6 +66,43 @@
                 site.visits_evolution    = parseInt(site.visits_evolution, 10);
                 site.pageviews_evolution = parseInt(site.pageviews_evolution, 10);
                 site.revenue_evolution   = parseInt(site.revenue_evolution, 10);
+
+                if (site.hasOwnProperty('ratio') && site.ratio != 1) {
+                    var percent = ((Math.round((site.ratio * 100))/100)*100)+'%';
```
sgiehl: using an appended `%` directly is not optimal, as in some languages the `%` might be prefixed or even another character is used. You should use `NumberFormatter.formatPercent()` instead.
Same applies for the other occurrences below.

tests/PHPUnit/System/expected/test_oneVisitor_oneWebsite_severalDays_DateRange_showColumns__MultiSites.getAll_day.xml
```diff
@@ -5,6 +5,10 @@
 		<nb_visits>2</nb_visits>
 		<visits_evolution>0%</visits_evolution>
 		<idsite>1</idsite>
+		<ratio>1</ratio>
```
sgiehl, line 8: Didn't see any tests where the ratio has another value than `1`. Was wondering if it might make sense to create a new system test case maybe where the ratio differs 🤔

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
```diff
@@ -101,6 +109,15 @@ public function compute(Row $row)
         $currentValue = $this->getMetric($row, $columnName);
         $pastValue = $pastRow ? $this->getMetric($pastRow, $columnName) : 0;
 
+        // Reduce past value proportionally to match the percent of the current period which is complete, if applicable
+        $ratio = $this->getRatio($this->currentData, $this->pastData, $row);
```
sgiehl: The method `getRatio` is static and therefor should be called that way.
```suggestion
        $ratio = self::getRatio($this->currentData, $this->pastData, $row);
```

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
```diff
@@ -63,10 +70,11 @@ class EvolutionMetric extends ProcessedMetric
      *                                          $wrapped's name with `'_evolution'` appended.
      * @param int $quotientPrecision The percent's quotient precision.
      */
-    public function __construct($wrapped, DataTable $pastData = null, $evolutionMetricName = false, $quotientPrecision = 0)
+    public function __construct($wrapped, DataTable $currentData = null, DataTable $pastData = null, $evolutionMetricName = false, $quotientPrecision = 0)
```
sgiehl: Those metric classes might actually be in use by any plugin. The class is even listed on https://developer.matomo.org/api-reference/Piwik/Plugins/CoreHome/Columns/Metrics/EvolutionMetric#__construct
So changing the constructor might cause problems in any plugin that might use it.
So we should mention this change in the change log. Or we move the new parameter to the end and adjust the code so it's optional. Currently the code would actually fail if `$currentData` would be null, as `getRatio` is always called, but doesn't allow `$currentData` to be null.

core/DataTable.php
```diff
@@ -1001,6 +1001,31 @@ public function getRowsMetadata($name)
...
+        foreach ($this->rows as $row) {
+            $row->deleteMetadata($name);
+
+            $subTable = $row->getSubtable();
+            if ($subTable) {
+                $subTable->deleteRowsMetadata($name, $deleteRecursiveInSubtables);
+            }
+        }
+        if (!is_null($this->summaryRow)) {
+                $this->summaryRow->deleteMetadata($name);
```
sgiehl:
```suggestion
            $this->summaryRow->deleteMetadata($name);
```

core/DataTable.php
```diff
@@ -1001,6 +1001,31 @@ public function getRowsMetadata($name)
...
+        }
+        if (!is_null($this->totalsRow)) {
+                $this->totalsRow->deleteMetadata($name);
```
sgiehl:
```suggestion
            $this->totalsRow->deleteMetadata($name);
```

sgiehl (changes requested):
Left some comments on possible improvements that came to my mind when looking though the code. I have not yet done some functional tests

core/Archive/DataCollection.php
bx80: Will do, I'm a big fan of type hints but haven't been using them in an effort to be consistent with existing code.

plugins/MultiSites/lang/en.json
bx80: Mein fehler, I've now split this into separate translations for each period type.

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
bx80: I've now moved `$currentData` to the end of the constructor and added null checks and adjusted the method signatures to allow `$currentData` to be null.

tests/PHPUnit/System/expected/test_oneVisitor_oneWebsite_severalDays_DateRange_showColumns__MultiSites.getAll_day.xml
bx80: plugins/CoreHome/tests/Unit/EvolutionMetricTest.php contains tests to confirm that the correct ratio is used when the period is complete and when it isn't. It calls `getRatio()` with a set of known date ranges so we can check the expected ratio is returned, which is 0.5 in the test for exactly half a week. Does this cover it?

plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js
```diff
@@ -66,6 +66,57 @@
                 site.visits_evolution    = parseInt(site.visits_evolution, 10);
                 site.pageviews_evolution = parseInt(site.pageviews_evolution, 10);
                 site.revenue_evolution   = parseInt(site.revenue_evolution, 10);
+
+                if (site.hasOwnProperty('ratio') && site.ratio != 1) {
+                    var percent = NumberFormatter.formatPercent(((Math.round((site.ratio * 100))/100)*100));
```
sgiehl: Seems the rounding here doesn't work as expected. Not sure if it makes sense to have floating digits at all in this case.
image

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
```diff
@@ -101,6 +112,15 @@ public function compute(Row $row)
         $currentValue = $this->getMetric($row, $columnName);
         $pastValue = $pastRow ? $this->getMetric($pastRow, $columnName) : 0;
 
+        // Reduce past value proportionally to match the percent of the current period which is complete, if applicable
+        $ratio = self::getRatio($this->currentData, $this->pastData, $row);
+        $period = $this->pastData->getMetadata(DataTableFactory::TABLE_METADATA_PERIOD_INDEX);
+        $row->setMetadata('ratio', $ratio);
+        $row->setMetadata('previous_'.$columnName, $pastValue);
+        $row->setMetadata('periodName', $period->getLabel());
+        $row->setMetadata('previousRange', $period->getDateStart()->setTime('00:00:00').' - '.$period->getDateEnd()->setTime('00:00:00'));
```
sgiehl: You could use `$period->getLocalizedShortString()` here instead. That way it would be shown as a localized date range, and would also work for days.

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
```diff
@@ -170,4 +190,56 @@ private function getPastDataTable()
...
+    public static function getRatio(?DataTable $currentData, ?DataTable $pastData, Row $row)
+    {
+        $ratio = 1;
+
+        if ($currentData != null && $pastData != null) {
+
+            $p = $pastData->getMetadata(DataTableFactory::TABLE_METADATA_PERIOD_INDEX);
+
+            $pStart = $p->getDateStart()->setTime('00:00:00');
+            $pEnd = $p->getDateEnd()->setTime('00:00:00');
```
sgiehl: This needs to be the end of the period, otherwise the last day won't count. That's also the reason why this doesn't work for a day period at the moment at all.
```suggestion
            $pEnd = $p->getDateEnd()->setTime('23:59:59');
```

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
```diff
@@ -170,4 +190,56 @@ private function getPastDataTable()
...
+
+            $c = $currentData->getMetadata(DataTableFactory::TABLE_METADATA_PERIOD_INDEX);
+            $cStart = $c->getDateStart()->setTime('00:00:00');
+            $cEnd = $c->getDateEnd()->setTime('00:00:00');
```
sgiehl:
```suggestion
            $cEnd = $c->getDateEnd()->setTime('23:59:59');
```

sgiehl (changes requested):

plugins/CoreHome/Columns/Metrics/EvolutionMetric.php
bx80: :facepalm: Good spot, thanks @sgiehl

plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js
bx80: It looks like the /100)*100) to get two decimal places is causes inaccuracy with certain values. I think you're right, floating digits aren't very useful here, so I've removed it.

plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js
```diff
@@ -66,6 +66,57 @@
...
+                    }
+                    if (model.sortColumn == 'pageviews_evolution') {
+                        previousTotal = site.previous_Actions_nb_pageviews;
+                        currentTotal = site.nb_pageviews;
+                        evolution = NumberFormatter.formatPercent(site.pageviews_evolution);
+                        metricName = _pk_translate("General_ColumnPageviews");
+                        previousTotalAdjusted = NumberFormatter.formatNumber(Math.round(site.previous_Actions_nb_pageviews * site.ratio));
+                    }
+                    if (model.sortColumn == 'revenue_evolution') {
+                        previousTotal = NumberFormatter.formatNumber(site.previous_Goal_revenue);
```
sgiehl: for the revenue values we could actually use `formatCurrency`, so it automatically adds the currency.

plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js
```diff
@@ -66,6 +66,57 @@
                 site.visits_evolution    = parseInt(site.visits_evolution, 10);
                 site.pageviews_evolution = parseInt(site.pageviews_evolution, 10);
                 site.revenue_evolution   = parseInt(site.revenue_evolution, 10);
```
sgiehl: @bx80 This seems to cause a problem with the evolution. For some randomly generated data it shows:
image
But comparing 379 to 15 visits would more be like a 2343% evolution, wouldn't it?
I would assume this might be some kind of formatting glitch. We often had that when values are formatted twice. 
It seems the API returns it already formatter and the UI tries to format it again here, causing everything after the thousand separator to be cut of. Actually I'm not sure why we try to format it again here. Maybe we can simply remove any formatting, if API already does that.

sgiehl (changes requested):
Found some other issues while testing it locally. Guess the formatting issue might actually have already existed before 🙈

plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js
bx80: Done, the evolution metric now passes the currency symbol as row metadata since each row is for a different site and might have different currency.

plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js
bx80: @sgiehl I wasn't able to recreate this with my data, I had 1748 visits so far this month vs 355 last month, ((1748-355)/355) = 392% which is what was displayed. Clearly there is some issue here as your screenshot should be showing something like 2426% ((379-15)/15). I tried removing the parseInt() formatting, but it looks like the fancy coloring and up/down arrows on the evolution column need it in this format, so instead I moved this formatting code after the tooltip is generated so the tooltip percent formatting shouldn't be affected. 

Actually just wondering if the issue is that your test data percent is more than 999 so a euro style thousand separator is added by the API formatting giving 2.426,00 then parseInt reads the . as a decimal place giving 2 point 426, which rounds to 2% ? Wild guess, but if that is the case then my last commit should fix the tooltip but the evolution value on the row would still be wrong, as that would be an existing bug not caused by this change.

bx80: Recreated! Looks like any percent higher than 999% will be broken by parseInt()

image

It looks like parseInt() is being used so that the site.directive.html can do a plus/zero/minus comparison to decide which arrows to show, so it can't just be removed. Instead I've committed a fix where the parseInt() values are stored in different model properties instead of overwriting the API formatted values, these new _int values are then used to decide on the arrow to use. Since we only need to know if the value is positive, zero or negative then the decimal places and parts after the thousand separator shouldn't matter. The original API formatted values can then be shown without being mangled by parseInt().

plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js
```diff
@@ -63,9 +63,62 @@
...
+                                site.tooltip += _pk_translate("MultiSites_EvolutionComparisonYear", [currentTotal, metricName, previousTotalAdjusted, site.previousRange, evolution]);
+                                break;
+                        }
+                    }
+
+                }
+
+                site.visits_evolution_int    = parseInt(site.visits_evolution, 10);
+                site.pageviews_evolution_int = parseInt(site.pageviews_evolution, 10);
+                site.revenue_evolution_int   = parseInt(site.revenue_evolution, 10);
```
sgiehl: That actually might not work in various languages. The are languages that are not using a "normal" minus, percent or currency sign. For example: 
https://github.com/matomo-org/matomo/blob/13b22afc4a105262754b51326962c94c305bd6b3/plugins/Intl/lang/sl.json#L878
`parseInt('−234')` returns `NaN`
Same might happen if the currency or percent signs are placed before the number.
Sending a `format_metrics=0` with the request for loading the metrics might not help, as it's overwritten here:
https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/plugins/MultiSites/Dashboard.php#L209
So maybe changing that to something like `$request['format_metrics'] = $request['format_metrics'] ?? 1;` might make that possible.

But that would mean the evolution metrics would be given as floating numbers (eg 0.13 instead of 13%,...)

bx80: @sgiehl Looking at `format_metrics` gave me the idea to add a new API option `format_metrics_add_raw_copy' which accepts a list of data table columns which will have non-formatted '_raw' copies added to the row before they are formatted. The dashboard-model.service can now request raw unformatted copies of the evolution columns which can be used for up/down arrow decisions instead of parseInt() while still having the main displayed evolution columns properly formatted by the API.

This option is disabled by default and should not incur any overhead unless the request utilizes it. Maybe it could be useful elsewhere when both a formatted and unformatted version of a column are needed by the front end.

sgiehl: Seems that approach works. At least some special language edge cases now seem to work:
image

Nevertheless I'm not sure about this new api parameter that allows adding unformatted values. Looks handy in that case, but actually it should also be possible to simply fetch all values unformatted and format them in the UI. Or did you have problems with that?
@tsteur what do you think about this additional api option approach?

plugins/MultiSites/API.php
```diff
@@ -313,6 +313,10 @@ function ($value) {
             );
         }
 
+        // Remove <ts_archived> row metadata, it's already been used by any filters that needed it
+        $dataTable->deleteRowsMetadata(DataTable::ARCHIVED_DATE_METADATA_NAME);
+        $dataTable->deleteColumn('_metadata');
```
sgiehl: Was wondering why the percentages were changing constantly. Debugging the code for a while, showed that the metadata removal is actually done to early here. The computed metrics (like evolution) are applied in the post processor, so after the stuff done in API directly. If you want to do that after computing the metrics you need to apply that using a queue filter, as that is executed in a later step. See
https://github.com/matomo-org/matomo/blob/69c09b4c12b22ef5578a335c487eee7bdf3eb312/core/API/DataTablePostProcessor.php#L108-L138

```suggestion
        $dataTable->queueFilter(function($dataTable) {
            $dataTable->deleteRowsMetadata(DataTable::ARCHIVED_DATE_METADATA_NAME);
            $dataTable->deleteColumn('_metadata');
        });
```

plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js
bx80: Turning off formatting in the API would result in unformatted currency values. Since for the multisites API each row is for a different site and may have a different currency, we'd then need to do something like add the currency symbol to each currency column in the datatable at UI level or add a new filter that formats everything except the evolution columns. It just seemed cleaner to let the API keep formatting the datatable values and have optional raw copies of just the required columns, but if that's not considered an acceptable approach then I can try to rework it provide the datatable unformatted and then apply datatable formatting at UI level.

plugins/MultiSites/API.php
bx80: Good spot! Nice that queueFilter can take a function for simple tasks rather than having to create an actual filter class.

plugins/MultiSites/angularjs/dashboard/dashboard-model.service.js
sgiehl: Well, you are actually already using `site.currencySymbol` to format some currency values, so the currency for each site already seems to be available.
But yes, we would need to format all metrics in javascript then.
Personally I think the best approach would be to load all metrics unformatted and format them as needed in the UI. We had that problem already a couple of times, that metrics were formatted twice (sometimes even with PHP, e.g. in API and the report rending it). At some point we need to find a way that is clean and easy to understand. And in case we need metrics to calculate with them in javascript we imho should load all metrics unformatted. Otherwise we might end up with having different solutions everywhere.
But maybe @tsteur has another point of view here?

tsteur: I'm not fully understand it all. Generally, if any possible, it be great to have all values coming from the API directly (whether formatted or unformatted). This makes sure we have things formatted consistently the same way and then API consumers like our mobile app also format the same way and show the same data etc. 

Like we could have a column `ratio` and a column `pretty_ratio`. 

If I understand this correctly this logic here generates the tooltip? Initially I was thinking to keep the tooltip simple as I figured it might be taking quite a bit of time. @mattab suggested in https://github.com/matomo-org/matomo/pull/18099#issuecomment-938030176 to maybe use more descriptive tooltips but it's not required if it makes the issue too complicated. I'm not too deep into it. Is there maybe something we can do to get like not the perfect solution but 80% there with little effort? I'm not too much into the problem so I'm not sure what that might be. Maybe it would be printing some numbers unformatted like `1343` instead of `1,343`.

Not sure if that helps?

sgiehl: @tsteur the problem actually is, that we try to display a trend picture (positiv / negative / no change) based on the evolution that is received from API. Before there was a "simple" `parseInt` on that number and than it checked if it's positiv, negativ or null. This didn't work properly already before!
While this works for English and a couple of other languages, this won't work for any language that use a different minus sign, or any other format that simply returns `NaN` for parseInt.
The current solution is to have an API parameter `format_metrics_add_raw_copy`, that adds a copy of a certain metric to the response, but unformatted. That works, but might not be the best approach.
Actually this is a general problem we have, whenever we want to calculate in javascript with formatted numbers we received through API.

tsteur: If this was a bug before, we could merge this PR and create a new issue for it? Not sure if I understand it correctly.

> The current solution is to have an API parameter format_metrics_add_raw_copy, that adds a copy of a certain metric to the response, but unformatted.

I'm not much into it. Can we add a new column if trend is positive or negative or no change to the API output? I think this will also help later when there are columns where lower is better etc. Generally it be great when these things come from the API I would say.

sgiehl: Adding a new trend column sound reasonable.
@bx80 could you remove the `format_metrics_add_raw_copy` stuff again and instead change the API method to add an additional column for each evolution metric holding a trend value? Guess this should be easily doable using the `ColumnCallbackAddColumn`. Note: filters will be applied before formatting, so there should be no issue with directly checking the evolution values.

bx80: @sgiehl I don't see what is wrong with the currently implemented and working approach of optionally returning raw metric values from the API where they are needed. Could you help me understand what issues you see with this approach? It's very similar to returning `ratio` and `pretty_ratio` as suggested above and is more generally useful than a positive / negative trend column since it is trivial to determine the trend from the raw figures. Is there a good reason to spend more time on this to achieve a less flexible outcome?

sgiehl: @bx80 Sorry, for not being explicit enough. Sometimes I might assume too much that other people would have the same weird thoughts as me 😀 So let me explain that a bit:

Using the raw metrics is something that works in this case. But for other metrics that won't be correct. For some metrics, like page load time, a lower value would actually be better than a higher value, but it wouldn't be possible to see that from the evolution only. So adding some trend value (like `-1`, `0`, `1`) feels like the best approach to me.
There is a method `Metrics::isLowerValueBetter`, that kind of returns if a negative evolution for a column would be a positive trend and wise versa.
So what we actually should imho do is changing the [evolution metric](https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/plugins/CoreHome/Columns/Metrics/EvolutionMetric.php) class to not only add a `_evolution` metric, but also a `_trend` metric. The trend would then be filled based on the evolution and the result of [`Metrics::isLowerValueBetter`](https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/core/Metrics.php#L238-L281) for the wrapped metric. (You can't use that method on the evolution metric directly, as it works for column metrics only).
That way this should be usable in all cases.
This trend value can the be used in the javascript and there shouldn't be any need for the `format_metrics_add_raw_copy` parameter.
Let me know if something is unclear or if you can think of a better solution.

bx80: @sgiehl Thanks for the detailed explanation, that makes a lot of sense when thinking about metrics where a decreasing value might actually be a good thing! :smiley: I've removed the add raw copy parameter. Instead, when calculating a processed metric, there is a now a check to see if it is an EvolutionMetric, if so then a '_trend' column is added to the datatable with 1, 0 or -1 depending on the calculated evolution and the isLowerValueBetter result. The Javascript then uses the _trend value to decide if up or down arrows are used.

core/API/DataTablePostProcessor.php
```diff
@@ -453,6 +455,17 @@ public function computeProcessedMetrics(DataTable $dataTable)
                 $computedValue = $processedMetric->compute($row);
                 if ($computedValue !== false) {
                     $row->addColumn($name, $computedValue);
+
+                    // Add a trend column for evolution metrics
+                    if ($processedMetric instanceof EvolutionMetric) {
+                        $isLowerBetter = Metrics::isLowerValueBetter($name);
```
sgiehl: This is actually not fully correct. `$name` holds the name of the evolution metric, but `Metrics::isLowerValueBetter` expects the column metric, which is the wrapped metric of the evolution metric. I will push a fix, so you can have a quick look and merge if it's fine

sgiehl (approved):
With the fixes / test updates I have pushed, this PR should be ready to merge.
@bx80 maybe have a look at the additional changes and feel free to merge afterwards

core/API/DataTablePostProcessor.php
bx80: Thanks, that makes sense.

PR #18101, opened 2021-10-06
scheduled tasks should not launch archiving when browser archiving is disabled

### LINKED ISSUE #17976: Scheduled tasks should not launch archiving
See https://github.com/matomo-org/matomo/commit/af44ee1de20bd2f9ace63e7f69f2312c08cfa61e#diff-f41c275e883996b8b69cb765b1818e12e82cbe034e648bc3f7d03610cc7abad1R449-R456

The trigger=archivephp flag will cause scheduled tasks to allow archiving. Seems this hack was needed for `purgeOutdatedArchives` to work. We should fix this issue so that scheduled tasks won't launch archiving when browser archiving is disabled.

This can otherwise cause many tasks to become very slow (including scheduled reports, custom alerts, some on the cloud). And it can cause issues with other archives that were invalidated in the sense that for example some reports might be archived multiple times or that they are archived in the wrong order and therefore causes results to be wrong under circumstances. Because usually the archiver makes sure to first run all daily archives, then the weekly, then ... vs this way there might be race conditions or other problems maybe.

peterhashair:
### Description:

Fixes #17976 

This PR remove $_GET['trigger'] == 'archivephp' from CronArchive.  

### Review


core/CronArchive.php
```diff
@@ -615,14 +615,6 @@ public function runScheduledTasks()
...
         });
 
-        // TODO: this is a HACK to get the purgeOutdatedArchives task to work when run below. without
-        //       it, the task will not run because we no longer run the tasks through CliMulti.
-        //       harder to implement alternatives include:
-        //       - moving CronArchive logic to DI and setting a flag in the class when the whole process
-        //         runs
-        //       - setting a new DI environment for core:archive which CoreAdminHome can use to conditionally
-        //         enable/disable the task
-        $_GET['trigger'] = 'archivephp';
```
tsteur, line 625: @peterhashair here we would need to enable `Rules::$disablePureOutdatedArchive= true`? (to replace the previous hack... and also leave the comment... otherwise the flag would never be enabled)

plugins/CoreAdminHome/tests/Integration/TasksTest.php
```diff
@@ -103,16 +103,14 @@ public function test_purgeInvalidatedArchives_PurgesCorrectInvalidatedArchives_A
...
-
         $wasPurged = $this->tasks->purgeOutdatedArchives();
         $this->assertFalse($wasPurged);
     }
 
     public function test_purgeOutdatedArchives_Purges_WhenBrowserArchivingEnabled_AndCronArchiveTriggerPresent()
     {
         Rules::setBrowserTriggerArchiving(false);
-        $_GET['trigger'] = 'archivephp';
+        Rules::$disablePureOutdatedArchive = true;
```
tsteur, line 113: maybe in `tearDown` method we could now call `Rules::$disablePureOutdatedArchive = false;`? This way the change would be undone again and won't persist

Issue #17941, opened 2021-08-31
Visits Over Time uses all available columns which causes the report generate slowly.


weixuezhang:
When no columns specified in parameter, Visits Over Time uses all available columns which causes report generate slowly.  In this situation, it should use columns from option table record option_name of viewDataTableParameters_***_VisitsSummary.getEvolutionGraph.

## Expected Behavior

When no columns specified in parameter, Visits Over Time should use columns from option table record option_name of viewDataTableParameters_***_VisitsSummary.getEvolutionGraph.

## Current Behavior

Visits Over Time uses all available columns so that generate many metrics not needed and lost time.

## Steps to Reproduce (for Bugs)

In global.ini.php, set "log_level = DEBUG" and "log_sql_queries = 1", we can see query for segment 'done90a5a511e1974bca37613b6daec137ba' which is in Frequency plugin, whose metrics are not shown in Visits Over Time.
DEBUG CoreAdminHome[2021-08-31 03:04:55 UTC] [a58b2] Db::fetchAll() executing SQL: SELECT option_name, option_value FROM `matomo_option` WHERE option_name LIKE ?
DEBUG CoreAdminHome[2021-08-31 03:04:55 UTC] [a58b2] Db::fetchAll() executing SQL: SELECT arc1.idarchive, arc1.value, arc1.name, arc1.ts_archived, arc1.date1 as startDate, arc2.value as nb_visits, arc3.value as nb_visits_converted
DEBUG CoreAdminHome[2021-08-31 03:04:55 UTC] [a58b2] FROM matomo_archive_numeric_2021_08 arc1
DEBUG CoreAdminHome[2021-08-31 03:04:55 UTC] [a58b2] LEFT JOIN matomo_archive_numeric_2021_08 arc2 on arc2.idarchive = arc1.idarchive and (arc2.name = 'nb_visits')
DEBUG CoreAdminHome[2021-08-31 03:04:55 UTC] [a58b2] LEFT JOIN matomo_archive_numeric_2021_08 arc3 on arc3.idarchive = arc1.idarchive and (arc3.name = 'nb_visits_converted')
DEBUG CoreAdminHome[2021-08-31 03:04:55 UTC] [a58b2] WHERE arc1.idsite = ?
DEBUG CoreAdminHome[2021-08-31 03:04:55 UTC] [a58b2] AND arc1.date1 = ?
[...]

PR #18123, opened 2021-10-10
Don't browser archive when segment is set to be pre-processed

tsteur:
### Description:

While commenting on https://github.com/matomo-org/matomo/issues/17941 I noticed that a segment that is selected to be pre-processed would still be archived when it's viewed from browser even though it's set to be preprocessed. When a segment is selected to be pre-processed then it should be only processed during cron archiving no matter the if browser archiving for segments is available or not.

I'm meaning this setting: 
image

When it's currently set to be pre-processed, it would be still archived during day time. Users confirm this makes Matomo slow!

The logic for whether archiving is enabled / disabled or not is currently where complicated IMO:

image

I'm thinking this new logic is more clear and less prone to side effects from when the `shouldProcessReportsAllPlugins` is shown.

image

## Behaviour to be specified

There's also one other issue I noticed in the implementation. When setting `archiving_range_force_on_browser_request=0` in config, then we were still archiving ranges in some cases. Like if browser archiving is enabled (as described in the config setting for it) or  if the `range` is requested during CLI archiving. Not sure if that's on purpuse but I kept this logic. 
image

### Review

### DIFF
--- a/core/ArchiveProcessor/Rules.php
+++ b/core/ArchiveProcessor/Rules.php
@@ -208,45 +208,39 @@ class Rules
 
     public static function isArchivingEnabledFor(array $idSites, Segment $segment, $periodLabel)
     {
-        return !self::isArchivingDisabledFor($idSites, $segment, $periodLabel);
-    }
-
-    public static function isArchivingDisabledFor(array $idSites, Segment $segment, $periodLabel)
-    {
         $generalConfig = Config::getInstance()->General;
 
-        if ($periodLabel == 'range') {
-            if (!isset($generalConfig['archiving_range_force_on_browser_request'])
-                || $generalConfig['archiving_range_force_on_browser_request'] != false
+        if ($periodLabel === 'range') {
+            if (isset($generalConfig['archiving_range_force_on_browser_request'])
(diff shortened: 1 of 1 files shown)


tsteur:
The tests don't seem to fail because of this change. @mattab can you maybe also have a look at the logic and if that makes sense to you?

core/ArchiveProcessor/Rules.php
```diff
@@ -208,45 +208,40 @@ public static function isBrowserArchivingAvailableForSegments()
...
-                return true;
-            }
+        if ($segment->isEmpty()) {
+            // viewing "All Visits"
+            return $isArchivingEnabled;
+        }
 
-            // Always allow processing one report
+        if (!$isArchivingEnabled
+            && (!self::isBrowserArchivingAvailableForSegments() || self::isSegmentPreProcessed($idSites, $segment))
```
tsteur, line 232: fyi looking at other usages of `isBrowserArchivingAvailableForSegments` this change should be fine as in relevant places we already check for `browser archiving available || segment pre processed`

sgiehl (approved):
@tsteur had a look through the changes and logically it imho makes sense. 
As all relevant tests still seem to pass I guess it would be fine to merge.

PR #18124, opened 2021-10-11
set table head sticky

### LINKED ISSUE #3513: Column names could stick to the top of screen when scrolling down report tables
It would be awesome that, when scrolling down and the header row with Column names is not visible, that instead this row would stick at the top of the screen and provide an always clear view of the data being shown. 

Even Piwik team members don't know about the order of columns it is hard to remember in all views... This would improve UX for All users at once!

peterhashair:
### Description:

Fixes: #3513 

update table heads to sticky.

The sticky head will trigger on the first time load, not on screen resize, if you change screen size probably needs to reload the page. let me know if we want to have that on-screen resize. 

On normal screen image

On the smaller device, I used to set the table body to scroll so the header stick on top,  but doesn't seems really good, and failed lots of UI test, so when the table x-axis is overflowed sticky head won't work. Let me know if we want to enable this. 

### Review


bx80 (dismissed):
Works well for me. Probably not important but I noticed that with this change Firefox now shows a console warning "Relative positioning of table rows and row groups is now supported. This site may need to be updated because it may depend on this feature having no effect."

peterhashair:
@bx80 good spot, update a little bit, should stop the warning now.

tsteur:
@peterhashair the tests are failing see https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/49914/ this also happens for me locally

image
image

peterhashair:
@tsteur right, sorry the firefox warning fix broken it, revert it back. should be fine now.

tsteur:
@peterhashair I haven't tested yet but looking at the tests there might be still issues? eg https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/49937/UIIntegrationTest_dashboard2.png 
image

peterhashair:
@tsteur  should work now 😬

tsteur:
I've tested this in few different browsers and mostly worked well. It's pretty awesome! 🚀 

I think it's expected it doesn't work on mobile?

There are still few issues though like this one below:

image

I see the same still in the UI test failures: eg in https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/50077/UIIntegrationTest_metric_tooltip.png

There is also https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/50077/UIIntegrationTest_goals_overview.png

Not sure if this is because of this PR? It doesn't seem related but then the test worked in another 4.x-dev build. Maybe it's a random build failure.

peterhashair:
@tsteur  it doesn't work on mobile, to work on mobile probably needs to restructure the entire table to a div, quite a big change. Let me know if we want to do that.

tsteur:
> to work on mobile probably needs to restructure the entire table to a div, quite a big change. Let me know if we want to do that.

Not needed  for now 👍 

Do you know why eg this screenshot test is failing now https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/50151/UIIntegrationTest_goals_overview.png ?

@sgiehl if you could give this one a test too that be great

peterhashair:
@tsteur I am not too sure why this [https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/50151/UIIntegrationTest_goals_overview.png](https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/50151/UIIntegrationTest_goals_overview.png ) failed, on my local it's fine.

plugins/CoreHome/stylesheets/dataTable/_dataTable.less
```diff
@@ -630,9 +641,36 @@ td.cellSubDataTable .loadingPiwik {
...
 
-div.dataTableScroller {
-  overflow-x: scroll;
+@media only screen and (max-width: 1360px) {
+  .l6 {
+    div.dataTableScroller {
+      overflow-x: scroll;
+    }
+  }
 }
```
sgiehl: Is there any specific reason for this "complexity"? Wondering why `overflow-x: scroll;` can't be set globally for `div.dataTableScroller` as before. If it needs to be changed for `.card-content`, guess it can be overwritten only there?
Btw. not having it globally causes the "regression" that the right border in widgetized reports isn't correct anymore. Which is visible in the UI screenshot changes.

sgiehl: Ah, I see. That's partially needed for the sticky header to work. But Imho the result now looks ugly if the table is too big:

image

peterhashair:
@sgiehl yeah, not the best appoach, I am trying to find those table looks ugly and not apply stick header, I was trying javascript find the width wider than the card, then don't apply stick header, but doesn't really work so well. Most people use scroll body instead of the scroll the whole page, but that means users need to hover the table to scroll. Here is a example [https://codepen.io/peterhashair/pen/NWvRaMv?editors=101](https://codepen.io/peterhashair/pen/NWvRaMv?editors=101)

sgiehl:
Yes. Having the body scroll y combined with an scroll x in the table seems not to work with sticky headers. I'm not sure what the best approach would be, but having overlapping or completely hidden content doesn't sound like a good solution. We for sure could try to implement sticky headers using javascript, but not sure if that would be out of scope for now. ping @tsteur

tsteur:
How much time would it take to get a JS solution to work roughly?

As it's more complex than expected I'm thinking maybe we just don't do it.

peterhashair:
@tsteur I don't think it will take long, the first commit is a javascript solution it works, but it failed the UI test, maybe just need to alter the UI test

tsteur:
👍  @peterhashair sounds good to give it a try then. Let's restrict it to spend max 4 hours for now and then re-assess if needed.

peterhashair:
@tsteur I think this is working, but needs to update a submodule screenshots

plugins/CoreHome/javascripts/dataTable.js
```diff
@@ -111,10 +111,21 @@ $.extend(DataTable.prototype, UIControl.prototype, {
...
+        this.enableStickHead(domElem);
         this.initialized = true;
 
     },
 
+    enableStickHead: function (domElem) {
+      $(window).resize(function () {
+        var tableScrollerWidth = $(domElem).find('.dataTableScroller').width();
+        var tableWidth = $(domElem).find('table').width();
+        console.log(tableWidth);
```
sgiehl:
```suggestion
```

plugins/CoreHome/javascripts/dataTable.js
```diff
@@ -111,10 +111,21 @@ $.extend(DataTable.prototype, UIControl.prototype, {
         this.isEmpty = $('.pk-emptyDataTable', domElem).length > 0;
         this.bindEventsAndApplyStyle(domElem);
         this._init(domElem);
+        this.enableStickHead(domElem);
         this.initialized = true;
 
     },
 
+    enableStickHead: function (domElem) {
+      $(window).resize(function () {
```
sgiehl: This needs to be triggered at least once when the table was loaded/rendered. Otherwise the table will overlap per default, but fit, once the window is resized.

sgiehl (changes requested):
The new solution actually doesn't make sticky headers work in all cases. It kind of disables them for a table that is too wide and only works if the table fits. But imho that approach would be better than having nothing. ping @tsteur

plugins/CoreHome/stylesheets/dataTable/_dataTable.less
```diff
@@ -630,9 +641,10 @@ td.cellSubDataTable .loadingPiwik {
     padding-left:8px;
   }
 }
-
-div.dataTableScroller {
-  overflow-x: scroll;
+.widgetContent,.reportContainer {
```
sgiehl: This actually prevents the sticky header from working in the goals reports, even if there would be enough space.
It's actually not needed anymore as the javascript will add the scrollbar if needed.
```suggestion
.widgetContent {
```

sgiehl (approved):
This should be good to merge now. Will wait for the tests before merging. Maybe they need another update.

sgiehl:
Failing tests are unrelated, so merging now.

PR #18135, opened 2021-10-12
Call setVisitorIdCookie later so disableCookie config can be set first

### LINKED ISSUE #17892: Don't create test cookies when cookies are disabled

justinvelluppillai:
### Description:

Fixes #17892

### Review


justinvelluppillai:
build js

js/piwik.js
```diff
@@ -5115,6 +5114,7 @@ if (typeof window.Matomo !== 'object') {
                 asyncTrackers.push(tracker);
 
                 Matomo.trigger('TrackerAdded', [this]);
+                setVisitorIdCookie();
```
tsteur: @justinvelluppillai I haven't fully tested this yet but generally there are also few other places where we create new tracker instances like `getTracker`, `createFirstTracker`. I suppose we would need to call the method there too?

Although I would assume that when someone calls

```js
var tracker = Matomo.getTracker();
tracker.disableCookies();
```

then `setVisitorIdCookie()` would be still called first before the cookies being disabled if we were to change eg the `getTracker()` method. Maybe we shouldn't call `setVisitorIdCookie()` in `Matomo.getTracker` but then there might be a chance we wouldn't call that method.

It's really difficult as we can't really call `setVisitorIdCookie()` while creating a tracker as they might later disable it. We can't know though when it's save to call that method.

We can obviously call `setVisitorIdCookie` at the latest when there's a tracking request. However, if there is no tracking request on that page, then  it may cause issues. However, looking at the code what `setVisitorIdCookie` does, maybe we don't need to call it at all and assume it will be called with a tracking request if there is one?

In the edge case where there is no tracking request on a page, it would mostly only impact the `RemainingVisitorCookieTimeout` if I understand this correctly.

Not sure if needed but maybe could call `setVisitorIdCookie` on page unload if there was no tracking request or so. This way if there are a few page loads without tracking request this way we would still be extending the visitor cookie timeout a bit.

Not sure this all makes sense. It's quite complicated. Long shorty short I'm thinking maybe we don't need to call `setVisitorIdCookie` there at all

tsteur:
ping @justinvelluppillai there's still one of my comments left

tsteur:
ping @justinvelluppillai

justinvelluppillai:
build js

js/piwik.js
justinvelluppillai: @tsteur  I've implemented the approach you suggested. It now doesn't explicitly set the visitorId cookie but on page unload it checks if there was no tracking request and sets it.

There are a few tests checking the visitorId is the same across a session which now fail. I'm thinking I could update these tests to add a trackPageView() call before the visitorId checks to make sure the cookie is there but wanted to check this makes sense or there's anything I've missed that would be better?

tsteur (approved):
This works for me, did quite some testing around it ( but didn't check the tests). @sgiehl can you think of any issue with this?

justinvelluppillai:
If I do it this way I will need to update quite a few tests which rely on the visitorId cookie being set when tracker is created.

eg see failing tests here https://github.com/matomo-org/matomo/pull/18789

tsteur:
Indeed, seeing https://app.travis-ci.com/github/matomo-org/matomo/jobs/559513056 there are few tests failing. Not sure if they highlight a regression or if they just need to be tweaked. Maybe `var visitorIdStart = tracker.getVisitorId();` would need to be executed after the first tracking request when the ID will be set?

justinvelluppillai:
> Indeed, seeing https://app.travis-ci.com/github/matomo-org/matomo/jobs/559513056 there are few tests failing. Not sure if they highlight a regression or if they just need to be tweaked. Maybe `var visitorIdStart = tracker.getVisitorId();` would need to be executed after the first tracking request when the ID will be set?

Yes I'd need to add a tracking request to all those calls for those expected outcomes to be right. Would need to understand the rationale for the tests to know if I can just change them or it would be a regression also.

sgiehl:
@justinvelluppillai does #18789 replace this PR? If so be good to close this one then.

justinvelluppillai:
This PR has most of the conversation history so worth keeping. The other is just a sandbox for running test experiments. I don't mind which one is closed eventually once I have ironed out the issues but was expecting the other one to be closed and this one merged. Neither has `needs review` label so shouldn't be a problem meantime, right?

PR #18141, opened 2021-10-12
Added functionality to filter getReportMetadata and getSegmentsMetada in SystemTestCase

AltamashShaikh:
Added functionality to filter getReportMetadata and getSegmentsMetada in SystemTestCase 
Issue: #DEV-2366

### Description:

### Review


tests/PHPUnit/Framework/TestCase/SystemTestCase.php
```diff
@@ -59,6 +60,41 @@ abstract class SystemTestCase extends TestCase
      */
     public static $fixture;
 
+    protected function setUp(): void
+    {
+        parent::setUp();
+        if (!isset(static::$fixture)) {
```
AltamashShaikh: @tsteur is this the correct way to initialise and use $fixture ?

tsteur: Each system test case would usually initalise the fixture themselves and usually at this point the fixture should already exist.

Maybe ideally we don't put any of that logic into the fixture as the fixture shouldn't be needed for this. So we could just have two methods in here that can be called. So a system test could do something like this:

```php
  public function setUp(): void
  {
      parent::setUp();
      $this->onlyAllowCategoryForApiMetadataReport('....');
  }
```

There's a chance these extra observers that may be added during the test may not be removed between the system tests so it's possible that the same action/filter is executed multiple times but that shouldn't be an issue. Alternatively, this could be done probably in the `static setUpBeforeClass` method in the test class.

AltamashShaikh: Updated logic and moved code from fixtures to systemTestCase

Plugins would add code like this
```
public static function setUpBeforeClass(): void
    {
        parent::setUpBeforeClass();
        self::setAllowedCategoryToFilterApiResponse('API.getSegmentsMetadata', 'ALLOWED_CATEGORYNAME');
    }
```

Is this fine?
Choose this since setUpBeforeClass is called before setUp and we need this data to be present before adding action

tsteur: I haven't tested it but reckon this could work. 👍

sgiehl:
@AltamashShaikh I think it might be useful to add or change a test in core that uses that new possibility so we can directly see how and if it works. That might also prevent future regressions. You could for example change the custom dimension tests: https://github.com/matomo-org/matomo/blob/d7dced0928c7457f4e14fb571d897ac927b95134/plugins/CustomDimensions/tests/System/ApiTest.php#L162

plugins/CustomDimensions/tests/System/ApiTest.php
```diff
@@ -23,6 +23,12 @@ class ApiTest extends SystemTestCase
      */
     public static $fixture = null; // initialized below class definition
 
+    public static function setUpBeforeClass(): void
+    {
+        parent::setUpBeforeClass();
+        self::setAllowedModuleToFilterApiResponse('API.getReportMetadata', 'CustomDimensions');
```
sgiehl: I guess this needs to be reset in `tearDownAfterClass`, otherwise other tests running afterwards might be affected by this. Could maybe also be done in `SystemTestCase`

tests/PHPUnit/Framework/TestCase/SystemTestCase.php
```diff
@@ -59,6 +60,39 @@ abstract class SystemTestCase extends TestCase
...
+            Piwik::addAction('API.getReportMetadata.end', function (&$reports, $info) use ($allowedModuleForApiMetadataReport) {
+                $this->filterReportsCallback($reports, $info, $allowedModuleForApiMetadataReport);
+            });
+        }
+
+        if (!empty($allowedModuleForApiSegmentsReport)) {
+            Piwik::addAction('API.API.getSegmentsMetadata.end', function (&$reports, $info) use ($allowedModuleForApiSegmentsReport) {
+                $this->filterReportsCallback($reports, $info, $allowedModuleForApiSegmentsReport);
+            });
+        }
```
sgiehl: I think this needs to be moved to the `setUpBeforeClass` method. Calling it in `setUp` causes each test to add a new observer. So running 10 tests would end up having 10 event observers in the last test.
If the event observer is added in the `setUpBeforeClass`, you could also always add it like:
```php
Piwik::addAction('API.getReportMetadata.end', function (&$reports, $info) {
    $allowedModuleForApiMetadataReport = self::getAllowedModuleToFilterApiResponse('API.getReportMetadata');
    if ($allowedModuleForApiMetadataReport) {
        self::filterReportsCallback($reports, $info, $allowedModuleForApiMetadataReport);
    }
 });

Piwik::addAction('API.API.getSegmentsMetadata.end', function (&$reports, $info) {
    $allowedModuleForApiSegmentsReport = self::getAllowedModuleToFilterApiResponse('API.getSegmentsMetadata');
    if ($allowedModuleForApiSegmentsReport) {
        self::filterReportsCallback($reports, $info, $allowedModuleForApiSegmentsReport);
    }
});
```

Note: EvenObserver will be cleared when a Fixture is teared down, so the shouldn't be a need to remove the observers manually.

sgiehl (changes requested):

tests/PHPUnit/Framework/TestCase/SystemTestCase.php
```diff
@@ -100,6 +103,20 @@ public static function setUpBeforeClass(): void
...
+            Piwik::addAction('API.getReportMetadata.end', function (&$reports, $info) use ($allowedModuleForApiMetadataReport) {
+                self::filterReportsCallback($reports, $info, $allowedModuleForApiMetadataReport);
+            });
+        }
+
+        if (!empty($allowedModuleForApiSegmentsReport)) {
+            Piwik::addAction('API.API.getSegmentsMetadata.end', function (&$reports, $info) use ($allowedModuleForApiSegmentsReport) {
+                self::filterReportsCallback($reports, $info, $allowedModuleForApiSegmentsReport);
+            });
+        }
```
sgiehl: As mentioned in a comment before, I think it would be better to always add the Event, but do the check if any allowed modules are set in the action function.
That way it doesn't matter if a class that overwrites the `setUpBeforeClass` calls it's parent before or after calling `setAllowedModuleToFilterApiResponse`. In theory it would then even be possible to call that in a single test only and reset it afterwards

AltamashShaikh: makes sense :+1:

plugins/CustomDimensions/tests/System/ApiTest.php
sgiehl: You could also add a filter for segments here. `API.getSegmentsMetadata` is also tested below

AltamashShaikh: Added

plugins/CustomDimensions/tests/System/ApiTest.php
```diff
@@ -23,6 +23,13 @@ class ApiTest extends SystemTestCase
...
+        self::setAllowedCategoryToFilterApiResponse('API.getSegmentsMetadata', 'Visitors');
```
sgiehl: Might be good if we could filter by multiple values here. The Custom Dimensions plugin adds Segments to `Visitors` and `Behaviour`

AltamashShaikh: Moved to array instead of string

sgiehl (approved):
Looks good to me now. Once travis tests are running again and we can see if all tests are still passing, this should be good to merge

PR #18142, opened 2021-10-12
update SEO Test

### LINKED ISSUE #17919: SEO Ranking plugin not working - problem with ping? - Fix the random test failures
Seeing this in the tests

image

and an error in the UI

image

Seems this regressed.

peterhashair:
### Description:

Fixes: #17919 
update SEO tests, It seems the print $_SERVER['REMOTE_ADDR'] doesn't give back the useful info. 

`bing-index expected non-zero rank, got [0], ip [127.0.0.1]`

### Review

### DIFF
--- a/plugins/SEO/Metric/Bing.php
+++ b/plugins/SEO/Metric/Bing.php
@@ -40,6 +40,8 @@ class Bing implements MetricsProvider
 
             if (preg_match('#([0-9,\.]+) results#i', $response, $p)) {
                 $pageCount = NumberFormatter::getInstance()->formatNumber((int)str_replace(array(',', '.'), '', $p[1]));
+            } elseif (preg_match('#There are no results#i', $response, $p)) {
+                $pageCount = Piwik::translate('General_Error');
             } else {
                 $pageCount = 0;
             }
--- a/plugins/SEO/Metric/Bing.php
+++ b/plugins/SEO/Metric/Bing.php
@@ -33,27 +33,28 @@ class Bing implements MetricsProvider
     public function getMetrics($domain)
     {
         $url = self::URL . urlencode($domain ?? '');
-
+        $suffix = '';
         try {
             $response = str_replace('&nbsp;', ' ', Http::sendHttpRequest($url, $timeout = 10, @$_SERVER['HTTP_USER_AGENT']));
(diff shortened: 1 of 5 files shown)


tsteur:
feel free to merge later @peterhashair 👍

peterhashair:
Did quite a lot of tests and research, 

I think the SEO tests randomly failed is because Bing is a dynamic system with an anti crawler, our crawler exceeds the number of trying in a period of time it will return a page like this. 

I Update bing `getMetrics` when there is no result instead of return 0, return a general error that should skip the tests

image

plugins/SEO/Metric/Bing.php
```diff
@@ -40,6 +40,8 @@ public function getMetrics($domain)
 
             if (preg_match('#([0-9,\.]+) results#i', $response, $p)) {
                 $pageCount = NumberFormatter::getInstance()->formatNumber((int)str_replace(array(',', '.'), '', $p[1]));
+            } elseif (preg_match('#There are no results#i', $response, $p)) {
+                $pageCount = Piwik::translate('General_Error');
```
tsteur: will this be shown in the UI? If so, we would maybe need to be a bit more specific as otherwise people will creates issues and contact support about it and ask for details.

tsteur (commented):
Nice find @peterhashair 👍 

btw just in case this will fail randomly again in the HTTP request we might want to set `$getExtendedInfo=true` so we can check the HTTP response code. Did a quick check and bing says it would return an HTTP 429 if someone exceeds request per minute limit.

I didn't think we issue requests to bing that often but maybe other people on travis running these tests contribute to the quota too.

peterhashair:
@tsteur Bing doesn't return an error status, it loads through javascript. I don't think it will change too much. Only when someone did a website that does not exist will display an error page. It seems like maybe this error needs a little bit of update on the code. Every search engine looks different, let me know if we want to update this. 
image

tsteur:
i see. Didn't realise others were already showing that simple error message. Possibly they don't error as often. We've had errors before on the cloud for bing so this happens more often. It be great to possibly use the same consistent message possibly. Even if it's just like an `Error. Please try again later.`

peterhashair:
@tsteur Sure, that should fix the SEO test and the error page.
image

PR #18143, opened 2021-10-13
Include lox/xhprof in the Matomo build release

### LINKED ISSUE #17939: Include lox/xhprof in the Matomo build release so Profiling works
Only works when installed from git but no longer works when installed from regular release zip.

That's likely because `lox/xhprof` is marked as development dependency in composer.json.

tsteur:
### Description:

fix https://github.com/matomo-org/matomo/issues/17939

### Review


sgiehl:
Wondering if we also need to update https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/core/Composer/ScriptHandler.php#L21 and the referenced shell script, as it still uses `facebook/`, while the package now uses `lox/`

tsteur:
@sgiehl changed those. Not really sure though it's generally still needed.

sgiehl:
If it's not needed, guess we should simply remove it. There also seems to be a `build-xhprof.sh`, not sure what it was good for and why it is not called in `ScriptHandler::buildXhprof` 🤷 
Btw. Guess it might be needed to run a composer update after this was merged, otherwise it might remain in the dev packages in composer.lock.

tsteur:
I removed these scripts for now. We can always add back if needed for some reason.

sgiehl (approved):

PR #18151, opened 2021-10-14
update database test work without section set

### LINKED ISSUE #17856: [database_tests] section no longer generated when executing php tests
to reproduce remove the section `[database_tests]` from your local tests. It will then result in a database access exception.

That's because in https://github.com/matomo-org/matomo/blob/4.4.1/tests/PHPUnit/bootstrap.php#L126 it uses a mock config which has "allowSave = false".

It must have regressed at some point but not sure where.

The goals are:

* [ ] Not throw an exception when `database_tests` section is not configured.
* [ ] Automatically pre-fill the database_tests section in the `config/config.ini.php`

image
image

or is this maybe now expected?

peterhashair:
### Description:
Fixes: #17856  
update database test work without section set. use `getenv('TEST_SUITE')` to diff the travis and local.

### Review


tests/PHPUnit/bootstrap.php
```diff
@@ -69,10 +69,12 @@ function setupRootContainer() {
...
     // phpunit's class loading (if a test class is loaded in bootstrap.php, phpunit can't load it from a file,
     // so executing the tests in a file will fail)
-    $vars = new TestingEnvironmentVariables();
-    $vars->delete();
+    if(getenv('TEST_SUITE')) {
+        $vars = new TestingEnvironmentVariables();
+        $vars->delete();
 
     Environment::setGlobalEnvironmentManipulator(new TestingEnvironmentManipulator($vars));
+    }
```
sgiehl, line 77: Could you explain what this change is needed for, eg what it should improve/solve?

peterhashair:
@sgiehl Sorry for the confusion. The purpose of this change is to run Matomo PHP tests locally without setup `[database_tests]` section in the `config.ini.php`.  So it should copy the `username` , `password` , `host` from `[database]` then make a matomo_tests database. 
[https://github.com/matomo-org/matomo/blob/852c3bdfa59c427d40c50286afe5e532e56736b8/tests/PHPUnit/bootstrap.php#L72-L75](https://github.com/matomo-org/matomo/blob/852c3bdfa59c427d40c50286afe5e532e56736b8/tests/PHPUnit/bootstrap.php#L72-L75)

I think those three lines init an empty test environment, the `forceSave` won't append default database connection to `[database_tests]` in the `config.ini.php` file, the local PHP tests doing database access deny . But if I removed it,  Travis start to fail, so I add `getenv` to different 2 environments.  



  
  [https://github.com/matomo-org/matomo/blob/852c3bdfa59c427d40c50286afe5e532e56736b8/tests/PHPUnit/Framework/Fixture.php#L265](https://github.com/matomo-org/matomo/blob/852c3bdfa59c427d40c50286afe5e532e56736b8/tests/PHPUnit/Framework/Fixture.php#L265)
  Remove this line. Because after the test runs locally, the code alters the `[database]` section. A `forceSave()` has been called.
   [https://github.com/matomo-org/matomo/blob/249adc21db61899ae1d908a17da4ada46e663b88/plugins/Installation/tests/resources/config.ini.php#L12](https://github.com/matomo-org/matomo/blob/249adc21db61899ae1d908a17da4ada46e663b88/plugins/Installation/tests/resources/config.ini.php#L12)
Add this `host = "127.0.0.1"`, purpose to pass `Plugins\CoreAdminHome\tests\Integration\Commands\ConfigGetTest`tests/PHPUnit/proxy/index.php` assert database access deny. Because I empty the `[database_tests]` localhost in  `global.ini.php` to auto pass from `[database]`.

sgiehl:
@peterhashair The problem with not using the `TestingEnvironmentManipulator` locally at all will cause tests to overwrite your local config. Similar to the database section update (you removed it), now any test can manipulate it. Try e.g. running the test `MultipleSitesArchivingTest`. This will then change your config file:
https://github.com/matomo-org/matomo/blob/6d744db7f3f7a55fc4bb1ca573e937f0832c642b/tests/PHPUnit/System/MultipleSitesArchivingTest.php#L47-L48

peterhashair:
@sgiehl that actually make sense, update it a little bit. so init again after the config database_tests is set.

peterhashair:
@sgiehl that one good to go?

sgiehl (approved):
checked that again and it seems to work as expected now.

PR #18156, opened 2021-10-14
Small optimisations to Archiving performance

samjf:
### Description:

issue: DEV-2055

Quite a few minor optimizations have been spotted while looking for regressions in archiving. Quite a few of them are minor in changes, but significant wins.

These changes should provide a few performance optimisations for archiving.

### Review

### DIFF
--- a/core/DataTable/Row.php
+++ b/core/DataTable/Row.php
@@ -473,6 +473,7 @@ class Row extends \ArrayObject
      */
     public function sumRow(Row $rowToSum, $enableCopyMetadata = true, $aggregationOperations = false)
     {
+        $operationsIsArray = is_array($aggregationOperations);
         foreach ($rowToSum as $columnToSumName => $columnToSumValue) {
             if (!$this->isSummableColumn($columnToSumName)) {
                 continue;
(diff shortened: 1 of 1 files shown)


core/DataTable/Row.php
```diff
@@ -650,6 +652,10 @@ public function setComparisons(DataTable $table)
      */
     protected function sumRowArray($thisColumnValue, $columnToSumValue, $columnName = null)
     {
+        if ($columnToSumValue === false || $columnToSumValue === null) {
```
samjf: @tsteur Is it realistic that `$columnToSumValue` is ever `null`? 
It seems when debugging that a lot of null values are passed in from page performance metrics, but that could be from broken local data.

tsteur: Not sure @samjf possibly it's in few cases but most of the time I don't expect this to be the case. This is now actually breaking some tests see eg https://app.travis-ci.com/github/matomo-org/matomo/jobs/543316947#L958

Seems it might be only the case for some bandwidth metrics. I'm not sure if this change would be expected or not but probably wouldn't check for `null` here as it could introduce some regression maybe (or it could fix something hard to say).

samjf: @tsteur Hmm yeah, I think it returns 0 if it is null in fact. I'm not sure if it is suppose to or not, but safer to remove that.

PR #18164, opened 2021-10-15
Install routine: Fix text contrast for accessibility

### LINKED ISSUE #17785: Install routine UI: improve text contrast
Install routine: text color

grafik
(checked with the development tool in Firefox)

## Expected Behavior
darker grey (better: black), that has a much better contrast

## Current Behavior
grey font color with low contrast.
font-color: #858585 (contrast ratio: 3.7)

## Possible Solution
font-color: #757575 (contrast ratio: 4.5 => minimum ratio https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast)

## Your Environment
* Matomo Version: 4.3.1
* Browser: Firefox 90

donovanglover:
Fixes #17785.

### DIFF
--- a/plugins/Installation/stylesheets/installation.css
+++ b/plugins/Installation/stylesheets/installation.css
@@ -81,7 +81,7 @@ h3 {
     margin-bottom: 15px;
 }
 .content p {
-    color: #858585;
+    color: #757575;
     margin-bottom: 10px;
 }
 p.next-step {
--- a/plugins/Installation/stylesheets/installation.css
+++ b/plugins/Installation/stylesheets/installation.css
@@ -26,6 +26,10 @@ body {
     color: #43a047;
 }
 
+#installation .form-group label {
+    color: #757575;
+}
+
 .header .languageSelection {
     color: #1976D2;
     font-size: 13px;


sgiehl:
@tsteur @mattab Is this color change fine, or should we use an even darker color?
The change can be seen here: https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/50103/Installation_start.png

tsteur:
It could be even darker but it's good enough for me as it's better than before.

sgiehl:
The headlines for the input boxes (`.form-group label`) is currently using `#76838f`, which is also to slightly too light. So maybe we should overwrite this in the install.css with the same value as the text so it matches.

sgiehl:
@donovanglover can you have a look at that?

donovanglover:
I added the CSS but I don't know which page has the label. Hopefully it works!

sgiehl:
@donovanglover It's the page to configure the database

sgiehl:
@donovanglover seems your change did the trick. I triggered another build and it looks fine now. Will merge this one and update our UI tests afterwards. Thanks for your contribution 👍

PR #18166, opened 2021-10-17
[Vue] Migrate content-block, enriched-headline, rate-feature + related

diosmosis:
### Description:

Changes:
* add static AjaxHelper.fetch method to make it seem more like piwikApi service
* migrate RateFeature component
* modify MatomoDialog to use v-model so it's simpler to use from vue
* made `vue:` property optional in scope mapping in createAngularAdapter
* migrate enrichedheadline
* allow translate function to take variadic args OR array of strings like _pk_translate
* migrate contentblock component
* (not related to vue) fix anchor links in trackingcodegenerator page (linking to `#anchor` instead of `#/anchor` causes angularjs to fail, happens on 4.x-dev too)
* fix nested transclude issue w/ createAngularAdapter

Notes:
* There is a cyclic dependency in terms of plugins here, with Feedback's RateFeature component depending on CoreHome's MatomoDialog, but CoreHome's EnrichedHeadline depending on Feedback's RateFeature. This was not a problem w/ angularjs since it was all just files depending on other files, but now there are UMD modules that depend on each other. This is mostly an issue for existing code, for new code it will be noticed early on. I was able to work around it like this: https://github.com/matomo-org/matomo/pull/18166/files#diff-0bb6c3fc35d58a5d2477dd55edaf3478e41bbf44ee5d0c051ff60c11f31123bbR79-R88

image

image

image

**Note: it seems adding components does bump the the JS payload a bit. The 4 new vue component templates add about 2kb to the gzipped content.** It might be good to set a threshold for when it would be good to explore either using async components or loading plugin UMD's lazily. Both are doable, but it would be easier to manage the latter option in the long run.

### Review


plugins/CoreHome/templates/_dataTable.twig
```diff
@@ -41,7 +41,7 @@
         <div class="card-description">{{ properties.description }}</div>
     {% endif %}
 
-    <div class="reportDocumentation">
+    <div class="reportDocumentation" data-content="{{ properties.documentation|e('html_attr') }}">
```
diosmosis: Note: this is for EnrichedHeadline so it can grab the HTML. In the angularjs directive, it gets it from ng-bind-html, but for vue we don't want to reference an angularjs directive like this.

sgiehl: does this need a `|default` like below, so it doesn't fail when `properties.documentation` is not defined for any reason?

diosmosis: :+1:

plugins/CoreHome/vue/src/ContentBlock/ContentBlock.vue
```diff
@@ -0,0 +1,120 @@
...
+
+<template>
+  <div class="card" ref="root">
+    <div class="card-content">
+      <h2
+        v-if="contentTitle && !actualFeature && !helpUrl && !actualHelpText"
+        class="card-title"
+      >{{ contentTitle }}</h2>
+      <h2
+        v-if="contentTitle && actualFeature || helpUrl || actualHelpText"
```
sgiehl: Is the operator precedence correct here? In the old template it had been `contentTitle && (feature || helpUrl || helpText)`

diosmosis: :+1: missed that, i'll double check the templates

diosmosis: @sgiehl fixed it, couldn't find any other occurrences. thanks for catching that!

plugins/CoreHome/vue/src/ContentBlock/ContentBlock.vue
```diff
@@ -0,0 +1,120 @@
...
+    if (this.actualFeature && (this.actualFeature === true || this.actualFeature === 'true')) {
+      this.actualFeature = this.contentTitle;
+    }
+
+    if (adminContent === null) {
+      // cache admin node for further content blocks
+      adminContent = document.querySelector('#content.admin');
+    }
+
+    let contentTopPosition: number;
```
sgiehl, line 99: I'm actually not familiar with how this is interpreted. Will `contentTopPosition` be initialized with `null` here or with `0` as it's a number?

plugins/CoreHome/vue/src/ContentBlock/ContentBlock.vue
```diff
@@ -0,0 +1,120 @@
...
+    if (adminContent) {
+      contentTopPosition = adminContent.offsetTop;
+    }
+
+    if (contentTopPosition || contentTopPosition === 0) {
+      const parents = root.closest('[piwik-widget-loader]');
```
sgiehl, line 105: This used to be `var parents = element.parentsUntil('.col', '[piwik-widget-loader]');`, which selects all `.col` elements up the tree till `[piwik-widget-loader]`. But now it directly looks for the closest `[piwik-widget-loader]`, which ihmo is something different. Or am I wrong here? 🤔

sgiehl (changes requested):
Left some more comments. Besides that, the code already looks fine. Will do some cross browser testing next week and hopefully travis tests will then run again, so we can see if all tests are passing or not.

plugins/CoreHome/vue/src/ContentBlock/ContentBlock.vue
diosmosis: It should be initialized as `undefined`.

plugins/CoreHome/vue/src/ContentBlock/ContentBlock.vue
diosmosis: It does, then it only uses the top most. So it's the first .col under [piwik-widget-loader]. I found it overly complicated, and the tests passed, so I was hoping we didn't need complicated select. Do you think it should look for .col?

plugins/CoreHome/vue/src/ContentBlock/ContentBlock.vue
diosmosis: Though it seems sometimes typescript complains about this, sometimes not...

plugins/CoreHome/vue/src/ContentBlock/ContentBlock.vue
sgiehl: Had a closer look at the code and that actually only seems to affect the admin widgets, where it seems to work correctly that way. So should be good to keep that simpler 👍

plugins/CoreHome/vue/src/ContentBlock/ContentBlock.vue
sgiehl: The reason for the question actually was to clarify if the default value would pass the condition below:
`contentTopPosition || contentTopPosition === 0`. 
Before it was initialized with `false`. Guess `undefined` should work correctly as well then.

sgiehl (approved):
Seems to work as expected. Once we have tests running again and they pass this would be good to merge.

sgiehl:
@diosmosis Unrelated to this PR, I get the following notice in the browser console
```
Source-Map-Fehler: Error: NetworkError when attempting to fetch resource.
Ressourcen-Adresse: webpack:////home/dizzy/Projects/matomo/node_modules/dompurify/dist/purify.js?
```
Seems the path is your local one...

diosmosis:
@sgiehl I grepped for it and its in the source map for the polyfill build... not sure why it's using absolute paths, the build config might be wrong there... I'll take a look, thanks for mentioning it!

diosmosis:
@sgiehl should be fixed

diosmosis:
Actually it seems the recommended approach is to not include the sourcemap in production: https://webpack.js.org/configuration/devtool/, but since we distribute unminified javascript with plugins (and much of Matomo is open source), perhaps this isn't something we need to care about... cc @sgiehl @tsteur

diosmosis:
Actually, actually, since we just merge everything into one big JavaScript asset, the sourcemaps are useless in production.

diosmosis:
@sgiehl some test failures but they don't seem related

sgiehl:
> Actually, actually, since we just merge everything into one big JavaScript asset, the sourcemaps are useless in production.

Should we remove them from the build then, so they aren't included in the release zip?

sgiehl:
Tests seem to pass. Will merge this now.

diosmosis:
@sgiehl 

> Should we remove them from the build then, so they aren't included in the release zip?

I guess we could, they're only really useful for development, and they'll be generated when we develop. I'll make a PR.

PR #18168, opened 2021-10-18
Cache translated timezone name in enrichSite

tsteur:
### Description:

Noticed the `getTimezoneName` method can be very slow and CPU consuming. Especially `new DateTimeZone($timezone);` etc. Could technically cache the `getTimezoneName` method itself but figured it's a bit easier to only cache it for now in `enrichSite` as we then don't need to worry about cache key for different getTimezoneName parameters etc.  The method isn't a concern when called from other places. Often most sites have the same timezone so this will be an improvement. On cloud we're calling getAllSites etc more often and we noticed it had a performance impact.

### Review


sgiehl (approved):
LGTM

PR #18170, opened 2021-10-18
Fix for HTML entity decoding of single quotes in campaigns and csv/xml exports

### LINKED ISSUE #14695: Special character apostrophe is not getting printed properly in visits log (campaign url)
Expected: apostrophe should be printed in visits log

Result: instead we can see HTML entity &#39 is getting printed in visits log (no issue in campaign report)

steps to reproduce:
1.     Create a Campaign URL with params containing apostrophe
2.     Visit the page through campaign URL
3.     Check the Visits log

Note: When using the campaign URL, in the address bar of the browser it's getting encoded properly as %27 and also in the Campaign report. Issue occurs only in "Visits log"
campaign_params_error
campaign_report

bx80:
### Description:

Fixes #14695 

Tweaked the campaign referrer name decoding so that HTML entity single quotes are properly rendered when displayed on the visitor log.

Also changed the XML and CSV datatable render to use HTML entity decode with the ENT_QUOTES option instead of the  ENT_COMPAT option so that exported data with entity encoded quotes is correctly converted.

### Review


sgiehl:
@bx80 Are you still working on that one? Had a quick look, but it didn't seem to work as expected. Feel free to readd the `Needs Review` label, once you are done.

bx80:
@sgiehl I've added an additional commit to fix issue #18190 where outlinks with single quotes are not decoded on the visits log. I've checked both the campaign referrer name and outlinks and they both correctly show the single quotes for me. If it's still not working for you, could you provide a screenshot or the test string used? Then I can try to recreate. Maybe character set is a factor?

My test steps for the campaign referrer:
1) Create a new campaign with a name containing a single quote
2) Visit a tracked page using the campaign URL.
3) Checked the visit log campaign name. (contained &#39; before the fix, single quote after)
visit log campaigns single quote fix

For the outlinks:
1) On my test tracked website, added a new link to the wikipedia article reported in the issue.
2) Followed the link to record a tracked visit.
3) Checked the visit log actions outlink. (contained &#39; before the fix, single quote after)
visit log outlink single quote fix

Thanks!

sgiehl:
@bx80 How did you track the campaign? I have a local site that simply tracks a page view. Adding `?pk_campaign=te%27%22st` to the url, triggers tracking of a campaign with the name `te'"st`.

Visitor log shows this for the visit:
image

The outlink seems to work correct now.

bx80:
@sgiehl That's pretty much how I did my tests, a local test site tracking page views with a campaign url. I've just copied and pasted your example campaign url from the comment above and repeated the test - it still works fine for me. It seems like there is some other factor here. I wonder if it could be database collation / character set dependent? `SELECT @@character_set_database, @@collation_database;` shows that I'm using utf8mb4 / utf8mb4_general_ci, how about you?

sgiehl:
My database collation ist actually still latin, but all tables are utf8mb4. But I don't think that should be the issue.
Maybe you could extend/change the UI tests of the Live plugin by some data containing such special chars, so we can see in the tests if it works or not.

plugins/Live/VisitorDetails.php
```diff
@@ -95,6 +95,10 @@ public function renderAction($action, $previousAction, $visitorDetails)
 
         $sitesModel = new \Piwik\Plugins\SitesManager\Model();
 
+        if (isset($action['type']) && $action['type'] == 'outlink' && isset($action['url'])) {
```
sgiehl, line 98: When adding/changing the tests it might be good to also add some internal urls and downloads containing such special chars. And maybe even some referrer urls. Wondering if that only doesn't work for outlinks, or if we need to fix that for all urls actually

bx80: @sgiehl I've added single quotes, double quotes, %27 and %22 special characters to a referrer URL, internal URL, outlink URL and campaign name in the live visitor UI tests.

Locally I get the following results:
- **Internal URL** (tablet visit): direct quotes (' and ") are preserved, entities (%27 and %22) are not converted.
- **Outlink** (vendor.site on smartphone visit): direct quotes preserved, entities not converted.
- **Referrer** (tablet visit, google query): direct quotes (' and ") are preserved, entities are converted.
- **Campaign name** (Visitor Z4F66G776HGI on next page): direct quotes preserved, entities are converted.

So the behavior for pre-encoded entities is inconsistent. Preservation of quote symbols is working everywhere, eg. ' and " will be shown properly. If the the tracker is passed an already encoded entity (eg. %27), should we be converting it to a quote or attempt to preserve exactly what was passed?

What do you think?

plugins/Live/tests/Fixtures/VisitsWithAllActionsAndDevices.php
```diff
@@ -32,13 +33,20 @@ public function setUp(): void
         CustomDimensionsApi::getInstance()->configureNewCustomDimension(1, 'currency', 'action', 1);
 
         $t = self::getTracker($this->idSite, $this->dateTime, $defaultInit = true);
+
+        // Campaign - needs a separate user otherwise it overrides the referrer
         $t->setTokenAuth(self::getTokenAuth());
+        $t->setUserId('Z4F66G776HGI');
+        $t->setVisitorId(substr(sha1('Z4F66G776HGI'), 0, 16));
```
sgiehl, line 39: If you want to create a new visitor the simplest way is to recreate the tracker instance with
`$t = self::getTracker($this->idSite, $this->dateTime, $defaultInit = true);`
That way all other possibly set parameters will also be unset.

justinvelluppillai:
@sgiehl it'd be good to know if you see any issues left on this needing fixed or what needs to be done to get it merged as well as the useful smaller tips.

sgiehl (approved):
Seems that actually works as expected. It took me a while to figure out why it doesn't work for campaigns locally. 🤯 It's actually the MarketingCampaignsReporting plugin that I have active, as it overwrites the campaign in that case. Would you mind checking how to fix that for that plugin as well, and maybe create a pull request there as well? Guess it might need to be added here: https://github.com/matomo-org/plugin-MarketingCampaignsReporting/blob/c7d437e9bf4232b29ec5e1f16dcd21dd457bec7e/VisitorDetails.php#L52

plugins/Live/tests/Fixtures/VisitsWithAllActionsAndDevices.php
bx80: Thanks, I'll make a note of that.

sgiehl:
@bx80 be good not to update a submodule to a specific branch without mentioning it. After merging the plugin PR, the submodule would now be broken as the reference doesn't exist anymore. Will update the submodule again and merge the PR if tests are passing afterwards

sgiehl:
failing tests seem to be random failures. Will merge this now 👍

PR #18179, opened 2021-10-19
update big int on cookie expire time to datatime format

### LINKED ISSUE #18019: Make the opt out work when a PHP 32 bit is used

peterhashair:
### Description:

Fixes: #18019 

 add a function in cookies. it accepts timestamp (int) and string eg: `+ 30 years` to avoid big int in PHP 32 bit

### Review


core/Cookie.php
```diff
@@ -468,4 +477,18 @@ private static function getSameSiteValueForBrowser($default)
 
         return $sameSite;
     }
+    /**
+     *  extend Cookie by years, by pass, default 30 years php 32 bit, max 100 years
+     * @param int $years
+     * @throws \Exception
+     */
+    public function extendExpireByYears($years = 30)
```
sgiehl: This name might be a bit misleading. Based on the name I would expect that an existing expire time would be extended if already set. But the method actually sets the expire time to a date X years in the future.

core/Cookie.php
```diff
@@ -468,4 +471,27 @@ private static function getSameSiteValueForBrowser($default)
...
+        try {
+            $expireTime = new DateTime();
+            if (is_int($time)) {
+                $expireTime->setTimestamp($time);
+            } else {
+                $expireTime->modify($time);
+            }
+        } catch (\Exception $e) {
+            $expireTime->setTimestamp($this->getDefaultExpire());
+        }
```
sgiehl: You maybe should readd the checks for `null` and `<0`. Not sure if there are any cases where that happens, but the behavior changed...
`DateTime::modify(null)`, doesn't do anything and also doesn't throw an exception, so the expire time remains to the current time. Instead it should use the default I guess.
Same applies for a negative value. `DateTime::setTimestamp(-50)` doesn't throw an exception but returns a time far in the past, so the cookie would be removed I guess. Before the default expire time was used as well.

sgiehl:
Seems there are also various tests failing...

peterhashair:
@sgiehl sorry, that's my bad, it actually fired a warning instead of an error.

peterhashair:
Not sure that's the best approach, it actually throws a warning to an exception enters the catch. Then reset it back. 
[https://github.com/matomo-org/matomo/blob/393bfa78d4c6e14eb8a30c6989248b27c70712ac/core/Cookie.php#L478](https://github.com/matomo-org/matomo/blob/393bfa78d4c6e14eb8a30c6989248b27c70712ac/core/Cookie.php#L478)

core/Cookie.php
```diff
@@ -468,4 +467,27 @@ private static function getSameSiteValueForBrowser($default)
...
+    /**
+     *  extend Cookie by timestamp or sting like + 30 years, + 10 months, default 2 years
+     * @param $time
+     * @return string
+     */
+    public function formatExpireTime($time = null)
+    {
+        $expireTime = new DateTime();
+        try {
+            ErrorHandler::registerWarningException();
```
sgiehl: Personally I think it might be a bit over bloated to use a custom error handler to catch some special cases.
Even with the error handler the case of a negative value is still not handled as before and doesn't result in the default expiration time.
If you want to validate if the value given to `modify` was valid, simply check the return value. Afaik it should return a boolean that indicates that.

core/Tracker/IgnoreCookie.php
```diff
@@ -50,21 +50,21 @@ public static function deleteThirdPartyCookieUIDIfExists()
...
     {
         $cookie_name = @Config::getInstance()->Tracker['ignore_visits_cookie_name'];
         $cookie_path = @Config::getInstance()->Tracker['cookie_path'];
-        
-        $thiryYears = time() + (86400 * 365 * 30);
 
-        $cookie = new Cookie($cookie_name, $thiryYears, $cookie_path, false);
-        
+
+        $cookie = new Cookie($cookie_name, "+ 30 years", $cookie_path, false);
```
tsteur, line 61: @peterhashair  we can maybe simplify this a bit like pass the date we want to set here like in https://3v4l.org/RB5Lp

```php
$date = gmdate('D, d-M-Y H:i:s', time()) . ' GMT';
$year = gmdate('Y', time());
echo str_replace($year, $year+30, $date);
```

Then in the `formatExpireTime` could be like 

```php
if (is_numeric($expire)) {
  // old logic
 $expire = gmdate('D, d-M-Y H:i:s', (int) $Expires) . ' GMT';
} 
 // in other cases we already get the full date
```

not sure that makes sense but could work maybe?

tsteur: Please ignore the comment if you can make it work otherwise as @sgiehl suggested to remove the error handler and checking return value or so

tests/PHPUnit/Unit/Tracker/RequestTest.php
```diff
@@ -8,6 +8,7 @@
 
 namespace Piwik\Tests\Unit\Tracker;
 
+use DateTime;
```
sgiehl: Guess the changes to this file are actually unneeded?

core/Cookie.php
```diff
@@ -468,4 +467,25 @@ private static function getSameSiteValueForBrowser($default)
...
+        $expireTime = new DateTime();
+        try {
+            if (is_int($time)) {
+                $expireTime->setTimestamp($time);
+            } else {
+                $expireTime->modify($time);
+            }
+        } catch (\Exception $e) {
+            $expireTime->modify("+2 years");
+        }
```
sgiehl: Guess this still doesn't have the same behavior as before. Maybe something like this would work better:

```php
if (is_null($time) || (is_int($time) && $time < 0)) {
    $expireTime->modify("+2 years");
} else if (is_int($time)) {
    $expireTime->setTimestamp($time);
} else if (!$expireTime->modify($time)) {
    $expireTime = new DateTime(); // might be unneeded, not sure if modify might change the object even if it fails...
    $expireTime->modify("+2 years");
}
```

sgiehl (changes requested):

peterhashair:
@sgiehl that's looks great, let me update it.

tests/PHPUnit/Unit/Tracker/RequestTest.php
sgiehl:
```suggestion
```

core/Cookie.php
```diff
@@ -8,7 +8,10 @@
  */
 namespace Piwik;
 
+use DateTime;
+use ErrorException;
 use Piwik\Container\StaticContainer;
+use Piwik\Exception\Exception;
```
sgiehl:
```suggestion
use Piwik\Container\StaticContainer;
```

sgiehl (approved):
Code looks good to merge for me now. Will wait for the tests before merging

PR #18195, opened 2021-10-21
Fixing unsafe comparison warnings (`==` => `===`, `!=` => `!==`)

hywak:
### Description:

https://github.com/matomo-org/matomo/issues/17686

Lot's of files are changed, but those are cosmetics changes only.

### Review

* [x] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)


core/API/DataTablePostProcessor.php
```diff
@@ -375,7 +375,7 @@ public function applyLabelFilter($dataTable)
     public function applyMetricsFormatting($dataTable)
     {
         $formatMetrics = Common::getRequestVar('format_metrics', 0, 'string', $this->request);
-        if ($formatMetrics == '0') {
+        if ($formatMetrics === '0') {
```
sgiehl, line 378: This might change the logic, unless we change the second parameter in the line above to `'0'`

core/API/DocumentationGenerator.php
```diff
@@ -191,7 +191,7 @@ private function checkIfCommentContainsInternalAnnotation($rClass)
      */
     public function checkDocumentation($moduleToCheck)
     {
-        if (strpos($moduleToCheck, '@hide') == true) {
+        if (strpos($moduleToCheck, '@hide') === true) {
```
sgiehl: This needs to be changed as well. `strpos` actually never returns true. Might be more correct to check for `!== false` instead

core/API/DataTableManipulator/LabelFilter.php
```diff
@@ -150,9 +150,7 @@ private function getLabelVariations($originalLabel)
         $sanitizedLabel = Common::sanitizeInputValue($label);
         $variations[] = $sanitizedLabel;
 
-        if ($this->apiModule == 'Actions'
-            && in_array($this->apiMethod, $pageTitleReports)
-        ) {
+        if ($this->apiModule == 'Actions' && in_array($this->apiMethod, $pageTitleReports)) {
```
sgiehl:
```suggestion
        if ($this->apiModule === 'Actions' && in_array($this->apiMethod, $pageTitleReports)) {
```

core/ArchiveProcessor/Loader.php
```diff
@@ -485,7 +485,7 @@ private function shouldForceInvalidatedArchive($value, $tsArchived)
         $params = $this->params;
 
         // the archive is invalidated and we are in a browser request that is allowed archive it
-        if ($value == ArchiveWriter::DONE_INVALIDATED
+        if ($value === ArchiveWriter::DONE_INVALIDATED
```
sgiehl: Looking at the code it might happen tat `$value` could be of type string, as it is returns from a database query. Therefore you might need to cast it to int before doing that kind of compare.

core/AssetManager.php
```diff
@@ -282,11 +282,11 @@ public function getAssetDirectory()
      */
     public function isMergedAssetsDisabled()
     {
-        if (Config::getInstance()->Development['disable_merged_assets'] == 1) {
+        if (Config::getInstance()->Development['disable_merged_assets'] === 1) {
```
sgiehl: Values returned from Config might be of type string.

core/AssetManager.php
```diff
@@ -282,11 +282,11 @@ public function getAssetDirectory()
...
             return true;
         }
         
-        if (isset($_GET['disable_merged_assets']) && $_GET['disable_merged_assets'] == 1) {
+        if (isset($_GET['disable_merged_assets']) && $_GET['disable_merged_assets'] === 1) {
```
sgiehl: I guess GET parameters might be of type string as well

core/CliMulti.php
```diff
@@ -390,7 +390,7 @@ private function executeNotAsyncHttp($url, StaticOutput $output)
         }
 
         $url = $piwikUrl . $url;
-        if (Config::getInstance()->General['force_ssl'] == 1) {
+        if (Config::getInstance()->General['force_ssl'] === 1) {
```
sgiehl: Config values might be strings

core/Concurrency/LockBackend/MySqlLockBackend.php
```diff
@@ -66,7 +66,7 @@ public function setIfNotExists($key, $value, $ttlInSeconds)
         try {
             Db::query($query, array($key, $value, (int) $ttlInSeconds));
         } catch (\Exception $e) {
-            if ($e->getCode() == 23000
+            if ($e->getCode() === 23000
```
sgiehl: Not fully sure if those database error codes are always of type int.

core/API/DataTablePostProcessor.php
hywak: Good catch. Changed

core/CronArchive.php
```diff
@@ -227,7 +227,7 @@ public function __construct(LoggerInterface $logger = null)
 
         $this->invalidator = StaticContainer::get('Piwik\Archive\ArchiveInvalidator');
 
-        $this->isArchiveProfilingEnabled = Config::getInstance()->Debug['archiving_profile'] == 1;
+        $this->isArchiveProfilingEnabled = Config::getInstance()->Debug['archiving_profile'] === 1;
```
sgiehl: Config values might be strings

core/CronArchive.php
```diff
@@ -243,7 +243,7 @@ public function __construct(LoggerInterface $logger = null)
 
     private function isMaintenanceModeEnabled()
     {
-        return Config::getInstance()->General['maintenance_mode'] == 1;
+        return Config::getInstance()->General['maintenance_mode'] === 1;
```
sgiehl: Config values might be strings

core/API/DocumentationGenerator.php
hywak: Changed as well.

core/DataAccess/Model.php
```diff
@@ -86,7 +86,7 @@ public function getInvalidatedArchiveIdsSafeToDelete($archiveTable, $setGroupCon
 
                 array_shift($duplicateArchives);
 
-                if ($value != ArchiveWriter::DONE_PARTIAL) {
+                if ($value !== ArchiveWriter::DONE_PARTIAL) {
```
sgiehl: Values returned from database queries might be strings or integers depending on the use database adapter

core/DataArray.php
```diff
@@ -393,7 +393,7 @@ protected function enrichWithConversions(&$data)
             $values[Metrics::INDEX_NB_CONVERSIONS] = $conversions;
 
             // 25.00 recorded as 25
-            if (round($revenue) == $revenue) {
+            if (round($revenue) === $revenue) {
```
sgiehl: I guess this won't work in all cases. `$revenue` is defined as Integer and as long as no float values will be added, it might remain as integer. calling `round` on an integer will return a float. And comparing a float with an integer holding the same value would still return false.

core/DataTable/Filter/Pattern.php
```diff
@@ -72,7 +72,7 @@ public static function getPatternQuoted($pattern)
      */
     public static function match($patternQuoted, $string, $invertedMatch = false)
     {
-        return preg_match($patternQuoted . "i", $string) == 1 ^ $invertedMatch;
+        return preg_match($patternQuoted . "i", $string) === 1 ^ $invertedMatch;
```
sgiehl: Even though it shouldn't happen, this compare won't any longer handle the error case when `preg_match` returns false.

core/Date.php
```diff
@@ -1010,7 +1010,7 @@ public static function addHourTo($timestamp, $n)
     {
         $isNegative = ($n < 0);
         $minutes = 0;
-        if ($n != round($n)) {
+        if ($n !== round($n)) {
```
sgiehl: This will always be true when `$n` is given as `int` as `round` always returns `float`

core/Db.php
```diff
@@ -821,7 +821,7 @@ private static function logSql($functionName, $sql, $parameters = array())
         self::checkBoundParametersIfInDevMode($sql, $parameters);
 
         if (self::$logQueries === false
-            || @Config::getInstance()->Debug['log_sql_queries'] != 1
+            || @Config::getInstance()->Debug['log_sql_queries'] !== 1
```
sgiehl: Config value might be strings or ints

core/Db.php
```diff
@@ -744,7 +744,7 @@ public static function getDbLock($lockName, $maxRetries = 30)
 
         while ($maxRetries > 0) {
             $result = $db->fetchOne($sql, array($lockName));
-            if ($result == '1') {
+            if ($result === '1') {
```
sgiehl: Values coming from database queries might be strings or ints depending on the database adapter

core/Db.php
```diff
@@ -764,7 +764,7 @@ public static function releaseDbLock($lockName)
         $sql = 'SELECT RELEASE_LOCK(?)';
 
         $db = self::get();
-        return $db->fetchOne($sql, array($lockName)) == '1';
+        return $db->fetchOne($sql, array($lockName)) === '1';
```
sgiehl: same as above

core/Db/Adapter/Mysqli.php
```diff
@@ -199,12 +199,12 @@ public static function isMysqliErrorNumber($e, $connection, $errno)
     {
         if (is_null($connection)) {
             if (preg_match('/(?:\[|\s)([0-9]{4})(?:\]|\s)/', $e->getMessage(), $match)) {
-                return $match[1] == $errno;
+                return $match[1] === $errno;
```
sgiehl: This and the compare below will only work if `$errno` is given as string. Not sure if this might be called with an integer at some places.

core/Db/Adapter/Pdo/Mysql.php
```diff
@@ -223,7 +223,7 @@ public function isErrNo($e, $errno)
     public static function isPdoErrorNumber($e, $errno)
     {
         if (preg_match('/(?:\[|\s)([0-9]{4})(?:\]|\s)/', $e->getMessage(), $match)) {
-            return $match[1] == $errno;
+            return $match[1] === $errno;
```
sgiehl: Depending of given `$errno` the types might differ here, too.

core/FileIntegrity.php
```diff
@@ -323,7 +323,7 @@ protected static function getMessagesFilesMismatch($messages)
 
             if (!file_exists($file) || !is_readable($file)) {
                 $messagesMismatch[] = Piwik::translate('General_ExceptionMissingFile', $file);
-            } elseif (filesize($file) != $props[0]) {
+            } elseif (filesize($file) !== $props[0]) {
```
sgiehl: The Manifest file contains the filesize as strings, so this compare will fail

core/FileIntegrity.php
```diff
@@ -336,7 +336,7 @@ protected static function getMessagesFilesMismatch($messages)
                     // convert end-of-line characters and re-test text files
                     $content = @file_get_contents($file);
                     $content = str_replace("\r\n", "\n", $content);
-                    if ((strlen($content) != $props[0])
+                    if ((strlen($content) !== $props[0])
```
sgiehl: same as above

core/FrontController.php
```diff
@@ -486,7 +485,7 @@ protected function prepareDispatch($module, $action, $parameters)
 
     protected function handleMaintenanceMode()
     {
-        if ((Config::getInstance()->General['maintenance_mode'] != 1) || Common::isPhpCliMode()) {
+        if ((Config::getInstance()->General['maintenance_mode'] !== 1) || Common::isPhpCliMode()) {
```
sgiehl: Config values might be strings

core/FrontController.php
```diff
@@ -565,13 +564,13 @@ private function closeSessionEarlyForFasterUI()
 
     private function handleProfiler()
     {
-        $profilerEnabled = Config::getInstance()->Debug['enable_php_profiler'] == 1;
+        $profilerEnabled = Config::getInstance()->Debug['enable_php_profiler'] === 1;
```
sgiehl: Config values might be strings

core/FrontController.php
```diff
@@ -565,13 +564,13 @@ private function closeSessionEarlyForFasterUI()
...
         if (!$profilerEnabled) {
             return;
         }
 
         if (!empty($_GET['xhprof'])) {
-            $mainRun = $_GET['xhprof'] == 1; // core:archive command sets xhprof=2
+            $mainRun = $_GET['xhprof'] === 1; // core:archive command sets xhprof=2
```
sgiehl: GET parameters might be strings

core/FrontController.php
```diff
@@ -787,7 +786,7 @@ private function isSupportedBrowserCheckNeeded()
         }
 
         $generalConfig = Config::getInstance()->General;
-        if ($generalConfig['enable_framed_pages'] == '1' || $generalConfig['enable_framed_settings'] == '1') {
+        if ($generalConfig['enable_framed_pages'] === '1' || $generalConfig['enable_framed_settings'] === '1') {
```
sgiehl: Config values can be string or ints.

core/Http.php
```diff
@@ -32,7 +32,7 @@ public static function getTransportMethod()
         $method = 'curl';
         if (!self::isCurlEnabled()) {
             $method = 'fopen';
-            if (@ini_get('allow_url_fopen') != '1') {
+            if (@ini_get('allow_url_fopen') !== '1') {
```
sgiehl, line 35: Not sure if `ini_get` may also return an integer 🤔

core/Http.php
```diff
@@ -775,9 +773,7 @@ public static function sendHttpRequestBy(
             @fclose($file);
 
             $fileSize = filesize($destinationPath);
-            if ($contentLength > 0
-                && $fileSize != $contentLength
-            ) {
+            if ($contentLength > 0 && $fileSize !== $contentLength) {
```
sgiehl, line 776: `$contentLength` might come from `@curl_getinfo($ch, CURLINFO_CONTENT_LENGTH_DOWNLOAD)`, not sure if that always returns an integer like `filesize`

core/IP.php
```diff
@@ -78,7 +78,7 @@ public static function getNonProxyIpFromHeader($default, $proxyHeaders)
             $proxyIps = array();
         }
 
-        $shouldReadLastProxyIp = Config::getInstance()->General['proxy_ip_read_last_in_list'] == 1;
+        $shouldReadLastProxyIp = Config::getInstance()->General['proxy_ip_read_last_in_list'] === 1;
```
sgiehl: Config values can be strings or ints

core/Nonce.php
```diff
@@ -197,7 +197,7 @@ public static function getAcceptableOrigins()
                 'http://' . $host,
                 'https://' . $host,
             );
-            if ($port != 443) {
+            if ($port !== 443) {
```
sgiehl: `$port` is the result from a `preg_match` and therefor likely is a string

core/NumberFormatter.php
```diff
@@ -149,7 +147,7 @@ public function formatCurrency($value, $currency, $precision=2)
 
         $pattern = $this->getPattern($value, 'Intl_NumberFormatCurrency');
 
-        if ($newValue == round($newValue)) {
+        if ($newValue === round($newValue)) {
```
sgiehl: the types might differ here, causing different results as before.

core/Period/Month.php
```diff
@@ -83,15 +83,15 @@ protected function generate()
     protected function processOptimalSubperiods($startDate, $endDate)
     {
         while ($startDate->isEarlier($endDate)
-            || $startDate == $endDate
+            || $startDate === $endDate
```
sgiehl: I don't think that will work, as both variables are an instance of `Piwik\Date` but aren't the same instance. This might work though
```suggestion
            || $startDate->toString() === $endDate->toString()
```

core/Period/Month.php
```diff
@@ -83,15 +83,15 @@ protected function generate()
...
         ) {
             $week        = new Week($startDate);
             $startOfWeek = $week->getDateStart();
             $endOfWeek   = $week->getDateEnd();
 
             if ($endOfWeek->isLater($endDate)) {
                 $this->fillDayPeriods($startDate, $endDate);
-            } elseif ($startOfWeek == $startDate) {
+            } elseif ($startOfWeek === $startDate) {
```
sgiehl: same here:
```suggestion
            } elseif ($startOfWeek->toString() === $startDate->toString()) {
```

core/Period/Month.php
```diff
@@ -109,7 +109,7 @@ protected function processOptimalSubperiods($startDate, $endDate)
      */
     private function fillDayPeriods($startDate, $endDate)
     {
-        while (($startDate->isEarlier($endDate) || $startDate == $endDate)) {
+        while (($startDate->isEarlier($endDate) || $startDate === $endDate)) {
```
sgiehl: here as well:
```suggestion
        while (($startDate->isEarlier($endDate) || $startDate->toString() === $endDate->toString())) {
```

core/Piwik.php
```diff
@@ -128,10 +128,10 @@ public static function getPercentageSafe($dividend, $divisor, $precision = 0)
      */
     public static function getQuotientSafe($dividend, $divisor, $precision = 0)
     {
-        if ($divisor == 0) {
+        if ($divisor === 0) {
```
sgiehl: I'm not sure here if `$divisor` might not be a numeric string in some cases.
```suggestion
        if ((int) $divisor === 0) {
```

core/Piwik.php
```diff
@@ -128,10 +128,10 @@ public static function getPercentageSafe($dividend, $divisor, $precision = 0)
...
             return 0;
         }
-        if ($dividend == 0 || $dividend === '-') {
+        if ($dividend === 0 || $dividend === '-') {
```
sgiehl:
```suggestion
        if ((int) $dividend === 0 || $dividend === '-') {
```

core/Plugin/ControllerAdmin.php
```diff
@@ -36,7 +36,7 @@ abstract class ControllerAdmin extends Controller
     private static function notifyWhenTrackingStatisticsDisabled()
     {
         $statsEnabled = PiwikConfig::getInstance()->Tracker['record_statistics'];
-        if ($statsEnabled == "0") {
+        if ($statsEnabled === "0") {
```
sgiehl: A config value might also be an integer.

core/ProxyHeaders.php
```diff
@@ -21,19 +21,19 @@ class ProxyHeaders
      */
     public static function getProtocolInformation()
     {
-        if (isset($_SERVER['SERVER_PORT']) && $_SERVER['SERVER_PORT'] == 443) {
+        if (isset($_SERVER['SERVER_PORT']) && $_SERVER['SERVER_PORT'] === 443) {
```
sgiehl: I guess `$_SERVER['SERVER_PORT']` is more likely returned as a string, but not sure if can also be an integer.

core/Settings/Storage/Backend/MeasurableSettingsTable.php
```diff
@@ -168,7 +168,7 @@ public static function removeAllSettingsForSite($idSite)
             $query = sprintf('DELETE FROM %s WHERE idsite = ?', Common::prefixTable('site_setting'));
             Db::query($query, array($idSite));
         } catch (Exception $e) {
-            if ($e->getCode() != 42) {
+            if ($e->getCode() !== 42) {
```
sgiehl: Not sure if database exception codes might not be strings instead.

core/Tracker/Model.php
```diff
@@ -144,7 +144,7 @@ public function createEcommerceItems($ecommerceItems)
         try {
             $this->getDb()->query($sql, $bind);
         } catch (Exception $e) {
-            if ($e->getCode() == 23000 ||
+            if ($e->getCode() === 23000 ||
```
sgiehl: Database exception codes might possibly also be strings...

sgiehl (changes requested):
Left a couple of comments. In general it might not be that easy to change `==` to `===` in all cases. Unfortunately our travis tests are currently not running as expected. As soon as they will work again we can trigger a new build to see how many tests are failing...

core/Db.php
hywak: @sgiehl Do you think it's safe to cast the result to string and compare `(string) $result === '1'` ?

core/AssetManager.php
hywak: @sgiehl Correct me if I'm wrong. It should be enough If I cast Config value to a string and compare it with the string `'1'`?

sgiehl: Yes, I guess it should be enough that way.

core/Tracker/Model.php
hywak: Changing condition to `if ((string) $e->getCode() === '23000'`

core/Settings/Storage/Backend/MeasurableSettingsTable.php
hywak: Changing condition to `if ((string) $e->getCode() !== '42') {`

core/ProxyHeaders.php
hywak: Changing condition to `(string) $_SERVER['SERVER_PORT'] === '443'`

core/Plugin/ControllerAdmin.php
hywak: Changed to `if ((string) $statsEnabled === '0') {`

core/CliMulti.php
hywak: Changed to `if ((string) Config::getInstance()->General['force_ssl'] === '1')`

core/AssetManager.php
hywak: Changing to `if ((string) Config::getInstance()->Development['disable_merged_assets'] === '1') {`

core/AssetManager.php
hywak: Changing to `(string) $_GET['disable_merged_assets'] === '1')`

core/Concurrency/LockBackend/MySqlLockBackend.php
hywak: Changed to `if ((string) $e->getCode() === '23000'`

core/CronArchive.php
hywak: Changed to `$this->isArchiveProfilingEnabled = (string) Config::getInstance()->Debug['archiving_profile'] === '1';`

core/CronArchive.php
hywak: Changed to `return (string) Config::getInstance()->General['maintenance_mode'] === '1';`

core/DataAccess/Model.php
hywak: Casting to int `if ((int) $value !== ArchiveWriter::DONE_PARTIAL) {` as `DONE_PARTIAL` value is `5`

core/Db.php
sgiehl: yes. should be

core/DataArray.php
hywak: It seems like this condition is to detect whether `$revenue` is int or float, depending on the value it's forced to be float. 🤔

core/Db.php
hywak: Changed to `@(string) Config::getInstance()->Debug['log_sql_queries'] !== '1') {`

core/Db.php
hywak: Changing to `if ((string) $result === '1') {`

core/Db.php
hywak: Changed to `return (string) $db->fetchOne($sql, [$lockName]) === '1';`

core/Db/Adapter/Mysqli.php
hywak: It seems like it can be. `mysqli_connect_errno()` and `mysqli_errno` return int, while `preg_match` returns string.  

I'm wondering whether below casting wouldn't do the job?
```        if (is_null($connection)) {
            if (preg_match('/(?:\[|\s)([0-9]{4})(?:\]|\s)/', $e->getMessage(), $match)) {
                return $match[1] === (string) $errno;
            }

            return mysqli_connect_errno() === (int) $errno;
        }

        return mysqli_errno($connection) === (int) $errno;
```

core/Db/Adapter/Pdo/Mysql.php
hywak: Changing to `            return $match[1] === (string) $errno;`

core/DataArray.php
sgiehl: Actually I'm not sure what that check is supposed to do. Guess simply casting `$revenue` to `float` would do the same, wouldn't it? 🤔

core/FrontController.php
hywak: Changed to `$profilerEnabled = (string) Config::getInstance()->Debug['enable_php_profiler'] === '1';`

core/FrontController.php
hywak: Changed to `$mainRun = (string) $_GET['xhprof'] === '1';`

core/FrontController.php
hywak: Changed to
 ```
        if ((string) $generalConfig['enable_framed_pages'] === '1' 
            || (string) $generalConfig['enable_framed_settings'] === '1'
        ) {
```

File: core/IP.php (hunk 28 above)
Comment (hywak) [reply core/IP.php]: Changed to `$shouldReadLastProxyIp = (string) Config::getInstance()->General['proxy_ip_read_last_in_list'] === '1';`

File: core/Nonce.php (hunk 29 above)
Comment (hywak) [reply core/Nonce.php]: changed to `if ((string) $port !== '443') {`

File: core/DataArray.php (hunk 12 above)
Comment (hywak) [reply core/DataArray.php]: `$revenue` field is being set to 0 in line 385. It's being modified later in line 389, theoretically, it can be either int or float depending on `$goalValues[Metrics::INDEX_GOAL_REVENUE];` result. So it seems like `round($revenue)` is done just for making sure that `$revenue` fill be float at the end.

But this condition doesn't make sense to me even before my change.
```
var_dump(round(25.00)); // float(25)

var_dump(round(25)); // float(25)
```

Comment (sgiehl) [reply core/DataArray.php]: `$goalValues[Metrics::INDEX_GOAL_REVENUE];` might also hold integer or float values I guess. Logically it seems that condition is nothing else than a "complexe" cast.
There can actually only be this two cases:
- `$revenue` is an integer like `18`. `round($revenue) == $revenue` would be `true` and `$revenue` would be changed to a `float` (by using round).
- `$revenue` is an float like `12.34`. `round($revenue) == $revenue` would be `false` and `$revenue` would stay a `float`

So imho simply writing `$revenue = (float) $revenue;` should be the same.

File: core/Db.php [hunk 40]
```diff
@@ -820,9 +820,7 @@ private static function logSql($functionName, $sql, $parameters = array())
     {
         self::checkBoundParametersIfInDevMode($sql, $parameters);
 
-        if (self::$logQueries === false
-            || @Config::getInstance()->Debug['log_sql_queries'] != 1
-        ) {
+        if (self::$logQueries === false || @(string) Config::getInstance()->Debug['log_sql_queries'] !== '1') {
```
sgiehl: I guess the `@` shouldn't be needed though

core/Db/Adapter/Mysqli.php
sgiehl: should be fine that way 👍

core/Settings/Storage/Backend/MeasurableSettingsTable.php
```diff
@@ -186,7 +186,7 @@ public static function removeAllSettingsForPlugin($pluginName)
             $query = sprintf('DELETE FROM %s WHERE plugin_name = ?', Common::prefixTable('site_setting'));
             Db::query($query, array($pluginName));
         } catch (Exception $e) {
-            if ($e->getCode() != 42) {
+            if ($e->getCode() !== 42) {
```
sgiehl:
```suggestion
            if ((string) $e->getCode() !== '42') {
```

core/Settings/Storage/Backend/PluginSettingsTable.php
```diff
@@ -174,7 +174,7 @@ public static function removeAllUserSettingsForUser($userLogin)
             $table = Common::prefixTable('plugin_setting');
             Db::get()->query(sprintf('DELETE FROM %s WHERE user_login = ?', $table), array($userLogin));
         } catch (Exception $e) {
-            if ($e->getCode() != 42) {
+            if ($e->getCode() !== 42) {
```
sgiehl:
```suggestion
            if ((string) $e->getCode() !== '42') {
```

core/Settings/Storage/Backend/PluginSettingsTable.php
```diff
@@ -195,7 +195,7 @@ public static function removeAllSettingsForPlugin($pluginName)
             $table = Common::prefixTable('plugin_setting');
             Db::get()->query(sprintf('DELETE FROM %s WHERE plugin_name = ?', $table), array($pluginName));
         } catch (Exception $e) {
-            if ($e->getCode() != 42) {
+            if ($e->getCode() !== 42) {
```
sgiehl:
```suggestion
            if ((string) $e->getCode() !== '42') {
```

core/Tracker/Model.php
```diff
@@ -422,7 +422,7 @@ public function hasVisit($idSite, $idVisit)
         $bindSql = array($idVisit);
 
         $val = $this->getDb()->fetchOne($sql, $bindSql);
-        return $val == $idSite;
+        return $val === $idSite;
```
sgiehl: Database query might return a string, while the value given to that method could actually be string or int I guess
```suggestion
        return (string) $val === (string) $idSite;
```

core/DataArray.php
hywak: Thinking the same. Changing to `$values[Metrics::INDEX_REVENUE] = (float) $revenue;`

core/Db.php
hywak: Removing then

core/Tracker/GoalManager.php
```diff
@@ -724,7 +724,7 @@ protected function recordStandardGoals(VisitProperties $visitProperties, Request
             }
 
             // If multiple Goal conversions per visit, set a cache buster
-            if ($convertedGoal['allow_multiple'] == 0) {
+            if ($convertedGoal['allow_multiple'] === 0) {
```
sgiehl: Not sure if we can assume this one to be an integer
```suggestion
            if ((int) $convertedGoal['allow_multiple'] === 0) {
```

core/Tracker/GoalManager.php
```diff
@@ -321,7 +321,7 @@ public function recordGoals(VisitProperties $visitProperties, Request $request)
      */
     protected function getRevenue($revenue)
     {
-        if (round($revenue) != $revenue) {
+        if (round($revenue) !== $revenue) {
```
sgiehl: I guess this is another check if we have fraction digits. To also handle integers correctly I think it would need to be
```suggestion
        if (round($revenue) !== (float) $revenue) {
```

core/Tracker/GoalManager.php
```diff
@@ -99,7 +99,7 @@ public static function getGoalDefinition($idSite, $idGoal)
         $goals = self::getGoalDefinitions($idSite);
 
         foreach ($goals as $goal) {
-            if ($goal['idgoal'] == $idGoal) {
+            if ($goal['idgoal'] === $idGoal) {
```
sgiehl: $goals is coming from database, so the values might be integer or string. Same applies for the value given to that method:
```suggestion
            if ((int) $goal['idgoal'] === (int) $idGoal) {
```

core/FileIntegrity.php
hywak: Casting `filesize($file)` to string should be safe, ` } elseif ((string) filesize($file) !== $props[0]) {`

core/FrontController.php
hywak: Changing to `if ((string) (Config::getInstance()->General['maintenance_mode'] !== '1') || Common::isPhpCliMode()) {`

core/FileIntegrity.php
hywak: Changing to `if ((strlen($content) !== (int) $props[0]) || (@md5($content) !== $props[1])) {`

core/Http.php
hywak: According to documentation it should be string or false https://www.php.net/manual/en/function.ini-get.php. I can cat it to be 100% sure, what do you think?

core/Profiler.php
```diff
@@ -193,7 +193,7 @@ private static function getSqlProfilingQueryBreakdownOutput($infoIndexedByQuery)
                 $avgTimeString = " (average = <b>" . round($avgTimeMs, 1) . "ms</b>)";
             }
             $query = preg_replace('/([\t\n\r ]+)/', ' ', $query);
-            $output .= "Executed <b>$count</b> time" . ($count == 1 ? '' : 's') . " in <b>" . $timeMs . "ms</b> $avgTimeString <pre>\t$query</pre>";
+            $output .= "Executed <b>$count</b> time" . ($count === 1 ? '' : 's') . " in <b>" . $timeMs . "ms</b> $avgTimeString <pre>\t$query</pre>";
```
sgiehl, line 196: Not sure if `$count` is always an integer. Might be good to simply cast it a few lines above.

core/NumberFormatter.php
sgiehl:
```suggestion
        if ((float) $newValue === round($newValue)) {
```

core/Http.php
sgiehl:
```suggestion
            if ($contentLength > 0 && $fileSize !== (int) $contentLength) {
```

core/FrontController.php
sgiehl:
```suggestion
        if (((string) Config::getInstance()->General['maintenance_mode'] !== '1') || Common::isPhpCliMode()) {
```

core/FileIntegrity.php
sgiehl:
```suggestion
            } elseif (filesize($file) !== (int) $props[0]) {
```

core/FileIntegrity.php
sgiehl:
```suggestion
                    if ((strlen($content) !== (int) $props[0])
```

core/Date.php
sgiehl:
```suggestion
        if ((float) $n !== round($n)) {
```

core/DataTable/Filter/ExcludeLowPopulation.php
```diff
@@ -68,7 +68,7 @@ public function __construct($table, $columnToFilter, $minimumValue, $minimumPerc
 
         $this->columnToFilter = $this->selectColumnToExclude($columnToFilter, $row);
 
-        if ($minimumValue == 0) {
+        if ($minimumValue === 0) {
```
sgiehl: based on the doc block `$minimumValue` can be `false`. This case was covered here as well. So guess this would be better:
```suggestion
        if ((int) $minimumValue === 0) {
```

core/DataTable/Filter/CalculateEvolutionFilter.php
```diff
@@ -191,9 +191,9 @@ public static function prependPlusSignToNumber($number)
      */
     private static function getPercentageValue($value, $divisor, $quotientPrecision)
     {
-        if ($value == 0) {
+        if ($value === 0) {
```
sgiehl: I guess we can't assume `$value` to be an integer. Could also be a `float`.
```suggestion
        if ((float) $value === 0.0) {
```

Note: Casting to integer instead would be incorrect as `(int) 0.25` would result in `0`

core/DataTable/Filter/CalculateEvolutionFilter.php
```diff
@@ -191,9 +191,9 @@ public static function prependPlusSignToNumber($number)
...
             $evolution = 0;
-        } elseif ($divisor == 0) {
+        } elseif ($divisor === 0) {
```
sgiehl:
```suggestion
        } elseif ((float) $divisor === 0.0) {
```

core/Http.php
hywak: $contentLength is being casted to `integer` in `core/Http.php:511`. I'm adding casting in `core/Http.php:761` as well `$contentLength = (int) @curl_getinfo($ch, CURLINFO_CONTENT_LENGTH_DOWNLOAD);`

core/DataArray.php
```diff
@@ -393,7 +393,7 @@ protected function enrichWithConversions(&$data)
...
                 $revenue = round($revenue);
             }
```
sgiehl:
```suggestion
            $revenue = (float) $revenue;
```

core/DataAccess/ArchiveWriter.php
```diff
@@ -342,7 +342,7 @@ private function checkDoneValueIsOnlyPartialForPluginArchives($doneValue)
     {
         // if the done flag is not like done%.PluginName, then it shouldn't be a partial archive.
         // log a warning.
-        if ($doneValue == self::DONE_PARTIAL && strpos($this->doneFlag, '.') == false) {
+        if ($doneValue === self::DONE_PARTIAL && strpos($this->doneFlag, '.') === false) {
```
sgiehl: `$doneValue` might be string though.

```suggestion
        if ((int) $doneValue === self::DONE_PARTIAL && strpos($this->doneFlag, '.') === false) {
```

core/DataAccess/ArchiveSelector.php
```diff
@@ -474,7 +474,7 @@ private static function findArchiveDataWithLatestTsArchived($results, $requested
             }
 
             $thisTsArchived = Date::factory($row['ts_archived']);
-            if ($row['value'] == ArchiveWriter::DONE_PARTIAL
+            if ($row['value'] === ArchiveWriter::DONE_PARTIAL
```
sgiehl:
```suggestion
            if ((int) $row['value'] === ArchiveWriter::DONE_PARTIAL
```

core/DataAccess/ArchiveSelector.php
```diff
@@ -439,7 +439,7 @@ private static function findArchiveDataWithLatestTsArchived($results, $requested
         foreach ($results as $result) {
             if (in_array($result['name'], $doneFlags)
                 && in_array($result['idarchive'], $idArchives)
-                && $result['value'] != ArchiveWriter::DONE_PARTIAL
+                && $result['value'] !== ArchiveWriter::DONE_PARTIAL
```
sgiehl:
```suggestion
                && (int) $result['value'] !== ArchiveWriter::DONE_PARTIAL
```

core/DataAccess/ArchiveSelector.php
```diff
@@ -234,7 +234,7 @@ public static function getArchiveIds($siteIds, $periods, $segment, $plugins, $in
                     if (strpos($doneFlag, '.') === false // all plugins archive
                         // sanity check: DONE_PARTIAL shouldn't be used w/ done archives, but in case we see one,
                         // don't treat it like an all plugins archive
-                        && $value != ArchiveWriter::DONE_PARTIAL
+                        && $value !== ArchiveWriter::DONE_PARTIAL
```
sgiehl:
```suggestion
                        && (int) $value !== ArchiveWriter::DONE_PARTIAL
```

core/NumberFormatter.php
hywak: Accepting your chnges

core/CronArchive/Performance/Logger.php
```diff
@@ -34,7 +34,7 @@ class Logger
 
     public function __construct(Config $config, LoggerInterface $logger = null)
     {
-        $this->isEnabled = $config->Debug['archiving_profile'] == 1;
+        $this->isEnabled = $config->Debug['archiving_profile'] === 1;
```
sgiehl:
```suggestion
        $this->isEnabled = (string) $config->Debug['archiving_profile'] === '1';
```

core/DataTable/Filter/Pattern.php
hywak: Reverting this change. I don't understand the logic here, don't want to mess up

core/API/DataTablePostProcessor.php
```diff
@@ -200,7 +200,7 @@ public function applyFlattener($dataTable)
      */
     public function applyTotalsCalculator($dataTable)
     {
-        if (1 == Common::getRequestVar('totals', '1', 'integer', $this->request)) {
+        if (1 === Common::getRequestVar('totals', '1', 'integer', $this->request)) {
```
sgiehl:
```suggestion
        if (1 === Common::getRequestVar('totals', 1, 'integer', $this->request)) {
```

sgiehl (changes requested):
Had another look through the changes.

core/Date.php
hywak: Accepting your changes

core/API/DataTablePostProcessor.php
```diff
@@ -214,7 +214,7 @@ public function applyTotalsCalculator($dataTable)
     public function applyGenericFilters($dataTable)
     {
         // if the flag disable_generic_filters is defined we skip the generic filters
-        if (0 == Common::getRequestVar('disable_generic_filters', '0', 'string', $this->request)) {
+        if (0 === Common::getRequestVar('disable_generic_filters', '0', 'string', $this->request)) {
```
sgiehl:
```suggestion
        if ('0' === Common::getRequestVar('disable_generic_filters', '0', 'string', $this->request)) {
```

core/API/DataTablePostProcessor.php
```diff
@@ -481,7 +481,7 @@ public function applyComputeProcessedMetrics(DataTableInterface $dataTable)
     public function applyComparison(DataTableInterface $dataTable)
     {
         $compare = Common::getRequestVar('compare', '0', 'int', $this->request);
```
sgiehl:
```suggestion
        $compare = Common::getRequestVar('compare', 0, 'int', $this->request);
```

core/Http.php
sgiehl: We can keep it as is for now.

core/Http.php
```diff
@@ -775,9 +773,7 @@ public static function sendHttpRequestBy(
             @fclose($file);
 
             $fileSize = filesize($destinationPath);
-            if ($contentLength > 0
-                && $fileSize != $contentLength
-            ) {
+            if ($contentLength > 0 && $fileSize !== (int) $contentLength) {
```
sgiehl: `$contentLength` is now casted before, so this is obsolete

```suggestion
            if ($contentLength > 0 && $fileSize !== $contentLength) {
```

sgiehl (changes requested):
Found a couple more things. But besides that it's looking fine. We may need to wait until mid of next week till the travis tests are running again. Once we had them running we can check if anything causes problems....

sgiehl:
@hywak would you mind merging in the latest changes from 4.x-dev branch. Travis builds should be up and running again, so we hopefully can see afterwards if all those changes break anything or not.

sgiehl:
@hywak Seems there are a plenty of tests failing. Might be good to first looking into why the Unit tests are failing, maybe fixing those might also fix some of the other test suites:
https://app.travis-ci.com/github/matomo-org/matomo/jobs/545181203
I'm not sure when I will find some time to have a quick look myself to maybe identify some first issues, but will try to do that soon.

core/Common.php
```diff
@@ -543,14 +543,14 @@ public static function getRequestVar($varName, $varDefault = null, $varType = nu
                     $ok    = true;
                 }
             } elseif ($varType === 'integer') {
-                if ($value == (string)(int)$value) {
+                if ($value === (string)(int)$value) {
```
sgiehl: Maybe this is causing some issues:
```suggestion
                if ((string)$value === (string)(int)$value) {
```

hywak:
@sgiehl I will take a look at it tomorrow or at the weekend :)

hywak:
@sgiehl Is there documentation on how to run tests locally?

sgiehl:
@hywak sure. See https://developer.matomo.org/guides/tests
If you have any problems with that, feel free to ask.

hywak:
Pushed changes where all unit tests were passing locally: https://github.com/matomo-org/matomo/pull/18195/commits/45a3aa5bbf1efc65f061e05b5659e571c5d97c6a

sgiehl:
The System tests are still failing though. Seems to be caused by some kind of rounding issue. Guess some check doesn't do the same as before...

sgiehl:
@hywak I will close this one now. I don't think we will be able to easily get this PR into a state where we can safely merge. The number of changes is simply to high to identify which are causing the test failures.
If you are still keen on contributing to Matomo, it would be very helpful to create smaller pull requests with the same changes instead. If only a couple of files are changed in each it's a lot easier to find the cause for failing tests and we will be a lot more faster in merging it...

PR #18208, opened 2021-10-24
github action tests

peterhashair:
### Description:

add GitHub action test. 

### PHP Tests
1. UnitTests   ✅. (Modified one tests, fully passed)
2. SystemTestsCore
3. SystemTestsPlugins
4. IntegrationTestsCore
5. IntegrationTestsPlugins
###  Js Tests
1. JavascriptTests.   ✅  (Not sure I did it correctly, but currenly shown as  passed)
2. AngularJSTests.   ✅  (Not sure I did it correctly, but currenly shown as passed)
### UI tests
1.  Group 0
2.  Group 1
3.  Group 2

around 22min runing all the php tests, we can reduce the run time a little more by add @group in tests split into smaller groups. Not sure the test faileds, it seems like the object in the function is not displayed. Could be xdebug not enable. About to set up the UI test, and js test.

### Review


.github/workflows/tests.yml
```diff
@@ -0,0 +1,107 @@
...
+      - name: Setup Mysql
+        run: |
+          docker run --name mariadb -p 3306:3306 --tmpfs /var/lib/mysql:rw -e MYSQL_DATABASE=piwik_tests -e MYSQL_ALLOW_EMPTY_PASSWORD=yes -d mariadb
+          docker ps -a
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          lfs: true
```
sgiehl: LFS will actually only be needed for UI tests, all other builds shouldn't use LFS as it takes a while to download the files and also takes a lot traffic (which we afaik pay for LFS above certain limits)

peterhashair:
@sgiehl for the php test, it's actually got coupe errors, do you have any idea if there are any configure I did wrong?

sgiehl:
I guess one point might be that we were actually using port 80 and not port 3000 if sudo was available on travis. See https://github.com/matomo-org/travis-scripts/blob/ca0dfcfa2ca0e552fc65b77649dffbf101a6a17e/setup_webserver.sh#L63-L64

peterhashair:
@sgiehl update this PR to port 80,  it's kind of working, it's skipping the error for me to debug. Still some errors on PHP tests but seems not hard to fix. Just wondering, how do I preview the Screenshot that failed to do a comparson. eg: 
image

sgiehl:
@peterhashair For Travis we have a script that uploads the tests to our server. See https://github.com/matomo-org/travis-scripts/blob/88a9409d2799b875c6762e0ed2efe02f549181c5/upload_artifacts.sh
But that won't be possible the same way using GitHub Actions, as it uses the travis api server side to validate the build number and stuff like that. Guess we might need to build something new if we want to switch to github actions

Btw. It seems there are still some php tests failing, but the build is green nevertheless. It should fail in that case for sure.

Another idea: Maybe it might also be an option to run php tests in github actions and only run UI tests on travis. That would lower the build time on travis

peterhashair:
@sgiehl yeah I did a failed ignore, just for debug purpose. should be able to fixes those php tests, just UI one is a little hard to see what's wrong for now

Issue #17923, opened 2021-08-27
Make SecurityPolicy methods a public API and fully enable the CSP


tsteur:
We introduced it in Matomo 4.5 in https://github.com/matomo-org/matomo/pull/17798

Now it's time to make it a public API see https://developer.matomo.org/guides/apis#process-for-declaring-something-as-public-api-for-plugins

We also want to disable the reporting only by default and instead actually make it apply the rules.

PR #18215, opened 2021-10-25
Fix CSP issue when viewing marketplace plugin details

sgiehl:
### Description:

When viewing the admin page, which contains a widget showing details on new plugins, or when opening the plugin details of a plugin, there is currently a CSP report shown:

```
[Report Only] Refused to load the image 'https://plugins.matomo.org/RerUserDates/images/4.0.1/RerUserDates-cal.png?w=400' because it violates the following Content Security Policy directive: "img-src 'self' 'unsafe-inline' 'unsafe-eval' data:".
```

To avoid this I've added `*.matomo.org` to `img-src`.

In addition premium plugins are loading the reviews from shop.matomo.org. This would also fail in the future, so added `*.matomo.org` to `default-src` as well.

refs #17923

### Review


plugins/CoreAdminHome/Controller.php
```diff
@@ -57,6 +57,10 @@ public function home()
         $isFeedbackEnabled = Plugin\Manager::getInstance()->isPluginLoaded('Feedback');
         $widgetsList = WidgetsList::get();
 
+        if ($isInternetEnabled && $isMarketplaceEnabled) {
```
tsteur, line 60: should this be done directly in the widget? I suppose similar issue would happen when adding a marketplace widget to the dashboard? I suppose it might not work as the widget is rendered in a separate request?

We might need to also check if the marketplace is being viewed in the reporting UI eg in a dashboard

tsteur (changes requested):

plugins/CoreAdminHome/Controller.php
sgiehl: Exactly. That doesn't work as it's another request. We could actually always add it if such a widget can be added to dashboard

sgiehl: I have checked the marketplace widgets for the dashboard. Both don't seem to show any images. So shouldn't be needed to add a CSP there.

tsteur: image
@sgiehl they show for me eg when clicking on "more details"

Also when clicking on the left menu on "Marketplace" then there may be screenshots shown when getting details 
image

Maybe when Marketplace is enabled, then *.matomo.org always need to be allowed in the reporting page?

sgiehl: ah. sure. Should be fixed now.

tsteur:
works 👍

PR #18221, opened 2021-10-26
Goals per page - new reports and metrics for tracking page conversions

### LINKED ISSUE #14974: Add Goals visualisation for Pages reports in behaviour category
see https://forum.matomo.org/t/goals-overview-for-page-interactions/34428 

It would be great if I could get a summary of how each page (under Behaviour/Pages) contributed to specific goals. Currently the only way to get this info is by clicking into the segmented visitor log and scrolling through each visitor profile one-by-one.

Ideally what I would really like to be able to do is:

* click on an icon that’s something like - Goals by page
* Then have an option to select the specific goal I want to analyse e.g. Downloaded Matomo
* Then once selected, have an option to display the goal by page at four different levels:
* viewed this page in the same sequence before achieving this goal
* viewed this page in the last 7 days before achieving this goal
* viewed this page in the last 30 days before achieving this goal
* viewed this page in the same sequence after achieving this goal

It would also be good to have the regular goals table visualisation for these reports.

Something like this anyway and then you can also open a segmented visitor log within to see it in a by profile basis.

bx80:
### Description:

Fixes #2030 

Adds new reports for goal conversions by page url, page title and entry page to the goals overview, see the detailed specification in #2030 for full details.

System tests have now been added.

### Review


justinvelluppillai:
@bx80 anything needed here to progress this?

bx80:
@justinvelluppillai Just needs reviewing.

plugins/Goals/DataTable/Filter/RemoveUnusedGoalRevenueColumns.php
```diff
@@ -0,0 +1,105 @@
...
+{
+    /**
+     * Constructor.
+     *
+     * @param DataTable $table The table to eventually filter.
+     */
+    public function __construct($table)
+    {
+        parent::__construct($table);
+    }
```
sgiehl: This should be unneeded

tests/PHPUnit/Framework/TestRequest/Collection.php
```diff
@@ -144,6 +144,9 @@ protected function _generateApiUrls()
         if ($this->testConfig->idGoal !== false) {
             $parametersToSet['idGoal'] = $this->testConfig->idGoal;
         }
+        if (isset($this->testConfig->otherRequestParameters['filter_update_columns_when_show_all_goals'])) {
+            $parametersToSet['filter_update_columns_when_show_all_goals'] = $this->testConfig->otherRequestParameters['filter_update_columns_when_show_all_goals'];
+        }
```
sgiehl: This shouldn't be needed, as the `otherRequestParameters` is automatically merged into `$parametersToSet` on line 134

plugins/Goals/tests/System/TrackGoalsPagesPageTest.php
```diff
@@ -0,0 +1,64 @@
+<?php
+/**
+ * Matomo - free/libre analytics platform
```
sgiehl: This file and `TrackGoalsPagesEntryTest.php` could be combined into one single test class. They are using the same test fixture and so merging into one file would save the time for setting up the fixture a second time when running the tests.

sgiehl (changes requested):
Only had a very rough look through the code, as it actually doesn't seem to work.
Running the archiving on this branch results in an error for me:
```
ERROR [2021-11-22 15:38:22] 23504  Got invalid response from API request: ?module=API&method=CoreAdminHome.archiveReports&idSite=1&period=week&date=2021-11-15&format=json&trigger=archivephp. Response was '{"result":"error","message":"SQLSTATE[42S22]: Column not found: 1054 Unknown column 'log_conversion.server_time' in 'on clause' - in plugin Goals. #0 \/srv\/matomo\/core\/ArchiveProcessor\/Loader.php(284): Piwik\\ArchiveProcessor\\PluginsArchiver-&gt;callAggregateAllPlugins('0', NULL, true) #1 \/srv\/matomo\/core\/ArchiveProcessor\/Loader.php(176): Piwik\\ArchiveProcessor\\Loader-&gt;prepareAllPluginsArchive('0', NULL) #2 \/srv\/matomo\/core\/ArchiveProcessor\/Loader.php(159): Piwik\\ArchiveProcessor\\Loader-&gt;insertArchiveData(0, 0) #3 \/srv\/matomo\/core\/ArchiveProcessor\/Loader.php(99): Piwik\\ArchiveProcessor\\Loader-&gt;prepareArchiveImpl('VisitsSummary') #4 \/srv\/matomo\/core\/Context.php(75): Piwik\\ArchiveProcessor\\Loader-&gt;Piwik\\ArchiveProcessor\\{closure}() #5 \/srv\/matomo\/core\/ArchiveProcessor\/Loader.php(103): Piwik\\Context::changeIdSite(1, Object(Closure)) #6 \/srv\/matomo\/plugins\/CoreAdminHome\/API.php(278): Piwik\\ArchiveProcessor\\Loader-&gt;prepareArchive('VisitsSummary') #7 \/srv\/matomo\/core\/Archive.php(847): Piwik\\Plugins\\CoreAdminHome\\API-&gt;archiveReports(1, Object(Piwik\\Period\\Day), '2021-11-17', '', 'VisitsSummary', '') #8 \/srv\/matomo\/core\/Archive.php(645): Piwik\\Archive-&gt;prepareArchive(Array, Object(Piwik\\Site), Object(Piwik\\Period\\Day)) #9 \/srv\/matomo\/core\/Archive.php(592): Piwik\\Archive-&gt;cacheArchiveIdsAfterLaunching(Array, Array) #10 \/srv\/matomo\/core\/Archive.php(518): Piwik\\Archive-&gt;getArchiveIds(Array) #11 \/srv\/matomo\/core\/Archive.php(328): Piwik\\Archive-&gt;get(Array, 'numeric') #12 \/srv\/matomo\/core\/ArchiveProcessor.php(630): Piwik\\Archive-&gt;getDataTableFromNumeric(Array) #13 \/srv\/matomo\/core\/ArchiveProcessor.php(265): Piwik\\ArchiveProcessor-&gt;getAggregatedNumericMetrics(Array, false) #14 \/srv\/matomo\/core\/ArchiveProcessor\/PluginsArchiver.php(306): Piwik\\ArchiveProcessor-&gt;aggregateNumericMetrics(Array) #15 \/srv\/matomo\/core\/ArchiveProcessor\/PluginsArchiver.php(104): Piwik\\ArchiveProcessor\\PluginsArchiver-&gt;aggregateMultipleVisitsMetrics() #16 \/srv\/matomo\/core\/ArchiveProcessor\/Loader.php(278): Piwik\\ArchiveProcessor\\PluginsArchiver-&gt;callAggregateCoreMetrics() #17 \/srv\/matomo\/core\/ArchiveProcessor\/Loader.php(176): Piwik\\ArchiveProcessor\\Loader-&gt;prepareAllPluginsArchive('22', 0) #18 \/srv\/matomo\/core\/ArchiveProcessor\/Loader.php(159): Piwik\\ArchiveProcessor\\Loader-&gt;insertArchiveData('22', 0) #19 \/srv\/matomo\/core\/ArchiveProcessor\/Loader.php(99): Piwik\\ArchiveProcessor\\Loader-&gt;prepareArchiveImpl(false) #20 \/srv\/matomo\/core\/Context.php(75): Piwik\\ArchiveProcessor\\Loader-&gt;Piwik\\ArchiveProcessor\\{closure}() #21 \/srv\/matomo\/core\/ArchiveProcessor\/Loader.php(103): Piwik\\Context::changeIdSite(1, Object(Closure)) #22 \/srv\/matomo\/plugins\/CoreAdminHome\/API.php(278): Piwik\\ArchiveProcessor\\Loader-&gt;prepareArchive(false) #23 [internal function]: Piwik\\Plugins\\CoreAdminHome\\API-&gt;archiveReports('1', Object(Piwik\\Period\\Week), '2021-11-15', false, false, false) #24 \/srv\/matomo\/core\/API\/Proxy.php(244): call_user_func_array(Array, Array) #25 \/srv\/matomo\/core\/Context.php(28): Piwik\\API\\Proxy-&gt;Piwik\\API\\{closure}() #26 \/srv\/matomo\/core\/API\/Proxy.php(335): Piwik\\Context::executeWithQueryParameters(Array, Object(Closure)) #27 \/srv\/matomo\/core\/API\/Request.php(266): Piwik\\API\\Proxy-&gt;call('\\\\Piwik\\\\Plugins\\\\...', 'archiveReports', Array) #28 \/srv\/matomo\/plugins\/API\/Controller.php(46): Piwik\\API\\Request-&gt;process() #29 [internal function]: Piwik\\Plugins\\API\\Controller-&gt;index() #30 \/srv\/matomo\/core\/FrontController.php(619): call_user_func_array(Array, Array) #31 \/srv\/matomo\/core\/FrontController.php(168): Piwik\\FrontController-&gt;doDispatch('API', false, Array) #32 \/srv\/matomo\/core\/dispatch.php(32): Piwik\\FrontController-&gt;dispatch() #33 \/srv\/matomo\/index.php(25): require_once('\/srv\/matomo\/cor...') #34 \/srv\/matomo\/core\/CliMulti\/RequestCommand.php(79): require_once('\/srv\/matomo\/ind...') #35 \/srv\/matomo\/vendor\/symfony\/console\/Symfony\/Component\/Console\/Command\/Command.php(257): Piwik\\CliMulti\\RequestCommand-&gt;execute(Object(Symfony\\Component\\Console\\Input\\ArgvInput), Object(Symfony\\Component\\Console\\Output\\ConsoleOutput)) #36 \/srv\/matomo\/vendor\/symfony\/console\/Symfony\/Component\/Console\/Application.php(874): Symfony\\Component\\Console\\Command\\Command-&gt;run(Object(Symfony\\Component\\Console\\Input\\ArgvInput), Object(Symfony\\Component\\Console\\Output\\ConsoleOutput)) #37 \/srv\/matomo\/vendor\/symfony\/console\/Symfony\/Component\/Console\/Application.php(195): Symfony\\Component\\Console\\Application-&gt;doRunCommand(Object(Piwik\\CliMulti\\RequestCommand), Object(Symfony\\Component\\Console\\Input\\ArgvInput), Object(Symfony\\Component\\Console\\Output\\ConsoleOutput)) #38 [internal function]: Symfony\\Component\\Console\\Application-&gt;doRun(Object(Symfony\\Component\\Console\\Input\\ArgvInput), Object(Symfony\\Component\\Console\\Output\\ConsoleOutput)) #39 \/srv\/matomo\/core\/Console.php(130): call_user_func(Array, Object(Symfony\\Component\\Console\\Input\\ArgvInput), Object(Symfony\\Component\\Console\\Output\\ConsoleOutput)) #40 \/srv\/matomo\/core\/Access.php(670): Piwik\\Console-&gt;Piwik\\{closure}() #41 \/srv\/matomo\/core\/Console.php(131): Piwik\\Access::doAsSuperUser(Object(Closure)) #42 \/srv\/matomo\/core\/Console.php(82): Piwik\\Console-&gt;doRunImpl(Object(Symfony\\Component\\Console\\Input\\ArgvInput), Object(Symfony\\Component\\Console\\Output\\ConsoleOutput)) #43 \/srv\/matomo\/vendor\/symfony\/console\/Symfony\/Component\/Console\/ ... ut))\n#57 \/srv\/matomo\/console(32): Symfony\\Component\\Console\\Application->run()\n#58 {main}"}'
```

The column exists in the table, so guess it might be something with a missing table alias or similar.

bx80:
@sgiehl Looks like I managed to come up with a query in LogAggregator::queryConversionsByPageView() that worked in MariaDB but not in MySQL with strict mode. :facepalm:
I've committed a fix and tested the query on both MariaDB and MySQL, hopefully that sorts the error you encountered.

core/DataAccess/LogAggregator.php
```diff
@@ -1100,6 +1100,129 @@ public function queryConversionsByDimension($dimensions = array(), $where = fals
...
+     * @return \Zend_Db_Statement|array
+     */
+    public function queryConversionsByPageView(string $linkField)
+    {
+        $dbSettings = new \Piwik\Db\Settings();
+        $tablePrefix = $dbSettings->getTablePrefix();
+        $subQuery = sprintf("
+            MAX((SELECT COUNT(am.idaction)
+             FROM %slog_conversion cam
+             LEFT JOIN %slog_link_visit_action vam ON vam.idvisit = cam.idvisit AND vam.server_time <= log_conversion.server_time
```
sgiehl: The query still throws the same error for me. Does this maybe need to be
```suggestion
             LEFT JOIN %slog_link_visit_action vam ON vam.idvisit = cam.idvisit AND vam.server_time <= cam.server_time
```

bx80: @sgiehl The log_link_visit_action.server_time from the sub-query (vam) is meant to join against the outer query log_conversion.server_time so `vam.server_time <= log_conversion.server_time` is intentional. This is working perfectly for me both on MariaDB and MySQL, so there must be some other difference in database setup. Are you using table prefixes? PDO or MySQLi? Which database version? I can try to match your test setup and recreate the issue.

Below is the extracted query in full, does it work for you if run directly? (it uses the table prefix 'matomo_' so will need a search and replace if you use something different)
```
SELECT
  log_conversion.idgoal AS idgoal,
  log_link_visit_action.idaction_url AS idaction,
  COUNT(*) AS `10`,
  COUNT(distinct log_conversion.idvisit) AS `3`,
  ROUND(SUM(log_conversion.revenue),2) AS `2`,
  ROUND(SUM(log_conversion.revenue_subtotal),2) AS `4`,
  ROUND(SUM(log_conversion.revenue_tax),2) AS `5`,
  ROUND(SUM(log_conversion.revenue_shipping),2) AS `6`,
  ROUND(SUM(log_conversion.revenue_discount),2) AS `7`,
  SUM(log_conversion.items) AS `8`,
  MAX((SELECT COUNT(am.idaction)
   FROM matomo_log_conversion cam
   LEFT JOIN matomo_log_link_visit_action vam ON vam.idvisit = cam.idvisit AND vam.server_time <= log_conversion.server_time
   LEFT JOIN matomo_log_action am ON am.idaction = vam.idaction_url
   WHERE cam.idgoal = log_conversion.idgoal AND vam.idvisit = log_link_visit_action.idvisit
   AND vam.idaction_url IS NOT NULL AND am.type = 1
   GROUP BY cam.idgoal, cam.idvisit
  )) AS `9`,
  1 AS `12`
FROM matomo_log_visit AS mv
INNER JOIN matomo_log_conversion AS log_conversion ON log_conversion.idvisit = mv.idvisit
LEFT JOIN matomo_log_link_visit_action AS log_link_visit_action ON log_link_visit_action.idvisit = log_conversion.idvisit
                                                                   AND log_link_visit_action.server_time <= log_conversion.server_time
LEFT JOIN matomo_log_action AS log_action ON log_action.idaction = log_link_visit_action.idaction_url
WHERE log_conversion.server_time >= '2021-11-03 11:00:00'
  AND log_conversion.server_time <= '2021-11-04 10:59:59'
  AND log_conversion.idsite IN (1) AND log_conversion.idgoal > 0
  AND log_link_visit_action.idaction_url IS NOT NULL
  AND log_action.type = 1
GROUP BY
  log_conversion.idgoal, log_conversion.idvisit, log_action.idaction;
```

sgiehl: @bx80 it fails on travis ci as well. My setup is similar. See https://app.travis-ci.com/github/matomo-org/matomo/jobs/549249575#L795

bx80: This query doesn't work in AWS Aurora MySQL 5.7.12, so leave it with me and I can test against that.

bx80: Ok, so the query worked with MariaDB 10.x and MySQL 8.x but not MySQL 5.7. Looks like some sort of issue with correlated subquery joins with more than one condition. I've adjusted the query to move the extra join condition to the where clause and the query now works on MySQL 5.7

sgiehl: Fine. Had a quick look and archiving seems to run through with those changes. Will continue my code review next week then.

core/API/DocumentationGenerator.php
```diff
@@ -228,6 +228,7 @@ public function getExampleUrl($class, $methodName, $parametersToSet = array())
             'idAlert'             => '1',
             'seconds'        => '3600',
 //            'segmentName'    => 'browserCode',
+            'filter_update_columns_when_show_all_goals' => '0',
```
sgiehl: Any good reason for adding that one? This causes all example urls to include that parameter. Which shouldn't be needed.
```suggestion
```

core/API/DocumentationGenerator.php
```diff
@@ -296,6 +297,7 @@ public function getExampleUrl($class, $methodName, $parametersToSet = array())
         $aParameters['compareSegments'] = false;
         $aParameters['comparisonIdSubtables'] = false;
         $aParameters['invert_compare_change_compute'] = false;
+        $aParameters['filter_update_columns_when_show_all_goals'] = false;
```
sgiehl: Guess this wouldn't be needed, as it is no new parameter and it worked before. But shouldn't hurt to have that in the list.  So no need to remove it here again.

core/DataTable/Filter/AddColumnsProcessedMetricsGoal.php
```diff
@@ -132,9 +136,13 @@ public function filter($table)
                     continue;
                 }
 
+                $extraProcessedMetrics[] = new ConversionsFloat($idSite, $idGoal); // PerGoal\Conversions or GoalSpecific\
```
sgiehl: Adding those metrics to this filter will actually add them to all reports using this filter, which are potentially all goal reports. But as those metrics are specific for the pages reports only, might be better to create an extended filter for them only. Alternatively it might be enough to add those processed metrics to the report classes.

core/Metrics.php
```diff
@@ -98,6 +98,12 @@ class Metrics
     const INDEX_GOAL_ECOMMERCE_REVENUE_SHIPPING = 6;
     const INDEX_GOAL_ECOMMERCE_REVENUE_DISCOUNT = 7;
     const INDEX_GOAL_ECOMMERCE_ITEMS = 8;
+    const INDEX_GOAL_NB_PAGES_UNIQ_BEFORE = 9;
+    const INDEX_GOAL_NB_CONVERSIONS_FLOAT = 10;
+    const INDEX_GOAL_NB_CONVERSIONS_PAGE_RATE = 11;
+    const INDEX_GOAL_NB_CONVERSIONS_PAGE_UNIQ = 12;
+    const INDEX_GOAL_NB_CONVERSIONS_ENTRY_RATE = 13;
+    const INDEX_GOAL_REVENUE_PER_ENTRY = 14;
```
sgiehl, line 105: I guess we can keep that for this reports, but for future stuff, we should avoid using such `id` to `name` mappings for metrics. We used to use them in the beginning of Matomo to save some space in the archived reports. But iirc we aren't using that anymore (for new stuff), as renaming those columns before serving the data takes some time, which can be avoided, by directly using the metric names... Or do you think this should be changed here @tsteur ?

plugins/Goals/Archiver.php
```diff
@@ -514,6 +524,274 @@ public function aggregateMultipleReports()
...
+    /**
+     * @var DataArray[]
+     */
+    protected $arrays = array();
+
+    public function aggregatePageGoalsDayReports()
+    {
+
+        $maximumRowsInDataTable = Config::getInstance()->General['datatable_archiving_maximum_rows_events'];
+        $maximumRowsInSubDataTable = Config::getInstance()->General['datatable_archiving_maximum_rows_subtable_events'];
```
sgiehl: Shouldn't that use `datatable_archiving_maximum_rows_actions` instead of the event configs?

plugins/Goals/Reports/GetPagesEntry.php
```diff
@@ -0,0 +1,81 @@
...
+{
+    protected $defaultSortColumn = '';
+
+    protected $processedMetrics = array('nb_actions_per_visit', 'avg_time_on_site', 'bounce_rate', 'nb_conversion_entry_rate');
+
+    protected function init()
+    {
+        parent::init();
+
+        $this->categoryId = 'Pages';
```
sgiehl: This should be `Goals_CategoryTextPages`, so it is translated automatically. Same applies for the other report files

plugins/Goals/Goals.php
```diff
@@ -292,6 +309,31 @@ private static function getAllReportsWithGoalMetrics()
...
+            'action'   => 'getPagesTitles',
+            'viewDataTable' => 'tableGoalsPages',
+        );
+        $reportsWithGoals[] = array('category' => 'Pages',
+            'name'     => Piwik::translate('Goals_EntryPagesTitles'),
+            'module'   => 'Goals',
+            'action'   => 'getPagesEntryTitles',
+            'viewDataTable' => 'tableGoalsEntryPages',
+            'noMenu'   => true
+        );
```
sgiehl: There actually shouldn't be any need for this special handling of the Pages reports. As they all set `hasGoalMetrics`, they would be included automatically. I assume you have done that for some kind of special handling regarding the `viewDataTable` and the `noMenu` flag?
To force a report always having a certain `viewDataTable` it should be possible simply implement this methods in the report class:
https://github.com/matomo-org/matomo/blob/b5c59ba7a4ebe0681dd6db7b9cb82d13ad93b2fa/core/Plugin/Report.php#L261-L279
But when looking through the code it seems that feature seems partially broken, as it is currently possible to overwrite a fixed viewdatatable with query parameters, which imho shouldn't be the case.
https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/core/ViewDataTable/Factory.php#L148
This should imho be `$type = $defaultViewType;`

Also I don't really see a good reason for the `noMenu` flag. Why not showing that report in the menu as well? 🤔

plugins/Goals/Reports/GetPagesEntry.php
```diff
@@ -0,0 +1,81 @@
...
+        if (null !== $this->getIdSiteFromInfos($infos)) {
+            parent::configureReportMetadata($availableReports, $infos);
+        }
+
+        $name = $this->name;
+
+        $this->addReportMetadataForEachGoal($availableReports, $infos, function ($goal) use ($name) {
+            return $goal['name'] . ' - ' . $name;
+        });
+    }
```
sgiehl: This seems to be the same for all new pages report classes. Why not having a new `BasePages` class to remove some duplicate code?

plugins/Goals/Visualizations/Goals.php
```diff
@@ -195,11 +195,32 @@ private function getGoals($idSite)
...
+            $excludedGoalColumns = json_decode($excludedGoalColumns);
+            foreach ($this->config->columns_to_display as $k => $v) {
+                foreach ($excludedGoalColumns as $excludeColName) {
+                    if ($v === $excludeColName) {
+                        unset($this->config->columns_to_display[$k]);
+                        break;
+                    }
+                }
+            }
+            $this->dataTable->deleteMetadata('excluded_goal_columns');
```
sgiehl: I still don't fully understand why you built this `excluded_goal_columns` feature. Could you maybe try to explain you intentions for that?

plugins/Goals/Visualizations/GoalsPages.php
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+    }
+
+    protected function setPropertiesForGoals($idSite, $idGoals)
+    {
+
+        $allGoals = $this->getGoals($idSite);
+
+        if ('all' == $idGoals) {
```
sgiehl:
```suggestion
        if ('all' === $idGoals) {
```

plugins/Goals/Visualizations/GoalsEntryPages.php
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        $this->config->columns_to_display[] = 'revenue_per_entry';
+    }
+
+    protected function setPropertiesForGoals($idSite, $idGoals)
+    {
+        $allGoals = $this->getGoals($idSite);
+
+        if ('all' == $idGoals) {
```
sgiehl:
```suggestion
        if ('all' === $idGoals) {
```

plugins/Goals/Visualizations/GoalsEntryPages.php
```diff
@@ -0,0 +1,82 @@
...
+namespace Piwik\Plugins\Goals\Visualizations;
+
+use Piwik\API\Request;
+use Piwik\Common;
+use Piwik\DataTable\Filter\AddColumnsProcessedMetricsGoal;
+use Piwik\Piwik;
+use Piwik\Plugins\CoreVisualizations\Visualizations\HtmlTable;
+use Piwik\Site;
+
+require_once PIWIK_INCLUDE_PATH . '/core/Twig.php';
```
sgiehl: What's that needed for?

plugins/Goals/tests/System/expected/test_trackGoals_pages__Goals.getPagesEntry_day.xml
```diff
@@ -0,0 +1,33 @@
...
+			</row>
+		</goals>
+		<nb_conversions>2</nb_conversions>
+		<revenue>20</revenue>
+		<nb_visits>0</nb_visits>
+		<conversion_rate>0%</conversion_rate>
+		<nb_actions_per_visit>0</nb_actions_per_visit>
+		<avg_time_on_site>0</avg_time_on_site>
+		<bounce_rate>0%</bounce_rate>
+		<revenue_per_visit>10</revenue_per_visit>
```
sgiehl: This exports should only contain those metrics that are relevant for the report. Stuff like `idaction`, `nb_visits`, `avg_time_on_site`,... shouldn't be included. They should not be tried to be computed at all, not sure where they are coming from.

plugins/Goals/Visualizations/GoalsPages.php
```diff
@@ -0,0 +1,82 @@
...
+            $idGoals = array_keys($allGoals);
+        } else {
+            // only sort by a goal's conversions if not showing all goals (for FULL_REPORT)
+            $this->requestConfig->filter_sort_column = 'goal_' . reset($idGoals) . '_nb_conversions';
```
sgiehl: This sorting does not seem to work. For me the page urls reports is displayed unsorted when loaded.

plugins/Goals/Visualizations/GoalsPages.php
```diff
@@ -0,0 +1,82 @@
...
+use Piwik\Piwik;
+use Piwik\Plugins\CoreVisualizations\Visualizations\HtmlTable;
+use Piwik\Site;
+
+require_once PIWIK_INCLUDE_PATH . '/core/Twig.php';
+
+/**
+ * DataTable Visualization that derives from HtmlTable and shows goals metrics for page conversions
+ */
+class GoalsPages extends Goals
```
sgiehl: I was wondering if those new Visualizations are needed. Overwriting some config variables should also be possible in the report classes. Did you maybe try that already and had any problems with it?
Generally there wouldn't be a problem with having additional visualizations, but there are a couple of checks for the `tableGoal` type in JS, that won't work for this new types automatically. e.g.
https://github.com/matomo-org/matomo/blob/b7bfb36a3d37e4d056ab3f0355672b85f318f3d9/plugins/CoreHome/javascripts/dataTable.js#L1975
https://github.com/matomo-org/matomo/blob/6b6f572693e1e4c0d0b371133bd2082e0c09525a/plugins/CoreHome/javascripts/dataTable_rowactions.js#L329
and maybe others

sgiehl (changes requested):
I have started looking through the code and testing some of the stuff. Already left a couple of comments. There might be some more when I dig deeper into it. Also almost all test suites are currently failing for various reasons. Be good to fix them at some point. 

One general thing actually came to my mind, where I'm unsure if that was kind of expected or not.
For most other goal related reports, we actually compute the goal metrics into the existing reports. So there aren't any new reports needed, and when the goal report is shown, those metrics are extracted out of the goals metadata.
Your implementation contains complete new reports. This for sure should work, but does not integrate into various mechanisms automatically. You e.g. won't be able to switch to goal metrics on the normal pages report. 
I'm aware that the pages reports are a bit more complicated, especially due to the virtual subtables for each path and such stuff.
Integrating the goals calculation there might be a bit harder though.

@tsteur do you have an opinion on that? Shall this be converted to integrated goal reports, or are additional reports fine? Guess the database queries would need to be done separately nevertheless.

tsteur:
@sgiehl @bx80 overall no big preference on that.

If we can do the same as for other reports and merge it into the regular pages report that might be beneficial and maybe few more things will function out of the box and be more consistent eg for API users. Queries would be still separate though indeed. 

If I understand this right there will be only one attribution model (linear attribution aka each page gets the same weight) and then it should be fine. If there were different attribution models, then it may be again a bit different, not sure. although then it would be just about finding different metric names for each attribution model.

I would recommend giving it a try for say 4-6 hours and check for blockers or whether there aren't any big issues.

plugins/Goals/Reports/GetPagesEntry.php
bx80: The category names are translated by `plugins/Goals/TranslationHelper::translateGoalMetricCategory()` which for some reason is prepending  the string 'Goals_CategoryText' to the categoryId, so 'Pages' becomes 'Goals_CategoryTextPages' and does get translated. It's a bit weird but it works.

plugins/Goals/Goals.php
bx80: The special handling of the pages reports menu was done partly as wasn't aware of getDefaultTypeViewDataTable() and partly because the original issue UI mock up specifically excludes the EntryPagesTitles report from the side menu in favor of having it as a related report. I've replaced the noMenu flag with a specifically excluding the report by name when building the menu in `buildGoalByDimensionView`, adjusted the reports to use getDefaultTypeViewDataTable() and removed the special handling.

plugins/Goals/Visualizations/GoalsEntryPages.php
bx80: It isn't needed, I must have forgotten to remove it when creating new visualization classes from the parent Goals visualization class. Removed.

plugins/Goals/Visualizations/Goals.php
bx80: Part of the specification is to remove goal revenue columns when there are no values to display for any of the rows. So I added a filter to check if any of the table rows had values in each of the revenue columns, but I couldn't find any way to have a filter remove columns, so instead the filter sets the `excluded_goal_columns` metadata to indicate unused revenue columns, then the visualization checks this value and hides the columns (if any). It seems to work okay, but maybe there is a better solution?

plugins/Goals/tests/System/expected/test_trackGoals_pages__Goals.getPagesEntry_day.xml
bx80: Some of there have already been fixed by other changes.  I've added a column delete filter to remove any remaining unnecessary metrics. There is a some extra post processing being done with filters to calculate page conversion metrics, so some metrics are needed for that but can then be removed. The exports should now only contain the same columns as shown on the reports.

plugins/Goals/Visualizations/GoalsPages.php
bx80: I've committed a fix for that, for specific goals on the page url report it was attempting to sort on conversions rather than conversions_float. For the goal overview where multiple goal columns are shown in the same table it should now sort on page views.

plugins/Goals/Visualizations/GoalsPages.php
bx80: The goals visualization adds the wrong goal metric columns to the report, I did initially try to set the correct metric columns via the report but the visualization then adds standard columns too. I've updated any javascript I could find that references the `tableGoals` visualization to also work with `tableGoalsEntryPages` and `tableGoalsPages`

core/ViewDataTable/Factory.php
```diff
@@ -145,7 +145,7 @@ public static function build($defaultType = null, $apiAction = false, $controlle
                 $type = $defaultType ?: self::DEFAULT_VIEW;
             }
         } else {
-            $type = $defaultType ?: $defaultViewType;
+            $type = $defaultViewType;
```
sgiehl: This seems to change the behavior. Was that change needed for anything to work?

plugins/ScheduledReports/API.php
```diff
@@ -473,6 +473,10 @@ public function generateReport($idReport, $date, $language = false, $outputType
             }
         }
 
+        usort($processedReports, function ($a, $b) {
+            return $a['metadata']['uniqueId'] <=> $b['metadata']['uniqueId'];
+        });
+
```
sgiehl: This looks unrelated

tests/PHPUnit/Integration/ReleaseCheckListTest.php
```diff
@@ -616,7 +616,7 @@ public function test_TotalPiwikFilesSize_isWithinReasonnableSize()
             // Don't run the test on local dev machine, as we may have other files (not in GIT) that would fail this test
             $this->markTestSkipped("Skipped this test on local dev environment.");
         }
-        $maximumTotalFilesizesExpectedInMb = 55;
+        $maximumTotalFilesizesExpectedInMb = 56;
```
sgiehl: Shouldn't be needed anymore

tests/PHPUnit/System/ApiGetReportMetadataTest.php
```diff
@@ -32,6 +32,11 @@ public function setUp(): void
         // ensure that they still work. This hack allows the API proxy to let us generate example
         // URLs for the ignored functions
         Proxy::getInstance()->setHideIgnoredFunctions(false);
+
+        if (version_compare(PHP_VERSION, '8.0.0') >= 0) {
+            $this->markTestSkipped('Skipping for PHP8');
+        }
+
```
sgiehl: Guess this can be remove again, or does that still fail on PHP 8?

sgiehl (changes requested):
@bx80 Before going further in the review:
Did you have a special reason for creating new API methods and reports for those goal specific page reports? I was wondering if it wouldn't have been enough to mark the existing page reports to support goal metrics by setting (`$this->hasGoalMetrics = true;`). Using that and enabling the goal viz for the view (`$view->config->show_goals = true;`) should make it possible to switch to the goals visualization on the normal page reports and those reports should automatically become visible in the `Goals By` reports. Let me know if you need more details on that. We can quickly have a chat or videocall if needed.

plugins/ScheduledReports/API.php
bx80: This was added to get the scheduled reports tests to pass as the sort order of the reports on Travis was different for PHP 7 vs PHP 8. It seems this might be a general issue not related to this PR, I'll revert it and retry the build.

tests/PHPUnit/System/ApiGetReportMetadataTest.php
bx80: It passed locally on PHP 8 but failed on Travis for some reason. I'll revert it and retry the build.

plugins/ScheduledReports/API.php
bx80: Removing this sort causes the scheduled reports tests to fail again, added it back and the tests pass.

tests/PHPUnit/System/ApiGetReportMetadataTest.php
bx80: This test still fails on PHP 8, so I've added back the version exclusion.

core/ViewDataTable/Factory.php
bx80: I don't think this is needed, I've reverted it.

bx80:
@sgiehl It would definitely be nicer to set the `show_goals` / `has_goals` options on the standard actions reports and let the visualization system take care of the rest.

I've tried this and it mostly works but I did run into a couple of issues, so before I get stuck into fixing them it'd be great to get your input and check I'm on the right track :slightly_smiling_face: 

1) The `$view->config->show_goals` setting makes assumptions about which goals visualization is to be used, the new page goal metrics need to be shown using `GoalPages` visualization instead of the usual `Goals` visualization. We could work around this by adding an alternative `$view->config->show_goals_pages` setting to use the different visualization and columns. Does that seem like a good approach?

2) The new page goal metrics cannot be entirely calculated within the archive and requires filters to be run on the dataset to populate some of the metric columns. This could be done on the `Actions.getPageX` API methods and we could remove the  `Goals.getPageX` API methods which always apply the filters, but I'm concerned about performance if these filters are run on every `Actions.getPageX` call, even when goal data isn't required.  
So I'm thinking we could either:
  a) Run the filters for every API call and hope it doesn't hurt performance too much or
  b) Perhaps there is a way to set the new Actions API parameter `includeGoals` to true just for calls made for the Goals visualization? If this could be done then we could run the goals filters only when they are needed which would minimize any performance hit. I'm not sure how to easily adjust the API call parameters based on the chosen visualization, but I can take a look or maybe you have some pointers on how this can be done?

sgiehl:
@bx80 I would have tried this way:
Don't use any extra visualization. Instead change the Goals viz and add all required stuff there. You could e.g. check with something like `$this->requestConfig->getApiModuleToRequest()` and `$this->requestConfig->getApiMethodToRequest()` if it is an action report in order to set different parameters.

If you need extra filters to be applied for the Viz only, you can actually do this with something like this in the visualization:
```php
public function beforeGenericFiltersAreAppliedToLoadedDataTable() {
    parent::beforeGenericFiltersAreAppliedToLoadedDataTable();
    $this->dataTable->filter('FilterToApply');
} 
```

Let me know if you need any further details.

bx80:
@sgiehl Thanks for the pointers, they were really helpful :+1: 

I've now removed the new reports and API methods previously added to the Goals plugin. The two additional visualizations have also been removed and their functionality added to the standard goals visualization. 

The actions page reports now have the proper flags set to indicate that they have goals metrics so the Goals visualization is automatically shown as an data table option and the reports show on the goals overview screen menu. The visualization will then check if one of the action pages API calls is being made and will set the `includeGoals` parameter on the API call which lets the API apply the additional filters instead of the visualization, this should allow external API calls to query page goals and still have the goals metrics data populated by the filters.

This is ready for the next stage of reviewing and hopefully is getting close to where it needs to be :slightly_smiling_face:

plugins/Goals/Goals.php
```diff
@@ -282,13 +307,16 @@ private static function getAllReportsWithGoalMetrics()
...
                     'category' => $report->getCategoryId(),
                     'name'     => $report->getName(),
                     'module'   => $report->getModule(),
                     'action'   => $report->getAction(),
-                    'parameters' => $report->getParameters()
+                    'parameters' => $report->getParameters(),
+                    'order' => $report->getOrder()
                 );
+
+                $reportsWithGoals[] = $r;
```
sgiehl: guess this one can actually be reverted

tests/PHPUnit/System/expected/test_noVisit__Goals.getPagesTitles_day.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
sgiehl: This and a couple of other files that had been added due to the previously added API methods can be removed again, now after those methods are removed again.

tests/PHPUnit/System/expected/test_OneVisitorTwoVisits_withCookieSupport__Actions.getPageUrls_day.xml
```diff
@@ -28,6 +28,17 @@
...
 		<max_bandwidth />
+		<goals>
+			<row idgoal='1'>
+				<nb_conversions>1</nb_conversions>
+				<revenue>42</revenue>
+				<nb_conv_pages_before>2</nb_conv_pages_before>
+				<nb_conversions_attrib>0.5</nb_conversions_attrib>
+				<nb_conversions_page_rate>0</nb_conversions_page_rate>
+				<nb_conversions_page_uniq>1</nb_conversions_page_uniq>
+				<revenue_attrib>21</revenue_attrib>
```
sgiehl, line 39: The other records in that file have some additional metrics like `revenue_entry` or `revenue_per_entry`. Is it on purpose that they are missing? Just wondering if we shouldn't have a consistent set of metrics even if they were `0`

tests/PHPUnit/System/BackwardsCompatibility1XTest.php
```diff
@@ -230,6 +234,10 @@ public function getApiForTesting()
 
             // did not exist before Matomo 4
             'PagePerformance.get',
+            'Goals.getPagesUrl',
+            'Goals.getPagesTitles',
+            'Goals.getPagesEntry',
+            'Goals.getPagesEntryTitles',
```
sgiehl: Can be removed again

plugins/Goals/Columns/Metrics/GoalSpecific/ConversionPageRate.php
```diff
@@ -0,0 +1,57 @@
...
+
+    public function getDocumentation()
+    {
+        return Piwik::translate('Goals_ColumnConversionRatePageViewedBeforeDocumentation', $this->getGoalNameForDocs());
+    }
+
+    public function getDependentMetrics()
+    {
+        return array('goals');
+    }
```
sgiehl: Is the goals metric really required here? Looks like it's not used anywhere 🤔

plugins/Goals/Columns/Metrics/GoalSpecific/ConversionPageRate.php
```diff
@@ -0,0 +1,57 @@
...
+use Piwik\Plugins\Goals\Goals;
+use Piwik\Tracker\GoalManager;
+
+/**
+ * The conversion rate for a specific goal. Calculated as:
+ *
+ *     goal's nb_conversions / sum_daily_nb_uniq_visitors
+ *
+ * The goal's nb_conversions is calculated by the Goal archiver and nb_visits
+ * by the core archiving process.
```
sgiehl: This seems not correct or not updated after it was copied 🙈 
Might be good to check the other metric files as well.

core/DataTable.php
```diff
@@ -1702,6 +1702,16 @@ public function setMetadata($name, $value)
         $this->metadata[$name] = $value;
     }
 
+    /**
+     * Deletes a metadata property by name.
+     *
+     * @param string $name The metadata name.
+     */
+    public function deleteMetadata($name)
```
sgiehl: We could maybe change this method so it has the same handling as `deleteMetadata` for a row:
https://github.com/matomo-org/matomo/blob/ec3e220b315dbd06a82508e3c0121dbe898cf243/core/DataTable/Row.php#L386-L394

Even though this might not be needed at the moment, it might be good to to have that kind of consistent.

plugins/Actions/API.php
```diff
@@ -472,17 +481,40 @@ protected function doFilterPageDatatableSearch($callBackParameters, $table, $sea
      *
      * @param DataTable|DataTable\Simple|DataTable\Map $dataTable
      * @param bool $isPageTitleType Whether we are handling page title or regular URL
+     * @param bool $includeGoals Whether goals columns should be included in the result
      */
-    private function filterActionsDataTable($dataTable, $isPageTitleType)
+    private function filterActionsDataTable($dataTable, $isPageTitleType, $includeGoals = false)
```
sgiehl: Looking at all other reports that can include goal data it seems they all include the `goals` metric by default. So maybe we should be consistent at that point and also include them here

plugins/Actions/ArchivingHelper.php
```diff
@@ -187,6 +188,155 @@ public static function updateActionsTableWithRowQuery($query, $fieldQueried, & $
...
+            if (!isset($row['idaction']) || !isset($row['type'])) {
+                continue;
+            }
+
+            // Match the existing action row in the datatable
+            $actionRow = self::getCachedActionRow($row['idaction'], $row['type']);
+            if ($actionRow === false || is_null($actionRow)) {
+                continue;
+            }
+
```
sgiehl, line 211: If I have seen that correctly `idaction` and `type` are only used to cache the action row. If that is the only purpose for them you could remove those entries from `$row` in that method, so it's not needed to do that in the API using a filter. This will avoid storing useless information in the archives.
The same is actually also done here:
https://github.com/matomo-org/matomo/blob/64f00ff4a7bdc519b384d08f68a5cb7256b25755/plugins/Actions/ArchivingHelper.php#L156-L159

plugins/Actions/DataTable/Filter/Actions.php
```diff
@@ -117,6 +117,7 @@ public function filter($table)
 
                 $row->deleteMetadata('folder_url_start');
                 $row->deleteMetadata('page_title_path');
+                $row->deleteMetadata('goals');
```
sgiehl: As mentioned on the API method, the goals details should be available by default, so it shouldn't be removed here.

plugins/Goals/Visualizations/Goals.php
```diff
@@ -30,8 +30,27 @@ class Goals extends HtmlTable
...
+    private $displayType = self::GOALS_DISPLAY_NORMAL;
+
     public function beforeLoadDataTable()
     {
+        // Check if one of the pages display types should be used
+        $requestMethod = $this->requestConfig->getApiModuleToRequest().'.'.$this->requestConfig->getApiMethodToRequest();
+        if (in_array($requestMethod, ['Actions.getPageUrls', 'Actions.getPageTitles'])) {
+            $this->displayType = self::GOALS_DISPLAY_PAGES;
+            $this->requestConfig->request_parameters_to_modify['includeGoals'] = true;
+            $this->requestConfig->request_parameters_to_modify['idGoal'] = -3;
```
sgiehl: Would in general be good to use the constants here instead. Otherwise it's hard to find if they are actually being in use anywhere.

plugins/Goals/Visualizations/Goals.php
```diff
@@ -30,8 +30,27 @@ class Goals extends HtmlTable
...
+
+        } else if (in_array($requestMethod, ['Actions.getEntryPageUrls', 'Actions.getEntryPageTitles'])) {
+            $this->displayType = self::GOALS_DISPLAY_ENTRY_PAGES;
+            $this->requestConfig->request_parameters_to_modify['includeGoals'] = true;
+            $this->requestConfig->request_parameters_to_modify['idGoal'] = -4;
```
sgiehl: Same here

plugins/Actions/API.php
```diff
@@ -472,17 +481,40 @@ protected function doFilterPageDatatableSearch($callBackParameters, $table, $sea
...
 
+        if (!$includeGoals) {
+            $dataTable->filter('Piwik\Plugins\Actions\DataTable\Filter\RemoveGoals');
+        } else {
+            $dataTable->queueFilter('Piwik\Plugins\Goals\DataTable\Filter\CalculateConversionPageRate');
+            $dataTable->queueFilter('ReplaceColumnNames');
+            $dataTable->queueFilter('Piwik\Plugins\Goals\DataTable\Filter\RemoveUnusedGoalRevenueColumns');
+            $colsToRemove = ['idaction', 'type', 'nb_conversions', 'revenue', 'nb_visits'];
+            $dataTable->queueFilter('ColumnDelete', array($colsToRemove));
+            $nestedColsToRemove = ['nb_conversions', 'nb_visits_converted', 'nb_conv_pages_before', 'nb_conversions_page_uniq'];
```
sgiehl: We shouldn't remove any columns here only because we don't show them in the UI. For some users it might be relevant to get those columns when using the API directly. Or do those column contain values that aren't correct due to any reasons?

plugins/Goals/Goals.php
```diff
@@ -257,21 +258,45 @@ public function getReportMetadataEnd(&$reports, $info)
...
+
+        $goalEntryMetrics = array(
+            'nb_conversions'  => Piwik::translate('Goals_ColumnConversions'),
+            'nb_conversions_entry_rate' => Piwik::translate('General_ColumnConversionRate'),
+            'revenue'         => Piwik::translate('General_ColumnRevenue')
+        );
+
         $reportsWithGoals = self::getAllReportsWithGoalMetrics();
 
+        $includeGoals = Common::getRequestVar('includeGoals', false);
```
sgiehl: Report metadata should not be dependent to any request parameters. But if the goal metrics are included by default guess this won't be needed anymore anyway.

plugins/Goals/Visualizations/Goals.php
sgiehl: It should actually be possible to do that all in the viz using a filter. You can simply do something like this here:
```php
$this->dataTable->filter(function($dataTable) {
  $columnsToRemove = [];
  
  // iterate over all rows and columns and check which columns to remove
  
  $dataTable->deleteColumns($columnsToRemove);
});
```

sgiehl (changes requested):
I'm not yet finished with looking through all the code, but already left a couple of comments. Feel free to already apply some changes if you have some time. Will continue my review tomorrow.

core/DataAccess/LogAggregator.php
```diff
@@ -1083,6 +1083,139 @@ public function queryConversionsByDimension($dimensions = array(), $where = fals
...
+        $subQuery = sprintf("
+            MAX((SELECT COUNT(am.idaction)
+             FROM %slog_conversion cam
+             LEFT JOIN %slog_link_visit_action vam ON vam.idvisit = cam.idvisit
+             LEFT JOIN %slog_action am ON am.idaction = vam.idaction_url
+             WHERE cam.idgoal = log_conversion.idgoal AND vam.idvisit = log_link_visit_action.idvisit
+             AND vam.idaction_url IS NOT NULL AND am.type = 1
+             AND vam.server_time <= log_conversion.server_time
+             GROUP BY cam.idgoal, cam.idvisit)) AS `9`
+        ", $tablePrefix, $tablePrefix, $tablePrefix);
```
sgiehl: using numbered placeholders makes it possible to use the same placeholder multiple times. Also it might be easier to understand when using constants instead of those "magic" numbers

```suggestion
             FROM %1$slog_conversion cam
             LEFT JOIN %1$slog_link_visit_action vam ON vam.idvisit = cam.idvisit
             LEFT JOIN %1$slog_action am ON am.idaction = vam.idaction_url
             WHERE cam.idgoal = log_conversion.idgoal AND vam.idvisit = log_link_visit_action.idvisit
             AND vam.idaction_url IS NOT NULL AND am.type = %2$s
             AND vam.server_time <= log_conversion.server_time
             GROUP BY cam.idgoal, cam.idvisit)) AS `%3$s`
        ", $tablePrefix, Action::TYPE_PAGE_URL, Metrics::INDEX_GOAL_NB_PAGES_UNIQ_BEFORE);
```
Note: this change might require additional use statements.

core/DataAccess/LogAggregator.php
```diff
@@ -1083,6 +1083,139 @@ public function queryConversionsByDimension($dimensions = array(), $where = fals
...
+                    "count(distinct log_conversion.idvisit) AS `3`",
+                    sprintf('%s AS `%d`', self::getSqlRevenue('SUM(log_conversion.revenue)'), 2),
+                    sprintf('%s AS `%d`', self::getSqlRevenue('SUM(log_conversion.revenue_subtotal)'), 4),
+                    sprintf('%s AS `%d`', self::getSqlRevenue('SUM(log_conversion.revenue_tax)'), 5),
+                    sprintf('%s AS `%d`', self::getSqlRevenue('SUM(log_conversion.revenue_shipping)'), 6),
+                    sprintf('%s AS `%d`', self::getSqlRevenue('SUM(log_conversion.revenue_discount)'), 7),
+                    "SUM(log_conversion.items) AS `8`",
+                    $subQuery,
+                    "1 AS `12`",
+                    "count(*) AS `1`",
```
sgiehl: Here you could also try to use the constants to make the query easier to understand.

core/DataAccess/LogAggregator.php
```diff
@@ -1083,6 +1083,139 @@ public function queryConversionsByDimension($dimensions = array(), $where = fals
...
+                    "log_action.type",
+                    "count(*) AS `1`",
+                    "count(distinct log_conversion.idvisit) AS `3`",
+                    sprintf('%s AS `%d`', self::getSqlRevenue('SUM(log_conversion.revenue)'), 17),
+                    sprintf('%s AS `%d`', self::getSqlRevenue('SUM(log_conversion.revenue_subtotal)'), 4),
+                    sprintf('%s AS `%d`', self::getSqlRevenue('SUM(log_conversion.revenue_tax)'), 5),
+                    sprintf('%s AS `%d`', self::getSqlRevenue('SUM(log_conversion.revenue_shipping)'), 6),
+                    sprintf('%s AS `%d`', self::getSqlRevenue('SUM(log_conversion.revenue_discount)'), 7),
+                    "SUM(log_conversion.items) AS `8`",
+                    "count(*) AS `16`",
```
sgiehl: same here

core/DataAccess/LogAggregator.php
```diff
@@ -1083,6 +1083,139 @@ public function queryConversionsByDimension($dimensions = array(), $where = fals
...
+                )
+            ),
+            // WHERE ... AND ...
+            implode(
+                ' AND ',
+                array(
+                    'log_conversion.server_time >= ?',
+                    'log_conversion.server_time <= ?',
+                    'log_conversion.idsite IN ('.Common::getSqlStringFieldsArray($this->sites).')',
+                    'log_conversion.idgoal > 0',
```
sgiehl: Shouldn't that be `>=0` so the metrics are also calculated for ecommerce?
```suggestion
                    'log_conversion.idgoal >= 0',
```

core/DataAccess/LogAggregator.php
```diff
@@ -1083,6 +1083,139 @@ public function queryConversionsByDimension($dimensions = array(), $where = fals
...
+                    "table"  => "log_visit",
+                    "joinOn" => "log_visit.idvisit = log_conversion.idvisit"
+                ),
+                array(
+                    "table" => "log_action",
+                    "joinOn" => "log_action.idaction = log_visit.".$linkField
+                )
+        );
+
+        $where   = $linkField.' IS NOT NULL AND log_conversion.idgoal > 0';
```
sgiehl: same here
```suggestion
        $where   = $linkField.' IS NOT NULL AND log_conversion.idgoal >= 0';
```

core/DataTable/Filter/AddColumnsProcessedMetricsGoal.php
```diff
@@ -129,13 +146,26 @@ public function filter($table)
...
+                if ($this->processOnlyIdGoal == self::GOALS_ENTRY_PAGES) {
+                    $extraProcessedMetrics[] = new ConversionsEntry($idSite, $idGoal); // PerGoal\Conversions or GoalSpecific\
+                    $extraProcessedMetrics[] = new GoalSpecificRevenuePerEntry($idSite, $idGoal); // PerGoal\Revenue entries
+                    $extraProcessedMetrics[] = new RevenueEntry($idSite, $idGoal); // PerGoal\Revenue entrances
+                    $extraProcessedMetrics[] = new ConversionEntryRate($idSite, $idGoal); // PerGoal\ConversionRate for entrances
+                }
+
                 if ($this->isEcommerce) {
                     $extraProcessedMetrics[] = new AverageOrderRevenue($idSite, $idGoal);
                     $extraProcessedMetrics[] = new ItemsCount($idSite, $idGoal);
```
sgiehl, line 182: Those columns are currently shown on the `Sales per pages` reports. I guess they are currently not calculated. So either we need to add the columns to the aggregating or don't show them in that case.

core/DataTable/Filter/AddColumnsProcessedMetricsGoal.php
```diff
@@ -58,6 +65,16 @@
...
+    /**
+     * Process metrics for entry page views
+     */
+    const GOALS_ENTRY_PAGES = -4;
+
+    /**
+     * Process for page views
+     */
+    const GOALS_PAGES = -3;
+
```
sgiehl, line 88: I'm not yet sure if introducing those new constants might have any implications. If I understood the implementation correct, the filter might get either passed one of those constant values or the id of a specific goal. 
So when viewing the goal reports for a specific goal, its id will be passed so only the metrics of that goal are included. 
Introducing those new constants might kind of change that behavior for the goals by pages reports only, as they might always include all goals (even though they might not be displayed afterwards). Or did I get something wrong here?

sgiehl (changes requested):
Left a couple more comments.
In general the code goes into the right direction 👍 
As mentioned in the comments currently the reports for `Sales by Pages` don't have any numbers.

tests/PHPUnit/System/expected/test_noVisit__Goals.getPagesTitles_day.xml
bx80: Done, I think I got them all.

plugins/Goals/Columns/Metrics/GoalSpecific/ConversionPageRate.php
bx80: It's not used on this metric now, at one point it was calculated. Removed.

plugins/Goals/Columns/Metrics/GoalSpecific/ConversionPageRate.php
bx80: I've corrected this comment and some of the comments on the other metrics.

core/DataTable.php
bx80: Makes sense, done :+1:

core/DataAccess/LogAggregator.php
bx80: Definitely easy to understand with constants. :+1:

core/DataTable/Filter/AddColumnsProcessedMetricsGoal.php
bx80: The new `AddColumnsProcessedMetricsGoal` constant set in the visualization was overwriting the ecommerce constant so these metrics were not added. Not sure if it's the nicest solution, but I've added extra two `AddColumnsProcessedMetricsGoal` constants for the page / entry page goals with ecommerce so that this combination is supported. This allows the ecommerce metrics to be correctly added (if ecommerce is enabled) when viewing the page goal metrics.

core/DataTable/Filter/AddColumnsProcessedMetricsGoal.php
bx80: The standard page and entry page reports do not currently have a way to select a specific goal, so instead columns get added for all goals. To support this the page / entry page constants always include all goals. Maybe in a future change we could add support to choose a specific goal on the page reports and revisit this?

plugins/Actions/API.php
bx80: Agreed it would be more consistent with the rest of the goals implementation. I'll change `includeGoals` to default to true. It was disabled by default because it could potentially add hundreds of columns to the reports, which might be problematic for existing third party API integrations. So maybe we'd need document this as a potential breaking change for the release. There will be quite a few additional tests to adjust too.

sgiehl:
@bx80 I had another look at the changes and was thinking about possible improvements for quite a while. As I wasn't that deep into the code and thus not sure what would work and not I decided to do the improvements locally to ensure they would work.
I don't think it makes much sense to add another dozens of comments and let you redo all the stuff.
I have now pushed the changes to a separate branch. See https://github.com/matomo-org/matomo/compare/m-2030-goals-per-page...goalpagesrefactor?expand=1
Maybe we can have a quick call/chat and go through the changes so I can explain you the reasons, for a better understanding.

bx80:
@sgiehl Thanks, much appreciated! :slightly_smiling_face: This will save a lot of time, I've read through the changes and they all make total sense to me :+1:
 - The `includeGoals` parameter and `RemoveGoals` filter probably weren't worth the extra complexity if goals are to be included by default. 
 - I'd missed a few places where constants should have been used instead of metric column names.
 - I hadn't realized `$table->deleteColumns()` would work from a filter, so that's a learning point and a lot more efficient than passing metadata around.
- `array_filter()` with no callback removes negative value array elements - I'll definitely be using this one! :writing_hand:  

Happy to have a chat if there's anything else of which that I should make note.

It'd probably be best if I merge your branch `goalpagesrefactor` into this one `m-2030-goals-per-page` to preserve this PR comment history and then recheck all build tests. Is that ok?

justinvelluppillai:
> array_filter() with no callback removes negative value array elements 

Does it? I thought it removed anything falsey

bx80:
> > array_filter() with no callback removes negative value array elements
> 
> Does it? I thought it removed anything falsey

Yes, sorry that's what I meant by 'negative' as anything not a positive value (and also zero), should have been more specific. `falsey` is a useful word for that, I'll add it to my vocabulary along with `truey` :smile:

sgiehl:
@bx80 Yes guess that were properly the basics. Your also commonly used `queueFilter`, which caused various issues I guess. It's good to lookup the difference between that and a direct `filter`, as that one is applied directly, and `queueFilter` after the `postProcessor` finished. Guess that also caused your issue with `nb_conversions_page_rate` metric, as it was caluclated too late and so missing in some cases. Let me know f you have any further questions.

Feel free to simply merge over the changes from my branch, or cherry pick that commit to here...

bx80:
@sgiehl @justinvelluppillai Changes merged in and all related tests now pass. Final check that we're all good to merge this into 4.x-dev?

sgiehl:
@bx80 I can have a final look at that tomorrow. Not sure if we shall merge it before 4.9 or for 4.10 though.

sgiehl:
@bx80 I had a final look at the code and applied some more fixes based on the requirements mentioned in the original issue. (Hiding the unused revenue columns in the UI didn't work correctly anymore)

What I did not yet do is some performance testing. I don't have a big database locally, but might be good to test the archiving with a big bunch of data to check if it runs smoothly. Not sure if we maybe could run an archiving query on our demo or another big account to avoid possible problems when releasing the feature.

bx80:
Performance testing results:

The two new archive queries were both tested against a large database.  

:heavy_check_mark: The query from `LogAggregator::queryConversionsByEntryPageView()` completed in 8 seconds for ~6,000 entry pages.
:x: The query from `LogAggregator::queryConversionsByPageView()` ran for 1,500 seconds before being terminated, the query plan showed ~200,000 conversions in the archived time period. Poor performance is likely due to the dependent sub-query, the `EXPLAIN` output has been logged. I will do some more query optimization and then we can re-test.

bx80:
@sgiehl I've updated the [tracker insert testing tool](https://github.com/matomo-org/matomo/tree/tracker-db-load-tester) to be able to generate goal conversions as part of generating random tracker data, this allowed me to create 1,000,000 visits, actions and conversions in a local database within a few minutes. I've then used this to test the slow  `LogAggregator::queryConversionsByPageView()` query.

Moving where clauses to joins and using `ORDER BY NULL` on the subquery to eliminate the use of `filesort` has improved the run time for me from 2mins45sec to 1min15sec.

The main outstanding reason for poor performance is the need to group the result by `idgoal`, `idvisit` and `idaction`. `idgoal` and `idvisit` are in the `log_conversion` table, but `idaction` is in the joined `log_action` table. Grouping by columns in different tables causes temporary table usage for the main result. Switching grouping from `log_action.idaction` to `log_conversion.idaction_url` improves the query time from 1min15sec to <15sec and avoids temporary table creation, however this will not work for page titles as `log_conversion.idaction_url` is only populated for page url hits.

The `log_conversion` table has a `idaction_url` column but not an `idaction_name` or `idaction` column. It seems inconsistent to store the `idaction_url` but not the `idaction_name`.

Could we add an `idaction_name` column to the `log_conversion` table, get new conversions to populate it and add a migration to fill it for existing data?

This would dramatically improve query performance for this query and possibly help future queries which need to aggregate conversions by page title action.

sgiehl:
@bx80 Thanks for the update and the improvements. 

Adding new columns to log tables isn't something we should do in a normal release. As it can take ages for bigger instances to add new columns to such tables, we are only doing that in major releases.

If we decide to add a new column, filling in data for old conversions shouldn't be done automatically. But we could provide a console command to achieve that.

Are there other possibilities to improve that? Did you check if an index would improve the runtime? Or we don't group by all columns and aggregate the results in PHP instead 🤔

tsteur:
Just FYI index changes to log tables can also only be added in major releases as it's the same problem where it can take many hours or days to add indexes and would cause unexpected issues when updating. Also new indexes on log tables if any possible we'd want to avoid as they're adding "cpu, disk and financial cost" on tracking requests

bx80:
Thanks for the info :+1: Makes sense that time consuming db updates shouldn't happen in normal releases :computer::hourglass::sleeping:

I can't see any way to improve the query performance further with the current schema, so if we can't easily add indices or new columns then the only solution is to do the aggregation in PHP. 

I've adjusted the `queryConversionsByPageView`  query to remove all grouping and also removed the sub-query. I've slightly reworked the code structure so the ungrouped data returned by the query is aggregated before being passed to a shared method for addition to the actions table. The simplified query now executes in ~250ms for 2,000,000 conversion records. There might be a lot of data for the archiving process to work though, but there are only ~13 numeric columns per row so it shouldn't use too much memory.

sgiehl:
@bx80 Are the number changes in the test files in your last commit expected?
Imho that would either mean, that the numbers were incorrect before, or the new calculation doesn't have the same result for some reason and might need to be adjusted.

Also why did you remove all the group by? Couldn't you keep those that were covered by an index and fast. That way we might not need to process too many records in PHP.

bx80:
@sgiehl The group by column that wasn't covered by the index was the last one `log_action.idaction`, if I leave the other group by columns then the data would already be partly aggregated for goal and visit, then it wouldn't be possible to breakdown by page.

There was a difference in `INDEX_GOAL_NB_VISITS_CONVERTED` values for some tests after moving the grouping to PHP. I suspect the numbers were incorrect before and the `TrackGoalsPagesTest` is still passing, but it would be good to be sure.

I'll extend the `SomePageGoalVisitsWithConversions` fixture to record multiple conversions in a single visit, it's very time consuming to debug log replay tests for this sort of issue, a known pattern of visits, page views and conversions will be better.

bx80:
@sgiehl The `TrackGoalsPagesTest` now covers multiple goal conversions in a single visit. I've manually calculated what the page goal metrics should be (screenshot of calculations attached) and then checked the test results against those values. From what I can see the correct figures are being returned in all cases.

image

sgiehl:
@bx80 awesome. Thanks for adjusting & checking it that much in detail! I looked through the changes and for me everything looked fine. But while clicking through the UI I then realized that there is a problem with calculating the "Viewed before rate". That one is done using a filter. It works without problems if the table only has one level. But when url or titles are tracked, that are displayed in a multi level table, the rate for the subtables is calculated incorrectly. The reason for that is easy to explain, but guess a lot harder to fix. When a subtable is loaded, the filter gets applied to the subtable only. So the filter uses the total number of goals based on that subtable only. That can then result in such numbers:

image

Besides that obviously visible reason, there might also be table limits that might make the numbers incorrectly. Let's assume a quite big page with thousands of sites. Each site might trigger goals. But the pages reports is limited to e.g. 500 records (by config). The current calculation would only count the goal conversions that happened on a site that is included in the report. 

I'm not sure what the best solution would be to solve those issues. One solution would be to let the filter query `Goals.get` to receive the exact number of conversions per goal. Another would be to do that while archiving and include that total conversion metric in each record. That total record could then be removed by the calculation filter.

Regarding your test case I guess it would be good to change it, so some url including the same paths are tracked - to get the page reports with multiple levels.
Also it might be good to add another tracking for a second (or third) day and also receive a weeks report in the test, so we can ensure the aggregated numbers are correct as well.

bx80:
@sgiehl Thanks for the pointers, really helpful :+1: 

I've taken the approach of having the `CalculateConversionPageRate` filter make an API call to `Goals.get` to find the total conversion for the parent data table. I briefly tried adding the conversions totals to each row when archiving, but this resulted in elevated conversion numbers (probably because rows were automatically summed again, I didn't go any further with it). The API call seems cleaner and easier to understand. 

As part of testing this I've improved the test case to include sub-pages, more conversions, more visits and visits across two days. An screenshot of the sanity check calculations is here:

image

The new `Goals_action_goals_visualization_page_urls_subtable.png` UI data table screenshot values now match this spreadsheet exactly.

For pages with sub-pages, the 'viewed before rate' can end up calculating at more than 100%, this is going to be confusing so I've tweaked the filter to cap the 'viewed before rate' to 100% - so if at least one of the sub-pages was viewed before every conversion then a 100% 'viewed before rate' has been achieved.

plugins/Goals/DataTable/Filter/CalculateConversionPageRate.php
```diff
@@ -0,0 +1,119 @@
...
+        $periodName = $period->getLabel();
+        $date = $period->getDateStart()->toString();
+
+        foreach ($goalIds as $idGoal => $g) {
+
+            $params = [
+                'period' => $periodName,
+                'date' => ($periodName === 'range' ? $date.','.$period->getDateEnd()->toString() : $date),
+                'idSite' => $idSite,
+                'idGoal' => $idGoal
```
sgiehl: I'd assume that this request might lag the segment.

plugins/Goals/DataTable/Filter/CalculateConversionPageRate.php
```diff
@@ -0,0 +1,119 @@
...
+            ];
+
+            /** @var DataTable $data */
+            $data = Request::processRequest( 'Goals.get', $params);
```
sgiehl: Depending on the intention you should set the third parameter to an empty array. Otherwise all request parameters from the original requests will be used as default. This would include the segment, but also some filter parameters, that might hide required columns.

plugins/Goals/DataTable/Filter/CalculateConversionPageRate.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Get the conversions total for each goal in the top level datatable
+     *
+     * @param DataTable $table
+     * @param array $goalIds
+     * @return array
+     */
+    private function getGoalTotalConversions(DataTable $table, array $goalIds)
```
sgiehl: We should also use a return type hint for methods where it's possible.
```suggestion
    private function getGoalTotalConversions(DataTable $table, array $goalIds): array
```

sgiehl (changes requested):
Had a rough look through the code changes and left some comments. Once fixed, I'll do a last functional testing and merge if everything works as expected.

plugins/Goals/DataTable/Filter/CalculateConversionPageRate.php
bx80: @sgiehl It could yes, the `Goals.get` method is going to load the whole blob data table again which could be slow on large datasets. Since we're storing the total conversions for each goal and period in the numeric archives I think it would be more efficient to directly lookup the total we need. `Goals.getConversions` does this but has unfortunately been disabled on the API with the @ignore annotation for some reason. 

I've committed a change where the conversion total is retrieved directly using the `Archive::getNumeric` method instead of an API call. This should be a lot faster with no API overhead and only a single value being queried instead of deserializing a datatable, applying filters, etc.

plugins/Goals/DataTable/Filter/CalculateConversionPageRate.php
```diff
@@ -0,0 +1,113 @@
...
+        $idSite = $site->getId();
+
+        $period = $table->getMetadata('period');
+        $periodName = $period->getLabel();
+        $date = $period->getDateStart()->toString();
+
+        foreach ($goalIds as $idGoal => $g) {
+
+            $date = ($periodName === 'range' ? $date.','.$period->getDateEnd()->toString() : $date);
+            $archive = Archive::build($idSite, $periodName, $date);
```
sgiehl: This one actually lags the segment as well. I'm wondering if it might be easier to move that to the Actions API. So we fetch the goal totals there and then pass the fetched totals to the filter call. That way it's a lot easier to access the required archive parameters and we don't need to rely on the datatable metadata.

sgiehl (changes requested):
Left another comment. 
One more thing that came to my mind: Did you check how the archiving behaves if no goals are configured and ecommerce is disabled? Are any queries fired nevertheless and could maybe be avoided? 🤔

plugins/Goals/DataTable/Filter/CalculateConversionPageRate.php
bx80: That makes sense, cleaner to have the goals totals injected into the filter than have the filter accessing archive data directly. I've committed an update, but I'm not sure that it will speed things up much.

bx80:
> Did you check how the archiving behaves if no goals are configured and ecommerce is disabled? Are any queries fired nevertheless and could maybe be avoided?

An attempt to lookup goal conversion totals is only made for goals that are contained in the row metadata. So if there isn't any goal metadata then no queries will be made and the filter will not run at all.

This does mean looping the table checking the metadata on each row and putting the goal ids into an array, but this should be fast with in-memory data, the alternative would be to make another API call to get all the available goals for the site and then check totals for each of these goals even if they are not present in the table - I think this would be much slower, especially if the site has dozens of goals but the data table is only being viewed for a single goal.

bx80:
There are some failed tests I'll need to check.

plugins/Actions/API.php
```diff
@@ -499,13 +500,38 @@ protected function doFilterPageDatatableSearch($callBackParameters, $table, $sea
...
+            if (isset($row[PiwikMetrics::INDEX_GOALS])) {
+                foreach ($row[PiwikMetrics::INDEX_GOALS] as $goalIdString => $metrics) {
+                    $goalIds[$goalIdString] = $goalIdString;
+                }
+            }
+        }
+
+        // For each goal in the table retrieve the period total conversions and add to the array
+        foreach ($goalIds as $idGoal => $g) {
+            $archive = Archive::build($idSite, $period, $date);
```
sgiehl: Should that `Archive::build` include the segment? Otherwise we will compare segmented action data, with unsegmented conversion numbers. Or is that intended. If so it would be good to add a comment

plugins/Goals/DataTable/Filter/CalculateConversionPageRate.php
bx80: Actually I don't think this is going to work, `filterActionsDataTable` can accept a `DataTable\Map` as well as a `DataTable` so the goal retrieval will need to happen at the filter level since we may be dealing with multiple different datatables. I'll revert the change.

plugins/Actions/API.php
bx80: It should probably include the segment as we'll want to show the viewed before rate relative to the conversions that happened for the segment. I've included the segment in the archive::build call.

bx80:
The missing conversion rate formatting on the `GoalsPages_individual_goal_updated.png` UI test should be fixed once #19295 is merged into 4.x-dev

sgiehl (approved):
The tests are now fixed. Awesome to get this finally merged 🎉

PR #18214, opened 2021-10-25
[Vue] migrate dropdown and related directives

diosmosis:
### Description:

**This PR is based off of https://github.com/matomo-org/matomo/pull/18213.**

Changes:
* add @types/materialize-css for materialize typings.
* migrate focus-anywhere-but-here directive.
* migrate focus-if directive.
* migrate menudropdown directive.

image

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #18225, opened 2021-10-27
[Vue] migrate notifications directive & service + UI.Notifications

diosmosis:
### Description:

**This PR is based off of https://github.com/matomo-org/matomo/pull/18214.**

This PR is fairly complicated since the original notifications code added and removed DOM nodes itself, where this management is meant to be done (in most cases) by the framework.

Changes:

* Migrated notification angularjs directive to Vue (in Notification.vue).

* Moved notification.js logic to new NotificationsStore class. This store houses a list of displayed notifications. Notifications are added to the store to display them, removed from the store to destroy them.

* Created NotificationGroup component to display a list of notifications based on what's in the store. This approximates the `placeat` functionality except instead of directly adding DOM nodes to an element, we are adding notifications to the store which a NotificationGroup will see and display. The specific NotificationGroup that displays the notification is determined by a new `group` property in client side notifications.
  * NotificationGroups are currently dynamically instantiated via NotificationsStore.show(). If we see that the placeat node is not initialized as the vue component, we do so. Then it pulls notifications from the store and displays them.
  * This dynamic instantiation logic will eventually be removed once Vue is used everywhere (then the <NotificationGroup> component will just be used directly).

* Added toast functionality to NotificationsStore (the previous implementation appears to be completely broken). 

* Implemented the scrollToNotification() functionality using an attribute we select for, since we no longer have the actual DOM node for a notification.

image

### Review


sgiehl:
@diosmosis could you please merge in the latest changes from 4.x-dev and resolve the conflicts. The diff currently also shows all changes from the previous PR, so it's hard to look through it.

diosmosis:
@sgiehl done

plugins/CoreHome/vue/src/Menudropdown/Menudropdown.vue
```diff
@@ -29,6 +29,7 @@
         <input
           type="text"
           v-model="searchTerm"
+          v-focus-if="{ focusIf: showItems }"
```
sgiehl: Should that give the search box a focus when the menu is opened? For me that doesn't work

plugins/UsersManager/tests/UI/UsersManager_spec.js
```diff
@@ -419,7 +419,7 @@ describe("UsersManager", function () {
 
         expect(await page.screenshotSelector('.usersManager')).to.matchImage('permissions_remove_access');
     });
-
+return;
```
sgiehl: Guess this one is left over from testing.

sgiehl (changes requested):
Overall this seems to work. The animation for appearing notifications looks slightly different, but imho that would be fine, as it doesn't make a big difference.
Left some small comments, besides that it would be good to merge.
Not sure if maybe @peterhashair wants to have a look as well.

plugins/CoreHome/vue/src/Menudropdown/Menudropdown.vue
peterhashair: I think that should work, but not perfect

```suggestion
         autofocus
```

plugins/CoreHome/vue/src/Notification/Notification.less
```diff
@@ -8,6 +8,10 @@
     margin-top: 1em;
   }
```
peterhashair, line 10: not sure if we need a `max-width:1300px` here. see the image, the notification is not lined up on a large screen.
image

peterhashair (changes requested):
Looks good, maybe one mirror CSS change.

plugins/CoreHome/vue/src/Menudropdown/Menudropdown.vue
diosmosis: > I think that should work, but not perfect

That could work here, but unfortunately we need the directive to continue working for backwards compatibility with existing anguarjs code.

Looks like Vue 3 directive values are static, once set they never change. I changed it instead to use the directive argument, which is supplied like `v-focus-if:[argument]`. It looks weirder, but it's the only way I could get it to work.

plugins/UsersManager/tests/UI/UsersManager_spec.js
diosmosis: :+1:

plugins/CoreHome/vue/src/Notification/Notification.less
diosmosis: @peterhashair this looks strange, I agree, but I think it's a problem w/ admin pages, since it looks fine on the dashboard and other report pages (the widgets stretch as far as the notification). I checked on 4.x-dev and it's an issue there so I'm guessing this might be a good candidate for a new issue :+1:

plugins/CoreHome/vue/src/Menudropdown/Menudropdown.vue
peterhashair: @diosmosis somehow the showItems is passed as false. I guess the internal times have to be before the if blind value.

diosmosis: @peterhashair it's always false, the value in the directive can't change after it's first set apparently. Only the argument is allowed to be dynamic according to https://v3.vuejs.org/guide/custom-directive.html.

sgiehl (approved):
LGTM. Merged in latest 4.x-dev changes. Will wait for the tests before merging.

PR #18228, opened 2021-10-27
Show correct metrics in row evolution triggered from goal view tables

sgiehl:
### Description:

This PR contains the following changes:

The row evolution will now show the goal metrics when a row evolution is opened from 
- a report on a goals report page (specific or overview) 
- the ecommerce sales page
- or any other report that has been switched to the goals visualization

Additionally it fixes this issues that existed before:

- When a row evolution was opened from a goals report page the url parameter contained the `idGoal`, loading this URL again resulted in an error page, as the previously used javascript did not remove the `idGoal` before sending the request. See eg https://demo.matomo.cloud/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=day&date=yesterday&segment=&category=Goals_Goals&subcategory=4&popover=RowAction$3ARowEvolution$3AReferrers.getReferrerType$3A$257B$2522idGoal$2522$253A$25224$2522$257D$3A$40Websites

- In a multi row evolution it is possible to choose the metric. Choosing the metric caused loosing any possible id parameters before, as the `extraParameters` were reset for this action. This caused a similar error as above, as an id param might be missing. This can be reproduced when e.g. switching the metric in a multi row evolution for forms: https://demo.matomo.cloud/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=day&date=yesterday&segment=&category=FormAnalytics_Forms&subcategory=7&popover=RowAction$3ARowEvolution$3AFormAnalytics.getPageUrls$3A$257B$2522action$2522$253A$2522getMultiRowEvolutionPopover$2522$252C$2522idForm$2522$253A$25227$2522$257D$3A$40divezone.net$252Fjobs$252Fapply$252Fchinese-speaking-ssi-instructor-or-padi-instructor$2C$40divezone.net$252Fjobs$252Fapply$252Fdivemaster-2

fixes #3883

### Review


sgiehl:
This can be already reviewed if someone has some time. I'll try to add some additional UI tests tomorrow that cover the "bugs" I mentioned in the description to ensure they won't reoccur in the future.

plugins/API/RowEvolution.php
```diff
@@ -90,6 +149,41 @@ public function getRowEvolution($idSite, $period, $date, $apiModule, $apiAction,
...
+                return $this->getAllGoalIds($idSite);
+            case Piwik::LABEL_ID_GOAL_IS_ECOMMERCE_ORDER:
+            default:
+                return [$goalId];
+        }
+    }
+
+    protected function getAllGoalIds($idSite)
+    {
+        static $goalIds = [];
```
tsteur: might need to use a transient cache here so it's automatically reset between tests?

tsteur (commented):
Works 👍  Code looks good too, left only one comment

plugins/API/RowEvolution.php
sgiehl: Guess I will simply remove that caching again. Had used the method somewhere else before. But as it's not needed anymore and actually only called once in a request, there's no need to cache it.

sgiehl: Ah no. It's the method that calls `getAllGoalIds`, which is still called twice. Will use Transient cache then

sgiehl:
@tsteur updated the caching and also added some more tests for row evolution.
Also decided to move some of the goals related UI tests to the Goals plugin. Imho it's easier to have smaller units of tests than everything in the big `UIIntegration` class, as it's easier to execute those single tests locally then.

PR #18243, opened 2021-10-29
fix concurrency archiving overview goals

### LINKED ISSUE #18088: Multiple requests may archive the same range request at the same time

To reproduce this:

```sql

-- same for blob
```

image

```sql
```

peterhashair:
### Description:
Fixes: #18088
Fix concurrency archiving overview goals with multiple ajax requests. Used a flock lock on the api request. Not the best solution, but should reduce the number of multiple archiving.

### Review


peterhashair:
@tsteur I think that PR fixed the concurrency for the goal overview archive

tsteur:
btw @peterhashair it be great if any lock would be only used if `Rules::isArchivingEnabledFor() &&  !SettingsServer::isArchivePhpTriggered();` because only then we will actually need it.

That means the lock will be only used if browser archiving is used or when a `range` is being requested. This way we avoid the overhead of locks when it's doing the regular archiving using cron as the cli archiver already pretty much makes sure we won't archive the same date concurrently. 

Also ideally the solution would automatically work for all reports if possible. If we only fix it eg in Goals then we could as well just use the JS fix to only request one widget instead of all of them

peterhashair:
@tsteur I locked `Rules::isArchivingEnabledFor() && !SettingsServer::isArchivePhpTriggered();` those two, it still has 1 duplicated after I tried some task 10 times. Is there any more places needs to lock?
https://github.com/matomo-org/matomo/blob/0475bc2717c52fb864739b64328c36aa508f6a4e/core/ArchiveProcessor/Loader.php#L136-L171

tsteur:
@peterhashair In chat the other day I mentioned we need to should only lock when `if ($this->params->isRootArchiveRequest() && !SettingsServer::isArchivePhpTriggered()) {`. Maybe try if that helps (we'd definitely need a check to only lock in root archive request and not archive if the archiving is triggered). 

If then there's still one duplicate that should be fine. It's also possible that the duplicate goes away once https://github.com/matomo-org/matomo/pull/18255 is merged

core/ArchiveProcessor/LoaderLock.php
```diff
@@ -0,0 +1,70 @@
...
+use Piwik\Option;
+use Piwik\Period;
+use Piwik\Piwik;
+use Piwik\SettingsServer;
+use Piwik\Site;
+use Psr\Log\LoggerInterface;
+use Piwik\CronArchive\SegmentArchiving;
+
+/**
+ * This class uses PluginsArchiver class to trigger data aggregation and create archives.
```
sgiehl: Seems that one was copied from there and don't match anymore. Also most use statements above a actually unneeded.

plugins/Goals/API.php
```diff
@@ -464,57 +466,57 @@ protected static function convertSpecialGoalIds($idGoal)
...
+            if (!empty($compare)) {
+                $getMetricsReport = ReportsProvider::factory('Goals', 'getMetrics');
+                $table->queueFilter(function (DataTable $t) use ($getMetricsReport) {
+                    $t->setMetadata(Metrics\Formatter::PROCESSED_METRICS_FORMATTED_FLAG, false);
+
+                    $formatter = new Metrics\Formatter();
+                    $formatter->formatMetrics($t, $getMetricsReport, $metricsToFormat = null, $formatAll = true);
+                });
+            }
+            return $table;
```
sgiehl: Did you actually change anything in that files? Looks like there are only whitespace changes 🤔

core/Concurrency/Lock.php
```diff
@@ -109,6 +108,22 @@ public function isLocked()
...
+        $lockValue = $this->backend->get($this->lockKey);
+        return !empty($lockValue);
+    }
+
+    public function isLockedByOtherProcess()
+    {
+        $lockValue = $this->backend->get($this->lockKey);
+        return !empty($lockValue) && $this->lockValue !== $lockValue;
+    }
+
```
sgiehl: Seems those methods are actually not in use anymore, so would be good to remove them again.

core/ArchiveProcessor/LoaderLock.php
```diff
@@ -0,0 +1,70 @@
...
+class LoaderLock
+{
+
+    const MAX_LOCK_TIME = 60; //in seconds
+    protected $id;
+
+    public function __construct($id)
+    {
+        if (strlen($id) > 64) {
+            throw new \Exception('excceed lenth');
```
sgiehl:
```suggestion
            throw new \Exception('exceeds length');
```

core/ArchiveProcessor/LoaderLock.php
```diff
@@ -0,0 +1,70 @@
...
+        }
+        $this->id = $id;
+    }
+
+    public function setLock()
+    {
+        Db::fetchOne('SELECT GET_LOCK(?,?)', array($this->id, self::MAX_LOCK_TIME));
```
sgiehl, line 36: Just FYI: We are still supporting MySQL 5.5+. Those lock features in MySQL have been fully rewritten in MySQL 5.7 / MariaDB 10.0.2. Before that rewrite only a single simultaneous lock can be acquired and GET_LOCK() releases any existing lock. Doesn't matter in the current use case, as we only use one lock. But maybe we could mention that in the doc block of the file, so if we ever need to use multiple locks and are running into problems we have that noted somewhere.

core/ArchiveProcessor/LoaderLock.php
```diff
@@ -0,0 +1,70 @@
...
+    public function unLock()
+    {
+        Db::query('DO RELEASE_LOCK(?)', array($this->id));
+    }
+
+    public function getLock()
+    {
+        $row = Db::fetchOne('SELECT IS_FREE_LOCK(?);', array($this->id));
+
+        if ($row[0] === 1) {
```
sgiehl: That won't work for all DB adapters. PDO and MySQLi differ in return types. Might be possible that a string is returned instead, so be good to cast that.
Btw. If you write some simple tests for that, you can see on travis if it works for both adapters.

core/ArchiveProcessor/Loader.php
```diff
@@ -119,20 +123,77 @@ private function prepareArchiveImpl($pluginName)
...
+
+        if (SettingsServer::isArchivePhpTriggered()) {
+            $this->logger->info("initiating archiving via core:archive for " . $this->params);
+        }
+
+        //get Lock ID
+        $lockId = $this->makeArchivingLock();
 
+        //ini lock
+        $this->lock = new LoaderLock($lockId);
```
sgiehl: If this is only used in that method, there is actually no need in having a class property for that and a local variable can be used.

core/ArchiveProcessor/Loader.php
```diff
@@ -144,24 +205,14 @@ private function prepareArchiveImpl($pluginName)
...
-
-        list($visits, $visitsConverted) = $this->prepareCoreMetricsArchive($visits, $visitsConverted);
-        list($idArchive, $visits) = $this->prepareAllPluginsArchive($visits, $visitsConverted);
-
-        if ($this->isThereSomeVisits($visits) || PluginsArchiver::doesAnyPluginArchiveWithoutVisits()) {
-            return [[$idArchive], $visits];
-        }
-
-        return [false, false];
+        return [$visits, $visitsConverted];
```
sgiehl: If I see that correctly, this return is actually the only case where `quickReturn` would be `false` and the return values are actually not used, as the archiving will be triggered, right?
Guess you could simply return `null` or `false` in that case and check for that in the main method to get rid of `quickReturn` at all maybe.

core/ArchiveProcessor/Loader.php
```diff
@@ -119,20 +123,77 @@ private function prepareArchiveImpl($pluginName)
...
+            }
+
+            //insert data
+            list($visits, $visitsConverted) = $this->prepareCoreMetricsArchive($visits, $visitsConverted);
+            list($idArchive, $visits) = $this->prepareAllPluginsArchive($visits, $visitsConverted);
+            if ($this->isThereSomeVisits($visits) || PluginsArchiver::doesAnyPluginArchiveWithoutVisits()) {
+                return [[$idArchive], $visits];
+            }
+        } catch (\Exception $e) {
+            throw new \Exception("exceed max time");
```
sgiehl: What case are you actually trying to catch here? Before there hadn't been any try catch around that code and the lock method is called outside 🤔

sgiehl (changes requested):
Left a couple of comments

core/ArchiveProcessor/LoaderLock.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        Db::fetchOne('SELECT GET_LOCK(?,?)', array($this->id, self::MAX_LOCK_TIME));
+    }
+
+    public function unLock()
+    {
+        Db::query('DO RELEASE_LOCK(?)', array($this->id));
+    }
+
+    public function getLock()
```
tsteur: @peterhashair it looks like this method is not used and can be removed

core/ArchiveProcessor/LoaderLock.php
```diff
@@ -0,0 +1,70 @@
...
+class LoaderLock
+{
+
+    const MAX_LOCK_TIME = 60; //in seconds
+    protected $id;
+
+    public function __construct($id)
+    {
+        if (strlen($id) > 64) {
+            throw new \Exception('exceeds length');
```
tsteur: can we here instead change the ID to fit into 64 characters? Like `substr($id,0,32).md5($id)`? This will make it more developer friendly and avoid if we accidentally create a too long id

core/ArchiveProcessor/LoaderLock.php
```diff
@@ -0,0 +1,70 @@
...
+ */
+class LoaderLock
+{
+
+    const MAX_LOCK_TIME = 60; //in seconds
+    protected $id;
+
+    public function __construct($id)
+    {
+        if (strlen($id) > 64) {
```
tsteur: We actually need to prepend `SettingsPiwik::getPiwikInstanceId()` here to the ID. Like 

```php
$id = SettingsPiwik::getPiwikInstanceId() . $id;
```

Otherwise there may be issues with a [multi tenant installation](https://developer.matomo.org/guides/multi-tenants) like we have on the cloud that one Matomo DB locks queries of another Matomo DB as these locks are across databases

core/ArchiveProcessor/LoaderLock.php
```diff
@@ -0,0 +1,70 @@
...
+use Piwik\Piwik;
+use Piwik\SettingsServer;
+use Piwik\Site;
+use Psr\Log\LoggerInterface;
+use Piwik\CronArchive\SegmentArchiving;
+
+/**
+ * This class uses PluginsArchiver class to trigger data aggregation and create archives.
+ */
+class LoaderLock
```
tsteur, line 14: @peterhashair can we write tests for this class to make sure it works?

core/ArchiveProcessor/Loader.php
```diff
@@ -119,20 +123,77 @@ private function prepareArchiveImpl($pluginName)
...
+        }
+
+        //get Lock ID
+        $lockId = $this->makeArchivingLock();
 
+        //ini lock
+        $this->lock = new LoaderLock($lockId);
+
+        //set mysql lock the entire process if another process is running
+        $this->lock->setLock();
```
tsteur: see https://github.com/matomo-org/matomo/pull/18243#issuecomment-960304570

We only want to get the lock if we are in a root archive, and not if we are archiving (since archiver should already avoid archiving the same thing twice) https://github.com/matomo-org/matomo/pull/18243#issuecomment-960304570 . This way we can avoid using locks when they aren't needed and can potentially also avoid deadlock situations.

tsteur (changes requested):

core/ArchiveProcessor/LoaderLock.php
peterhashair: @tsteur add a test,  simple one, let me know if that doesn't meet the requirements

tsteur: @peterhashair can you move this test into a `LoaderLockTest.php`? For each class/file we also create a different test class.

What you additionally want to test is that

* Create another lock with a different idea, check that the initially created lock doesn't also lock this lock with different ID. To make sure it only locks the specific ID.
* A test where you enter a long ID with > 64 characters
* Call "unlock" when it's not locked yet
* Are there other edge cases that can be tested?

peterhashair: @tsteur update it, but it seems like Travis number 7 angular test is started to fail.  [https://app.travis-ci.com/github/matomo-org/matomo/jobs/547651253](https://app.travis-ci.com/github/matomo-org/matomo/jobs/547651253). There may be a regression somewhere

core/ArchiveProcessor/Loader.php
```diff
@@ -119,20 +122,91 @@ private function prepareArchiveImpl($pluginName)
...
         ) {
             $this->invalidatedReportsIfNeeded();
         }
+        // load existing data from archive
+        list($visits, $visitsConverted) = $this->loadArchiveData();
+        if ($this->quickReturn) {
+            return [$visits, $visitsConverted];
+        }
+
+        if (SettingsServer::isArchivePhpTriggered()) {
```
tsteur: this method for the logging will need to go into `insertArchiveData` in case we launch the archiving around line 157

tests/PHPUnit/Integration/ArchiveProcessor/LoaderLockTest.php
```diff
@@ -0,0 +1,85 @@
...
+class LoaderLockTest extends IntegrationTestCase
+{
+
+    public function test_singleLocking()
+    {
+        $lockId = "lock1";
+        $lockOne = new LoaderLock($lockId);
+        $lockOne->setLock();
+        $formatLockKey = $lockOne->getId();
+        $isLocked = Db::fetchOne('SELECT IS_FREE_LOCK(?)', [$formatLockKey]);
```
tsteur: maybe we can move the `Db::fetchOne('SELECT IS_FREE_LOCK(?)', [$formatLockKey]);` after all back into the `LoaderLock` class to remove some duplicate code? Initially it was removed as it wasn't used but now it would be used so could move it there?

tests/PHPUnit/Integration/ArchiveProcessor/LoaderLockTest.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public function test_lockerIdMaxLength()
+    {
+        $lockId = $this->generateRandomString(128);
+        $this->assertEquals(strlen($lockId),128);
+
+        $lock = new LoaderLock($lockId);
+        $check = strlen($lock->getId())<64;
+        $this->assertEquals($check,1);
```
tsteur: just fyi (no change required).

Here you can use ` $this->assertTrue(strlen($lock->getId())<64);` or even better `$this->assertLessThanOrEqual(64,strlen($lock->getId()));`

tsteur: actually in this case you would even want to check that `$this->assertSame(64,strlen($lock->getId()));`

because the test would otherwise pass even if an empty ID was returned so we can't actually check that the logic works.

The very best solution would be to not produce a random string, but created a fixed string like `$lockId = str_pad('test', 128, '1');`

and then test the exact returned value for the lock id like 

`$this->assertSame(' the generated lock id', $lock->getId())`

This way you make sure that the method works as intended. This should be actually changed so we make sure it shortens the ID correctly as expected.

core/ArchiveProcessor/Loader.php
tsteur: This be good to change indeed. Setting a property on the object to read it here is rather an anti-pattern and shouldn't be needed. We'd want to remove `$quickReturn`

core/ArchiveProcessor/LoaderLock.php
```diff
@@ -0,0 +1,45 @@
...
+{
+
+    const MAX_LOCK_TIME = 60; //in seconds
+    protected $id;
+
+    public function __construct($id)
+    {
+        // for multi tenant database solution
+        $id = md5(SettingsPiwik::getPiwikInstanceId() . $id);
```
tsteur: It be great to hash the ID only if it's longer than 64 characters (too long). This will help better understand what lock is being set eg if we had to troubleshoot some locking issues directly in the database but also while debugging the code.

If the ID is longer than 64 characters, then we will want to do what I mentioned in https://github.com/matomo-org/matomo/pull/18243#discussion_r746152922

The benefit of something like `substr($id,0,32).md5($id)` is that most of the ID is still readable which is very useful when debugging.

tsteur (changes requested):
@peterhashair this worked nicely for me. I left a few more comments to look at.

There is also a test failing that needs to be fixed: https://app.travis-ci.com/github/matomo-org/matomo/jobs/547651249#L787
image

The failing angular test should be unrelated to this change.

I expect this to be the last iteration and will do a final test again after these changes.

tsteur:
@peterhashair I've made few tweaks like:

* removing requires MySQL 5.6 comment as it's been there at least in 4.1 http://download.nust.na/pub6/mysql/doc/refman/4.1/en/miscellaneous-functions.html . Otherwise we would have needed to catch potential exceptions etc. Or did you maybe mean some other reason why it required MySQL 5.6 for the locks?
* When the lockID is < 64 characters then we don't hash the ID. I mentioned this in the comment as there's not always a need to hash it 
* The log `"initiating archiving via core:archive for..."` wasn't always logged and moved it now into `insertArchiveData`.

I tested it again and worked nicely. If the tests pass then we can merge 🎉 . Made these changes quickly myself so we can merge it before the release to avoid another round of changes and reviews.

PR #18255, opened 2021-11-02
Fix archiving too many segments that aren't needed and showing 0 conversions for new visits/returning visitors

tsteur:
### Description:

Noticed while looking into https://github.com/matomo-org/matomo/issues/18088 that we're archiving way too many segments when eg viewing a goals overview page see the issue on how to reproduce it.

Previously we would have archived 9 archives when no segment was used, and 13 archives when a segment was used. Now we're only archiving 6 archives each which is:

* an archive for all segment (or the given segment)
* an archive for the segment + new visitors
* an archive for the segment + returning visitors

Each of these get archived twice. Once for goals and once for visits summary. Previously we would have archived a lot of segments but never looked at these.

Eg previously we would have archived roughly like this
* an archive for all segment (or the given segment)
* an archive for the segment + new visitors
* an archive for the segment + returning visitors
* new visitors + an archive for the segment 
* returning visitors +  an archive for the segment
* an archive for the segment wrong encoding + new visitors
* an archive for the segment wrong encoding + returning visitors
* new visitors + an archive for the segment + returning visitors
* returning visitors + an archive for the segment + new visitors

### Review


plugins/VisitFrequency/API.php
```diff
@@ -60,7 +61,7 @@ public function get($idSite, $period, $date, $segment = false, $columns = false)
         }
 
         foreach ($visitTypes as $columnSuffix => $visitorTypeSegment) {
-            $modifiedSegment = $this->appendVisitorTypeSegment($segment, $visitorTypeSegment);
+            $modifiedSegment = Segment::combine($segment, SegmentExpression::AND_DELIMITER, $visitorTypeSegment);
```
tsteur, line 64: this was just to reuse existing logic

plugins/Goals/API.php
```diff
@@ -476,7 +476,16 @@ public function get($idSite, $period, $date, $segment = false, $idGoal = false,
...
+            if (!empty($predefinedSegment)) {
+                // we are disabling the archiving of these segments as the archiver archives them already using
+                // archiveProcessDependend logic. Otherwise we would eg archive reports that we don't need:
+                // userid=5;visitorType%3D%3Dnew;visitorType%3D%3Dreturning%2CvisitorType%3D%3DreturningCustomer
+                // userid=5;visitorType%3D%3Dreturning%2CvisitorType%3D%3DreturningCustomer;visitorType%3D%3Dnew;
+                // it would also archive dependends for these segments that we already combined here and then combine
+                // segments again when archiving dependends
+                Archiver::$ARCHIVE_DEPENDENT = false;
+            }
+            $segmentToUse = $this->appendSegment($segment, $predefinedSegment);
```
tsteur, line 488: flipped `segment` and `predefinedSegment` parameter.

Previously, in API we were using `$predefinedSegment, $segment` and then in the archiver we would have used `$segment, $predefinedSegment` meaning they actually resulted in different segment strings and resulted in quite a few additional segment archives that weren't even needed.

Since the archiver archives it the correct way, changed the API so we're always using the correct existing/previously archived archives.

It can prevent some cases where we might have shown 0 visits for these metrics.

tsteur:
I can reproduce this actually on our own Matomo how we're seeing 0 conversions for new visits/returning visitors depending on what segment is selected.

image

core/ArchiveProcessor.php
```diff
@@ -673,7 +673,13 @@ public function processDependentArchive($plugin, $segment)
...
         $idSites = [$params->getSite()->getId()];
 
-        $newSegment = Segment::combine($params->getSegment()->getString(), SegmentExpression::AND_DELIMITER, $segment);
+        // important to use the original segment string when combining. As the API itself would combine the original string.
+        // this prevents a bug where the API would use the segment
+        // userId!@%2540matomo.org;userId!=hello%2540matomo.org;visitorType==new
+        // vs here we would use
+        // userId!@%40matomo.org;userId!=hello%40matomo.org;visitorType==new
+        // thus these would result in different segment hashes and therefore the reports would either show 0 or archive the data twice
+        $newSegment = Segment::combine($params->getSegment()->getOriginalString(), SegmentExpression::AND_DELIMITER, $segment);
```
tsteur, line 682: for some segments where originalString != getString this would actually result in a different segment hash and therefore a different done flag. However, it would have also not worked earlier AFAIK as the UI/API would have requested a different segment string and therefore also would have used a different hash.

sgiehl (approved):
Good catch. Should indeed safe the archiving of unneeded segments 👍 
Do you think it might make sense to somehow invalidate the segments that were archived without any use. We could eg go through all pre-archived segments and prepend new & returning visitors segment and invalidate them? But maybe that's not worth the effort in order to safe some database space...

tsteur:
@sgiehl I noticed on cloud that Matomo will try to re-archive some data because "originalString" is now used in dependent archives. I thought it wouldn't be a problem as it would only archive current day, week, ... 

But of course when it eg tries to archive the current year, it will also archive all subperiods for that period then.

I wrote a migration script that I partially tested on our Matomo to migrate basically previous done flags to the new done flag so this re-archiving won't be happening as it can cause quite a bit of load. 

Can you have a look at this script to see if you notice any issues with that https://gist.github.com/tsteur/13721a2657838b6b71c1f688c9ac3fdd ? 

I could potentially add it as a DB upgrade script in Matomo then (if wanted)

sgiehl:
@tsteur had a look at it. In general I guess it should work. The only edge case that might possibly happen, is that archives for both segment variations might already exist upfront, and then the rename could fail with a duplicate key error. not sure if that is something that should be handled.

tsteur:
Thanks @sgiehl I reckon duplicates shouldn't be an issue as the index is on `idarchive,name` meaning duplicate names are no problem as long as the idarchive is different 
image

tsteur:
Added a migration script @sgieh
Screenshot 2021-11-05 at 10 18 13 AM
l've also tested the script locally and on our cloud. Making sure it now generates the same segment hashes as would be used by the logic itself and seems to all work.

tsteur:
We could check if the archives actually include any of the segment hashes and only show the needed update statements. Makes a few more selects though.

sgiehl:
@tsteur not sure if it would be better to first do some simple selects. Currently it would fire an update statement on each archive table, even though it might not need to change anything. Guess for people executing those queries manually it would be better to have only required queries listed.

tsteur:
@sgiehl added the select query 👍

sgiehl (approved):
@tsteur I've merged in the latest 4.x-dev changes and bumped the version number, so the update script is triggered. For me locally, not even one query was needed. Seems the segments were already generated correctly, for my case.
Nevertheless, if that worked for you, feel free to merge this PR if tests are passing.

tsteur:
@sgiehl it's only a problem for some segments. Like `actions>1` isn't a problem, but more complicated segments like `userId!@%40matomo.org;userId!=hello%40matomo.org;visitorType==new` would be affected. It worked for me and we'll run this on production soon. Can wait till then.

tsteur:
The migration script seems to have worked quite well for us.

Issue #18232, opened 2021-10-27
Remove "refer us" feature


tsteur:
In https://github.com/matomo-org/matomo/pull/17234 we added the "refer us" feature which shows a popup asking to share on twitter, facebook, linkedin.

We couldn't see too many shares so far now we would like to remove this feature again to have less popups.

tsteur:
@peterhashair seems this one is maybe done and can be closed?

peterhashair:
@tsteur yes can be closed.

PR #18262, opened 2021-11-03
add feedback question, remove other related.

### LINKED ISSUE #18250: Remove review popup
We have a logic to show a popup like below to ask for a review on few review sites. The feature was added in https://github.com/matomo-org/matomo/pull/14432

image

We'd like to remove this feature as it's not much used. 

We're also removing the "refer us" feature in https://github.com/matomo-org/matomo/issues/18232

We're still asking for a review after a user rated a feature positive. We would keep this feature.

### LINKED ISSUE #18251: Collect feedback about what features people are using?
We are wanting to learn more about what features or services people are missing in Matomo so we can improve it further.

## How we ask for feedback

Therefore, we want to ask for feedback in the app like this:

Screenshot 2021-10-27 at 7 56 01 PM

* [x] Important to have the close button/icon visible clearly
* [x] It should basically look exactly like on our cloud (the above initial screen has too much of a padding-top/bottom for example):

image

## Questions to show

* [ ] The first question is selected randomly out of our pool of questions
* [x] We will rotate between below questions.  Meaning next time it shows again, we show a different question. 
* [x] Once all questions are shown, we start again with the first question.

### Questions

* What are your biggest problems or pain points with Matomo and why?
* What is the one thing you’d like to see improved in Matomo and why?
* What is the feature you are missing the most in Matomo and why?
* What is your main goal for using Matomo?
* What is your main concern about Matomo?

## When to show this

* [x] When the message is supposed to show, we show it again on every subsequent page view until they either clicked on close, or clicked on the button in the bar
  * [x] When someone closes it, we don’t show it for another 6 months. 
  * [ ] When someone clicks the feedback button, we don’t show it for another 6 months. 
     * (we could also change this to when user actually sent feedback in case they want to do it later)... or maybe if easy to do the popup could have a button "remind me later" and then we remind again in couple days -> let's maybe ignore it for now?
* [ ] We only show it on desktop, not on mobile/tablet because it might take too much space and close button might be hard to see/reach etc.
* [ ] We show it the first time 6 months after Matomo has been installed. This is important because we don't want to show it to all users at once to not get too much feedback at once, and because it means by then they have used Matomo for a while and know better what is possible and what is missing etc. 
  * [ ] Basically, on each day we show it to the people where it was installed 6 months ago. If someone installed it 7 months ago, then we don't want to show it. Otherwise if we rolled out this feature, then we would suddenly get feedback from all Matomo installations. Basically, we want to do something like `$daysSinceInstallation % 6 months == 0`. Problems: We don't have an install date I think, and what if they log in 6 months+3 days but not exactly 6 months ago to the day? We'd need to have some flexibility in there. Like we want to show it also to people that haven't logged in the last 1 week. If users only use it every few weeks that's fine then they might not see the popup. Likely they have little usage and requirements anyway. Maybe the next best thing is to check the first site creation date? (although when someone imports historical data then this may be changed). 
* [x] For now we only show it to super users, we might later expand to regular users depending on the amount of feedback we get

Also:
* [ ] We won't show this when WhiteLabel is installed (I believe this plugin might disable Feedback plugin anyway)
* [ ] To be defined: Maybe we also don't show it when there is any premium feature installed already?

## On button click to get actual feedback

We show our feedback popup 
image

* [x] The title of the popup is the same question we were showing in the UI initially
* [x] When it opens, the cursor should be already in the text field
* [x] We require at least 10 characters to be entered
* [x] We show a privacy policy with text eg `Your message and your email address will be sent to us by submitting this form. The information will be only used to improve our products and services. Your data will be processed by us, our hosting provider, and our support ticket provider. For more information have a look at Matomo's Privacy Policy.`  Link for privacy policy: https://matomo.org/privacy-policy/ 

## Confirmation message 

image

* [x] People might expect a response and send support questions. The idea was to show a message that we might reply if we need more information, to lower expectations of a reply. Note: Maybe at some point could also show a message that if they submitted a support question, to ask it in our forum instead and if they reported a bug, to create a github issue. We could set up a dedicated page for this and say something like `If your message was not feedback, please [read here]` (to be improved). 
* [x] The feedback is sent to us by email similar to the other feedback
* [x] The email includes the asked question, and the email of the user so we can follow up if needed

At a later stage we might ask a single select follow up question on the thank you confirmation message.

peterhashair:
### Description:
Fixes:#18251  #18250 #18232

merge @andyjdavis #18277 remove refer us to this branch. 

Please ignore the *umd.js file it VUE compile js, when doing a review. 

This PR includes: 
1. remove feedback popup and related(remove old angular js, PHP and related test and UI screen shoot) 
   `plugin/feedback/**`
2. add feedback question banner(VUE template, adapter, Php render, UI tests, integration Tests)
`plugin/feedback/tests/**`
3. extend icon with heart (icon-heart)
4. extend some date function with add month
5. extend get current user register date and last seen
7. extend modal adapter with validation, which does not close the modal. 

### Review


diosmosis:
@peterhashair will take a look at this now.

diosmosis:
@peterhashair The AJAX issue is fixed by this:

https://github.com/matomo-org/matomo/blob/vue-comparisons/plugins/CoreHome/vue/src/AjaxHelper/AjaxHelper.ts#L154

There are a couple fixes like this in that branch, I can create a quick PR for that specific one.

I'm not sure why `./console vue:build Feedback` doesn't work for you, it's working for me. Are you running it, then loading the app w/ development mode enabled? If so, can you check in the chrome sources whether the new file is actually loaded? If it's a file created with `--watch`, it will have a lot of `eval(...)`s.

diosmosis:
Also, @sgiehl would you have time to checkout this branch and see if you can reproduce the `vue:build Feedback` vs `vue:build --watch Feedback` issue?

peterhashair:
@diosmosis sorry, I may be confused you, the command itself works. But the interface looks different from watch. 
Here is the screenshot, I guess it maybe related to `<style scope>`
image

diosmosis:
@peterhashair oh I see, we're not supporting the use of styles from within Vue right now, instead we're still using the less files and core asset pipeline to define styling. This is mainly to keep the finished asset size down.

EDIT: If removing the style from the vue file doesn't change things, let me know.

peterhashair:
@diosmosis yes it's the line `<style scope>` that causes the problem.

plugins/Feedback/FeedbackReminder.php
```diff
@@ -10,6 +10,7 @@
 
 use Piwik\Piwik;
 use Piwik\Option;
+use Piwik\Site;
```
sgiehl:
```suggestion
```

plugins/Feedback/tests/UI/FeedbackQuestion_spec.js
```diff
@@ -0,0 +1,45 @@
...
+describe('FeedbackQuestion', function () {
+  this.timeout(0);
+  this.fixture = 'Piwik\\Plugins\\Feedback\\tests\\Fixtures\\FeedbackQuestionBannerFixture';
+
+  var url = '?module=CoreHome&action=index&idSite=1&period=day&date=2019-07-11&forceFeedbackTest=1';
+
+  before(async function () {
+    await page.webpage.setViewport({
+      width: 1250,
+      height: 768,
```
sgiehl: This one shouldn't be needed, unless you want to test that with a specific width. The tests should reset this after each test to `1350x768`

core/Date.php
```diff
@@ -985,6 +985,20 @@ public function addDay($n)
...
+     *
+     * @param int $n Number of days to add, must be > 0.
+     * @return \Piwik\Date
+     */
+
+    public function addMonth($n)
+    {
+        $ts = strtotime("+$n month", $this->timestamp);
+        return new Date($ts, $this->timezone);
+    }
```
sgiehl, line 1000: Be good to also add some simple tests here: https://github.com/matomo-org/matomo/blob/7cde69f763f2cbf2b9a27c8a7dcf7bcaa950c3bd/tests/PHPUnit/Unit/DateTest.php#L276

core/Date.php
```diff
@@ -985,6 +985,20 @@ public function addDay($n)
...
     }
 
+    /**
+     * Adds `$n` Month to `$this` date and returns the result in a new Date.
+     * instance.
+     *
+     * @param int $n Number of days to add, must be > 0.
+     * @return \Piwik\Date
+     */
+
```
bx80, line 995:
```suggestion
    */
    public function addMonth($n)
```

core/Piwik.php
```diff
@@ -179,6 +179,24 @@ public static function getCurrentUserEmail()
         return $user['email'] ?? '';
     }
 
+
+    public static function getCurrentUserCreationData()
+    {
+        $user = APIUsersManager::getInstance()->getUser(Piwik::getCurrentUserLogin());
+        return $user['date_registered'] ?? '';
+    }
```
bx80, line 187:
```suggestion
    }
    
```

plugins/Morpheus/javascripts/piwikHelper.js
```diff
@@ -275,6 +275,11 @@ window.piwikHelper = {
             $button.hide();
 
             var button = $('<a href="javascript:;" class="modal-action modal-close waves-effect waves-light btn-flat "></a>');
+
+            if(role === 'validation'){
+            button = $('<a href="javascript:;" class="modal-action waves-effect waves-light btn-flat "></a>');
+            }
```
bx80:
```suggestion
            if(role === 'validation'){
                button = $('<a href="javascript:;" class="modal-action waves-effect waves-light btn-flat "></a>');
            }
```

bx80 (commented):
Looks good to me, just flagged couple of minor code tidy ups.

peterhashair:
@bx80 good spot for the cancel and reopen the bug, fixed it :)

plugins/Feedback/vue/src/FeedbackQuestion/FeedbackQuestion.vue
```diff
@@ -0,0 +1,154 @@
...
+export default defineComponent({
+
+  props: {
+    showQuestionBanner: String,
+  },
+  components: {
+    MatomoDialog,
+  },
+  computed: {
+    isHide() {
```
sgiehl: Guess `isHidden` would be a better name

plugins/Feedback/Feedback.php
```diff
@@ -171,22 +109,41 @@ public function getShouldPromptForFeedback()
...
             return false;
         }
 
-        $now = Date::now()->getTimestamp();
+
+        // if user is not log in for a week. Display banner
+        $userLastSeen = Piwik::getCurrentUserLastSeen();
+        if (!empty($userLastSeen) && Date::factory($userLastSeen)->addDay(7)->getTimestamp() < $now) {
+            return true;
+        }
```
sgiehl: This can't work I guess. The last seen option is updated on `Platform.initialized`, which is done in `FrontController::init` and afaik happens before the render method of the Feedback plugin is triggered. Therefor I guess this can actually never be triggered. Btw. the `Platform.initialized` is even triggered on the 2FA page. So if someone has 2FA enabled the date is updated even before he finishes the login.

plugins/Feedback/Feedback.php
```diff
@@ -171,22 +109,41 @@ public function getShouldPromptForFeedback()
...
+        // if user is created less than 6 month ago, it won't show.
+        // I am not sure if this test really works. Because fake access is trade as isUserIsAnonymous
+        $userCreatedDate = Piwik::getCurrentUserCreationData();
+        if (!empty($userCreatedDate) && Date::factory($userCreatedDate)->addMonth(6)->getTimestamp() < $now) {
+            return true;
+        }
```
sgiehl: I'm a bit confused. Should the banner show if a user was created within the last 6 months or not? The comment says no, but the code isn't clear. Shouldn't it return false, if the users is newer, to prevent the code from returning `true` as default? Currently this actually forces the banner to be displayed always for users older than 6 months. Also I wonder if it shouldn't show for newer users, why should it then show for such users, if they weren't logged for 7 days (even though that doesn't work yet)

sgiehl: On the other side I guess this would actually only be needed to exclude the banner from being shown for new users if they didn't login recently. In all other cases it shouldn't be shown. If a new user is created and logs in the first time, the reminder date will be set to 6 months in the future. So the first time it would be shown, will always be 6 months after the creating date (unless someone removes the option from the database manually.)

plugins/Feedback/Feedback.php
```diff
@@ -171,22 +109,41 @@ public function getShouldPromptForFeedback()
...
+
         $nextReminderDate = Date::factory($nextReminderDate);
+        // if user click close extend to 6 moth
```
sgiehl: The comment sounds like the date would be updated, but it actually isn't updated here.

plugins/Feedback/Feedback.php
sgiehl: Looking at the requirement I guess this check is not even needed. I guess it wasn't meant that someone always gets the banner if he wasn't logged in for a week.  It was more like that he doesn't need to log in exactly after 6 months to get the banner at all I guess.

sgiehl (changes requested):
Left a couple of comments.
Regarding the feature itself a couple of questions came to my mind while testing:
- Is it on purpose that the new feedback question will actually never be shown to a user who clicked `Never show again` in the old feedback popup? (The new banner actually doesn't have that feature)
- The new feature uses the old option. So I guess this actually already kind of spreads the feedback at least for those users that didn't click on `never shown again`
- The popup says the email address will be sent with the details. I was wondering if we should display the mail address, and maybe even give the user the possibility to change it. Was just wondering if it might prevent people that might use a shared account, or shared email address  from clicking the send button, as they don't want to receive mails on the configured address.
- According to #18251 the banner should not be shown on mobile. Can't see any code to prevent that.

peterhashair:
@sgiehl yeah, that logic on `banner show` code is a little bit confusing. Just update it, hopefully, it makes sense now. 😀

bx80 (approved):

bx80:
@peterhashair One minor thing I noticed - the post question 'thanks' dialog doesn't have a 'Close' or 'Ok' button. Not sure if that is intentional?
image

peterhashair:
@bx80 yes, that makes sense, added a close button to the finished screen.

peterhashair:
@sgiehl sorry to bother you, do you want to do a final check on that one 😀

plugins/Feedback/lang/en.json
```diff
@@ -43,12 +42,16 @@
...
+        "Question0": "What are your biggest problems or pain points with Matomo and why?",
+        "Question1": "What is the one thing you’d like to see improved in Matomo and why?",
+        "Question2": "What is the feature you are missing the most in Matomo and why?",
+        "Question3": "What is your main goal for using Matomo?",
+        "Question4": "What is your main concern about Matomo?",
+        "FeedbackTitle": "Please help us improve Matomo",
+        "FeedbackSubtitle": "Here's your chance to ask for it! Please describe in as much details as you can. <br/> Your feedback will help us improve Matomo. %1$s",
+        "ThankYourForFeedback": "We really appreciate your feedback! %1$s We will be in touch in case we require more information about your feedback.",
+        "Policy": "Your message and your email address will be sent to us by submitting this form. The information will be only used to improve our products and services. Your data will be processed by us, our hosting provider, and our support ticket provider. For more information have a look at %1$sMatomo's Privacy Policy%2$s.",
+        "FormEmptyBody": "The message body can't not be blank."
```
sgiehl:
```suggestion
        "FormEmptyBody": "The message body can't be blank."
```

plugins/Feedback/vue/src/FeedbackQuestion/FeedbackQuestion.vue
```diff
@@ -0,0 +1,159 @@
...
+      expanded: false,
+      showFeedbackForm: false,
+      feedbackMessage: null,
+      errorMessage: null,
+    };
+  },
+  watch: {
+    showFeedbackForm() {
+      // eslint-disable-next-line no-underscore-dangle
+      this.questionText = window._pk_translate(`Feedback_Question${this.question}`);
```
sgiehl: Guess you could actually do a `import translate from '../translate';` and use `translate()` instead.

plugins/Feedback/vue/src/FeedbackQuestion/FeedbackQuestion.vue
```diff
@@ -0,0 +1,159 @@
...
+      const currentCookie = document.cookie.match(`(^|;)\\s*${cookieName}\\s*=\\s*([^;]+)`);
+      return currentCookie ? currentCookie.pop() : null;
+    },
+    setCookieValue(value) {
+      const now = new Date();
+      const time = now.getTime();
+      const expireTime = time + 1000 * 36000;
+      now.setTime(expireTime);
+      document.cookie = `${cookieName}=${value};expires=${now.toUTCString()};path=/`;
+    },
```
sgiehl: Might it maybe make sense to have some global class for handling cookie related stuff in vue (e.g. get/set/remove)? I guess we might maybe need that in other places as well and it might not make sense to implement that again every time.

plugins/Feedback/API.php
```diff
@@ -60,6 +62,71 @@ public function sendFeedbackForFeature($featureName, $like, $message = false)
...
+
+        // if message is test content then don't send email
+        if ($message == 'test') {
+            $feedbackReminder = new FeedbackReminder();
+            $feedbackReminder->setUserOption("-1");
+            return 'success';
+        }
+
+        Piwik::checkUserIsNotAnonymous();
+        Piwik::checkUserHasSomeViewAccess();
```
sgiehl, line 77: This should always be done at top of the method, before validating anything. And I guess `Piwik::checkUserHasSomeViewAccess();` might be unneeded here, but actually doesn't hurt either.

plugins/Feedback/API.php
```diff
@@ -60,6 +62,71 @@ public function sendFeedbackForFeature($featureName, $like, $message = false)
...
+    /**
+     * Sends feedback for a specific feature to the Matomo team or alternatively to the email address configured in the
+     * config: "feedback_email_address".
+     *
+     * @param $question
+     * @param string|bool $message A message containing the actual feedback
+     * @throws \Piwik\NoAccessException
+     * @throws \Exception
+     */
+    public function sendFeedbackForSurvey($question,  $message = false)
```
sgiehl, line 74: Actually I'm wondering if should implement some CSRF token in the form, to prevent that someone could call that method with custom content. Otherwise you could call that method in the url directly and kind of spam us with feedback.
Alternatively we could check if the feedbackreminder option has a value that triggers the feedback popup and don't accept it otherwise maybe.

plugins/Feedback/vue/src/FeedbackQuestion/FeedbackQuestion.vue
```diff
@@ -0,0 +1,159 @@
...
+    showFeedbackForm() {
+      // eslint-disable-next-line no-underscore-dangle
+      this.questionText = window._pk_translate(`Feedback_Question${this.question}`);
+    },
+  },
+  created() {
+    if (this.getCookieValue(cookieName)) {
+      // eslint-disable-next-line radix
+      this.question = parseInt(this.getCookieValue(cookieName));
+      const nextQuestion = (this.question + 1 > 4) ? 0 : this.question + 1;
```
sgiehl: This is actually always executed even if the feedback banner isn't shown. Maybe `created` isn't the correct place for this.

sgiehl (changes requested):
Left some more comments. Also my questions/comments from the last review are still "open":

> - Is it on purpose that the new feedback question will actually never be shown to a user who clicked Never show again in the old feedback popup? (The new banner actually doesn't have that feature)
> - The new feature uses the old option. So I guess this actually already kind of spreads the feedback at least for those users that didn't click on never shown again
> - The popup says the email address will be sent with the details. I was wondering if we should display the mail address, and maybe even give the user the possibility to change it. Was just wondering if it might prevent people that might use a shared account, or shared email address from clicking the send button, as they don't want to receive mails on the configured address.
> - According to #18251 the banner should not be shown on mobile. Can't see any code to prevent that.

In addition to those I got some new ones:

- The content of the feedback popup will get translated into multiple languages. Currently we do not mention something like `Please write your message in English if possible`, so it would likely happen that we will receive messages in various languages, not sure if we want that.
- The question shown in the banner is currently rotated not with the next request. The first request sets a cookie with value `0`, so the next request uses the same question again.
- Currently when someone sends feedback, he will never see the feedback banner again. Is that on purpose? I thought in the issue it said to be shown again after 6 months as well.
- Additionally some of the other requirements of #18251 aren't met:
  - When it opens, the cursor should be already in the text field
  - We require at least 10 characters to be entered
  - The first question is selected randomly out of our pool of questions
  - For now we only show it to super users, we might later expand to regular users depending on the amount of feedback we get

plugins/Feedback/API.php
peterhashair: @sgiehl ah, I think it needs CSRF or Recaptcha as well, but then I saw in the post it actually POST `token_auth` in the form, should I add `$this->checkTokenInUrl()` to here? I mean copy the function from the controller.php

peterhashair:
there could be a bug, because I didn't record the question they answers, after 6 months, they maybe answer the same question again.

peterhashair:
@sgiehl  I think this is ready for another review 😀

core/Plugin/API.php
```diff
@@ -9,7 +9,10 @@
 
 namespace Piwik\Plugin;
 
+use Piwik\Common;
 use Piwik\Container\StaticContainer;
+use Piwik\NoAccessException;
+use Piwik\Piwik;
```
sgiehl:
```suggestion
use Piwik\Container\StaticContainer;
```

plugins/CoreHome/vue/src/CookieHelper/CookieHelper.adapter.ts
```diff
@@ -0,0 +1,11 @@
...
+
+function CookieHelper() {
+  return {
+    setCookie,
+    getCookie,
+    deleteCookie,
+  };
+}
+
+window.angular.module('piwikApp.service').factory('cookieHelper', CookieHelper);
```
sgiehl: Do we need to have an angularjs adapter for a fully new vue component?

plugins/Feedback/API.php
sgiehl: @peterhashair No. API methods actually should automatically check for a valid session or a valid token_auth, so no need to check for the token_auth. But thinking a bit more about that I guess we actually won't need an additional check. I guess if someone really wants to spam us, there are simpler ways than triggering the API method.

plugins/CoreHome/vue/src/CookieHelper/CookieHelper.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ * General utils for managing cookies in Typescript.
+ */
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export function setCookie(name: string, val: string, day: number) {
```
sgiehl: I guess the most common definition uses seconds or milliseconds to define the expiration time. Also it should maybe be possible to leave the expiration time blank, to set a cookie with session lifetime.

plugins/CoreHome/vue/src/CookieHelper/CookieHelper.ts
```diff
@@ -0,0 +1,37 @@
...
+  const value = `; ${document.cookie}`;
+  const parts = value.split(`; ${name}=`);
+
+  // eslint-disable-next-line eqeqeq
+  if (parts.length == 2) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    return parts.pop().split(';').shift();
+  }
+}
```
sgiehl, line 37: Just out of curiosity as I'm not yet such familiar with typescript and the resulting javascript. Will that method return `undefined` when no cookie is available or is it `null`. Just wondering what you would need to compare against to check if no cookie was set.

plugins/Feedback/Feedback.php
```diff
@@ -171,22 +113,35 @@ public function getShouldPromptForFeedback()
...
-            $nextReminder = Date::now()->getStartOfDay()->addDay(90)->toString('Y-m-d');
-            $feedbackReminder->setUserOption($nextReminder);
 
+            // if user is created more than 6 month ago, set reminder to today and show banner
+            $userCreatedDate = Piwik::getCurrentUserCreationData();
+            if (!empty($userCreatedDate) && Date::factory($userCreatedDate)->addMonth(6)->getTimestamp() < $now) {
+                $nextReminder = Date::now()->getStartOfDay()->subDay(1)->toString('Y-m-d');
+                $feedbackReminder->setUserOption($nextReminder);
+                return true;
+            }
```
sgiehl, line 132: I think we should also handle the case, where a user was created within the last 6 months and set the reminder date to `creation_date + 6 months`.

plugins/Feedback/lang/en.json
```diff
@@ -43,12 +42,16 @@
...
         "SearchHelpResources": "Search matomo.org help resources",
         "PopularHelpTopics": "Popular help topics",
-        "ReferMatomo": "Refer Matomo",
-        "ReferBannerTitle": "Believe in Matomo?",
-        "ReferBannerLonger": "Please help us spread the word about us so more people get control over their analytics data.",
-        "ReferBannerEmailShareSubject": "Refer them to Matomo Analytics now to take back control!",
-        "ReferBannerEmailShareBody": "I choose Matomo, an ethical alternative to Google Analytics that gives me 100%% data ownership and protects the data of my website visitors.\r\nI’m sharing this message in the hope that you too will take back the power from Google and get complete ownership of your own data.\r\n\r\nCheck out Matomo at https://matomo.org",
-        "ReferBannerSocialShareText": "If you’re using Google Analytics, stay in control by switching to an ethical alternative like Matomo now!",
-        "WontShowAgain": "We won't show this message again."
+        "WontShowAgain": "We won't show this message again.",
```
sgiehl, line 45: This message is now actually unused and can be removed.
```suggestion
```

plugins/Feedback/Feedback.php
```diff
@@ -65,97 +62,42 @@ public function getClientSideTranslationKeys(&$translationKeys)
         $translationKeys[] = 'Feedback_PleaseLeaveExternalReviewForMatomo';
         $translationKeys[] = 'Feedback_RemindMeLater';
         $translationKeys[] = 'Feedback_NeverAskMeAgain';
-        $translationKeys[] = 'Feedback_ReferMatomo';
-        $translationKeys[] = 'Feedback_ReferBannerTitle';
-        $translationKeys[] = 'Feedback_ReferBannerLonger';
-        $translationKeys[] = 'Feedback_ReferBannerSocialShareText';
-        $translationKeys[] = 'Feedback_ReferBannerEmailShareSubject';
-        $translationKeys[] = 'Feedback_ReferBannerEmailShareBody';
         $translationKeys[] = 'Feedback_WontShowAgain';
```
sgiehl, line 65: This key isn't used anymore.
```suggestion
```

plugins/Feedback/Feedback.php
```diff
@@ -65,97 +62,42 @@ public function getClientSideTranslationKeys(&$translationKeys)
         $translationKeys[] = 'Feedback_PleaseLeaveExternalReviewForMatomo';
         $translationKeys[] = 'Feedback_RemindMeLater';
         $translationKeys[] = 'Feedback_NeverAskMeAgain';
```
sgiehl, line 64: This one isn't used anymore and can also be removed from `en.json`
```suggestion
```

plugins/Feedback/Feedback.php
```diff
@@ -171,22 +113,35 @@ public function getShouldPromptForFeedback()
 
         $feedbackReminder = new FeedbackReminder();
         $nextReminderDate = $feedbackReminder->getUserOption();
+        $now = Date::now()->getTimestamp();
 
+        //user answered question
         if ($nextReminderDate === self::NEVER_REMIND_ME_AGAIN) {
             return false;
         }
```
sgiehl, line 121: Do we still provide an option to never show that again? Can't see any possibility in the code (besides writing `test` as feedback). If this isn't used anymore we should remove that code. (The constant can be removed as well)
```suggestion
```

plugins/Feedback/API.php
```diff
@@ -60,6 +62,73 @@ public function sendFeedbackForFeature($featureName, $like, $message = false)
...
+        Piwik::checkUserIsNotAnonymous();
+        Piwik::checkUserHasSomeViewAccess();
+
+        // if message is test content then don't send email
+        if ($message == 'test') {
+            $feedbackReminder = new FeedbackReminder();
+            $feedbackReminder->setUserOption("-1");
+            return 'success';
+        }
+
```
sgiehl: Personally I don't like such "special" test case handling in normal code if there are other ways possible. If you want to prevent any mails being sent in a test, you can easily achieve that by setting the config `[General] emails_enabled = 0` or by observing the event `Mail.shouldSend`

Btw. for your local installation I would suggest to use something like a mailcatcher to ensure no mails are sent anywhere while testing.

```suggestion
```

plugins/Feedback/API.php
```diff
@@ -60,6 +62,73 @@ public function sendFeedbackForFeature($featureName, $like, $message = false)
...
+
+        $body = sprintf("Question: %s\n", $featureName);
+
+        $feedbackMessage = "";
+        if (!empty($message) && $message != 'undefined') {
+            $feedbackMessage = sprintf("Answer:\n%s\n", trim($message));
+        }
+        $body .= $feedbackMessage ? $feedbackMessage : " \n";
+
```
sgiehl: just some style improvements

```suggestion

        $featureName = $this->getEnglishTranslationForFeatureName($question);
        $body = sprintf("Question: %s\n", $featureName);
        $feedbackMessage = "";
        
        if (!empty($message) && $message !== 'undefined') {
            $feedbackMessage = sprintf("Answer:\n%s\n", trim($message));
        }

        $body .= $feedbackMessage ? $feedbackMessage : " \n";

```

plugins/Feedback/Feedback.php
```diff
@@ -171,22 +113,35 @@ public function getShouldPromptForFeedback()
...
 
+        // if is new user or old user field not exist
         if ($nextReminderDate === false) {
```
sgiehl: If we remove the never remind me option check above, but still use the old option, we actually need to handle that here I guess
```suggestion
        if ($nextReminderDate === false || $nextReminderDate <= 0) {
```

plugins/Feedback/vue/src/FeedbackQuestion/FeedbackQuestion.vue
```diff
@@ -0,0 +1,173 @@
...
+  methods: {
+    initQuestion() {
+      if (getCookie(cookieName)) {
+        // eslint-disable-next-line radix
+        this.question = parseInt(getCookie(cookieName));
+        const nextQuestion = (this.question + 1 > 4) ? 0 : this.question + 1;
+        setCookie(cookieName, nextQuestion, 7);
+      } else {
+        setCookie(cookieName, this.getRandomIntBetween(0, 4), 7);
+      }
```
sgiehl: Actually the first question shown is still always number 0. Guess this is as initially there is no cookie, and the cookie set in that case is then used for the next page load. Guess this could be fixed and simplified with this:
```suggestion
      if (!getCookie(cookieName)) {
        this.question = this.getRandomIntBetween(0, 4);
      } else {
        // eslint-disable-next-line radix
        this.question = parseInt(getCookie(cookieName));
      }
      
      const nextQuestion = (this.question + 1) % 4;
      setCookie(cookieName, nextQuestion, 7);
```

plugins/Feedback/vue/src/FeedbackQuestion/FeedbackQuestion.vue
```diff
@@ -0,0 +1,173 @@
...
+    },
+  },
+  created() {
+    if (this.showQuestionBanner !== '0') {
+      // eslint-disable-next-line no-restricted-globals
+      if (location.href.includes('forceFeedbackTest')) {
+        this.question = 0;
+        setCookie(cookieName, 0, 7);
+      } else {
```
sgiehl: I guess this one is to ensure the UI test always shows the same question? Guess this could be achieved also by setting the cookie in the UI test using [`page.setCookie`](https://pptr.dev/#?product=Puppeteer&version=v10.2.0&show=api-pagesetcookiecookies)

sgiehl (changes requested):
Had another look at the changes. In general most stuff seems to work as expected now :+1:
Left a couple of comments. Most of them are actually related to code cleanup and improvements. 
Guess afterwards this will be ready to merge

plugins/Feedback/vue/src/FeedbackQuestion/FeedbackQuestion.vue
peterhashair: @sgiehl I tried set page.setCookie in the Test, but its return method was not found. Then I tried to add `setCookies` in `tests/lib/screenshot-testing/support/page-renderer.js` still no luck

sgiehl: You may need to add `setCookie` to the list of proxied methods here: https://github.com/matomo-org/matomo/blob/4.x-dev/tests/lib/screenshot-testing/support/page-renderer.js#L53
If that method is not proxied it will only be available through `page.webpage.setCookie`

PR #18279, opened 2021-11-08
add disable archiving segment for plugins

### LINKED ISSUE #18134: Add new configuration to disable archiving the segment reports for specific plugins
We noticed some plugins may cause a lot of CPU load. A solution is to disable a plugin that causes a lot of load. However, sometimes it be great to still have at least report data for "All visits" but not segments. This way say 95% of the load is reduced while the biggest value might be still there.

It be great if we had a new config ini flag to disable archiving the archiving for some plugins when the archiving is for any segment. 

When this is disabled for a plugin then:

* Don't archive any data for this plugin if segment is not empty
* In the report itself, when a segment is viewed and there is no data, then show a footer message in the report that viewing segments for this report is disabled.
* Reports for "all visits" (segment is empty) should still be archived regularly

Basically, a config setting like 

```
[General]
disable_archiving_segment_for_plugin="Plugin1,Plugin2,Plugin3"
```

Where we can provide the name of plugins in a comma separated list. Could also use an array if that makes it easier like
 
```
[General]
disable_archiving_segment_for_plugin[]="Plugin1"
disable_archiving_segment_for_plugin[]="Plugin1"
disable_archiving_segment_for_plugin[]="Plugin1"
```

The config name is just a random thought without too much thinking. Feel free to change the name.

In tracking we can configure some settings on a per site basis see https://matomo.org/faq/how-to/how-do-i-configure-a-config-setting-in-the-config-ini-php-differently-for-a-specific-site/

If we could do this for this setting as well that would be even better. Eg if we could set below setting to only apply this for site 42 that would be even better.

```
[General_42]
disable_archiving_segment_for_plugin[]="Plugin1"
```

peterhashair:
### Description:
Fixes: #18134
add disable archiving segment configuration for plugins

- Under `[General]`-> `disable_archiving_segment_for_plugins` accept string, sting with comma, array. 
- `[General_{siteId}]` can be enable by Site ID. 

I guess this will be 4.7

### Review


config/global.ini.php
```diff
@@ -385,6 +385,14 @@
 ; This feature will not work with the MYSQLI extension.
 archiving_query_max_execution_time = 7200
 
+
+; Allow you to disable archiving segment on selected plugins
+; When this is on, it won't archive any data for selected plugin if segment is not empty
+; This field accept split by comma eg: disable_archiving_segment_for_plugin="Plugin1,Plugin2,Plugin3"
```
justinvelluppillai: Maybe better "This field accepts more than one plugin separated by comma eg:..."

config/global.ini.php
```diff
@@ -385,6 +385,14 @@
...
+; This field also accept array format eg: disable_archiving_segment_for_plugin[]="Plugin1", disable_archiving_segment_for_plugin[]="Plugin2"
```
justinvelluppillai: Should be "accepts"

config/global.ini.php
```diff
@@ -385,6 +385,14 @@
...
+; if you like to disable on specify site, please append the siteId to the [General_siteId]. eg siteId=2 [General_2]
```
justinvelluppillai: "if you'd like to disable on a specific site"

config/global.ini.php
```diff
@@ -385,6 +385,14 @@
 ; This feature will not work with the MYSQLI extension.
 archiving_query_max_execution_time = 7200
 
+
+; Allow you to disable archiving segment on selected plugins
+; When this is on, it won't archive any data for selected plugin if segment is not empty
+; This field accepts more than one plugin separated by comma eg: disable_archiving_segment_for_plugin="Plugin1,Plugin2,Plugin3"
+; This field also accepts array format eg: disable_archiving_segment_for_plugin[]="Plugin1", disable_archiving_segment_for_plugin[]="Plugin2"
+; if you'd like to disable on specify site, please append the siteId to the [General_siteId]. eg siteId=2 [General_2]
```
bx80:
```suggestion
; if you'd like to disable on a specific site, please append the siteId to the [General_siteId]. eg siteId=2 [General_2]
```

bx80 (approved):

peterhashair:
The last test runs fully passed [https://app.travis-ci.com/github/matomo-org/matomo/builds/241828886](https://app.travis-ci.com/github/matomo-org/matomo/builds/241828886). the Last update just one wording in global.config.php, I think no need to run the test again.

justinvelluppillai:
@peterhashair changing the config changes one of the screenshots

peterhashair:
@justinvelluppillai right, let me run it again.

sgiehl:
> [General_{siteId}] can be enable by Site ID.

If I see that right, the site specific config would then only work for that specific configuration value, right? Imho we should do it the same way like in tracker config. So it would be available for all values in General section.

See https://github.com/matomo-org/matomo/blob/e0417845f8fd7031e00e97c3576415c998abb2ce/core/Tracker/TrackerConfig.php#L28-L47

I'm not sure if it's easy to adjust the config class to support that. Maybe we need to introduce a new class `GeneralConfig` (like `TrackerConfig`) and replace all usages of the general config over time then...

peterhashair:
@sgiehl  right, that makes sense let me do the adjustment.

core/ArchiveProcessor/Rules.php
```diff
@@ -356,4 +364,41 @@ public static function shouldProcessSegmentsWhenReArchivingReports()
...
+                $pluginArchivingSetting = explode(",", "$pluginArchivingSetting");
+            }
+            if (in_array($pluginName, $pluginArchivingSetting)) {
+                return true;
+            }
+        }
+
+        if (!$siteId) {
+            return false;
+        }
```
sgiehl: Shouldn't this part now be obsolete? The `GeneralConfig` below should now return either the site specific or the global config

config/global.ini.php
```diff
@@ -385,6 +385,14 @@
 ; This feature will not work with the MYSQLI extension.
 archiving_query_max_execution_time = 7200
 
+
+; Allow you to disable archiving segment on selected plugins
+; When this is on, it won't archive any data for selected plugin if segment is not empty
+; This field accepts more than one plugin separated by comma eg: disable_archiving_segment_for_plugin="Plugin1,Plugin2,Plugin3"
+; This field also accepts array format eg: disable_archiving_segment_for_plugin[]="Plugin1", disable_archiving_segment_for_plugin[]="Plugin2"
+; if you'd like to disable on a specify site, please append the siteId to the [General_siteId]. eg siteId=2 [General_2]
```
sgiehl: If we aim to support this feature for all settings in General config, we maybe don't need to mention it in each setting. 🤔

sgiehl (changes requested):

peterhashair:
@sgiehl yes, don't need load `General`. sorry, my brain was dead there 😅

config/global.ini.php
peterhashair: @sgiehl do we want to remove this part?

sgiehl: Might be better to mention that somewhere else, like in the FAQ

peterhashair: @sgiehl right that makes sense, let me convert to FAQ tomorrow. 👍

peterhashair:
FAQ for review https://matomo.org/?post_type=faq&p=47059&preview=true

config/global.ini.php
```diff
@@ -385,6 +385,10 @@
 ; This feature will not work with the MYSQLI extension.
 archiving_query_max_execution_time = 7200
 
+
+; Allow you to disable archiving segment on selected plugins for more details please see https://matomo.org/faq/how-to-disable-archiving-the-segment-reports-for-specific-plugins
```
sgiehl:
```suggestion
; Allows you to disable archiving segments for selected plugins. For more details please see https://matomo.org/faq/how-to-disable-archiving-the-segment-reports-for-specific-plugins
```

core/ArchiveProcessor/Rules.php
```diff
@@ -356,4 +364,24 @@ public static function shouldProcessSegmentsWhenReArchivingReports()
...
+
+        if (is_string($pluginArchivingSetting)) {
+            $pluginArchivingSetting = explode(",", "$pluginArchivingSetting");
+        }
+
+        if (in_array($pluginName, $pluginArchivingSetting)) {
+            return true;
+        }
+        return false;
```
sgiehl: This can be simplified as:

```suggestion
        return in_array($pluginName, $pluginArchivingSetting);
```

tests/PHPUnit/Integration/ArchiveProcessor/LoaderTest.php
```diff
@@ -1463,6 +1463,56 @@ public function test_canSkipArchiveForSegment_returnsTrueIfHasNoInvalidationForR
...
+        $params->setRequestedPlugin('testPlugin');
+        $params->setArchiveOnlyReport('myReport');
+        $loader = new Loader($params);
+        $this->assertTrue($loader->canSkipArchiveForSegment());
+    }
+
+    public function test_canSkipArchiveForSegment_returnTrueIfPluginIsDisabledBySiteId()
+    {
+        Rules::setBrowserTriggerArchiving(false);
+        Config::setSetting('General_1','disable_archiving_segment_for_plugins','testPlugin');
```
sgiehl, line 1491: Might be good to also add a test if overwriting works correctly. e.g. a config like

```
[General]
disable_archiving_segment_for_plugins = myplugin

[General_1]
disable_archiving_segment_for_plugins = testplugin
```

Guess for idSite=1 only `testplugin` should then be skipped and not `myplugin`

core/ArchiveProcessor/Rules.php
```diff
@@ -356,4 +364,24 @@ public static function shouldProcessSegmentsWhenReArchivingReports()
...
+
+        if (empty($pluginArchivingSetting)) {
+            return false;
+        }
+
+        if (is_string($pluginArchivingSetting)) {
+            $pluginArchivingSetting = explode(",", "$pluginArchivingSetting");
+        }
+
+        if (in_array($pluginName, $pluginArchivingSetting)) {
```
sgiehl: @tsteur should we compare the plugin name case insensitive here? Or should that only work if the plugin name is exactly the same?

sgiehl (changes requested):
Left a couple of minor comments, afterwards I guess this should be ready to merge

core/ArchiveProcessor/Rules.php
tsteur: No big preference. Maybe case insensitive be better if done easily.

peterhashair:
@sgiehl update a little bit, 
-  add site Id configure test overwrite general test. 
- case insensitive, 
- special case ignored.

peterhashair:
@sgiehl add test if overwriting to here  [https://github.com/matomo-org/matomo/blob/d6b4e69967233a8181e23bfa2295980fbfb26614/tests/PHPUnit/System/ArchiveInvalidationTest.php#L198-L201](https://github.com/matomo-org/matomo/blob/d6b4e69967233a8181e23bfa2295980fbfb26614/tests/PHPUnit/System/ArchiveInvalidationTest.php#L198-L201)

core/ArchiveProcessor/Rules.php
```diff
@@ -356,4 +364,29 @@ public static function shouldProcessSegmentsWhenReArchivingReports()
...
+            //special case return false
+            if (preg_match('/[\'^£$%&*()}{@#~?><>|=_+¬-]/', $pluginArchivingSetting)) {
+                return false;
+            }
+
+            $pluginArchivingSetting = explode(",", "$pluginArchivingSetting");
+        }
+
+        //case-insensitive search
+        return (bool)preg_grep('/'.$pluginName.'/i',$pluginArchivingSetting);
```
sgiehl: Won't that also return partial matches. e.g `testplugin` would also match if `testplugin2` would be disabled, right?

core/ArchiveProcessor/Rules.php
```diff
@@ -356,4 +364,29 @@ public static function shouldProcessSegmentsWhenReArchivingReports()
...
+            return false;
+        }
+
+
+        if (is_string($pluginArchivingSetting)) {
+
+            //special case return false
+            if (preg_match('/[\'^£$%&*()}{@#~?><>|=_+¬-]/', $pluginArchivingSetting)) {
+                return false;
+            }
```
sgiehl: To check for valid plugin name there is also `Plugin\Manager::isValidPluginName`

tests/PHPUnit/System/ArchiveInvalidationTest.php
```diff
@@ -128,6 +140,70 @@ public function testAnotherApi()
...
+        //test siteId 2 not affect siteId1
+        Config::setSetting('General_2', 'disable_archiving_segment_for_plugins','testPlugin' );
+        $this->assertFalse(Rules::isSegmentPluginArchivingDisabled('testPlugin',1));
+
+
+        //test general setting not affect siteId1
+        Config::setSetting('General', 'disable_archiving_segment_for_plugins','myPlugin' );
+        Config::setSetting('General_1', 'disable_archiving_segment_for_plugins','testPlugin' );
+        $this->assertFalse(Rules::isSegmentPluginArchivingDisabled('myPlugin',1));
+
```
sgiehl, line 202: Maybe a test to ensure no partial matches work would be good:

```suggestion

        Config::setSetting('General_1', 'disable_archiving_segment_for_plugins', 'testPlugin2');
        $this->assertFalse(Rules::isSegmentPluginArchivingDisabled('testPlugin', 1));
```

core/ArchiveProcessor/Rules.php
```diff
@@ -356,4 +364,29 @@ public static function shouldProcessSegmentsWhenReArchivingReports()
...
+
+            $pluginArchivingSetting = explode(",", "$pluginArchivingSetting");
+        }
+
+        //case-insensitive search
+        return (bool)preg_grep('/'.$pluginName.'/i',$pluginArchivingSetting);
+
+    }
```
sgiehl: To fix the previous mentioned problems I would suggest to use something like this:

```suggestion
    public static function isSegmentPluginArchivingDisabled($pluginName, $siteId = null)
    {
        $pluginArchivingSetting = GeneralConfig::getConfigValue('disable_archiving_segment_for_plugins', $siteId);

        if (empty($pluginArchivingSetting)) {
            return false;
        }

        if (is_string($pluginArchivingSetting)) {
            $pluginArchivingSetting = explode(",", $pluginArchivingSetting);
            $pluginArchivingSetting = array_filter($pluginArchivingSetting, function($plugin){
                return Manager::getInstance()->isValidPluginName($plugin);
            });
        }

        $pluginArchivingSetting = array_map('strtolower', $pluginArchivingSetting);

        return in_array(strtolower($pluginName), $pluginArchivingSetting);
    }
```

sgiehl (changes requested):

peterhashair:
@sgiehl that make more sense. 👍

sgiehl (dismissed):
Should be good to merge now if tests are passing.

sgiehl:
@peterhashair Just had another look at the original issue and it seems there is one requirement, that isn't fulfilled currently:

> In the report itself, when a segment is viewed and there is no data, then show a footer message in the report that viewing segments for this report is disabled.

If you need some hints to find the correct places to implement this. Maybe start having a look here:
https://github.com/matomo-org/matomo/blob/5d9cf649aeefa7a1390574c2b147ce972b1b5c39/core/Plugin/Visualization.php#L228-L235
https://github.com/matomo-org/matomo/blob/1023fb7d691ebf47c27aa8a1cab4702c55c2630d/plugins/CoreHome/templates/_dataTable.twig#L58-L68

peterhashair:
@sgiehl I did a mock-up for the requirement. Not sure that's the right approach for this, do I need to implement a UI test for that one?

> In the report itself, when a segment is viewed and there is no data, then show a footer message in the report that viewing segments for this report are disabled.

core/Plugin/Visualization.php
```diff
@@ -600,6 +602,17 @@ private function hasReportBeenPurged()
         return PrivacyManager::hasReportBeenPurged($this->dataTable);
     }
 
+
+    private function hasReportSegmentDisabled()
+    {
+        $module = $this->requestConfig->getApiModuleToRequest();
+        var_dump($module);
```
sgiehl:
```suggestion
```

plugins/CoreHome/lang/en.json
```diff
@@ -11,6 +11,7 @@
         "CloseWidgetDirections": "You can close this widget by clicking on the 'X' icon at the top of the widget.",
         "ChooseX": "Choose %1$s",
         "DataForThisReportHasBeenPurged": "The data for this report is more than %s months old and has been purged.",
+        "DataForThisReportHasBeenDisabled": "The archive for this plugin is disabled, please check %1$shttps:\/\/matomo.org\/faq\/how-to-disable-archiving-the-segment-reports-for-specific-plugins%2$s for more info",
```
sgiehl: We should try to write messages that are useful for all kind of users. This one might be a bit too technical and the FAQ link might only be relevant for the admin. Guess something simple like `Segmentation is currently disabled for this report` would do it. Not sure if we maybe should mention to contact the admin for details 🤔 Maybe @tsteur has an even better suggestion for that.

And a general thing: We should not include an URL in a translation, as that might be error prone for translations. In that case you could have only used one placeholder to include everything, or link a text like `this FAQ` instead.

core/Plugin/Visualization.php
```diff
@@ -600,6 +602,17 @@ private function hasReportBeenPurged()
...
+        if (Rules::isSegmentPluginArchivingDisabled($module)) {
```
sgiehl: You also need to check if a segment was actually applied to this report, Otherwise it will also be shown for unsegmented reports, that have no data for other reasons.

plugins/CoreHome/templates/_dataTable.twig
```diff
@@ -59,6 +59,8 @@
                 <div class="pk-emptyDataTable">
                 {% if showReportDataWasPurgedMessage is defined and showReportDataWasPurgedMessage %}
                     {{ 'CoreHome_DataForThisReportHasBeenPurged'|translate(deleteReportsOlderThan) }}
+                {% elseif showPluginArchiveDisabled %}
```
sgiehl: Might be better to check if the variable is defined, otherwise it might throw an exception if that is not the case.
```suggestion
                {% elseif showPluginArchiveDisabled is defined and showPluginArchiveDisabled %}
```

plugins/CoreHome/templates/_dataTable.twig
```diff
@@ -59,6 +59,8 @@
...
+                    {{ 'CoreHome_DataForThisReportHasBeenDisabled'|translate('<a target="_blank" href="'~'https://matomo.org/faq/how-to-disable-archiving-the-segment-reports-for-specific-plugins'|e('html_attr') ~'">', '</a>') }}
```
sgiehl: In case you output a translation in a twig template that contains html, you need to apply the `raw` filter, otherwise the HTML is outputted encoded.

sgiehl (commented):
@peterhashair your approach has the correct direction ;-)

core/Config/GeneralConfig.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public static function getConfigValue($name, $idSite = null)
+    {
+        $config = self::getConfig();
+        if (!empty($idSite)) {
+            $siteSpecificConfig = self::getSiteSpecificConfig($idSite);
+            $config = array_merge($config, $siteSpecificConfig);
+        }
+        return $config[$name];
```
sgiehl: Could improve this, to avoid random array access errors when we start using this class more.
```suggestion
        return $config[$name] ?? null;
```

config/global.ini.php
```diff
@@ -385,6 +385,10 @@
 ; This feature will not work with the MYSQLI extension.
 archiving_query_max_execution_time = 7200
 
+
+; Allows you to disable archiving segments for selected plugins. For more details please see https://matomo.org/faq/how-to-disable-archiving-the-segment-reports-for-specific-plugins
```
sgiehl, line 389: Wouldn't it be good to directly mention that it's an comma separated list?

plugins/CoreHome/lang/en.json
```diff
@@ -11,6 +11,7 @@
         "CloseWidgetDirections": "You can close this widget by clicking on the 'X' icon at the top of the widget.",
         "ChooseX": "Choose %1$s",
         "DataForThisReportHasBeenPurged": "The data for this report is more than %s months old and has been purged.",
+        "DataForThisReportHasBeenDisabled": "Segmentation is currently disabled for this report, please check %1$sthis FAQ%2$s for more info",
```
sgiehl:
```suggestion
        "DataForThisReportHasBeenDisabled": "Segmentation is currently disabled for this report. Please check %1$sthis FAQ%2$s for more details.",
```

plugins/CoreHome/templates/_dataTable.twig
```diff
@@ -59,6 +59,8 @@
                 <div class="pk-emptyDataTable">
                 {% if showReportDataWasPurgedMessage is defined and showReportDataWasPurgedMessage %}
                     {{ 'CoreHome_DataForThisReportHasBeenPurged'|translate(deleteReportsOlderThan) }}
+                {% elseif showPluginArchiveDisabled is defined and showPluginArchiveDisabled%}
+                    {{ 'CoreHome_DataForThisReportHasBeenDisabled'|translate('<a target="_blank" href="'~'https://matomo.org/faq/how-to-disable-archiving-the-segment-reports-for-specific-plugins'|e('html_attr') ~'">', '</a>')|raw }}
```
sgiehl: The escaping of the link would actually not be needed, as it won't break the html. This is only required for variables were we don't now what content it might add. But as it doesn't hurt, there is no need to change that.

sgiehl (changes requested):
@peterhashair Left a couple of comments for smaller improvements. Regarding a UI test: I guess it would be good to have one, as it would also be kind of an integration test, as it ensures that tracked data isn't archived and displayed. Besides that guess it should be ready to merge after the adjustments

tests/PHPUnit/Fixtures/DisablePluginArchive.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    public $idSite = 1;
+    public $dateTime = '2009-01-04 00:11:42';
+
+    public $trackInvalidRequests = true;
+
+    public function setUp(): void
+    {
+      $this->setUpWebsitesAndGoals();
+      $this->setUpConfig();
```
sgiehl: It would be good to also track some visits in this fixture, maybe one or two with different referrers. That would ensure that no archiving is done. Otherwise the reports would show the data...

sgiehl (changes requested):
@peterhashair it seems you have updated a couple of screenshots that are unrelated to this issue. The updater screenshots are failing for a specific reason, and they should work again, once that issue is fixed. So the changes to them should be reverted. Also you added `tests/UI/expected-screenshots/DisablePluginArchive.png`, which is not used.

peterhashair:
@sgiehl right, update below. I am not too sure the refer tracking is correct, copied from other fixtures. 
- remove the unrelated UI changes.
- set up a refer tracking.
- update screenshot name

plugins/CoreHome/lang/en.json
```diff
@@ -11,6 +11,7 @@
         "CloseWidgetDirections": "You can close this widget by clicking on the 'X' icon at the top of the widget.",
         "ChooseX": "Choose %1$s",
         "DataForThisReportHasBeenPurged": "The data for this report is more than %s months old and has been purged.",
+        "DataForThisReportHasBeenDisabled": "Segmentation is currently disabled for this report. Please check %1$sthis FAQ%2$s for more details.",
```
sgiehl, line 14: @tsteur Not sure if this will be a feature we are going to use on cloud maybe. In that case I was wondering if linking an FAQ in the message would actually be helpful, as the user can't change that 🤔

sgiehl (approved):
The remaining failing tests seem unrelated, so should be good to merge now. Left a comment regarding cloud, so might be good to wait for an answer.

tsteur:
@sgiehl linking to an FAQ be great, the FAQ could also mention specifically the cloud that if they are on the cloud, to reach out to our support.

sgiehl:
@peterhashair I have merged in the latest changes from `4.x-dev`. If tests are passing, feel free to merge

PR #18284, opened 2021-11-09
[Vue] wait for angular to be initialized to post events to avoid loading race condition

diosmosis:
### Description:

events can sometimes be posted before angular is ready. Waiting for angular seems to fix the issue.

### Review


sgiehl (approved):

sgiehl:
seems the vue files weren't updated for some reason. Will merge in latest 4.x-dev, maybe that triggers the build.

sgiehl:
@diosmosis I guess I know why the vue files didn't built. The action only ran on `synchronize`, but seems that isn't triggered when the PR is opened. As you didn't push any further commits afterwards it was never triggered. I've added a commit so the action should also run when a PR is opened. Feel free to have a look and merge if you think that's fine.

diosmosis:
@sgiehl I see, thanks for pushing a fix!

plugins/CoreHome/vue/src/Matomo/Matomo.ts
```diff
@@ -65,8 +65,10 @@ piwik.postEvent = function postMatomoEvent(
   piwik.postEventNoEmit(eventName, ...args);
 
   // required until angularjs is removed
-  const $rootScope = piwik.helper.getAngularDependency('$rootScope') as any; // eslint-disable-line
-  return $rootScope.$oldEmit(eventName, ...args);
+  angular.element(() => {
```
sgiehl, line 68: @diosmosis seems this broke the js tests. See https://app.travis-ci.com/github/matomo-org/matomo/jobs/547700442#L2281-L2291
does there maybe need to be a `return` before the `angular.element`? Or the return type needs to be adjusted

diosmosis: No, we can't return anything here unfortunately. It doesn't seem like the result is used in angular.js so it should be fine. I've included it in another PR, but will extract the fix into a new one.

sgiehl: Not returning anything is fine as well. Just wasn't sure how this needs to be fixed

PR #18285, opened 2021-11-09
Ensure DNS records are only requested for IP allowlist when needed

### LINKED ISSUE #18244: Ensure DNS records are only requested for IP allowlist when needed
## Expected Behavior

No DNS Request for network entries just for hostnames in login_whitelist_ip/login_allowlist_ip

## Current Behavior

All entries in login_whitelist_ip/login_allowlist_ip for networks that was specified with an "/" are attempted to be resolved using DNS. 
For Example: 
We have "192.168.50.0/24" in our login_whitelist_ip/login_allowlist_ip. Now our DNS Server recivies a ton of request (around 175000 per day) trying to resolve a IP for hostname "192.168.50.0/24" and "192.168.50.0/24.ourdomain.com"

login_whitelist_ip/login_allowlist_ip still is working as expected.
We have this issue since our update to matomo 4.5.0

## Possible Solution

Don't try to resolve a DNS Name for a network.

1. Enter a "/x" Network into login_whitelist_ip or login_allowlist_ip like "192.168.50.0/24"
2. Check all DNS Request send by Matomo

## Your Environment

* Matomo Version: 4.5.0
* PHP Version: 7.4
* Server Operating System: Ubuntu 20.04.3 LTS

sgiehl:
### Description:

- only request dns records if it's not an IP range
- cache DNS results for 30 seconds to reduce DNS requests

Note: Tracking requests never go through the allow-list, as that's excluded here: https://github.com/matomo-org/matomo/blob/bc75f9dafc9287a7864f3ecd556a5ae6b699546a/plugins/CoreHome/CoreHome.php#L78-L81

fixes #18244

### Review

### DIFF
--- a/config/global.php
+++ b/config/global.php
@@ -170,26 +170,32 @@ return array(
             if (filter_var($ip, FILTER_VALIDATE_IP) || \Matomo\Network\IPUtils::getIPRangeBounds($ip) !== null) {
                 $ipsResolved[] = $ip;
             } else {
-                $ipFromHost = @gethostbyname($ip);
-                if (!empty($ipFromHost) && $ipFromHost !== $ip) {
-                    $ipsResolved[] = $ipFromHost;
-                } 
-                
-                if (function_exists('dns_get_record')) {
-                    $lazyCache = \Piwik\Cache::getLazyCache();
-                    $cacheKey = 'DNS.' . md5($ip);
-                    if ( !$lazyCache->contains($cacheKey) ) {
-                        $entry = @dns_get_record($ip, DNS_AAAA);
-                        $lazyCache->save($cacheKey, $entry, 30);
-                    } else {
-                        $entry = $lazyCache->fetch($cacheKey);
+                $lazyCache = \Piwik\Cache::getLazyCache();
+                $cacheKey = 'DNS.' . md5($ip);
+
+                $resolvedIps = $lazyCache->fetch($cacheKey);
+
+                if (!is_array($resolvedIps)) {
+                    $resolvedIps = [];
+
+                    $ipFromHost = @gethostbyname($ip);
+                    if (!empty($ipFromHost) && $ipFromHost !== $ip) {
+                        $resolvedIps[] = $ipFromHost;
                     }
 
-                    if (!empty($entry['0']['ipv6'])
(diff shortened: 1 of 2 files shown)


config/global.php
```diff
@@ -167,19 +167,24 @@
 
         foreach ($ips as $ip) {
             $ip = trim($ip);
-            if (filter_var($ip, FILTER_VALIDATE_IP)) {
+            if (filter_var($ip, FILTER_VALIDATE_IP) || \Matomo\Network\IPUtils::getIPRangeBounds($ip) !== null) {
```
tsteur, line 170: can we add a test for this one @sgiehl ?

config/global.php
```diff
@@ -167,19 +167,24 @@
...
-                    // we don't check using filter_var if it's an IP as "gethostbyname" will return the $ip if it's not a hostname
-                    // and we then assume it is an IP range. Otherwise IP ranges would not be added. Ideally would above check if it is an
-                    // IP range before trying to get host by name.
+                if (!empty($ipFromHost) && $ipFromHost !== $ip) {
                     $ipsResolved[] = $ipFromHost;
                 } 
                 
                 if (function_exists('dns_get_record')) {
-                    $entry = @dns_get_record($ip, DNS_AAAA);
+                    $lazyCache = \Piwik\Cache::getLazyCache();
```
tsteur: Can we also cache `gethostbyname` as I would assume it's equally "slow" and can add load for DNS when it resolves IPv4 there?

config/global.php
```diff
@@ -167,19 +167,24 @@
...
+                    $cacheKey = 'DNS.' . md5($ip);
+                    if ( !$lazyCache->contains($cacheKey) ) {
```
tsteur: Using the `contains` can maybe cause a bit of a race condition. As when we fetch the entry, the cache might be already outdated. Also `contains` already does a `fetch` I believe (not 100% sure) so we might actually fetch the cache twice in the end. I wonder if we can fetch the cache here using `$entry = $lazyCache->fetch($cacheKey);` and then only get the value if none is there yet?

tsteur (changes requested):
Works overall @sgiehl . Few tweaks be needed though.

sgiehl:
@tsteur Thanks for the suggestions. I've updated the code accordingly.

tsteur:
Works 🚀

PR #18290, opened 2021-11-10
GitHub actions

peterhashair:
### Description:
All the files that failed will be downloadable in the artifact at the moment. Add a new command for GitHub action. Not sure if that needed `./console tests:sync-ui-screenshots -a github -r matomo-org/matomo 1528756403` just used for my local screen compare and update image.

### Failed UI
image
 

Group all the bash to `.github/scripts/bash/prepare.sh`
This file split PHP XML into smaller fragments `tests/PHPUnit/formatXML.php`

### MYSQLI adapter setup
`mysqli.allow_local_infile = On`

### Submodule implementation: 
To set up Submodule on GitHub need to update plugin_tests.yml. Once approved, github-action should be 4.x-dev
```yaml
    env:
      TARGET_BRANCH: github-aciton
      PLUGIN: AnonymousPiwikUsageMeasurement

```

- AnonymousPiwikUsageMeasurement https://github.com/matomo-org/plugin-AnonymousPiwikUsageMeasurement/pull/67 √
- Bandwidth https://github.com/matomo-org/plugin-Bandwidth/pull/40 √
- SecurityInfo https://github.com/matomo-org/plugin-SecurityInfo/pull/57 
- MarketingCampaignsReporting https://github.com/matomo-org/plugin-MarketingCampaignsReporting/pull/89
- CustomerAlerts https://github.com/matomo-org/plugin-CustomAlerts/pull/104
- CustomVariables https://github.com/matomo-org/plugin-CustomVariables/pull/29
- DeviceDetectorCache https://github.com/matomo-org/plugin-DeviceDetectorCache/pull/33
- LogViewer https://github.com/matomo-org/plugin-LogViewer/pull/40
- VisitorGenerator https://github.com/matomo-org/plugin-VisitorGenerator/pull/86
- LoginLdap https://github.com/matomo-org/plugin-LoginLdap/pull/256
- TasksTimetable https://github.com/matomo-org/plugin-TasksTimetable/pull/29
- QueuedTracking https://github.com/matomo-org/plugin-QueuedTracking/pull/166
- tag-manager https://github.com/matomo-org/tag-manager/pull/386
- TreemapVisualization https://github.com/matomo-org/plugin-TreemapVisualization/pull/25
- TrackingSpamPrevention https://github.com/matomo-org/plugin-TrackingSpamPrevention/pull/40

### Review


peterhashair:
I think this is kind of working. coupe random fails 
image

peterhashair:
I guess this is ready for review, just 2 submodule image needs an update.
update UIs are all sparkline-related. Somehow the sparkline got 100px diff, although it looks the same.

I am not too sure about those. @sgiehl do you know why those `$` not find is failing
[https://github.com/matomo-org/matomo/runs/4175351607?check_suite_focus=true](https://github.com/matomo-org/matomo/runs/4175351607?check_suite_focus=true)  

`
  1) Overlay
       should load correctly:
          Evaluation failed: ReferenceError: $ is not defined
           at __puppeteer_evaluation_script__:2:17
            Url to reproduce: http://localhost/tests/PHPUnit/proxy/?module=Overlay&period=year&date=today&idSite=3#?l=http$3A$2F$2F127.0.0.1$2Ftests$2Fresources$2Foverlay-test-site-real$2F
            Screenshot of failure: /home/runner/work/matomo/matomo/tests/UI/processed-ui-screenshots/should_load_correctly_failure.png
  `

sgiehl:
Haven't had a closer look yet, but regarding

> I am not too sure about those. @sgiehl do you know why those $ not find is failing
https://github.com/matomo-org/matomo/runs/4175351607?check_suite_focus=true

Those tests are trying to open a page on `http://127.0.0.1/tests/resources/overlay-test-site-real/`, which should be served by the configured webserver. Not sure if your implementation actually serves something using the IP or maybe creating that site fails for some reason here: https://github.com/matomo-org/matomo/blob/2404b24c3a52f04475b63f300269a129399882ec/tests/PHPUnit/Fixtures/UITestFixture.php#L282

tsteur:
I haven't tested it but something that will need to be looked at is to make this work for our many plugins. With travis, a plugin only adds a `.travis.yml` and it works see eg https://github.com/matomo-org/plugin-QueuedTracking/blob/4.x-dev/.travis.yml . This `travis.yml`  doesn't have much logic as otherwise developers might often need to change the plugin travis.yml and they wouldn't know what needs changing.

I don't know what this all means for the code in here. Effectively, a plugin should only have to create a workflow yml with ideally little logic to make the tests work.

I would assume that some of these scripts in the matomo-tests.yml will need to be moved to external scripts 
image

Also I'm assuming that for plugins the directory structure will be different 
image

Not sure if we want to cover this separately in a separate PR or what is best/easiest? 

I'm mostly mentioning in case this means there will be bigger refactorings needed in the way it works.

peterhashair:
@tsteur right, I was wondering why in bash it has so many variables set for plugin names, now I get it. Maybe I should convert all the run command into a bash file take env variables like it was. Then we can reuse it in the plugins, but I guess all the plugins have `.travis.yml` has to convert to GitHub actions as well? Or we can leave Travis for plugins for now. Only Core on Github Action. what would be the best approach?

tsteur:
> . Then we can reuse it in the plugins, but I guess all the plugins have .travis.yml has to convert to GitHub actions as well? 

That will be needed indeed. 

We'll also need to set up the GitHub secrets for each plugin as currently the secrets are configured in Travis.

We'll also need to adjust various docs like https://developer.matomo.org/guides/tests-travis and https://developer.matomo.org/guides/tests-travis-extended

> Or we can leave Travis for plugins for now. Only Core on Github Action. what would be the best approach?

I would say they should all use the same system. That will simplify a lot of things.

peterhashair:
@tsteur that makes sense. I was wondering if we make a docker image for this. Because converting to bash and run it in Github Action like the Travis it becomes a unique format, at some stage we move again, it needs to be rewritten as well. But if we build a docker image, it can be portable to any test engine, even locally as long as docker exist.

tsteur:
@matomo-org/core-reviewers are there any thoughts on using docker? Any benefits or downsides there?

peterhashair:
@tsteur convert it to bash, still fixing some ui failed mismatch, should I start covert some of the plugin travis.

tsteur:
@peterhashair yeah it be great to test a plugin and checking that it works running tests on github actions just by putting in a minor workflow.yml. Not sure if we need to merge this one before to make this work? 

Before doing this, maybe someone should also review this and we'd want tests to pass already here ideally?

peterhashair:
@tsteur yes I will fix all the failed tests, I had a preview most of them are tiny fonts different.

peterhashair:
fix almost every UI Test just a couple of common ones remain. 
image

peterhashair:
@tsteur convert coupe plugins, it seems working, probably needs someone to review.

sgiehl:
@peterhashair As I just updated my tests for device detector: The php install action meanwhile supports PHP 8.1. So in theory you could change that here, so it will already run tests on PHP 8.1

peterhashair:
@sgiehl 8.1 works. do you want  to update all the tests to 8.1?

sgiehl:
@peterhashair On travis we had builds running on PHP7.2 and PHP8, to cover minimum and maximum supported version. There I would have updated the PHP8 builds only. I didn't yet have a detailed look at this PR, so not sure which builds are running on which PHP version...

peterhashair:
@sgiehl do you want to start reviewing this, it's kind of working, and all the submodules are passing the test except one screenshot I haven't fixed. Also got an update for ui-test-view https://github.com/matomo-org/ui-tests-viewer/pull/37/files.

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,235 @@
...
+        env:
+          MYSQL_DATABASE: piwik_tests
+          MYSQL_ALLOW_EMPTY_PASSWORD: yes
+        options: --tmpfs /var/lib/mysql:rw --health-cmd="mysqladmin ping" --health-interval=5s --health-timeout=2s --health-retries=3
+
+    strategy:
+      fail-fast: false
+      matrix:
+        tests: [ 'UITests' ]
+        parts: [ 0,1,2,3,4,5,6,7 ]
```
sgiehl: Having too many parts might actually even make it slower. The database for our fixtures (e.g. UITestFixture) is reused when it was already set up by a previous test, that for sure can't work for different runners.

sgiehl:
@peterhashair Sure I can start reviewing that one. Do you want me to have a look at something specific or is this one fully finished? If so please ensure you had a look yourself, that only required changes are included and that our new test suites are testing the same as on travis (e.g. same suites on same PHP versions). I don't want to waste time looking at unfinished PRs.

peterhashair:
@sgiehl Did self-review and Update the PHP version Should be the same as the Travis now. Still needs to clean up the Travis folder and relates. 
There is a test failure because `7.2.34-28+ubuntu20.04.1+deb.sury.org+1` still needs to be fixed. currently working on it. 
All the screenshot updates are fonts-related. 

But I guess this didn't need reviewed before this https://github.com/matomo-org/ui-tests-viewer/pull/37/files deployed first otherwise, the upload won't work.  


PHP version listed. 
- All the Integration, unit and system and core tests run on 7.2 and 8.0,  
- javascript test run on 7.2
-  UI test run on 7.2.

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,238 @@
...
+        env:
+          MYSQL_DATABASE: piwik_tests
+          MYSQL_ALLOW_EMPTY_PASSWORD: yes
+        options: --tmpfs /var/lib/mysql:rw --health-cmd="mysqladmin ping" --health-interval=5s --health-timeout=2s --health-retries=3
+
+    strategy:
+      fail-fast: false
+      matrix:
+        tests: [ 'UITests' ]
+        php: [ '8.0' ]
```
sgiehl: UI Tests should always run on the minimal required PHP version (7.2. at the moment). That way we can ensure not to use any PHP code features that aren't supported by that version. Running the Integration & System tests on the minimal version isn't enough though, as they cover a lot less than the UI tests

core/SupportedBrowser.php
```diff
@@ -60,7 +61,7 @@ private static function browserNotSupported($shortName, $version)
         return array_key_exists($shortName, self::$notSupportedBrowsers) && $version > 0 && $version <= self::$notSupportedBrowsers[$shortName];
     }
 
-    private static function throwException($client)
+    private static function throwException($client, $userAgent)
```
sgiehl: Can't see any benefit in that change, as `$userAgent` isn't used...

plugins/CoreAdminHome/CustomLogo.php
```diff
@@ -16,6 +16,7 @@
 use Piwik\Piwik;
 use Piwik\Plugin\Manager;
 use Piwik\SettingsPiwik;
+use function DI\get;
```
sgiehl:
```suggestion
```

plugins/CoreAdminHome/tests/UI/TrackingFailures_spec.js
```diff
@@ -13,6 +13,10 @@ describe("TrackingFailures", function () {
     var manageUrl = '?module=CoreAdminHome&action=trackingFailures&idSite=1&period=day&date=today';
     var widgetUrl = '?module=Widgetize&action=iframe&moduleToWidgetize=CoreAdminHome&actionToWidgetize=getTrackingFailures&idSite=1&period=day&date=today&widget=1';
 
+  before(async function(){
+    await page.setUserAgent('Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36');
+  });
```
sgiehl, line 18: Just out of curiosity: Was that needed to fix any problem?

tests/PHPUnit/Integration/HttpTest.php
```diff
@@ -76,6 +76,10 @@ public function testCustomByteRange($method)
             return; // not supported w/ this method
         }
 
+        if (($method == 'socket' || $method == 'curl') && getenv('GITHUB')) {
+            $this->markTestSkipped("GITHUB using php -C don't have socket");
```
sgiehl: Are you still using `php -C` ?

tests/PHPUnit/Integration/HttpTest.php
```diff
@@ -106,6 +110,10 @@ public function testHEADOperation($method)
             return; // not supported w/ this method
         }
 
+        if ($method == 'socket' && getenv('GITHUB')) {
+            $this->markTestSkipped("GITHUB using php -C don't have socket");
```
sgiehl: Are you still using `php -C` ?

.github/scripts/bash/plugins/pluginTests.sh
```diff
@@ -0,0 +1,84 @@
...
+      ./console tests:run-js --matomo-url='http://localhost'
+    elif [ "$TEST_SUITE" = "UI" ]
+    then
+        if [ -n "$PLUGIN_NAME" ]
+        then
+            # HACK: this is a hack to get UI test jobs to run. the --extra-options option was added for 2.15.1, but
+            #       older Matomo's will end up w/ a tests:run-ui command that doesn't support it.
+            # a better fix would be to decouple the Matomo testing framework from Matomo in a way that allowed us to
+            # change code for all versions of Matomo as well as selectively for individual Matomo versions.
+            # git checkout 4.x-dev ../../plugins/TestRunner/Commands/TestsRunUI.php
```
sgiehl: That comment can be removed I guess, as that won't be needed for a github action anymore, especially as the `extra-options` flag is supported long enough now.

.github/scripts/bash/prepare.sh
```diff
@@ -0,0 +1,136 @@
...
+SET='\033[0m'
+
+
+# set up fonts
+if [ "$PIWIK_TEST_TARGET" = "UI" ]
+then
+  echo -e "${GREEN}Setup fonts${SET}"
+  git clone --recursive https://github.com/google/woff2.git ../travis_woff2
+  cd ../travis_woff2
+  make clean all
```
sgiehl: Actually I don't think woff2 is required for UI tests. The woff2 file is included in the repo. So that lib is only needed to check if the files is up to date, which is done in a normal test iirc.

.github/scripts/bash/prepare.sh
```diff
@@ -0,0 +1,136 @@
...
+mkdir -p ./tmp/latest
+mkdir -p ./tmp/logs
+mkdir -p ./tmp/sessions
+mkdir -p ./tmp/templates_c
+mkdir -p ./tmp/templates_c/d2
+mkdir -p ./tmp/templates_c/2f
+mkdir -p ./tmp/nonexistant
+mkdir -p ./tmp/tcpdf
+mkdir -p ./tmp/climulti
+mkdir -p /tmp
```
sgiehl, line 97: If that is always done, why is that already done for UI and JS tests as well a couple of lines above?

.github/scripts/bash/prepare.sh
```diff
@@ -0,0 +1,136 @@
...
+  sudo systemctl start nginx
+  sudo cp ./.github/artifacts/ui_nginx.conf /etc/nginx/conf.d/
+  sudo unlink /etc/nginx/sites-enabled/default
+  sudo nginx -t
+  sudo systemctl reload nginx
+  sudo systemctl restart nginx
+fi
+
+echo -e "${GREEN}set folder Permission${SET}"
+cp .github/artifacts/config.ini.github.ui.php config/config.ini.php
```
sgiehl: The config that was set for UI and JS tests before, will here be overwritten. Maybe one config for all is enough then? 🤔

.github/scripts/bash/prepare.sh
```diff
@@ -0,0 +1,136 @@
...
+sed -i 's/3000/\//g' ./config/config.ini.php
+fi
+
+if [ "$PIWIK_TEST_TARGET" = "UI" ];
+then
+  echo -e "${GREEN}setup screenshot folder${SET}"
+  mkdir -p ./tests/UI/processed-ui-screenshots
+  chmod a+rw ./tests/lib/geoip-files || true
+  chmod a+rw ./plugins/*/tests/System/processed || true
+  chmod a+rw ./plugins/*/tests/Integration/processed || true
```
sgiehl: That seems also be done earlier already

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,238 @@
...
+          ./vendor/phpunit/phpunit/phpunit --configuration ./tests/PHPUnit/phpunit.xml --testsuite UnitTests
+
+      - name: Test runing
+        if: startsWith(matrix.tests,'SystemTests') || startsWith(matrix.tests,'IntegrationTests')
+        run: |
+          ./vendor/phpunit/phpunit/phpunit --configuration ./tests/PHPUnit/phpunit.xml --testsuite ${{ matrix.tests }}
+        env:
+          TRAVIS: true
+          GITHUB: true
+          PHP_VERSION: ${{ matrix.php }}
```
sgiehl: Can't this be simplified to:
```suggestion
      - name: Test running
        run: |
          ./vendor/phpunit/phpunit/phpunit --configuration ./tests/PHPUnit/phpunit.xml --testsuite ${{ matrix.tests }}
        env:
          TRAVIS: true
          GITHUB: true
          PHP_VERSION: ${{ matrix.php }}
```
Or do the additional env flags touch the unit tests in any way?

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,238 @@
...
+
+  Javascript:
+    runs-on: ubuntu-20.04
+    services:
+      redis:
+        image: redis:5.0
+        ports:
+          - 6379:6379
+        options: --entrypoint redis-server
```
sgiehl: Don't think redis is required for JS tests

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,238 @@
...
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+      mariadb:
+        image: mariadb:latest
+        ports:
+          - 3306:3306
+        env:
+          MYSQL_DATABASE: piwik_tests
```
sgiehl: We should directly use `matomo` instead of `piwik` where possible
```suggestion
          MYSQL_DATABASE: matomo_tests
```

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,238 @@
...
+          php-version: ${{ matrix.php }}
+          extensions: bcmath, ldap, curl, zip, pdo_mysql, mysqli, gd, redis
+          ini-values: post_max_size=256M, max_execution_time=10000, memory_limit=-1, always_populate_raw_post_data=-1,zend.exception_ignore_args = Off
+          tools: composer:v2
+          coverage: none
+
+      - name: Setup Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '12'
```
sgiehl: As we are suggestion to use `Node 16` for developing with Vue, we maybe should try using a newer version in our builds as well (if that doesn't break anything or needs extra effort)

tests/PHPUnit/Framework/Fixture.php
```diff
@@ -615,6 +615,8 @@ public static function getRootUrl()
             $piwikUrl .= $piwikUri;
         }
 
+        $piwikUrl = getenv('GITHUB') ? "http://localhost/" : $piwikUrl;
+
```
sgiehl, line 620: Any reason why the action needs a special handling here?

tests/PHPUnit/Integration/DependencyTest.php
```diff
@@ -47,6 +47,10 @@ public function test_getMissingDependencies_EmptyVersion_ShouldBeIgnored()
 
     public function test_getMissingDependencies_multipleConditions()
     {
+        //todo this is need to be fix skip for now
+        if (getenv('PHP_VERSION') == '7.2') {
+            $this->markTestSkipped('version number not exist');
+        }
```
sgiehl: Why shouldn't those tests don't work on 7.2 anymore?

tests/PHPUnit/Integration/SqlTest.php
```diff
@@ -36,6 +36,10 @@ public function tearDown(): void
 
     public function testOptimize()
     {
+        if(getenv('GITHUB'))
+        {
+            $this->markTestSkipped('Mariadb does\'t work for some reason');
```
sgiehl: `optimize table` should work on MariaDB, so might be good to investigate that.

tests/PHPUnit/Integration/TravisEnvironmentTest.php
```diff
@@ -18,6 +18,10 @@ class TravisEnvironmentTest extends IntegrationTestCase
 {
     public function testUsageOfCorrectMysqlAdapter()
     {
+        if (getenv('GITHUB')) {
+            $this->markTestSkipped("Github action does't need travis env test");
+        }
+
```
sgiehl: That shouldn't be needed. If the `MYSQL_ADAPTER` env variable isn't set, the test should simply `return` anyways.

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,238 @@
...
+          - 3306:3306
+        env:
+          MYSQL_DATABASE: piwik_tests
+          MYSQL_ALLOW_EMPTY_PASSWORD: yes
+        options: --tmpfs /var/lib/mysql:rw
+    strategy:
+      fail-fast: false
+      matrix:
+        tests: [ 'UnitTests', 'SystemTestsPlugins' ,'SystemTestsCore' , 'IntegrationTestsCore', 'IntegrationTestsPlugins' ]
+        php: [ '7.2','8.0' ]
```
sgiehl: Before the tests on `8.0` where running the `AllTests` Suite. That made a difference as there might be some tests cases that aren't executed in the other units, due to various reasons (like incorrect folder names,...). So guess we would somehow need to check if all tests are executed correctly at the moment.
In addition the all tests suite was actually running with another database adapter. Would be good to continue running tests on PDO and MYSQLI, as there are differences in the values returned from database.

.github/scripts/bash/plugins/pluginTests.sh
```diff
@@ -0,0 +1,84 @@
...
+else
+    if [ "$COVERAGE" = "Unit" ]
+    then
+        echo "Executing tests in test suite UnitTests..."
+        ./vendor/phpunit/phpunit/phpunit --configuration ./tests/PHPUnit/phpunit.xml --testsuite UnitTests --colors --coverage-clover $TRAVIS_BUILD_DIR/build/logs/clover-unit.xml $PHPUNIT_EXTRA_OPTIONS || true
+    elif [ "$COVERAGE" = "Integration" ]
+    then
+        echo "Executing tests in test suite IntegrationTests..."
+        ./vendor/phpunit/phpunit/phpunit --configuration ./tests/PHPUnit/phpunit.xml --testsuite IntegrationTests --colors --coverage-clover $TRAVIS_BUILD_DIR/build/logs/clover-integration.xml $PHPUNIT_EXTRA_OPTIONS || true
+    fi;
```
sgiehl: I would say we can remove all that code coverage stuff. It wasn't used in a long time and if we ever consider using that again, I think it's better to rethink that completely.

tests/PHPUnit/formatXML.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
sgiehl: Is that test splitter still in use? Can see it is used somewhere, but there is actually only one test suite. Guess for the beginning it fine to have similar test splitting as on travis. We can later always look into splitting them up some more...

sgiehl (changes requested):
Had a look through the changes and left some comments. In addition:

* You have added the file `plugins/CustomJsTracker/tests/resources/file-made-non-writable.js`, which should be unneeded, as it's a file created during tests. Maybe we could add it to `.gitignore` instead.

plugins/CoreAdminHome/tests/UI/TrackingFailures_spec.js
peterhashair: @sgiehl that was one of the screenshot, return browser not supported in the error message.

tests/PHPUnit/formatXML.php
peterhashair: @sgiehl I guess I can remove this file. it was used in the test, but I realize the max concurrency GitHub action allow is 20. And the exceed part has to wait for the available proccess, it actually makes the whole process slower, when we excceed 20 matrix

tests/PHPUnit/Integration/SqlTest.php
peterhashair: @sgiehl It seems like OPTIMIZE TABLE works for InnoDB (before MariaDB 10.1.1). https://mariadb.com/kb/en/optimize-table/. I think this one failed. `CREATE TABLE table3 (c INT) ENGINE=InnoDB`.

peterhashair:
I guess we missing this part in here
./console development:sync-system-test-processed {buildnumber}.

sgiehl:
@peterhashair the system test builds used to upload their artifacts to our build server as well.

peterhashair:
@sgiehl modified this a little bit, the Mariadb version using the in github action is higher than 10.1.1. `DB::isOptimizeInnoDBSupported()` is returning true. added the bottom function for the test.
```php
      if (DB::isOptimizeInnoDBSupported()) {
            $this->assertTrue(Db::optimizeTables(array('table3', 'table4')) !== false);
        } else {
            $this->assertTrue(Db::optimizeTables(array('table3', 'table4')) === false);
        }
```

tests/PHPUnit/Integration/DependencyTest.php
peterhashair: @sgiehl is the PHP version on names format a little different cause problem for this test `7.2.34-28+ubuntu20.04.1+deb.sury.org+1` maybe need a little name format on that test

peterhashair:
@sgiehl sorry to bother got a question. Those two keep failing after switch to the MYSQLI adapter. Any idea why?
```
1) Piwik\Tests\Integration\ArchiveProcessingTest::testTableInsertBatch
2) Piwik\Tests\Integration\CronArchiveTest::test_output

```

sgiehl:
@peterhashair Is #18679 a replacement for this one

peterhashair:
@sgiehl yes, this one got some extra changes on screenshot, which doesn't needed

Issue #9198, opened 2015-11-11
Time spent on page calculation is buggy


tsteur:
1. Avg time spent on a page is calculated by dividing the sum of all `time_spent_ref_action` divided by the number of visits `nb_visits`. Not all visits have `time_spent_ref_action` though. Instead `sum_time_spent` should be divided by something like `nb_hits_with_time_spent`. 

2. In tracker we calculate the `time_spent_ref_action` wrong. It calculates `visit_last_action_time - currentTimestamp` but `visit_last_action_time` is updated on any tracking call, meaning also on any hit. 

To make it a bit more clear let's say there are the following tracking calls
* pageview
* event
* pageview 

The time spent for first pageview is calculated by the time difference between the event and the pageview, not between the two pageviews. This means for many common scenarios where one triggers a pageview and then an event, search, content impression, ... the time spent information is not accurate.

hpvd:
uuh good find.- that's quite important!

SR-mkuhn:
Seems like we are experiencing this too -> http://forum.piwik.org/t/update-to-2-15-changed-average-visit-duration-and-time-on-page/16744

tsteur:
I think the issue I mentioned here has been an issue for a long time and not only since the last update. From which Piwik version did you update? I presume the problem you are describing might be actually a different one

mattab:
Marking this issue as duplicate of #9199 - which was renamed to include in its scope this bug

SR-mkuhn:
@tsteur : from 2.14.3 to 2.15.0

SR-mkuhn:
Main Question is: which counting is correct?

tsteur:
@SR-mkuhn this particular issue has  been buggy for quite a while and not only since last update I think.

tsteur:
@SR-mkuhn maybe create a new issue for your problem and describe it there

sebastianpiskorski:
I've also encountered this issue recently. And I've found that there is problem with Metric calculation: 

Time spent on site is defined as `sum_time_spent` and calculated as SUM():
https://github.com/piwik/plugin-CustomDimensions/blob/master/Archiver.php#L164

[...]

PR #18294, opened 2021-11-11
Fix time on page calculation in visits log

### LINKED ISSUE #18270: Average time spent on each page is incorrect
I am tracking events and pages , matomo is calacating the time spent on each page as the difference between the last event done on the page and the next page view and not the difference between the two page views. How to solve this

sgiehl:
### Description:

This change would actually remove the time on page for all actions other than pageviews. For pageviews the time on page would be summed up till the next pageview.

This is an approach to fix #18270 

Did the changes in a couple of minutes while checking for a possible reason. To finish this some manual testing and updating/adding tests would be needed.

@tsteur If this is something that should be worked on further, feel free to assign the PR to a milestone and I will resume the work then.

### Review


tsteur:
@sgiehl how much more time would it take to finish it? I guess originally the idea was to eventually fix the root cause of https://github.com/matomo-org/matomo/issues/9198 which then would fix this as well. As this might take a while until we work on https://github.com/matomo-org/matomo/issues/9198 could maybe get this done now if it's real quick.

There's a bit of a problem that then eg people who aggregate the page view time manually will find it's different to the reported average time. But we could probably ignore this.

sgiehl:
@tsteur I think the code in the PR should actually work. It would remove the "time on page" metric for all non page view actions in the visits log API. The time on page views should be summed up.
To ensure that really works as expected I would need to build a simple system test case that tracks multiple different actions and if that works also adjust the now failing tests. Guess that would be doable in a couple of hours.

tsteur:
@sgiehl this already changes quite a lot of system tests so maybe no new system test would need to be written ? https://app.travis-ci.com/github/matomo-org/matomo/jobs/547727804

sgiehl:
I'd rather like to write a very simple new test that tracks some different actions and validate if the resulting number are correct and then simply update the expected results of all other tests assuming they are correct.
Could also try to do that with the existing tests, but might be more effort to look through the fixtures and check which results would be expected, as we often don't turn time forward when tracking additional actions.

sgiehl:
This PR should be ready for a first review. The time on page should now only be calculated for page views. For all other action types the `timeSpent` and `timeSpentRef` columns are removed. If a visit doesn't contain a page view, there will actually be no action that contains a `timeSpent`

Some System tests of submodule plugins are still failing, but those can be updated once this one would be ready to merge.

tsteur (approved):

sgiehl:
@tsteur shall we merge this one for 4.7 then?

tsteur:
OK sounds good

sgiehl:
@AltamashShaikh This most likely will break the tests for various plugins. The time spent in visit details may change in various tests. Those can be simply updated, if there are no other changes.

AltamashShaikh:
Okay Thanks,
Will keep this in mind

PR #18308, opened 2021-11-14
Update Feedback Question banner

### LINKED ISSUE #18251: Collect feedback about what features people are using?

peterhashair:
### Description:

Update feedback Question Banner, Fix random UI failed for UIIntegrationTest I think this is related to #18124, update selection.json missing heart icon #18262.

Also, follow up on request here [https://github.com/matomo-org/matomo/issues/18251#issuecomment-968298122](https://github.com/matomo-org/matomo/issues/18251#issuecomment-968298122)
- Hide question banner on mobile
- add event hook for banner
- update event show condition

### Review


plugins/Feedback/Feedback.php
```diff
@@ -121,15 +129,7 @@ public function getShouldPromptForFeedback()
...
-        if ($nextReminderDate === false || $nextReminderDate <= 0) {
-
-            // if user is created more than 6 month ago, set reminder to today and show banner
-            $userCreatedDate = Piwik::getCurrentUserCreationData();
-            if (!empty($userCreatedDate) && Date::factory($userCreatedDate)->addMonth(6)->getTimestamp() < $now) {
-                $nextReminder = Date::now()->getStartOfDay()->subDay(1)->toString('Y-m-d');
-                $feedbackReminder->setUserOption($nextReminder);
-                return true;
-            }
+        if ($nextReminderDate === false) {
```
sgiehl, line 132: This actually makes it worse, as everyone will now get notified 6 months after the update. I will merge this PR now nevertheless, so the other fixes are in `4.x-dev`. Will create a PR to improve those checks myself later

PR #18314, opened 2021-11-15
Fix some PHP8.1 deprecation warnings occurred while running tests

sgiehl:
### Description:

Tried running the tests locally and a couple of deprecation warnings showed up. Not sure if they would occur normally. But should be good to fix them nevertheless.

refs #17686 

### Review

### DIFF
--- a/core/Common.php
+++ b/core/Common.php
@@ -298,7 +298,7 @@ class Common
     {
         try {
             // phpcs:ignore Generic.PHP.ForbiddenFunctions
-            return unserialize($string, ['allowed_classes' => empty($allowedClasses) ? false : $allowedClasses]);
+            return unserialize($string ?? '', ['allowed_classes' => empty($allowedClasses) ? false : $allowedClasses]);
         } catch (\Throwable $e) {
             if ($rethrow) {
                 throw $e;
@@ -743,8 +743,12 @@ class Common
 
     public static function stringEndsWith($haystack, $needle)
     {
-        if ('' === $needle) {
+        if (empty($needle)) {
             return true;
+        }
+
+        if (empty($haystack)) {
+            return false;
         }
 
         $lastCharacters = substr($haystack, -strlen($needle));
--- a/core/Common.php
+++ b/core/Common.php
@@ -743,11 +743,11 @@ class Common
 
     public static function stringEndsWith($haystack, $needle)
     {
-        if (empty($needle)) {
+        if (strlen(strval($needle)) === 0) {
             return true;
         }
 
-        if (empty($haystack)) {
+        if (strlen(strval($haystack)) === 0) {
             return false;
         }
 
--- a/core/DataArray.php
+++ b/core/DataArray.php
@@ -286,10 +286,10 @@ class DataArray
 
         $newRowToAdd[Metrics::INDEX_EVENT_SUM_EVENT_VALUE] = round($newRowToAdd[Metrics::INDEX_EVENT_SUM_EVENT_VALUE], static::EVENT_VALUE_PRECISION);
         $oldRowToUpdate[Metrics::INDEX_EVENT_SUM_EVENT_VALUE] += $newRowToAdd[Metrics::INDEX_EVENT_SUM_EVENT_VALUE];
-        $oldRowToUpdate[Metrics::INDEX_EVENT_MAX_EVENT_VALUE] = round(max($newRowToAdd[Metrics::INDEX_EVENT_MAX_EVENT_VALUE], $oldRowToUpdate[Metrics::INDEX_EVENT_MAX_EVENT_VALUE]), static::EVENT_VALUE_PRECISION);
+        $oldRowToUpdate[Metrics::INDEX_EVENT_MAX_EVENT_VALUE] = round(max(0, $newRowToAdd[Metrics::INDEX_EVENT_MAX_EVENT_VALUE], $oldRowToUpdate[Metrics::INDEX_EVENT_MAX_EVENT_VALUE]), static::EVENT_VALUE_PRECISION);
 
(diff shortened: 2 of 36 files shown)


core/DataArray.php
```diff
@@ -286,10 +286,10 @@ protected function doSumEventsMetrics($newRowToAdd, &$oldRowToUpdate)
 
         $newRowToAdd[Metrics::INDEX_EVENT_SUM_EVENT_VALUE] = round($newRowToAdd[Metrics::INDEX_EVENT_SUM_EVENT_VALUE], static::EVENT_VALUE_PRECISION);
         $oldRowToUpdate[Metrics::INDEX_EVENT_SUM_EVENT_VALUE] += $newRowToAdd[Metrics::INDEX_EVENT_SUM_EVENT_VALUE];
-        $oldRowToUpdate[Metrics::INDEX_EVENT_MAX_EVENT_VALUE] = round(max($newRowToAdd[Metrics::INDEX_EVENT_MAX_EVENT_VALUE], $oldRowToUpdate[Metrics::INDEX_EVENT_MAX_EVENT_VALUE]), static::EVENT_VALUE_PRECISION);
+        $oldRowToUpdate[Metrics::INDEX_EVENT_MAX_EVENT_VALUE] = round(max(0, $newRowToAdd[Metrics::INDEX_EVENT_MAX_EVENT_VALUE], $oldRowToUpdate[Metrics::INDEX_EVENT_MAX_EVENT_VALUE]), static::EVENT_VALUE_PRECISION);
 
         // Update minimum only if it is set
-        if ($newRowToAdd[Metrics::INDEX_EVENT_MIN_EVENT_VALUE] !== false) {
+        if ($newRowToAdd[Metrics::INDEX_EVENT_MIN_EVENT_VALUE] !== false && $newRowToAdd[Metrics::INDEX_EVENT_MIN_EVENT_VALUE] !== null) {
```
sgiehl, line 292: Seems this one actually fixes some kind of bug, that was never recognized.
Before when an event value was `null`, which imho means it is not set, it was indirectly casted to int (caused by the round) and thus always resulted in `0`. So the minimum event value was incorrectly reported as `0`.
This can be seen in the expected test results I have updated. In the Fixture we are actually not setting `0` as event value, so imho this should be a valid change

sgiehl:
I'll put this into review. There might still be some PHP 8.1 issues around. But might be good to already merge this before releasing an RC. Will continue searching for additional PHP 8.1 issues the coming days and create a new PR if I find any.

core/Common.php
```diff
@@ -743,10 +743,14 @@ public static function getLastJsonError()
 
     public static function stringEndsWith($haystack, $needle)
     {
-        if ('' === $needle) {
+        if (empty($needle)) {
```
tsteur: This wouldn't work if needle is `0`?

core/Common.php
```diff
@@ -743,10 +743,14 @@ public static function getLastJsonError()
...
             return true;
         }
 
+        if (empty($haystack)) {
```
tsteur: same here it wouldn't work if haystack is `0`?

core/DataArray.php
tsteur: Nice!

core/Metrics/Formatter.php
```diff
@@ -71,7 +71,7 @@ public function getPrettyTimeFromSeconds($numberOfSeconds, $displayTimeAsSentenc
             } else {    
                 $time    = sprintf(Piwik::translate('Intl_NDays'), $days) . " " . sprintf("%02s", $hours) . ':' . sprintf("%02s", $minutes) . ':' . sprintf("%02s", $seconds);
             }
-            $centiSeconds = ($numberOfSeconds * 100) % 100;
+            $centiSeconds = intval($numberOfSeconds * 100) % 100;
```
tsteur, line 74: note to myself, this avoids `Deprecated: Implicit conversion from float xxxxx to int loses precision`. There was no rounding happening before either https://3v4l.org/OK980 so all good

core/Url.php
```diff
@@ -668,13 +668,17 @@ protected static function getHostsFromConfig($domain, $key)
      */
     public static function getHostFromUrl($url)
     {
-        $parsedUrl = parse_url($url);
+        if (empty($url)) {
```
tsteur: should we just check for `is_string` here maybe? Not sure if that makes sense? I suppose host or Url can't be 0?

core/Common.php
```diff
@@ -298,7 +298,7 @@ public static function safe_unserialize($string, $allowedClasses = [], $rethrow
     {
         try {
             // phpcs:ignore Generic.PHP.ForbiddenFunctions
-            return unserialize($string, ['allowed_classes' => empty($allowedClasses) ? false : $allowedClasses]);
+            return unserialize($string ?? '', ['allowed_classes' => empty($allowedClasses) ? false : $allowedClasses]);
```
tsteur, line 301: note to myself: it'll return always false which it also did before https://3v4l.org/qC9u3

tsteur (changes requested):
Looks good overall. Left a few comments while reviewing mostly for myself. I think it's only `stringEndsWith` that might need looking at

core/Common.php
sgiehl: ah yes. will change that.

core/Url.php
sgiehl: makes sense

sgiehl:
@tsteur thanks for the review. I've applied some changes.

tsteur (approved):
@sgiehl 👍  checked with https://3v4l.org/dAKrC and works

Be good to merge once tests pass (unless they fail for another reason)

PR #18315, opened 2021-11-15
Determine the first date to show feedback question banner more randomly

sgiehl:
### Description:

This should randomize the first time the feedback banner is shown

@tsteur is this random enough?

follow-up to #18308

### Review

### DIFF
--- a/plugins/Feedback/Feedback.php
+++ b/plugins/Feedback/Feedback.php
@@ -123,15 +123,19 @@ class Feedback extends \Piwik\Plugin
         $nextReminderDate = $feedbackReminder->getUserOption();
         $now = Date::now()->getTimestamp();
 
-        //user answered question
-        if ($nextReminderDate === self::NEVER_REMIND_ME_AGAIN) {
-            return false;
-        }
+        // If there isn't any reminder date set, or never remind me was selected previously (-1) we determine a new date
+        if ($nextReminderDate === false || $nextReminderDate <= 0) {
 
-        // if is new user or old user field not exist
-        if ($nextReminderDate === false) {
-            //new user extend to 6 month, don't show banner
-            $nextReminder = Date::now()->getStartOfDay()->addMonth(6)->toString('Y-m-d');
+            // if user was created within the last 6 months, we set the date to 6 months after his creation date
+            $userCreatedDate = Piwik::getCurrentUserCreationData();
+            if (!empty($userCreatedDate) && Date::factory($userCreatedDate)->addMonth(6)->getTimestamp() > $now) {
+                $nextReminder = Date::now()->getStartOfDay()->subDay(1)->toString('Y-m-d');
+                $feedbackReminder->setUserOption($nextReminder);
+                return false;
+            }
+
+            // Otherwise we set the date to somewhen within the next 6 months
+            $nextReminder = Date::now()->getStartOfDay()->addDay(Common::getRandomInt(1, 6*30))->toString('Y-m-d');
             $feedbackReminder->setUserOption($nextReminder);
             return false;
         }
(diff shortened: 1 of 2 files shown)


plugins/Feedback/Feedback.php
```diff
@@ -123,15 +123,19 @@ public function showQuestionBanner()
...
-
-        // if is new user or old user field not exist
-        if ($nextReminderDate === false) {
-            //new user extend to 6 month, don't show banner
-            $nextReminder = Date::now()->getStartOfDay()->addMonth(6)->toString('Y-m-d');
+        // If there isn't any reminder date set, or never remind me was selected previously (-1) we determine a new date
+        if ($nextReminderDate === false || $nextReminderDate <= 0) {
+
+            // if user was created within the last 6 months, we set the date to 6 months after his creation date
+            $userCreatedDate = Piwik::getCurrentUserCreationData();
```
tsteur: There's a typo in the method name, it should be `date` probably ? Be good to change it in this release before it ends up being used somewhere else

tsteur (changes requested):
@sgiehl it looks random. However, if I see this correct there is one problem:

> Date::factory($userCreatedDate)->addMonth(6)->getTimestamp() > $now

This would apply to a lot of users after updating? Would potentially most users get to see this after updating if their users was created more than 6 months ago? I think I mention something like we may want to check something like if the user was created between 6 and 6.5 months ago. 

For this to work we might actually need to do a modulo as a user might have been created 1 year ago. Would we need to find out number of days since signing up, and then do a modulo (like $numDaysSinceRegister modulo 6*30)?

plugins/Feedback/Feedback.php
sgiehl: fixed that

sgiehl:
@tsteur updated the code. I had actually used the wrong line of code. It should now do what I had described in the PR.
Users created within the last 6 months will see the banner 6 months after the creation. Users that are older will get a random date within the next 6 months.

tsteur:
@sgiehl I'm not sure if it would still cause every user that was created like a year ago to see this at the same time?

Like assuming creation date was 1.5 years ago. That means `!empty($userCreatedDate) && Date::factory($userCreatedDate)->addMonth(6)->getTimestamp() > $now === true` meaning it goes into the if block.

It then adds 6 months to the creation date in `$nextReminder = Date::factory($userCreatedDate)->addMonth(6)->toString('Y-m-d');`. Meaning nextReminder is in this case 1 year ago. Meaning next page load we would show it?

The same would happen for users signed up 1 year ago and 2 years ago etc. Basically, as soon as they upgrade to the next version, they would all get to see the feedback screen?

Or maybe I'm missing something?

sgiehl:
@tsteur it actually should only go into the if block if the creation date is within the past 6 months. If it is older `(creationdate + 6 months)` should be lower than `$now`. Or did I get something wrong in the if statement :thinking:

justinvelluppillai:
Looks right to me, if the creation date plus six months is less than now it will skip the if block

tsteur:
I see, makes sense. Sorry about that. So in that case we then just pick something randomly. I thought initially to make it then continuous so we show it after 1 year, 1.5 years, 2 years but we make it instead randomly. That's why I read it wrong. That should be all fine though and works too 👍 

Ideally there were tests for it @sgiehl as working with date etc there's always quickly a bug introduced. But I'm happy to skip it to get it into the RC.

tsteur:
Meaning feel free to merge if tests pass.

sgiehl:
@tsteur it should only be the first date that is picked randomly in that case. Once it was shown and clicked away it will be shown again after 6 months.

tsteur:
Yep 👍  all good. 

I meant the first one we show after 1 year, or 1.5 year etc. But that's not important actually just something I had in my mind somehow. Picking the first date randomly is probably even better 👍  so all good

PR #18321, opened 2021-11-16
Sort segment static lists by usage

### LINKED ISSUE #15620: Specific countries appear in visit log but return no results when segmented
See eg https://demo.matomo.org/index.php?module=API&method=API.getSuggestedValuesForSegment&segmentName=countryName&idSite=62&format=JSON&token_auth=anonymous

We return the first 100 countries alphabetically instead of ideally checking which countries are actually visited sorted by the number of visits desc rather than alpabetically

bx80:
### Description:

Fixes #15620 

API calls to suggested values for segments that are static lists (eg. countries) will now be returned sorted by the most used values first. This should mitigate the issue where suggested value lists do not contain the most frequently used values. 

For example, the getSuggestedValuesForSegment list for countryName will now return a list of country names with the countries that have the most visits at the top rather than an alphabetical list potentially containing countries that have no visits or missing countries with many visits but names towards the end of the alphabet.

This has been applied to segments: countryName, browserName and operatingSystemName

### Review


plugins/API/API.php
```diff
@@ -614,7 +614,9 @@ public function getSuggestedValuesForSegment($segmentName, $idSite)
             $values = $this->getSuggestedValuesForSegmentName($idSite, $segment, $maxSuggestionsToReturn);
         }
 
-        $values = $this->getMostFrequentValues($values);
+        if ($this->doesSegmentNeedMostFrequentValues($segmentName)) {
+            $values = $this->getMostFrequentValues($values);
+        }
```
sgiehl: What is this new condition needed for? I guess actually that check now prevents all other segments from showing suggestions at all.

bx80: `getMostFrequentValues()` is grouping and sorting the values list, but as the static lists have already been sorted by most frequently used and do not contain any duplicates we don't want them re-sorted alphabetically. So this check calls a new method that returns true for all segments that are not in a hard coded exclude list, which currently only contains countryName, browserName and operatingSystemName. This is similar to the way `doesSegmentNeedActionsData()` works.

I think it would be nicer if there was some way to have an `isStaticList` boolean property for the segment rather than a hard coded list of segment names, but I couldn't see a clean way to do this without having a segment class.

core/Plugin/Dimension/VisitDimension.php
```diff
@@ -346,4 +347,45 @@ public static function getDimensions(Plugin $plugin)
...
+        $flat = [];
+        $i = 0;
+        foreach ($array as $k => $v) {
+            $flat[$k] = $v['name'];
+            $i++;
+            if ($i == ($maxValuesToReturn - 1)) {
+                break;
+            }
+        }
+        $flat[] = 'Unknown';
```
sgiehl: Does it make sense to have `Unknown` always in the list? Even if it might not have visits?

sgiehl (changes requested):

core/Plugin/Dimension/VisitDimension.php
bx80: The previous code appended `Unknown` to the bottom of the list, but you are right, that no longer makes sense. I've changed this to add `Unknown` the list from the beginning and any unmatched row values should now be counted as `Unknown`. So `Unknown` should only be returned in the suggested values list if it has enough visits.

sgiehl:
Looks good and works as expected. 
As you mentioned it might be better not to have a static list of segments that don't need the most frequent values. A possibility might be to change the `Segment` class. So it has a new property that defines if it requires most frequent values or not. And the method `setSuggestedValuesCallback` could have a second (optional) parameter that sets the property.

bx80:
I've added a property to the `Segment` class to indicate if it requires the frequent values method to be run. The static list of segments has been removed.

sgiehl (approved):
Looks good now 👍

PR #18325, opened 2021-11-16
Remove update script that might cause problems

sgiehl:
### Description:

Was wondering a few times lately why the file `/plugins/VisitFrequency/Archiver.php` was removed locally after running tests.
Seems the reason is, that some updater tests are trying to update from a very old version, causing all update scripts to be executed.

I guess the same might actually happen if someone tries updating from Piwik before 2.2 to Matomo after 4.1.2. The update would first create the Archiver file and the update script would then remove it again when it is executed.
Should be better to remove that one.

ping @diosmosis You had removed that file in one of your PRs by accident as well. guess this is the reason.

### Review


tsteur (approved):
LGTM . could put this into `next_release` branch but not too important

tsteur:
ACtually better we merge this into 4.x-dev as only few users would upgrade some such an old version anyway, and the next update would bring the file again.

PR #18328, opened 2021-11-16
Updates Symfony components to 5.4.x

sgiehl:
### Description:

We are currently still using Symfony Console (and other components) in version 2.6.
This one is quite outdated and has some issues with PHP 8.1 (deprecation warnings).

This PR will update all components to the last available version 5.4.x.

So far there were two issues I have identified after the update:

- The monolog class `ConsoleFormatter` does not longer have multiple constructor parameters, but takes an option array instead. This was adjusted in `plugins/Monolog/config/cli.php`

- The `execute` method of a console command is now kind of forced to return an integer. Before that was kind of expected, but all other values were silently treated as `0`. I have adjusted our command class, so it will catch and log that error and return `0` in that case. That way commands that may not yet be adjusted won't fail. I have nevertheless changed all commands in core here and already created PRs for the submodule plugins.

- The DialogHelper has been deprecated and removed meanwhile. I've added it back for BC, but we need to migrate our usage to the QuestionsHelper instead.

### Review

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,9 +54,9 @@
         "php-di/php-di": "^6.0.0",
         "phpmailer/phpmailer": "^6.1",
         "psr/log": "~1.0",
-        "symfony/console": "~5.3.0",
-        "symfony/event-dispatcher": "~5.3.0",
-        "symfony/monolog-bridge": "~5.3.0",
+        "symfony/console": "~5.4.0",
+        "symfony/event-dispatcher": "~5.4.0",
+        "symfony/monolog-bridge": "~5.4.0",
         "symfony/polyfill-iconv": "^1.20",
         "symfony/polyfill-mbstring": "^1.20",
         "szymach/c-pchart": "~3.0.13",
@@ -70,8 +70,8 @@
         "slevomat/coding-standard": "~7.2.0",
         "squizlabs/php_codesniffer": "~3.6",
         "phpstan/phpdoc-parser": "~1.5.0",
-        "symfony/var-dumper": "~5.3.0",
-        "symfony/yaml": "~5.3.0"
+        "symfony/var-dumper": "~5.4.0",
+        "symfony/yaml": "~5.4.0"
     },
     "replace": {
         "symfony/polyfill-php54": "*",
--- a/composer.json
+++ b/composer.json
@@ -54,9 +54,9 @@
         "php-di/php-di": "^6.0.0",
         "phpmailer/phpmailer": "^6.1",
         "psr/log": "~1.0",
-        "symfony/console": "~2.6.0",
-        "symfony/event-dispatcher": "~2.6.0",
-        "symfony/monolog-bridge": "~2.6.0",
+        "symfony/console": "~5.3.0",
+        "symfony/event-dispatcher": "~5.3.0",
+        "symfony/monolog-bridge": "~5.3.0",
         "symfony/polyfill-iconv": "^1.20",
         "symfony/polyfill-mbstring": "^1.20",
         "szymach/c-pchart": "~3.0.13",
@@ -70,8 +70,8 @@
         "slevomat/coding-standard": "~7.2.0",
         "squizlabs/php_codesniffer": "~3.6",
         "phpstan/phpdoc-parser": "~1.5.0",
-        "symfony/var-dumper": "~2.6.0",
-        "symfony/yaml": "~2.6.0"
+        "symfony/var-dumper": "~5.3.0",
+        "symfony/yaml": "~5.3.0"
     },
     "replace": {
         "symfony/polyfill-php54": "*",
--- a/core/Console/DialogHelper.php
+++ b/core/Console/DialogHelper.php
@@ -1,0 +1,505 @@
+<?php
+
+// This file is a copy of DialogHelper from symfony console 2.8
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Piwik\Console;
+
+use Symfony\Component\Console\Exception\InvalidArgumentException;
+use Symfony\Component\Console\Exception\RuntimeException;
+use Symfony\Component\Console\Formatter\OutputFormatterStyle;
+use Symfony\Component\Console\Helper\InputAwareHelper;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * The Dialog class provides helpers to interact with the user.
+ *
+ * @author     Fabien Potencier <fabien@symfony.com>
+ *
+ * @deprecated since version 2.5, to be removed in 3.0.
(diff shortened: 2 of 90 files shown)


composer.json
```diff
@@ -50,9 +50,9 @@
         "php-di/php-di": "^6.0.0",
         "phpmailer/phpmailer": "^6.1",
         "psr/log": "~1.0",
-        "symfony/console": "~2.6.0",
-        "symfony/event-dispatcher": "~2.6.0",
-        "symfony/monolog-bridge": "~2.6.0",
+        "symfony/console": "~5.3.0",
```
mickaelandrieu: Can we upgrade to 5.4 ? I don't think this will imply more works but this will for sure secure a little bit more the support of these dependencies 😉

core/Console/DialogHelper.php
```diff
@@ -0,0 +1,505 @@
+<?php
+
+// This file is a copy of DialogHelper from symfony console 2.8
```
mickaelandrieu, line 3: You may want to use the "new"  [QuestionHelper](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Console/Helper/QuestionHelper.php) instead

core/Console/DialogHelper.php
```diff
@@ -0,0 +1,505 @@
...
+ * @deprecated since version 2.5, to be removed in 3.0.
+ *             Use {@link \Symfony\Component\Console\Helper\QuestionHelper} instead.
+ */
+class DialogHelper extends InputAwareHelper
+{
+    private $inputStream;
+    private static $shell;
+    private static $stty;
+
+    public function __construct($triggerDeprecationError = true)
```
mickaelandrieu, line 37: ... but if you plan to use it anyway, I don't see the value to keep the deprecation error : WDYT ?

core/Plugin/ConsoleCommand.php
```diff
@@ -54,4 +55,21 @@ protected function checkAllRequiredOptionsAreNotEmpty(InputInterface $input)
...
+    {
+        // This method used to return `0` if a non numeric value was returned from execute()
+        // To have our commands still compatible till the next major release we imitate that behaviour but output
+        // an info message so old commands will get updated
+        // @todo remove in Matomo 5
+        try {
+            return parent::run($input, $output);
+        } catch (\TypeError $e) {
+            if (strpos($e->getMessage(), 'Return value of "') === 0) {
+                Log::info('Deprecation warning: ' . $e->getMessage() . "\nPlease update the command implementation, as this won't be supported by Matomo 5 anymore");
```
mickaelandrieu:
```diff
- Log::info('Deprecation warning: ' . $e->getMessage() . "\nPlease update the command implementation, as this won't be supported by Matomo 5 anymore");
+ Log::info('Deprecation warning: ' . $e->getMessage() . "\nPlease update the command implementation to return an int, as this won't be supported by Matomo 5 anymore");
```

plugins/CoreAdminHome/Commands/DeleteLogsData.php
```diff
@@ -100,6 +100,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
         if ($input->getOption('optimize-tables')) {
             $this->optimizeTables($output);
         }
+
+        return 0;
```
mickaelandrieu: Maybe can we use SymfonyCommand::SUCCESS and SymfonyCommand::FAILURE here ?

Theses [constants are now available in matomo](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Console/Command/Command.php#L33) project

mickaelandrieu (commented):
I'm surprised you have no BC from the Symfony Event Dispatcher as the main function (dispatch()) got a BC break. Lucky you 😇

justinvelluppillai:
Thanks for the comments @mickaelandrieu. This PR and the symfony updates are now scheduled for the 5.0.0 milestone as our initial look suggests it will require BC breaks.

core/Console/DialogHelper.php
sgiehl: I have readded that one for BC only. Some community plugins might still use the DialogHelper, and in order not to break them, we would have this class, at least for a while. But the plan would be to convert our usages. Guess that would be something we could already do before this is merged as iirc the QuestionHelper was also available in 2.6

composer.json
Findus23: 5.4 wasn't released when this PR was created, but with security support until 30 Nov 2025 we might want to switch to 5.4 now.
@sgiehl ?

core/Console/DialogHelper.php
mickaelandrieu: It make sense, but I'm not sure we should convert these commands in that contribution because it will make it too big and maybe difficult to be accepted/reviewed.

Once this one is merged, I'll give a hand to help : I don't know the Matomo code base but this will be a perfect topic for beginner contributions, right ? 😇

sgiehl: I guess we could actually already start to convert the commands. At least replacing DialogHelper with QuestionHelper should be doable without that PR merged, right?

mickaelandrieu: Yes 👍

ulcuber (approved):

sgiehl:
Updated the PR so it will require Symfony 5.4 instead. Once I've fixed all tests, this one should be ready to review.

ulcuber (approved):

justinvelluppillai:
Is this ready for review/merge now @sgiehl ? Maybe if you fix the conflicts we can push it along

sgiehl:
Conflicts solved. Should be ready to review again.

bx80 (approved):
This looks good to to me. Automated tests are 100%, can't see any issues with the code and was able to run a selection of console commands with and without input with no issues, `return self::SUCCESS` :+1: 

* [x] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [x] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [x] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [x] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [x] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [x] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [x] Test updated [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [x] Has breaking changes, change log updated [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [x] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [x] n/a [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [x] n/a Existing documentation updated if needed

PR #18333, opened 2021-11-16
Cache API.getSegmentMetadata better in Segment

### LINKED ISSUE #17944: Cache API.getSegmentMetadata better in Segment - Improve performance on every UI request
I have here a regular UI call, and it triggers more than 90 `API.getSegmentMetadata` calls which then triggers calls to funnels, heatmaps, goals, ... That's a site with like 50 segments.

This is executed here: https://github.com/matomo-org/matomo/blob/4.5.0-b1/core/Segment.php#L186-L194

It's cached in a property but we often create a new segment instance like `new Piwik/Segment()`. So the cache isn't really used and for every segment we create a new instance and fetch segment metadata over and over again. 

The goal of this issue is to have a cache be reused across segment instances. Meaning when a new segment instance is created, and we fetched this data previously, then it should reuse the cache from the other instance. The data may need to be cached per site and maybe other criteria.

image

image

This all can be quite slow.

peterhashair:
### Description:

fixes: #17944 
add cache and lock to segment load, using lazy cache 

### Review


core/Segment.php
```diff
@@ -180,17 +182,45 @@ public function getSegmentExpression()
...
         if (empty($this->availableSegments)) {
-            $this->availableSegments = Request::processRequest('API.getSegmentsMetadata', array(
-                'idSites' => $this->idSites,
-                '_hideImplementationData' => 0,
-                'filter_limit' => -1,
-                'filter_offset' => 0,
-                '_showAllSegments' => 1,
-            ), []);
+            //init lock
+            $lock = new LoaderLock($lockId);
```
tsteur: @peterhashair is there a particular reason the lock is used? It be fine if multiple processes try to populate this cache at the same time.

core/Segment.php
```diff
@@ -180,17 +182,45 @@ public function getSegmentExpression()
         return $this->segmentExpression;
     }
 
+    /**
+     * @throws Exception
+     */
     private function getAvailableSegments()
     {
+        // start lazy cache
+        $cache = PiwikCache::getLazyCache();
```
tsteur: I know there's no needs review label yet but wanted to mention using the lazy cache may be quite complicated as you will need to invalidate this cache whenever eg a custom report/form/goal/dimension... is added or changed or removed. Just some quick feedback. Maybe it's good enough to use a transient cache (haven't looked in detail)

peterhashair:
@tsteur just testing, I saw on the goal page, the same problem we have last time appears, happens here as well.  each graphic called the `API.getSegmentsMetadata`.  only happened when we have lots of goals. Not sure that can be fixed, because the cache won't carry to another request.

tsteur:
Using lazy cache would be ideal for performance but be quite a bit of work to invalidate all the caches in the right moments which makes it very hard and a long job. I guess the idea with transient cache is that it may be an easy fix and we already make it maybe say 50% better compared to making it 90% better but a lot of work. Be good to double check if with a few segments this method is called quite often during one request.

core/Segment.php
```diff
@@ -180,17 +180,30 @@ public function getSegmentExpression()
...
+    /**
+     * @throws Exception
+     */
     private function getAvailableSegments()
     {
+        // start cache
+        $cache = PiwikCache::getTransientCache();
+
+        //covert cache id
+        $cacheId = 'API.getSegmentsMetadata'.SettingsPiwik::getPiwikInstanceId() . implode(",", $this->idSites);
```
sgiehl: Even though it might be a very rare race condition and shouldn't even happen with transient cache, I would suggest to add some additional separators to the cache key.
Assuming two instances `matomo` and `matomo2` and each having multiple sites, the cache key for `matomo` with site id `21` and `matomo2` with site id `1`, would currently be the same.
```suggestion
        $cacheId = 'API.getSegmentsMetadata.'.SettingsPiwik::getPiwikInstanceId() . '.' . implode(",", $this->idSites);
```

sgiehl (changes requested):
Left a comment for a small improvement. Otherwise this should be good to merge.
Checked the changes on a dashboard with a couple of widgets. Without the changes here `API.getSegmentsMetadata` was called over 2000 times when loading the page. With the changes it was only twice 👍

core/Segment.php
```diff
@@ -180,17 +180,30 @@ public function getSegmentExpression()
...
+    /**
+     * @throws Exception
+     */
     private function getAvailableSegments()
     {
+        // start cache
+        $cache = PiwikCache::getTransientCache();
+
+        //covert cache id
+        $cacheId = 'API.getSegmentsMetadata.'.SettingsPiwik::getPiwikInstanceId() . '.' . implode(",", $this->idSites);
```
tsteur, line 192: just fyi, no change needed. For regular caches we don't need the instance ID in there. That was for the `get_lock` mysql method only.

peterhashair: @tsteur ah, this is already merged, should I update this in another PR

tsteur: @peterhashair no changed needed (mention this in the comment above). It's just for future when we work with caches again.

PR #18335, opened 2021-11-17
Handle scheduled task failures

### LINKED ISSUE #17938: Some scheduled tasks might be skipped on error - meaning eg a scheduled report might not be sent
refs L3-141

## Problem 1 - If task fails, it won't be executed / retried again.
Generally I can see reschedules the task here before executing it [matomo/Scheduler.php](https://github.com/matomo-org/matomo/blob/4.4.1/core/Scheduler/Scheduler.php#L120-L121) meaning if eg a scheduled report fails to send the report (or execute any task) it won’t be executed again. Meaning if it's a weekly scheduled report that week no report will be sent. Same applies to any other task that if there is an exception they won't be completed until they run again according to their schedule. They won't be retried the next time the scheduler runs as they were already rescheduled according to their schedule.

## Problem 2 - Schedule reports specific

Maybe at the same time a `cache:clear` happened while we were trying to send the report? I’m thinking the execution stopped here [matomo/API.php](https://github.com/matomo-org/matomo/blob/4.4.1/plugins/ScheduledReports/API.php#L618-L619). 

Error logs are for example

> [2021-08-16 08:39:05] piwik.WARNING: /vendor/tecnickcom/tcpdf/tcpdf.php(7370): Warning - imagepng(/tmp/tcpdf/__tcpdf_12c3eb6b7d6f9bd4a51494b31b58fb98_imgmask_plain_e40746f50d0211c25b80d87ea0473683): failed to open stream: No such file or directory - Matomo 4.4.1 - 

> [2021-08-16 08:39:06] piwik.WARNING: /vendor/tecnickcom/tcpdf/include/tcpdf_static.php(296): Notice - tempnam(): file created in the system's temporary directory - Matomo 4.4.1 - 

> [2021-08-16 08:39:08] piwik.WARNING: /vendor/tecnickcom/tcpdf/include/tcpdf_static.php(1825): Warning - fopen(file:///tmp/assets/PDF Email Report - 3.2021-08-09.2.fr - rJ4AgwmQKD_bZwrLMtR7kvAkfZCFr8Vo9092Ud3c.pdf): failed to open stream: No such file or directory - Matomo 4.4.1 - 

> [2021-08-16 08:39:05] piwik.WARNING: /vendor/tecnickcom/tcpdf/tcpdf.php(7365): Warning - imagepng(/tmp/tcpdf/__tcpdf_12c3eb6b7d6f9bd4a51494b31b58fb98_imgmask_alpha_e40706f40d5211c25b80d87ea0473683): failed to open stream: No such file or directory - Matomo 4.4.1 

## Problem 3 - Exceptions  are never logged or not visible

It seems we are catching such exceptions but never maybe throw them again or log them? See [matomo/Scheduler.php](https://github.com/matomo-org/matomo/blob/4.4.1/core/Scheduler/Scheduler.php#L269)

## Some thoughts

I don't think we need to fix all the problems I suppose if we fix problem 1 then won't need to fix problem 2 and the other way around.

We could probably ignore problem 2 if we fix problem 1 as it means we would retry the task next time the task runner runs. Not sure what potential problems this could cause especially when multiple task runners run at the same time. There could be small edge cases where some reschedule time gets lost but could be likely ignored. It likely would be fine if we were to set the original time again for this task (as it was before rescheduling the task). However, depending on where a task failed this may cause issues: For example let’s say the error happens while sending custom alerts after sending an email. If then an exception happens we would next time retry the same custom alert maybe and send the same email again. It’s like a task would need to define on which exception a task should be retried and when not. For example if there is an error in writing the Geo IP to disk we would run this command on every archiving run and download the DB again and again etc. In any case if there are exceptions during a failed task then we would want to be notified. For example custom alerts currently sends all alerts as part of one task. If one of them plays up then all the other alerts will be resend hourly and cause people to be paged/alerted at night (happened in the past for users). We would maybe need to refactor the task to have one task per alert similar to scheduled reports.

Problem 3 would need to be solved for better debugging I suppose. Meaning logging information so it can be fixed. Or we could let the entire archive run fail so we get an email meaning we stop executing any exceptions after that and stop executing things. This can have downsides though that important tasks after this task wouldn’t be executed. Another one could be to rethrow all thrown exceptions after executing all the tasks. There could be downsides to this too.

We could also decide to fix problem 2 and keep the existing behaviour in scheduled tasks and instead make sure that the tasks itself make sure of retrying things when there is an exception. This might be easiest way to fix things without side effects but requires maybe to pay more attention in each task. Like ScheduledReports could schedule the tasks smarter depending on when they were sent last based on `report.ts_last_sent` column. We would also need to check other tasks if we would need to change them too. 

Maybe we could fix problem 3 first to see how often this happens and then assess priority. If we were to receive emails on cloud when this happens then we'd get probably a good understanding how often this possibly happens. Nonetheless it needs to be fixed eventually either way as it does seem to happen and people rely on the reports.

Any change here might also affect our cloud/account tasks so we would need to make sure things don't cause side effects there too.

bx80:
### Description:

Fixes #17938 (partial)

Catches and logs exceptions that occur when running scheduled tasks

### Review


bx80:
This PR currently just addresses problem 3 of the three problems listed in #17938 by logging any exceptions.

@tsteur Do we want to just add this logging for now to help determine priority or look for a solution for rescheduling/retrying failed tasks? 

As previously noted in the issue, it seems like there would need to be some indication passed from the failed task as to whether to reschedule or not (eg. if an email has been sent then don't reschedule again) and perhaps additionally some sort of retry counter option to avoid getting stuck in a pointless reschedule/fail loop. (reschedule x times, then give up)

So for example:
GeoIP download, no diskspace. Task indicates that it can be rescheduled, task is run three times but disk is still full. Done.
Notification task, error after sending some notification. Task indicates that it should not be rescheduled. Done.
Report fails because of cache clear. Task indicates that it can be rescheduled, second run works. Done.

Perhaps we could start by adding a mechanism to return a reschedule=true/false indication when tasks encounter an exception and then have the scheduler exception handler use this to reschedule if required. Initially we could just have the email reports task check for exceptions that are worth retrying (eg. the 'No such file or directory' error) and then pass back reschedule=true. If this works well then other scheduled tasks could also use the same mechanism as needed.

What do you think?

tsteur:
> Perhaps we could start by adding a mechanism to return a reschedule=true/false indication when tasks encounter an exception and then have the scheduler exception handler use this to reschedule if required.

@bx80 I think that makes sense and maybe it's not too complicated to implement? Then we could decide on a per task level whether we want to reschedule or not if it fails.

bx80:
I've added a fairly simple rescheduling mechanism, for it to work the task must catch an exception and re-throw it with the error code 1. The scheduler execute task exception handler will check if the exception has error code 1 and if so will set a flag to reschedule the task for a retry. The task will then be rescheduled for one hour into the future by the main schedule loop.

To prevent infinite retry loops a new option 'TaskScheduler.retryList' has been added, it is a simple list of task names and the number of times they have been scheduled for a retry, when this count reaches 3 then the failing task will no longer be rescheduled one hour into the future and should fall back to it's normal next scheduled time.

The scheduled reports 'sendReport' method has been adjusted to catch any exceptions thrown on the `generateReport()` method and re-throw with error code = 1. So failures to generate the report should now be retried, but failures in the send part of the process will not be retried.

I'm not sure if a one hour bump and three retries are the best values to use, happy to change them to whatever makes the most sense.

core/Scheduler/Scheduler.php
```diff
@@ -275,8 +306,15 @@ private function executeTask($task)
...
             call_user_func($callable, $task->getMethodParameter());
             $message = $timer->__toString();
-        } catch (Exception $e) {
+        } catch (\Throwable $e) {
+            $this->logger->error("Scheduler: Error {errorMessage} for task '{task}'",
+                ['errorMessage' => $e->getMessage(), 'task' => $task->getName()]);
             $message = 'ERROR: ' . $e->getMessage();
+
+            // If the task has indicated that retrying on exception is safe then flag for rescheduling
+            if ($e->getCode() == 1) {
```
sgiehl: This actually doesn't look like a very clean solution. In theory every exception could have the error code 1. So either we use a custom exception (like `RetryableException`) for that, so we can check for that specific class. Or if we want to define that on a task level, dependently from a certain exception, we could have a new property in the Task class `retryCount`, which defaults to `0`, but can be overwritten in the constructor (and the various factories).

bx80: Agreed, a custom exception is a much better approach, I've committed a change for this. Since it might not be a good idea to retry the task depending on the stage that the task reached when it failed, a custom exception will allow the task code to decide whether a retry should happen or not, which should be more flexible then defining retries at the task class level.

core/Scheduler/Scheduler.php
```diff
@@ -147,6 +152,33 @@ public function run()
...
 
+                    // Task has thrown an exception and should be scheduled for a retry
+                    if ($this->scheduleRetry) {
+
+                        if($this->timetable->getRetryCount($task->getName()) == 3) {
+
+                            // Task has already been retried three times, give up
+                            $this->timetable->clearRetryCount($task->getName());
+
+                            $this->logger->info("Scheduler: '{task}' has already been retried three times, giving up",
```
sgiehl: Shouldn't we at least log a warning in that case, so we still notice tasks that keep failing and only random errors are hidden?

plugins/ScheduledReports/API.php
```diff
@@ -608,9 +609,9 @@ public function sendReport($idReport, $period = false, $date = false, $force = f
                         $report['period_param']
                     );
 
-            } catch (Exception $e) {
+            } catch (\Throwable $e) {
                 $this->enableSaveReportOnDisk = false;
-                throw $e;
+                throw new RetryableException($e->getMessage());
```
sgiehl, line 614: This would actually mean sending the report is only retried when it fails to be generated. If the task fails sending the reason for e.g. some temporary mailing issue it wouldn't be retried. Not sure if that is intended.

sgiehl (changes requested):
Left some comments.

@tsteur would you mind having a quick look if that solution fits. I think it should work fine, but we would need to go through some our tasks afterwards and check if we would need to throw the `RetryableException` somewhere else as well.

plugins/ScheduledReports/API.php
bx80: That's mostly intended. Failing to generate the report should always be safe to retry because we know it hasn't been sent. Mailing issues are a bit trickier as in theory there could a be an issue where the mail will send for some recipients but still return an error, in which case a retry could cause the report to be emailed three times. 

It's probably a balance between having the report email not arrive at all because of a temporary mail issue vs the risk of some partial error where multiple email reports arrive. I can add retry exception on mail send failure if we think it's a good idea?

sgiehl: At least for email report there is a check that reports aren't sent twice
https://github.com/matomo-org/matomo/blob/affd3737c1c608923ec45f484ddb9afd2650aa6e/plugins/ScheduledReports/ScheduledReports.php#L327
But guess such a thing doesn't exist for sms reports, so guess it's fine not to change anything here for now. We can also adjust / refactor that when needed.

core/Scheduler/Scheduler.php
```diff
@@ -147,6 +152,33 @@ public function run()
                     $readFromOption = true;
                     $message = $this->executeTask($task);
 
+                    // Task has thrown an exception and should be scheduled for a retry
```
tsteur, line 156: just to be safe does it make sense to call `$this->scheduleRetry = false;` before `$message = $this->executeTask($task);` just in case it's not reset? Or maybe we reset it directly after the `if ($this->scheduleRetry) {`? Just some thought. It might work now with no problem, but it would make it less likely to break in the future depending on what gets refactored

core/Scheduler/Scheduler.php
```diff
@@ -147,6 +152,33 @@ public function run()
...
+                            // Task has already been retried three times, give up
+                            $this->timetable->clearRetryCount($task->getName());
+
+                            $this->logger->warning("Scheduler: '{task}' has already been retried three times, giving up",
+                                ['task' => $task->getName()]);
+
+                        } else {
+
+                            $readFromOption = true;
+                            $rescheduledDate = $this->timetable->rescheduleTaskAndRunInOneHour($task);
```
tsteur, line 170: very random thought, not sure it's an issue. What if the task would run in one hour again anyway because it's running hourly? If I understand it right we would still only execute it once in the next hour and seems all good 👍  and probably in that case we don't need to do anything special.

core/Scheduler/Timetable.php
```diff
@@ -149,4 +163,71 @@ public function readFromOption()
...
+    }
+
+    /**
+     * Increment the retry counter for a task
+     *
+     * @param string $taskName
+     */
+    public function incrementRetryCount(string $taskName)
+    {
+        if (!isset($this->retryList[$taskName])) {
```
tsteur, line 214: would we need to read the list before retrying in case a different process already increased the counter?

core/Scheduler/Timetable.php
```diff
@@ -149,4 +163,71 @@ public function readFromOption()
...
+     * Return the current number of retries for a task
+     *
+     * @param string $taskName
+     *
+     * @return int
+     */
+    public function getRetryCount(string $taskName) : int
+    {
+        // Ignore excessive retry counts, workaround for SchedulerTest mock
+        if (!isset($this->retryList[$taskName]) || $this->retryList[$taskName] > 10000) {
```
tsteur, line 233: would we need to here also first call again `readRetryList` in case a difference process updated it ?

core/Scheduler/Timetable.php
```diff
@@ -19,12 +19,15 @@
...
     const TIMETABLE_OPTION_STRING = "TaskScheduler.timetable";
+    const RETRY_OPTION_STRING = "TaskScheduler.retryList";
 
     private $timetable;
+    private $retryList;
 
     public function __construct()
     {
         $this->readFromOption();
+        $this->readRetryList();
```
tsteur: maybe that one is actually not needed to call if we only read it when needed in the get/save

tsteur (changes requested):
Haven't tested it but looked through the code and left a few comments.

Also few things here:

* Any chance we can add a few (basic) tests for the timetable methods like `incrementCounter` etc?
* Can we also document this new exception in https://developer.matomo.org/guides/scheduled-tasks ?
* we'll also want to create a follow up issue to implement this in custom alerts https://github.com/matomo-org/plugin-CustomAlerts/issues/new to retry these as well should there be any exception.

core/Scheduler/Scheduler.php
bx80: Yes, if a task was scheduled hourly then the retry schedule would be the same as it's next execution so there shouldn't be any difference.

bx80:
I've added a new integration test for the various timetable retry methods and improved the option loading/reloading code as suggested.

There is a new paragraph submitted for the scheduled task developer guide https://github.com/matomo-org/developer-documentation/pull/605 and a new issue added to implement this in the CustomAlerts plugin.

bx80:
All tests are fixed, this should now be ready for a final review.

sgiehl:
@bx80 you could also try adding some integration tests for the retry mechanism. Like creating some kind of test task that does nothing but throwing a retry exception and test that the system behaves correctly with rescheduling and aborting if retry limit was reached. Otherwise this PR already looks fine.

core/Scheduler/Schedule/SpecificTime.php
```diff
@@ -30,4 +30,10 @@ public function setDay($_day)
     {
         throw new \Exception('not supported');
     }
+
+    public function setScheduledTime($scheduledTime)
+    {
+        $this->scheduledTime = $scheduledTime;
+    }
```
sgiehl, line 37: We should in general try to avoid adding new methods that are "needed" in tests only. In that case you could have avoided that need by using the mock builder manually instead of `createPartialMock`.

PR #18337, opened 2021-11-17
[Vue] detect dependent plugins in vue UMDs and order plugins appropriately

diosmosis:
### Description:

Plugin UMD modules can depend on other ones, but the loading order currently is just how it is encountered when reading filesystem folders. This is an issue if, for example, CorePluginsAdmin is loaded before CoreHome when CorePluginsAdmin depends on CoreHome. This PR fixes this by:

- detecting UMD module dependencies via the vue.config.js file. Webpack allows us to specify what imports should result in accessing the window object directly. Currently we do this for plugins, eg, `import blahblah from 'CoreHome'`. We did this before by using an object mapping `'CoreHome'` with the object in `window` (also 'CoreHome' currently). webpack also allows using a function instead of an object, so now we're using a function that will detect and remember when a plugin is imported.
- then a new webpack plugin is used to output that data to a umd.metadata.json file.
- in JScriptUIFetcher, we read these files and do a quick DFS to order the plugins correctly for loading UMD files.

Note: this PR does not currently detect cycles.

### Review


diosmosis:
Added docs for this here: https://github.com/matomo-org/developer-documentation/pull/587

sgiehl (approved):

PR #18340, opened 2021-11-17
Fixes possible notice

### LINKED ISSUE #18338: WARNING Undefined index: host - Matomo 4.5.0
## Expected Behavior

I get no warning 😆

## Current Behavior

`WARNING [2021-11-16 07:30:20] 29776  ...core/Http.php(229): Notice - Undefined index: host - Matomo 4.5.0`

This warning only occures sometimes when running `console core:archive` directly after this log: `Will invalidate archived reports for yesterday in site ID = 1's timezone (2021-11-16 00:00:00).`

It seems like this only occurs, when the script is run via a crontab. When I run it in the terminal, I never noticed the issue. Maybe this could be a clue to identify the root cause. I'll try to dig deeper…

## Your Environment

* Matomo Version: 4.5.0
* PHP Version: 7.4
* Server Operating System: Debian
* Additionally installed plugins: -

sgiehl:
### Description:

fixes #18338 

### Review

### DIFF
--- a/core/Http.php
+++ b/core/Http.php
@@ -226,7 +226,7 @@ class Http
             $isBlocked = false;
 
             foreach ($disallowedHosts as $host) {
-                if (preg_match(self::convertWildcardToPattern($host), $parsedUrl['host']) === 1) {
+                if (!empty($parsedUrl['host']) && preg_match(self::convertWildcardToPattern($host), $parsedUrl['host']) === 1) {
                     $isBlocked = true;
                     break;
                 }


tsteur:
@sgiehl does this mean the user changed `http.blocklist.hosts` as it wouldn't be expected otherwise such a notice? I wonder if in that case we would rather need to throw an exception regarding a misconfiguration?

sgiehl:
That shouldn't be the problem in that case. The parsed url (where a request should be sent to) doesn't contain a host. This could have various reasons I guess. e.g. a request to `file:///tmp/file`, is actually valid, but won't contain a host. Not sure what url was requested in that case.

tsteur (approved):

Issue #13321, opened 2018-08-20
Invite new users in Matomo, rather than creating them directly


mattab:
The goal is to change the way Matomo creates new users (ie. team members), by introducing an *Invitation* system. Instead of directly creating accounts for people, a Super User (or a user with Admin access) would invite a new person to join the Matomo platform, and they have to then accept the invitation (click a button in the invitation email,  within a period of up to 4 weeks or so.). When accepting the invitation they can set their secure password.

The invitation solution is used in products like Github and Slack. Let us look at how they do it and get inspired?

Notes:

* For flexibility reasons, the API should allow bypassing the invitation process, and directly force-create a new valid user. 
 * For API BC reasons, should we keep the default behavior as "no invite" or do we switch to invite by default?
* Initially we thought of having a Wordpress-like email notification system in https://github.com/matomo-org/matomo/issues/12797 but then realised it's better to always invite people so they can agree to Terms & conditions, for example.
* more feedback?

fdellwing:
Take a look at https://github.com/opf/openproject, is has such a system (a good one in my opinion) and is completely open source.

tsteur:
Here already a few notes:

* In UI "Add new user" button becomes "Invite a new user".
  * Password field removed.
  * Username field stays. Why? Because it won't be easy to configure that user without username in Matomo as it's the ID.
* New API for "invite user". Old "addUser" we still keep for BC and possibly also beyond that
* Activity log should recognise "invite user" and "invite user accepted".
* Need a UI to "manage invitations" 
	* remove a pending invite.
	* resend a pending invite
	* see when it expires
	* see a list of all invited people
	* We show this as part of the "Manage users" screen above the list of regular users below the "invite a new user" button.
* Have eg a new column in user table "invite_token" and "invite_expire". The token be hashed in the DB and not in plain text. Could use eg password_hash for example. By default, the invite is valid for 7 days. This needs to be configurable using the API parameter  and as a configuration option. 
* While user is being invited, a super user or admin can already configure access to sites for that invited user
* The invited user receives an email with a button to accept the invite.
  *  Then the screen shows up to sign up where a user enters password etc see below
  * A user can also  decline the invite in which case the person that created the invite is being notified by email and we remove the user entry to have no personal data in there.  (we might need to store in the user table who invited the user, if that user login no longer exists then we don't notify anyone)
  * If the invite is expired, then we also remove the user from the DB and send an email to the person that created that user
[...]

PR #18351, opened 2021-11-19
invite user replace add user

peterhashair:
### Description:

 
Just a note, I think this PR will be put on pause for now, currently is on the transaction from angular to VUE. Once the Vue implementation is finished, will convert it to VUE and fix the remaining task.

Add Two table column 

- `invite_status`  sent or null
- `invite_at` datetime

Using currently token from `user_auth_token` table. Resend invite will send a link to the user, clicking the link will mark the user active and land on the password setting page. 

Email template: 

image

**Missing**: 
- UI Test
- Integration Test

### Review


peterhashair:
sorry to bother you, 3 questions. And find one bug. 

**Bug:**

- If you come from a `quick link` URL it appends `showadduser=1`, after that you will always stay add user form. even refresh. simple fix, I will in this PR. 

**Questions**

- To addColumn to exiting database table. do I use `4.7.0-b1?
- Introduce user status, currently depending on the `invite_token` field, just checking, do we want to introduce this field as an individual field not relied on `invite_token`. So it accepts pending, suspended, disabled, etc. 

- Resend invites,  after people send invites do we need a status said sent, and the next send can only be done after 15mins? Otherwise, it maybe adds a mail server to span?

tsteur:
@peterhashair I saw your questions in the email but they are removed in the comment here. I'll comment quick anyway

> To addColumn to exiting database table. do I use `4.7.0-b1?

Likely yes, it depends when we merge it and whether by then maybe already a b1 has been released. Likely it will be -b1 though.

> Introduce user status, currently depending on the invite_token field, just checking, do we want to introduce this field as an individual field not relied on invite_token. So it accepts pending, suspended, disabled, etc.

Not 100% sure. It really depends whether it's needed or not. It might not be needed as I think we delete a user if we withdraw the invite etc. We probably won't have a suspended. I'd say it's probably not needed, but it's hard to say without being too much into it.

> Resend invites, after people send invites do we need a status said sent, and the next send can only be done after 15mins? Otherwise, it maybe adds a mail server to span?

Yes that be good 👍

peterhashair:
@tsteur sorry to bother you, I am trying to implement integration tests, but it seems like the database missing the new field I happened to, do I need to add a `install()` function inside of UserManager other than `4.7.0-b1.php`

tsteur:
@peterhashair you will also need to adjust the schema in https://github.com/matomo-org/matomo/blob/4.6.0-rc2/core/Db/Schema/Mysql.php#L47 which is used for new installs

peterhashair:
@tsteur  it seems like the tests still don't apply that database extends fixture on my local does not update the database. It's there a way I can manually run updates, I mean manually run `4.7.0-b1.php`

tsteur:
@peterhashair you will also need to set the version number in the version class to 4.7.0-b1

peterhashair:
Not sure if how we test the email templates on screenshot

tests/resources/OmniFixture-dump.sql
```diff
@@ -1256,6 +1256,8 @@ CREATE TABLE `user` (
   `superuser_access` tinyint(2) unsigned NOT NULL DEFAULT '0',
   `date_registered` timestamp NULL DEFAULT NULL,
   `ts_password_modified` timestamp NULL DEFAULT NULL,
+  `invite_status` varchar(40) DEFAULT NULL,
+  `invite_at` timestamp NULL DEFAULT NULL,
```
sgiehl: That should actually not be needed. When the fixture is set up it tries to perform any available updates that were added after the Omnifixture was last created (based on the version in the option table). So as long as you have added an update script and the version in `core\Version.php` has been increased, the columns should be added automatically

peterhashair:
the question I have for that one is are we missing an email Screenshot test, still thinking about how we implement this.

tsteur:
@peterhashair do you want to test what the email looks like? In that case we could create an integration test and check that the created email content matches the expected HTML or text content. And basically not have a visual test for the email content.

peterhashair:
@tsteur that make more sense, Will do a HTML code comparison

justinvelluppillai:
What's needed here to move this PR along @peterhashair? Can you update and request review?

peterhashair:
@justinvelluppillai nothing needed here, just code review :)

tsteur:
* In the Invite User UI before the "Invite" button can we show some text what will happen after clicking on the button? Like we can mention that the user will receive an email, that the user has X days to confirm this email. And we can also inform them they can reject an unaccepted invite by doing xyz? What else would be important?

image

* The email contains `%s` text 
image


* Can we disable maybe the "Password" form field (maybe also in the API) until a user confirmed the invite? Currently a super user can set the password for a person who was invited but has not accepted anything yet. I checked it wasn't possible to log in but may be still good to prevent this? 

* I'm not sure but it looks like I was automatically logged in if after accepting the invite without having to confirm the password. Possible that this was the case because the super user set a password on my behalf but doesn't look like it. I tested it with another user as well and I did not have to configure any password I was directly logged in and could click around in the UI and it counted as accepted.
Instead it was accepted that we're showing the user a screen where the user confirms the invite, configures password, and also we need to make sure that they'll know what their login is. They won't be automatically logged in. If a privacy or terms or imprint URL is configured in general settings then they would need to agree to the configured terms and policy to confirm the invite. And if an Imprint URL is configured, then we would need to show this URL too but they would not need to agree to it. This could be shown at the bottom.

I stopped for now with reviewing as it looks like there may be still some bigger things left to do

peterhashair:
@tsteur will do an update. :)

plugins/UsersManager/lang/en.json
```diff
@@ -5,10 +5,11 @@
         "TwoFactorAuthentication": "Two-factor authentication",
         "ResetTwoFactorAuthentication": "Reset two-factor authentication",
         "ResetTwoFactorAuthenticationInfo": "If the user can no longer log in due to lost recovery codes or a lost authentication device, you can reset two-factor authentication for the user, so they can log in again.",
-        "AddUser": "Add a new user",
+        "AddUser": "Invite a new user",
```
sgiehl, line 8: Please use new translation keys in case the sense of the message changed to ensure existing translations won't be used anymore.

core/Updates/4.7.0-b1.php
```diff
@@ -0,0 +1,62 @@
...
+use Piwik\Segment\SegmentExpression;
+use Piwik\Updater;
+use Piwik\Updates as PiwikUpdates;
+use Piwik\Updater\Migration;
+use Piwik\Updater\Migration\Factory as MigrationFactory;
+
+/**
+ * Update for version 4.7.0-b1.
+ */
+class Updates_4_7_0_b1 extends PiwikUpdates
```
sgiehl, line 28: This needs to be changed to 4.8.0-b1

plugins/CoreAdminHome/lang/en.json
```diff
@@ -161,6 +161,10 @@
...
         "YouMayOptOut2": "You may choose to prevent this website from aggregating and analyzing the actions you take here.",
         "YouMayOptOut3": "Doing so will protect your privacy, but will also prevent the owner from learning from your actions and creating a better experience for you and other users.",
-        "YouMayOptOutBis": "To make that choice, please click below to receive an opt-out cookie."
+        "YouMayOptOutBis": "To make that choice, please click below to receive an opt-out cookie.",
+        "UserInviteSubject": "%1$s send you an invite to join Matomo Analytics for %2$s.",
+        "AcceptInvite": "Accept Invite",
+        "UserInviteContent": "You can accept this invitation by click the accept button below.",
+        "UserInviteNotes": "This invitation was intended for %1$s access, If you were not expecting this invitation, you can ignore this email. If %2$s is sending you too many emails, you can block them or report abuse. The invitation expire in 7 days. "
```
sgiehl:
```suggestion
        "UserInviteNotes": "This invitation was intended for %1$s access. If you were not expecting this invitation, you can ignore this email. If %2$s is sending you too many emails, you can block them or report abuse. The invitation expire in 7 days."
```

plugins/UsersManager/Controller.php
```diff
@@ -436,6 +443,26 @@ public function setIgnoreCookie()
         Piwik::redirectToModule('UsersManager', 'userSettings', array('token_auth' => false));
     }
 
+    public function activeInviteUser()
```
sgiehl, line 446: `activeInviteUser` sounds like a bad method name to me. It actually doesn't really say what it does. I guess something like `acceptInvitation` would be a lot more meaningful.

plugins/UsersManager/Controller.php
```diff
@@ -436,6 +443,26 @@ public function setIgnoreCookie()
...
+    {
+        $userLogin = Common::getRequestVar('login');
+        $token = Common::getRequestVar('token');
```
sgiehl: When using request variables we should try to always provide the expected type to avoid possible issues when using them later. This can be easily achieved using the correct parameters. Would be good to keep that in mind for future PRs as well.

```suggestion
        $userLogin = Common::getRequestVar('login', null, 'string');
        $token = Common::getRequestVar('token', null, 'string');
```

core/Db/Schema/Mysql.php
```diff
@@ -52,6 +52,8 @@ public function getTablesCreateSql()
                           superuser_access TINYINT(2) unsigned NOT NULL DEFAULT '0',
                           date_registered TIMESTAMP NULL,
                           ts_password_modified TIMESTAMP NULL,
+                          invite_status varchar(40) DEFAULT NULL,
```
sgiehl, line 55: Is there any benefit and storing the invite_status at all? Actually if there is an `invited_at` timestamp that means, that the user has been invited. And after he accepted, we can simply remove the timestamp and it indicates the users is active.

plugins/UsersManager/API.php
```diff
@@ -764,6 +768,64 @@ public function addUser($userLogin, $password, $email, $_isPasswordHashed = fals
...
+        Piwik::checkUserHasSomeAdminAccess();
+        UsersManager::dieIfUsersAdminIsDisabled();
+
+        if (!Piwik::hasUserSuperUserAccess()) {
+            if (empty($initialIdSite)) {
+                throw new \Exception(Piwik::translate("UsersManager_AddUserNoInitialAccessError"));
+            }
+
+            Piwik::checkUserHasAdminAccess($initialIdSite);
+        }
```
sgiehl, line 794: Does that mean that a super user can invite someone without any initial access to a website?
Btw. might also be good to check if a site with the id `$initialIdSite` actually exists.

plugins/UsersManager/API.php
```diff
@@ -764,6 +768,64 @@ public function addUser($userLogin, $password, $email, $_isPasswordHashed = fals
...
+          'emailAddress' => Piwik::getCurrentUserEmail(),
+          'userLogin' => $userLogin,
+        ));
+        $mail->safeSend();
+
+
+        // we reload the access list which doesn't yet take in consideration this new user
+        Access::getInstance()->reloadAccess();
+        Cache::deleteTrackerCache();
+
```
sgiehl, line 814: Why should that be needed? Adding a new user shouldn't change anything on the currents user auth. Not should it cause any changes to the tracker, so no need to clear the cache.

plugins/UsersManager/API.php
```diff
@@ -1444,6 +1512,58 @@ public function newsletterSignup()
...
+    }
+
+    private function sendInvite($userLogin)
+    {
+        //add token
+        $this->model->deleteAllTokensForUser($userLogin);
+        $generatedToken = $this->model->generateRandomTokenAuth();
+        $this->model->addTokenAuth($userLogin, $generatedToken, "Invite Token", Date::now()->getDatetime(),  Date::now()->addDay(7)->getDatetime());
```
sgiehl, line 1542: The following requirements aren't fulfilled here:
> By default, the invite is valid for 7 days. This needs to be configurable using the API parameter and as a configuration option.

sgiehl (changes requested):
Had rough look through the code and left various comments. Didn't go on with the review and UI testing as actually various requirements aren't obviously met (by looking at the code only). Please review the requirements here https://github.com/matomo-org/matomo/issues/13321#issuecomment-964729030 and adjust the implementation accordingly.

peterhashair:
@sgiehl that one is not ready yet :) , but will do a update

peterhashair:
Just a note, I think this PR will be put on pause for now, currently is on the transaction from angular to VUE. Once the Vue implementation is finished, will transfer it to VUE and fix the remaining task.

sgiehl:
closing in favor of #18868

PR #18352, opened 2021-11-19
[Vue] migrate piwik-field and related directives

diosmosis:
### Description:

**This PR is based off of https://github.com/matomo-org/matomo/pull/18292.**

`piwik-field` is a recursive directive so they all had to be converted at once, so this PR is also quite large.

Changes:
- Migrate `piwik-field` directive to the vue Field directive. Note: the `templateFile` property is not longer supported. If plugin developers use it in the wild, we'll have to provide a different way of extending Field.
- Migrate `matomo-multi-pair-field` directive to MultiPairField.
- Migrate `matomo-field-array` directive to FieldArray.
- Add extra parameters to createAngularJsAdapter() scope property transform function.
- Migrate `piwik-form-field` directive to FormField. Every template that was included in the angularjs directive was converted to a new vue component used by FormField.
- Add new mechanism to support extending form fields that allows passing in a Vue component, rather than angularjs template (this method is still supported for BC, until Matomo 5).
- Do not use Vue UMDs if there is a dist folder w/ files but no src folder (can happen during development).
- Fix bug visible in 4.x-dev w/ report export: changing the report limit selector did not change the filter_limit in the export URL. For some reason, angularjs couldn't modify $parent.reportLimit, possibly because there is also a scope.reportLimit.
- Put the defineAsyncComponent() function used to get components from other plugins when we don't know if they're loaded into the new `useExternalPluginComponent()` function.
- Removed the selectedSite internal state in the SiteSelector component, it now follows the v-model contract properly.

image

image

### Review


core/AssetManager/UIAssetFetcher/JScriptUIAssetFetcher.php
```diff
@@ -102,11 +102,15 @@ private function addUmdFilesIfDetected($plugins)
...
             $minifiedUmd = "plugins/$plugin/vue/dist/$plugin.umd.min.js";
-
-            if (Development::isEnabled() && is_file(PIWIK_INCLUDE_PATH . '/' . $devUmd)) {
-                $this->fileLocations[$plugin] = $devUmd;
-            } else if (is_file(PIWIK_INCLUDE_PATH . '/' . $minifiedUmd)) {
-                $this->fileLocations[$plugin] = $minifiedUmd;
+            $umdSrcFolder = "plugins/$plugin/vue/src";
+
+            // in case there are dist files but no src files, which can happen during development
+            if (is_dir($umdSrcFolder)) {
```
diosmosis, line 108: During development it can be possible for there to be a dist/ folder but no src/ folder (happens when checking out different branches where one branch doesn't have a vue folder I believe).

plugins/CoreHome/CoreHome.php
```diff
@@ -164,7 +165,7 @@ public function getJsFiles(&$jsFiles)
         $jsFiles[] = "node_modules/jquery.scrollto/jquery.scrollTo.min.js";
         $jsFiles[] = "node_modules/sprintf-js/dist/sprintf.min.js";
         $jsFiles[] = "node_modules/mousetrap/mousetrap.min.js";
-        $jsFiles[] = 'node_modules/angular/angular.min.js';
+        $jsFiles[] = Development::isEnabled() ? 'node_modules/angular/angular.js' : 'node_modules/angular/angular.min.js';
```
diosmosis: Found myself having to debug angularjs quite often w/ the unminified code, so I thought I'd just add this.

plugins/CoreHome/CoreHome.php
```diff
@@ -256,16 +257,11 @@ public function getJsFiles(&$jsFiles)
...
         $jsFiles[] = "plugins/CoreHome/angularjs/period-selector/period-selector.controller.js";
 
-        $jsFiles[] = "plugins/CoreHome/angularjs/multipairfield/multipairfield.directive.js";
-        $jsFiles[] = "plugins/CoreHome/angularjs/multipairfield/multipairfield.controller.js";
-
-        $jsFiles[] = "plugins/CoreHome/angularjs/field-array/field-array.directive.js";
-        $jsFiles[] = "plugins/CoreHome/angularjs/field-array/field-array.controller.js";
-
+        // TODO: [Vue] used by RollUpReporting, has to be kept for now.
+        $jsFiles[] = "plugins/CoreHome/angularjs/siteselector/siteselector-model.service.js";
```
diosmosis, line 268: I thought about using an adapter, but since the data in the angularjs service is meant to be mutable from outside of the service, and the Vue counterpoint is not, I thought it would be better to simply leave it for now.

plugins/CoreHome/angularjs/report-export/reportexport.directive.js
```diff
@@ -253,7 +256,7 @@
                     if (scope.maxFilterLimit > 0) {
                         scope.$watch('reportLimit', function (newVal, oldVal) {
                             if (parseInt(newVal, 10) > parseInt(scope.maxFilterLimit, 10)) {
-                                scope.reportLimit = oldVal;
+                                scope.limitContainer.reportLimit = oldVal;
```
diosmosis, line 259: Using an object was the only way I could get this to work. This bug (described in the issue description) is also in 4.x-dev.

plugins/CoreHome/angularjs/report-export/reportexport.popover.html
```diff
@@ -38,24 +38,27 @@
                  data-title="{{ 'CoreHome_RowLimit'|translate }}"
                  ng-model="$parent.reportLimitAll"
                  ng-hide="maxFilterLimit > 0"
-                 full-width="false"
+                 full-width="true"
```
diosmosis, line 41: Note: `full-width="false"` never seemed to work.

plugins/CoreHome/vue/src/SiteSelector/SiteSelector.adapter.ts
```diff
@@ -62,22 +72,32 @@ export default createAngularJsAdapter<[ITimeoutService]>({
...
+            vm.modelValue = ngModel.$viewValue;
+          }
+        });
       };
     }
 
     $timeout(() => {
       if (attrs.siteid && attrs.sitename) {
-        vm.modelValue = { id: attrs.siteid, name: attrs.sitename };
+        vm.modelValue = { id: attrs.siteid, name: Matomo.helper.htmlDecode(attrs.sitename) };
```
diosmosis: sitename is assumed to be encoded in places.

plugins/CoreHome/vue/src/createAngularJsAdapter.ts
```diff
@@ -64,10 +72,21 @@ function toAngularJsCamelCase(arg: string): string {
...
+export function processScopeProperty<T>(newValue: T): T {
+  if (typeof newValue === 'object'
+    && newValue !== null
+    && Object.getPrototypeOf(newValue) === Object.prototype
+  ) {
+    return Object.fromEntries(Object.entries(newValue).filter((pair) => !/^\$/.test(pair[0]))) as T;
+  }
+
+  return newValue;
+}
```
diosmosis, line 84: AngularJS sometimes adds its own properties to scope properties (like `$$hashKey`), this removes them when sending them to Vue.

plugins/CoreHome/vue/src/createAngularJsAdapter.ts
```diff
@@ -134,7 +153,7 @@ export default function createAngularJsAdapter<InjectTypes = []>(options: {
                   rootVueTemplate += ` @${eventName}="onEventHandler('${eventName}', $event)"`;
                 }
               } else {
-                rootVueTemplate += ` :${info.vue}="${info.vue}"`;
+                rootVueTemplate += ` :${toKebabCase(info.vue)}="${info.vue}"`;
```
diosmosis, line 159: Seemed to work w/o kebab case, but as far as I know it shouldn't, so changed it.

plugins/CoreHome/vue/src/createAngularJsAdapter.ts
```diff
@@ -253,3 +288,27 @@ export default function createAngularJsAdapter<InjectTypes = []>(options: {
...
+  if (typeof v === 'undefined') {
+    return undefined;
+  }
+
+  if (v === null) {
+    return null;
+  }
+
+  return parseInt(v, 10);
+}
```
diosmosis, line 317: Utility functions to deal w/ `@` angularjs attributes that are meant to be booleans/numbers.

plugins/CorePluginsAdmin/vue/src/Field/Field.adapter.ts
```diff
@@ -0,0 +1,211 @@
...
+          return shallowRef(FieldAngularJsTemplate);
+        }
+
+        const { plugin, component } = value;
+        if (!plugin || !component) {
+          throw new Error("Invalid component property given to piwik-field directive, must be {plugin: '...',component: '...'}");
+        }
+
+        return shallowRef(useExternalPluginComponent(plugin, component));
+      },
```
diosmosis, line 166: Converts the angularjs `component` property which will be a JSON object to a Vue component. From Vue, we'll just pass a component directly.

tests/lib/screenshot-testing/support/page-renderer.js
```diff
@@ -121,14 +120,19 @@ PageRenderer.prototype._reset = function () {
...
     console.log('Using page.waitFor is deprecated, please use one of this instead: waitForSelector, waitForFunction, waitForTimeout');
     if (typeof selectorOrTimeoutOrFunction === 'function') {
-        this.webpage.waitForFunction(selectorOrTimeoutOrFunction)
+        return this.webpage.waitForFunction(selectorOrTimeoutOrFunction)
     } else if (typeof selectorOrTimeoutOrFunction === 'number') {
-        this.webpage.waitForTimeout(selectorOrTimeoutOrFunction)
+        return this.webpage.waitForTimeout(selectorOrTimeoutOrFunction)
     } else if (typeof selectorOrTimeoutOrFunction === 'string') {
-        this.webpage.waitForSelector(selectorOrTimeoutOrFunction)
+        return this.webpage.waitForSelector(selectorOrTimeoutOrFunction)
```
diosmosis, line 127: waitFor was not working before since it didn't return the promises returned by these functions. It would return immediately instead of waiting.

tests/lib/screenshot-testing/support/page-renderer.js
```diff
@@ -121,14 +120,19 @@ PageRenderer.prototype._reset = function () {
...
     }
 }
 
+PageRenderer.prototype.type = async function (...args) {
+  await this.webpage.type(...args);
+  await this.waitForTimeout(100); // puppeteer types faster than vue can update the model state
```
diosmosis: puppeteer types too fast for vue to update the model, so adding a timeout here fixed a bunch of UI tests.

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ This is the Developer Changelog for Matomo platform developers. All changes in o
 
 The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)** lets you see more details about any Matomo release, such as the list of new guides and FAQs, security fixes, and links to all closed issues. 
 
+## 4.7.0
```
sgiehl:
```suggestion
## Matomo 4.7.0
```

plugins/CoreHome/CoreHome.php
sgiehl: In theory that might cause problems if someone enables development mode in a Matomo not installed from git. As the the unminified files is removed in builds.

tests/lib/screenshot-testing/support/page-renderer.js
sgiehl: good catch. Guess I didn't notice as I directly converted all occurrences 🙈

sgiehl:
@diosmosis I'm currently doing a review. Might take a while as the changes are quite huge.
Looking through some of the expected UI file changes I have already seen, that the title for text `FieldArray`s are not displayed for each input anymore. See https://github.com/matomo-org/matomo/commit/d343d20a429bc86b205fb9762c498df84a566896?short_path=ecaef42#diff-ecaef42b8254004961e2206553ed6b910ac38155a6fa329264534e400d068fc0
Feel free to fix that already, while I'm still reviewing the code.

plugins/CorePluginsAdmin/vue/src/FormField/FormField.vue
```diff
@@ -0,0 +1,329 @@
...
+    },
+    processedModelValue() {
+      const field = this.formField;
+
+      // convert boolean values since angular 1.6 uses strict equals when determining if a model
+      // value matches the ng-value of an input.
+      if (field.type === 'boolean') {
+        const valueIsTruthy = this.modelValue && this.modelValue > 0 && this.modelValue !== '0';
+
+        // for checkboxes, the value MUST be either true or faluse
```
sgiehl:
```suggestion
        // for checkboxes, the value MUST be either true or false
```

plugins/CorePluginsAdmin/vue/src/FormField/FormField.vue
```diff
@@ -0,0 +1,329 @@
...
+        } catch (e) {
+          // invalid JSON
+        }
+
+        if (defaultParsed !== null && typeof defaultParsed === 'object') {
+          return '';
+        }
+      }
+
+      // TODO: change all instanceof Array to Array.isArray
```
sgiehl: Can this comment be removed or is anything left over?

plugins/CoreHome/angularjs/report-export/reportexport.popover.html
```diff
@@ -38,24 +38,27 @@
...
-                 ng-show="$parent.reportLimitAll == 'no' && maxFilterLimit <= 0">
+                 ng-model="limitContainer.reportLimit"
+                 value="{{ $parent.reportLimit }}"
+                 full-width="true"
+                 ng-if="$parent.reportLimitAll == 'no' && maxFilterLimit <= 0">
             </div>
 
             <div piwik-field uicontrol="number" name="filter_limit"
                  min="1"
                  max="{{ scope.reportLimit }}"
```
sgiehl: Unrelated to this issue. But I guess this actually should be `maxFilterLimit`.

sgiehl (changes requested):
Left a couple of comments while looking through the code only. Will now check tests & do some local testing.

Btw. We won't be able to merge this one, until #18406 is merged, as your branch already contains some of the commits from `next_release` branch.

sgiehl:
@diosmosis While clicking through the UI / tests I so far spotted this issues:

- [x] Opening the configuration for GoogleAnalyticsImporter (which is configured) results in a JavaScript error:
  image
  The same also happens for CustomTranslations plugin, same error caused by `multitupletext` field
- [x] The expected test file for mobile messaging changed here: https://github.com/matomo-org/matomo/commit/f0d1f279d552105955150e4442366a751937c9af?short_path=0ad0702#diff-0ad0702aa329c2b750bd3f088bc8316ac7d524d1b98f06a2e01d930dfd4d53ed
   I don't think we should update the test file, but adjust the test instead to work the same again. The problem seems to be, that you kind of fixed an issue in select boxes. Before the mobile provider select box had an empty entry in the beginning. After the vue changes that empty entry is gone (which imho is correct). So guess we need to adjust this: https://github.com/matomo-org/matomo/blob/9db3daa0349a2a08a8a50b28537722a611c3868d/tests/UI/specs/UIIntegration_spec.js#L818
- [x] Opening the AB-Tests management, results in: 
   image
   Same happens for FormAnalytics and when editing a goal
- [x] Opening the tracking code generator doesn't show the tracking codes anymore till some property is changed. I think this regressed earlier, as it worked in 4.5.0, but doesn't on 4.x-dev. (That's also the reason for the [failing tests](https://app.travis-ci.com/github/matomo-org/matomo/jobs/549742225#L1679)) If that's unrelated to Vue, feel free to move that to a new issue.
- [x] On the GDRP Tools page, there is another error:
   image

sgiehl:
I'm done for now. Couldn't find anything else while clicking around. Feel free to re-request a review once everything is fixed.

diosmosis:
@sgiehl this should be ready for review again

sgiehl (approved):
@diosmosis I have clicked through the issues I had reported and everything seems to work fine now. One last thing needs to be updated. It seems you have updated the CustomAlerts submodule to a specific branch, this might cause problems when we merge that. If there are changes required in the submodule please create PRs there. Guess we need to reset the submodule to `4.x-dev` here before merging. Otherwise good to merge, if tests are passing again.

diosmosis:
@sgiehl I'll create a PR on CustomAlerts :+1:

sgiehl:
fixed a failing test. Will merge this now. @diosmosis feel free to ping me, once the PR for custom alerts is created.

PR #18354, opened 2021-11-19
Avoid any archiving queries in Goals Archiver if no goals or ecommerce are active

### LINKED ISSUE #18283: Don't archive anything in Goals unless a goal is configured or ecommerce is used
I noticed that 2/3rd of sites use neither goals nor ecommerce feature.

This means the goals archiver wouldn't need to archive anything when there is no goal and no ecommerce configured for a site.

If we didn't archive in that case, this would be especially useful because in the Goals archiver, we process two dependent archives for new visitors and returning visitors in [click](https://github.com/matomo-org/matomo/blob/4.6.0-b3/plugins/Goals/Archiver.php#L121-L122) and in [click](https://github.com/matomo-org/matomo/blob/4.6.0-b3/plugins/Goals/Archiver.php#L505-L506). This can be quite slow because it means for every segment (including the All Visits segments), we need to archive two additional segments. This means we need to create two additional temporary tables for the segment to get this data and it can be quite slow to resolve these segment queries. Not doing this can greatly improve archiving overall.

I've prepared a rough fix in https://github.com/matomo-org/matomo/compare/improvegoalarchive?expand=1 but noticed many tests are now failing (see [click](https://app.travis-ci.com/github/matomo-org/matomo/builds/241406215)) because they don't have the ecommerce feature enabled but were tracking ecommerce data nonetheless.

The goal of this issue would be to check if the so far implemented logic makes sense and make the tests work. If tests aren't easy fixable or if there is any issue then we could otherwise also think about only skipping the `processDependentArchive` in the `Goals\Archiver` when there is no goal and no ecommerce enabled.

Eg in the tests some fixtures would probably need to be updated that they no longer return data. In some other tests we'd want the ecommerce feature enabled for a site. This way we can test some sites where the archiver skips it and some sites where it's enabled.

The only downside to this is, once you create a goal or enable ecommerce, it will try to archive the data for the entire year for "processDependentArchive" when the yearly archive is being requested. That should be fine though I suppose. Any thoughts on this?

sgiehl:
### Description:

fixes #18283

Note: The test changes are kind of expected. We had a couple of fixtures where ecommerce data was tracked, even though it actually wasn't enabled for that site.

For some Fixtures I have activate the ecommerce flag, while for others I kept it off on purpose. That way we can see that there are null records archived, even though some data would actually be tracked and available.

### Review


plugins/Goals/Archiver.php
```diff
@@ -128,59 +127,79 @@ public function aggregateDayReport()
         }
     }
 
+    private function hasAnyGoalOrEcommerce($idSite)
```
tsteur, line 130: above in line 120 it should be maybe `$this->usesEcommerce($this->getSiteId())` instead of `if (Manager::getInstance()->isPluginActivated('Ecommerce')) {` ?

tsteur (changes requested):
@sgiehl I suppose it won't cause any issue if we don't implement similar logic in aggregate archives? I had this done initially in https://github.com/matomo-org/matomo/compare/improvegoalarchive?expand=1 but maybe it keeps code simpler if we don't do it like it did. Just not entirely sure if there could be downsides like if you request certain days and it doesn't find that archive, whether it would maybe try to archive it. But I think it won't as long as there is a done flag there. Just double checking.

The biggest benefit of this issue is to not having to go into `$this->getProcessor()->processDependentArchive` when there is no goals and archive. I think that's currently still the case. Could we ideally not call it when there are no goals and no ecommerce?

This saves us having to create a temporary table for the new_visit and returning visit segment  and not having to fetch these visitors (which can be very slow)

sgiehl:
@tsteur won't `$this->getProcessor()->processDependentArchive` also (only) call the goals archiver, but with a segment? It should skip doing any queries in that case as well, but create empty archives. Thought it would be good to have empty ones there as well, to ensure they won't be created later maybe.

plugins/Goals/Archiver.php
sgiehl: That's actually on purpose. So it will create an empty archive if the plugin is activated, but not used for the current site.

tsteur:
@sgiehl AFAIK it would still try to resolve the segment because every time we process dependent archives it's also trying to resolve the number of visits we had for that segment even before we go into the goals archiver

image

tsteur:
The only way would be to initialise a new archive manually for that segment and writing empty entries and finishing the archive without even going in there

sgiehl:
Haven't yet debugged that through, but won't those "VisitsSummary" numbers be archived by the `VisitFrequency` archiver nevertheless. So it should at least be no additional "work" to do 🤔

tsteur:
It's something general the archiver does for any kind of archive. Even if it was a say customreports specific archive. I'm not sure if we can check for any other archives there as they may be outdated or be invalidated soon after this or something. Also would partially depend on the order in which plugins archive in the same archive etc.

sgiehl:
@tsteur added some code to skip the dependent archives if there are no goals and ecommerce isn't used. I guess that's the simplest solution for now even though that might mean that those archives might be archived later when that changes.

tsteur:
@sgiehl we might need to do the same for the `processDependentArchive` calls in `aggregateMultipleReports`? I'm not entirely sure what Matomo does here but usual behaviour when Matomo tries to aggregate data, and it doesn't have the data for the subperiods, then it would launch the archiving for these subperiods. Aka it might try to launch the day archive for each dependent archive. I haven't double checked but wanted to check if you tested that case maybe?

sgiehl:
@tsteur Haven't tested that. But as there won't be a done flag for that segment, I'm pretty sure the day periods would then be processed. Just pushed an update so it will be skipped as well

tests/PHPUnit/Integration/ArchiveProcessor/LoaderTest.php
```diff
@@ -181,28 +177,12 @@ public function test_pluginOnlyArchivingDoesNotRelaunchChildArchives_whenReusing
             ],
             [
                 'idarchive' => '3',
```
sgiehl: The changes to this files are caused by the fact, that the dependent archives are no longer been processed when no goals are available and ecommerce is disabled. Therefor the `Goals` archives (for segments) disappeared. The `VisitsSummary` archives are still present, as they are archived by the `VisitFrequency` archiver.

tsteur:
@sgiehl I've just done some testing what happens when previously we were archiving days/weeks/months. 

Then we add a goal.

Then it goes into the aggregateMultipleReports for a week or month.

It will then launch a processDependentArchive.

It will notice there's no such archive there yet.

It will initiate a new launch of archiving for each day. The parameters are then set again to root request and `onlyArchiveRequestedPlugin` is set to false meaning it will archive all plugins by the looks for that segment instead of only the goal specific one. That makes it potentially worse. It probably also means that we might sometimes launch the entire day of archiving for a dependent archive if the weeks or month being processed before a day which may be especially bad for browser archiving (in general, not just for this issue).

The call stack is quite long.

Screenshot 2021-11-30 at 1 34 42 PM
Screenshot 2021-11-30 at 1 35 54 PM


Even worse it will then launch processDependentArchive again for that day, and because `self::$ARCHIVE_DEPENDENT` is not set to false, it will eg combine the returning visit with the new visit segment and archive this too.

image

We could probably work around it by writing empty archives somewhat similar to this when no goal is configured:

image

But it makes me think maybe there is a problem with processDependentArchive overall as even if the processDependentArchive launches archiving for missing periods, then it should not suddenly lose these archive process parameters 
image

How can we make sure that we wouldn't suddenly launch archiving for all plugins in that case and to make sure we only archive data we need? or maybe there are other ideas/fixes?

sgiehl:
@tsteur I tried working further on that. But actually I did not yet find a solution that would really work.
What we actually tried to avoid is processing the `VisitsSummary` metrics when the dependent segment metrics for Goals are processed, when it is actually not needed.
Tried inserting an empty archive (so there is a done flag for goals), as well as changing the code, so `VisitsSummary` won't be processed at all if only one plugin is requested. In theory that works. Running the archiving only creates archives and `done` flags for the Goals plugin. The `VisitsSummary` metrics aren't archived, at least not in the Goals plugin. 
_The VisitFrequency plugin will always archive the `VisitsSummary` metrics from the same segments. So that will be done in most cases nevertheless._
But that actually does not really solve the problem. Because as soon as the report `Goals.get` is requested all those `VisitsSummary` archives would be processed, which is caused by:
https://github.com/matomo-org/matomo/blob/d5997bd1dd1218c09d8369b880a732af2c69e685/plugins/Goals/API.php#L472-L500
The `getMetrics` API tries to fetch the numeric record for `nb_visits`, which is a `VisitsSummary` metric...

The only useful solution I still would see is to process the dependent archives for goals in all cases (like it was in the beginning). The day archiver inserts empty archives and skips all queries. That for sure will also cause the core metrics being processed (which `VisitsFrequency` would do nevertheless). 

~Nevertheless I've tried to do some other changes that might prevent that the requested plugin is lost during the archiving process.~ Had to revert that, as it broke almost all tests.

tsteur:
👍  we can keep it basically to as before. The only other thing would be that if no goals and no ecommerce is configured, to manually create these archives 
image

Not sure if it's worth looking at this or not? I'm generally thinking it could make archiving quite a bit faster when we don't need to do the processDependentArchives as segments are always a bit slow to resolve to create this temporary table etc. But it's also not the end of the world if we just keep it now as is in the PR and always process the dependent archives.

sgiehl:
@tsteur I tried that, but it led to strange behavior when the VisitFrequency plugin was disabled.
Running the normal archiving created an empty done flag for Goals only. So far so good, but once the report `Goals.get` was requested an additional `VisitsSummary` done flag for those segment appeared, without having any additional data. Didn't debug that through, but guess that might cause other issues if that segment would be used later for other purpose, as the core metrics might stay empty.

plugins/Goals/Archiver.php
```diff
@@ -450,28 +474,30 @@ public function aggregateMultipleReports()
...
 
         /*
          *  Archive General Goal metrics
          */
         $goalIdsToSum = GoalManager::getGoalIds($this->getProcessor()->getParams()->getSite()->getId());
 
         //Ecommerce
-        $goalIdsToSum[] = GoalManager::IDGOAL_ORDER;
-        $goalIdsToSum[] = GoalManager::IDGOAL_CART; //bug here if idgoal=1
+        $goalIdsToSum = array_merge($goalIdsToSum, $this->getEcommerceIdGoals());
```
tsteur: should we only add the ecommerceGoalIds if ecommerce is enabled? and then `$this->getProcessor()->aggregateNumericMetrics($fieldsToSum);` we could run only if `$fieldsToSum` is not empty? $this->getProcessor()->aggregateNumericMetrics($fieldsToSum);

tsteur (commented):
Makes sense. We wouldn't want to create any such archive record for the VisitsSummary plugin either as we would be writing 0 entries probably when there could be visits etc.

Left only one comment to look at. Not sure we need to do anything there. Then good to merge 👍

plugins/Goals/Archiver.php
sgiehl: I have added a check for ecommerce plugin. But the `$goalIdsToSum` array actually always contains the entry `false`, for processing the overview. Not sure if we should skip that one if no other goal is processed or not.

tsteur: Guess we could always keep the `false` 👍

sgiehl:
@tsteur I've updated the tests. From my side everything should be done. Let me know if we can merge or if anything else needs to be adjusted.

tsteur:
sounds good to merge @sgiehl 👍

PR #18357, opened 2021-11-22
Log archiving query timeouts

### LINKED ISSUE #18025: Max execution timeout exceptions for archiving are not handled
The maximum database query execution timeout for archiving queries defaults to two hours and may also be overridden with a custom timeout value in config.ini.php

When an archiving query exceeds this timeout a  'maximum statement execution time exceeded' exception is returned.

## Expected Behavior
The max execution time exception for archiving queries should be caught and used to trigger an event in the same way that the live query Model::handleMaxExecutionTimeError() method works. This event can then be used to provide notifications that the archiving operation has failed due to query timeout.

We need to make sure the user will be aware if queries run into timeouts. 

The goal of this issue is that when you run the archiving using core:archive command, then the archiving will fail and an exception message is shown and it's clear what happened. Ideally also which query caused the issue (if any possible).

## Current Behavior
The max execution time exception is currently not caught for archiving queries and the archiving operation will silently fail. If an archiving query is slow enough to consistently exceed the timeout then archiving may never complete for some reports and without notification it could be difficult to determine why.

bx80:
### Description:

Fixes #18025 

Added an exception handler to the archiving db adapter which will catch any exceptions thrown because of the max_execution_time hint and log the exception and query details.

### Review

### DIFF
--- a/core/DataAccess/ArchivingDbAdapter.php
+++ b/core/DataAccess/ArchivingDbAdapter.php
@@ -13,6 +13,7 @@ use Piwik\Config;
 use Piwik\Db\AdapterInterface;
 use Piwik\DbHelper;
 use Psr\Log\LoggerInterface;
+use Exception;
 
 class ArchivingDbAdapter
 {
@@ -48,7 +49,7 @@ class ArchivingDbAdapter
         $sql = DbHelper::addMaxExecutionTimeHintToQuery($sql, $this->maxExecutionTime);
         $this->logSql($sql);
 
-        return call_user_func_array([$this->wrapped, __FUNCTION__], func_get_args());
+        return call_user_func_array([$this, "callFunction"], array_merge([__FUNCTION__], func_get_args()));
     }
 
     public function query($sql)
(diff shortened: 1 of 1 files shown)


core/DataAccess/ArchivingDbAdapter.php
```diff
@@ -98,4 +111,25 @@ private function logSql($sql)
...
+        unset($args[0]);
+
+        try {
+            return call_user_func_array([$this->wrapped, $function], $args);
+        } catch (\Exception $e) {
+            if ($this->isErrNo($e, \Piwik\Updater\Migration\Db::ERROR_CODE_MAX_EXECUTION_TIME_EXCEEDED_QUERY_INTERRUPTED) ||
+                $this->isErrNo($e, \Piwik\Updater\Migration\Db::ERROR_CODE_MAX_EXECUTION_TIME_EXCEEDED_SORT_ABORTED)
+            )
+            {
+                $this->logger->info('Archiver query exceeded maximum execution time: {details}',
```
sgiehl: I doubt that many people would actually log a `info` on an instance. As the exception is rethrown, and the archiving will fail with an error in any case, wouldn't it be more helpful to log that as a warning, so it appears with normal log configuration?

bx80: Yes, makes sense :+1:

sgiehl (approved):

PR #18364, opened 2021-11-23
Revert "[Vue] do not decode model value of site name in site selector, so it is always matches the DB value"

diosmosis:
Reverts matomo-org/matomo#18356

From more investigation, the merged PR looks like it will cause regressions:

- Some existing code seems to expect the site name to be decoded, see: https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/CoreAdminHome/angularjs/trackingcode/jstrackingcode.controller.js#L172

- Some API responses auto decode the value (every API request with `format = json`): https://github.com/matomo-org/matomo/blob/next_release/core/DataTable/Renderer/Json.php#L69-L73

So it is just piwik.siteName that isn't decoded I think? I'll look at fixing this another quick way, though there are a lot of inconsistencies that need to be ironed out here.


peterhashair:
Test it works for me. @diosmosis maybe there is a little tiny bug, not really relevant. so when you copy a full text to the search autocomplete, it does trigger the search for the first time. 


image
image

sgiehl:
@peterhashair The reverted PR aimed to fix that the site selector as form field might display the pre selected value with htmlentities. You can see that when creating an alert.

sgiehl:
@diosmosis feel free to merge. If you see more inconsistencies that you are not going to fix with your PRs might be good to create issues for it, so we maybe can improve that another time.

diosmosis:
@peterhashair I think this is what you're referring to? https://github.com/matomo-org/matomo/pull/18367

diosmosis:
@sgiehl will merge, the site name encoding issues are pretty longstanding (as I'm sure you're aware), they'll have to be fixed properly in matomo 5... though it seems much more complicated than I originally thought it was...

PR #18136, opened 2021-10-12
Remove root array correctly for jsonp API requests

justinvelluppillai:
### Description:

Fixes #18133 regression

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #18366, opened 2021-11-23
Configuration option to disable transition periods

### LINKED ISSUE #18332: Add new feature to disable Transitions for specific periods
Transitions is using raw data to generate its reports.

This is known to be slow on some high traffic installations see https://matomo.org/faq/how-to/faq_161/

Similar to unique visitors (see https://matomo.org/faq/how-to/faq_113/ ) we would also want a config setting to disable Transitions for specific periods.

Ideally this was configurable additionally per site as it can be useful when there are many lower traffic sites in there and only one or few high traffic sites like

```ini
[Transitions] ; applies to all sites
setting_name=...
[Transitions_61] ; applies only to site 61
setting_name=...
```

We would want to configure whether it's enabled for `year`, `month`, `week`, or `day`. If a range is selected and it's bigger than the not allowed period, then it should not work on that range. If only `month|week|day` is allowed, then any range selecting more than 31 days wouldn't be possible. Maybe we allow to configure the max period like 

```
max_period_allowed=any|year|month|week|day
```

Just a quick thought and naming could be likely improved. I guess once you allow month, it should also automatically allow week and day.

* [ ] When the feature is disabled for a period, it should ideally not be shown in a row action.
* [ ] When the feature is disabled for a period, then the Transitions report should say something like "This feature is not available for this period, try selecting a period with less ..." and not even try to load any data.

bx80:
### Description:

Fixes #18332 

 Adds an option to specify the max allowed period for transitions, either globally or per site

```
[Transitions]
max_period_allowed = "month"
```

```
[Transitions_1]
max_period_allowed = "week"
```
Site level settings override the general settings.
All period smaller than the max period are allowed, so month will also allow week and day.
Ranges with more days than the max period will not be allowed, eg. an 8 day range isn't allowed if the `max_period_allowed = "week"`, but a 7 day range would be ok.

Row transitions obey the same rules, but I haven't hidden the row transition action icon if the period is not allowed as it might be confusing to have to icon disappear entirely without any explanation. It's also a bit awkward having the datatable row action registration methods make an async API call each time to check period availability - open to suggestions on how to do this efficiently if we really want to go ahead and hide the row action :slightly_smiling_face: 

### Review


plugins/Transitions/API.php
```diff
@@ -656,4 +660,63 @@ protected function processTransitionsToFollowingPages($type, $actions)
...
+     *
+     * @param $period
+     * @param $idSite
+     * @param $date
+     *
+     * @return bool
+     */
+    private function getPeriodAllowed($period, $idSite, $date) : bool
+    {
+        $generalMaxPeriodAllowed = Config::getInstance()->Transitions['max_period_allowed'];
```
sgiehl: This will throw a notice, as the `Transitions` section isn't available by default.

plugins/Transitions/API.php
```diff
@@ -656,4 +660,63 @@ protected function processTransitionsToFollowingPages($type, $actions)
...
+        $sectionName = 'Transitions_' . $idSite;
+        $section = Config::getInstance()->$sectionName;
+        $siteMaxPeriodAllowed = $section['max_period_allowed'];
```
sgiehl: Same here

plugins/Transitions/javascripts/transitions.js
```diff
@@ -1586,8 +1586,8 @@ Piwik_Transitions_Ajax.prototype.callApi = function (method, params, callback) {
 
                 if (typeof Piwik_Transitions_Translations == 'undefined') {
                     self.callApi('Transitions.getTranslations', {}, function (response) {
-                        if (typeof response[0] == 'object') {
-                            Piwik_Transitions_Translations = response[0];
+                        if (typeof response == 'object') {
+                            Piwik_Transitions_Translations = response;
```
sgiehl, line 1629: Why was this change required? Wasn't that working correctly before?

sgiehl (changes requested):
Left some comments. Besides that, according to the original issue it would be good to hide the transitions row action when it's disabled for the current period.
Guess to achieve that, the simples solution would be to pass the config setting as a global javascript variable using the even `Template.jsGlobalVariables`, and then you can check that variable in the row action method that returns it's availability: https://github.com/matomo-org/matomo/blob/c83fe5a1beeb0dd2e2c17448e6cf15c00183f40f/plugins/Transitions/javascripts/transitions.js#L125-L137

plugins/Transitions/javascripts/transitions.js
bx80: This wasn't working and the translations were not being loaded. The API returns an object with the translation strings but the response handler is checking for a sub-object for some reason.

justinvelluppillai: Probably related to a change a while back where we removed the wrapping array from json responses?

See https://github.com/matomo-org/matomo/commit/66bcb98556e5b92b7b6187a5a904e13a408d0ed6#diff-287d49d80349087bbbaa1186c9f825689026f878c9bad60f46a735b65011acb2R64

and https://github.com/matomo-org/matomo/pull/18136

plugins/Transitions/API.php
bx80: I've added some extra null checks.

bx80:
@sgiehl Thanks, `Template.jsGlobalVariables` was the pointer I needed. I've added code to disable the transitions row action if the period isn't allowed. Also added a system test to make sure the config setting results in the correct exception.

plugins/Transitions/Transitions.php
```diff
@@ -55,4 +59,24 @@ public function getClientSideTranslationKeys(&$translationKeys)
...
+
+        $api = API::getInstance();
+        if($api->getPeriodAllowed($period, $idSite, $date)) {
+            $allowed = 'true';
+        } else {
+            $allowed = 'false';
+        }
+
+        $out .= "      
+        piwik.transitionsPeriodAllowed = $allowed;\n
```
sgiehl: This won't work properly. I guess you will actually need to pass over the configured allowed period to javascript and do the check in javascript as well.
When switching the period/date in the date selector the page won't be fully reloaded, and so the global variables won't be refreshed. So it currently always uses the initial state matomo was opened with.

sgiehl (changes requested):

plugins/Transitions/Transitions.php
bx80: Good point :+1:  I've changed it to pass over the max period allowed config setting and replicate the API check in Javascript. The transitions row action is now disabled/enabled when report dates are changed (without reloading)

plugins/CoreHome/vue/src/Periods/Range.ts
```diff
@@ -176,6 +176,10 @@ export default class RangePeriod {
   containsToday(): boolean {
     return todayIsInRange(this.getDateRange());
   }
+
+  getDayCount(): number {
+    return Math.ceil((this.endDate.getTime() - this.startDate.getTime()) / (1000 * 3600 * 24));
```
sgiehl: I think that is actually not accurate. It might always return one day less than expected.
Let's assume a two days range from 2021-11-24 to 2021-11-25. As we use `00:00:00` as time the result of `getDayCount` will be `1`, as there is actually a difference of one day. But the reports will show two days. So guess we always need to add `1`. The same also applies for the php code.

core/Period/Range.php
```diff
@@ -498,6 +498,16 @@ public static function getDateXPeriodsAgo($subXPeriods, $date = false, $period =
...
 
+    /**
+     * Return the number of days contained in this range
+     *
+     * @return int
+     * @throws Exception
+     */
+    public function getDayCount()
+    {
+         return self::getNumDaysDifference($this->getDateStart(), $this->getDateEnd());
```
sgiehl: See the other comment

plugins/Transitions/API.php
```diff
@@ -656,4 +660,80 @@ protected function processTransitionsToFollowingPages($type, $actions)
...
     }
+
+    /**
+     * Retrieve the period allowed config setting for a site or all sites if null
+     *
+     * @param $idSite
+     *
+     * @return string|null
+     */
+    public function getPeriodAllowedConfig($idSite) : ?string
```
sgiehl: In which case can this method actually return null? If no config setting is set it will return `all`, in all other cases set config value that is set, which is always not `null`

plugins/Transitions/API.php
```diff
@@ -656,4 +660,80 @@ protected function processTransitionsToFollowingPages($type, $actions)
...
+            $transitionsSiteConfig = Config::getInstance()->$sectionName;
+            $siteMaxPeriodAllowed = ($transitionsSiteConfig && !empty($transitionsSiteConfig['max_period_allowed']) ? $transitionsSiteConfig['max_period_allowed'] : null);
+        }
+
+        if (!$generalMaxPeriodAllowed && !$siteMaxPeriodAllowed) {
+            return 'all'; // No config setting, so all periods are valid
+        }
+
+        // Site setting overrides general, if it exists
+        return ($siteMaxPeriodAllowed ? $siteMaxPeriodAllowed : $generalMaxPeriodAllowed);
```
sgiehl: could be shortened to

```suggestion
        return $siteMaxPeriodAllowed ?? $generalMaxPeriodAllowed;
```

plugins/Transitions/API.php
```diff
@@ -656,4 +660,80 @@ protected function processTransitionsToFollowingPages($type, $actions)
...
+    /**
+     * Check if a period is allowed by config settings
+     *
+     * @param $period
+     * @param $idSite
+     * @param $date
+     *
+     * @return bool
+     */
+    public function getPeriodAllowed($period, $idSite, $date) : bool
```
sgiehl: We have quite a lot methods using the order `$idSite, $period, $date`, might be good to change it here as well to have some kind of consistency.

sgiehl: Also I would suggest to rename this method to `isPeriodAllowed`, as the method checks if the given period is allowed and doesn't return the allowed period

plugins/Transitions/API.php
sgiehl: Another note: When adding public methods to API classes, please always keep in mind, that all those methods will be listed in the API docs. We should avoid adding to many methods that are actually useless for a user or are for internal use only. Personally I wouldn't consider this method as much useful, as everything that a user might need is provided by the second method below.
I guess you made this method public to be able to use it in the Transitions plugin class. In that case maybe simply move this method to the Transitions plugin class as a static one, and call it from within the other API method.

plugins/Transitions/tests/System/TransitionsMaxAllowedPeriodTest.php
```diff
@@ -0,0 +1,45 @@
...
+use Piwik\Config;
+use Piwik\Plugins\Transitions\API;
+
+/**
+ * Tests the transitions plugin max_period_allowed setting
+ *
+ * @group TransitionsMaxAllowedPeriodTest
+ * @group Plugins
+ */
+class TransitionsMaxAllowedPeriodTest extends IntegrationTestCase
```
sgiehl: Integration tests should be located at `tests/Integration/`

plugins/Transitions/tests/System/TransitionsMaxAllowedPeriodTest.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        // Attempt to get transition data for a week when the max period is set to a day
+        Config::setSetting('Transitions', 'max_period_allowed', 'day');
+        $this->expectException(\Exception::class);
+        $this->expectExceptionMessage('PeriodNotAllowed');
+        $this->api->getTransitionsForAction('http://example.org/page/one.html', 'url', self::$fixture->idSite, 'week', self::$fixture->dateTime);
+    }
+}
+
+TransitionsMaxAllowedPeriodTest::$fixture = new SomeVisitsManyPageviewsWithTransitions();
```
sgiehl: Integration tests don't really need fixture
```suggestion
```

plugins/Transitions/tests/System/TransitionsMaxAllowedPeriodTest.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public function test_ShouldThrowException_IfPeriodNotAllowed()
+    {
+        // Attempt to get transition data for a week when the max period is set to a day
+        Config::setSetting('Transitions', 'max_period_allowed', 'day');
+        $this->expectException(\Exception::class);
+        $this->expectExceptionMessage('PeriodNotAllowed');
+        $this->api->getTransitionsForAction('http://example.org/page/one.html', 'url', self::$fixture->idSite, 'week', self::$fixture->dateTime);
+    }
```
sgiehl: Would be good to add some more tests to test other periods and range edge cases. Also maybe some tests that valid requests don't return an exception

sgiehl (changes requested):
Left a couple of comments.

plugins/Transitions/tests/System/TransitionsMaxAllowedPeriodTest.php
bx80: I've added some tests for valid and invalid combinations, that ranges that are larger than periods throw exceptions and that per set config values override the general config setting.

plugins/Transitions/API.php
bx80: Makes sense, done.

plugins/CoreHome/vue/src/Periods/Range.ts
bx80: Fixed. I was incorrectly assuming the range had start and end times, but obviously it doesn't.

plugins/Transitions/API.php
bx80: Good point, I'll make a note to avoid adding any public API methods unless they'd actually be useful for someone using the API.

sgiehl (approved):
Looked through the code and tested everything locally. Works as expected 👍 
You could add an additional UI test with the period that should be shown disabled, but that's a nice to have. Besides that, feel free to merge.

PR #18371, opened 2021-11-24
Allow to distinguish between MySQL and MariaDB (also in installation)

### LINKED ISSUE #18360: Allow configuration of max execution time when using MariaDB
To provide feature equality between MySQL and MariaDB the same feature as implemented in #14858 (setting a max_execution_time for live queries) should also be supported when using MariaDB.

MariaDB uses a different syntax and seconds instead of milliseconds, but apart from that, the feature should work the same.

See https://mariadb.com/kb/en/library/aborting-statements/ for more information about this.

ping @tomper00 and @jorgeuos

sgiehl:
### Description:

This is a small draft how to solve issues that might occur in MariaDb only. 
I've only spent a couple of minutes to do the changes, so I'm not even sure if it would even fully work.

I currently don't have time to finish this PR or even test if everything works as expected.
If someone has some time, feel free to pick it up.

When finished this aims to fix #18360

This PR will change some test runs to use MariaDb instead of MySQL. This currently only affects PHP tests (Unit, Integration, System). Due to some fixtures using the same datetime for their visits, the UI tests would currently fail on MariaDb due to different sorting behaviour compared to MySQL.

Requires https://github.com/matomo-org/github-action-tests/pull/16 and https://github.com/matomo-org/plugin-MarketingCampaignsReporting/pull/152

### Review


jorgeuos:
I'm curios though if there's other functionality that looks for `SELECT` in their queries. And if it would break backwards compatibility. Haven't had time to look into that yet.
Considering the comment from @tsteur here: https://github.com/matomo-org/matomo/issues/18360#issuecomment-975820840
Mostly just core DB updates I would guess, but IDK right now.
Although don't think it's related to why the build failed: https://app.travis-ci.com/github/matomo-org/matomo/jobs/549521319#L1078 
I would need to run the tests locally first.

Pojielp (approved):

plugins/Diagnostics/Diagnostic/DatabaseAbilitiesCheck.php
```diff
@@ -47,6 +47,13 @@ public function execute()
         $result->addItem($this->checkTemporaryTables());
         $result->addItem($this->checkTransactionLevel());
 
+        $databaseVersion = Db::fetchOne('SELECT VERSION();');
+
+        if (strpos(strtolower($databaseVersion), 'mariadb')) {
```
michalkleiner:
```suggestion
        if (strpos(strtolower($databaseVersion), 'mariadb') !== false) {
```

.github/workflows/matomo-tests.yml
```diff
@@ -54,10 +61,11 @@ jobs:
           submodules: true
           path: matomo
       - name: running tests
-        uses: matomo-org/github-action-tests@main
+        uses: matomo-org/github-action-tests@mariadb
```
sgiehl:
```suggestion
        uses: matomo-org/github-action-tests@main
```

This needs to be changed back once https://github.com/matomo-org/github-action-tests/pull/16 has been merged.

plugins/Installation/FormDatabaseSetup.php
```diff
@@ -92,9 +97,10 @@ function init()
             'host'          => '127.0.0.1',
             'type'          => $defaultDatabaseType,
             'tables_prefix' => 'matomo_',
+            'schema' => 'Mysql',
```
michalkleiner:
```suggestion
            'tables_prefix' => 'matomo_',
            'schema'        => 'Mysql',
```

tests/PHPUnit/Unit/DbHelperTest.php
```diff
@@ -62,20 +64,29 @@ public function getVariousDbNames()
...
+            ['SELECT  /*+ MAX_EXECUTION_TIME(20000) */  column FROM (SELECT * FROM table)', 'SELECT column FROM (SELECT * FROM table)', 20, 'Mysql'],
+            ['SELECT * FROM table', 'SELECT * FROM table', 0, 'Mysql'],
+            ['SELECT /*+ MAX_EXECUTION_TIME(1000) */ * FROM table', 'SELECT /*+ MAX_EXECUTION_TIME(1000) */ * FROM table', 3.5, 'Mysql'], // should not append/change MAX_EXECUTION_TIME hint if already present
+            ['UPDATE table SET column = value', 'UPDATE table SET column = value', 150, 'Mysql'],
+            // MariaDB Schema
+            ['SET STATEMENT max_statement_time=2 FOR SELECT * FROM table', 'SELECT * FROM table', 1.5, 'Mariadb'],
+            ['SET STATEMENT max_statement_time=20 FOR SELECT column FROM (SELECT * FROM table)', 'SELECT column FROM (SELECT * FROM table)', 20, 'Mariadb'],
+            ['SELECT * FROM table', 'SELECT * FROM table', 0, 'Mariadb'],
+            ['SET STATEMENT max_statement_time=2 FOR SELECT * FROM table', 'SET STATEMENT max_statement_time=2 FOR SELECT * FROM table', 3.5, 'Mariadb'], // should not append/change MAX_EXECUTION_TIME hint if already present
```
michalkleiner:
```suggestion
            ['SET STATEMENT max_statement_time=2 FOR SELECT * FROM table', 'SET STATEMENT max_statement_time=2 FOR SELECT * FROM table', 3.5, 'Mariadb'], // should not append/change max_statement_time hint if already present
```

michalkleiner (approved):

plugins/UsersManager/tests/Integration/ModelTest.php
```diff
@@ -77,11 +77,16 @@ public function testGetSitesAccessFromUserMultipleSites()
         $this->model->addUserAccess($this->login, Write::ID, array(3));
         $this->model->addUserAccess($this->login, Write::ID, array(2));
         $this->model->addUserAccess($this->login, View::ID, array(1));
+        $access = $this->model->getSitesAccessFromUser($this->login);
+        // The order might differ depending on the database, so sort by 'idaction'
```
mneudert:
```suggestion
        // The order might differ depending on the database, so sort by 'site'
```

tests/PHPUnit/Fixtures/ManyVisitsWithGeoIP.php
```diff
@@ -158,7 +158,7 @@ protected function trackVisit(\MatomoTracker $t, $fixtureCounter, $visitorCounte
         }
 
         // first visit
-        $date = Date::factory($this->dateTime)->addDay($visitorCounter);
+        $date = Date::factory($this->dateTime)->addDay($visitorCounter)->addPeriod($fixtureCounter, 'SECOND');
```
mneudert:
```suggestion
        $date = Date::factory($this->dateTime)->addDay($visitorCounter)->addPeriod($fixtureCounter, 'second');
```

plugins/Diagnostics/Diagnostic/DatabaseAbilitiesCheck.php
```diff
@@ -50,6 +50,13 @@ public function execute()
         $result->addItem($this->checkTemporaryTables());
         $result->addItem($this->checkTransactionLevel());
 
+        $databaseVersion = Db::fetchOne('SELECT VERSION();');
+
+        if (strpos(strtolower($databaseVersion), 'mariadb') !== false && Config\DatabaseConfig::getConfigValue('schema') !== 'Mariadb') {
```
mneudert, line 55: Should we have that check in the other direction (with `===` operators)?

Having MariaDB configured but connecting to MySQL would break the max execution time features. Could happen during the installation, as there is no safe guard against choosing that combination.

plugins/API/tests/System/expected/test_AutoSuggestAPITest__Live.getLastVisitsDetails_range.xml
```diff
@@ -2125,10 +2125,10 @@
 	</row>
 	<row>
 		<idSite>1</idSite>
-		<idVisit>8</idVisit>
-		<visitIp>1.2.4.3</visitIp>
+		<idVisit>30</idVisit>
```
mneudert, line 2128: _Picked a semi random file just to have this as a comment thread._

The change in [`plugin/CustomVariables`](https://github.com/matomo-org/plugin-CustomVariables/compare/fixturechange) should be merged in the plugin with a following submodule update.

plugins/Diagnostics/Diagnostic/DatabaseAbilitiesCheck.php
sgiehl: You mean checking for `schema === 'Mysql'`? I did it on purpose that way, if we might add other schema later 🤔

mneudert: More like `databaseVersion !== x && configValue === x`, only saying "you have choosen poorly".

Having those checks if no `databaseVersion !== x && configValue !== x` was previously triggered should always give guidance for the correct one first, falling back to "please remove the config" if no suggestion is possible :thinking:

sgiehl: The current check will only suggest to switch to Mariadb in config if the database version contains `mariadb` and it is not yet configured. In all other cases nothing will be shown.
Can we be sure that it is not a MariaDb server if the version does not contain MariaDb?
I'm actually not sure if we can fully rely on the returned database version. Will every MariaDb server contain `mariadb` in the returned version, or could that be hidden some way? 🤔

mneudert: Of course the version can be manipulated at will: https://mariadb.com/kb/en/server-system-variables/#version

We will never be sure, but if we only show it as a "please verify your setup" information, it may also not hurt. Could also be added once we have more databases available (and find a good wording for the message) and/or some user reports :+1:

sgiehl: Ok. maybe we should rephrase it a bit and change it to informational instead of warning

sgiehl: changed that.

mneudert (approved):

michalkleiner (approved):

PR #18382, opened 2021-11-26
[Vue] migrate piwik-period-selector angularjs component

diosmosis:
### Description:

**This PR is based off of #18352.**

Changes:

* Migrate piwik-period-selector directive.
* Modify initialization of material selects in corehome.js to exclude ui datepicker controls.
* Add new MatomoUrl.updateLocation() method that updates the hash or whole URL depending on whether we are in a widgetized context or not. Refactor comparisons store to use this method.
* Modify ExpandOnClick/ExpandOnHover to allow string values for `expander` that reference Vue $refs. Without something like this, the directive will be initiated before the rest of the component is mounted, so refs for child elements will be undefined.

image

### Review


matomo.js
```diff
@@ -68,6 +68,109 @@ function ah(av,au){var at=new T(av,au);L.push(at);_paq=c(_paq,E);for(H=0;H<_paq.
...
 }return new T(au,at)},getAsyncTrackers:function(){return L},addTracker:function(av,au){var at;if(!L.length){at=ah(av,au)}else{at=L[0].addTracker(av,au)}return at},getAsyncTracker:function(ax,aw){var av;if(L&&L.length&&L[0]){av=L[0]}else{return ah(ax,aw)}if(!aw&&!ax){return av}if((!M(aw)||null===aw)&&av){aw=av.getSiteId()}if((!M(ax)||null===ax)&&av){ax=av.getTrackerUrl()}var au,at=0;for(at;at<L.length;at++){au=L[at];if(au&&String(au.getSiteId())===String(aw)&&au.getTrackerUrl()===ax){return au}}},retryMissedPluginCalls:function(){var au=al;al=[];var at=0;for(at;at<au.length;at++){aj(au[at])}}};if(typeof define==="function"&&define.amd){define("piwik",[],function(){return u});define("matomo",[],function(){return u})}return u}())}
 /*!!! pluginTrackerHook */
+
+/* GENERATED: tracker.min.js */
```
sgiehl: Guess you accidentally committed the changes to the tracker js files.

sgiehl (changes requested):
Left a comment. Otherwise the code is looking fine.
While clicking through the UI I found this regressions:
* The compare dropdown has an incorrect z-index:
image

* In range selection: when changing a date in the input fields above the date chooser are the choosers only updated when the input field looses the focus. Before it was triggered with each change (guess on keypress or so)

diosmosis:
@sgiehl 

> The compare dropdown has an incorrect z-index:

This happens on 4.5.0 too, it's a materialize issue.

> In range selection: when changing a date in the input fields above the date chooser are the choosers only updated when the input field looses the focus. Before it was triggered with each change (guess on keypress or so)

Fixed + another related bug (will now only change the date if it is a full YYYY-MM-DD date).

> Guess you accidentally committed the changes to the tracker js files.

Undid those changes.

plugins/CoreHome/vue/src/DateRangePicker/DateRangePicker.vue
```diff
@@ -5,57 +5,59 @@
...
+    <div id="calendarRangeTo">
+      <h6>
+        {{ translate('General_DateRangeTo') }}
+        <input
+          type="text"
+          id="inputCalendarTo"
+          name="inputCalendarTo"
+          class="browser-default"
+          v-model="endDateText"
+          @change="onRangeInputChanged('to', $event)"
```
sgiehl: Guess this should be `keydown` as well.
```suggestion
          @keydown="onRangeInputChanged('to', $event)"
```

sgiehl (changes requested):
Left a comment. Besides that the CustomAlerts submodule was changed in this PR as well. This should be reverted before merging. When that is changed and tests are passing, feel free to merge this one.

diosmosis:
There are a couple unrelated failures, one is fixed in another branch, but I'll create a new branch for it just so it's there. I'll look into the other one.

tassoman:
Hello there, does this migration Vue makes possible to override/modify the period selector calendar? 
My [Plugin RerUserDates](https://github.com/RegioneER/RerUserDates) rely on this but using angular, cannot hook in it .. 🤔

diosmosis:
@tassoman how specifically do you override/modify it? I can't see any angularjs code in the plugin.

diosmosis:
Hi @tassoman, I took another look and if you mean the HTML element removal your plugin does then it won't work with Vue anymore. Vue uses a virtual DOM and keeps HTML in sync with it, so if you remove an element, it will just be put back again. I would suggest a different approach:

1. when you determine the element should not be there, add a css class to `html` (eg, `remove-period-option`)
2. then in a LESS file hide the element based on that, eg:

    ```
    html.remove-period-option {
        #periodString ... {
            display: none;
        }
    }
    ```

tassoman:
Thanks for inspecting my code. I was stuck after the adoption of Angular. 
I'm going to try this trick, looks like quite short and straight. Whithout involving diving deep in the ui frontend.

PR #18383, opened 2021-11-26
disable cron task when segment is deleted

### LINKED ISSUE #16152: When a custom dimension is deleted, disable cron archiving for all segments that use this custom dimension
## Context 

When a custom dimension is delete, and possibly when it's deactivated (to be tested), then any segment that still use this custom dimension, will fail to process.
* When the segment is processed from cron, it leads to the whole archiving process to fail and error, preventing other reports to run. The error looks like: `ERROR [2018-03-12 00:12:06] Got invalid response from API request: ?module=API&method=API.get&idSite=2&period=day&date=last52&format=php&trigger=archivephp. Response was 'a:2:{s:6:"result";s:5:"error";s:7:"message";s:81:"Segment 'dimension7' is not a supported segment. - caused by plugin CustomReports";}'`
* When the segment is processed via the UI only, the user would see the error message in the UI `Segment 'dimension7' is not a supported segment. - caused by plugin CustomReports`

## Solution

The goal of this issue is to address the problem of the Cron archive only. Proposed solution:

* Whenever a custom dimension is deleted (and likely also when it is de-activated), then we look for all segments that use this custom dimension
* for each segment, we mark them as "autoArchive = false" (see SegmentEditor.add/update API parameter), which maps to auto_archive=0 in the `segment` table

As a result, the cron processing will not fail anymore, and all reports will be available, as expected.

peterhashair:
### Description:
Fixes:#16152
disable cron task when segment is deleted

### Review


peterhashair:
not sure that's the right solution. but it should skip the segment mark as deleted.

core/CronArchive/QueueConsumer.php
```diff
@@ -650,4 +657,13 @@ public function setMaxSitesToProcess($newValue = null)
...
     }
+
+    /*
+     *Check if segment is deleted
+     *
+     */
+    public function isSegmentDeleted()
+    {
+
+    }
```
sgiehl: Guess this can be removed again

core/CronArchive/QueueConsumer.php
```diff
@@ -546,6 +546,13 @@ private function findSegmentForArchive(&$archive)
             return false;
         }
 
+        //if Segment deleted
+        if ($storedSegment['deleted']) {
```
sgiehl: This should actually never happen, as the method above only loads segments that aren't deleted.
The problem isn't about deleted segments. The problem are segments that contain dimensions that don't exist anymore. I'm not fully sure if this would be the correct place, but it might maybe already help to simply try initializing the segment, and catch possible exceptions. e.g.
```php
try {
    $segment = new Segment($archive['segment']);
} catch (\Exception $e) {
    $this->logger->debug("Segment is not valid anymore.");
    return false;
}
```

sgiehl (changes requested):
Your approach won't work.
Also this would only solve one part of the requirements described here: https://github.com/matomo-org/matomo/issues/16152#issuecomment-372429476

peterhashair:
@sgiehl right, that makes more sense now. Update a little bit, hopeful I am on the right direction :)

core/CronArchive/QueueConsumer.php
```diff
@@ -550,6 +555,23 @@ private function findSegmentForArchive(&$archive)
         return $this->segmentArchiving->isAutoArchivingEnabledFor($storedSegment);
     }
 
+
+    /**
+     * check if segments that contain dimensions that don't exist anymore
+     * @param $invalidatedArchive
+     * @return bool
+     */
+    private function isSegmentAvaible($invalidatedArchive)
```
sgiehl: There's a typo in the method name

core/CronArchive/QueueConsumer.php
```diff
@@ -201,6 +201,11 @@ public function getNextArchivesToProcess()
                 break;
             }
 
+            if (!$this->isSegmentAvaible($invalidatedArchive)) {
+                $this->logger->debug("Segment is not valid anymore.");
```
sgiehl: This is already logged in the method, so no need to do that twice.

core/CronArchive/QueueConsumer.php
```diff
@@ -550,6 +555,23 @@ private function findSegmentForArchive(&$archive)
...
+     * check if segments that contain dimensions that don't exist anymore
+     * @param $invalidatedArchive
+     * @return bool
+     */
+    private function isSegmentAvaible($invalidatedArchive)
+    {
+        try {
+            new Segment($invalidatedArchive['segment'], [$invalidatedArchive['idsite']]);
+           } catch (\Exception $e) {
+            $this->logger->debug("Segment is not valid anymore.");
```
sgiehl: You could add some details to the log message, like the segment and idsite, so it's possible to check which segment it it caused.

core/CronArchive/QueueConsumer.php
```diff
@@ -201,6 +201,11 @@ public function getNextArchivesToProcess()
                 break;
             }
 
+            if (!$this->isSegmentAvaible($invalidatedArchive)) {
+                $this->logger->debug("Segment is not valid anymore.");
+                break;
```
sgiehl: Actually I don't think that a break is correct here. That would actually stop the whole invalidation queue. It should more likely be a `continue`. In addition it could make sense to remove the invalidation, so it doesn't happen in each run. Similar to the block below, where it checks if the plugin is still active.

sgiehl (changes requested):

core/CronArchive/QueueConsumer.php
sgiehl: The removal of the invalidation is still missing. The other changes look fine.

core/CronArchive/QueueConsumer.php
```diff
@@ -201,6 +201,11 @@ public function getNextArchivesToProcess()
                 break;
             }
 
+            //check if isSegmentAvailable
```
sgiehl: That comment is kind of useless as it doesn't provide any additional information than the code below.

core/CronArchive/QueueConsumer.php
```diff
@@ -201,6 +201,11 @@ public function getNextArchivesToProcess()
...
+            if (!$this->isSegmentAvailable($invalidatedArchive) && !Manager::getInstance()->isPluginActivated($invalidatedArchive['plugin'])) {
```
sgiehl: The plugin check shouldn't be needed here as it's done a couple of lines below. Actually it might even fail if the plugin is empty.

core/CronArchive/QueueConsumer.php
```diff
@@ -201,6 +201,10 @@ public function getNextArchivesToProcess()
                 break;
             }
 
+            if (!Manager::getInstance()->isPluginActivated($invalidatedArchive['plugin'])) {
```
sgiehl: I guess now you accidentally removed the wrong part 🙈

core/CronArchive/QueueConsumer.php
```diff
@@ -201,6 +201,10 @@ public function getNextArchivesToProcess()
                 break;
             }
 
+            if (!$this->isSegmentAvailable($invalidatedArchive)) {
+                continue;
```
sgiehl: Ok. So this would mean a segment that doesn't exist anymore or is invalid for other reasons will simply be skipped. But as the record isn't removed it will be checked with each run again. I'm not sure if there would be any benefit in keeping that record, besides the fact, that it would get invalidated as soon as it is valid again. Guess should be safe to remove it. Or am I seeing anything wrong here @tsteur 
```suggestion
                $this->model->deleteInvalidations([$invalidatedArchive]);
                continue;
```

Pojielp (approved):

tests/PHPUnit/Integration/CronArchive/QueueConsumerTest.php
```diff
@@ -715,6 +715,137 @@ public function test_skipSegmentsToday()
         $this->assertEquals($uniqueInvalidationDescs, $invalidationDescs, "Found duplicate archives being processed.");
     }
 
+    public function testDeletedSegment()
+    {
```
sgiehl: If I see that correctly this test checks how a segment is handled that was removed from the segment table.
This actually doesn't reflect what the fix is meant for. It should actually ensure that segments that might still exist, but got invalid (e.g. due to removing a plugin) don't break anything.
So a proper test would be to (maybe manually) insert invalidations for an invalid segment (e.g. a condition that is invalid like `InvalidSegment==val`) and check if running the cron doesn't fail, but silently removes such invalidations.

peterhashair: @sgiehl just checking, if I am right. So I have to manually insert a record into `archive_invalidations` table with invalid segment, but it seems like the `archive_invalidations` reset all the time the manually insert will be clear.

sgiehl: The archiving and invalidations stuff is quite complex. It's even hard for me to say something without looking deeper into it. But I would guess that it might work with adding it to the ReArchiveList like here:
https://github.com/matomo-org/matomo/blob/fe255ed247035b2deaadd058cbbaf041a92d3ad7/core/Archive/ArchiveInvalidator.php#L541-L548

peterhashair:
@sgiehl Have a deep look at this one, if I am right I think a new function doesn't even needed, it seems like already checked in here. I guess it's somewhere else.

https://github.com/matomo-org/matomo/blob/8e1aabfc68ad16ea8af059a9108a54e6053aa414/core/CronArchive/SegmentArchiving.php#L83-L104

sgiehl:
@peterhashair Were you actually able to reproduce the error that is described in the issue in any way?
The only way I was able to reproduce the exact same error message was when I disabled a plugin while archiving was running. So the archiving requests then failed due to the missing segments. If the segment is already invalid before the archiving starts it seems to be skipped without resulting in an error.

I have created a test case, that reflects the way I was able to reproduce. So the process that starts the archiving receives a segment from `UserLanguage` plugin. But for the archiving requests that are sent the plugin is disabled, causing the segment to be invalid.
The test currently fails. I guess the target would be to adjust the archiving, so such requests with an invalid segment won't throw an error or that specific error is handled properly, without resulting in a failure.

If you were able to reproduce that another way, feel free to add an additional test case for that as well.

peterhashair:
@sgiehl I saw the new test, reproduced it. Should be fixed now.

image

image

sgiehl:
@peterhashair Actually you have adjusted the test case so it covers another problem. Disabling the plugin before the archiver runs, let's the archiver skip the segments. That actually was already the case before, even though the message was not that clear.
But the error message of the original issue is
```
ERROR [2018-03-12 00:12:06] Got invalid response from API request: ?module=API&method=API.get&idSite=2&period=day&date=last52&format=php&trigger=archivephp. Response was 'a:2:{s:6:"result";s:5:"error";s:7:"message";s:81:"Segment 'dimension7' is not a supported segment. - caused by plugin CustomReports";}'
```
That actually happens if the archiver already started, but while looping over the archiving requests the plugin gets disabled. So a request for archiving an invalid segment is sent, causing the error above.

I guess we need to ignore if a archiving request results in that specific error. And log an info about that instead.

peterhashair:
@sgiehl right, that makes more sense. Update a little. Hopefully is correct :)

core/CronArchive.php
```diff
@@ -715,6 +715,12 @@ public function logError($m)
 
     private function logNetworkError($url, $response)
     {
+
+        if (preg_match("/Segment (.*?) is not a supported segment/i", $response, $match)) {
+            $this->logger->info($match[0]);
```
sgiehl, line 725: That already looks fine. But the log then contains 
```
Segment 'languageCode' is not a supported segment\n
Archived website id 1, period = day, date = 2019-12-12, segment = 'languageCode==fr', 0 visits found. Time elapsed: 4.590s\n
```

Might be fine, but actually it didn't archive anything as the segment was invalid.
If easily possible it would be nice to have something like this instead:

```
Segment 'languageCode' is not a supported segment\n
Skipped archiving website id 1, period = day, date = 2019-12-12, segment = 'languageCode==fr'. Time elapsed: 4.590s\n
```

sgiehl (commented):
Left a comment. If it's to much effort to change that, I guess it should also be good to merge it that way.

peterhashair:
@sgiehl update the log info. 😀

sgiehl (approved):
Looks good now.

PR #18386, opened 2021-11-26
Ensure javascript of a datatable only initializes itself

### LINKED ISSUE #18373: Goals by dimension view might not fully show
image

I think this has happened the first time around merging https://github.com/matomo-org/matomo/pull/18243/files 

I can reproduce this sometimes locally when I have say 15 goals, and load a range etc.

It's like the "by dimension" view loads faster than any of the goal reports and then maybe thinks there's less width available or something.

Not 100% sure it's a regression but haven't really noticed this before

sgiehl:
### Description:

When a report containing some kind of datatable is loaded, the delivered code also includes some javascript to initialize the datatable (see `_dataTableJS.twig `). We used to simply initialize all (new) datatables found on the page. This actually led to #18373. When switching the date range, all reports on the goals page are reloaded. In most cases the evolution chart is finished first. This chart already triggers a datatable initialize. In most cases the detail report below is already loaded but not yet rendered into the page. The javascript already tries to initialize the datatable, resulting in the weird behavior. 

This PR will change this, so only the report contained in the response (based on the reportid) will be initialized.
Locally that seems to work fine for me.

fixes #18373 

### Review

### DIFF
--- a/plugins/CoreHome/javascripts/dataTable.js
+++ b/plugins/CoreHome/javascripts/dataTable.js
@@ -40,7 +40,7 @@ function DataTable(element) {
 DataTable._footerIconHandlers = {};
 
 DataTable.initNewDataTables = function (reportId) {
-    var selector = reportId ? '[data-report="'+reportId+'"]' : 'div.dataTable';
+    var selector = typeof reportId === 'string' ? '[data-report='+JSON.stringify(reportId)+']' : 'div.dataTable';
     $(selector).each(function () {
         if (!$(this).attr('id')) {
             var tableType = $(this).attr('data-table-type') || 'DataTable',
--- a/plugins/CoreHome/javascripts/dataTable.js
+++ b/plugins/CoreHome/javascripts/dataTable.js
@@ -39,8 +39,9 @@ function DataTable(element) {
 
 DataTable._footerIconHandlers = {};
 
-DataTable.initNewDataTables = function () {
-    $('div.dataTable').each(function () {
+DataTable.initNewDataTables = function (reportId) {
+    var selector = reportId ? '[data-report="'+reportId+'"]' : 'div.dataTable';
+    $(selector).each(function () {
         if (!$(this).attr('id')) {
             var tableType = $(this).attr('data-table-type') || 'DataTable',
                 klass = require('piwik/UI')[tableType] || require(tableType);
@@ -376,7 +377,7 @@ $.extend(DataTable.prototype, UIControl.prototype, {
         self.handleSearchBox(domElem);
         self.handleColumnDocumentation(domElem);
         self.handleRowActions(domElem);
-		self.handleCellTooltips(domElem);
+        self.handleCellTooltips(domElem);
         self.handleRelatedReports(domElem);
         self.handleTriggeredEvents(domElem);
         self.handleColumnHighlighting(domElem);
--- a/plugins/CoreHome/javascripts/dataTable.js
+++ b/plugins/CoreHome/javascripts/dataTable.js
@@ -1956,17 +1956,13 @@ $.extend(DataTable.prototype, UIControl.prototype, {
                     // ensure the tooltips of parent elements are hidden when the action tooltip is shown
                     // otherwise it can happen that tooltips for subtable rows are shown as well.
                     open: function() {
-                        var tooltip = $(this).parents().filter(function() {
-                            return jQuery.hasData(this) && $(this).data('ui-tooltip');
-                        }).tooltip('instance');
+                        var tooltip = $(this).parents('[piwik-widget]').tooltip('instance');
                         if (tooltip) {
                             tooltip.disable();
                         }
                     },
                     close: function() {
-                        var tooltip = $(this).parents().filter(function() {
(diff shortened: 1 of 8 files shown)


sgiehl:
Seems the UI tests are passing so far. There are a couple of failures that are unrelated though. All those are already fixed on `4.x-dev`. So fixing them here might result in even more merge conflicts when merging `next_release` back...

tsteur:
👍

plugins/CoreHome/javascripts/dataTable.js
```diff
@@ -39,8 +39,9 @@ function DataTable(element) {
 
 DataTable._footerIconHandlers = {};
 
-DataTable.initNewDataTables = function () {
-    $('div.dataTable').each(function () {
+DataTable.initNewDataTables = function (reportId) {
+    var selector = reportId ? '[data-report="'+reportId+'"]' : 'div.dataTable';
```
tsteur: not sure if we need to escape the selector? Not sure if we have any user generated reportIds or so that could somehow cause issues?

plugins/CoreHome/templates/_dataTableJS.twig
```diff
@@ -1,5 +1,5 @@
 <script type="text/javascript" defer="defer">
     $(document).ready(function () {
-        require('piwik/UI/DataTable').initNewDataTables();
+        require('piwik/UI/DataTable').initNewDataTables('{{ reportId }}');
```
tsteur: here we would need to escape maybe as `js`? Or use maybe `json` filter?

tsteur (changes requested):
Generally sounds like a good fix 👍  Maybe this will also prevent random other issues.

Not sure if we should maybe merge this now into the 4.6 release or 4.7 release in case it breaks something worse? If I understand it right this has been an issue for a while?

sgiehl:
@tsteur thanks for the feedback. I've applied some changes.

tsteur:
@sgiehl I just tested this with a range and I still had the issue:

image

I double checked that there was no browser caching.

sgiehl:
@tsteur moved some js calculation to simpler css usage. hope that should fix it finally. At least I wasn't able to reproduce in local testing.

plugins/CoreHome/javascripts/dataTable.js
```diff
@@ -445,19 +438,19 @@ $.extend(DataTable.prototype, UIControl.prototype, {
...
             if (parentDataTable && parentDataTable.length) {
                 // makes sure dataTableWrapper and DataTable has same size => makes sure maxLabelWidth does not get
                 // applied in getLabelWidth() since they will have the same size.
 
                 if (dataTableInCard.length) {
-                    dataTableInCard.width(maxTableWidth);
+                    dataTableInCard.css('max-width', maxTableWidth);
                 } else {
-                    parentDataTable.width(maxTableWidth);
+                    parentDataTable.css('max-width', maxTableWidth);
```
sgiehl, line 453: Changing this caused the screenshot changes. Those are actually kind of expected, as the datatable will now only use the space available on the screen rather than exactly `1200px`

plugins/CoreHome/javascripts/dataTable.js
```diff
@@ -1963,17 +1956,13 @@ $.extend(DataTable.prototype, UIControl.prototype, {
...
+                        var tooltip = $(this).parents('[piwik-widget]').tooltip('instance');
                         if (tooltip) {
                             tooltip.disable();
                         }
                     },
                     close: function() {
-                        var tooltip = $(this).parents().filter(function() {
-                            return jQuery.hasData(this) && $(this).data('ui-tooltip');
-                        }).tooltip('instance');
+                        var tooltip = $(this).parents('[piwik-widget]').tooltip('instance');
```
sgiehl, line 1965: This should be a small improvement. The widget instance above actually holds the tooltip for the rows. Before it was iterating up all parents till the body to search for tooltips. This is actually useless as we know where the tooltip is.

tsteur: 👍 🚀

tsteur (commented):
Nice, I couldn't reproduce the issue anymore and also worked in smaller resolutions. 👍  The only thing that would have been nice if in changed screenshots like https://github.com/matomo-org/matomo/pull/18386/files?short_path=606a79c#diff-606a79c8142da1db13cf950ec3f838ae72befbdcbc69bdc86874ede720ec3acc we could still see the entire output but I suppose that might not be easily doable?

I see some tests that also need updating. https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/51923/

Once tests are passing LGTM for me. If we could make the pivot test for example show the full width that be awesome but not required.

sgiehl:
@tsteur I actually had a closer look at those tests. The full page of the pivot test looks like this:
image

So the reason that it looks cut off, is actually that the resolution in tests is too low, to display the full size. Before it was displayed with a fixed width of 1200px, causing the full page to be scrollable, and that way it had enough space to be displayed fully.

The screenshots that still needs to be updated, are located in a submodule. Will do that now, before merging this

PR #18391, opened 2021-11-28
Add option to allow skipping logic that exits w/ an error code if warnings are detected in the output

diosmosis:
### Description:

This is for automation for the vue migration, specifically being able to run vue:build and tell if there is a legitimate error in the file.

### Review

### DIFF
--- a/core/Console.php
+++ b/core/Console.php
@@ -55,7 +55,7 @@ class Console extends Application
         $this->getDefinition()->addOption($option);
 
         $option = new InputOption('ignore-warn', null, InputOption::VALUE_NONE,
-            'Return 0 exit code even if there are warnings detected.');
+            'Return 0 exit code even if there are warning logs or error logs detected in the command output.');
 
         $this->getDefinition()->addOption($option);
     }
--- a/core/Console.php
+++ b/core/Console.php
@@ -53,6 +53,11 @@ class Console extends Application
         );
 
         $this->getDefinition()->addOption($option);
+
+        $option = new InputOption('ignore-warn', null, InputOption::VALUE_NONE,
+            'Return 0 exit code even if there are warnings detected.');
+
+        $this->getDefinition()->addOption($option);
     }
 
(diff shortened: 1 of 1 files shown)


bx80 (approved):

bx80 (changes requested):
Just spotted an integration test exception on the travis build:

 ```
Uncaught exception: /home/travis/build/matomo-org/matomo/vendor/symfony/console/Symfony/Component/Console/Input/Input.php(180): The "ignore-warn" option does not exist. [Query: , CLI mode: 1]
[InvalidArgumentException]                
The "ignore-warn" option does not exist.  
```

diosmosis:
@bx80 thanks for the catch!

ping @tsteur since this is an API change

tsteur:
@diosmosis not sure I fully understand the regression. As long as `ignore-warn` is optional and we keep existing behaviour it should be fine?

diosmosis:
@tsteur it's not a regression, I need the change for some automation, specifically to be able to run vue:build and check whether it actually built something rather than it failing because a WARNING was emitted (see https://github.com/matomo-org/matomo/pull/18392 for some details on the use case). I'm not sure how good the parameter name is.

tsteur:
Name is fine by me. I can't think of a better name except some really long ones but they be too long.

I would however adjust the description to also include error logs `Return 0 exit code even if there are warnings or errors detected.` Although maybe that's misleading too as it then might sound like we would catch any kind of error and return 0 exit code?

diosmosis:
@tsteur

> I would however adjust the description to also include error logs `Return 0 exit code even if there are warnings or errors detected`. Although maybe that's misleading too as it then might sound like we would catch any kind of error and return 0 exit code?

That sounds better. I can make it `Return 0 exit code even if there are warning logs or error logs detected in the command output.` to be more specific.

tsteur:
Awesome, thanks 👍

PR #18398, opened 2021-11-29
Add system check warnings for php-fpm and nginx if config files are accessible

### LINKED ISSUE #13589: Detect Nginx during install and suggest taking a look at the official config
Matomo can be used with Nginx instead of Apache.
As Nginx doesn't recognize .htaccess (by default) it is important to use a proper nginx config that blocks access to e.g. `tmp/`.

Therefore I'd suggest detecting nginx during the installer and show a notice to the admin urging them to take a look at the Matomo nginx config.

One way to detect it would be using `$_SERVER["SERVER_SOFTWARE"]` which returns `nginx/1.14.0` for me. While it isn't 100% accurate, it should be okay for a notice.

### LINKED ISSUE #18132: error private directories are accessible
The security check fails, but security files have been generated.

./console core:create-security-files

## Expected Behavior
No errors

## Current Behavior
We found that the above URLs are accessible via the browser, but they should NOT be. Allowing them to be accessed can pose a potential security risk since the contents can provide information about your server and potentially your users. Please restrict access to them.

We also found that Matomo's config directory is publicly accessible. While attackers can't read the config now, if your webserver stops executing PHP files for some reason, your MySQL credentials and other information will be available to anyone. Please check your webserver config and deny access to this directory.

## Possible Solution
Fix bug.

1. Upgrade to Matomo 4.5

## Context
None

## Your Environment

* Matomo Version: 4.5
* PHP Version: 7.3
* Server Operating System: Debian 9
* Browser: Google Chrome
* Operating System: Windows 10 20H2

bx80:
### Description:

Fixes #13589
Fixes #18132

Nginx doesn't use .htaccess files by default, therefore will not apply access restriction rules for the /config directory
Apache will ignore .htaccess files for any php files passed to php-fpm via fpm-fcgi, therefore will also not apply access restriction rules.

This PR adds additional checks to the system diagnostics page to detect these scenarios and provide informative warnings.

Accessibility of the /config/global.ini.php files is used a general test of whether the access rules are being applied. If the global config file is not accessible then no warnings will be shown even if running php-fpm or nginx, as we can assume that non-htaccess rules have been added to the server config and we don't want to show a warning once the issue has been corrected.

For the PHP-SAPI check:
If the config is accessible and fpm-fcgi is being used...
 ...if Apache then show a warning suggesting adding a ProxyPass rule
 ...if nginx then show a warning and link to the official nginx server configuration
 ...otherwise show a generic warning that an access rule should be added to prevent access to /config

For the server info check:
If the config is accessible and nginx is being used then show a warning and link to the official nginx server configuration

### Review


plugins/Diagnostics/Diagnostic/ServerInformational.php
```diff
@@ -28,8 +28,22 @@ public function execute()
     {
         $results = [];
 
-        if ( ! empty( $_SERVER['SERVER_SOFTWARE'] ) ) {
-            $results[] = DiagnosticResult::informationalResult('Server Info', $_SERVER['SERVER_SOFTWARE']);
+        if (!empty($_SERVER['SERVER_SOFTWARE'])) {
+
+            $rpd = new RequiredPrivateDirectories($this->translator);
+            $isGlobalConfigIniAccessible = $rpd->isGlobalConfigIniAccessible();
```
sgiehl: You can't perform this check here. This information is also shown during the install. But at that state neither a `config.ini.php` is available, nor a piwik url is already set. Currently when trying to install Matomo the system check throws an exception as no database is configured.

sgiehl (changes requested):

plugins/Diagnostics/Diagnostic/ServerInformational.php
bx80: I've changed this so that the file access check will only be performed if matomo is already installed, so on the installation page the warnings will always be shown, but on the post-installation system check they will only be shown if config files are accessible. I've changed `isGlobalConfigIniAccessible()` to actually check access for `global.ini.php` not `config.ini.php` as they're both covered by the same rule.

sgiehl (commented):
The code looks fine so far. But I'm not able to test that fully, as I don't have a nginx setup locally and no time to create one to only test that. I also haven't fully read the both issues this PR should solve, so not sure if that is the case. Maybe @Findus23 can have a quick look if this PR makes sense or not?

justinvelluppillai:
@Findus23 do you have time to look at this? If not we can get @bx80 to check it again to confirm in relevant environments and then we can merge if no issues

plugins/Diagnostics/lang/en.json
```diff
@@ -29,6 +29,10 @@
...
         "ConfigIniAccessible": "We also found that Matomo's config directory is publicly accessible. While attackers can't read the config now, if your webserver stops executing PHP files for some reason, your MySQL credentials and other information will be available to anyone. Please check your webserver config and deny access to this directory.",
         "AllPrivateDirectoriesAreInaccessible": "All private directories are inaccessible from the internet.",
-        "UrlsAccessibleViaBrowser": "We found that the above URLs are accessible via the browser, but we recommend they should not be. If possible, please restrict access to them."
+        "UrlsAccessibleViaBrowser": "We found that the above URLs are accessible via the browser, but we recommend they should not be. If possible, please restrict access to them.",
+        "PHPFPMWarningApache": "PHP FPM may ignore .htaccess rules for .php files. To ensure that sensitive files cannot be accessed directly it is recommended to exclude certain directories from being handled by PHP FPM by adding the line %1$s to the %2$s section in you apache virtual host config just above the %3$s line.",
```
Findus23:
```suggestion
        "PHPFPMWarningApache": "PHP FPM may ignore .htaccess rules for .php files. To ensure that sensitive files cannot be accessed directly it is recommended to exclude certain directories from being handled by PHP FPM by adding the line %1$s to the %2$s section in your Apache virtual host config just above the %3$s line.",
```

plugins/Diagnostics/lang/en.json
```diff
@@ -29,6 +29,10 @@
...
+        "PHPFPMWarningNginx": "PHP FPM may ignore .htaccess rules for .php files. To ensure that sensitive files cannot be accessed directly it is recommended to exclude certain directories from being handled by PHP FPM. For more information please see the %1$s official nginx server configuration %2$s",
```
Findus23: For nginx I wouldn't use "may" as I couldn't think of any setup where it would be respected (as blocking things isn't the job of PHP FPM)

plugins/Diagnostics/lang/en.json
```diff
@@ -29,6 +29,10 @@
...
+        "PHPFPMWarningGeneric": "PHP FPM may ignore .htaccess rules for .php files. To ensure that sensitive files cannot be accessed directly it is recommended to configure your web server to exclude the /config directory from being handled by PHP FPM.",
+        "HtaccessWarningNginx": "To ensure that sensitive files cannot be accessed directly it is recommended to configure your web server to restrict accesss to certain directories. For more information please see the %1$s official nginx server configuration %2$s"
```
Findus23:
```suggestion
        "HtaccessWarningNginx": "To ensure that sensitive files cannot be accessed directly it is recommended to configure your web server to restrict access to certain directories. For more information please see the %1$s official nginx server configuration %2$s"
```

Findus23 (changes requested):
Just a quick spell checking

justinvelluppillai (changes requested):
Seems to work ok. Pending fixing @Findus23 changes this should be ok to merge, and then would be good to ask the users experiencing the problem originally if this resolves it for them.

plugins/Diagnostics/lang/en.json
bx80: For nginx and Apache I've changed it to `PHP FPM will ignore .htaccess rules...`. but I've left it as `...may ignore...` for other web server types as we've no way of knowing how they handle things.

justinvelluppillai (approved):

PR #18403, opened 2021-11-29
dependent archive never run

### LINKED ISSUE #18362: Dependent archives never processed for All visits segment causing some data to not show
I was testing/debugging https://github.com/matomo-org/matomo/pull/18354

Noticed there is a check in https://github.com/matomo-org/matomo/blob/4.6.0-rc2/core/ArchiveProcessor.php#L683 that when `$newSegment === $segment` we don't process the dependent archive.

However, `$newSegment === $segment` is always true when the segment is the All Visits segment is given. Then the segment is an empty string (`''`) and therefore the combine segment equals the regular segment.

Instead it probably should look like this https://github.com/matomo-org/matomo/compare/4.x-dev...processdependentarchive?quick_pull=1#diff-0423f9c2b4343926ff8deebfed3605b3f1905daa9e8fecc1ec1f00f65bf3e997R683

As when no original segment is given, then the segment will be always different.

As a result, eg Visitor Map might be showing empty data see https://demo.matomo.cloud/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=week&date=2021-11-16&segment=&category=MediaAnalytics_Media&subcategory=MediaAnalytics_TypeAudienceMap if I see this right.

I'm thinking it might still partially work by accident because `Goals.get` in the end still requests the segment and it may end up archiving it but not sure entirely. I think this looks like a bug to me?

Be great to see if we can reproduce this.

peterhashair:
### Description:
Fixes: #18362 
Fix for dependent archive never run, not sure that's the best approach

### Review


sgiehl:
@peterhashair were you able to reproduce the reported issue? I don't think that specific solution is something we want to go with. We should check for a global solution, as if that can happen for one report, it can also happen for any similar report.

peterhashair:
@sgiehl I kind of do not fully understand that one, just quickly applied @tsteur solution, I will look deep for that one. remove the need review tag for now.

tsteur:
@peterhashair we could catch up in a call tomorrow morning or afternoon if that helps?  generally I would have thought that https://github.com/matomo-org/matomo/compare/4.x-dev...processdependentarchive?quick_pull=1#diff-0423f9c2b4343926ff8deebfed3605b3f1905daa9e8fecc1ec1f00f65bf3e997R683 should fix it.

sgiehl:
The changes are logically looking fine. I was wondering if we are able to reproduce an issue that those changes will fix. If so we could possibly add a test for that, to ensure that won't regress in the future.

tsteur:
fyi @sgiehl we actually noticed that there might not have been an actual issue and it worked potentially "by luck" as MediaAnalytics requested a different plugin to be archived and then it wouldn't ever `return` in that case: https://github.com/matomo-org/matomo/blob/4.6.2/core/ArchiveProcessor.php#L683 because plugin is different. Also in the case of "goals" it still worked probably by luck because the UI would also requested these metrics.

This kind of also means there might be another bug in there as even if the segment is already now being archived, in some cases it would archive it again (I think there's an issue for cron archiving but this check may be needed for browser archiving or if only a specific plugin is requested, quite difficult). I might create a separate issue for this.

A quick test could be maybe added though

sgiehl:
@peterhashair could you check if it's possible to add a useful test for that?

peterhashair:
Had another look for that one, if I am right if `$originSegmentString` is not empty, `$newSegment === $segment` will always be false. `$newSegment === $segment` will be true only if `$originSegmentString` is empty.

plugins/Goals/tests/System/ProcessDependentArchiveTest.php
```diff
@@ -34,6 +34,12 @@ public function tearDown(): void
         parent::tearDown();
     }
 
+   public function test_numArchivesCreated_day()
+    {
+        API::getInstance()->getMetrics(self::$fixture->idSite, 'day', '2009-01-04');
+        $this->assertNumRangeArchives(5, 1); // days;
```
sgiehl, line 40: Doesn't this test have the same result on 4.x-dev? Wondering how that should cover the changes though.

tsteur: In 4.x-dev it only generates 3 archives. We debugged and tested together to make sure this tests the right use case.

sgiehl: Hm... Seems i didn't checkout the test changes correctly when I tried running that test on 4.x-dev last time. Tried that again right now, and it indeed fails on 4.x-dev...

sgiehl (approved):

PR #18404, opened 2021-11-29
Added table of contents for general settings

### LINKED ISSUE #13837: Add Table of Content to General Settings
Just a random thought, but now that Matomo has more and more features (and quite a few plugins that also have options) General Settings is becoming quite long (12137px for me)

One easier fix would be adding a table of content to the top so one could quickly jump to the plugin one wants to configure.

bx80:
### Description:

Fixes #13837

Adds a table of contents to the general settings page. Anchors and table of content links for plugin sections are automatically added by the controller.

### Review


justinvelluppillai (commented):
I've got this going on

Screen Shot 2021-12-03 at 9 58 21 AM

I can't exactly see why it's line breaking where it is.

Other than this TOC working well and code looks good.

tsteur (commented):
FYI I've done a quick test here as well because I saw the screenshot and was wondering if we could maybe somehow shorten some of the titles because they are quite long and it takes quite a bit to read/scan this menu? Like  could we rename somehow `CONFIGURATION FOR SERVER VARIABLES USED BY GEOIP 2 SERVER` to `GeoIp`

btw the link for geoip did not work for me 
image


Also noticed the CORS link is missing 
image

I'm also wondering if maybe regular case would be better readable 
image

and thinking maybe we have to put a space before capital letters there just to have it bit more human readable? We're doing the same on the [Marketplace](https://plugins.matomo.org/) using something like below but it's in PHP

```
    function matomoAddSpacesToCamcelCase($name)
    {
        $parts = preg_split('/(?=[A-Z0-9])/', $name);
        $name = '';

        foreach ($parts as $part) {
            if (strlen($part) > 1) {
                $name .= ' ';
            }
            $name .= $part;
        }

        return trim($name);
    }
```

bx80:
The text wrapping issue was caused by there not being a suitable white space character between links, they look like they have spaces but it's actually a right margin applied by the materialize classes.

The long plugin names were because I was using the plugin title rather than the plugin name, I've switched to using the plugin name which is shorter, this has also fixed the broken GeoIp2 link.

The 'CORS' and 'Trusted Matomo Hostname' sections both come from the `CoreAdminHome` plugin, so creating menu items based on the plugin name didn't work, I've added some extra code to handle `CoreAdminHome` sub sections.

I've also overridden the materialize card uppercase text CSS and added a space before each capital letter for plugin names, which does look a lot better.

justinvelluppillai (approved):
Looks good to me. Not sure if we want the tracking code TOC to look the same now (maybe a separate issue anyway)?

Will leave open in case @tsteur has any other feedback.

tsteur:
@bx80 fyi just saw there's actually maybe an issue with SEOWebVitals plugin where it makes `S E O Web Analytics`

image

PR #18405, opened 2021-11-30
Rendering performance improvement for reports with many rows

### LINKED ISSUE #18394: Rendering performance improvement for reports with many rows
Refs L3-181

When loading User ID reports with several thousand rows (For example 3k-10k rows) the report generation in the browser causes very high system resource usage on the device loading the report.

For example an matomo server that has a User ID row limit of 10,000 and selecting the row limit in the UI to "All" takes anywhere from 3-6 minutes (or more) to generate the report in the browser. During this time, that browser tab uses for example 3gb+ of RAM and presents the popup "Browser tab is unresponsive"

If the device being used doesn't have enough spare RAM or just isn't powerful enough to run the javascript to generate the report, that report can basically get stuck for that user and would force them to reset their visualisations to fix it (https://matomo.org/faq/troubleshooting/what-do-i-do-when-the-browser-tab-becomes-unresponsive/).

bx80:
### Description:

refs #18394

Removes the obsolete handleColumnHighlighting() method which causes a significant performance hit when rendering a datatable with many rows.

Column width and row highlighting is now handled by CSS so this method should no longer be needed.

### Review


sgiehl (changes requested):
This one can't be simply removed. The row highlight is covered by CSS. But when you hover a column, that column receives a class in each row. I don't think this can be covered by CSS. Now the ratio values aren't displayed (e.g. for visits or page views) anymore as it relies on that class. See 
https://github.com/matomo-org/matomo/blob/8066b3536c2399e1c2992c9ee4607b3985b94b42/plugins/CoreHome/stylesheets/dataTable/_dataTable.less#L36-L38

bx80:
The 'give all values consistent width' part of handleColumnHighlighting() is the real performance hit, adding mouse event handlers to add/remove classes doesn't really impact loading speed. So I've restored the method and tested again with just the width setting code disabled and performance is basically the same as with the handleColumnHighlighting() method disabled.
image
This means the ratio values will work and any other code that relies on the column highlight class being set.

bx80:
@sgiehl Following the theme of naively deleting random datatable functions... :wink:

I think the performance heavy setFixWidthToMakeEllipsisWork() method can be replaced with a small CSS tweak. Setting the table cell width to calc(80%) should allow the ellipsis to be applied to the table label column using pure CSS without needing javascript to calculate and set a fixed cell width both initially and every time the table is resized. 

I've committed this change for review and it seems to work well in the tests I've done. Can you think of any problems with this approach or anything I should specifically test?

plugins/CoreHome/stylesheets/dataTable/_dataTable.less
```diff
@@ -78,7 +78,7 @@ table.dataTable tr.subDataTable {
     word-break: break-all;
     overflow: hidden;
     text-overflow: ellipsis;
-    width: inherit;
+    width: calc(80%);
```
sgiehl: Why is that `calc` needed here? Isn't `80%` a fixed value? Also `calc` is a less method used here, to use a css `calc` it would need to be `width: ~"calc(80%)"`

sgiehl (changes requested):
Oh dear. I would really love to get rid of that method. Struggling with it already for a while in #18386. The "problem" with all those methods is, that they are doing a lot more, than the name would let you think. The method `setFixWidthToMakeEllipsisWork` is actually also used to limit the width of a datatable to `1200px` (if it's not widgetized), but it also ensures that a datatable doesn't always take all the space that would be available, but only what it really needs. In addition if fixes the size of the tables in "reportByDimension" view, which can be seen in the goal reports (the one with the additional dimension menu on the left). And I'm pretty sure it's responsible for some more stuff I don't see right now.

Removing that method actually broke most of the additional stuff...
Maybe for now remove that again, so we can already merge the other improvement. If needed you can then create new PRs with further improvements.

plugins/CoreHome/stylesheets/dataTable/_dataTable.less
bx80: My mistake, I was thinking CSS. Ellipsis need a fixed width in order to work, calc() outputs a fixed width but handles resizing etc without javascript. Clearly there is a lot more going on with `setFixWidthToMakeEllipsisWork` than I'd initially realized, thanks for pointing it out. I'll revert the change and back away slowly.

sgiehl (changes requested):
The code changes look fine. But there are actually a lot failures in the UI tests. Once the new build is finished you may need to look through all those changes, and check if they are actually expected, or if there are some regressions, where the table sizing maybe doesn't work correctly anymore.

bx80:
I've updated all the screenshots for UI tests that failed from this change.

sgiehl (approved):
Looks got to merge now. 
Btw. you forgot to create PR in submodule plugin to update the screenshots. I've done that for you. Also changing any submodule to another branch than 4.x-dev always needs to be reverted before a PR is merged. I'll do that once I saw if all related screenshot tests are passing.

PR #18419, opened 2021-12-01
[Vue] emit/broadcast on scope the method was called on not rootScope

diosmosis:
### Description:

AngularJS scopes inherit directly from their parent, so overwriting $emit/$broadcast on the $rootScope changes it for everyone. Which is a problem if the overwritten function uses $rootScope directly to broadcast, and not the given scope. ('$destroy' on a child scope for example can destroy many scopes it shouldn't).

### Review


sgiehl (approved):

tsteur:
Do we know what this breaks for the user? Generally we could merge it as long as we don't break anything else with it. 
For it to be in 4.6.2 we would need a PR against `next_release` branch

sgiehl:
This caused at least the problem in ActivityLog, not sure if anything else is still undiscovered

diosmosis:
Updated the base branch

tsteur:
@diosmosis some UI tests are now failing, not sure if this is expected?

diosmosis:
@tsteur no not expected, looking into it now

diosmosis:
@tsteur most failures are random, the rest are due to database version being different in the screenshot. I updated those.

tsteur:
👍  good to merge then

tsteur:
it's probably not needed to update screenshots in `next_release` as it might be fixed then in 4.x-dev I would assume. Not sure what best flow is there

diosmosis:
@tsteur ok, I won't update (I actually pushed them to the wrong branch)

PR #18423, opened 2021-12-01
[Vue] migrate piwik-reporting-menu directive to vue

diosmosis:
### Description:

Changes:
* Add jquery.js (no min.js) to asset ordering so it can be specified in addJavaScript files if its needed to debug jquery. Without this, it will be added out of order.
* Migrate side-nav directive.
* Migrate reporting-menu model service to store.
* Migrate piwik-reporting-menu directive to vue.
* Add a $timeout() call after AjaxHelper finishes a request (since angularjs triggers a digest after $http uses).
* Move URL comma replacing logic to MatomoUrl.
* Allow prepending notifications even if not going to a modal dialog.
* Make reporting pages store adapter's more immutable.
* Add clone helper method for creating angularjs adapters (angularjs adapter results are not allowed to modify Vue store data, but must still be mutable because angularjs expects it).
* Fix to jqueryNativeEventTrigger.ts for click events on links.
* Add `window.widgetsHelper` typings to index.d.ts.

image

### Review


peterhashair:
not sure if this is done in another PR, but it seems like the menu has a function that groups large datasets to a autocomplete search box 

Expected
image

Current
image

diosmosis:
Tests should be passing for this one. Fixed a bunch of smaller bugs towards the end.

core/AssetManager/UIAssetFetcher/JScriptUIAssetFetcher.php
```diff
@@ -74,8 +74,10 @@ protected function getPriorityOrder()
     {
         return array(
             'node_modules/jquery/dist/jquery.min.js',
+            'node_modules/jquery/dist/jquery.js',
             'node_modules/materialize-css/dist/js/materialize.min.js', // so jquery ui datepicker overrides materializecss
             'node_modules/jquery-ui-dist/jquery-ui.min.js',
+            'node_modules/jquery-ui-dist/jquery-ui.js',
```
diosmosis, line 80: Added so jquery/jquery-ui can be debugged using non-minified JS.

plugins/CoreHome/vue/src/SideNav/SideNav.adapter.ts
```diff
@@ -0,0 +1,35 @@
...
+    priority: 10,
+    link: function linkPiwikSideNav(scope, element, attr) {
+      const binding = {
+        instance: null,
+        value: {
+          activator: $(attr.piwikSideNav)[0],
+        },
+        oldValue: null,
+        modifiers: {},
+        dir: {},
```
sgiehl, line 23: Just out of curiosity: Are those attributes needed for anything? Can't see that they are used anywhere 🤔

plugins/CoreHome/vue/src/ReportingMenu/ReportingMenu.store.ts
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  private sortMenuItems<T extends Orderable>(menu: T[]): T[] {
+    const result = [...menu];
+    result.sort((lhs, rhs) => {
+      if (lhs.order < rhs.order) {
+        return -1;
+      }
+
+      if (rhs.order > lhs.order) {
```
sgiehl: Isn't that the same as the check above? Guess it should be
```suggestion
      if (lhs.order > rhs.order) {
```

plugins/Dashboard/tests/UI/DashboardManager_spec.js
```diff
@@ -98,7 +98,7 @@ describe("DashboardManager", function () {
 
         expect(await page.screenshot({ fullPage: true })).to.matchImage('create_new');
     });
-
+return;
```
sgiehl:
```suggestion

```

plugins/CoreHome/vue/src/ReportingMenu/ReportingMenu.vue
```diff
@@ -0,0 +1,359 @@
...
+        <a
+          class="item"
+          tabindex="5"
+          href=""
+          @click.prevent="loadCategory(category)"
+        >
+          <span
+            :class="`menu-icon ${category.icon ? category.icon : 'icon-arrow-right'}`"
+          />
+          {{ category.name }}
```
sgiehl: The additional space between the icon and the menu title actually causes the "regression" that can be seen in the UI file changes. Long menu items are now always wrapped into a new line after the icon. This should fix that:
```suggestion
          />{{ category.name }}
```

plugins/CoreHome/vue/src/SiteSelector/SiteSelector.adapter.ts
```diff
@@ -82,11 +75,18 @@ export default createAngularJsAdapter<[ITimeoutService]>({
...
     });
 
+    if (attrs.siteid && attrs.sitename) {
+      vm.modelValue = { id: attrs.siteid, name: Matomo.helper.htmlDecode(attrs.sitename) };
+    } else if (Matomo.idSite) {
+      vm.modelValue = {
+        id: Matomo.idSite,
+        name: Matomo.helper.htmlDecode(Matomo.siteName),
+      };
+    }
```
sgiehl: Did you do those changes on purpose? It seems now pre selection in the site selector don't work anymore. You can see that in the siteselector within the usersettings.

sgiehl (changes requested):
Left a couple of comments for stuff I found while looking through the code & test changes. Did not yet complete the manual testing. will add further comments if I find anything else...

plugins/CoreHome/vue/src/SideNav/SideNav.adapter.ts
diosmosis: These are required by the `DirectiveBinding` interface in Vue. If you remove them, phpstorm should complain. Typescript will also complain, but possibly not when calling vue:build. For the adapter we're passing a fake object just to trigger the directive logic correctly.

plugins/CoreHome/vue/src/SiteSelector/SiteSelector.adapter.ts
diosmosis: They were to fix another issue. Seems like the new problem is due to two reasons:

1. The original `vm.modelValue =` in postCreate() wasn't being applied. Moved it to a default() function in the adapter definition so the vue app data would be initiated to that value.
2. It turns out if the `default` value for a Vue property is a factory function, it is applied no matter what, even if an original value is supplied. Now we check if there's a value set first.

Hopefully the fix won't break something elsewhere.

plugins/CoreHome/vue/src/ReportingMenu/ReportingMenu.vue
diosmosis: Applied.

plugins/Dashboard/tests/UI/DashboardManager_spec.js
diosmosis: Removed.

plugins/CoreHome/vue/src/ReportingMenu/ReportingMenu.store.ts
diosmosis: Good catch, thanks :+1:

plugins/CoreHome/vue/src/SiteSelector/SiteSelector.adapter.ts
diosmosis: UI tests need to be updated (I'll do that if the PR is reviewed, otherwise tomorrow), but it looks like the change worked.

plugins/CoreHome/stylesheets/layout.less
```diff
@@ -190,7 +190,7 @@ nav {
       padding-bottom: 20px;
 
       .menu-icon {
-        padding-right: 13px;
+        padding-right: 9.5px;
```
sgiehl: This needs to be reverted, guess it was needed due to the additional space between icon and text.

sgiehl (changes requested):
Had a quick look at the UI file changes. Seems one css change needs to be reverted. And on the PrivacyManager tests it looks like that before the `All Websites` entry was preselected in the site selector. Now that switched to a specific site. Would it be easy to change that? Otherwise we could also define that as expected behavior and add according attributes to the site selector on the privacy settings pages, so that `all websites` is preselected

diosmosis:
@sgiehl

> And on the PrivacyManager tests it looks like that before the All Websites entry was preselected in the site selector. Now that switched to a specific site. Would it be easy to change that?

Oh, I thought it didn't have an initial site set, but it is initialized in the controller. Looked into a bit, this is a different issue, though I'm not sure what... Setting data on the vue app isn't working, it doesn't propagate to the root component in the app. Will see if I can figure it out.

diosmosis:
@sgiehl fixed it w/ an extra nextTick(). Not sure why this works, but it seems to.

plugins/CoreHome/stylesheets/layout.less
sgiehl:
```suggestion
        padding-right: 13px;
```

plugins/CoreHome/vue/src/AjaxHelper/AjaxHelper.ts
```diff
@@ -358,13 +359,26 @@ export default class AjaxHelper<T = any> { // eslint-disable-line
...
+    try {
+      $timeout = Matomo.helper.getAngularDependency('$timeout');
+    } catch (e) {
+      // ignore
+    }
+
     const result: AbortablePromise<T> = new Promise<T>((resolve, reject) => {
-      this.requestHandle!.then(resolve).fail((xhr: jqXHR) => {
+      this.requestHandle!.then((...args) => {
+        resolve(...args);
```
sgiehl: Seems the tests are failing because of this line
```
plugins/CoreHome/vue/src/AjaxHelper/AjaxHelper.ts:371:17 - error TS2556: A spread argument must either have a tuple type or be passed to a rest parameter.
371         resolve(...args);
```

diosmosis: :+1: fixed in last commit

sgiehl:
@diosmosis there still seem to be some changes in the UI files that look unexpected. The usermanagers results changed, but imho they shouldn't have

peterhashair (approved):
working as expected, I think the test is being fixed.

diosmosis:
@sgiehl: @peterhashair was faster by a couple minutes, but it looks like the change was a fix to some broken behavior, the first failure is for `role_for.png`, in the test we select for a site named "relentless": https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/UsersManager/tests/UI/UsersManager_spec.js#L68, but in the 4.x-dev screenshot the wrong site is selected. Will merge in a couple hours if there are no further comments.

sgiehl:
@diosmosis I guess some ui changes might still be some sort of incorrect:
`tests/UI/expected-screenshots/Menus_admin_loaded.png` shouldn't have been updated. Guess that one will now fail on `4.x-dev`. Maybe you can directly fix that there.
`plugins/Dashboard/tests/UI/expected-screenshots/DashboardManager_removed.png` did you check why the menu switched from being displayed fully, to be collapsible? Fo me it looks like the menu isn't updated anymore when a dashboard has been removed.

diosmosis:
> Maybe you can directly fix that there.
plugins/Dashboard/tests/UI/expected-screenshots/DashboardManager_removed.png did you check why the menu switched from being displayed fully, to be collapsible?

This works for me locally (when using tests/PHPUnit/proxy). It might be a race condition. The menu is updated, that's something I specifically had to fix (specifically due to a race condition w/ the jquery dashboard code not waiting for dashboard load to finish) though it might be possible the fix is in a different PR. I'll look a bit more to see if I can reproduce it.

PR #18430, opened 2021-12-02
[Vue] migrate reporting page directive

diosmosis:
### Description:

**This PR is based off of #18429.**

Changes:
* Migrate piwik-reporting-page directive.
* Migrate reportingPageModel service to store.
* Moving interfaces around so different services can use them.

image

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #18431, opened 2021-12-02
[Vue] add site selector model adapter and fix siteselector.vue bug

diosmosis:
### Description:

**This PR is based off of #18430.**

Changes:
* Add an adapter for sitesstore => siteselector model so the angularjs code can be removed.\
* Fix issue in SiteSelector.vue, the previous angularjs version only set the current site to the first site in the list if there was only one site in the Matomo overall. Changing that breaks the manage rollup field.

image

### Review


diosmosis:
UI tests should be passing for this PR

peterhashair:
@diosmosis working as expected, test VUE parts, interface, and CSS. I did conflicts resolve for this PR. Can you review it, in case I merge the wrong parts?

diosmosis:
@peterhashair I'll check eventually, but these PRs are all based on each other like a chain. The next one to review is https://github.com/matomo-org/matomo/pull/18430, then when that gets merged I'll merge into the next PR, etc. This PR, eg, contains changes in other PRs so I wouldn't want to get this merged right away, instead would look at the PR it's based on first.

This is the link I've been using to look at the PRs in the order that they were created: https://github.com/matomo-org/matomo/pulls?q=is%3Apr+%5BVue%5D+sort%3Acreated-asc+is%3Aopen

plugins/CoreHome/vue/src/ReportingPage/ReportingPage.vue
```diff
@@ -0,0 +1,292 @@
...
+    type: 'transient',
+  });
+}
+
+function hideOnlyRawDataNoticifation() {
+  NotificationsStore.remove('onlyRawData');
+}
+
+export default defineComponent({
+  props: {
```
peterhashair: it seems an empty props here, maybe can be remove

plugins/CoreHome/vue/src/ReportingPage/ReportingPage.vue
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        this.hasNoVisits = true;
+
+        if (this.hasRawData) {
+          showOnlyRawDataNotification();
+          return;
+        }
+
+        AjaxHelper.fetch({
```
peterhashair: (just suggestion doesn't affect the performance or function) it seems like ajax is waiting another ajax finished. maybe we can use await.
```javascript
   async showOnlyRawDataMessageIfRequired() {
      try {
        const json = await AjaxHelper.fetch({ method: 'VisitsSummary.getVisits' });
      } catch (e) {
      }
    },
```

peterhashair (commented):
Review Vue and interface, working as expected, made some suggestions for the code style, but can be ignored 😀

plugins/CoreHome/vue/src/ReportingPage/ReportingPage.vue
diosmosis: I'll look into using .then(). FYI, the Vue migration doesn't support using async/await since the regenerator runtime is not added (it would have to be added to every UMD not just MatomoPolyfills). There are details in one of the first [Vue] prs.

diosmosis:
@peterhashair fixed

sgiehl:
@diosmosis this seems to work fine. While testing I discovered anther bug in RollUpReporting that seems to be caused by the vue migration (at least it worked with Matomo 4.5).
When the `all sites` object is added to an roll up, it can't be removed anymore. There is no javascript error, but the dom keeps updating the site elements, without really removing it...
Feel free to merge this one and create another PR for that if you find a fix.

diosmosis:
@sgiehl :+1: just found the fix, will create a new PR

PR #18432, opened 2021-12-02
[Vue] migrate the plugin-settings directive

diosmosis:
### Description:

**This PR is based off of #18431.**

Changes:
* Migrate the plugin-settings directive.
* Allow FormField.vue to execute conditions without supplying all-settings param. Doesn't make sense in a Vue context since the parent component can keep track of the list of all values. Still there, but only for BC.
* Add AjaxHelper.post() method.

**IMPORTANT NOTE:** This PR is still using angularjs for expression evaluation. There doesn't seem to be an established library to do something similar, though a handful exist (none of them seem very popular). We'll either have to use one of those or roll our own. I'm not sure which is the best option, but suggestions are welcome.

image

### Review


diosmosis:
UI tests passing for this one (some screenshots in plugins need an update, but holding off on that until after the specific PRs are merged)

plugins/CorePluginsAdmin/vue/src/PluginSettings/PluginSettings.vue
```diff
@@ -0,0 +1,260 @@
...
+        }).modal('open');
+
+        return;
+      }
+
+      this.isSaving[requestedPlugin] = true;
+
+      const settingValuesPayload = this.getValuesForPlugin(requestedPlugin);
+
+      AjaxHelper.post(
```
peterhashair, line 194: just a question. do we want to implement a test for the AjaxHelper, or is actually a part of the integration UI test?

peterhashair:
@diosmosis hasn't finished reviewing the VUE code on that one, but did resolve conflicts for that one, can you review the merge conflicts, in case I merge the wrong parts.

package.json
```diff
@@ -77,6 +77,7 @@
     "nock": "^13.1.4",
     "ts-jest": "^26.5.6",
     "typescript": "^4.3.5",
-    "vue-jest": "^5.0.0-0"
+    "vue-jest": "^5.0.0-0",
+    "yet-another-abortcontroller-polyfill": "^0.0.4"
```
sgiehl, line 81: Is that one only needed for building the matomo polyfill? Guess it was on purpose that it's code wasn't added to `node_modules`?

plugins/CorePluginsAdmin/vue/src/PluginSettings/PluginSettings.vue
```diff
@@ -0,0 +1,260 @@
...
+      settingsPerPlugin: [],
+      settingValues: {},
+    };
+  },
+  created() {
+    AjaxHelper.fetch({ method: this.apiMethod }).then((settingsPerPlugin) => {
+      this.isLoading = false;
+      this.settingsPerPlugin = settingsPerPlugin;
+
+      settingsPerPlugin.forEach((settings) => {
```
sgiehl, line 112: The part that was added in https://github.com/matomo-org/matomo/pull/18404/files#diff-3fe2ecdd62c8a5f38ede22ec60896debbfe870bdde10b889f78d129fc5a3ffa2 got lost

plugins/CorePluginsAdmin/vue/src/PluginSettings/PluginSettings.vue
```diff
@@ -0,0 +1,260 @@
...
+    },
+  },
+  methods: {
+    save(requestedPlugin: string) {
+      const { saveApiMethod } = this;
+      const { root } = this.$refs;
+
+      const $root = $(root);
+
+      if (this.mode === 'admin' && !this.passwordConfirmation) {
```
sgiehl: There is kind of a bug around the `passwordConfirmation`. Wondering if it would make sense to directly apply a fix here, instead of later adjusting https://github.com/matomo-org/matomo/pull/18453

plugins/CorePluginsAdmin/vue/src/PluginSettings/PluginSettings.vue
```diff
@@ -0,0 +1,260 @@
...
+      <div class="modal-footer">
+        <a
+          href=""
+          class="modal-action modal-close btn"
+          :disabled="!passwordConfirmation ? 'disabled' : undefined"
+          @click="$event.preventDefault(); save(this.settingsToSave)"
+        >{{ translate('General_Yes') }}</a>
+        <a
+          href=""
+          class="modal-action modal-close modal-no"
```
sgiehl, line 67: Wondering if adding that might make sense, as otherwise the page reloads when clicking on "no"
```suggestion
          class="modal-action modal-close modal-no"
          @click="$event.preventDefault();"
```
Or maybe that needs to be fixed in the modal, as it wasn't the case before 🤔

plugins/CorePluginsAdmin/vue/src/PluginSettings/PluginSettings.vue
```diff
@@ -0,0 +1,260 @@
...
+
+      return () => {
+        if (!conditionScope) {
+          const $rootScope = Matomo.helper.getAngularDependency('$rootScope');
+          conditionScope = $rootScope.$new(true);
+        }
+
+        // TODO: this is definitely not as performant. would probably need a separate component
+        // for a single plugin's settings so we can make this and other types of transforms
+        // computed properties.
```
sgiehl: If that's useful and/or needed, might be good to create an follow up issue on github instead of a code comment. Otherwise I guess this will be never addressed 🙈

sgiehl (changes requested):
Generally seems to work. Found a couple of smaller issues that needs to be fixed.

package.json
diosmosis: Yes, the contents will be included in MatomoPolyfills.umd.min.js, so it's not needed to include the node_module as well in git. Only the libraries that are included directly in Matomo's asset pipeline have to live in git.

plugins/CorePluginsAdmin/vue/src/PluginSettings/PluginSettings.vue
diosmosis: If the other one is ready to merge, it could be merged and I can then convert it to Vue here?

peterhashair: should I pull this PR and make a change on here?

peterhashair:
@diosmosis did a fix for one of the issues in #18453 for this version here. When we type a wrong password in the modal, then close it, `passwordConfirmation` still has the value. Separate the `modal-show` and `save` it into 2 different functions.
Other than that, it's working as expected.

sgiehl:
@diosmosis would be good to review #18515 and merge it before pushing any further changes, to avoid merge conflicts.

plugins/CorePluginsAdmin/vue/src/PluginSettings/PluginSettings.vue
diosmosis: Adding new tests is out of scope for the Vue migration (for me at least), I'm sure it's possible to add them some day in the future however. Perhaps creating a new issue would be a good idea.

diosmosis:
@sgiehl I will take a look as soon as I can but I will be pretty busy with other things for the moment.

sgiehl:
@diosmosis no problem. Btw. I found another regression in TagManager (top controls aren't visible anymore). Shall I create a issue for that and assign it to you?

diosmosis:
@sgiehl :+1: sure, go ahead!

plugins/CorePluginsAdmin/vue/src/PluginSettings/PluginSetting.vue
```diff
@@ -0,0 +1,75 @@
...
+import FormField from '../FormField/FormField.vue';
+
+// TODO: have to use angularjs here until there's an expression evaluating alternative
+let conditionScope: IScope;
+
+export default defineComponent({
+  props: {
+    pluginName: String,
+    setting: Object,
+    modelValue: Object,
```
sgiehl: Opening the settings page logs a couple of warnings in the browser console:
image
Not sure if that's something that needs to be fixed or not.

diosmosis: 👍 worth fixing

sgiehl (approved):

PR #18433, opened 2021-12-02
[Vue] migrate remaining directives in CorePluginsAdmin

diosmosis:
### Description:

**This PR is based off of #18432.**

Changes:
* Migrate piwik-save-button.
* Migrate piwik-form.
* Migrate piwik-plugin-filter.
* Migrate piwik-plugin-management.
* Migrate piwik-plugin-upload.
* Migrate piwik-save-button.
* Allow event names to be different in vue than in angularjs.
* Support '&?' scope binding type in createAngularJsAdapter.

Changes made to make UI tests pass:
* Adding a polyfill for AbortController (since hacking an .abort() method on a promise isn't good enough). The polyfill is recommended by github's `fetch()` polyfill library: https://github.com/github/fetch. See the AbortController spec: https://developer.mozilla.org/en-US/docs/Web/API/AbortController

image

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #18440, opened 2021-12-03
[Vue] migrate report export directive and popover

diosmosis:
### Description:

**This PR is based off of #18433.**

Changes:
* Migrate piwik-report-export directive.
* Fixes an issue w/ format_metrics query param in report export where the checkbox had no effect if it was already set in the predefined request params.
* Create helper method `createVueApp()` that calls createApp and sets some globals so the globals are set in one place.

image

### Review


diosmosis:
The UI tests should be passing for this PR

peterhashair (approved):
probably need to merge the conflicts, but other than that the VUE and the JS look good to me failed tests seem unrelated. j

peterhashair (approved):

plugins/CoreHome/vue/src/ReportExport/ReportExportPopover.vue
```diff
@@ -0,0 +1,425 @@
...
+        ) {
+          exportUrlParams.include_aggregate_rows = 1;
+        }
+      }
+
+      if (!this.optionFlat && this.optionExpanded) {
+        exportUrlParams.expanded = 1;
+      }
+
+      exportUrlParams.format_metrics = this.optionFormatMetrics ? 1 : 0;
```
sgiehl: before the parameter was only set if the checkbox was checked. Guess that makes a difference, as when the value is not given it uses internally `bc` as default, which will at least format percent metrics.

sgiehl (changes requested):
Found one small regression, besides that it looks good to merge.

plugins/CoreHome/vue/src/ReportExport/ReportExportPopover.vue
diosmosis: Even if it's set to `0`? I changed this since I noticed the checkbox had no effect for evolution graphs where format_metrics was set in the datatable parameters. I tried testing just now and I cant get the report export to use bc if format_metrics is set to 0, can you confirm?

sgiehl: Not sure if I understood that right. But I see what you mean regarding the evolution charts. Maybe in that case we should manipulate the datatable parameters and remove those that can be set in the export popover and maybe use the value from datatable parameters as default?

diosmosis: I think we can just set it based on whether the checkbox is checked, eg, checked: `format_metrics=1`, unchecked `format_metrics=0`, then it's always supplied and never defaults to `format_metrics=bc`. Would that work or am I missing something?

sgiehl: But `bc` is kind of the default if no value is given. So always providing `0` will change the result of the exports (compared to the behavior before). See https://github.com/matomo-org/matomo/blob/3b00013819863472afe30cc04de49c4365c22d21/core/API/Request.php#L108-L110

diosmosis: It doesn't change the result of the API just the URL in the export popover, which I would say is more useful to the user than having half metrics formatted, and the other half not. But this isn't relevant to vue so I'll just change it back.

PR #18444, opened 2021-12-03
Ensure UTF8mb4 conversion command always enables tracking again

### LINKED ISSUE #18441: When convert to utf8mb4 command fails, enable tracking again
When running the `core:convert-to-utf8mb4` command, this command may fail eg because of a reason like below:

> Error: {"message":"SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs","file":"\/var\/www\/html\/libs\/Zend\/Db\/Adapter\/Pdo\/Abstract.php","line":280,"backtrace":" on \/var\/www\/html\/libs\/Zend\/Db\/Adapter\/Pdo\/Abstract.php(280)\n#0 \/var\/www\/html\/plugins\/CoreUpdater\/Commands\/ConvertToUtf8mb4.php(91): 

Afterwards, the tracking is disabled because the command sets `record_statistics=0` and only enables it if the command succeeds. If any query fails, we would also expect the setting to be enabled again.

It might otherwise take a user a few days that tracking may be disabled and then data is lost.

In some case we just got alerted instantly that tracking isn't working anymore and we thought there may be a DB issue when in reality tracking was simply disabled.

sgiehl:
### Description:

fixes #18441 

### Review

### DIFF
--- a/plugins/CoreUpdater/Commands/ConvertToUtf8mb4.php
+++ b/plugins/CoreUpdater/Commands/ConvertToUtf8mb4.php
@@ -87,7 +87,6 @@ class ConvertToUtf8mb4 extends ConsoleCommand
             $output->writeln("\n" . Piwik::translate('CoreUpdater_ConsoleStartingDbUpgrade'));
 
             try {
-                throw new \Exception();
                 foreach ($queries as $query) {
                     $output->write("\n" . 'Executing ' . $query . '... ');
                     Db::get()->exec($query);
@@ -101,8 +100,8 @@ class ConvertToUtf8mb4 extends ConsoleCommand
                 if (!$keepTracking) {
                     $output->writeln("\n" . Piwik::translate('Enabling Matomo Tracking'));
                     $config->Tracker['record_statistics'] = '1';
-                    $config->forceSave();
                 }
+                $config->forceSave();
             }
 
             $this->writeSuccessMessage($output, array('Conversion to utf8mb4 successful.'));
--- a/plugins/CoreUpdater/Commands/ConvertToUtf8mb4.php
+++ b/plugins/CoreUpdater/Commands/ConvertToUtf8mb4.php
@@ -86,22 +86,24 @@ class ConvertToUtf8mb4 extends ConsoleCommand
 
             $output->writeln("\n" . Piwik::translate('CoreUpdater_ConsoleStartingDbUpgrade'));
 
-            foreach ($queries as $query) {
-                $output->write("\n" . 'Executing ' . $query . '... ');
-                Db::get()->exec($query);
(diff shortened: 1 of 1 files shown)


plugins/CoreUpdater/Commands/ConvertToUtf8mb4.php
```diff
@@ -86,23 +86,25 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
             $output->writeln("\n" . Piwik::translate('CoreUpdater_ConsoleStartingDbUpgrade'));
 
-            foreach ($queries as $query) {
-                $output->write("\n" . 'Executing ' . $query . '... ');
-                Db::get()->exec($query);
-                $output->write(' done.');
+            try {
+                throw new \Exception();
```
justinvelluppillai: Testing line?

plugins/CoreUpdater/Commands/ConvertToUtf8mb4.php
```diff
@@ -86,23 +86,25 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                }
+
+                $output->writeln("\n" . 'Updating used database charset in config.ini.php.');
+                $config                      = Config::getInstance();
+                $config->database['charset'] = 'utf8mb4';
+            } finally {
+                if (!$keepTracking) {
+                    $output->writeln("\n" . Piwik::translate('Enabling Matomo Tracking'));
+                    $config->Tracker['record_statistics'] = '1';
+                    $config->forceSave();
```
justinvelluppillai: `forceSave()` should be outside the if statement to save the charset also?

justinvelluppillai (changes requested):
Made two comments needing fixing, seems fine otherwise.

plugins/CoreUpdater/Commands/ConvertToUtf8mb4.php
sgiehl: Ooops :grimacing:

plugins/CoreUpdater/Commands/ConvertToUtf8mb4.php
sgiehl: ah, you're right. changed that.

justinvelluppillai (approved):

sgiehl:
@justinvelluppillai I have pushed another small fix. In theory it could have happened that `$config` might not have been defined in the `finally`, if `$keepTracking` is `false`, and one of the db queries fails.

PR #18445, opened 2021-12-03
[Vue] migrate popover handler to popover service

diosmosis:
### Description:

**This PR depends on #18443.**

Changes:
* The current popover handler directive which if placed on a page handles the popover parameter has been replaced with a service class that does the same thing but doesn't need any HTML placed on the page.

image

### Review


diosmosis:
UI tests should be passing for this one

peterhashair:
maybe we need to resolve the conflicts, I am trying to fix those local, but it seems to return a blank page without any javascript errors on my tests.

plugins/CoreHome/vue/src/AjaxForm/AjaxForm.vue
```diff
@@ -0,0 +1,113 @@
...
+    return {
+      isSubmitting: false,
+      successfulPostResponse: null,
+      errorPostResponse: null,
+    };
+  },
+  emits: ['update:modelValue'],
+  mounted() {
+    // on submit call controller submit method
+    $(this.$refs.root).on('click', 'input[type=submit]', () => {
```
peterhashair: just wondering, will $emit function works here.

peterhashair (commented):
just one suggestion, if that works

plugins/CoreHome/vue/src/AjaxForm/AjaxForm.vue
diosmosis: I'm not sure what you mean, how would $emit be used here?

peterhashair: ah, actually i confused myself, I was thinking this 
```html
 <div ref="root" @click=‘this.submitForm()’>
```

peterhashair: ah, actually I confused myself, I see why this way.

PR #18453, opened 2021-12-06
Fixes white-label with auto password

peterhashair:
### Description:

Fixes white-label, when modal open password manager autofill the field in the back, update to read-only fields on modal opens. Revert to normal when modal closed. 
https://innocraft.atlassian.net/jira/software/projects/L3/boards/20?label=Core&selectedIssue=L3-161

### Review


plugins/CorePluginsAdmin/angularjs/plugin-settings/plugin-settings.controller.js
```diff
@@ -66,9 +66,15 @@
...
-                    }}).modal('open');
+                            $('.modal.open #currentUserPassword').focus();
+                            $('.modal.open #currentUserPassword').off('keypress').keypress(onEnter);
+                            $('.pluginSettings input').prop("readonly",true);
+                            $('.confirm-password-modal input').prop("readonly",false);
+                        },
+                        onCloseStart: function () {
+                            $('.pluginSettings input').prop('readonly', false);
+                        },
+                    }).modal('open');
```
sgiehl: I'm not sure if that will really solve the issue. I guess the `readonly` attributes are set to prevent password managers from filling in the field. On the one side, it might actually be too late to set that attribute when opening the modal. The HTML is already placed earlier and a password manager could already fill the field. On the other side I doubt that password managers will all respect the `readonly` attribute. There should also be a `autocomplete=off`, which is already ignore by certain password managers, so why should they respect other attributes?
I guess the cleanest solution would be to disconnect the password input field from angular and set the `passwordConfirmation` property to the input fields value once the modal is closed. That way the modal should always be shown, even if it is already filled.

peterhashair:
@sgiehl I did a screen capture on that one, it seems like autocomplete=off didn't prevent the autofill, but the read-only does, for Bitwarden and another third part password.  See the screen capture. 

https://user-images.githubusercontent.com/3783072/145738110-eede3d3f-f0d2-4779-a174-2f44d9ed9f0b.mp4

sgiehl:
Yes. It works for my password manager too. But I actually can only repeat my last comment. It works for the tested password managers. But as soon as any password manager is able to set a value for that field it will break again.
Btw. disconnecting the password input from angular would even solve another issue: When the modal is currently opened and you type something into the password field, but click `No`, the modal will be closed without clearing the password field. Due to this the next click on the save button, will simply use the password typed before, without opening the modal again.

peterhashair:
@sgiehl did a simple approach, hopefully that disconnect the password input

sgiehl:
@peterhashair That wasn't what I had in my mind, also it actually changes the behavior. Before the password confirm was only shown when saving some admin configuration. Now it is also shown for user configuration. What I had in mind, was removing the `ng-model` from the password field and instead fetch/set the value with javascript when it's needed.

peterhashair:
@sgiehl update its discount password field. I notice this is in the process of converting to VUE on this PR #18432, did a VUE version as well which is here #18515, I probably can drop this PR.

sgiehl:
Yep. closing this one in favor of the vue PR.

PR #18467, opened 2021-12-08
Dashboards per site

tomper00:
### Description:
This branch adds the feature to make Dashboards connected to specific idSites. 
This is important especially if we work with plugins like Custom reports where reports are bound to specific idSites so if we add a custom report to a Dashboard it will generate errors to the users on all other sites.

I added support for All sites (0) in the db and logic but there is not yes a GUI for users to manage this (we will always set. the siteId of the site you are currently viewing when you create a new dashboard.

**Note**: This still needs more testing I have created the DB updates but these are not tested yet for instance.

I would really need some help testing this.

### DIFF
--- a/plugins/Dashboard/API.php
+++ b/plugins/Dashboard/API.php
@@ -9,6 +9,7 @@ namespace Piwik\Plugins\Dashboard;
 
 use Piwik\API\Request;
 use Piwik\Piwik;
+use Piwik\Common;
 
 /**
  * This API is the <a href='http://matomo.org/docs/analytics-api/reference/' rel='noreferrer' target='_blank'>Dashboard API</a>: it gives information about dashboards.
@@ -36,15 +37,16 @@ class API extends \Piwik\Plugin\API
      *
      * @return array[]
      */
-    public function getDashboards($login = '', $returnDefaultIfEmpty = true)
+    public function getDashboards($login = '', $returnDefaultIfEmpty = true )
     {
         $login = $login ? $login : Piwik::getCurrentUserLogin();
+        $idSite = Common::getRequestVar('idSite');
 
         $dashboards = [];
 
         if (!Piwik::isUserIsAnonymous()) {
             Piwik::checkUserHasSuperUserAccessOrIsTheUser($login);
-            $dashboards = $this->getUserDashboards($login);
+            $dashboards = $this->getUserDashboards($login,$idSite);
         }
 
         if (empty($dashboards) && $returnDefaultIfEmpty) {
@@ -74,8 +76,9 @@ class API extends \Piwik\Plugin\API
         if ($addDefaultWidgets) {
             $layout = $this->dashboard->getDefaultLayout();
         }
+        $idSite = Common::getRequestVar('idSite');
 
-        return $this->model->createNewDashboardForUser($login, $dashboardName, $layout);
+        return $this->model->createNewDashboardForUser($login, $dashboardName, $layout,$idSite);
     }
 
     /**
--- a/plugins/Dashboard/API.php
+++ b/plugins/Dashboard/API.php
@@ -37,10 +37,9 @@ class API extends \Piwik\Plugin\API
      *
      * @return array[]
      */
-    public function getDashboards($login = '', $returnDefaultIfEmpty = true )
+    public function getDashboards($login = '', $returnDefaultIfEmpty = true, $idSite = 0 )
     {
         $login = $login ? $login : Piwik::getCurrentUserLogin();
-        $idSite = Common::getRequestVar('idSite');
 
         $dashboards = [];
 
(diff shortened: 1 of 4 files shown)


plugins/Dashboard/Model.php
```diff
@@ -133,10 +133,11 @@ public function updateLayoutForUser($login, $idDashboard, $layout)
...
     {
         $dashboard = "login VARCHAR( 100 ) NOT NULL ,
-					  iddashboard INT NOT NULL ,
-					  name VARCHAR( 100 ) NULL DEFAULT NULL ,
-					  layout TEXT NOT NULL,
-					  PRIMARY KEY ( login , iddashboard )";
+                      iddashboard INT NOT NULL ,
+                      name VARCHAR( 100 ) NULL DEFAULT NULL ,
+                      layout TEXT NOT NULL,
+                      idsite bigint(20) NOT NULL ,
```
sgiehl: `idsite` is defined as `int(11)` in most other tables, so would be enough to use that.

plugins/Dashboard/API.php
```diff
@@ -36,15 +37,16 @@ public function __construct(Dashboard $dashboard, Model $model)
      *
      * @return array[]
      */
-    public function getDashboards($login = '', $returnDefaultIfEmpty = true)
+    public function getDashboards($login = '', $returnDefaultIfEmpty = true )
     {
         $login = $login ? $login : Piwik::getCurrentUserLogin();
+        $idSite = Common::getRequestVar('idSite');
```
sgiehl: This should be a method parameter instead. Maybe we could use `0` as default (which would be `all websites`)

plugins/Dashboard/API.php
```diff
@@ -74,8 +76,9 @@ public function createNewDashboardForUser($login, $dashboardName = '', $addDefau
         if ($addDefaultWidgets) {
             $layout = $this->dashboard->getDefaultLayout();
         }
+        $idSite = Common::getRequestVar('idSite');
```
sgiehl: Same here. `Common::getRequestVar` should be avoided in API, as request parameters would automatically be mapped to the method parameters if possible.

plugins/Dashboard/Updates/4.7.0.php
```diff
@@ -0,0 +1,61 @@
...
+
+use Piwik\Updater;
+use Piwik\Updates as PiwikUpdates;
+use Piwik\Updater\Migration;
+use Piwik\Updater\Migration\Factory as MigrationFactory;
+
+/**
+ * Update for version 1.0.8-beta.
+ */
+class Updates_4_7_0 extends PiwikUpdates
```
sgiehl, line 19: This should have the name of the next possible beta release. Currently that would be `4.7.0-b2` and in order to have the script applied when someone fetches your changes you might also need to change the version in `core\Version.php`

sgiehl (changes requested):
Hi @tomper00. Thanks for creating the PR. I had a very rough look through the code and left some comments. Might have some more time tomorrow to look into that in order to test that a bit maybe.

tsteur:
@tomper00 great. That will be very useful indeed. From a product perspective, when creating or copying a dashboard, it would be great to be able to select whether a dashboard should be available for all sites, or only the given site (later we could maybe allow to select it to share it with multiple sites). From what I see this PR would otherwise change all the dashboards to be per site? There is for some dashboards/use cases also value in having dashboards across sites.

tomper00:
Ok I started to change the code according to you suggestions @sgiehl, still some testing to be done locally though - Thanks for this, very appreciated! 

@tsteur Yes there is a bit of thinking needed to be done around the full functionality for sure so perhaps this should be part of a bigger feature update rather than just a small improvement.

Here are some examples of things we discussed recently that would make dashboards more powerful:

- Make it possible to add custom content to a dashboard (Texts and perhaps images to better describe what you see).
- Add an admin view for all Dashboards (so that its easier to help users and see all their dashboards)
- Make it possible to have Shared dashboards with users instead of a lot of copies to make sure changes are replicated properly

sgiehl:
@tomper00 Thanks for the additional suggestions:

> Make it possible to add custom content to a dashboard (Texts and perhaps images to better describe what you see).

This should actually be something that could easily be implemented in a third party plugin. I don't think that is something we plan to integrate into core currently.

> Add an admin view for all Dashboards (so that its easier to help users and see all their dashboards)

This could also be done in an custom plugin, but might be something that could be interesting for core as well.

> Make it possible to have Shared dashboards with users instead of a lot of copies to make sure changes are replicated properly

Nice idea. Guess this won't be easily implemented in a custom plugin, unless we extend the Dashboard API a lot. 
Btw. how shall shared dashboards behave?
  * Can it only be edited by an owner or can everyone it was shared with edit the dashboard (or should that be configurable)
  * Can someone remove a share of a dashboard again (either the owner or the user it was shared with)
  * I guess it would somehow need to be visible, that a dashboard is currently shared, and maybe also whom it was shared with.

PR #18218, opened 2021-10-25
Add guide links on tracking code page and no data email

bx80:
### Description:

Fixes #15429 
Added setup guide links for Wordpress, Squarespace, Wix, SharePoint,  Joomla, Shopify and Google Tag Manager to the 'No data' email and to the tracking code page.

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #18469, opened 2021-12-09
Updated shell_exec recommended message

### LINKED ISSUE #14782: [Docs] Explain shell_exec requirement
I am a bit confused by the usage of `shell_exec` in Matomo and can't find any docs for it.

- https://matomo.org/docs/requirements/ doesn't mention that it is required
- In the few places I checked there is often a `if (!function_exists('shell_exec')) {` indicating that is only recommended, but not required
- There are no docs explaining what Matomo needs shell_exec for (or even what it is and how to enable it)
   - Therefore there are many recurring questions about it on the forum like https://forum.matomo.org/t/shell-exec-sie-mussen-diese-eingebaute-funktion-aktivieren/23079/2
- The system check is called `RecommendedFunctionsCheck` indicating that it isn't required
- but the warning it gives is `You need to enable this built-in function.` and `Sie müssen diese eingebaute Funktion aktivieren` indicating that it is required.

Can someone who knows the background help clarifying all docs?

justinvelluppillai:
### Description:

Resolves #14782 by changing the shell_exec missing message from "You need to" to "We recommend you", also adding a link to the reasons for shell_exec being used in Matomo.

### Review

### DIFF
--- a/plugins/Diagnostics/Diagnostic/RecommendedFunctionsCheck.php
+++ b/plugins/Diagnostics/Diagnostic/RecommendedFunctionsCheck.php
@@ -73,7 +73,7 @@ class RecommendedFunctionsCheck implements Diagnostic
         );
 
         $translation_params = array(
-            'shell_exec'     => ["<a href='https://matomo.org/faq/troubleshooting/how-to-make-the-diagnostic-managing-processes-via-cli-to-display-ok/'>", "</a>"]
+            'shell_exec'     => ["<a href='https://matomo.org/faq/troubleshooting/how-to-make-the-diagnostic-managing-processes-via-cli-to-display-ok/' rel='noopener' target='_blank'>", "</a>"]
         );
 
         return $this->translator->translate($messages[$function], $translation_params[$function] ?? []);
--- a/plugins/Diagnostics/Diagnostic/RecommendedFunctionsCheck.php
+++ b/plugins/Diagnostics/Diagnostic/RecommendedFunctionsCheck.php
@@ -72,6 +72,10 @@ class RecommendedFunctionsCheck implements Diagnostic
             'gzopen'         => 'Installation_SystemCheckZlibHelp',
         );
 
-        return $this->translator->translate($messages[$function]);
+        $translation_params = array(
+            'shell_exec'     => ["<a href='https://matomo.org/faq/troubleshooting/how-to-make-the-diagnostic-managing-processes-via-cli-to-display-ok/'>", "</a>"]
+        );
+
+        return $this->translator->translate($messages[$function], $translation_params[$function] ?? []);
     }
 }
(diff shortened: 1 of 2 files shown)


plugins/Diagnostics/Diagnostic/RecommendedFunctionsCheck.php
```diff
@@ -72,6 +72,10 @@ private function getHelpMessage($function)
             'gzopen'         => 'Installation_SystemCheckZlibHelp',
         );
 
-        return $this->translator->translate($messages[$function]);
+        $translation_params = array(
+            'shell_exec'     => ["<a href='https://matomo.org/faq/troubleshooting/how-to-make-the-diagnostic-managing-processes-via-cli-to-display-ok/'>", "</a>"]
```
sgiehl: I guess this one should have a `target='_blank'`, otherwise clicking it will leave Matomo

Findus23: And noopener, right?

plugins/Installation/lang/en.json
```diff
@@ -88,7 +88,7 @@
         "SystemCheckFileIntegrity": "File integrity",
         "SystemCheckFilterHelp": "You need to configure and rebuild PHP with \"filter\" support enabled (don't use --disable-filter).",
         "SystemCheckFunctions": "Required functions",
-        "SystemCheckFunctionHelp": "You need to enable this built-in function.",
```
Findus23, line 91: Wouldn't this also affect the other functions that might still be required?

plugins/Diagnostics/Diagnostic/RecommendedFunctionsCheck.php
justinvelluppillai: Thanks 👍

justinvelluppillai: Understanding from https://github.com/matomo-org/matomo/pull/18218 we don't need noreferrer anymore.

Findus23: Ah, indeed (at least as long as people use modern browsers)

justinvelluppillai: I saw your other comment from email @Findus23 but I guess maybe you removed it later? The translation text isn't currently used anywhere other than for `shell_exec` from my usages search but I could leave that translation there and create another if you're aware of any plugins or other places that may use it. I think possibly better to not leave unused translations in there but I'm open to other ideas.

Findus23: I noticed too afterwards that it is only used for shell_exec, that's why I deleted it. Maybe the translation key could be changed to reflect this (as we are already invalidating it). But leaving everything as-is also seems fine by me.

justinvelluppillai: Good idea.

sgiehl (approved):

PR #18474, opened 2021-12-09
[Vue] fix build + race condition fixes

diosmosis:
### Description:

Changes:
* update screenshot failure due to submodule that hasn't been updated
* better race condition fix for creating dashboard (rebuild menu and reload dashboards before loading new page, or page load will abort ongoing requests)
* race condition fix for removing dashboard (wait for menu rebuild before loading new page or page load will abort ongoing requests)

### Review


sgiehl:
@diosmosis the regression when removing a dashboard seems to be fixed. But I found something else while testing it.
If you add a new dashboard, and the amount of dashboard is or gets bigger than 5, the dropdown menu is displayed. When the new dashboard is created, it is loaded correctly, but the value in the drop down isn't updated to it. Opening the drow down shows it as selected though.

sgiehl:
And another thing that might be nice to fix at some point. When switching the dashboard the hash is updated to always include `&segment=&comparePeriods=&compareDates=&compareSegments=`. Would it be easily possible to remove empty parameters from the hash?

sgiehl:
And another one 🙈 
When I initially choose a goal in the dropdown menu, the value is always updated to the first goals name, even though another has been clicked and is loaded. (It works correctly when switching to another goal afterwards.)

sgiehl:
Seems the reported issues are now fixed. But I actually found something else. It seems the drop downs in the menu are somehow connected. E.g. when I load the second dashboard and open the goals submenu, the second goal is preselected. Changing the dashboard also preselects another goal....

diosmosis:
@sgiehl pushed another fix

plugins/Dashboard/javascripts/dashboard.js
```diff
@@ -29,11 +29,12 @@ function createDashboard() {
...
-
-                        $('#dashboardWidgetsArea').dashboard('loadDashboard', id);
-                        $('#dashboardWidgetsArea').dashboard('rebuildMenu');
-                    });
+                  Promise.all([
+                    dashboardsModel.reloadAllDashboards(),
+                    reportingMenuModel.reloadMenuItems(),
+                  ]).then(function () {
+                    $('#dashboardWidgetsArea').dashboard('loadDashboard', id);
+                  });
```
sgiehl: This actually breaks the dashboad menu in widgetize mode. `$('#dashboardWidgetsArea').dashboard('rebuildMenu');` still needs to be triggered as this would also update the widgetize menu. See 
https://github.com/matomo-org/matomo/blob/934eb450419edaf4acf2fc21e3b52723f11c40f1/plugins/Dashboard/javascripts/dashboardObject.js#L535-L597

plugins/Dashboard/javascripts/dashboardObject.js
```diff
@@ -204,8 +204,9 @@
             }, 'get');
             ajaxRequest.setCallback(
                 function () {
-                    methods.loadDashboard.apply(this, [1]);
-                    rebuildMenu();
+                    Promise.resolve(rebuildMenu()).then(function () {
```
sgiehl, line 207: This doesn't work either. When a dashboard is removed, the list in widgetized dashboard isn't refreshed. I guess the problem is, that `rebuildMenu`, doesn't return a promise in widgetized mode. simply returning the `ajaxRequest.send()` at the methods end might solve this I guess.

diosmosis:
@sgiehl pushed another update

sgiehl (approved):
now everything seems to work as expected 👍

PR #18477, opened 2021-12-09
[Vue] two bug fixes: use correct base for site change + correct URL for launching new site

diosmosis:
### Description:

Fixes:

* First: when changing the site, we can't use hash values in the new URL since some hash parameters will break Matomo if present in the main URL. To reproduce, just change a site on a report page, the hash will merge w/ the rest of the URL and the page content will not load.

* Second: when launching a site in a new tab, the hash needs to contain the period/date even if it's not in the hash, since some angularjs code will only forward what is in the hash in API requests. TO reproduce, create a site w/ no visits, go to a site w/ visits, then from the site selector cmd+click the site with no visits. It will load in a new tab, but the AJAX request for siteWithoutData content will fail because no date/period was specified in it. (The root problem is in the widgetloader, not sure if this is fixed automatically in the Vue migration PR).

~Given how very specific/fragile the URL handling is in different places I wonder if there's a specific API that will avoid these sorts of problems in the future...~

Fixed the second one also by just making sure period/date/segment are always present in the hash params.

### Review


sgiehl (changes requested):
This doesn't seem to work as expected. Now when I'm a report page and click another site in the site selector, the `idSite` is updated in the query string but not in the hash, causing ,the site to be selected in the site selector, but the loaded content is from the previous site.
And when I'm on a site without data (url without hash) and open another site in a new tab (strg + click), the new url contains a hash with the `idSite` only, and the request fails. I guess the reason for that is here:
https://github.com/matomo-org/matomo/blob/d25cc336745057f889afa0f15a0f2e3c5b2a93fb/plugins/CoreHome/vue/src/SiteSelector/SiteSelector.vue#L359-L373

diosmosis:
@sgiehl updated

> And when I'm on a site without data (url without hash) and open another site in a new tab (strg + click), the new url contains a hash with the idSite only, and the request fails. I guess the reason for that is here:

can't reproduce this, the URL I get if I ctrl + click a site entity looks like `http://localhost/index.php?module=CoreHome&action=index&idSite=3&period=day&date=yesterday#?period=day&date=yesterday&idSite=3&category=Dashboard_Dashboard&subcategory=1` which works. What kind of URL are you seeing? What URL did you click from?

sgiehl:
@diosmosis I'm not able to reproduce that either with the latest changes. 
Seems some JS tests are currently failing. Would be good to merge, once they are fixed.

PR #18482, opened 2021-12-10
Added code to filter metadata response before xml comparison

AltamashShaikh:
### Description:

Added code to filter metadata response before XML comparison.

The previous code changes added to filter metadata response was breaking other API tests like `API.get` which internally tries to call `API.getSegmentsMetadata`. The result of `API.getSegmentsMetadata` gets filtered out for `API.get` and system testcases starts to fail.

To solve the issue, will be filtering the XML response just before we compare the processed and expected output.

### Review


AltamashShaikh:
@tsteur I have 1 better way which can work by tweaking the existing logic
I can show it to you tomorrow and we can decide which 1 should we choose

tsteur:
OK. Generally 1 way is enough though since it's quite small logic and not too complex it be probably fine as it is and don't need to optimise further. Unless there's some other benefit?

AltamashShaikh:
Created #18510
As it is a much simpler approach closing this.

PR #18493, opened 2021-12-12
convert sparkline api to one request - fixes archive

### LINKED ISSUE #18088: Multiple requests may archive the same range request at the same time

peterhashair:
### Description:

Description:
Fixes: #18088.

Follow the fix on UI tests

### Review


peterhashair:
this is the follow-up on the UI failed fixes. Should pass the test now.

bx80:
@peterhashair The only remaining failing UI screenshot `UIIntegrationTest_dashboard2.png` is expecting the goals overview widget goal titles to be wrapped in single quotes, where as the processed screenshot has them without.
image
It seems like it might be related to this change?

bx80 (approved):
UI tests are now all passing. 

I noticed that because the `GetMetrics` report is a now a descendant of the `Get` report that the view is configured slightly differently with overview Goal Names shown as `Goal 'Something'` instead of `Goal Something`. This would seem to be more consistent, what do you think @sgiehl ?

plugins/Goals/Pages.php
```diff
@@ -54,9 +54,21 @@ public function createGoalsOverviewPage($goals)
...
+        $config = $this->factory->createCustomWidget('getSparklines');
+        $config->setSubcategoryId($subcategory);
+        $config->setName('');
+        $config->setOrder(25);
+        $config->setIsNotWidgetizable();
+        $widgets[] = $config;
+
+        $container = $this->createWidgetizableWidgetContainer('GoalsOverview', $subcategory, $widgets);
+
         foreach ($goals as $goal) {
```
sgiehl: I guess this `foreach` is no obsolete. As `$widgets` isn't used afterwards anymore, there is no sense it filling it more.

plugins/Goals/Controller.php
```diff
@@ -219,6 +223,37 @@ public function getEvolutionGraph(array $columns = array(), $idGoal = false, arr
...
+    {
+        $module = "Goals";
+        $action = "getMetrics";
+        $content = "";
+
+        $idSite = Common::getRequestVar('idSite', null, 'int');
+        $goals = Request::processRequest('Goals.getGoals', ['idSite' => $idSite, 'filter_limit' => '-1'], []);
+
+        $apiProxy = Proxy::getInstance();
+        $queryString = $_SERVER['QUERY_STRING'];
```
sgiehl: seems unused
```suggestion
```

plugins/Goals/Controller.php
```diff
@@ -219,6 +223,37 @@ public function getEvolutionGraph(array $columns = array(), $idGoal = false, arr
...
     }
 
+    public function getSparklines()
+    {
+        $module = "Goals";
+        $action = "getMetrics";
+        $content = "";
+
+        $idSite = Common::getRequestVar('idSite', null, 'int');
+        $goals = Request::processRequest('Goals.getGoals', ['idSite' => $idSite, 'filter_limit' => '-1'], []);
```
sgiehl:
```suggestion
        $goals = Request::processRequest('Goals.getGoals', ['idSite' => $this->idSite, 'filter_limit' => '-1'], []);
```

plugins/Goals/Controller.php
```diff
@@ -219,6 +223,37 @@ public function getEvolutionGraph(array $columns = array(), $idGoal = false, arr
...
+
+        $apiProxy = Proxy::getInstance();
+        $queryString = $_SERVER['QUERY_STRING'];
+        if (!$apiProxy->isExistingApiAction($module, $action)) {
+            throw new \Exception("Invalid action name '$action' for '$module' plugin.");
+        }
+
+        $apiAction = $apiProxy->buildApiActionName($module, $action);
```
sgiehl: Can't see much value in checking and building the API action name using those methods. Guess you could also simply use `Goals.getMetrics`. If that one gets deleted somewhen, this method will throw an exception nevertheless, as the report can't be found at another point while processing the view.

sgiehl (changes requested):
Generally seems to work now. Left a couple of comments for code improvements

plugins/Goals/Controller.php
```diff
@@ -219,6 +223,25 @@ public function getEvolutionGraph(array $columns = array(), $idGoal = false, arr
...
+
+        foreach ($goals as $goal) {
+            //this is used to hack query params
+            $_GET['idGoal'] = $goal['idgoal'];
+            $_GET['allow_multiple'] = (int)$goal['allow_multiple'];
+            $_GET['only_summary'] = 1;
+            //load Visualisations Sparkline
+            $view = ViewDataTableFactory::build(Sparklines::ID, 'Goals.getMetrics', 'Goals.' . __METHOD__, true);
+            $view->config->show_title = true;
+            $content .= $view->render();
```
sgiehl: Here you could use `Context::executeWithQueryParameters`, which will ensure the request parameters are reset afterwards.
```suggestion
            $params = [
                'idGoal' => $goal['idgoal'],
                'allow_multiple' => (int) $goal['allow_multiple'],
                'only_summary' => 1,
            ];

            \Piwik\Context::executeWithQueryParameters($params, function() use (&$content) {
                //load Visualisations Sparkline
                $view = ViewDataTableFactory::build(Sparklines::ID, 'Goals.getMetrics', 'Goals.' . __METHOD__, true);
                $view->config->show_title = true;
                $content .= $view->render();
            });
```

sgiehl (changes requested):
Had another idea for further code improvement. Otherwise this should be good to merge, if tests are passing.

PR #18503, opened 2021-12-14
Ensure updates are always done with super user permission

sgiehl:
### Description:

Our process for performing updates is currently designed in a way that allows anyone (even without permission) to trigger the database updates (once the files were updated).
This can currently lead to errors or differing results depending on the current users permissions.
I saw this basically for this update script where I saw differing results:
https://github.com/matomo-org/matomo/blob/4fe950ae94b0fe3bd13f928544ff876275fe6733/core/Updates/4.6.0-b4.php#L54-L58
Initializing a segment using `new Segment` tries to perform an API request to `API.getSegmentsMetadata` in the background. Based on the current user this might return either the segments (user with view permission) or an error (user without permission). This causes the resulting list of generated queries to be different depending on the user.
In that case this might not be that problematic, but there could be cases where it could even break something.
Therefor it's imho required to perform the migration scripts as superuser always.

### Review


tsteur (changes requested):
@sgiehl is it possible to only execute the needed methods as superuser? Eg 

*   $this->doExecuteUpdates($viewDone, $updater, $componentsWithUpdateFile);
* $componentsWithUpdateFile = $updater->getComponentUpdates();

Just to avoid potential security issues etc. Eg some fatal error and then suddenly it keeps super user or maybe some view would expose information a regular user shouldn't see or so?

sgiehl:
@tsteur moved setting the super user permission to the parts in Updater class, where it's needed. Guess that makes more sense, than doing it in a controller

core/Updater.php
```diff
@@ -233,37 +233,39 @@ public function hasMajorDbUpdate()
...
 
-                /** @var Updates $update */
-                $update = StaticContainer::getContainer()->make($className);
-                $migrationsForComponent = $update->getMigrations($this);
-                foreach ($migrationsForComponent as $index => $migration) {
-                    $migration = $this->keepBcForOldMigrationQueryFormat($index, $migration);
-                    $queries[] = $migration;
+                    /** @var Updates $update */
+                    $update                 = StaticContainer::getContainer()->make($className);
+                    $migrationsForComponent = $update->getMigrations($this);
```
tsteur: @sgiehl here we could still maybe narrow it down to only the `$update->getMigrations($this);` call maybe?

sgiehl: done.

tsteur (approved):

PR #18507, opened 2021-12-15
[Vue] fix infinite loop when using custom field angularjs template

diosmosis:
### Description:

In the FieldAngularJsTemplate.vue component we clone the form field value before it goes to angularjs and vice versa in order to make sure mutations in angularjs, where the data is meant to be mutable, doesn't create unexpected changes in the Vue layer, where the data is not meant to be modified by the form field components. Currently this causes an infinite loop in some cases as we check instance equality, and the instance changes when we clone the value. Fixed by doing a deep equal before propagating changes from either side.

### Review


sgiehl (approved):
works 👍

PR #18525, opened 2021-12-20
Installation pages send API request to API.getPagesComparisonsDisabledFor

### LINKED ISSUE #18402: Installation pages might send API request to API.getPagesComparisonsDisabledFor
When opening Matomo in an uninstalled state, Matomo will display the installation.
At this state the URL might not contain a `module`. Thus the check doesn't work here:

https://github.com/matomo-org/matomo/blob/24c07cd3967b41952da9d015d24705b2a0891a0d/plugins/CoreHome/vue/src/Comparisons/Comparisons.store.ts#L320-L325

This results in an failing API request.

peterhashair:
### Description:
Fixes:#18402
Installation pages send API request to API.getPagesComparisonsDisabledFor

### Review


plugins/CoreHome/vue/src/Comparisons/Comparisons.store.ts
```diff
@@ -292,6 +292,11 @@ export default class ComparisonsStore {
 
   private loadComparisonsDisabledFor() {
     const matomoModule: string = MatomoUrl.parsed.value.module as string;
+
+    if (typeof matomoModule === 'undefined') {
```
sgiehl: In theory Matomo can also be opened without a module given in the URL when it's installed. In that case it will be handled as `CoreHome`. e.g. a URL like this should work: `index.php?action=index&idSite=1&period=day&date=yesterday#?period=day&date=yesterday&category=Dashboard_Dashboard&subcategory=1`

Maybe it's better to check if the current page is the installation page by checking the html element. e.g.
https://github.com/matomo-org/matomo/blob/680ac30b2078ce7f0fb587c145161dab0bce2c24/plugins/Installation/templates/layout.twig#L15

sgiehl (changes requested):

plugins/CoreHome/vue/src/Comparisons/Comparisons.store.ts
```diff
@@ -293,10 +293,13 @@ export default class ComparisonsStore {
   private loadComparisonsDisabledFor() {
     const matomoModule: string = MatomoUrl.parsed.value.module as string;
 
-    if (typeof matomoModule === 'undefined') {
+    // check if body id #installation exist
+    const install = document.getElementById('installation');
```
sgiehl: Guess this should be good for now, but it actually might have the risk, that this regresses if any element has this id, maybe from a custom plugin. We could harden the check for `body#installation`, as the body element shouldn't have that id outside installation.

plugins/CoreHome/vue/src/Comparisons/Comparisons.store.ts
```diff
@@ -292,6 +292,14 @@ export default class ComparisonsStore {
 
   private loadComparisonsDisabledFor() {
     const matomoModule: string = MatomoUrl.parsed.value.module as string;
+
+    // check if body id #installation exist
+    const install = document.getElementById('body#installation');
```
sgiehl: `getElementById` doesn't work with such an selector. 
```suggestion
    const install = document.querySelector('body#installation');
```

sgiehl (approved):

PR #18526, opened 2021-12-20
update Performance Report Average Load Time query

### LINKED ISSUE #17012: Improve average page load time query accuracy (reporting too low in some cases)
The metric **Page load time** is defined as `How long it takes to load the whole page, from requesting a page until it is fully rendered within the browser. This is the sum of all previous metrics.`

When looking at some of the websites reports, it seems this metric is (sometimes?) calculated wrongly.

For example [here](https://demo.matomo.org/index.php?module=CoreHome&action=index&date=today&period=week&idSite=26#?idSite=26&period=week&date=2020-12-15&segment=&category=General_Actions&subcategory=PagePerformance_Performance) seeing this report which says ` 0.2s Avg. page load time` but we'd expect the value to be approx 4s or 5s instead.

Screenshot from 2020-12-24 09-46-35

the "Avg page load time" are shown correctly in the report table below though:

Screenshot from 2020-12-24 09-51-16

so maybe the issue is only in this sparkline metric and how it's calculated?

The strange thing is that it sometimes seems to work correctly for example in the same instance for another website [here.](https://demo.matomo.org/index.php?module=CoreHome&action=index&date=today&period=week&idSite=36&updated=1#?idSite=36&period=week&date=2020-12-15&category=General_Actions&subcategory=PagePerformance_Performance) 

(but the problem is  reproduced in other instances for example [here](https://web.innocraft.cloud/index.php?module=CoreHome&action=index&idSite=12&period=day&date=yesterday#?idSite=12&period=day&date=yesterday&segment=&category=General_Actions&subcategory=PagePerformance_Performance)

peterhashair:
### Description:

Fixes: #17012. 
It seems like one of our queries is wrong, update it should work now. But I guess it won't change the previously achieved data.

### Review


peterhashair:
It seems like those 2 queries produce 2 different results, I think the second one is correct, which will load the correct average load time, I guess if column is null, it reset sum back to 0?
### Current Query
```mysql

select sum(time_dom_completion+time_dom_processing+time_network+time_on_load+time_server+time_transfer) as total from log_link_visit_action where server_time like '2021-11-08%';
```

### Convert to Query
```mysql
select sum(+
IFNULL(`time_dom_completion`, 0) 
+IFNULL(`time_dom_processing`,0)+IFNULL(`time_network`,0)+IFNULL(`time_on_load`,0)+IFNULL(`time_server`,0)+IFNULL(`time_transfer`,0))
as total from log_link_visit_action where server_time like '2021-11-08%';

File: plugins/PagePerformance/Archiver.php [hunk 1]
```diff
@@ -54,7 +54,7 @@ public function aggregateDayReport()
             $column = $dimension->getColumnName();
             $selects[] = "sum($table.$column) as {$column}_total";
             $selects[] = "sum(if($table.$column is null, 0, 1)) as {$column}_hits";
-            $totalColumns[] = "$table.$column";
+            $totalColumns[] = "IFNULL($table.$column,0)";
```
sgiehl, line 58: Good catch. Checked that locally, and that seems to work 👌

tests/PHPUnit/System/expected/test_ImportLogs__PagePerformance.get_month.xml
```diff
@@ -6,5 +6,5 @@
 	<avg_time_dom_processing>0</avg_time_dom_processing>
 	<avg_time_dom_completion>0</avg_time_dom_completion>
 	<avg_time_on_load>0</avg_time_on_load>
-	<avg_page_load_time>0</avg_page_load_time>
+	<avg_page_load_time>1.67</avg_page_load_time>
```
sgiehl: This change actually lets me wonder if our calculation for the avg page load time is correct. Why should the page load time be a lot higher then the avg server time in this case?

Actually I think we might maybe also change the calculation of the `page_load_hits`
```
$selects[] = sprintf('(SUM(%s)/%s) as page_load_hits', implode(' + ', $hitsColumns), count($hitsColumns));
```
That actually means that a visit only fully counts if all fields are filled. This works if normally all fields have a value I guess, but as here in the test only one column is filled, causing the avg page load time to be much higher than it should be. I guess we should count a hit as soon as one field is given instead.

peterhashair: @sgiehl yes, I was wondering the same, not sure that will work, add a `having`  condition. it actually had PDF test, failed `test_TwoVisitors_twoWebsites_differentDays_schedrep_in_pdf_tables_only__ScheduledReports.generateReport_month.original.pdf` not sure that's right.

sgiehl: Thinking more on that I was actually wondering if we can't simply sum all other average values to get the average page load time. Guess it shouldn't be less accurate then now, but a lot easier to calculate.
The only problem I can see is that it might be inaccurate when the tracking is a bit "abnormal". E.g. you have only one visit where the network time is tracked, but hundreds of visits where the transfer time is tracked. The average network time would be calculated based on that one visit only, while the transfer time out of hundreds. Summing up their averages might then be not accurate. But that might an edge case we can ignore maybe.

tests/PHPUnit/System/expected/test_ImportLogs__PagePerformance.get_month.xml
```diff
@@ -6,5 +6,5 @@
 	<avg_time_dom_processing>0</avg_time_dom_processing>
 	<avg_time_dom_completion>0</avg_time_dom_completion>
 	<avg_time_on_load>0</avg_time_on_load>
-	<avg_page_load_time>0</avg_page_load_time>
+	<avg_page_load_time>0.13</avg_page_load_time>
```
sgiehl: Shouldn't that be the same as `avg_time_server` as all other values are `0`? Maybe the `HAVING` doesn't work as expected though?

plugins/PagePerformance/Archiver.php
```diff
@@ -54,16 +55,19 @@ public function aggregateDayReport()
...
             $hitsColumns[] = "if($table.$column is null, 0, 1)";
+            $allColumns[]  = $table.$column;
```
sgiehl: That can't work, as the `.` is used for concatenation, but is required in sql.
```suggestion
            $allColumns[]  = "$table.$column";
```

sgiehl (approved):

PR #18540, opened 2021-12-24
Fix error where no data is shown on campaign with segments and periods

peterhashair:
### Description:
L3: https://innocraft.atlassian.net/browse/L3-199
fix error no data notification.  
Show on the `campaign` with data show. 
To reproduce it, need to select a period, and require a segment. The request is that one, in my case.  
```
segment=visitorType==returning&date=2021-09-01,2021-09-15&module=MarketingCampaignsReporting&format=html&action=getGroup&period=range&idSite=1&widget=&showtitle=1&random=2789 
```

The code runs into here https://github.com/matomo-org/matomo/blob/1aec00ff0affa727ef134e230b465091b97e6863/core/DataAccess/ArchiveSelector.php#L102-L107

### Review

### DIFF
--- a/plugins/SegmentEditor/SegmentEditor.php
+++ b/plugins/SegmentEditor/SegmentEditor.php
@@ -252,7 +252,7 @@ class SegmentEditor extends \Piwik\Plugin
         // check if segment archive does not exist
         $processorParams = new \Piwik\ArchiveProcessor\Parameters(new Site($idSite), $period, $segment);
         $archiveIdAndStats = ArchiveSelector::getArchiveIdAndVisits($processorParams, null);
-        if (!empty($archiveIdAndStats[0])) {
+        if (!empty($archiveIdAndStats[0]) || !empty($archiveIdAndStats[1])) {
             return null;
         }
 


bx80 (commented):
I've been able to recreate the original issue and this fix prevents the 'no data' message from showing. 

I might be misunderstanding something here, but If the archive selector is identifying the archive as having an invalid done flag value then should we be accepting a result without an `idarchive` value? 

Would it be better to determine why the archive done flag is incorrect and either fix that if the archive is truly valid for use or if the archive isn't really valid then keep the 'no data' message and prevent the data being shown?

peterhashair:
@bx80 if I am right, I remember the problem was, there are multiple javascript ajax calling this function https://github.com/matomo-org/matomo/blob/9ed7e1815e263aa8a1a52332311ba0fd3489bb01/plugins/SegmentEditor/SegmentEditor.php#L186
 If one of the ajax returns an empty result,  the `no data` message will show up. In my case is the `Campaign Groups`, which is `no data` is expected, but the segment is processed as an archive.

bx80:
@peterhashair That makes sense when considering the multiple ajax calls. I can't really think of any scenario where there would be a reason to show the `no data (check back later)` message when we know that an archive has already been generated for the segment. Just wanted to be sure we're not preventing the message being shown when it is needed. :slightly_smiling_face:

bx80 (approved):

PR #18446, opened 2021-12-03
[Vue] migrate smaller core home common directives and deprecate others

diosmosis:
### Description:

**This PR is based off of #18445.**

Changes:
* Migrate dropdown-button.
* Migrate piwik-show-sensitive-data. (Unused, mostly for BC.)
* Mark other common directives as deprecated, they will be removed rather than migrated.

image

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #18545, opened 2021-12-27
[Vue] enable full typescript compilation + fix typing errors in existing Vue code

diosmosis:
### Description:

**This PR is based off of #18446.**

By default Vue sets the 'transpile only' option in the TypeScript compiler which only does a cursory check of typing related errors. This means the current code has a lot of typing errors (and also some bugs) that were never detected. It also means we can't output typing information for plugin UMDs, so type information for CoreHome would be lost when using CoreHome exports in another plugin. This PR fixes that. Note: Vue doesn't actually support doing a full compile natively (there's at least one open issue), so we workaround it.

(This is also why typescript errors in unit tests do not appear in vue:build.)

Changes:
* In vue.config.js if in production mode, turn on full compilation. In development mode this is not done so development will still be fast (using full type checking takes a bit longer since tsc will do more work).
* Output typings for plugin UMDs to the `/path/to/matomo/@types` directory. These will be picked up by a change to tsconfig.json and typescript will be able to apply typing information to JS imported from other plugins.
* Fixes for all typing errors. Some are superficial, but there are some actual bug fixes in here as well.
  Note: one "fix" is a workaround for vue's DeepReadonly type not being able to handle recursive interfaces (eg, our Widget data structure). Fixed by separating the part that recurses and casting to a sub-interface when needed.

### Review


plugins/CoreHome/vue/src/ContentBlock/ContentBlock.vue
```diff
@@ -78,7 +79,7 @@ export default defineComponent({
       }
     }, 0);
 
-    if (this.actualFeature && (this.actualFeature === true || this.actualFeature === 'true')) {
+    if (this.actualFeature && this.actualFeature === 'true') {
```
sgiehl, line 82: Was that change on purpose? Wondering as it doesn't seem the same logically.

sgiehl (commented):
I had a look through all the changes. As I'm not very experienced with typescript I can't really assess if all those changes would be needed or what they are good for. But as no tests are failing I guess it should be good.
I also clicked roughly through the UI, but for sure didn't retest every feature that had been migrated before. So in theory there could be some regressions introduced I didn't see.

plugins/CoreHome/vue/src/ContentBlock/ContentBlock.vue
diosmosis: Yes, actualFeature is initialized to this.feature which is typed to be a String, so the value will never be a boolean `true`. I don't think it was possible in angularjs either since the binding value was `@`, not `=` or `<`. Typescript complained about this being unnecessary if I recall correctly.

diosmosis:
@sgiehl 

> As I'm not very experienced with typescript I can't really assess if all those changes would be needed or what they are good for.

You can see why a change was made doing the following:

- revert the change locally
- run vue:build $pluginName. tsc will fail with an error and that will tell you why the change was made.

This PR is just fixing typing errors that tsc complains about when you set it to actually use and generate typing information (by default vue does not enable this, so much of the typescript code that was written was incorrect). This PR would not be large or even necessary if this setting change was done in the beginning, but vue does not seem to mention that it doesn't do real type checking.

sgiehl:
Yeah sure. Feel free to merge it then. If any regressions may pop up later we can fix them then as well. I don't think it's needed to test everything in detail again now

PR #18547, opened 2021-12-28
remove strong type on SendFeedbackForFeature function

### LINKED ISSUE #18537: Fatal error: In Feedback.sendFeedbackForFeature API
Getting below error.

> Error: {"message":"Argument 2 passed to Piwik\\Plugins\\Feedback\\API::sendFeedbackForFeature() must be of the type int, string given","file":"plugins\/Feedback\/API.php","line":38," "request_id":"d5fc2","backtrace":" on /plugins\/Feedback\/API.php(38)\n#0 [internal function]: Piwik\\Plugins\\Feedback\\API->sendFeedbackForFeature('Help', 'index.php\/.', 'dislikeother', 'e')\

I noticed this API is enforcing specific types. I'm thinking in API's we should likely not make use of PHP types like `(?string $featureName, int $like` as it could cause issues like fatal errors when you actually want a regular error message response that the input is not correct.

Could we maybe remove those types there and in general document to not use any types there? We could maybe even have an automated test that checks that no API has specific parameters enforced.

It can be frustrating otherwise when consuming the API and all you get is an HTTP 500 and you don't know what you're doing wrong.  Also we get emailed about it every time this happens which can be annoying too.

peterhashair:
### Description:

Fixes: #18537 
remove strong type on SendFeedbackForFeature function

### Review


plugins/Feedback/API.php
```diff
@@ -31,11 +31,11 @@ class API extends \Piwik\Plugin\API
      * config: "feedback_email_address".
      *
      * @param string|null $featureName  The name of a feature you want to give feedback to.
-     * @param int         $like         Whether you like the feature or not
+     * @param string|null $like         Whether you like the feature or not
```
justinvelluppillai, line 34: I guess this is a bit strange to use `string|null` and then set the default to `$like = false`.

plugins/Feedback/API.php
```diff
@@ -31,11 +31,11 @@ class API extends \Piwik\Plugin\API
...
      * @param string|null $choice       Multiple choice option chosen
      * @param string|null $message      A message containing the actual feedback
      */
-    public function sendFeedbackForFeature(?string $featureName, int $like, ?string $choice, ?string $message = null)
+    public function sendFeedbackForFeature($featureName, $like = false, $choice = null, $message = null)
```
justinvelluppillai: Maybe this should be `$like = null`

plugins/Feedback/vue/src/RateFeature/RateFeature.vue
```diff
@@ -257,7 +257,7 @@ export default defineComponent({
       AjaxHelper.fetch({
         method: 'Feedback.sendFeedbackForFeature',
         featureName: this.title,
-        like: this.like ? '1' : '0',
```
justinvelluppillai, line 260: Suggest keeping this line as is otherwise both `'true'` and `'false'` may be true in php

justinvelluppillai (changes requested):
Left a couple comments needing fixing

justinvelluppillai (approved):

sgiehl:
@justinvelluppillai @peterhashair didn't you see that the vue build action failed for this PR due to some code formatting issues? Would be good to always check that before merging, as now the action fails for all PRs after merging in latest changes...
We maybe should consider to require this action to pass for a PR to be mergeable.

justinvelluppillai:
@sgiehl sorry missed that - yes certainly the build should always be required to pass at least before merging.

sgiehl:
I have pushed a fix here: https://github.com/matomo-org/matomo/pull/18525/
Will merge the PR soon, so it will be fixed on 4.x-dev as well.

peterhashair:
@sgiehl yes, sorry, I forgot to update the VUE format.

PR #18542, opened 2021-12-24
[Vue] remove support in vue for FormField.allSettings

diosmosis:
### Description:

Fixes #18539

Deep watching for properties in an object property doesn't seem to work in FormField, and it's not particularly performant, so removing the allSettings property from Vue. It may be better in the long run to move this functionality further up the component tree also (instead of FormFIeld doing `v-show=showThisField`, we have `<FormField v-show=shouldShow>`).

Changes:

* Remove allSettings property from FormField.vue and just make the condition function handle that.
* Support deep watching in angularjs adapter where it works.
* In FieldTextAreaArray, check if ref is set before using it. Think this only happens in error states, but should still be useful to have it I guess.

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #18554, opened 2021-12-30
Revert "[Vue] remove support in vue for FormField.allSettings"

Swansoffie02:
Reverts matomo-org/matomo#18542


sgiehl:
Hi @Swansoffie02 
Is there any specific reason for reverting that one?

PR #18562, opened 2022-01-03
[Vue] convert sites manager single site template/controller to vue component

diosmosis:
### Description:

**This PR is based off of https://github.com/matomo-org/matomo/pull/18553.**

Changes:
* Fix typing error in SiteSelector.
* Convert sites-manager-site.controller.js to SiteFields.vue component.
* Add stores for timezone/currency/site type for use in SitesManager.
* Created lazyInitSingleton function that constructs a singleton when it is first accessed so we don't end up sending a bunch of ajax requests on page load for stores that may not be used.

image

### Review


sgiehl:
Here a few issues found while clicking through the UI. Some might possibly be unrelated to the changes here.
Only did some quick testing yet. Maybe something else pops up when I continue the review...

- [ ] When adding a new site the cancel button does not work
- [ ] When searching for a site has exactly 10 results the `next` button is clickable (even though there aren't any further results)

diosmosis:
> When searching for a site has exactly 10 results the next button is clickable (even though there aren't any further results)

This is a problem w/ 4.x-dev. It can be fixed easily but better to wait for these fixes until the next PR since the fix would still be in the angularjs code here.

diosmosis:
> When adding a new site the cancel button does not work

Fixed.

sgiehl:
@diosmosis Adding a rollup doesn't work with the changes here. The field for selecting the sites isn't displayed without any error. It works on 4.x-dev, so seems to be related to the changes here. Can you check if we can get that running again without changes in the RollUp plugin?

diosmosis:
@sgiehl should be fixed

plugins/SitesManager/vue/src/SiteFields/SiteFields.vue
```diff
@@ -0,0 +1,499 @@
...
+      let suffix = '';
+      if (this.isInternalSetupUrl) {
+        suffix = MatomoUrl.stringify({
+          idSite: site.idsite,
+          period: MatomoUrl.parsed.value.period,
+          date: MatomoUrl.parsed.value.date,
+          updated: 'false',
+        });
+      }
+      return `${this.howToSetupUrl}${suffix}`;
```
sgiehl: Here it seems a `&` is missing between the url and suffix

sgiehl (changes requested):
Found some more issues:

- When adding a new site, it automatically added at the top of the list. But the currency and timezone is always empty for those new sites:
  image
  When editing a site the display values in the list are also not updated.

Besides that it looks fine and should be good to merge, once those issues are fixed.

diosmosis:
> When adding a new site, it automatically added at the top of the list. But the currency and timezone is always empty for those new sites:

This is broken on 4.x-dev too. It should be fixed in the next PR.

> Here it seems a & is missing between the url and suffix

Fixed.

sgiehl (approved):

PR #18569, opened 2022-01-04
Fix for page overlay sidebar authentication failure

### LINKED ISSUE #18538: Fix page overlay sidebar authentication failure
When viewing the page overlay, on some sites, the sidebar may fail to authenticate and shows the following message:

> Error: Your session has expired due to inactivity

When this happens the AJAX call to the `renderSidebar` controller method is sent with a different session cookie than all the other requests which causes the API request to fail as the session is not authenticated.

See L3-197

## Expected Behavior
The page overlay sidebar should consistently use the currently logged in session and not show an error.

## Current Behavior
The page overlay sidebar intermittently fails to use the current session and loads with an error.

## Possible Solution

If a Javascript breakpoint is added to `Piwik_Overlay.js` at line 33 (the beginning of the `loadSidebar()` method) and execution is immediately continued when that breakpoint is hit, then `loadSidebar()` sends the correct cookie and the error does not occur. As soon as the breakpoint is removed the session expired error returns on the next reload.

It appears to be some sort of race condition where `loadSidebar()` somehow executes before the session cookie is available and the small delay introduced by the breakpoint prevents this happening.

Adding a small delay to the `loadSidebar()` call might fix this, but to confirm the fix this issue would need to be recreated in development environment first.

This is hard to recreate and has only been seen on one site so far.
1. Go to Behaviour > Pages
2. Choose a page url and click the 'Open Page Overlay' row action
3. In the new tab that opens the sidebar will display the 'session has expired' error

## Context
This prevents usage of the page overlay feature for affected sites.

## Your Environment

* Matomo Version: 4.6.2
* Matomo cloud
* Browser: Firefox, Chromium
* Operating System: Linux, MacOs

bx80:
### Description:

Fixes #18538 

When using the page overlay feature, the `notifyParentIframe` call intermittently gets sent without a session cookie. This is quite random and only occurs approximately once in every 20-30 calls. It appears to be a race condition related iframe access to cookies. 

Because there is no session cookie then a new session is started and the response to the `notifyParentIframe` sets a new session cookie which invalidates the session, this is apparent when the `loadSidebar` call is made shortly after and the `token_auth` is rejected and the user is forcibly logged out.

Without a solution to the race condition issue this PR is a workaround to prevent the session breaking. It's not possible to check the session cookie via javascript so a server side check is implemented which will remove the `set-cookie` header when responding to a `notifyParentIframe` without any cookies.

To recreate:
1) Create some visits to a test site clicking between one or two pages.
2) Open the page overlay view
3) Set the browser console to view network requested filtered by `notifyParentIframe`
4) Click back and forth between the pages that were tracked
5) Eventually the sidebar will show a 'session expired' error.
6) Check the last `notifyParentIframe` call in the console, it will be missing a request session cookie and will receive a set cookie for a new session

With this fix the `notifyParentIframe` call will still be made without a session cookie, but no set cookie will be returned in the response and the session will not break.

See also L3-197

### Review

### DIFF
--- a/plugins/Overlay/Overlay.php
+++ b/plugins/Overlay/Overlay.php
@@ -18,7 +18,8 @@ class Overlay extends \Piwik\Plugin
     {
         return array(
             'AssetManager.getJavaScriptFiles'        => 'getJsFiles',
-            'Translate.getClientSideTranslationKeys' => 'getClientSideTranslationKeys'
+            'Translate.getClientSideTranslationKeys' => 'getClientSideTranslationKeys',
+            'Platform.initialized'                   => 'handleIframeMissingCookie'
         );
     }
 
@@ -36,5 +37,18 @@ class Overlay extends \Piwik\Plugin
     {
         $translationKeys[] = 'General_OverlayRowActionTooltipTitle';
         $translationKeys[] = 'General_OverlayRowActionTooltip';
+    }
+
+    /**
+     * The Overlay.notifyParentIframe call is sometimes made without a session cookie due to an iframe related race
+     * condition, this results in a new session being created which logs the current user out.
+     * As a hacky workaround we detect this condition here and remove the set cookie header from the response to
+     * protect the session
+     */
+    public function handleIframeMissingCookie()
+    {
+        if ($_GET['action'] == 'notifyParentIframe' && count($_COOKIE) == 0) {
+            header_remove('Set-Cookie');
+        }
     }
 }
--- a/plugins/Overlay/templates/notifyParentIframe.twig
+++ b/plugins/Overlay/templates/notifyParentIframe.twig
@@ -9,10 +9,7 @@
     var piwikWindow = window.parent.parent;
     // notify piwik of location change
     // the location has been passed as the hash part of the url from the overlay session
-
-    setTimeout(function() {
-        piwikWindow.Piwik_Overlay.setCurrentUrl(window.location.hash.substring(1));
-    }, 500);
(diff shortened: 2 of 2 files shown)


sgiehl:
I tried to reproduce the issue for a while, but wasn't able to with my normal setup. So I started to have some deeper thoughts about this issue. I excluded a possible browser bug, so it needed to be something within Matomo that let's the browser struggle with the session cookie.

So I took a closer look at the requests that are sent by the Overlay. All requests that are sent to `index.php` are actually setting the session cookie. But there is one big difference those requests are having.
We have a special handling for the `samesite` cookie attribute implemented for all requests that are going to the Overlay module:
https://github.com/matomo-org/matomo/blob/9873cb71e66be0f80839c76e923c3b866dd23b46/core/Session.php#L166-L183

The overlay page is sending multiple requests to the API. Such requests are actually setting the value to `lax`, while the requests  going to the overlay module are setting the value to `none`.

That wasn't a problem in my setup as the tracked url and matomo were both running on the same domain. So I set up another vhost to have different urls for matomo and the tracked site. Using this set up it was easy to reproduce the problem.

So the problem actually is that depending on the request that last finished, before the `notifyParentIframe` request is sent, the samesite attribute is set to `lax` or `none`. With `none` everything works fine, but if the attribute is `lax` the iframe can't access the cookie, so it is not sent with the request, causing a new session to be started.

Your workaround would actually solve that, but it's not really a proper solution. I'll close this PR and create a new one with a better fix soon.

PR #18599, opened 2022-01-10
add disable finger print options

### LINKED ISSUE #18448: Add an option to disable fingerprinting / config_id entirely
## Update - the specs

- Fix in 5.0.0 - See the proposed solution in https://github.com/matomo-org/matomo/issues/18448#issuecomment-1512338767
- Ideally we would backport but it might be OK in that case not to backport

## Summary

This is continuation of #16361 and the [related discussion](https://forum.matomo.org/t/does-device-fingerprinting-require-consent/42869) in the forums. Right now, the stance of Matomo is that in cookie-less mode, [it doesn't use fingerprinting](https://matomo.org/faq/general/how-is-the-visitor-config_id-processed/) so it doesn't require consent. Under GDPR this line of thought is perfectly sound and reasonable. However, under the ePrivacy directive (which is now implemented in German law since 1 December, 2021 with the [TTDSG](https://dsgvo-gesetz.de/ttdsg/)), this is once again a debatable point.

The key phrase in the ePrivacy directive is the 'access to information' already stored on the user's device. Of course, it's not entirely clear what this means in practice. But call it a fingerprint or not, Matomo does [access some data from the user's device](https://github.com/matomo-org/matomo/blob/4.x-dev/js/piwik.js#L3130-L3187) to create the config_id (in particular, the screen size and supported mime types). It's really up for debate if this constitutes 'access to information' – but as long as there's no legal precedence, some clients will want to cover all their bases. So it will once again be necessary to require consent before using Matomo, even without cookies or 'real' fingerprinting in place. So having an option to disable the device detection further would be great. Instead, the config_id could be based solely on User-Agent and anonymised IP address (which the server receives implicitly, so no data access is required).

Of course, this would further limit the usefulnes of some reports. But this would be acceptable in order to be 100% sure that using Matomo without consent is completely safe, legally speaking.

peterhashair:
### Description:
Fixes: #18448 
add disable finger print

FAQ Preview: https://matomo.org/faq/how-do-i-disable-browser-feature-detection-completely/

### Review


peterhashair:
build js

peterhashair:
@tsteur sorry to bother, just checking if I am on the right track.

js/piwik.js
```diff
@@ -2439,7 +2439,9 @@ if (typeof window.Matomo !== 'object') {
                 uniqueTrackerId = trackerIdCounter++,
 
                 // whether a tracking request has been sent yet during this page view
-                hasSentTrackingRequestYet = false;
+                hasSentTrackingRequestYet = false,
+
+                fingerPrintEnable = true;
```
tsteur: We usually prefix these variables with `config` see eg `configConsentRequired` or `configCookiesToDelete`

js/piwik.js
```diff
@@ -3749,11 +3751,14 @@ if (typeof window.Matomo !== 'object') {
...
                     '&send_image=0';
 
-                var browserFeatures = detectBrowserFeatures();
-                // browser features
-                for (i in browserFeatures) {
-                    if (Object.prototype.hasOwnProperty.call(browserFeatures, i)) {
-                        request += '&' + i + '=' + browserFeatures[i];
+                //Browser FingerPrint
+                if(fingerPrintEnable) {
+                    var browserFeatures = detectBrowserFeatures();
```
tsteur, line 3759: there are also other places that use `detectBrowserFeatures`. It's probably best to simply return an empty object maybe or so in `detectBrowserFeatures` if the feature is not supposed to be used. We basically don't want to read any of these features when it's disabled.

js/piwik.js
```diff
@@ -4960,6 +4965,10 @@ if (typeof window.Matomo !== 'object') {
                 linkTrackingInstalled = false;
                 linkTrackingEnabled   = false;
             };
+            this.disableFingerPrint = function ()
+            {
+                fingerPrintEnable = false;
```
tsteur: I think when `setConsentGiven` or `setCookieConsentGiven` is called, then we can set `fingerPrintEnable = true` as the user has given consent and it should be fine to read those properties

js/piwik.js
```diff
@@ -4960,6 +4965,10 @@ if (typeof window.Matomo !== 'object') {
                 linkTrackingInstalled = false;
                 linkTrackingEnabled   = false;
             };
+            this.disableFingerPrint = function ()
```
tsteur: I wonder if we can find a better name here. Like `disableAccessBrowserFeatures` since someone could argue it's not a fingerprint that we create

tsteur (changes requested):
@peterhashair left a few comments, overall goes in the right direction

peterhashair:
build js

js/piwik.js
```diff
@@ -4960,6 +4967,10 @@ if (typeof window.Matomo !== 'object') {
                 linkTrackingInstalled = false;
                 linkTrackingEnabled   = false;
             };
+            this.disableBrowserFeatureDetection = function ()
```
sgiehl: I would assume that we need to add that new method to the list of methods applied first here:
https://github.com/matomo-org/matomo/blob/41b79fe4dbd179338df22c90f2a44aa7c63b4183/js/piwik.js#L7109

Am I right @tsteur?

tsteur: @sgiehl yes definitely 👍  we'd want to execute that method quite early in the process to make sure we don't access the browser features. We probably want to execute it before `disableCookies` or after. The earlier the better I guess.

peterhashair:
build js

peterhashair:
build js

justinvelluppillai:
@peterhashair why are you adding those eslint rules?

peterhashair:
@justinvelluppillai 
Ah maybe we need those ignore it throw lots of errors for pikwi.js, see screen shots
image

js/piwik.js
```diff
@@ -7106,7 +7118,7 @@ if (typeof window.Matomo !== 'object') {
...
+        var applyFirst = ['disableBrowserFeatureDetection', 'addTracker', 'forgetCookieConsentGiven', 'requireCookieConsent', 'disableCookies', 'setTrackerUrl', 'setAPIUrl', 'enableCrossDomainLinking', 'setCrossDomainLinkingTimeout', 'setSessionCookieTimeout', 'setVisitorCookieTimeout', 'setCookieNamePrefix', 'setCookieSameSite', 'setSecureCookie', 'setCookiePath', 'setCookieDomain', 'setDomains', 'setUserId', 'setVisitorId', 'setSiteId', 'alwaysUseSendBeacon', 'enableLinkTracking', 'setCookieConsentGiven', 'requireConsent', 'setConsentGiven', 'disablePerformanceTracking', 'setPagePerformanceTiming', 'setExcludedQueryParams'];
```
tsteur: @peterhashair fyi it should be right before `disableCookies` or after, not the first entry as this could cause issues that the method call is not applied to other added trackers

peterhashair:
build js

justinvelluppillai:
@peterhashair be good to get this one ready for next review and request reviewer if it is being ignored. Also looking at the FAQ I think maybe it is better to add the information to this page which describes how the config_id works more generally also https://matomo.org/faq/general/how-is-the-visitor-config_id-processed/ (or at least link to this page and the page on fingerprinting).

peterhashair:
@justinvelluppillai sure, will do.

peterhashair:
build js

js/piwik.js
```diff
@@ -112,7 +112,7 @@
      "", "\b", "\t", "\n", "\f", "\r", "\"", "\\", apply, call, charCodeAt, getUTCDate, getUTCFullYear, getUTCHours,
     getUTCMinutes, getUTCMonth, getUTCSeconds, hasOwnProperty, join, lastIndex, length, parse, prototype, push, replace,
     sort, slice, stringify, test, toJSON, toString, valueOf, objectToJSON, addTracker, removeAllAsyncTrackersButFirst,
-    optUserOut, forgetUserOptOut, isUserOptedOut, withCredentials, visibilityState
+    optUserOut, forgetUserOptOut, isUserOptedOut, withCredentials, visibilityState,
```
justinvelluppillai:
```suggestion
    optUserOut, forgetUserOptOut, isUserOptedOut, withCredentials, visibilityState
```

justinvelluppillai (approved):
Working for me, looking good also. @tsteur did you want to have a look?

tsteur:
@justinvelluppillai I haven't fully tested it but I think we could merge for now so it's in the release. Then add some follow up work like add also an `enable...` method (which they can call if they get consent) and maybe there could be also one or more automated tests for this? Like making sure when the feature is disabled, that we're not sending any browser resolution in the tracking request?

bx80 (approved):
Works for me. It would probably be a good idea to have a tracker test to make sure screen resolution and other fingerprint data are not recorded when this option is enabled to prevent any future regression.

justinvelluppillai:
This is looking good thanks @peterhashair, still good to publish your FAQ and update the other mentioned doc, after that once these tests pass this is good to merge and will probably be the last merge before the next RC.

peterhashair:
@justinvelluppillai updated the FAQ, just reference a link at the bottom. publish it now.

peterhashair:
FAQ https://matomo.org/faq/how-do-i-disable-browser-feature-detection-completely/

PR #18601, opened 2022-01-11
[Marketing]adding help link or search tracking campaign params

### LINKED ISSUE #18598: Add URL campaign parameters to in-app search and help links
We use these campaign parameters as we're partially using them already in other cases (we could change if it doesn't make sense)

```
mtm_campaign=App_Help&mtm_source=Matomo_App&mtm_medium=$DEPENDS_ON_CONTEXT
```

When someone searches for something in the quick search, or on the help page, and we then open "matomo.org" to view the search results, add campaign parameters so we know how many people used this feature.

Also add campaign parameters to the help links in report documentation (not sure there are any URLs)

image

and category help 
image

help links in the help page

image

peterhashair:
### Description:
adding help link or search tracking campaign params. 
Fixes: #18598 

### Review


core/Category/Subcategory.php
```diff
@@ -152,4 +152,11 @@ public function getHelp()
     {
         return null;
     }
+    public function getHelpWithTracking()
+    {
+        $helpText =  $this->getHelp();
+        preg_match('/"https:\/\/matomo.org(.*?)"/',$helpText,$matches);
+
+        return $matches[0];
+    }
```
sgiehl: This method seems unused

core/Piwik.php
```diff
@@ -889,4 +889,20 @@ public static function translate($translationId, $args = array(), $language = nu
...
+    /**
+     * @param $helpText
+     * @return string
+     */
+    public static function helpLinkFormat($helpText)
+    {
+        preg_match('/"https:\/\/matomo.org(.*?)"/', $helpText, $matches);
+        if (!empty($matches)) {
+            $replace = substr_replace($matches[0], "",
+                -1) . 'mtm_campaign=App_Help&mtm_source=Matomo_App&mtm_medium=cpc"';
```
sgiehl: Those parameters need to be added to the query. So a `?` needs to be added in between. Also currently links that already contain a query might get broken, as the `preg_match` only looks for a `/` in the end.

plugins/CoreHome/vue/src/helpLink.ts
```diff
@@ -0,0 +1,15 @@
...
+ * @link https://matomo.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ */
+
+export const keywordSearch = (keyword: string): string => {
+  let mtm_medium = 'cpc';
+  // test if is email
+  if (keyword.toLowerCase().match(/^([a-zA-Z0-9_\-.]+)%40([a-zA-Z0-9_\-.]+)\.([a-zA-Z]{2,5})$/)) {
+    mtm_medium = 'email';
+  }
```
sgiehl: Would you mind explaining what this medium detection is good for? If I understand that correctly, the medium should reflect which medium was used. So `cpc` (cost-per-click) would be some sort of advertisement someone clicked on, and `email` would be used, when the link is sent by mail and someone clicks on it. So I guess we could simply leave the medium away.

plugins/Feedback/templates/index.twig
```diff
@@ -23,12 +23,12 @@
 
             <h4>{{'Feedback_PopularHelpTopics'|translate}}</h4>
             {% for helpTopic in popularHelpTopics %}
-                <p> &bull;  <a rel="noreferrer noopener" target="_blank" href="{{helpTopic.url|e('html_attr')}}">{{helpTopic.title|translate}}</a></p>
+                <p> &bull;  <a rel="noreferrer noopener" target="_blank" href="{{ helpLink(helpTopic.url) }}">{{helpTopic.title|translate}}</a></p>
```
sgiehl: Here and for the links below it would be good to keep `|e('html_attr')` instead of `|raw`. This ensures that the html wouldn't break if the returned link would contain a character like `"`

sgiehl (changes requested):
Left a couple of comments for improvements.
Btw. did you think about solving this issue in javascript only? Instead of having the links with all campaign parameters within the API responses, we could have a new angular/vue component, which is used instead of a normal `<a>` tag. And when it's rendered it can automatically parse the link and append the campaign parameters if needed...

plugins/CoreHome/vue/src/helpLink.ts
peterhashair: @sgiehl right, I wasn't to sure about the mtm_medium.

peterhashair:
> did you think about solving this issue in javascript only? 

Yes, group most of them into a vue/angular directory. but the getHelp() doesn't like `span matomo-help-link base="https://matomo.org/docs/" text="test"></span>` or VUE version `<helpLink base="url" text='test'/>`, not too sure why, but update the regex, should be better

plugins/API/WidgetMetadata.php
```diff
@@ -191,7 +191,7 @@ private function buildCategoryMetadata($category)
             'name'  => $category->getDisplayName(),
             'order' => $category->getOrder(),
             'icon' => $category->getIcon(),
-            'help' => Piwik::translate($category->getHelp()),
+            'help' => Piwik::translate(Piwik::helpLinkFormat($category->getHelp())),
```
sgiehl: Guess that needs to be switched. So it replaces the link, after it was translated. On the other side I wonder if that is needed at all. I don't think we need links with campaigns in the API response (unless they are used somewhere in Matomo where we can't use VUE to add the campaign)

plugins/API/WidgetMetadata.php
```diff
@@ -209,7 +209,7 @@ private function buildSubcategoryMetadata($subcategory)
             'id'    => (string) $subcategory->getId(),
             'name'  => Piwik::translate($subcategory->getName()),
             'order' => $subcategory->getOrder(),
-            'help' => Piwik::translate($subcategory->getHelp()),
+            'help' => Piwik::translate(Piwik::helpLinkFormat($subcategory->getHelp())),
```
sgiehl: same as above

js/piwik.js
```diff
@@ -7122,7 +7122,7 @@ if (typeof window.Matomo !== 'object') {
                 }
             }
 
-            // replace initialization array with proxy object
+            // replace initcore/Category/Subcategory.phpialization array with proxy object
```
sgiehl: guess that was changed by accident. Might be good to quickly look through your changes in the PR, before sending something into review.

core/Twig.php
```diff
@@ -294,6 +295,14 @@ protected function addFunction_linkTo()
         $this->twig->addFunction($urlFunction);
     }
 
+    protected function addFunction_HelpLink()
```
sgiehl: I guess this one could be removed again, now as we are using an VUE component instead.

plugins/CoreHome/vue/src/HelpLink/HelpLink.vue
```diff
@@ -0,0 +1,32 @@
...
+  {{ text }}
+  </a>
+</template>
+<script lang="ts">
+import { defineComponent } from 'vue';
+
+export default defineComponent({
+  props: {
+    keyword: String,
+    base: String,
```
sgiehl: `base` sounds to me a bit misleading. It actually may contain a full link and in theory might even contain a query string already or a hash. To have that bulletproof I guess we would need to use some kind of url parser to first parse the url, adjust the query parameters and then recreate the URL again so it contains the correct components again. Not sure if there is an easy way in vue for that. ping @diosmosis

sgiehl (dismissed):

plugins/CoreHome/vue/src/HelpLink/HelpLink.vue
diosmosis: I would use a computed property to process the input data.

EDIT: If you're asking for a URL parsing utility, you can use https://developer.mozilla.org/en-US/docs/Web/API/URL (for query string parsing broadcast.getValuesFromUrl() unless you need deep associative array query parameters which is not supported by any library I know of)

tsteur: BTW we could maybe also simply hard code all the links without any component or something. Or at least in the JS. If we look at all the changes needed for this, and maintenance coming from it, might be easier to just add the parameters to the URL where needed and it's simple. If things ever change or needed somewhere else it's also easy to just add the parameters again since it's harder to figure out if there's a component for it etc.

tsteur: @sgiehl @peterhashair be good to think about if we can simplify all this so we also have to maintain less (component, tests, ...) and "know" (not needing to know there's a component for it) less. It also makes maybe things more clear. I know it was previously suggested to maybe use a component so sorry for sending another signal here but it's worth thinking about it. I'm not too much into it if it makes sense but all this complexity may not be needed for simply adding few campaign parameters that could be done in a few lines of code.

sgiehl: My thought was that we actually don't want to have the additional parameters in the API response, as that could be used anywhere outside of Matomo. Also adjusting the links in every help text is a lot more and we need to keep that in mind every time we adjust or add a new one. Having a component makes that automatically and is easy to adjust later

tsteur: That could be still the case though without a component that we set the parameters in a central place? I would assume that we won't change anything there too much and when a change is needed, it be extremely quick to do. And it might be more likely that we first run into test issues with this where tests need to be fixed or the component eventually needs to be migrated again or we're making the campaign tracking more complex and then need to refactor it all again etc. Generally for such things be good to keep things simple.

peterhashair: @tsteur So the idea is to do a global campaign variable, then append it to HTML href manually?

tsteur: I would have just put in the campaign parameters every time as a string, not even a global campaign variable. It's ultimately up to the team though. Just meaning it might make a lot more things simpler and faster

peterhashair: @tsteur right, hardcode all the href params, update it now.

sgiehl: I still have the opinion that a more global solution would have had more benefits. Like it would have been easier to adjust that later again. Or it would (if done right) e.g. have been possible to inject an additional tracking parameter on cloud, so we would have additional data how many people on cloud are clicking those links vs how many on self hosted.
Also there are a lot more links to documentation or faq spread across Matomo like in the diagnostic checks or in some no data or footer messages of reports. In addition the are more of such links in some of our plugins.

@tsteur feel free to go ahead and review and merge this PR if it sophisticates your needs.

tsteur: @peterhashair I would have eg expected to see the `plugins/Feedback/templates/index.twig` template changed. Can you go over the original issue and make sure that all the relevant articles are changed?

We'd also want to add a "medium" for some campaigns so we know on which link they clicked. 

Overall this looks a lot lot better though. Even if we had to change it eventually, it could be potentially changed with a search and replace. So far we don't really have any plans to change anything for cloud re the campaign parameters. If we plan to do that, then we can still see if there's a simple solution to it.

It must have been a lot faster to develop this way, easier to maintain and read, easier to review, easy to change and customise, etc.

plugins/Feedback/templates/index.twig
```diff
@@ -23,14 +23,14 @@
...
             {% for helpTopic in popularHelpTopics %}
-                <p> &bull;  <a rel="noreferrer noopener" target="_blank" href="{{helpTopic.url|e('html_attr')}}">{{helpTopic.title|translate}}</a></p>
+                <p> &bull;  <a rel="noreferrer noopener" target="_blank" href="{{helpTopic.url|e('html_attr')}}?mtm_campaign=App_Help&mtm_source=Matomo_App&mtm_keyword=UserGuides">{{helpTopic.title|translate}}</a></p>
             {% endfor %}
 
             <h4>{{ 'Feedback_CommunityHelp'|translate }}</h4>
-            <p> &bull; <a target="_blank" rel="noreferrer noopener" href="https://matomo.org/docs/">{{ 'Feedback_UserGuides'|translate }}</a> - {{ 'Feedback_ViewUserGuides'|translate("","") }}.</p>
-            <p> &bull; <a target="_blank" rel="noreferrer noopener" href="https://matomo.org/faq/">{{ 'Feedback_FAQs'|translate }}</a> - {{ 'Feedback_ViewAnswersToFAQ'|translate("","") }}.</p>
+            <p> &bull; <a target="_blank" rel="noreferrer noopener" href="https://matomo.org/docs?mtm_campaign=App_Help&mtm_source=Matomo_App&mtm_keyword=docs">{{ 'Feedback_UserGuides'|translate }}</a> - {{ 'Feedback_ViewUserGuides'|translate("","") }}.</p>
```
justinvelluppillai:
```suggestion
            <p> &bull; <a target="_blank" rel="noreferrer noopener" href="https://matomo.org/docs/?mtm_campaign=App_Help&mtm_source=Matomo_App&mtm_keyword=docs">{{ 'Feedback_UserGuides'|translate }}</a> - {{ 'Feedback_ViewUserGuides'|translate("","") }}.</p>
```

plugins/Feedback/templates/index.twig
```diff
@@ -23,14 +23,14 @@
...
+            <p> &bull; <a target="_blank" rel="noreferrer noopener" href="https://matomo.org/faq?mtm_campaign=App_Help&mtm_source=Matomo_App&mtm_keyword=FAQs">{{ 'Feedback_FAQs'|translate }}</a> - {{ 'Feedback_ViewAnswersToFAQ'|translate("","") }}.</p>
```
justinvelluppillai:
```suggestion
            <p> &bull; <a target="_blank" rel="noreferrer noopener" href="https://matomo.org/faq/?mtm_campaign=App_Help&mtm_source=Matomo_App&mtm_keyword=FAQs">{{ 'Feedback_FAQs'|translate }}</a> - {{ 'Feedback_ViewAnswersToFAQ'|translate("","") }}.</p>
```

plugins/Feedback/templates/index.twig
justinvelluppillai: pedantic, but otherwise it leads to an unnecessary 301 redirect

justinvelluppillai (approved):
Looking good here, I've just made one suggestion to avoid 301s when I click two of the links changed.

PR #18632, opened 2022-01-17
update build js auto trigger

peterhashair:
### Description:

- build js auto-trigger, when there is a change in js/piwik.js, no need to leave comments. (I guess this is to prevent test run before js updated)

### Review

### DIFF
--- a/.github/workflows/buildtrackerjs.yml
+++ b/.github/workflows/buildtrackerjs.yml
@@ -1,8 +1,10 @@
 name: Build Tracker JS files
 
 on:
-  issue_comment:
-    types: [created]
+  push:
+    paths:
+      - 'js/piwik.js'
+      - 'matomo.js'
 
 permissions:
   actions: read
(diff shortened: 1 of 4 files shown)


.github/workflows/buildtrackerjs.yml
```diff
@@ -1,8 +1,10 @@
 name: Build Tracker JS files
 
 on:
-  issue_comment:
-    types: [created]
+  push:
+    paths:
+      - 'js/piwik.js'
```
sgiehl, line 6: The reason I didn't add that in that way in the first place was that it might cause trouble when your branch gets silently updated by a github action without any notice. It might be kind of fine, when that happens for a pull request but I wouldn't do that for a random branch when something is pushed.
Also I'm not sure if we should run such actions on our "protected" branches, like 4.x-dev or the release branches. We wouldn't want to have any automatic commits there.

.github/workflows/buildvue.yml
```diff
@@ -1,8 +1,9 @@
 name: Build Vue files
 
 on:
-  pull_request:
-    types: [synchronize, opened]
+  push:
+    paths:
+      - 'plugins/**/vue/src/**'
```
sgiehl, line 6: This can't work without changing the action, as it currently works with some pull request refs, which aren't available on push

peterhashair:
@sgiehl just playground  😀

PR #18633, opened 2022-01-17
add enable browser detection and some tests

### LINKED ISSUE #18448: Add an option to disable fingerprinting / config_id entirely

peterhashair:
### Description:

Fixes: #18448 

add enable browser detection and some tests

[develop document](https://github.com/matomo-org/developer-documentation/pull/606)

[FAQ draft](https://matomo.org/wp-admin/post.php?post=48524&action=edit&lang=en)

### Review


peterhashair:
build js

peterhashair:
@sgiehl sorry to bother, just checking what would be the best approach for piwik.js _pag.push tests I found here, https://github.com/matomo-org/plugin-AnonymousPiwikUsageMeasurement/pull/68 add it here. not sure if that's correct.

sgiehl (changes requested):
@peterhashair seems there are some JSLint errors, that need to be fixed. Otherwise looks good.
@tsteur would that be something that should be included in 4.7.0 release?

tsteur:
Very low risk of breaking something so that could be included.

Be also good to add more tests. We could make sure no browser resolution and detected plugins information is in the tracking request when disabled. Seems the other usages of `detectBrowserFeatures` might be less testable because they are partially including random uids etc.

We'd also need to update the previously FAQ to mention the new method.

We also need to update https://developer.matomo.org/api-reference/tracking-javascript to mention the new methods.

BTW are these new methods mentioned in the developer changelog?

This is documented in https://developer.matomo.org/guides/jstracker-core#adding-a-new-public-js-tracker-method

peterhashair:
@tsteur regarding the tests, I was thinking this test include the browser resolution check. Should I do a fake tracking in Integration tests? 
```javascript 
    test("Browser detector feature Disable and enable", function() {
      var pattern = /res/;
      var tracker = Piwik.getTracker();
      tracker.disableBrowserFeatureDetection();
      var request = tracker.getRequest('hello=world');
      equal(pattern.test(request), false, 'Disable browser fingerprint');

      tracker.enableBrowserFeatureDetection();
      var request = tracker.getRequest('hello=world');
      equal(pattern.test(request), true, 'Enable browser fingerprint set');

    });
```

tsteur:
Above test looks good. You'll also want to check this for the other tracking parameters. And could also adjust the messages like `Disable browser fingerprint'` change it to `When browser fingerprint is disabled the request should not include browser resolution'`

sgiehl:
@peterhashair There are still some JSLint errors to be fixed. If you need help with that, let me know.

peterhashair:
@sgiehl found it, fixed it. But it seems like the release check failed because the file size increased?

tests/javascript/index.php
```diff
@@ -5220,6 +5220,27 @@ function testWithWrappedConsole(callback) {
...
+    test("Browser detector feature Disable and enable", function() {
+        var pattern = /(res=)|(cookie=)/;
+        var tracker = Piwik.getTracker();
+        var siteIdPattern = /idsite/;
+
+        tracker.enableBrowserFeatureDetection();
+        var requestWithFingerprint = tracker.getRequest('hello=world');
+
+        equal(siteIdPattern.test(requestWithFingerprint), true);
+        equal(pattern.test(requestWithFingerprint), true, 'When browser fingerprint is enabled the request should include browser resolution and cookie');
```
sgiehl: Doesn't the pattern actually tests if the string contains resolution or cookie?

tests/javascript/index.php
```diff
@@ -5220,6 +5220,27 @@ function testWithWrappedConsole(callback) {
...
+
+        tracker.disableBrowserFeatureDetection();
+        var requestWithoutFingerprint = tracker.getRequest('hello=world');
+
+        equal(siteIdPattern.test(requestWithoutFingerprint), true);
+        equal(pattern.test(requestWithoutFingerprint), false, 'When browser fingerprint is disabled the request should not include browser resolution or cookie');
+
+
+
```
sgiehl: maybe try to avoid too many useless empty lines

sgiehl (changes requested):
Left some minor comments. Otherwise looks good.

tests/javascript/index.php
peterhashair: @sgiehl ah should I change this part ` var pattern = /(res=)|(cookie=)/;`

PR #18634, opened 2022-01-17
Allow disabling the usage of the default geolocation provider as fallback

### LINKED ISSUE #18625: Geolocation wrong on internal network even with enable_language_to_country_guess=0
Our websites are accessed externally and internally (with an IP from the local network range).
GeoIp is installed and enabled.

Despite having set enable_language_to_country_guess=0, the geolocation information for users from the local network is set based on their browser language and therefore totally wrong.

This seems to happen in Common::extractCountryCodeFromBrowserLanguage and I’ve got the feeling that this function does not respect the config value correctly.

```
public static function extractCountryCodeFromBrowserLanguage($browserLanguage, $validCountries, $enableLanguageToCountryGuess)
{
	/** @var LanguageDataProvider $dataProvider */
	$dataProvider = StaticContainer::get('Piwik\Intl\Data\Provider\LanguageDataProvider');

	$langToCountry = $dataProvider->getLanguageToCountryList();

	if ($enableLanguageToCountryGuess) {
		if (preg_match('/^([a-z]{2,3})(?:,|;|$)/', $browserLanguage, $matches)) {
			// match language (without region) to infer the country of origin
			if (array_key_exists($matches[1], $langToCountry)) {
				return $langToCountry[$matches[1]];
			}
		}
	}

        // the following code should run only if $enableLanguageToCountryGuess is true

	if (!empty($validCountries) && preg_match_all('/[-]([a-z]{2})/', $browserLanguage, $matches, PREG_SET_ORDER)) {
		foreach ($matches as $parts) {
			// match location; we don't make any inferences from the language
			if (array_key_exists($parts[1], $validCountries)) {
				return $parts[1];
			}
		}
	}
	return self::LANGUAGE_CODE_INVALID;
}

```
## Expected Behavior
When setting "enable_language_to_country_guess=0" in config, the country code should never be guessed from the browser language. It should be "unknown".

## Current Behavior
Despite having enable_language_to_country_guess=0 set in config, the geolocation for users with local network IP addresses is extracted from the browser language resulting in completely wrong countries.

## Possible Solution
The function Common::extractCountryCodeFromBrowserLanguage should respect the setting correctly

sgiehl:
### Description:

fixes #18625

### Review


config/global.ini.php
```diff
@@ -773,6 +773,9 @@
 ; By setting this option to 0, you can prevent Super User from editing the Geolocation settings.
 enable_geolocation_admin = 1
 
+; By setting this option to 0, you can prevent the geolocator to fall back to the default provider, if the configured provider doesn't provide any results.
```
tsteur: haven't reviewed the PR or tested anything. Just wanted to mention that reading this config name and description it wasn't clear what it does or means. Maybe we can partially rename the setting name and/or the description? What does it do to the data or to the user when disabling this setting?

sgiehl: What about
```
; By setting this option to 0, you can prevent the geolocator to guess the country based on the users language, if the configured provider doesn't provide any results.
enabled_geolocation_guess_country_fallback = 1
```

tsteur: 👍  better

sgiehl: adjusted.

tsteur (commented):
Can we create an FAQ for this and explain the use case when someone might do this?

BTW can this also be used to disable geolocation? We've had few people asking for this in the past and if so, we should create a separate FAQ

sgiehl:
> BTW can this also be used to disable geolocation? We've had few people asking for this in the past and if so, we should create a separate FAQ

Actually this only disables the fallback. So if no geolocation provider is configured -> the default is used guessing the country by the language will still be done. But if we want to make that possible, we could simply create some kind of `Dummy` provider, which always returns an empty result set.

tsteur:
@sgiehl getting bit confused. So if we disable `enabled_geolocation_guess_country_fallback`, then there's still some guessing and a fallback happening within the default provider? That might be getting bit confusing as there's then also a different guessing happening?

Also I see there's a setting `[Tracker]enable_language_to_country_guess`. Would that mean this config should be in Tracker category too?

sgiehl:
Guess could make sense to move it to Tracker. The default provider is doing the language to country guess.
If you have configured e.g. GeoIP2 and this provider doesn't return any results then the GeoLocator falls back to the default provider. And that fallback can be disabled with the config. By default Matomo uses the default provider and that currently can't be disabled.
Hope that makes it a bit clearer.

tsteur:
Does it make sense in that case to name the setting like `enable_default_location_provider`?

When disabled, we use a `Dummy` provider. This way we can tackle this specific use case, plus we can also handle the use case for disabling geolocation altogether if I see this right? Also the name would be a bit more clear maybe and it be easier to explain/understand.

Would also need some FAQs then.

sgiehl:
@tsteur I've now adjusted everything a bit again. We now have a new `Disabled` provider. If that one is chosen geolocation is disabled. The current default provider is still used as default and fallback unless the Tracker option `enable_default_location_provider` is disabled. In that case the `Disabled` provider will be used as default and fallback.
Hope that makes everything a bit easier to understand.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ This is the Developer Changelog for Matomo platform developers. All changes in o
 
 The Product Changelog at **[matomo.org/changelog](https://matomo.org/changelog)** lets you see more details about any Matomo release, such as the list of new guides and FAQs, security fixes, and links to all closed issues. 
 
+## Matomo 4.8.0
+
+### New config.ini.php settings
+
+* The config setting `enabled_geolocation_guess_country_fallback` in `General` has been added. By setting this option to 0, you can prevent the geolocator to guess the country based on the users language, if the configured provider doesn't provide any results.
```
tsteur: Changelog still needs to be updated

tsteur (approved):
Left one more comment, otherwise looks good. Feel free to merge after the FAQs have been created.

sgiehl:
As of Matomo 4.8. there will be a new config setting `[Tracker] enable_default_location_provider = 0`, which can be used to disable the default geolocation provider. That way the fallback to this provider can also be disabled.

PR #18636, opened 2022-01-17
Enrich log_visit table index by a third column

JasonMortonNZ:
### Description:

**Issue:** [dev-2431](https://innocraft.atlassian.net/browse/DEV-2431)

This PR improves the existing log_visit table index by including a third column within the index. The schema change will allow all new installations to benefit from the improved index, while the update script is earmarked for the 5.0.0 release.

### Review

### DIFF
--- a/core/Updates/5.0.0-b1.php
+++ b/core/Updates/5.0.0-b1.php
@@ -39,9 +39,7 @@ class Updates_5_0_0_b1 extends PiwikUpdates
 
     public function doUpdate(Updater $updater)
     {
-        if ($this->requiresUpdatedLogVisitTableIndex()) {
-            $updater->executeMigrations(__FILE__, $this->getMigrations($updater));
-        }
+        $updater->executeMigrations(__FILE__, $this->getMigrations($updater));
     }
 
     public function getMigrations(Updater $updater)
--- a/core/Updates/5.0.0-b1.php
+++ b/core/Updates/5.0.0-b1.php
@@ -16,12 +16,12 @@ use Piwik\Updater;
 use Piwik\Updater\Migration\Db as DbAlias;
 use Piwik\Updater\Migration\Db\DropIndex;
 use Piwik\Updater\Migration\Db\Sql;
-use Piwik\Updates;
+use Piwik\Updates as PiwikUpdates;
 
 /**
  * Update for version 5.0.0-b1
  */
-class Updates_5_0_0_b1 extends Updates
+class Updates_5_0_0_b1 extends PiwikUpdates
 {
     private $tableName;
     private $indexName;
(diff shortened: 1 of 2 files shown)


core/Updates/5.0.0-b1.php
```diff
@@ -0,0 +1,81 @@
...
+
+            // Using the base `Sql` class instead of the AddIndex class as it doesn't support DESC collation
+            $sql = "ALTER TABLE `{$this->tableName}` ADD INDEX `{$this->indexName}` (`idsite`, `idvisitor`, `visit_last_action_time` DESC)";
+            $addIndex = new Sql($sql, [DbAlias::ERROR_CODE_DUPLICATE_KEY, DbAlias::ERROR_CODE_KEY_COLUMN_NOT_EXISTS]);
+            $addIndex->exec();
+        } catch (\Exception $e) {
+            if (!$dropIndex->shouldIgnoreError($e)) {
+                throw $e;
+            }
+        }
```
sgiehl: Is there a reason why this isn't done using migrations like e.g. here:
https://github.com/matomo-org/matomo/blob/0e7530e93f0ada22024e7226badc1d94c8309f3c/core/Updates/4.0.0-rc3.php#L20-L44

Using migrations makes it possible for the user, to see the queries before running the update (and run them manually instead)

JasonMortonNZ: @sgiehl Yes, the AddIndex class wouldn't allow me to easily use the `DESC` collation with the `visit_last_action_time` column. I'm open to ideas if you know of any methods to make this work?

sgiehl: You can still use a custom sql with migrations. e.g. `$this->migration->db->sql('....')`

core/Updates/5.0.0-b1.php
```diff
@@ -16,44 +16,49 @@
...
     public function doUpdate(Updater $updater)
     {
         if ($this->requiresUpdatedLogVisitTableIndex()) {
-            $this->updateLogVisitTableIndex();
+            $updater->executeMigrations(__FILE__, $this->getLogVisitTableMigrations());
         }
     }
 
-    private function updateLogVisitTableIndex()
+    private function getLogVisitTableMigrations()
```
sgiehl, line 54: You should use the methods names that are defined in the base class, so the updater is able to determine the updates before they are executed. So there should be a `getMigrations` method that returns the migrations that are required to execute (it should return nothing if `requiresUpdatedLogVisitTableIndex` return true though) and the `doUpdate` should execute the updates, if any are returned from `getMigrations`

JasonMortonNZ: @sgiehl I've updated this to make use of the `getMigrations` method in the base class.

sgiehl (commented):
@JasonMortonNZ This looks good now and should be good to merge
Just one last thing that needs to be clarified. How much of a need is the descending order? That is actually only supported by MySQL 8. Older MySQL versions will simply ignore the `DESC`

@tsteur If we are starting to use some MySQL 8 features, we maybe should recommend using MySQL 8 in our [docs](https://matomo.org/docs/requirements/), even though we are still supporting older versions.

tsteur:
@sgiehl changed that guide to recommend MySQL 8. 

> Just one last thing that needs to be clarified. How much of a need is the descending order

It's not needed. On Cloud we don't run it on MySQL 8 either but it still helped.

sgiehl:
@tsteur if that's the case, why do we know that a `DESC` key would be better, and shouldn't we then maybe change the other keys  containing `visit_last_action_time` to `DESC` as well? :thinking:

tsteur:
The find visitor query does a DESC and so the most others if I remember correctly. I don't think it's needed for the others but we could always test later down the road. We haven't seen any issues there so far.

sgiehl (approved):
Should be good to merge

sgiehl:
@tsteur feel free to put this into a milestone it should be included or even merge it right away

PR #18638, opened 2022-01-17
setting save success, scroll to notification

### LINKED ISSUE #18637: Saving general settings successfully no longer scrolls to the success notification
I checked with Matomo 4.6 and this seems to have worked there so it looks like this is a regression.

When saving settings, you now never really know whether it worked and it's confusing.

peterhashair:
### Description:

Fixes: #18637 
after setting save success, should scroll to notification

### Review


plugins/CorePluginsAdmin/vue/src/PluginSettings/PluginSettings.vue
```diff
@@ -227,7 +227,7 @@ export default defineComponent({
           context: 'success',
           type: 'transient',
         });
-        NotificationsStore.scrollToNotification('generalSettings');
```
diosmosis, line 225: w/ the new code the way to do this is:

```
const notificationInstanceId = NotificationsStore.show({
  message: translate('CoreAdminHome_PluginSettingsSaveSuccess'),
  id: 'generalSettings',
  context: 'success',
  type: 'transient',
});
NotificationsStore.scrollToNotification(notificationInstanceId);
```

The reason for this change is because Notification.vue makes an AJAX request in markNotificationAsRead() if an ID is specified, but that isn't needed for notifications added client side only. So if you want to avoid making that request, you don't have to specify the `id:` property.

diosmosis: I'll add documentation for this later unless it's desired to revert this to the old behavior.

cc @sgiehl

peterhashair: @diosmosis that makes more sense. updating it now

diosmosis (approved):
looks good if tests pass

peterhashair:
@diosmosis sorry to bother got a question. when I run `./console vue:build` it returns random errors on 4.x-dev. 
image

Also here, should that be a catch method instead of finally? https://github.com/matomo-org/matomo/blob/165be4f4ea40cbca50ae71821e517e6edb6054ba/plugins/CoreAdminHome/vue/src/ArchivingSettings/ArchivingSettings.vue#L178

diosmosis:
> when I run ./console vue:build it returns random errors on 4.x-dev.

I've noticed this, seems to be a tsc issue/bug, but I don't know how to fix it. Currently just rebuilding when it happens. It doesn't happen very often for me (once a day at most).

> Also here, should that be a catch method instead of finally? 

It should be finally, unless you want to show the loading gif if there's an error. (EDIT: "no error" not an error.)

peterhashair:
@diosmosis I saw someone solved it by add  this https://www.npmjs.com/package/eslint-import-resolver-typescript

diosmosis:
Thanks, @peterhashair! Do you have info on how they used it, specifically w/ vue CLI?

peterhashair:
@diosmosis haven't try myself, but they claim work like this https://github.com/nklayman/vue-cli-plugin-electron-builder/issues/1059

PR #18643, opened 2022-01-18
enable case insensitive on segment search

### LINKED ISSUE #18640: Expandable search field no longer search case insensitive and neither
image

When I search for `event` it no longer finds `Event category`. I have to use a capital `E`.

Pretty sure this used to work and seems to be a regression. 

Not sure if similar issu

peterhashair:
### Description:
Fixes: #18640
enable case insensitive on segment search

### Review


peterhashair:
@diosmosis did a simple update for segment insensitive search. Hopefully, I did it right.

plugins/CorePluginsAdmin/vue/src/FormField/FieldExpandableSelect.vue
```diff
@@ -48,7 +48,7 @@
           <ul v-show="showCategory === options.group || searchTerm" class="collection secondLevel">
             <li
               class="expandableListItem collection-item valign-wrapper"
-              v-for="children in options.values.filter(x => x.value.indexOf(searchTerm) !== -1)"
+              v-for="children in options.values.filter(x => searchRegExp.test(x.value))"
```
sgiehl: This would actually enable the possibility to use regular expressions in the search box. The problem is that it would throw an error if you type an invalid regex like `[a` in the search box.
To only change that to a insensitive match I guess the simples solution would be to apply a `toLowerCase` on both strings.

sgiehl:
btw @peterhashair  As the issue is meant to be included in the 4.7.0 final, your pull request should have been based on and targeted to the `next_release` branch. I will rebase your branch and change the PR base. You may need to do a hard reset to the origin before doing any further changes though

plugins/CorePluginsAdmin/vue/src/FormField/FieldExpandableSelect.vue
peterhashair: @sgiehl that make sense, should reverse to that https://github.com/matomo-org/matomo/pull/18643/commits/70dd0ed253e5e92e9cb6c23a760aa515ccb9b307 solution. Was trying to make the function short :)

sgiehl:
@peterhashair This looks good for me now. 
For regressions its always good to check if we can adjust our tests so this will be covered for the future. Not sure if we already have a UI test that uses such a field for searching, but guess that could be added to the SegmentSelectorEditor tests maybe.

justinvelluppillai:
Maybe would be better to have a UI test that performs a search with a capital letter in it and then checks the results are the same?

sgiehl (changes requested):
@peterhashair the `capitalized` screenshot needs to be updated. It's currently an empty file.
Once tests are passing, that is good to merge.

peterhashair:
I guess that should be good to go

PR #18648, opened 2022-01-18
Fix session timeouts in overlay session

### LINKED ISSUE #18538: Fix page overlay sidebar authentication failure

sgiehl:
### Description:

As commented in #18569:

> I tried to reproduce the issue for a while, but wasn't able to with my normal setup. So I started to have some deeper thoughts about this issue. I excluded a possible browser bug, so it needed to be something within Matomo that let's the browser struggle with the session cookie.
> 
> So I took a closer look at the requests that are sent by the Overlay. All requests that are sent to `index.php` are actually setting the session cookie. But there is one big difference those requests are having.
> We have a special handling for the `samesite` cookie attribute implemented for all requests that are going to the Overlay module:
> https://github.com/matomo-org/matomo/blob/9873cb71e66be0f80839c76e923c3b866dd23b46/core/Session.php#L166-L183
> 
> The overlay page is sending multiple requests to the API. Such requests are actually setting the value to `lax`, while the requests  going to the overlay module are setting the value to `none`.
> 
> That wasn't a problem in my setup as the tracked url and matomo were both running on the same domain. So I set up another vhost to have different urls for matomo and the tracked site. Using this set up it was easy to reproduce the problem.
> 
> So the problem actually is that depending on the request that last finished, before the `notifyParentIframe` request is sent, the samesite attribute is set to `lax` or `none`. With `none` everything works fine, but if the attribute is `lax` the iframe can't access the cookie, so it is not sent with the request, causing a new session to be started.

This fix actually checks if the referrer contains `module=Overlay` to set the `samesite=none` for those requests as well. This should properly work unless the referrer gets blocked for some reason.
An alternative would be to send some kind of magic parameter with all requests that are sent by the overlay, so we can check for that parameter.

While testing that I also saw, that the Overlay was sending a request to `API.getPagesComparisonsDisabledFor`, which is not needed, so I added the Overlay module to the exclude list for that.

@tsteur I have targeted the `next_release` branch as it might be nice to include that fix maybe. If you think it's to risky, feel free to change the milestone and I'll rebase the branch to `4.x-dev` instead.

fixes #18538
replaces #18569

### Review


core/Session.php
```diff
@@ -171,9 +171,12 @@ public static function getSameSiteCookieValue()
         $module = Piwik::getModule();
         $action = Piwik::getAction();
 
+        $referrerUrlQuery = parse_url(Url::getReferrer() ?? '', PHP_URL_QUERY);
+        $comingFromOverlay = $referrerUrlQuery && strpos($referrerUrlQuery, 'module=Overlay') !== false;
+
         $isOptOutRequest = $module == 'CoreAdminHome' && $action == 'optOut';
         $isOverlay = $module == 'Overlay';
-        $shouldUseNone = !empty($general['enable_framed_pages']) || $isOptOutRequest || $isOverlay;
+        $shouldUseNone = !empty($general['enable_framed_pages']) || $isOptOutRequest || $isOverlay || $comingFromOverlay;
```
tsteur: Haven't tested it but generally looks good. I wonder if there's any kind of security issue that you could get Matomo to use `None` just by setting any kind of referrer like `foobarmodule=Overlay`? I don't know what the request looks like and if we can further restrict / limit this to a specific module/action etc?

bx80 (approved):
Makes sense now that the sequence of the multiple API calls was the race condition combined with different same site values eventually affecting iframe cookie availability, great spot! :exploding_head:  :+1: 

This is a much cleaner fix than my approach of knobbling the new session cookie.

I've been doing all testing with a separate vhost/domain - with this fix applied I did 100 page overlay link clicks without issue, then switched the the 4.x-dev branch and got a session expired error within three clicks, back to the fix branch and another 50 link clicks without issue. So looks like it solves the issue well.

core/Session.php
bx80: I just tested adding `&foobarmodule=Overlay` to a random url and it forces `none`.

Maybe matching `&module=Overlay` instead of `module=Overlay` would prevent this?

sgiehl: @tsteur getting a same site none cookie was also possible before by sending a request to the overlay module.
But also using a parameter for this would have the same risk. Not sure if there is another solution. At least none I can think of right now.

tsteur: Which request is it that this change is needed for? 
Ideally we could do something like `$referrerUrlQuery && strpos($referrerUrlQuery, '&module=Overlay') !== false; && $module === '' && $action = ''` or `in_array($module, ...)`

sgiehl: I'll check that and add an allow list to the check, but we need to keep that in mind when changing the overlay later

sgiehl: I've added the check and moved it to the overlay plugin, so it's easier to find

plugins/Overlay/Overlay.php
```diff
@@ -37,4 +41,39 @@ public function getClientSideTranslationKeys(&$translationKeys)
...
+     * See https://github.com/matomo-org/matomo/pull/18648
+     */
+    public static function isOverlayRequest()
+    {
+        $module = Piwik::getModule();
+        $action = Piwik::getAction();
+        $method = Common::getRequestVar('method', '', 'string');
+
+        $isOverlay = $module == 'Overlay';
+        $referrerUrlQuery = parse_url(Url::getReferrer() ?? '', PHP_URL_QUERY);
```
tsteur: just to be safe we could additionally get the host of the referrer, and validate it's a valid host

plugins/Overlay/Overlay.php
```diff
@@ -37,4 +41,39 @@ public function getClientSideTranslationKeys(&$translationKeys)
...
+        $isPossibleOverlayRequest = (
+            $module === 'Proxy' // JS & CSS requests
+            || ($module === 'API' && 0 === strpos($method, 'Overlay.')) // Overlay API data
+            || ($module === 'CoreHome' && $action === 'getRowEvolutionPopover') // Row evolution
+            || ($module === 'CoreHome' && $action === 'getRowEvolutionGraph') // Row evolution (graph)
+            || ($module === 'CoreHome' && $action === 'saveViewDataTableParameters') // store chart changes (within row evolution & transitions)
+            || $module === 'Annotations' // required to interact with annotations in evolution charts (within row evolution)
+            || ($module === 'Transitions' && $action === 'renderPopover') // Transitions
+            || ($module === 'API' && 0 === strpos($method, 'Transitions.')) // Transitions API data
+            || ($module === 'Live' && $action === 'indexVisitorLog') // Visits Log
```
tsteur, line 69: is it actually needed for visits log, transitions and row evolution? Not sure I understand the issue right but would have thought that these requests aren't issued from within an iframe? i'm not too much into the problem though so feel free to ignore it.

tsteur (commented):
Added another comment to check. Will also create an issue that we document how `Overlay` works technically as it's always a bit confusing and this might help understand things better.

plugins/Overlay/Overlay.php
sgiehl: The cookie is shared between the overlay page and the iframe. If you open the rowevolution the cookie would then be changed to samesite=lax and the iframe would loose access to it. It might not be required at that point but I wanted to avoid possible further regressions for any edge cases.

plugins/Overlay/Overlay.php
sgiehl: Added an additional check for valid hosts

tsteur: 👍

plugins/Overlay/Overlay.php
tsteur: Not sure what to do here. From a security point of view I would maybe rather remove them. I'm thinking for these UIs you could otherwise quite easily force a `None` samesite attribute by using a URL like `/index.php?module=Overlay&action=generalSettings&idSite=1&period=day&date=yesterday&module=CoreAdminHome`.

Not sure how big of an issue it is, maybe it would allow more attacks so would rather restrict it where possible.

Eventually could also check if `module` is twice in the URL etc but these things can get quite complicated and break things for users.

sgiehl: @tsteur I've improved the code a bit, so the query of the referrer is parsed. This at least ensures that the module parameter isn't provided twice...

plugins/Overlay/Overlay.php
```diff
@@ -37,4 +41,41 @@ public function getClientSideTranslationKeys(&$translationKeys)
...
+     */
+    public static function isOverlayRequest()
+    {
+        $module = Piwik::getModule();
+        $action = Piwik::getAction();
+        $method = Common::getRequestVar('method', '', 'string');
+
+        $isOverlay = $module == 'Overlay';
+        $referrerUrlQuery = parse_url(Url::getReferrer() ?? '', PHP_URL_QUERY);
+        parse_str($referrerUrlQuery, $referrerUrlQueryParams);
```
tsteur: not sure if it makes sense to use `UrlHelper::getArrayFromQueryString()` which we also use in other places? Probably not too important as it might do the same, just wondering for consistency

tsteur (commented):
Looks good. Be great though to add some tests to make sure eg when passing the Overlay parameter twice in the URL, that it behaves correctly as this is something that could break again in the future. Be great to add a few tests for `isOverlayRequest` (eg could make it `isOverlayRequest($module, $action, $method, $referrer)`  for better testability )

plugins/Overlay/Overlay.php
sgiehl: I guess it doesn't do the exactly same for parameter arrays, but should be fine to use that one as well. will change that

sgiehl:
@tsteur I've added a bunch of tests for the new method.

tsteur:
@sgiehl seems various other commits are included here? As the release might be today have to maybe target 4.x-dev after all?

sgiehl:
@tsteur Sorry I merged in 4.x-dev without thinking much about it. Rebased the branch so the changes aren't included anymore. Feel free to merge if needed or change the milestone and I can rebase to 4.x-dev again

tsteur:
@sgiehl I'll change the milestone 👍

sgiehl:
@tsteur changed the base to 4.x-dev. Should be ready for a last review now....

tsteur:
Looks good to me from a code perspective. @bx80 not sure if you can still easily reproduce the issue locally. If so, could you maybe also confirm it works?

bx80 (approved):
I retested clicking around the overlay repeatedly without encountering the original issue which is quickly apparent on 4.x-dev. The unit tests all pass. 

Might be outside the scope of this change, but on PHP8.1 the tests show a deprecation warning `PHP Deprecated:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/matomo/core/UrlHelper.php on line 205` which is easily fixed by adding a null check to that line in UrlHelper:
```
if ($urlQuery === null || strlen($urlQuery) == 0) { 
```

PR #18653, opened 2022-01-19
add date, period params into common function

### LINKED ISSUE #18639: strtolower(): Argument #1 ($string) must be of type string, array given in Date.php
## First problem:

> Error: {"message":"strtolower(): Argument #1 ($string) must be of type string, array given","file":"\/core\/Date.php","line":1079,"request_id":"ba369","backtrace":" on \/core\/Date.php(1079)\n#0 \/core\/Date.php(1079): strtolower(Array)\n#1 \/core\/Period\/Range.php(255): Piwik\\Date->addPeriod(-29, Array)\n#2 \/core\/Period.php(131): Piwik\\Period\\Range->generate()\n#3 \/core\/Period\/Range.php(152): Piwik\\Period->getDateStart()\n#4 \/core\/Period\/Range.php(533): Piwik\\Period\\Range->getDateStart()\n#5 \/core\/Plugin\/Controller.php(502): Piwik\\Period\\Range::getRelativeToEndDate(Array, 'last30', Object(Piwik\\Date), Object(Piwik\\Site))\n#6 \/plugins\/MultiSites\/Controller.php(97): Piwik\\Plugin\\Controller->getGraphParamsModified()\n#7 \/plugins\/MultiSites\/Controller.php(37): Piwik\\Plugins\\MultiSites\\Controller->getSitesInfo(false)\n#8 [internal function]: Piwik\\Plugins\\MultiSites\\Controller->index()\n#9 \/core\/FrontController.php(619): call_user_func_array(Array, Array)\n#10 \/core\/FrontController.php(168): Piwik\\FrontController->doDispatch('MultiSites', 'index', Array)\n#11 \/core\/dispatch.php(32): Piwik\\FrontController->dispatch()\n#12

Happens on
* `/index.php?action=index&date=yesterday&idSite=1&module=MultiSites&period[]=day`

## Second problem

> Error: {"message":"urldecode(): Argument #1 ($string) must be of type string, array given","file":"\/core\/Date.php","line":138,"request_id":"83d50","backtrace":" on \/core\/Date.php(138)\n#0 \/core\/Date.php(138): urldecode(Array)\n#1 \/core\/Period.php(121): Piwik\\Date::factory(Array)\n#2 \/core\/Plugin\/Controller.php(644): Piwik\\Period::checkDateFormat(Array)\n#3 \/core\/Plugin\/Controller.php(616): Piwik\\Plugin\\Controller->setGeneralVariablesViewAs(Object(Piwik\\View), 'basic')\n#4 \/plugins\/MultiSites\/Controller.php(100): Piwik\\Plugin\\Controller->setGeneralVariablesView(Object(Piwik\\View))\n#5 \/plugins\/MultiSites\/Controller.php(37): Piwik\\Plugins\\MultiSites\\Controller->getSitesInfo(false)\n#6 [internal function]: Piwik\\Plugins\\MultiSites\\Controller->index()\n#7 \/core\/FrontController.php(619):

Happens on 
* `index.php?action=index&date[]=yesterday&idSite=1&module=MultiSites&period=day`

## Expected outcome

No fatal error

peterhashair:
### Description:

Fixes: #18639 

Extend `common:getRequestVar`, if type = null, but array passed, convert to string.

### Review


core/Common.php
```diff
@@ -574,6 +575,11 @@ public static function getRequestVar($varName, $varDefault = null, $varType = nu
             settype($value, $varType);
         }
 
+        if (is_array($requestArrayToUse[$varName])) {
+            $value = implode(",", $value);
+        }
```
sgiehl: I don't think that is a proper solution, as it might break other places, where we expect an array, but don't pass a type maybe.

peterhashair: @sgiehl right, I guess another solution is to do a name check, but that looks urgly. Or we update everyplace has common:getRequestVar('period') and common:getRequestVar('date').  sth like common:getRequestVar('date')->toString(). but normal an error notification expected to show, when the params is on wrong type.
```php
 if (is_array($requestArrayToUse[$varName]) && in_array($varName,['period', 'date']) ) {
            $value = implode(",", $value);
        }
```

sgiehl: We should avoid such special handling I would say. Personally I would prefer a method like `Piwik::getDate($default=null)` / `Piwik::getPeriod($default=null)`, that performs the `getRequestVar` correctly and use that everywhere. Like we have `Piwik::getModule()`
But maybe let's discuss that on #18639

core/Piwik.php
```diff
@@ -889,4 +889,22 @@ public static function translate($translationId, $args = array(), $language = nu
 
         return $translator->translate($translationId, $args, $language);
     }
+
+    public static function getPeriod($paramName = 'period', $default = false)
```
sgiehl: I guess there is no sense in making the paramName an parameter, as it is always `period`, and the method shouldn't be used for anything else. Also providing `null` as default makes more sense

```suggestion
    public static function getPeriod($default = null)
```

core/Piwik.php
```diff
@@ -889,4 +889,22 @@ public static function translate($translationId, $args = array(), $language = nu
...
+    {
+        $period = Common::getRequestVar($paramName, $default);
```
sgiehl:
```suggestion
        $period = Common::getRequestVar('period', $default, 'string');
```

Using that one, will automatically throw an exception if period is provided as an array and convert it to string otherwise. If at some point we don't want to throw and exception if the period is missing we can provide an empty string as default.

core/Piwik.php
```diff
@@ -889,4 +889,22 @@ public static function translate($translationId, $args = array(), $language = nu
...
+        if (is_array($period)) {
+            $period = implode(',', $period);
+        }
```
sgiehl: There is no need fo that special handling. If it's and array, it's fine if the `Common::getRequestVar` above throws and exception. As it is simply unsupported and was manually provided by the user. Concatenating would cause other issues later, which would need to handle again.

core/Piwik.php
```diff
@@ -889,4 +889,22 @@ public static function translate($translationId, $args = array(), $language = nu
...
+    }
+
+    public static function getDate($paramName = 'date', $default = false)
+    {
+        $date  = Common::getRequestVar($paramName, $default);
+        if (is_array($date)) {
+            $date = implode(',', $date);
+        }
+        return $date;
+    }
```
sgiehl: Same applies to that method.

sgiehl (changes requested):

core/Piwik.php
```diff
@@ -889,4 +889,14 @@ public static function translate($translationId, $args = array(), $language = nu
 
         return $translator->translate($translationId, $args, $language);
     }
+
+    public static function getPeriod($default = null)
+    {
+        return Common::getRequestVar('period', $default, 'string');
+    }
+
+    public static function getDate( $default = false)
```
sgiehl: The default should be null here as well

```suggestion
    public static function getDate($default = null)
```

core/Plugin/Controller.php
```diff
@@ -477,7 +477,7 @@ protected function getLastUnitGraphAcrossPlugins($currentModuleName, $currentCon
     protected function getGraphParamsModified($paramsToSet = array())
     {
         if (!isset($paramsToSet['period'])) {
-            $period = Common::getRequestVar('period');
```
sgiehl, line 480: Haven't had a closer look at the usage here. But we need to check everywhere we change that, that it might change the behavior. If no `vartype` is passed to `Common::getRequestVar` it simply returns `null` if the parameter can't be found in the request. The new methods are both passing `string` as `varType`, causing an exception if no default is provided.

peterhashair: @sgiehl that make sense. Would that works 
```php
  try {
     try {
            $period = isset($paramsToSet['period']) ?? Piwik::getPeriod();
        } catch (\Exception $e) {
            $period = null;
        }
```

PR #18665, opened 2022-01-20
Avoid loading widget urls twice

### LINKED ISSUE #18662: Admin home issues requests and then cancels them again
This might be a regression in Matomo 4.7. At least I can't reproduce it on the cloud where an older version of Matomo runs. I noticed this because sometimes it would then show me an alert notification in the admin home page but not always.

image

Only managed to see this error when the developer tools are closed. It happens after clicking on the Admin cog icon 
image

Noticed this eg on https://demo.matomo.org/index.php?module=CoreAdminHome&action=home&idSite=1&period=day&date=yesterday

See network tab in browser developer tools:

image

sgiehl:
### Description:

@diosmosis It seems that watching the parameters is also triggered after the component was mounted, currently causing the widget url to be loaded twice.
Tried avoiding that by checking if the parameters actually changed. Locally that worked well. Or is there a better solution for that?

fixes #18662 

### Review


plugins/CoreHome/vue/src/WidgetLoader/WidgetLoader.vue
```diff
@@ -77,8 +77,8 @@ export default defineComponent({
     };
   },
   watch: {
-    widgetParams(parameters: QueryParameters) {
-      if (parameters) {
+    widgetParams(parameters: QueryParameters, oldParameters: QueryParameters) {
+      if (parameters && JSON.stringify(parameters) !== JSON.stringify(oldParameters)) {
```
diosmosis, line 81: If this is also executed on mount, the mounted() code could just be removed?

sgiehl: Wasn't sure if that is always the case or some kind of race condition because of the angularjs <> vue mixup maybe.

diosmosis: Hm, I'll investigate as I don't understand exactly what's happening here.

diosmosis: I couldn't reproduce this, tried in the dashboard and in a report page. It's possible the parameters object is set to a new object that has the exact same value, maybe when loading the same page? If it's something you could reproduce it would be good to figure out where `JSON.stringify(oldParameters) === JSON.stringify(newParameters)` gets triggered (via `debugger;` for instance). Not doing anything in this case might result in nothing happening when the user tries to reload the current page (ie, user is on Visitors Overview, then clicks Visitors Overview to reload the widgets).

sgiehl: @diosmosis It only happens on the admin page and only with disabled developer mode (merged assets).

diosmosis: @sgiehl fixed it here: https://github.com/matomo-org/matomo/pull/18672

It happens w/ developer mode enabled too.

PR #18679, opened 2022-01-23
[Github Action] Test 4.x branch

### LINKED ISSUE #20179: Migrate from Travis CI to Github Actions to run our CI
as per title

work done in https://github.com/matomo-org/matomo/pull/18679

## Steps

- Update Global action to deal with Missing dependencies in some builds
- Update console command to generate the file. So that it's easy to update the Travis files. So you can create github action file by using a command. so it's easy to push the github actions to all available plugins without manually coping the github action files 
- check if Github builds at least as fast as travis? and report the runtime here to compare
- we make it work for Matomo core 
  - and also for the other repositories under the team
  - and we provide basic instructions to colleagues on how to convert a plugin repo to github
- Update the guides that mention Travis to replace with Github and fix appropriately -> Delete any outdated content and write a very basic guide on how Github works
  - https://developer.matomo.org/guides/tests-travis-extended
  - https://developer.matomo.org/guides/tests-travis
- Document how to handle dependencies in a plugin
  - eg. how would it work for LDAP plugin for example "[LoginLdap](https://github.com/matomo-org/plugin-LoginLdap), for example, tests itself against a live LDAP server and thus needs to install and setup OpenLDAP on travis. To accomplish this, [LoginLdap](https://github.com/matomo-org/plugin-LoginLdap) adds extra steps to its generated .travis.yml file." -> how would we do this on github action? Yes, with a local action it's possible.

Notes:

- we don't fix random failing tests for now
- we do fix any tests that  consistently fail

peterhashair:
### Description:
Fork https://github.com/matomo-org/matomo/pull/18290
update github-action branch has some unnecessary UI updates Test 4.x branch, not sure why but seems like php8.0 MySQLi driver has a poor performance here 

UI viewer test:
  Because the silex is out of sup. Here is the Symfony version of it, but at the moment only tested the GitHub action part not fully tested Travis. https://github.com/matomo-org/ui-tests-viewer/pull/40
  
  
 YML, PDF: 
 when tests failed, the system will upload the artifact as zip and stay on Github for 5 days
 
 Once this is merged to 4.x-dev, we can use the below to wait for the other process finished. Only works on default branch
 ```yml
   workflow_run:
    workflows: [ Build Vue files,PHPCS check ]
    types:
      - completed
 ```
 
 
 Performance
 
 It seems like the `MYSQLI` is twice slower as the `PDO` driver, not sure why. 

### Review


core/Db/Schema/Mysql.php
```diff
@@ -191,7 +191,7 @@ public function getTablesCreateSql()
                                 PRIMARY KEY(idvisit),
                                 INDEX index_idsite_config_datetime (idsite, config_id, visit_last_action_time),
                                 INDEX index_idsite_datetime (idsite, visit_last_action_time),
-                                INDEX index_idsite_idvisitor (idsite, idvisitor, visit_last_action_time DESC)
```
sgiehl: That shouldn't be changed

core/Updates/4.9.0-b1.php
```diff
@@ -18,9 +18,9 @@
 use Piwik\Updates as PiwikUpdates;
 
 /**
- * Update for version 5.0.0-b1
+ * Update for version 4.9.0-b1
  */
-class Updates_5_0_0_b1 extends PiwikUpdates
+class Updates_4_9_0_b1 extends PiwikUpdates
```
sgiehl: What's the reason for renaming that one?

plugins/SitesManager/stylesheets/SitesManager.less
```diff
@@ -122,11 +122,6 @@ td.editable-site-field:hover {
 }
 
 .site-without-data {
-    .card-title:after {
```
sgiehl: Seems you have included the changes from another unmerged PR already.

plugins/UserId/tests/Fixtures/TrackFewVisitsAndCreateUsers.php
```diff
@@ -8,15 +8,14 @@
...
-use Piwik\Plugins\UserId\API;
 use Piwik\Tests\Framework\Fixture;
 
 /**
  * Generates visits with user IDs and creates the user IDs index for testing
  */
 class TrackFewVisitsAndCreateUsers extends Fixture
 {
-    public $dateTime = '2010-02-01 11:22:33';
+    public $dateTime = '2010-03-01 11:22:33';
```
sgiehl: Any reason for changing the fixture?

tests/PHPUnit/Integration/ReleaseCheckListTest.php
```diff
@@ -50,6 +50,9 @@ public function test_TestCaseHasSetGroupsMethod()
 
     public function test_woff2_fileIsUpToDate()
     {
+        if (getenv('GITHUB')) {
+            $this->markTestSkipped('this needs review');
```
sgiehl: Did you check why that one is still skipped?

sgiehl (changes requested):
Had a rough look through the code to check if there are still some unexpected changes. Left a couple of comments...

core/Updates/4.9.0-b1.php
peterhashair: that's wrong change should revert :)

plugins/UserId/tests/Fixtures/TrackFewVisitsAndCreateUsers.php
peterhashair: That shouldn't be changed

peterhashair:
@sgiehl revert some unnecessary code form change. that should make review easier.

peterhashair:
### advantage: 

-  The last commit will be canceled the preview build.
-  It seems like Github action overall will be 2-10 mins quicker. 

### disadvantage:
- Entire Pipeline is running as a whole, you can not rerun a single job.
- It seems php8.0 + MySQLI driver is very slow. not sure why could be sth to do with MariaDB. 
- Github Action is free on the public repo, but the limit is 20 concurrency jobs per repo. Meaning the max pipeline at the same time are 3 PR requests. Fork will remove the limit if we want to go this path. 
- The main test should wait for VUE/PHPCS build finished, then continue. But that option only works on the default branch, which is `4.x-dev`. Or we could merge VUE, PHPCS build and test to one pipeline as a step 



### Push Live Require.  Just a list for this to progress. 


- [ ] [Review] Code review on this branch. 
- [ ] [Decision] do we need Travis and Github action running at the same time, or completely remove Travis
     Suggestion: We keep both running for at least 1 week.
- [ ] [Review] Submodule PR review. Submodule target change to 4. x-dev
- [ ] [Decision] do we allow UI upload for public fork repo. Do we consider displaying the failed UI HTML inside GitHub action?  
    * the benefit of fork repo is not limited by concurrency.
      Suggest: we move screen compared to GitHub action, instead of using https://builds-artifacts.matomo.org/

- [ ] Doc] Documentation for development guild
- [ ] [Doc] Documentation for internal staff use.

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,270 @@
...
+  pull_request:
+
+permissions:
+  actions: read
+  checks: none
+  contents: read
+  deployments: none
+  issues: write
+  packages: none
+  pull-requests: write
```
sgiehl: Does the action do anything besides running tests? I wonder why `issues` and `pull-requests` should require a `write` permission

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,270 @@
...
+        image: redis:5.0
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+      mariadb:
+        image: mariadb:latest
```
sgiehl: So all our tests would then run on mariadb? I'm not sure if that is a way we should go. Actually we suggest to use MySQL. Even if many people might use MariaDB instead, we should still be sure everything runs correctly with MySQL. So might be better to keep MySQL for tests

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,270 @@
...
+
+      - name: Test running
+        run: |
+          ./vendor/phpunit/phpunit/phpunit --configuration ./tests/PHPUnit/phpunit.xml --testsuite ${{ matrix.tests }}
+        env:
+          TRAVIS: true
+          GITHUB: true
+          PHP_VERSION: ${{ matrix.php }}
+          MYSQL_ADAPTER: ${{ matrix.adapter }}
+  Javascript:
```
sgiehl, line 63: Any benefits in having all sort of tests in one action with mutliple jobs? If we split them in multiple action, it would be possible to rerun specific ones

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,270 @@
...
+          submodules: recursive
+
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 8.0
+          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, pdo_sqlite, gd, redis
+          ini-values: post_max_size=256M, max_execution_time=10000, memory_limit=4024M
+          tools: composer:v2
```
sgiehl: I guess for javascript tests the extensions, ini-values and tools are acutually not needed

tests/PHPUnit/Integration/SqlTest.php
```diff
@@ -43,6 +43,10 @@ public function testOptimize()
         $this->assertTrue(Db::optimizeTables(array('table1', 'table2', 'table3', 'table4')) !== false);
 
         // make sure innodb tables are skipped
-        $this->assertTrue(Db::optimizeTables(array('table3', 'table4')) === false);
+        if (DB::isOptimizeInnoDBSupported()) {
+            $this->assertTrue(Db::optimizeTables(array('table3', 'table4')) !== false);
+        } else {
+            $this->assertTrue(Db::optimizeTables(array('table3', 'table4')) === false);
+        }
```
sgiehl: Guess using the proper assertion methods would be better:
```suggestion
        if (DB::isOptimizeInnoDBSupported()) {
            $this->assertTrue(Db::optimizeTables(array('table3', 'table4')));
        } else {
            $this->assertFalse(Db::optimizeTables(array('table3', 'table4')));
        }
```

tests/UI/config.dist.js
```diff
@@ -19,7 +19,7 @@ exports.piwikUrl = "http://localhost/";
 exports.phpServer = {
     HTTP_HOST: 'localhost',
     REQUEST_URI: '/',
-    REMOTE_ADDR: '127.0.0.1'
+    REMOTE_ADDR: 'localhost'
```
sgiehl: What's the reason for changing `REMOTE_ADDR` to `localhost`. Shouldn't that be an IP address? 🤔

sgiehl (changes requested):
Some first review comments. Will continue next week

peterhashair:
@tsteur somehow this test on MySQL running really slow, but MariaDB is a little bit quick. Any idea? [PHP (SystemTestsCore, 8.0, MYSQLI)](https://github.com/matomo-org/matomo/runs/5440499701?check_suite_focus=true#logs)

tsteur:
@peterhashair no idea why but you could either try setting `testdox=true` in the phpunit.xml.dist and then monitor if it's one particular test that takes a while. Or I believe there is also another feature maybe to print how long each test took to execute in phpunit. This would then help you figure out if there's one or few tests that are slow or all of them.

I just checked and testdox gave me this output for example:
image

tests/PHPUnit/Integration/DataAccess/LogAggregatorTest.php
```diff
@@ -148,7 +148,7 @@ public function test_generateQuery_withSegment_shouldUseTmpTableWhenEnabled()
 
     public function testSetMaxExecutionTimeOfArchivingQueries()
     {
-        if (SystemTestCase::isMysqli()) {
+        if (SystemTestCase::isMysqli() || getenv('GITHUB')) {
```
sgiehl: This shouldn't be needed anymore when running on MySQL

.github/workflows/javascript-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+    runs-on: ubuntu-20.04
+    services:
+      mysql:
+        image: mysql:5.7
+        ports:
+          - 3306:3306
+        env:
+          MYSQL_DATABASE: piwik_tests
+          MYSQL_ALLOW_EMPTY_PASSWORD: yes
+        options: --tmpfs /var/lib/mysql:rw
```
sgiehl: Is mysql needed for javascript tests? If not we could maybe check if leaving it away speeds up the running time maybe.

.github/workflows/ui-tests.yml
```diff
@@ -0,0 +1,100 @@
...
+          - 3306:3306
+        env:
+          MYSQL_DATABASE: matomo_tests
+          MYSQL_ALLOW_EMPTY_PASSWORD: yes
+        options: --tmpfs /var/lib/mysql:rw --health-cmd="mysqladmin ping" --health-interval=5s --health-timeout=2s --health-retries=3
+
+    strategy:
+      fail-fast: false
+      matrix:
+        tests: [ 'UITests' ]
```
sgiehl: This line should be unneeded.

tests/PHPUnit/Integration/FrontControllerTest.php
```diff
@@ -48,8 +48,13 @@ public function test_thrownExceptionInFrontControllerPrintsBacktrace()
...
+test message on {includePath}/tests/resources/trigger-fatal-exception.php(23) #0 [internal function]: {closure}('CoreHome', 'index', Array) #1 {includePath}/core/EventDispatcher.php(141): call_user_func_array(Object(Closure), Array) #2 {includePath}/core/Piwik.php(845): Piwik\EventDispatcher-&gt;postEvent('Request.dispatc...', Array, false, Array) #3 {includePath}/core/FrontController.php(606): Piwik\Piwik::postEvent('Request.dispatc...', Array) #4 {includePath}/core/FrontController.php(168): Piwik\FrontController-&gt;doDispatch('CoreHome', 'index', Array) #5 {includePath}/tests/resources/trigger-fatal-exception.php(31): Piwik\FrontController-&gt;dispatch('CoreHome', 'index') #6 {main}
 FORMAT;
+        if (getenv('GITHUB') && getenv('PHP_VERSION') != '7.2') {
```
sgiehl: This check might be a bit error prone. I guess the expected format below is meant for PHP 8 and maybe 8.1.
If it doesn't match for PHP 7.3 or PHP 7.4, the check should be improved, as it might happen that we drop support for PHP 7.2 and maybe need to switch to PHP 7.4

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,104 @@
...
+        env:
+          PHP_VERSION: ${{ matrix.php }}
+          MATOMO_TEST_TARGET: 'PHP'
+          MYSQL_ADAPTER: ${{ matrix.adapter }}
+
+      - name: Test running
+        run: |
+          ./vendor/phpunit/phpunit/phpunit --configuration ./tests/PHPUnit/phpunit.xml --testsuite ${{ matrix.tests }}
+        env:
+          TRAVIS: true
```
sgiehl: Is there a reason for setting `TRAVIS` env variable?

.github/workflows/javascript-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+jobs:
+  Javascript:
+    runs-on: ubuntu-20.04
+    services:
+      mysql:
+        image: mysql:5.7
+        ports:
+          - 3306:3306
+        env:
+          MYSQL_DATABASE: piwik_tests
```
sgiehl: we should maybe use `matomo_tests` here to get rid of the old name somewhen.

.github/workflows/javascript-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Javascript Tests
+        if: matrix.tests == 'JavascriptTests'
+        run: |
+          ./console tests:run-js --matomo-url='http://localhost'
+      - name: Running angularjs && vue tests
+        if: matrix.tests == 'AngularJSTests'
+        working-directory: ./tests/angularjs
+        run: |
+          npm install
+          ./node_modules/karma/bin/karma start karma.conf.js --browsers ChromeHeadless --single-run
```
sgiehl: Running the AngularJS tests might not need PHP to be installed at all. We should consider splitting JS & Angular into two actions, so we only set up what is required

.github/artifacts/config.ini.github.php
```diff
@@ -0,0 +1,32 @@
+; <?php exit; ?> DO NOT REMOVE THIS LINE
+; This configuration is used for automatic integration
+; tests on Travis-CI. Do not use this in production.
```
sgiehl: This i now longer Travis-CI 🙈

sgiehl (changes requested):
I left a couple of comments. 
In general I was wondering if it wouldn't make sense to move all parts that sets up the git action container for any test to https://github.com/matomo-org/github-action-tests and use that one also for core tests. That way we don't need everything twice.

peterhashair:
@sgiehl move all tests to here https://github.com/matomo-org/github-action-tests  makes more sense, but needs a little work, start working on it.

peterhashair:
There is one random PHP test that failed remain, JS test won't end.

peterhashair:
@sgiehl update some scripts, haven't grouped them into a GitHub Action package yet. But I got 2 questions,

-  it seems like js got extra 3 tests, do you know any settings that skip those 3 tests

- It seems the PDF has been updated, do we upload expected pdf to the  https://builds-artifacts.matomo.org/ as well?

peterhashair:
ping @sgiehl I think PHP 8 is fully passed, but PHP 8.1 got an error and failed on my local as well, not sure if want to use php8.1 for the testing.

justinvelluppillai:
@peterhashair we would want to use php8.1, if not already then soon. Be worth spending an hour or so understanding what's happening in PHP8.1

.github/workflows/buildtrackerjs.yml
```diff
@@ -1,8 +1,9 @@
 name: Build Tracker JS files
 
 on:
-  issue_comment:
-    types: [created]
+  push:
+    paths:
+      - 'js/piwik.js'
```
sgiehl: I guess it shouldn't have been part of this issue to change the behavior of the action to build js files.

plugins/TestRunner/Commands/SyncScreenshots.php
```diff
@@ -60,30 +63,31 @@ protected function configure()
...
         $repository = $input->getOption('repository');
         $httpUser = $input->getOption('http-user');
         $httpPassword = $input->getOption('http-password');
 
-        $screenshots = $this->getScreenshotList($repository, $buildNumber, $httpUser, $httpPassword);
+        $screenshots = $this->getScreenshotList($repository, $buildNumber, $httpUser, $httpPassword, $agent);
 
         $this->logger->notice('Downloading {number} screenshots', array('number' => count($screenshots)));
         foreach ($screenshots as $name => $url) {
-            foreach ($screenshotsRegex as $regex) {
```
sgiehl, line 72: Is there any good reason for removing the possibility to provide multiple regexes here?

tests/PHPUnit/Integration/FrontControllerTest.php
```diff
@@ -48,11 +48,15 @@ public function test_thrownExceptionInFrontControllerPrintsBacktrace()
...
+test message on {includePath}/tests/resources/trigger-fatal-exception.php(23) #0 [internal function]: {closure}('CoreHome', 'index', Array) #1 {includePath}/core/EventDispatcher.php(141): call_user_func_array(Object(Closure), Array) #2 {includePath}/core/Piwik.php(845): Piwik\EventDispatcher-&gt;postEvent('Request.dispatc...', Array, false, Array) #3 {includePath}/core/FrontController.php(606): Piwik\Piwik::postEvent('Request.dispatc...', Array) #4 {includePath}/core/FrontController.php(168): Piwik\FrontController-&gt;doDispatch('CoreHome', 'index', Array) #5 {includePath}/tests/resources/trigger-fatal-exception.php(31): Piwik\FrontController-&gt;dispatch('CoreHome', 'index') #6 {main}
 FORMAT;
+
         //remove all the numbers
         $expectedFormat = preg_replace('/[0-9]+/', 'x', $expectedFormat);
+        $expectedFormat = preg_replace('/".*?"|\'.*?\'/', 'xxx', $expectedFormat);
```
sgiehl, line 56: What is this needed for?

tests/PHPUnit/Integration/ServeStaticFileTest.php
```diff
@@ -336,6 +336,7 @@ public function test_compressionCache()
         curl_setopt($curlHandle, CURLOPT_RETURNTRANSFER, true);
         curl_exec($curlHandle);
         curl_close($curlHandle);
+        sleep(1);
```
sgiehl: Do you have a good reason for slowing down the tests by waiting a second here?

sgiehl (changes requested):
Had a very rough look through the changes and left some comments.
Also didn't we discuss to use a custom action for running tests. Guess that was why you created https://github.com/matomo-org/github-action-tests for, right? But why isn't it in use here?

peterhashair:
@sgiehl yes, should use  https://github.com/matomo-org/github-action-tests, just trying to pass all the tests first, then I can covert them to a package.

plugins/TestRunner/Commands/SyncScreenshots.php
peterhashair: @sgiehl ops, that's me, revert it

tests/PHPUnit/Integration/ServeStaticFileTest.php
peterhashair: @sgiehl was failed, but figure it out, it was caused by else, revert it

tests/PHPUnit/Integration/FrontControllerTest.php
peterhashair: @sgiehl Github Action adds some extra on there. will revert it, and see if it works.

sgiehl:
> @sgiehl yes, should use https://github.com/matomo-org/github-action-tests, just trying to pass all the tests first, then I can covert them to a package.

So if you know that your work isn't finished yet, why exactly are you requesting a review?? I'm happy to also have a first look in between to e.g. check if something goes into the right direction. But in that case you need to leave a comment so I'm aware of the current state and what exactly to look at. Otherwise this is a total waste of my time.

peterhashair:
@sgiehl sorry about that, just trying to check, if all the test performance is correctly, before converting to package.

peterhashair:
Updates, group them into a GitHub package now. Still, some tests fail to try to fix. Seems like GitHub Action provides only run failed jobs options, now all the tests are in one file.

peterhashair:
@sgiehl Group them into the package, maybe can give a init review, I believe is just the angular keep failed.

plugins/PrivacyManager/tests/System/AnonymizationTest.php
```diff
@@ -42,7 +42,7 @@ public function setUp(): void
      */
     public function testApi($api, $params)
     {
-        $params['xmlFieldsToRemove'] = array();
+        $params['xmlFieldsToRemove'] = array('userId');
```
sgiehl: Whats that needed for?

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,173 @@
...
+          - php: '8.1'
+            adapter: 'PDO_MYSQL'
+          - php: '7.2'
+            adapter: 'MYSQLI'
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          lfs: false
+          submodules: recursive
```
sgiehl: We should do the checkout in the github action directly. Passing through the sha/branch/tag should make it possible to use the version, that should be used for testing. This will also be required for testing plugins. We used to test our plugin against various Matomo versions. In most cases, this is the minimum require version as well as the latest 4.x-dev branch

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,173 @@
...
+        ports:
+          - 26379:26379
+      mysql:
+        image: mysql:5.7
+        ports:
+          - 3306:3306
+        env:
+          MYSQL_DATABASE: matomo_tests
+          MYSQL_ALLOW_EMPTY_PASSWORD: yes
+        options: --tmpfs /var/lib/mysql:rw
```
sgiehl: should we define the required services in the github action as well? Only services that might differ from the default should be defined.

For plugins it would in general be awesome to have the action in each plugin as small as possible. That way we won't need to change every plugins action and only need to adjust our global test action if something needs to be adjusted.

sgiehl (changes requested):
Left some more thoughts here. Also it would be awesome if you also could at least adjust one PR in a plugin to use the new global action, so we can approve it works there as well.

plugins/PrivacyManager/tests/System/AnonymizationTest.php
peterhashair: @sgiehl Github Action somehow put userID as a random UUID.

.github/workflows/matomo-tests.yml
peterhashair: @sgiehl this may be a little bit tricky because Github Action composite does not accept containers. But I guess we can manually build MySQL and reds in ubuntu, instead of using docker.

peterhashair:
@sgiehl convert the mysql, redis etc to the package, now should be a simple setup. Convert one plugin as a test, here https://github.com/matomo-org/plugin-TasksTimetable/pull/29

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,108 @@
+name: Matomo Tests
+
+on:
+  pull_request:
```
sgiehl, line 9: I don't think, that our tests should in the future only run for pull requests. They should also run for every commit on any base branch (4.x-dev, 5.x-dev) and maybe even manually.
Allowing that will make some definitions below incorrect, as there will not always be a reference like `${{ github.event.pull_request.head.sha }}`

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,108 @@
...
+        uses: matomo-org/github-action-tests@v1
+        with:
+          test-type: ${{ matrix.tests }}
+          php-version: ${{ matrix.php }}
+          test-command: ${{ matrix.parts }}
+          run_number: ${{ github.run_number }}
+          node-version: ${{ matrix.node }}
+          repository: ${{ github.repository }}
+          artifect_pass: ${{ secrets.ARTIFACTS_PASS }}
+          test-target: ${{ github.event.pull_request.base.ref }}
```
sgiehl: Wouldn't that run tests against the target branch?

sgiehl (changes requested):
@peterhashair would you mind cleaning up your stuff and checking if everything actually works as expected before putting something into review? You should have already been able to identify a couple of issues only by looking through the github action logs:
* https://github.com/matomo-org/matomo/runs/8105848809?check_suite_focus=true#step:2:281
* https://github.com/matomo-org/matomo/runs/8105848809?check_suite_focus=true#step:2:340
* https://github.com/matomo-org/matomo/runs/8105848809?check_suite_focus=true#step:2:341

peterhashair:
@sgiehl fixed the error, but I am confused about that one. Do you mind giving an example? 

> I don't think, that our tests should in the future only run for pull requests. They should also run for every commit on any base branch (4.x-dev, 5.x-dev) and maybe even manually.

sgiehl:
> I don't think, that our tests should in the future only run for pull requests. They should also run for every commit on any base branch (4.x-dev, 5.x-dev) and maybe even manually.

We need to make sure that not only pull request have test running, but also our main branches after each commit/merge. Otherwise we can't be sure that tests would still be passing after something is merged.

Guess it would need to be something like
```
on:
  pull_request:
    types: [opened, synchronize]
  push:
    branches:
      - '**.x-dev'
      - 'next_release'
```

peterhashair:
@justinvelluppillai both Travis and GitHub action is enabled, still, some random PHP failed as expected, also for the UI test, I think we can only use one instead of the other, it seems like the 2 system has some IP or fonts difference, make the UI unique. 

The other funny part seems like the php8.1 + MySQLI driver is significantly slower than php7.2 + PDO driver, not too sure why.

peterhashair:
As we discussed, we can merge with Travis, temp disables GitHub action UI tests, once we confirm the GitHub action is solid, then we enable the UI tests again.

justinvelluppillai:
That sounds good to me @peterhashair as long as we have no security issues and can run them concurrently.

peterhashair:
@justinvelluppillai I don't think there is a security issue there, but if someone can review this would be good, still a coupe random fails, but I think that's expected.

justinvelluppillai:
@bx80 are you at all familiar with this work to be able to give it a final review?

bx80:
@justinvelluppillai, @peterhashair I can take a look :+1:

peterhashair:
@bx80 all the Gtihub Action codes tests code actually are here: https://github.com/matomo-org/github-action-tests.

sgiehl:
@peterhashair For the action repo it might be helpful to create a temporary branch and create a PR from main against that branch. It's a lot easier to do a review, if there is a PR that can be commented.

peterhashair:
> @peterhashair For the action repo it might be helpful to create a temporary branch and create a PR from main against that branch. It's a lot easier to do a review, if there is a PR that can be commented.

refer https://github.com/matomo-org/matomo/pull/19843

sgiehl:
> > @peterhashair For the action repo it might be helpful to create a temporary branch and create a PR from main against that branch. It's a lot easier to do a review, if there is a PR that can be commented.
> 
> refer #19843

I guess you misunderstood that. I meant creating a PR on that repo: https://github.com/matomo-org/github-action-tests.
That repo contains the most relevant code and needs a detailed review...

sgiehl:
This PR needs to be reviewed together with https://github.com/matomo-org/github-action-tests/pull/2 and one of the plugin PRs like https://github.com/matomo-org/plugin-TasksTimetable/pull/29
Only if all work smoothly together this one can be merged.

peterhashair:
@sgiehl do we need anything to get this merged?

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,115 @@
...
+  contents: read
+  deployments: none
+  issues: read
+  packages: none
+  pull-requests: read
+  repository-projects: none
+  security-events: none
+  statuses: none
+concurrency:
+  group: php-${{ github.head_ref }}
```
sgiehl: `github.head_ref` is only set when the workflow is triggered by a pull request. If it's triggered by a push or manually for a branch, this will be empty.
Not sure if we should use `github.ref` instead. That would be set for all cases, but for pull request it would target to the merge branch instead of the branch itself.

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,115 @@
...
+    steps:
+      - name: running tests
+        uses: matomo-org/github-action-tests@v1
+        with:
+          test-type: ${{ matrix.type }}
+          mysql-driver: ${{ matrix.adapter }}
+          php-version: ${{ matrix.php }}
+          test-command: ${{ matrix.command }}
+          redis-service: true
+          git-head: ${{ github.head_ref }}
```
sgiehl: as mentioned above `github.head_ref` might not always be defined

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,115 @@
...
+        tests: [ 'JS' ]
+        node: [ '12' ]
+    steps:
+      - name: running tests
+        uses: matomo-org/github-action-tests@v1
+        with:
+          test-type: ${{ matrix.tests }}
+          php-version: ${{ matrix.php }}
+          node-version: ${{ matrix.node }}
+          git-head: ${{ github.head_ref }}
```
sgiehl: same here

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,115 @@
...
+        node: [ '16' ]
+    steps:
+      - name: running tests
+        uses: matomo-org/github-action-tests@v1
+        with:
+          test-type: ${{ matrix.type }}
+          php-version: ${{ matrix.php }}
+          test-command: ${{ matrix.tests }}
+          mysql-service: false
+          git-head: ${{ github.head_ref }}
```
sgiehl: and here

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,115 @@
...
+#          php-version: ${{ matrix.php }}
+#          test-command: ${{ matrix.parts }}
+#          run-number: ${{ github.run_number }}
+#          node-version: ${{ matrix.node }}
+#          repository: ${{ github.repository }}
+#          artifect_pass: ${{ secrets.ARTIFACTS_PASS }}
+#          run-id: ${{ github.run_id }}
+#          php-memory: "-1"
+#          redis-service: true
+#          git-head: ${{ github.head_ref }}
```
sgiehl: and here

plugins/PrivacyManager/tests/System/AnonymizationTest.php
sgiehl: This should be reverted. There should be no need to adjust the tests. If tests are failing we need to narrow down the exact reason instead of simply hiding something in the tests.

plugins/TestRunner/Commands/SyncScreenshots.php
```diff
@@ -48,42 +50,56 @@ protected function configure()
...
-                            . 'from travis artifacts to the tests/UI/expected-screenshots/ folder');
+            .'from travis artifacts to the tests/UI/expected-screenshots/ folder');
         $this->addArgument('buildnumber', InputArgument::REQUIRED, 'Travis build number you want to sync.');
-        $this->addArgument('screenshotsRegex', InputArgument::OPTIONAL | InputArgument::IS_ARRAY,
-            'A regex to use when selecting screenshots to copy. If not supplied all screenshots are copied.', ['.*']);
-        $this->addOption('repository', 'r', InputOption::VALUE_OPTIONAL, 'Repository name you want to sync screenshots for.', 'matomo-org/matomo');
-        $this->addOption('http-user', '', InputOption::VALUE_OPTIONAL, 'the HTTP AUTH username (for premium plugins where artifacts are protected)');
-        $this->addOption('http-password', '', InputOption::VALUE_OPTIONAL, 'the HTTP AUTH password (for premium plugins where artifacts are protected)');
+        $this->addOption('agent', 'a', InputOption::VALUE_OPTIONAL, 'Build agent using you want to choose');
+        $this->addArgument('screenshotsRegex', InputArgument::OPTIONAL,
```
sgiehl: Any reason why you removed the ` | InputArgument::IS_ARRAY`?

tests/PHPUnit/Integration/ReleaseCheckListTest.php
```diff
@@ -48,11 +48,22 @@ public function test_TestCaseHasSetGroupsMethod()
...
+
+
+        } else {
+            //todo this should be removed once travis retired
+            link(PIWIK_INCLUDE_PATH."/plugins/Morpheus/fonts/matomo.ttf", "temp.ttf");
+            $command = PIWIK_INCLUDE_PATH."/../travis_woff2/woff2_compress 'temp.ttf'";
+            $log = shell_exec($command);
+            $this->assertFileEquals('temp.woff2', PIWIK_INCLUDE_PATH."/plugins/Morpheus/fonts/matomo.woff2",
+                "woff2 file is out of date.\nCommand output:\n".$log);
+        }
```
sgiehl: Note: This part will be removed once #20081 has been merged.

.github/workflows/matomo-tests.yml
```diff
@@ -0,0 +1,115 @@
...
+  group: php-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  PHP:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        command: [ 'UnitTests', 'SystemTestsPlugins' ,'SystemTestsCore' , 'IntegrationTestsCore', 'IntegrationTestsPlugins' ]
```
sgiehl: What we will loose when using this one is our "AllTests" suite. I agree that normally all test suites above should cover all tests, but we had it a couple of times in the past, that some tests did not have the correct groups set and therefor did not run in any of the above groups.
It might be fine to only use the groups in the future, but for that we need to ensure, that all tests are executed. So we might need to rethink how our tests are structured

sgiehl (dismissed):
Some first comments here. Further review will be done in the github action

plugins/TestRunner/Commands/SyncScreenshots.php
peterhashair: @sgiehl that should be an accident. revert it

.github/workflows/matomo-tests.yml
peterhashair: @sgiehl good point will implement AllTests.

peterhashair: @sgiehl I had a look at All tests, and it seems like is the same thing with the split tests.

```bash
    elif [ "$TEST_SUITE" = "AllTests" ]
    then
        exit_code=0

        if [ "$ALLTEST_EXTRA_OPTIONS" = "--run-first-half-only" ]
        then
            echo "Executing tests in test suite UnitTests"
            travis_wait ./../../vendor/phpunit/phpunit/phpunit --configuration phpunit.xml --testsuite UnitTests --colors $PHPUNIT_EXTRA_OPTIONS || exit_code=$?
            echo "Executing tests in test suite SystemTests for Plugins"
            travis_wait ./../../vendor/phpunit/phpunit/phpunit --configuration phpunit.xml --testsuite SystemTestsPlugins --colors $PHPUNIT_EXTRA_OPTIONS || exit_code=$?
            echo "Executing tests in test suite SystemTests for Core"
            travis_wait ./../../vendor/phpunit/phpunit/phpunit --configuration phpunit.xml --testsuite SystemTestsCore --colors $PHPUNIT_EXTRA_OPTIONS || exit_code=$?
        elif [ "$ALLTEST_EXTRA_OPTIONS" = "--run-second-half-only" ]
        then
            echo "Executing tests in test suite IntegrationTests"
            travis_wait ./../../vendor/phpunit/phpunit/phpunit --configuration phpunit.xml --testsuite IntegrationTests --colors $PHPUNIT_EXTRA_OPTIONS || exit_code=$?
        else
```

sgiehl: Seems you are right. Guess no need to add it then.

plugins/CustomDimensions/tests/System/ApiTest.php
```diff
@@ -264,6 +264,10 @@ public function getApiForTesting()
 
     public function testScheduledReport()
     {
+        //todo this should be removed once github action is alive.
+        if (!getenv('GITHUB')) {
+            $this->markTestSkipped("Github and Travis has some fonts diff, please see Github Action tests");
+        }
```
sgiehl: Might be better to check with `Fixture::canImagesBeIncludedInScheduledReports()` if the test should be run or not.
I guess the reason for the differing result might be a different gd lib version. The method above should check that accordingly.

sgiehl:
As I will be away the next days here a small update on the current state of the github test action.
I have refactored and simplified various parts of the action and added some missing parts that are required for running tests for plugins.
In core, as well as for public and private plugins the action seems to run more or less correctly. There are still a couple of failures that I need to investigate. Once everything works as expected I will start implementing a command to generate the action files for plugins. So we can finally start rolling out the action to all repos.
Will continue to work on it when I'm back.

sgiehl:
Build finally succeeded. Even though there are still some randomly failing tests like before.

PR #18691, opened 2022-01-26
[Performance]Cache whether a DB supports transaction level

### LINKED ISSUE #18138: Cache whether a DB supports transaction level or not
We noticed on Cloud that https://github.com/matomo-org/matomo/blob/4.x-dev/core/Db/TransactionLevel.php#L57 might be called very often. Especially when having many funnels but also in general. It can cause a bit of load. It be better to instead cache this information whether uncommitted can be set or not instead of retrying for support every time. We should cache this eg on the `$this->db` as it's DB dependent as if someone configures different DBs for writer and reader for example then we cannot reuse / cache the same result.

I'm meaning something in this direction (not tested): 

```
diff --git a/core/Db/TransactionLevel.php b/core/Db/TransactionLevel.php
index c19699084..fd730ced9 100644
--- a/core/Db/TransactionLevel.php
+++ b/core/Db/TransactionLevel.php
@@ -40,6 +40,10 @@ class TransactionLevel
 
     public function setUncommitted()
     {
+        if ($this->db->supportsUncommitted === false) {
+            return false;
+        }
+
         try {
             $backup = $this->db->fetchOne('SELECT @@TX_ISOLATION');
         } catch (\Exception $e) {
@@ -54,8 +58,13 @@ class TransactionLevel
             $this->db->query('SET SESSION TRANSACTION ISOLATION LEVEL READ UNCOMMITTED');
             $this->statusBackup = $backup;
 
-            Option::set(self::TEST_OPTION_NAME, '1'); // try setting something w/ the new transaction isolation level
+            if (!isset($this->db->supportsUncommitted)) {
+                Option::set(self::TEST_OPTION_NAME, '1'); // try setting something w/ the new transaction isolation level
+
+                $this->db->supportsUncommitted = true;
+            }
         } catch (\Exception $e) {
+            $this->db->supportsUncommitted = false;
             // catch eg 1665 Cannot execute statement: impossible to write to binary log since BINLOG_FORMAT = STATEMENT and at least one table uses a storage engine limited to row-based logging. InnoDB is limited to row-logging when transaction isolation level is READ COMMITTED or READ UNCOMMITTED
             $this->restorePreviousStatus();
             return false;
diff --git a/plugins/Cloud b/plugins/Clo
```

peterhashair:
### Description:
Fixes: #18138 
Cache whether a DB supports transaction on all the DB adapter to improve perfomance. 
### Review


tests/PHPUnit/Integration/Db/TransactionLevelTest.php
```diff
@@ -44,16 +44,23 @@ public function test_canLikelySetTransactionLevel()
 
 	public function test_setUncommitted_restorePreviousStatus()
 	{
-		$value = $this->db->fetchOne('SELECT @@TX_ISOLATION');
+        $version = $this->db->fetchOne('SELECT VERSION()');
+        if (preg_match("/8[.].*/", $version)) {
```
sgiehl: wouldn't that also match `5.8.0` or even something like `5.7.36-0ubuntu0.18.04.1`? 
Maybe it would be easier to run something like `SHOW GLOBAL VARIABLES LIKE 't%_isolation'` to check if a variable is supported?

sgiehl (changes requested):
Code looks fine. Left a comment for improving the tests

sgiehl (approved):
Looks good to me.
@tsteur do you want to have a last quick view before this is merged?

core/Db/TransactionLevel.php
```diff
@@ -54,8 +60,12 @@ public function setUncommitted()
             $this->db->query('SET SESSION TRANSACTION ISOLATION LEVEL READ UNCOMMITTED');
             $this->statusBackup = $backup;
 
-            Option::set(self::TEST_OPTION_NAME, '1'); // try setting something w/ the new transaction isolation level
+            if (!isset($this->supportsUncommitted)) {
```
tsteur: not sure this `if` is needed?

tsteur (changes requested):
In the original issue in https://github.com/matomo-org/matomo/issues/18138 I mention this:

> We should cache this eg on the $this->db as it's DB dependent as if someone configures different DBs for writer and reader for example then we cannot reuse / cache the same result.

@peterhashair @sgiehl the current solution wouldn't 100% work for us.

We want to cache it per DB (in case writer and reader are somehow different) and also to have better cache when many new instances of this class are created. Otherwise it be only cached in the instance itself. We are creating this class very often so it wouldn't really help.

peterhashair:
@tsteur right, so this `$supportsUncommitted` should be in the `Db.php`.

core/Tracker/Db.php
```diff
@@ -30,6 +30,9 @@ abstract class Db
 
     protected $connection = null;
 
+    // this is used for indicate TransactionLevel Cache
+    public $supportsUncommitted = null;
```
sgiehl, line 34: Is the TransactionLevel only used with Tracker\Db instances. If not we might also add that new variable to `Piwik\Db` maybe?

tsteur: I would say so it be good to do it for both as the regular DB uses it as well eg when archiving

tsteur: the test below also uses the regular DB

sgiehl:
@peterhashair most tests are now failing with messages like `Undefined property: Piwik\Db\Adapter\Pdo\Mysql::$supportsUncommitted`. Maybe it needs to be added to the adapters as well...

core/Db/Adapter/Pdo/Mysql.php
```diff
@@ -29,6 +29,10 @@ class Mysql extends Zend_Db_Adapter_Pdo_Mysql implements AdapterInterface
      *
      * @param array|Zend_Config $config database configuration
      */
+    
+    // this is used for indicate TransactionLevel Cache
+    public $supportsUncommitted;
```
sgiehl, line 34: This needs to be part of the `MySQLi` adapter as well...

peterhashair:
@sgiehl I think that one is good to go?

sgiehl:
From my point of view this looks good to merge.

core/Db/TransactionLevel.php
```diff
@@ -40,6 +40,10 @@ public function canLikelySetTransactionLevel()
 
     public function setUncommitted()
     {
+        if ($this->db->supportsUncommitted === false) {
```
tsteur, line 43: Not sure if I'm seeing this right. Wouldn't this need to be `if ($this->db->supportsUncommitted !== null) {` because otherwise if `supportsUncommitted = true;` then it wouldn't use the cached result? @peterhashair @sgiehl

tsteur (changes requested):
There might be still an issue. Also can we add a test for the cache by setting the property on the `db` object in a test and then we get expected result or something?

core/Db/TransactionLevel.php
peterhashair: @tsteur right, that make sense, just checking should it be return like that.
```php
        if ($this->db->supportsUncommitted !== null) {
            return $this->db->supportsUncommitted;
        }
```

tsteur: I believe that makes sense. Be good to test it and double check it

core/Db/TransactionLevel.php
```diff
@@ -40,6 +40,10 @@ public function canLikelySetTransactionLevel()
 
     public function setUncommitted()
     {
+        if ($this->db->supportsUncommitted !== null) {
+            return $this->db->supportsUncommitted;
```
sgiehl: If we also "cache" and return if `supportsUncommitted` is `true`, that means we won't perform the setting of the transaction level anymore. That might be fine, but don't we need to unset `supportsUncommitted` then in `restorePreviousStatus`, to ensure it's checked and set again if needed?

sgiehl: Note: This might also cause issues if the transaction level might be changed due to any other reasons...

tsteur: good point, we still want to set the transaction mode, but we don't need to execute the `Option::set` for example to check if it worked.

peterhashair: @tsteur @sgiehl did a little adjustment, would that work

core/Db/TransactionLevel.php
```diff
@@ -40,6 +40,10 @@ public function canLikelySetTransactionLevel()
 
     public function setUncommitted()
     {
+        if ($this->db->supportsUncommitted) {
+            return true;
+        }
```
sgiehl: That doesn't seem to make sense to me. Doesn't that mean, that if setting the transaction level isn't supported, that it will be retried every time it's requested. And on the other side - if it's supported the transaction level won't be set again once it was detected.

peterhashair:
@sgiehl right, sorry wasn't thinking, update again, if that makes sense.

core/Db/TransactionLevel.php
```diff
@@ -54,7 +59,13 @@ public function setUncommitted()
             $this->db->query('SET SESSION TRANSACTION ISOLATION LEVEL READ UNCOMMITTED');
             $this->statusBackup = $backup;
 
-            Option::set(self::TEST_OPTION_NAME, '1'); // try setting something w/ the new transaction isolation level
+            if ($this->db->supportsUncommitted === null) {
```
sgiehl, line 63: This will currently always be true, as the method already returns in the beginning

core/Db/TransactionLevel.php
```diff
@@ -40,12 +40,17 @@ public function canLikelySetTransactionLevel()
 
     public function setUncommitted()
     {
+        if ($this->db->supportsUncommitted !== null) {
+            return $this->db->supportsUncommitted;
+        }
```
sgiehl: We should only skip the rest if `supportsUncommitted` was set to false.

```suggestion
        if ($this->db->supportsUncommitted === false) {
            return false;
        }
```

core/Db/TransactionLevel.php
```diff
@@ -69,6 +80,7 @@ public function restorePreviousStatus()
         if ($this->statusBackup) {
             $value = strtoupper($this->statusBackup);
             $this->statusBackup = null;
+            $this->db->supportsUncommitted = null;
```
sgiehl: We shouldn't reset that one here anymore.

core/Db/TransactionLevel.php
```diff
@@ -54,7 +59,13 @@ public function setUncommitted()
             $this->db->query('SET SESSION TRANSACTION ISOLATION LEVEL READ UNCOMMITTED');
             $this->statusBackup = $backup;
```
sgiehl, line 61: Shouldn't the `statusBackup` only be set if it wasn't set before? Just wondering what would happen when calling the method twice, without calling `restorePreviousStatus` before. In that case the second call would overwrite the `statusBackup` with `READ UNCOMMITTED` and the previously set level gets lost.

sgiehl (changes requested):

core/Db/TransactionLevel.php
peterhashair: @sgiehl i was wonder the same thing. Otherwise the restore will be always skipped. I am right.

core/Db/TransactionLevel.php
```diff
@@ -54,7 +59,13 @@ public function setUncommitted()
...
 
-            Option::set(self::TEST_OPTION_NAME, '1'); // try setting something w/ the new transaction isolation level
+            if ($this->db->supportsUncommitted === null) {
+                // the first time we need to check if the transaction level actually works by
+                // trying to set something w/ the new transaction isolation level
+                Option::set(self::TEST_OPTION_NAME, '1');
+            }
+            $this->db->supportsUncommitted = true;
+
```
tsteur: when it goes in the exception below, then we need to set `$this->db->supportsUncommitted = false;`

core/Db/TransactionLevel.php
tsteur: 👍  absolutely, otherwise we never set the uncommitted status again when it is supported. @peterhashair I had also sent you this in slack

core/Db/TransactionLevel.php
tsteur: This would be a different issue overall maybe and not as part of this. Generally, because we create each time a separate instance of the transaction level it should work and be fine. Like 

```php
$t = new TransactionLevel();
$t->setUncommitted();
....
$x = new TransactionLevel();
$x->setUncommitted();
$x->restorePrevious();
...
$t->restorePrevious();
```

AFAIK `$t` should then restore the previous one again correctly. `$x` would technically not need to try and set the status if `backup` is already `READ UNCOMMITTED`.

core/Db/TransactionLevel.php
```diff
@@ -40,12 +40,18 @@ public function canLikelySetTransactionLevel()
 
     public function setUncommitted()
     {
+        if ($this->db->supportsUncommitted === false) {
+            // we know "Uncommitted" transaction level is not supported, we don't need to do anything as it won't work to set the status
+            return;
```
sgiehl: shouldn't that be to have consistent return types:
```suggestion
            return false;
```

sgiehl (changes requested):
Left a comment. Otherwise guess it looks good now, as the status backup doesn't seem to be an issue.

sgiehl (approved):
Should be good to merge now. @tsteur do you want to have another look before merging?

core/Db/TransactionLevel.php
```diff
@@ -69,6 +83,7 @@ public function restorePreviousStatus()
         if ($this->statusBackup) {
             $value = strtoupper($this->statusBackup);
             $this->statusBackup = null;
+            $this->db->supportsUncommitted = null;
```
tsteur: I'm not understanding why we would reset this one? This would clear the cache every time meaning the cache would effectively never be used?

core/Db.php
```diff
@@ -40,6 +40,8 @@ class Db
 
     private static $logQueries = true;
 
+    // this is used for indicate TransactionLevel Cache
+    public $supportsUncommitted = null;
```
tsteur: not sure this one is actually needed but can leave it there I guess

core/Tracker/Db.php
tsteur: This one still needs to be done. I don't think this is needed in here but rather in `Tracker/Mysqli` and `Tracker/pdo/mysql` DB?

tsteur (changes requested):
@peterhashair @sgiehl I believe there are still few issues. Could we get the these resolved?

core/Tracker/Db.php
sgiehl: @tsteur aren't `Tracker/Mysqli` and `Tracker/PDO/Mysql` inheriting from `Tracker/Db`?

tsteur: They do indeed 👍

tsteur (approved):
I haven't tested it but looks all good to me now 👍

PR #18723, opened 2022-02-02
Make unique index compatible with database without long key support

### LINKED ISSUE #18721: 4.7.0-b2 Update fails with MySQL 5.5
## Expected Behavior
Updating to 4.7.0-b2 via Webupdater shouldnt be a thing

## Current Behavior
Fatal error: 
`matomo/core/Updates/4.7.0-b2.php: Error trying to execute the migration 'ALTER TABLE `matomo_changes` ADD UNIQUE KEY unique_plugin_version_title (`plugin_name`, `version`, `title`);'. The error was: SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes`

## Possible Solution
I guess it works with MySQL 5.7, but as 5.5 is still supported in general, it should work

sgiehl:
### Description:

Not sure if that is the most elegant solution. But can't think of something better right now.

fixes #18721

### Review

### DIFF
--- a/core/Db/Schema/Mysql.php
+++ b/core/Db/Schema/Mysql.php
@@ -362,14 +362,14 @@ class Mysql implements SchemaInterface
             'changes'             => "CREATE TABLE `{$prefixTables}changes` (
                                       `idchange` INTEGER UNSIGNED NOT NULL AUTO_INCREMENT,
                                       `created_time` DATETIME NOT NULL,
-                                      `plugin_name` VARCHAR(40) NOT NULL,
+                                      `plugin_name` VARCHAR(60) NOT NULL,
                                       `version` VARCHAR(20) NOT NULL, 
                                       `title` VARCHAR(255) NOT NULL,                                      
                                       `description` TEXT NULL,
                                       `link_name` VARCHAR(255) NULL,
                                       `link` VARCHAR(255) NULL,       
                                       PRIMARY KEY(`idchange`),
-                                      UNIQUE KEY unique_plugin_version_title (`plugin_name`, `version`, `title`(131))                            
+                                      UNIQUE KEY unique_plugin_version_title (`plugin_name`, `version`, `title`(100))                            
                                   ) ENGINE=$engine DEFAULT CHARSET=$charset
             ",
         );
--- a/core/Db/Schema/Mysql.php
+++ b/core/Db/Schema/Mysql.php
@@ -362,14 +362,14 @@ class Mysql implements SchemaInterface
             'changes'             => "CREATE TABLE `{$prefixTables}changes` (
                                       `idchange` INTEGER UNSIGNED NOT NULL AUTO_INCREMENT,
                                       `created_time` DATETIME NOT NULL,
-                                      `plugin_name` VARCHAR(255) NOT NULL,
+                                      `plugin_name` VARCHAR(40) NOT NULL,
                                       `version` VARCHAR(20) NOT NULL, 
                                       `title` VARCHAR(255) NOT NULL,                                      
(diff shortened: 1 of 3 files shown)


MatomoForumNotifications:

https://forum.matomo.org/t/4-7-0-b2-update-fails-with-mysql-5-5/44603/10

core/Db/Schema/Mysql.php
```diff
@@ -362,14 +362,14 @@ public function getTablesCreateSql()
             'changes'             => "CREATE TABLE `{$prefixTables}changes` (
                                       `idchange` INTEGER UNSIGNED NOT NULL AUTO_INCREMENT,
                                       `created_time` DATETIME NOT NULL,
-                                      `plugin_name` VARCHAR(255) NOT NULL,
+                                      `plugin_name` VARCHAR(40) NOT NULL,
```
tsteur: @sgiehl seems in other tables we have a 60 character limit in the DB like `plugin_setting`, `site_setting`. Does it make sense to apple this here as well? We could create a follow up issue for Matomo 5 (to not break anything) to adjust `isValidPluginName` to limit to max 60 characters. Checked our longest plugin name so far is around 30 characters.

core/Db/Schema/Mysql.php
```diff
@@ -362,14 +362,14 @@ public function getTablesCreateSql()
...
                                       `version` VARCHAR(20) NOT NULL, 
                                       `title` VARCHAR(255) NOT NULL,                                      
                                       `description` TEXT NULL,
                                       `link_name` VARCHAR(255) NULL,
                                       `link` VARCHAR(255) NULL,       
                                       PRIMARY KEY(`idchange`),
-                                      UNIQUE KEY unique_plugin_version_title (`plugin_name`, `version`, `title`)                            
+                                      UNIQUE KEY unique_plugin_version_title (`plugin_name`, `version`, `title`(131))
```
tsteur: Should we make this like 100 then? For a title it's already very long.

tsteur (changes requested):

core/Db/Schema/Mysql.php
sgiehl: the developer docs say the max length should be 40. But we can also use 60. No preference there.

core/Db/Schema/Mysql.php
sgiehl: If we extend the plugin name to `60`, guess we can also lower the title to `100`

core/Db/Schema/Mysql.php
tsteur: OK we can leave it at 40. I will create a follow up issue then to change the other columns to 40 as well.

tsteur: or maybe let's use 60 @sgiehl for consistency with the other tables. I'm sure it be fine to use 40, but better have consistency for now. I created follow up issue for Matomo 5 to decide on a number.

bx80 (approved):
I've tested the update and it works without issues for me. The plugin_name column length is adjusted and the index is recreated correctly.

core/Db/Schema/Mysql.php
bx80: I've updated the plugin_name length to 60

core/Db/Schema/Mysql.php
bx80: I've updated the title index limit to 100

MatomoForumNotifications:

https://forum.matomo.org/t/does-4-71-fix-the-fatal-sql-error-on-upgrade-using-mysql-version-5-5/44656/2

PR #18728, opened 2022-02-02
[Regression] Expand event name table column change the table size. 

### LINKED ISSUE #18699: Label shortened when opening multiple subtables
Not sure if we already have an issue.
* To reproduce, go to https://demo.matomo.cloud/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=day&date=2022-01-27&segment=&category=General_Actions&subcategory=Events_Events
* Click on `Event Names`
* First the table will look like this 
image

* when clicking on a row with a little bit longer title like `added - Diving Boots` and then row `Diving the Ribbon Reefs on Australia's Great Barrier Reef`, then the rows gets truncated and barely any text is readable 
image
* The more subtables are open, the less space is available 
image

peterhashair:
### Description:

Fixes: #18699 

Problem: When expanding the event name table column change the table size.

It seems like the line below,`$('tr:nth-child(1) td.label', domElem).length` is increase by 1 when expand. That makes the table label smaller. Should just first parent label length.
https://github.com/matomo-org/matomo/blob/f870e36a8993f91fcb5b5fd814c3fe6efbfd2690/plugins/CoreHome/javascripts/dataTable.js#L489
### Review


peterhashair:
It seems on the mobile this part won't work as well, should I fix that issue as well @tsteur 

image

tsteur:
@peterhashair which part doesn't work there?

peterhashair:
@tsteur it seems the right side table is cut half on the screen.

sgiehl:
@peterhashair your approach doesn't look right. The reason why the table gets smaller or displays a table fully cut of, is that we have a javascript that tries to determine the expected / required width of the table. It then sets a `max-width` for the table. I guess if the calculation is fired too early the resulting values are incorrect. Maybe this is a regression of the migration of the widget loader to vue. But I guess we need to adjust the javascript

peterhashair:
@sgiehl right that make sense, will update

tsteur:
@peterhashair re the mobile issue: we don't want to create a table mobile view for now

peterhashair:
@sgiehl @tsteur sorry was trying the quick solution on demo.matomo.cloud browser CSS. Is there a way to generate some `event name` or can I copy the demo database to reproduce it locally

tsteur:
see  https://matomo.org/docs/event-tracking/#implementing-event-tracking-with-matomo  on how to track events. I believe this same issue applies though also to other reports, not just events.

I believe the visitor generator might generate events too.

peterhashair:
It seems like the line below, `$('tr:nth-child(1) td.label', domElem).length` is increase by 1 with expand. That makes the table label smaller, removed this divide, not sure if that breaks other parts.
```javascript 
            return parseInt(labelWidth / $('tr:nth-child(1) td.label', domElem).length, 10);
```

sgiehl:
Checked that on the events report. It seems when opening the first subtable the full table increases the width slightly. But opening further subtables don't change the width anymore and the automatically fit into the other tables width.
@peterhashair can you have a quick look if that can be fixed quickly. If it takes longer I would say this should be good enough as it is.

peterhashair:
@sgiehl should be stable size now. 😃

plugins/CoreHome/javascripts/dataTable.js
```diff
@@ -567,6 +566,11 @@ $.extend(DataTable.prototype, UIControl.prototype, {
 
             if (labelColumnWidth) {
                 $('td.label', domElem).each(function() {
+                    var tableClass =  $(this).closest('table').attr('class');
+                    // check if the table is sub table. remove the width by 20px to make sure the table size wont change.
+                    if (tableClass.includes('subDataTable')) {
+                        labelColumnWidth = labelColumnWidth - 20;
+                    }
```
sgiehl: That seems to cause that the label column of subtables is smaller than the one of the parent table, which looks a bit ugly. Guess then it's better if the table becomes bigger when opening the subtable.

peterhashair:
@sgiehl just checking is that one good to merge :)

sgiehl:
@peterhashair So the current implementation still has the "issue" that the table might get a bit bigger when a subtable is opened, right?
For me that solution would be fine for now, as it at least improves it a lot. @tsteur is that fine, or shall we again check if it's possible to fix the increasing size as well?

tsteur:
Not sure if it's caused by this or if it's a different issue. I had a case where the table now got a lot wider than the widget itself. but only in some cases when I opened eg the first row. When I opened the second row it was scrolling within the available area. I couldn't reproduce it on an older version but maybe it only happens in specific cases:

https://user-images.githubusercontent.com/273120/153954648-4f04b64d-dca5-4800-aa44-909adbe15962.mp4

This happens sometimes that eventually the width gets too wide and the content overflows.

Apart from this I couldn't really reproduce the issue anymore I believe.

peterhashair:
@tsteur ah, right, I think this because the expanded has an extra column and the first column size changed. Did an update, it only recalculated once, I think should fix that problem.

peterhashair:
@sgiehl find another solution, just resize once, should fits nows :)

sgiehl:
There is still the issue that smaller tables, like the report for events might slightly get wider when a subtable is opened.
I would suggest to leave that as is for now, at it seems not that easy to fix. The other issues seem fix.
@tsteur what do you think?

tsteur:
That be fine for me 👍 . Does the table only get a bit wider or is the content overflowing? Overflowing be a bigger issue. If that's the case, could we create an issue for this and fix it as part of Matomo 4.9? 

Either way be good to create a follow up issue that describes the problem and where it happens / when it happens. Then we can see if / when we fix it

sgiehl (approved):

sgiehl:
@peterhashair could you create a follow-up issue and merge this one afterwards.

peterhashair:
@sgiehl sure

PR #18732, opened 2022-02-03
Fixed round fatal error in php8

AltamashShaikh:
### Description:

Fixed round fatal error in php8.
Fixes: #DEV-2577

In PHP7.x
echo round('82,460.82'); // results in 82

Whereas for PHP 8
echo round('82,460.82'); // results in fatal error

### Review

### DIFF
--- a/core/Columns/Dimension.php
+++ b/core/Columns/Dimension.php
@@ -456,10 +456,7 @@ abstract class Dimension
             case Dimension::TYPE_DURATION_S:
                 return $formatter->getPrettyTimeFromSeconds($value, $displayAsSentence = false);
             case Dimension::TYPE_DURATION_MS:
-                $val = number_format($value / 1000, 2, '.', '');
-                if ($val > 60) {
-                    $val = round($val);
-                }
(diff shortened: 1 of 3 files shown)


core/Columns/Dimension.php
```diff
@@ -456,7 +456,7 @@ public function formatValue($value, $idSite, Formatter $formatter)
             case Dimension::TYPE_DURATION_S:
                 return $formatter->getPrettyTimeFromSeconds($value, $displayAsSentence = false);
             case Dimension::TYPE_DURATION_MS:
-                $val = number_format($value / 1000, 2);
+                $val = number_format($value / 1000, 2, '.', '');
                 if ($val > 60) {
                     $val = round($val);
                 }
```
sgiehl: The number format is actually only used to limit the precision. So can't that be simplified with
```suggestion
                $val = round($val / 1000, $val > 60 ? 0 : 2);
```

AltamashShaikh: makes sense and works a it is :+1:

AltamashShaikh: Updated as suggested

sgiehl (approved):

PR #18739, opened 2022-02-04
[tests] Increase Release tests file lmited size from 55 to 56

peterhashair:
### Description:

Seems the Travis php@8.0 all tests-> release check tests. Start to fail, because file size increases. Update from 55 to 56.

Increase Release tests file limited size from 55 to 56

### Review

### DIFF
--- a/core/Updates/4.9.0-b1.php
+++ b/core/Updates/4.9.0-b1.php
@@ -18,9 +18,9 @@ use Piwik\Updater\Migration\Factory;
 use Piwik\Updates as PiwikUpdates;
 
 /**
- * Update for version 5.0.0-b1
+ * Update for version 4.9.0-b1
  */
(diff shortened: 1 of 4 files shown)


sgiehl:
@peterhashair I would guess that the test actually needs to be updated. It tries to exclude all files that are not included in the release. But the list of files is kind of outdated and the detection for dev only composer packages also seems incorrect.

This list needs to be checked: 
https://github.com/matomo-org/matomo/blob/f46abedba2ebf29b19d76f016e58b5ca686c4733/tests/PHPUnit/Integration/ReleaseCheckListTest.php#L816-L818

Also we are currently reading the `composer.json` file in order to check for dev only packages. But actually we should use `composer.lock`, as this file also includes several dependencies required by the dev packages we require.

sgiehl:
closing in favor of #18751

Issue #18082, opened 2021-09-30
Fix randomly failing UI tests


tsteur:
These ones fail randomly quite often AFAIK 

* [ ] https://builds-artifacts.matomo.org/matomo-org/matomo/4.x-dev/49535/UsersManager_permissions_bulk_access_set.png
* [ ] here it seems to not load the CSS randomly https://builds-artifacts.matomo.org/matomo-org/matomo/next_release/49592/
* [ ] This is another one I've seen that doesn't load the css https://builds-artifacts.matomo.org/matomo-org/matomo/m-6032-export-transitions/49534/UIIntegrationTest_admin_themes.png

Could we maybe retry them when they fail? or try to fix them in a different way?

bx80:
This is another one I've seen that doesn't load the css https://builds-artifacts.matomo.org/matomo-org/matomo/m-6032-export-transitions/49534/UIIntegrationTest_admin_themes.png

peterhashair:
There is one that always failed on my pr.
```
  1) Overlay
       should load overlay correctly when coming from an widgetized action report:
          Cannot read property 'hover' of null
```
Just a thought, is that possible we only run tests on the parts we changed. like a subModule or Composer package :)

sgiehl:
There is a general problem with automatically retrying UI tests. We have a lot tests that rely on the state of the previous test or that perform actions that can't be done again (like adding or removing stuff). So we could only rerun the whole test suite if one test failed, which then could cause timeouts if a long running test suite is rerun automatically.

Regarding the not loaded CSS/JS. I already added a lot code that automatically tries to load the CSS/JS again if the request fails, but there still seem to be some circumstances where all requests serve an empty response, so I guess the generated and cached CSS/JS is kind of empty and re-requesting it doesn't help.
I was never able to reproduce that locally, so this might be some travis specific error 🤷 

@peterhashair I'll try to fix that failing Overlay UI test. Not yet sure why it fails, as it's working correctly locally.
And we are running all UI tests on purpose. Only running some specific ones might have the risk, that a change accidentally changed something else and you might not see that.

sgiehl:
The overlay UI test should be fixed with https://github.com/matomo-org/matomo/pull/18084

peterhashair:
Get this on the UI test, I guess there is a timeout or ssl on the test. 
![image](https://user-images.githubusercontent.com/3783072/135953132-021e9d8b-2c3a-42cb-b9e1-293b1446cb7b.png)

[...]

PR #18748, opened 2022-02-07
Fix failing UserId tests

sgiehl:
### Description:

The UserId tests were actually failing due to this change: #18636

The reason for that seems to be the weird tracking done for the UserId tests, causing the tracked data to be slightly different.
I'll add some comments in the code to make that a bit more clear what the issue was.

refs #18082 

### Review

### DIFF
--- a/plugins/UserId/tests/Fixtures/TrackFewVisitsAndCreateUsers.php
+++ b/plugins/UserId/tests/Fixtures/TrackFewVisitsAndCreateUsers.php
@@ -38,12 +38,6 @@ class TrackFewVisitsAndCreateUsers extends Fixture
             for ($numVisits = 0; $numVisits < ($key+1) * 10; $numVisits++) {
                 $t->setUserId($userId);
                 $t->setVisitorId(str_pad($numVisits.$key, 16, 'a'));
-                if ($numVisits % 5 == 0) {
-                    $t->doTrackSiteSearch('some search term');
-                }
-                if ($numVisits % 4 == 0) {
-                    $t->doTrackEvent('Event action', 'event cat');
-                }
                 $t->setForceNewVisit();
                 $t->setUrl('http://example.org/my/dir/page' . ($numVisits % 4));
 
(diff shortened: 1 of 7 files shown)


plugins/UserId/tests/Fixtures/TrackFewVisitsAndCreateUsers.php
```diff
@@ -38,19 +38,25 @@ private function trackVisits()
             for ($numVisits = 0; $numVisits < ($key+1) * 10; $numVisits++) {
                 $t->setUserId($userId);
                 $t->setVisitorId(str_pad($numVisits.$key, 16, 'a'));
-                if ($numVisits % 5 == 0) {
-                    $t->doTrackSiteSearch('some search term');
-                }
-                if ($numVisits % 4 == 0) {
-                    $t->doTrackEvent('Event action', 'event cat');
-                }
```
sgiehl, line 46: Doing the sitesearch and event tracking upfront actually caused the problems. At that point the forced date time was not yet set. So the date time from the previous iteration was used. For the first iteration of a userid that means it used the date time from the last iteration from the previous user id.

So the first iteration of userid3 inserted a visit for e.g. '2010-02-20' for the event tracking and sitesearch, but the pageview below was than inserted for '2010-02-01'. Each iteration then forces a new visit to be inserted. So a couple iterations later a second visit for '2010-02-20' with the exact same time will be inserted. The next iteration then tracks the sitesearch and event. To map those actions to the correct visit the tracker tries to load the visit to increase the action counters. But there are now two visits with the exact same parameters (idvisitor, user_id, visit_last_action_time, ...), so it's actually up to the database to decide which of the records to return first. That behavior changed due to the new index, as that seems to have changed the order in which the records are returned.

As this is actually only a very test specific issue, that won't occur while normal tracking, I've adjusted the tracking to be a bit more realistic and avoid duplicate visit entries...

peterhashair (approved):
looks great

PR #18770, opened 2022-02-10
[Vue] patch vue compiler so it does not decode &nbsp; entities

diosmosis:
### Description:

As title. I thought of skipping the UMD files, but I'm not sure why the test was originally added, and it looked related to security. Since `&nbsp;` could be easily needed in the future it also doesn't seem acceptable to just never use it.

I found the place where this happens in the vue compiler and found a way to patch it.

### Review


peterhashair:
@diosmosis it still failed https://app.travis-ci.com/github/matomo-org/matomo/jobs/559193206. I guess just replace`<p>&nbsp;</p>` to `<br/>`  maybe

diosmosis:
@peterhashair that will fix it now, but it means using `&nbsp;` in vue templates is not allowed which seems like it could become a problem.

diosmosis:
I can find one simple way to fix this but it the result still requires remembering and explaining something, so I don't think its much better than having to explain that &nbsp; cannot be used directly in templates. Will close this.

PR #18790, opened 2022-02-14
Prevent archiving of data for time periods that start in the future

### LINKED ISSUE #18773: Dependent archives show only partial data or no report data
refs DEV-2418 for more details.

When archiving a week, then in https://github.com/matomo-org/matomo/blob/4.7.1/core/Archive.php#L629-L643 we archive up to 2 days into the future. That means if today is `2022-02-09`, then when archiving this week we also archive `2022-02-10` and `2022-02-11`. I noticed this in the archive tables:

image
image

In above screenshots we can see that the archive was created eg on 2nd Feb for a date in the future.

That means we create archives for in the future and of course these archives will have 0 data.

I assume that these archives in the future are also created for regular archives, not only dependent archives. However, I believe it works for these regular archives because they are invalidated again regularly and therefore the data is re-archived one or two days later. The logic for invalidating archives currently doesn't know which dependent archives exist with what segments. Meaning I'm assuming these get never invalidated.

I'm meaning calls like https://github.com/matomo-org/matomo/blob/4.7.1/plugins/Goals/Archiver.php#L124-L127
`$this->getProcessor()->processDependentArchive('Goals', VisitFrequencyAPI::NEW_VISITOR_SEGMENT);`. These calls also happen in non-core plugins specifically MediaAnalytics for example. As a result, the media map always shows 0 data unless if for some reason there is rarely already some data there.

Not sure if we still need to allow archiving any date in the future and why this is there in https://github.com/matomo-org/matomo/blob/4.7.1/core/Archive.php#L629-L643 . This code goes back 9 years. There are likely few ways to fix this:

The goals of this issue is that Media Analytics Map works again correctly (see DEV-2418 for all the details). We can see this eg on [demo](https://demo.matomo.cloud/index.php?module=CoreHome&action=index&idSite=1&period=day&date=yesterday#?idSite=1&period=month&date=2022-02-04&segment=&category=MediaAnalytics_Media&subcategory=MediaAnalytics_TypeAudienceMap)

1. Maybe we don't need to archive this data in the future then the issue be fixed. 
2. Or maybe we need to have some event in something like https://github.com/matomo-org/matomo/blob/4.7.1/core/DataAccess/Model.php#L119 to also trigger archives for segments that are used by `processDependentArchive`. 

I've also created a related issue: "invalidation for dependent archives doesn't work" in #18772 . Depending how this issue is fixed, the other issue will be fixed too. I separated them because this one might be easier to fix by changing to not archive data in the future anymore.

bx80:
### Description:

Fixes #18773 

When archiving a week, month or year some sub-periods may fall in the future (eg. it's Thursday so we want to archive this week, but Friday, Saturday and Sunday are in the future).

Writing blank archive data for the future causes problems (and is inefficient) so the code in Archive.php#L629-L643 skips any archives for periods with a start date in the future or before the website was created. However for some reason it only skips archives with start dates two days greater than the current date, this causes blank archives to be written for today+1 and today+2.

This is a simple fix to skip any periods which start on a date later than the current date.

Prior to this change a running `./console core:archive` command would result in `Archive::prepareArchive()` being called for dates two days into the future, after the change the method is only called up to and include the current date.

### Review

### DIFF
--- a/core/Archive.php
+++ b/core/Archive.php
@@ -610,8 +610,6 @@ class Archive implements ArchiveQuery
      */
     private function cacheArchiveIdsAfterLaunching($archiveGroups, $plugins)
     {
-        $today = Date::today();
-
         foreach ($this->params->getPeriods() as $period) {
             $twoDaysAfterPeriod = $period->getDateEnd()->addDay(2);
 
@@ -633,6 +631,9 @@ class Archive implements ArchiveQuery
                         $idSite, $period->getLabel(), $period->getPrettyString());
                     continue;
                 }
+
+                // Allow for site timezone, local time may have started a new day ahead of UTC
+                $today = \Piwik\Date::factory('now', $site->getTimezone());
 
                 // if the starting date is in the future we know there are no visits
                 if ($period->getDateStart()->isLater($today)) {
--- a/core/Archive.php
+++ b/core/Archive.php
@@ -613,7 +613,6 @@ class Archive implements ArchiveQuery
         $today = Date::today();
 
         foreach ($this->params->getPeriods() as $period) {
-            $twoDaysBeforePeriod = $period->getDateStart()->subDay(2);
             $twoDaysAfterPeriod = $period->getDateEnd()->addDay(2);
 
             foreach ($this->params->getIdSites() as $idSite) {
@@ -635,8 +634,8 @@ class Archive implements ArchiveQuery
                     continue;
                 }
 
-                // if the starting date is in the future we know there is no visiidsite = ?t
-                if ($twoDaysBeforePeriod->isLater($today)) {
+                // if the starting date is in the future we know there are no visits
+                if ($period->getDateStart()->isLater($today)) {
                     Log::debug("Archive site %s, %s (%s) skipped, archive is after today.",
                         $idSite, $period->getLabel(), $period->getPrettyString());
                     continue;
--- a/tests/PHPUnit/Integration/ArchiveTest.php
+++ b/tests/PHPUnit/Integration/ArchiveTest.php
@@ -273,9 +273,9 @@ class ArchiveTest extends IntegrationTestCase
 
         // archive range and day
         Rules::setBrowserTriggerArchiving(true);
-        API::getInstance()->get(1, 'day', '2020-03-04');
-        API::getInstance()->get(1, 'day', '2020-03-05');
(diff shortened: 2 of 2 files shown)


sgiehl:
@bx80 could you maybe add some tests for that? Could be a simple test that triggers the archiving and checks which records are created in the archive tables.
Note: To ensure the test assumes a certain date is in the future, you can manipulate the "current time" by setting `Date::$now`

sgiehl:
The code looks fine. But what I'm not fully sure about is, if that might cause issues when using other timezones. But guess even if the website is in a timezone that is ahead UTC, the start date should still not be in the future. 🤔 🤯

sgiehl:
@tsteur would you mind having a quick look at this one. I guess it should be fine. But I'm always unsure if those changes could result in any timezone related issues. Maybe we should check if we have enough tests that are using other timezones to ensure this won't cause any problems?

tsteur (changes requested):
At first I was going to say it's no problem because it's all in UTC. However, I was then double checking and set timezone to `Pacific/Auckland` for a site. It's now Feb 24th 10AM in Auckland.

UTC today is Feb 23rd. 

Matomo correctly wants to archive data for Feb 24th. It should already allow archiving Feb 24th here but it would not allow it. It goes into `continue`. That's a problem.

In case it's useful:
```php
$today->getDatetime(); //2022-02-23 00:00:00
 \Piwik\Date::now()->getDatetime(); // 2022-02-23 21:02:02
 \Piwik\Date::factory('now', 'Pacific/Auckland')->getDatetime(); //2022-02-24 10:02:02
```

sgiehl:
Ok. Guess we should then start adding some archiving tests for sites with timezones ahead of UTC, while faking a specific time to ensure the timezone is already a day ahead.

bx80:
I've added an additional test which will expect the current local date to be included for archiving even if it's ahead of UTC. The range in the future check is now based of the site local time date rather than the UTC date at midnight.

sgiehl:
@bx80 guess that logically makes sense. Would you mind also adding a test case where the site's timezone is behind UTC and UTC time is already a day ahead? Guess in that case it shouldn't create archives as well, right?

sgiehl (approved):
Code looks fine now and the added tests should cover all cases I think. I did some local testing by creating some sites with timezones far ahead and behind UTC and checked that no archives in the future were created. That all seem to work as expected.

PR #18796, opened 2022-02-15
[Bug] Cant not Copy User Dashboard to another user

### LINKED ISSUE #18692: Fixes problem where dashboards can't be copied to users in certain circumstances
Refs DEV-2411
Not sure exactly under what circumstances this bug occurs, but in some cases its not possible to copy a dashboard to a different user. 

For example we had a case where an `Admin` user was not able to copy a dashboard to a specific `View` user. They were able to copy to **some** View users and other Admin users, but several View users were not showing on the list when trying to copy.
Other Admin users were also not able to copy to that specific View users until that View user was made an Admin user.

Thomas mentioned that this might be something that only happens under certain circumstances depending on what users exist:
image

peterhashair:
### Description:

Fixes: #18692 
 
Cant not Copy User Dashboard for different sites with different access. 

My guess was the getUser API endpoints, do have related param site Id, when site selector changes, getUser still return default site sitting maybe.

Question: do we allow users to copy the dashboard to themselves? 

### Review

### DIFF
--- a/plugins/Dashboard/javascripts/dashboard.js
+++ b/plugins/Dashboard/javascripts/dashboard.js
@@ -103,16 +103,13 @@ function copyDashboardToUser() {
     var ajaxRequest = new ajaxHelper();
     ajaxRequest.addParams({
         module: 'API',
-        method: 'UsersManager.getUsers',
+        method: 'UsersManager.getSiteAccessUsers',
         format: 'json',
-        filter_limit: '-1'
+        filter_limit: '-1',
     }, 'get');
     ajaxRequest.setCallback(
         function (availableUsers) {
             $(makeSelectorLastId('copyDashboardUser')).empty();
-            $(makeSelectorLastId('copyDashboardUser')).append(
-                $('<option></option>').val(piwik.userLogin).text(piwik.userLogin)
-            );
             $.each(availableUsers, function (index, user) {
                 if (user.login != 'anonymous' && user.login != piwik.userLogin) {
                     $(makeSelectorLastId('copyDashboardUser')).append(
--- a/plugins/UsersManager/Model.php
+++ b/plugins/UsersManager/Model.php
@@ -132,7 +132,21 @@ class Model
     {
         $db = $this->getDb();
         $users = $db->fetchAll("SELECT login FROM " . Common::prefixTable("access")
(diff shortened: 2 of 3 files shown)


peterhashair:
Notes: This is a draft, not sure I am on the right track if that's not the case. I will look at it here
https://github.com/matomo-org/matomo/blob/8121427ba53127451bc74fdc3b0887e9843237e4/plugins/UsersManager/UserAccessFilter.php#L168-L182

plugins/UsersManager/Model.php
```diff
@@ -132,7 +132,21 @@ public function getUsersLoginWithSiteAccess($idSite, $access)
     {
         $db = $this->getDb();
         $users = $db->fetchAll("SELECT login FROM " . Common::prefixTable("access")
-                               . " WHERE idsite = ? AND access = ?", array($idSite, $access));
+                               . " WHERE idsite = ? AND access in (?)", array($idSite, $access));
```
sgiehl, line 135: Any reason for changing that line? Can't see any benefits in using an `IN` for a singe value only.

plugins/Dashboard/javascripts/dashboard.js
```diff
@@ -103,16 +103,13 @@ function copyDashboardToUser() {
     var ajaxRequest = new ajaxHelper();
     ajaxRequest.addParams({
         module: 'API',
-        method: 'UsersManager.getUsers',
```
sgiehl, line 106: Doesn't `getUser` already filter the returned list for users having access to the same site? If that doesn't work here, than the api method to copy a dashboard would have the same issue:
https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/plugins/Dashboard/API.php#L114-L126

sgiehl (changes requested):
@peterhashair If there are issues with users not being displayed, it might be worth checking the tests first. Maybe there are test cases for `UsersManager.getUsers` that are currently uncovered. If so it might be better to start with writing tests for such cases in order to find the problem. Your solution looks like a wild guess, without any evidence it actually solves something.

peterhashair:
@sgiehl that makes sense, will write tests for that, see if `getUsers()` covers different site ID permission.

PR #18820, opened 2022-02-18
Fix cookies with session lifetime

### LINKED ISSUE #18819: Fix cookies with session lifetime
The language change option for the anonymous user doesn't seem to be working at the moment (Tested in 4.6.1 and 4.7.1)

It shows the drop down menu and you can select a different language, but it just kicks it back to English every time.
image

This happens on any instance that has Anonymous user access enabled including demo.matomo.cloud and demo2.matomo.org

sgiehl:
### Description:

As a regression from https://github.com/matomo-org/matomo/pull/18179 cookies with session lifetime don't work anymore.

As stated in the Cookie class an expire of `0` should indicate a session cookie:
https://github.com/matomo-org/matomo/blob/9c666ab8cf4c591ba26cd7d992972ef47d5369a2/core/Cookie.php#L81-L82

But that doesn't work anymore as the lifetime will be always converted to a datetime, resulting in `01-Jan-1970 00:00:00 UTC`, causing the cookie to be never set.

fixes #18819

### Review

### DIFF
--- a/core/Cookie.php
+++ b/core/Cookie.php
@@ -146,7 +146,10 @@ class Cookie
             }
         }
 
-        $Expires = $this->formatExpireTime($Expires);
+        // Format expire time only for non session cookies
+        if (0 !== $Expires) {
+            $Expires = $this->formatExpireTime($Expires);
+        }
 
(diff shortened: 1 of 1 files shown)


peterhashair (approved):
Looks good to me

PR #18838, opened 2022-02-23
[UI] add report date on overview reports and fix date display

### LINKED ISSUE #18806: Inline Help - show report generation date on "Overview" reports with Evolution Graphs
In #18689 It was reported that the "Report Freshness" was not showing correctly which was fixed with #18498 
However, I noticed that on the "Overview" reports (The ones that contain the evolution graphs) don't show this date at all:
image
image

Is it possible for this to be shown on these reports, or is this data not stored for the evolution graphs?

peterhashair:
### Description:

Fixes:  #18806 

add report date on overview reports and fix date display. 
When the overview type is DataMap, get the last item `ts_archived` as the report date.  
Screen Shot 2022-02-23 at 3 46 13 PM

### Review


plugins/CoreHome/templates/_dataTable.twig
```diff
@@ -13,7 +13,7 @@
     {% if properties.title %}
         <h2 class="card-title"
             {% if properties.title_edit_entity_url %}edit-url="{{ properties.title_edit_entity_url }}"{% endif %}
-            {% if reportLastUpdatedMessage is defined and reportLastUpdatedMessage %}report-generated="{{ reportLastUpdatedMessage|raw }}"{% endif %}
+            {% if reportLastUpdatedMessage is defined %}report-generated="{{ reportLastUpdatedMessage|raw }}"{% endif %}
```
sgiehl: Doesn't that mean, that the attribute will be also set if the value is empty? If the attribute should be set in any case, it would be better to use something like this:
```suggestion
            report-generated="{% if reportLastUpdatedMessage is defined %}{{ reportLastUpdatedMessage|raw }}{% endif %}"
```

core/DataTable/Map.php
```diff
@@ -540,4 +540,10 @@ public function getColumns()
         }
         return array();
     }
+
+    public function getLastMetaData()
+    {
+        $lastItem = end($this->array);
+        return $lastItem->getAllTableMetadata();
+    }
```
sgiehl: When adding new methods to such core classes, please ensure to always add a doc block, that describes the method.
In that case I wouldn't actually add that method at all here, as it isn't a method that will get used often.

core/Plugin/Visualization.php
```diff
@@ -461,10 +461,15 @@ private function postDataTableLoadedFromAPI()
         if ($this->dataTable instanceof DataTable) {
             $this->metadata = $this->dataTable->getAllTableMetadata();
 
-            if (isset($this->metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
-                $this->reportLastUpdatedMessage = $this->makePrettyArchivedOnText();
+
+        } else {
+            if ($this->dataTable instanceof DataTable\Map) {
+                $this->metadata = $this->dataTable->getLastMetaData();
```
sgiehl: Using the metadata of the last table in the map might be an approach that might be good enough. But it is actually inaccurate. The date stored in that table might not be the newest date. Someone could e.g. invalidate and regenerate older data. To have that correct, you would need to iterate over all tables and look for the newest date.

sgiehl (changes requested):

core/DataTable/Map.php
```diff
@@ -540,4 +540,25 @@ public function getColumns()
...
+
+        $largestKey = null;
+        // find the largest key
+        foreach ($this->array as $key => $item) {
+            if (!$largestKey || (strtotime($largestKey) < strtotime($key))) {
+                $largestKey = $key;
+            }
+        }
+
+        $latestItem = $this->array[$largestKey];
```
sgiehl: For an empty map this method would throw a warning or error, as it would try to access `$this->array[null];`

core/DataTable/Map.php
```diff
@@ -540,4 +540,25 @@ public function getColumns()
...
+     * @return array
+     * @throws Exception
+     */
+    public function getLatestMetaData()
+    {
+
+        $largestKey = null;
+        // find the largest key
+        foreach ($this->array as $key => $item) {
+            if (!$largestKey || (strtotime($largestKey) < strtotime($key))) {
```
sgiehl: This won't work in all cases. Datatable maps can be indexed by date, which you seem to assume, but they can also be index by idsite. And actually it can also be a map of maps, on indexed by date, the other by idsite.
As it might be hard to handle all those cases here in a global method that is actually only needed for one special purpose, I would still recommend to move that handling to the Visualization class instead, where you can be sure it can't be a map of maps or indexed by idsite.
Also iterating over the index doesn't actually fix what I mentioned before. An older date can still have a newer generation date, due to invalidations.

sgiehl (changes requested):

peterhashair:
@sgiehl update a little, but I guess here could be the problem, eg: the first item ts_created not exist. Any suggestions?

```php
 if (!isset($metaData[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
                        break;
 }
```

core/Plugin/Visualization.php
```diff
@@ -461,10 +461,30 @@ private function postDataTableLoadedFromAPI()
...
+            // if the dataTable is Map
+            if ($this->dataTable instanceof DataTable\Map) {
+                // load all the data
+                $dataTable = $this->dataTable->getDataTables();
+                // find the latest key
+                foreach ($dataTable as $item) {
+                    $metaData = $item->getAllTableMetadata();
+                    // if ts_created not exist exit loop
+                    if (!isset($metaData[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
+                        break;
```
sgiehl: shouldn't this be a continue, so the remaining tables will be checked for a valid metadata entry?

core/Plugin/Visualization.php
```diff
@@ -461,10 +461,30 @@ private function postDataTableLoadedFromAPI()
...
+                    // if ts_created not record in the metadata or if current is more recent up the metadata
+                    if (!isset($this->metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])
+                      || strtotime($metaData[DataTable::ARCHIVED_DATE_METADATA_NAME]) > strtotime($this->metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
+                        $this->metadata = $metaData;
+                    }
+               }
             }
         }
+        if (isset($this->metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
+            $this->reportLastUpdatedMessage = $this->makePrettyArchivedOnText();
```
sgiehl: I would suggest to pass `$metadata[DataTable::ARCHIVED_DATE_METADATA_NAME]` as parameter to `makePrettyArchivedOnText` instead of accessing the private property in that method. As the metadata property only seems to be used for that one special case, it can be removed afterwards.

core/Plugin/Visualization.php
```diff
@@ -461,10 +461,30 @@ private function postDataTableLoadedFromAPI()
...
+                    }
+                    // if ts_created not record in the metadata or if current is more recent up the metadata
+                    if (!isset($this->metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])
+                      || strtotime($metaData[DataTable::ARCHIVED_DATE_METADATA_NAME]) > strtotime($this->metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
+                        $this->metadata = $metaData;
+                    }
+               }
             }
         }
+        if (isset($this->metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
```
sgiehl: I know it has been that way before, but wouldn't an `!empty` check make more sense here? If for any reason the metadata entry exist, but has no value it shouldn't be used

sgiehl (changes requested):

core/Plugin/Visualization.php
```diff
@@ -461,10 +461,30 @@ private function postDataTableLoadedFromAPI()
...
+                foreach ($dataTable as $item) {
+                    $metaData = $item->getAllTableMetadata();
+                    // if ts_created not exist exit loop
+                    if (empty($metaData[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
+                        continue;
+                    }
+                    // if ts_created not record in the metadata or if current is more recent up the metadata
+                    if (!empty($this->metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])
+                      || strtotime($metaData[DataTable::ARCHIVED_DATE_METADATA_NAME]) > strtotime($this->metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
+                        $this->metadata = $metaData;
```
sgiehl: `$this->metadata` could now be converted to a local variable in the method only. As the property is only used in that one method, there is no use in holding it as private class property.

core/Plugin/Visualization.php
```diff
@@ -458,13 +457,34 @@ private function postDataTableLoadedFromAPI()
...
+                foreach ($dataTable as $item) {
+                    $metaData = $item->getAllTableMetadata();
+                    // if ts_created not exist exit loop
+                    if (empty($metaData[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
+                        continue;
+                    }
+                    // if ts_created not record in the metadata or if current is more recent up the metadata
+                    if (!empty($metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])
+                      || strtotime($metaData[DataTable::ARCHIVED_DATE_METADATA_NAME]) > strtotime($metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
+                        $metadata = $metaData;
```
sgiehl: Don't you think it would be a bit easier to read by renaming one variable. When quickly looking over the code it might be hard to see that those variables differ in their writing case. This might also be a bit error prone when the code needs to be changed.

core/Plugin/Visualization.php
```diff
@@ -458,14 +457,35 @@ private function postDataTableLoadedFromAPI()
...
+                    $itemMetaData = $item->getAllTableMetadata();
+                    // if ts_created not exist skip
+                    if (empty($itemMetaData[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
+                        continue;
+                    }
+                    // initial metadata and update metadata if current is more recent
+                    if (!empty($itemMetaData[DataTable::ARCHIVED_DATE_METADATA_NAME])
+                      || strtotime($itemMetaData[DataTable::ARCHIVED_DATE_METADATA_NAME]) > strtotime($itemMetaData[DataTable::ARCHIVED_DATE_METADATA_NAME])) {
+                        $metadata = $itemMetaData;
+                    }
```
sgiehl: Does that logically still make sense? Isn't every available record of `$itemMetaData[DataTable::ARCHIVED_DATE_METADATA_NAME]` overwriting `$metadata`

```suggestion
                    // initial metadata and update metadata if current is more recent
                    if (!empty($itemMetaData[DataTable::ARCHIVED_DATE_METADATA_NAME])
                        && (
                            empty($metadata[DataTable::ARCHIVED_DATE_METADATA_NAME]) 
                            || strtotime($itemMetaData[DataTable::ARCHIVED_DATE_METADATA_NAME]) > strtotime($metadata[DataTable::ARCHIVED_DATE_METADATA_NAME])
                        )
                    ) {
                        $metadata = $itemMetaData;
                    }
```

sgiehl (approved):

PR #18843, opened 2022-02-24
Use browser client hints for detection

### LINKED ISSUE #16125: support Sec-CH-UA Client Hints in addition to UserAgent
(I thought I wrote about this before, but couldn't find an issue)

The Chrome team is working on a large change that plans to deprecate the user-agent string. 
The current(?) Chrome 83 already freezes the string and Chrome 85 will unify it to one common string for all desktop and one for all mobile users.

As a replacement the client hints standard is proposed: https://github.com/WICG/ua-client-hints

This allows to get similar data about a device already separated by things like brand, browser, os, etc. (meaning device-detector would not be needed anymore).

There is both an HTTP header (that only returns non-vague results if the server requests them) and a JS API for accessing the data.

#### the main UA value is not a string, but a list/set

`Chrome"; v="73", "ChromiumBasedBrowser"; v="60", "Chromium"; v="73"` (and the corresponding dictionary from JS) is a valid response and browser vendors are encouraged to "randomly [include] additional, intentionally incorrect, comma-separated entries with arbitrary ordering".

---------------

It seems like at the moment only Chromium-based browsers are working on implementing this standard, but as it seems like these are by far the majority now, that doesn't really matter.

I added this to 4.0 RC for now as it impacts every Matomo user as the most popular user agent will display incorrect results.

Links:
https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/-2JIRNMWJ7s/yHe4tQNLCgAJ
https://www.chromestatus.com/feature/5995832180473856
https://github.com/WICG/ua-client-hints
https://wicg.github.io/client-hints-infrastructure/
`https://github.com/mozilla/standards-positions/issues/202#issuecomment-558294095` (Position of the firefox team on this standard)

sgiehl:
### Description:

this requires https://github.com/matomo-org/device-detector/pull/6989 to be finished and released

fixes #16125

### Review


sgiehl:
build js

core/DeviceDetector/DeviceDetectorFactory.php
```diff
@@ -20,37 +20,39 @@ class DeviceDetectorFactory
...
+        self::$deviceDetectorInstances[$cacheKey] = $deviceDetector;
 
         return $deviceDetector;
     }
 
-    public static function getNormalizedUserAgent($userAgent)
+    public static function getNormalizedUserAgent($userAgent, $clientHints = [])
     {
-        return mb_substr(trim($userAgent), 0, 500);
+        return mb_substr(md5(json_encode($clientHints) ?? '') . trim($userAgent), 0, 500);
```
Findus23: I don't know what this is used for exactly, but keep in mind that user agents are encouraged to randomize the exact values, so this might not be useful as a cache string. (See https://wicg.github.io/ua-client-hints/#grease)

sgiehl: It's used to cache the results. I know that the brands might be shuffled. Not yet sure how to improve it.

justinvelluppillai: sort it alphabetically first might work?

justinvelluppillai:
How are we looking here @sgiehl? Anything we need to do to push this forward?

sgiehl:
@justinvelluppillai Had to rewrite the JS a bit. As the former version didn't work in all cases. Once tests are fixed and added, this should be ready for a first review.

core/Cookie.php
```diff
@@ -456,7 +456,7 @@ private static function getSameSiteValueForBrowser($default)
             } else {
                 $userAgent = Http::getUserAgent();
                 $ddFactory = StaticContainer::get(\Piwik\DeviceDetector\DeviceDetectorFactory::class);
-                $deviceDetector = $ddFactory->makeInstance($userAgent);
+                $deviceDetector = $ddFactory->makeInstance($userAgent, Http::getClientHintsFromServerVariables());
```
sanchezzzhak, line 459: public static function getClientHintsFromServerVariables()
  it seems this method does not skip X_HTTP_REQUESTED_WITH, which is used to identify anonymous browsers

sgiehl:
Had some more thoughts on that one today. The implementation is kind of finished, but some more tests would be needed.
We also need to consider to make it possible to set the client hints in Matomo PHP Tracker, so we are also able to add some tests where those data is provided.
The general implementation should be ready for a first review.

core/Tracker/Request.php
```diff
@@ -634,6 +635,14 @@ public function getUserAgent()
         return Common::getRequestVar('ua', $default, 'string', $this->params);
     }
 
+    public function getClientHints()
+    {
+        // use headers as default if no data was send with the tracking request
+        $default = Http::getClientHintsFromServerVariables();
+
+        return Common::getRequestVar('uadata', $default, 'json', $this->params);
```
Findus23, line 643: This new parameter would need to be documented on https://developer.matomo.org/api-reference/tracking-api, right?

js/piwik.js
```diff
@@ -3130,6 +3197,19 @@ if (typeof window.Matomo !== 'object') {
...
              */
             function detectBrowserFeatures() {
+                detectClientHints(function() {
+                    var i, requestType;
+                    clientHintsResolved = true;
+                    for (i = 0; i < clientHintsRequestQueue.length; i++) {
+                        requestType = typeof clientHintsRequestQueue[i];
+                        if (requestType === 'string') {
+                            sendRequest(clientHintsRequestQueue[i], configTrackerPause);
+                        } else if (requestType === 'object') {
```
Findus23, line 3253: Maybe I am misunderstanding this block, but would that mean that matomo.js is sending more requests than it would before?

Findus23 (approved):
This doesn't count as a proper code review, but I looked through everything and it seems reasonable to me.

js/piwik.js
sgiehl: No. This part actually does the same as the cookie request queue. As long as the client hints are not available all tracking requests are added to a queue. As soon as the client hints are available this queue will be send. This is required as the client hints are resolved using a promise. So we don't know when the promise is resolved. At the moment all browsers resolve the promise instantly. So waiting some milliseconds would be enough. But that could change in the future to the users e.g. must accept sending clients hints or similar before the promise resolves...

Findus23: Thanks, I forgot that the client hints check is asynchronous.

MatomoForumNotifications:

https://forum.matomo.org/t/windows-11-recorded-as-windows-10/45882/2

justinvelluppillai:
@bx80 although still in draft this is needing a more detailed review. Are you able to take ownership of reviewing this PR and take a look this week?

bx80:
build js

bx80 (approved):
I've looked through the code and done some tracker testing using both Chome providing client hints and firefox without. It all works well as far as I can see :+1:

One minor thing I noticed is that the stored browser version is a lot more detailed when obtained via client hints. On 4.x-dev I get config_browser_name = `Chrome`, config_browser_version = `101.0` whereas with client hints the same browser is recorded as config_browser_name = `Chrome`, config_browser_version = `101.0.4951.54`. This will reduce the usefulness of the browser version metric where there may be many 101.0.xxxx.xx rows. Would it perhaps be worth attempting to truncate the version string to the major version number for known browsers?

sgiehl:
> One minor thing I noticed is that the stored browser version is a lot more detailed when obtained via client hints. On 4.x-dev I get config_browser_name = Chrome, config_browser_version = 101.0 whereas with client hints the same browser is recorded as config_browser_name = Chrome, config_browser_version = 101.0.4951.54. This will reduce the usefulness of the browser version metric where there may be many 101.0.xxxx.xx rows. Would it perhaps be worth attempting to truncate the version string to the major version number for known browsers?

Good catch. The device detector is actually also able to detect such detailed versions, but returns the details requested. For Matomo we are using the default, which automatically truncates to the minor version number. So guess we should do that for client hints as well.

sgiehl:
I've update the PR so it now also includes some changes on related libs & plugins.
After final review has been done the following steps need to be performed before merging:
- [x] Review & merge https://github.com/matomo-org/matomo-php-tracker/pull/104
  ❗ Afterwards a new version needs to be released, and the composer requirement in this branch needs to be changed back to  `^3.0` (including a composer update, so the lock file is updated as well.)
- [x] Review & merge https://github.com/matomo-org/developer-documentation/pull/643
- [x] Review & merge https://github.com/matomo-org/plugin-DeviceDetectorCache/pull/44
  ❗ Afterwards the submodule in this branch needs to be updated
- [ ] Finally merge this PR

justinvelluppillai:
@bx80 it'd be great if you could give this a final review after the latest changes

plugins/DevicesDetection/templates/detection.twig
```diff
@@ -6,6 +6,32 @@
 
     <script type="text/javascript">
 
+        if (!navigator.userAgentData || typeof navigator.userAgentData.getHighEntropyValues !== 'function') {
+            $('#noclienthints').show();
+            $('[name=clienthints],#usech').hide();
```
bx80: The 'Your Browser does not support client hints.' message doesn't show for me on Firefox, I think the span visibility is being toggled before page rendering is complete. Wrapping the whole `if..else` block with `$(document).ready(functino() {..});` gets the message to show.

bx80 (approved):
The changes and code improvements all look good. I've repeated the functional testing with Chromium and Firefox testing with/without client hints, works fine. The tests all complete correctly (after a composer update). Minor issue with the devices detection diagnostic message, but I can't see any other issues :+1:

sgiehl:
@bx80 did you also check the PRs I have mentioned in the previous comment? Would be good to have them reviewed as well.

bx80:
@sgiehl I've now reviewed the additional PRs, can't see any issues other than a test needing an update :+1:

sgiehl:
All dependencies have been merged and updated in the PR. Once the tests have run through, I think this one can be merged.

MatomoForumNotifications:

https://forum.matomo.org/t/when-will-support-sec-ch-ua-be-implemented/46570/2

PR #18862, opened 2022-03-01
With multi_server_environment enable maintain won't send 503 as response

### LINKED ISSUE #18054: Multi server environment in maintenance mode shouldn't result in 503 response
In the FAQ https://matomo.org/faq/new-to-piwik/faq_134/ it is written
"When Matomo is used behind a load balancer, here are the steps to follow:
..... in your config.ini.php, below the section [General], add the following line: multi_server_environment=1. This setting lets Matomo know you’re using multiple servers....."

When publishing a web-service via loadbalancers, you use a health-monitor that checks that the web-application answers as expected. As described in https://matomo.org/faq/how-to/faq_20278/

When activating maintenance_mode, the Reporting API and Core platform changes status-code from 200 to 503.
This will cause the loadbalancers to stop sending traffic to the web-server. And the maintenance page is not reachable for the visitors/clients.

## Expected Behavior

When multi_server_environment = 1  (or new property indicating that there is a loadbalancers in front of matomo) is set.
And maintenance_mode = 1 is set,  the status code from Reporting API and Core platform should still be http 200.
So clients are displayed with the maintenance page.

## Current Behavior

Reporting API and Core platform returns 503 when maintenance_mode is set to 1.

## Possible Solution

When using multi_server_environment ( or new property),  and maintenance_mode is set to 1. Keep the response code 200.

1. maintenance_mode = 0 
2. `curl -ik 'https://matomo.example.com/index.php?module=API&method=API.getMatomoVersion&token_auth=<token>`
Results in a 200 OK
3. maintenance_mode = 1
4. same command as in step 2 results in a 503  Service Unavailable.

## Your Environment

* Matomo Version: 4.4.1
* PHP Version: PHP 7.4.3
* Server Operating System: Ubuntu 20.04.3

peterhashair:
### Description:
Fixes: #18054 

With multi_server_environment enable maintain won't send 503 as a response. Missing tests

Docs update
https://matomo.org/faq/how-to/faq_20278/
 ```
+ When Matomo is maintenance_mode, the API will return `503`, unless Matomo is configured behind a load balancer, with `General` config `multi_server_environment=1`, should return http status code `200` 
```

### Review

### DIFF
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -639,7 +639,7 @@ assume_secure_protocol = 0
 ; By enabling this flag we will for example not allow the installation of a plugin via the UI as a plugin would be only
 ; installed on one server or a config one change would be only made on one server instead of all servers.
 ; This flag doesn't need to be enabled when the config file is on a shared filesystem such as NFS or EFS.
-; During maintenance mode, multiple environment will continue return 200 as response on API request.
+; When enabled, Matomo will return the response code 200 instead of 503 in maintenance mode.
 multi_server_environment = 0
 
 ; List of proxy headers for client IP addresses
--- a/config/global.ini.php
+++ b/config/global.ini.php
@@ -639,6 +639,7 @@ assume_secure_protocol = 0
 ; By enabling this flag we will for example not allow the installation of a plugin via the UI as a plugin would be only
 ; installed on one server or a config one change would be only made on one server instead of all servers.
 ; This flag doesn't need to be enabled when the config file is on a shared filesystem such as NFS or EFS.
+; During maintenance mode, multiple environment will continue return 200 as response on API request.
 multi_server_environment = 0
 
 ; List of proxy headers for client IP addresses
--- a/core/FrontController.php
+++ b/core/FrontController.php
@@ -11,6 +11,7 @@ namespace Piwik;
 
 use Exception;
 use Piwik\API\Request;
+use Piwik\Config\GeneralConfig;
 use Piwik\Container\StaticContainer;
 use Piwik\DataTable\Manager;
 use Piwik\Exception\AuthenticationFailedException;
@@ -493,12 +494,12 @@ class FrontController extends Singleton
 
     protected function handleMaintenanceMode()
     {
-        if ((Config::getInstance()->General['maintenance_mode'] != 1) || Common::isPhpCliMode() ) {
+        if ((GeneralConfig::getConfigValue('maintenance_mode') != 1) || Common::isPhpCliMode() ) {
             return;
         }
 
         // as request matomo behind load balancer should not return 503. https://github.com/matomo-org/matomo/issues/18054
-        if (Config::getInstance()->General['multi_server_environment'] != 1) {
+        if (GeneralConfig::getConfigValue('multi_server_environment') != 1) {
             Common::sendResponseCode(503);
         }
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: core/FrontController.php [hunk 1]
```diff
@@ -493,7 +493,8 @@ protected function prepareDispatch($module, $action, $parameters)
 
     protected function handleMaintenanceMode()
     {
-        if ((Config::getInstance()->General['maintenance_mode'] != 1) || Common::isPhpCliMode()) {
+        // as request matomo behind load balancer should not return 503. https://github.com/matomo-org/matomo/issues/18054
+        if ((Config::getInstance()->General['maintenance_mode'] != 1) || Common::isPhpCliMode() || Config::getInstance()->General['multi_server_environment']) {
```
sgiehl: I don't think it's on purpose to disable the maintenance mode on multi server environments, is it?
I actually understood the issue in a way, that only the response code should be changed to `200` in that case...

core/FrontController.php
```diff
@@ -493,10 +493,14 @@ protected function prepareDispatch($module, $action, $parameters)
...
     protected function handleMaintenanceMode()
     {
-        if ((Config::getInstance()->General['maintenance_mode'] != 1) || Common::isPhpCliMode()) {
+        if ((Config::getInstance()->General['maintenance_mode'] != 1) || Common::isPhpCliMode() ) {
             return;
         }
-        Common::sendResponseCode(503);
+
+        // as request matomo behind load balancer should not return 503. https://github.com/matomo-org/matomo/issues/18054
+        if (Config::getInstance()->General['multi_server_environment'] != 1) {
```
sgiehl: I was wondering if we maybe should replace the usage of Config with the `GeneralConfig::getConfigValue`, when adjusting such code.
It's not really needed, but might be good at some point to only use `GeneralConfig`

sgiehl (commented):
The code looks fine now. We should maybe mention this behavior in the description of the config flag here:
https://github.com/matomo-org/matomo/blob/2ba0f7a67cb4aa336ef312e2850de1f325adcb69/config/global.ini.php#L637-L642

And not sure if this should also be mentioned in the changelog, as this is something that might be relevant there. @tsteur what do you think?

tsteur:
@sgiehl it doesn't hurt to mention it if it can be explained easily 👍  

* We also need to update https://matomo.org/faq/how-to/faq_111/
* and https://matomo.org/faq/new-to-piwik/faq_134/

config/global.ini.php
```diff
@@ -639,6 +639,7 @@
 ; By enabling this flag we will for example not allow the installation of a plugin via the UI as a plugin would be only
 ; installed on one server or a config one change would be only made on one server instead of all servers.
 ; This flag doesn't need to be enabled when the config file is on a shared filesystem such as NFS or EFS.
+; During maintenance mode, multiple environment will continue return 200 as response on API request.
```
sgiehl: Not sure if that makes it clear enough. Actually this is not limited on API requests. I would suggest something like `When enabled, Matomo will return the response code 200 instead of 503 in maintenance mode.`
Or maybe @justinvelluppillai has a better idea here.

sgiehl (approved):
Should be good to merge if all FAQs were updated accordingly.

sgiehl:
@peterhashair did you already update the FAQs on this one?

peterhashair:
@sgiehl yes, but haven't reviewed yet. They are here.

- https://matomo.org/wp-admin/revision.php?revision=50566
- https://matomo.org/wp-admin/revision.php?revision=50753
- https://matomo.org/wp-admin/revision.php?revision=50754

PR #18868, opened 2022-03-02
invite user replace add user 

### LINKED ISSUE #13321: Invite new users in Matomo, rather than creating them directly

peterhashair:
### Description:
fixes: https://github.com/matomo-org/matomo/issues/13321

Since we move UsersManager to VUE.  Redo the entire PR in the VUE version and fixes the remaining comments

This PR is VUE version of https://github.com/matomo-org/matomo/pull/18351

This probably needs docs for it, still working on the docs, not sure we want people inject their only term & condition on the set password page.

### Changes 
add one table column
  - invite_at datetime

### Review


peterhashair:
@sgiehl I got a question for the structure, is that ok for me to move some of the `API.php` private methods to a different class, just the public method remains in the API.php. In this case, is the UsersManager/API.php

sgiehl:
> @sgiehl I got a question for the structure, is that ok for me to move some of the API.php private methods to a different class, just the public method remains in the API.php. In this case, is the UsersManager/API.php

If private API methods could be reused somewhere else, it makes sense to move them to another class.

peterhashair:
@diosmosis sorry to bother, just checking for the Build Vue, I keep getting errors from plugin `MobileMessaging ` any idea, what causing that one?

diosmosis:
@peterhashair ScheduledReports isn't in MobileMessaging's umd.metadata.json so it's not being built first. Rebuilding locally adds it back for me. Or you can add it manually, the file should look like:

```
{
  "dependsOn": [
    "CoreHome",
    "CorePluginsAdmin",
    "ScheduledReports"
  ]
}
```

I'm not sure why it isn't there, though.

bx80 (changes requested):
@peterhashair Looking good overall :+1: 

I noticed a couple of issues: 

The `Resend Invite` action shows for all users even if they haven't been invited, this could be confusing and an error is shown when attempting to reinvite existing users. It would be good to hide `Resend Invite` if the user hasn't been invited.
image

After creating an invite for a new user, following the accept link and entering a new password an error is shown:
image
This happens when already logged in as one user who creates the invite, then the accept link is opened in the same browser. Probably not a common scenario but it could happen, especially when testing.

peterhashair:
@bx80 fixed it.

bx80 (approved):
The resend invite action is only showing when an invite has been sent :+1:

The `defaultReportDate` was due to an out of date plugin I had installed - no issue once the plugin was deactivated :+1:

core/Db/Schema/Mysql.php
```diff
@@ -53,6 +53,7 @@ public function getTablesCreateSql()
                           date_registered TIMESTAMP NULL,
                           ts_password_modified TIMESTAMP NULL,
                           idchange_last_viewed TIMESTAMP NULL,
+                          invite_status enum('accept','pending','decline','expired') DEFAULT 'accept',
```
sgiehl, line 56: Using enums is not always the best choice. For this table it might not hurt too much, but for tables that might contain a lot data, we should never use enums, as it is a very expensive operation to change them again later. Personally I would have preferred using a simple integer column and assigning the logical values in PHP only.

core/Validators/Login.php
```diff
@@ -0,0 +1,59 @@
...
+ *
+ */
+
+namespace Piwik\Validators;
+
+use Piwik\Piwik;
+use Piwik\SettingsPiwik;
+use Piwik\Plugins\UsersManager\API as APIUsersManager;
+
+class Login extends BaseValidator
```
sgiehl, line 16: Is this Validator actually used anywhere? If I see that correct, it's also included in a similar way in the UsersManager plugin.

plugins/Login/Controller.php
```diff
@@ -497,4 +510,141 @@ public function logout()
...
+        $model = new UsersModel();
+        $passwordHelper = new Password();
+        $view = new View('@Login/invitation');
+
+        $token = Common::getRequestVar('token', null, 'string');
+        $form = Common::getRequestVar('invitation_form', false, 'string');
+
+        //check token is valid
+        $user = $model->getUserByTokenAuth($token);
+        if ($user['invite_status'] !== 'pending') {
```
sgiehl, line 529: Won't this throw a notice if the user can't be found with the token? Guess `$user` would be `null` in that case.

plugins/Login/Controller.php
```diff
@@ -497,4 +510,141 @@ public function logout()
...
+            throw new Exception(Piwik::translate('Login_InvalidOrExpiredToken'));
+        }
+
+        //if user not match the invite user
+        if (!$user) {
```
sgiehl, line 534: This check would only make sense above the previous one, as it can't be reached here. But actually it doesn't make sense at all. There should be only "invalid token" error, if no user can't be found with the token, or the invite status isn't pending.

plugins/Login/Controller.php
```diff
@@ -497,4 +510,141 @@ public function logout()
...
+                    $mail = StaticContainer::getContainer()->make(UserAcceptInvitationEmail::class, array(
+                      'login'        => $user['login'],
+                      'emailAddress' => $user['email'],
+                      'userLogin'    => $user['login'],
+                    ));
+                    $mail->safeSend();
+                } catch (\Exception $e) {
+
+                }
+
```
sgiehl, line 592: The try/catch is actually useless, as `safeSend` already catches any errors and doesn't throw any.

plugins/Login/Controller.php
```diff
@@ -497,4 +510,141 @@ public function logout()
...
+    public function declineInvitation()
+    {
+        $model = new UsersModel();
+
+        $token = Common::getRequestVar('token', null, 'string');
+        $form = Common::getRequestVar('invitation_form', false, 'string');
+
+        $user = $model->getUserByTokenAuth($token);
+        if ($user['invite_status'] !== 'pending') {
```
sgiehl, line 613: Like in the other action, this might trigger a notice/warning.

plugins/Login/Controller.php
```diff
@@ -497,4 +510,141 @@ public function logout()
...
+            throw new Exception(Piwik::translate('Login_InvalidOrExpiredToken'));
+        }
+        //if user not match the invite user
+        if (!$user) {
```
sgiehl, line 617: Like in the other action, this check can't be reached.

tests/resources/OmniFixture-dump.sql
```diff
@@ -1256,6 +1256,7 @@ CREATE TABLE `user` (
   `superuser_access` tinyint(2) unsigned NOT NULL DEFAULT '0',
   `date_registered` timestamp NULL DEFAULT NULL,
   `ts_password_modified` timestamp NULL DEFAULT NULL,
+  `invite_status`  enum('accept','pending','decline','expired'),
```
sgiehl, line 1259: Changes to the omnifixture sql are **NOT** needed. Once the SQL is set up for tests an update is performed, which will trigger all update scripts that were added after the omnifixture was last built (based on the version included in this sql).

plugins/UsersManager/vue/src/UsersManager/UsersManager.vue
```diff
@@ -262,6 +263,17 @@ export default defineComponent({
         // ignore (errors will still be displayed to the user)
       }).then(() => this.fetchUsers());
     },
+    onResendInvite(user: User) {
+      console.log(user);
```
sgiehl, line 267: that should be removed

plugins/UsersManager/vue/src/UsersManager/UsersManager.vue
```diff
@@ -262,6 +263,17 @@ export default defineComponent({
...
+      AjaxHelper.fetch<AjaxHelper>(
+        {
+          method: 'UsersManager.resendInvite',
+          userLogin: user.login,
+        },
+      ).then((res) => {
+        console.log(res);
```
sgiehl, line 274: Why is there a single `console.log` in the `then`? Is nothing supposed to happen after the resend was sent? Shouldn't there be some sort of success notification? The user currently does not get any feedback it it worked or not!

plugins/Login/templates/inviteLayout.twig
```diff
@@ -0,0 +1,46 @@
+{% extends '@Morpheus/layout.twig' %}
+
+{% block meta %}
```
sgiehl, line 3: Adding this file could have been avoided. It actually has the same content as `Login\loginLayout.twig`. Only the title differs, which can be simply overwritten in the files using the layout.

plugins/Login/Controller.php
```diff
@@ -497,4 +510,141 @@ public function logout()
...
+            throw new Exception(Piwik::translate('Login_InvalidUsernameEmail'));
+        }
+
+        if ($form) {
+            $model->deleteAllTokensForUser($user['login']);
+            $model->updateUserFields($user['login'], ['invite_status' => 'decline']);
+            $this->redirectToIndex('Login', 'index');
```
sgiehl, line 624: Simply redirecting the user to the login, doesn't give them any feedback if the decline actually worked. Also does showing the login form not make any sense. The user just declined to be invited so why should he want to use the login form? Makes more sense to only show a message, that the invite was declined.

plugins/Login/templates/invitation.twig
```diff
@@ -0,0 +1,83 @@
...
+                                           autocorrect="off" autocapitalize="none"
+                                           tabindex="2"/>
+                                    <label for="password_confirm"><i class="icon-locked icon"></i> {{ 'Login_NewPasswordRepeat'|translate }}</label>
+                                </div>
+                            </div>
+                            <div class="row actions">
+                                <div class="col s12">
+                                    <label>
+                                        <input name="terms" type="checkbox" id="terms" value="1" tabindex="90"/>
+                                        <span>{{ 'PrivacyManager_TermsAndConditions'|translate }}</span>
```
sgiehl, line 45: Sorry, but that checkbox is crap, this way:

image

If the user should accept any terms and conditions, it should be written that way. E.g. "I accept the Terms & Conditions". Also the terms and conditions should be linked. Why should anyone accept any terms, if he actually can't read them.

The PrivacyManager plugin has a setting for the Terms & Condition URL. If it is set, it should be linked here.
If it isn't set, I guess it doesn't make sense to have the checkbox at all.

plugins/Login/lang/en.json
```diff
@@ -54,6 +54,15 @@
...
         "SuspiciousLoginAttemptsInLastHourEmail4": "Setup two-factor auth so attackers will need more information than just your password in order to login.",
         "SuspiciousLoginAttemptsInLastHourEmail5": "Additionally, if your Matomo has a limited set of users or IPs through which users will access it, it may be beneficial to setup a IP address allowlist. %1$sRead our docs for more information.%2$s",
-        "LoginNotAllowedBecauseUserLoginBlocked": "Login functionality is temporarily disabled since we've a suspicious amount of failed login attempts in the last hour."
+        "LoginNotAllowedBecauseUserLoginBlocked": "Login functionality is temporarily disabled since we've a suspicious amount of failed login attempts in the last hour.",
+        "InvitationTitle": "Accept Invitation",
+        "InvitationDeclineTitle": "Decline Invitation",
+        "InvitationDeclineBody": "Are you sure you want to decline this Invitation?",
+        "InvitationHints": "(to cancel this action, just leave the page)",
+        "Accept": "Accept",
+        "Yes": "Yes",
```
sgiehl, line 63: There is already "General_Yes", which could have been used instead of adding a new translation with the same content & meaning"

plugins/UsersManager/API.php
```diff
@@ -1539,12 +1430,37 @@ private function getDeviceDescription()
...
+    /**
+     * resend the invite email to user
+     * @param $userLogin
+     * @throws NoAccessException
+     * @throws \DI\DependencyException
+     * @throws \DI\NotFoundException
+     */
+    public function resendInvite($userLogin, $expired = 7)
+    {
+        Piwik::checkUserIsNotAnonymous();
```
sgiehl, line 1451: This check is redundant, as the anonymous user can't have super user access, which is checked below.

plugins/UsersManager/Emails/UserInviteEmail.php
```diff
@@ -0,0 +1,84 @@
...
+        $this->addTo($this->user['email']);
+        $this->setSubject($this->getDefaultSubject());
+        $this->addReplyTo($this->getFrom(), $this->getFromName());
+        $this->setWrappedHtmlBody($this->getDefaultBodyView());
+    }
+
+    protected function getDefaultSubject()
+    {
+        return Piwik::translate('CoreAdminHome_UserInviteSubject',
+          [$this->currentUser, $this->user['login']]);
```
sgiehl, line 61: The resulting sentence doesn't make much sense to me. It's currently something like
`admin sent you an invite to join Matomo Analytics for {newuser}.`
It should either be something like `admin sent you an invite to join Matomo Analytics as {newuser}.` Or the second placeholder should contain the site the user will have access to.

plugins/UsersManager/Emails/UserInviteEmail.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    private function getDefaultSubjectWithStyle()
+    {
+        return Piwik::translate('CoreAdminHome_UserInviteSubject',
+          ["<strong>" . $this->currentUser . "</strong>", "<strong>" . $this->user['login'] . "</strong>"]);
```
sgiehl, line 67: Same as above

plugins/UsersManager/Emails/UserInviteEmail.php
```diff
@@ -0,0 +1,84 @@
...
+        $view = new View('@UsersManager/_userInviteEmail.twig');
+        $view->login = $this->user['login'];
+        $view->emailAddress = $this->user['email'];
+        $view->token = $this->token;
+
+        // content line for email body
+        $view->content = $this->getDefaultSubjectWithStyle();
+
+        //notes for email footer
+        $view->notes = Piwik::translate('CoreAdminHome_UserInviteNotes', [$this->user['login'], $this->currentUser]);
```
sgiehl, line 81: Using the login as placeholder doesn't seem to make sense here either:
`This invitation was intended for {newuser} access. If you...`
What should that actually tell the user?

plugins/UsersManager/Repository/UserRepository.php
```diff
@@ -0,0 +1,225 @@
+<?php
+
+namespace Piwik\Plugins\UsersManager\Repository;
```
sgiehl, line 3: A general note: For new classes and methods it would be helpful if we can directly try to use more strict types. e.g. using type hints for method parameters as well as using return type hints where possible.

plugins/UsersManager/Model.php
```diff
@@ -681,11 +728,19 @@ public function getUsersWithAccessToSites($idSites)
...
         $loginSql = 'SELECT DISTINCT ia.login FROM ' . Common::prefixTable('access') . ' ia WHERE ia.idsite IN ('
-            . implode(',', $idSites) . ')';
+          . implode(',', $idSites) . ')';
 
         $logins = \Piwik\Db::fetchAll($loginSql);
         $logins = array_column($logins, 'login');
         return $logins;
     }
 
+    public function getPendingUser($userLogin)
```
sgiehl, line 738: That method should better be named `isPendingUser` as it returns the current state and not the user details. Also the return value might be better casted to boolean.

plugins/UsersManager/Model.php
```diff
@@ -444,36 +478,42 @@ public function getUserByTokenAuth($tokenAuth)
...
         );
 
+
+        if ($inviteStatus) {
+            $user['invite_status'] = 'pending';
+        }
+
         $db = $this->getDb();
         $db->insert($this->userTable, $user);
+        return $user;
```
sgiehl, line 500: Any reason for changing the return signature? Can't see that the return value is actually used somewhere

plugins/Login/Controller.php
```diff
@@ -497,4 +510,141 @@ public function logout()
...
+            if (!$error) {
+                $password = UsersManager::getPasswordHash($password);
+                $passwordInfo = $passwordHelper->info($password);
+
+                if (!isset($passwordInfo['algo']) || 0 >= $passwordInfo['algo']) {
+                    // password may have already been fully hashed
+                    $password = $passwordHelper->hash($password);
+                }
+
+                //update pending user to active user
```
sgiehl, line 572: Shouldn't the invite token be removed at that point? Otherwise it will remain in the database.

plugins/UsersManager/API.php
```diff
@@ -704,72 +692,40 @@ private function checkEmail($email, $userLogin = null)
...
+    {
+        //create User
+        $this->userRepository->create($userLogin, $email, $initialIdSite);
 
-        /**
-         * Triggered after a new user is created.
-         *
-         * @param string $userLogin The new user's login handle.
-         */
-        Piwik::postEvent('UsersManager.addUser.end', array($userLogin, $email, $password));
```
sgiehl, line 761: Removing this event and using `UsersManager.inviteUser.end` instead is a breaking change as various of our plugins currently use that event and will fail in the future.
Also new or removed events should always be mentioned the in changelog.md

plugins/UsersManager/Repository/UserRepository.php
```diff
@@ -0,0 +1,225 @@
...
+
+            //remove all previous token
+            $this->model->deleteAllTokensForUser($userLogin);
+
+            //generate Token
+            $generatedToken = $this->model->generateRandomTokenAuth();
+
+            //attach token to user
+            $this->model->addTokenAuth($userLogin, $generatedToken, "Invite Token", Date::now()->getDatetime(),
+              Date::now()->addDay($expired)->getDatetime());
```
sgiehl, line 118: Are you aware of the implications using a temporary token auth for this has?
The invite token can also be used to authenticate against the full API. That means a user can use the API with this token even if he didn't click on accept or decline. And even if he clicks, the token isn't removed and can be used until it expires. This might be a security problem.

plugins/UsersManager/Repository/UserRepository.php
```diff
@@ -0,0 +1,225 @@
...
+
+    public function __construct(Model $model, UserAccessFilter $filter, Password $password)
+    {
+        $this->model = $model;
+        $this->filter = $filter;
+        $this->password = $password;
+    }
+
+
+    public function index($userLogin, $pending)
```
sgiehl, line 41: Is that method actually used anywhere? If so it could for sure have a better name.

plugins/UsersManager/Repository/UserRepository.php
```diff
@@ -0,0 +1,225 @@
...
+use Piwik\Plugins\UsersManager\UserAccessFilter;
+use Piwik\Plugins\UsersManager\UsersManager;
+use Piwik\Plugins\UsersManager\Validators\Email;
+use Piwik\Plugins\UsersManager\Validators\Login;
+use Piwik\Validators\BaseValidator;
+use Piwik\Validators\IdSite;
+use Piwik\Plugin;
+
+
+class UserRepository
```
sgiehl, line 24: Actually I wonder why you moved some of the methods to this new `Repository` class. Looking at the methods it actually does not much what I would expect from a repository pattern. Checking access for anything would be something I wouldn't expect for example. That should be done by methods using a repository pattern to access data.

plugins/Login/Controller.php
```diff
@@ -497,4 +510,141 @@ public function logout()
...
+                $model->updateUserFields($user['login'], ['password' => $password, 'invite_status' => 'accept']);
+                $sessionInitializer = new SessionInitializer();
+                $auth = StaticContainer::get('Piwik\Auth');
+                $auth->setTokenAuth(null); // ensure authenticated through password
+                $auth->setLogin($user['login']);
+                $auth->setPassword($passwordConfirmation);
+                $sessionInitializer->initSession($auth);
+
```
sgiehl, line 580: It would be good to introduce new events like `Login.invitation.accept` and `Login.invitation.decline`, that we could listen on e.g. for plugins like `ActivityLog`

plugins/Login/tests/UI/Decline_spec.js
```diff
@@ -0,0 +1,25 @@
+/*!
+ * Matomo - free/libre analytics platform
+ *
+ * login & password reset screenshot tests.
```
sgiehl, line 4: either remove such comments when you copy a file or update it so it matches the content.

plugins/Login/tests/UI/Invite_spec.js
```diff
@@ -0,0 +1,38 @@
+/*!
+ * Matomo - free/libre analytics platform
+ *
+ * login & password reset screenshot tests.
```
sgiehl, line 4: either remove such comments when you copy a file or update it so it matches the content.

plugins/Login/Controller.php
```diff
@@ -497,4 +510,141 @@ public function logout()
...
+        try {
+            $mail = StaticContainer::getContainer()->make(UserDeclinedInvitationEmail::class, array(
+              'login'        => $user['login'],
+              'emailAddress' => $user['email'],
+              'userLogin'    => $user['login'],
+            ));
+            $mail->safeSend();
+        } catch (\Exception $e) {
+
+        }
```
sgiehl, line 642: the try/catch here is unneeded as well.

plugins/Login/lang/en.json
```diff
@@ -54,6 +54,15 @@
         "SuspiciousLoginAttemptsInLastHourEmail3": "Ensure your password is a secure, random password of an appropriate length (for example, at least 30 characters).",
         "SuspiciousLoginAttemptsInLastHourEmail4": "Setup two-factor auth so attackers will need more information than just your password in order to login.",
         "SuspiciousLoginAttemptsInLastHourEmail5": "Additionally, if your Matomo has a limited set of users or IPs through which users will access it, it may be beneficial to setup a IP address allowlist. %1$sRead our docs for more information.%2$s",
-        "LoginNotAllowedBecauseUserLoginBlocked": "Login functionality is temporarily disabled since we've a suspicious amount of failed login attempts in the last hour."
+        "LoginNotAllowedBecauseUserLoginBlocked": "Login functionality is temporarily disabled since we've a suspicious amount of failed login attempts in the last hour.",
+        "InvitationTitle": "Accept Invitation",
+        "InvitationDeclineTitle": "Decline Invitation",
+        "InvitationDeclineBody": "Are you sure you want to decline this Invitation?",
```
sgiehl, line 60: We should try to use correct english where possible. As far as I know in English most stuff should be written in lower case, unless it's a name or similar.
```suggestion
        "InvitationTitle": "Accept invitation",
        "InvitationDeclineTitle": "Decline invitation",
        "InvitationDeclineBody": "Are you sure you want to decline this invitation?",
```

plugins/Login/templates/invitation.twig
```diff
@@ -0,0 +1,83 @@
...
+                            <div class="row">
+                                <div class="col s12 input-field">
+                                    <input type="text" name="login" value="{{ user.login }}" size="20" readonly
+                                           tabindex="0"/>
+                                    <label><i class="icon-user icon"></i> {{ 'Login_LoginOrEmail'|translate }}</label>
+                                </div>
+                                <div class="col s12 input-field">
+                                    <input type="password" placeholder="" name="password" id="password" class="input" value="" size="20"
+                                           autocorrect="off" autocapitalize="none"
+                                           tabindex="1" required/>
```
sgiehl, line 31: Is there are reason for having `required` for the password, but not for the password confirmation. Actually it would make sense to have it on both, so the browser prevents sending an empty password confirmation.

plugins/Login/Controller.php
```diff
@@ -497,4 +510,141 @@ public function logout()
...
+            }
+
+            if (!$error) {
+                $password = UsersManager::getPasswordHash($password);
+                $passwordInfo = $passwordHelper->info($password);
+
+                if (!isset($passwordInfo['algo']) || 0 >= $passwordInfo['algo']) {
+                    // password may have already been fully hashed
+                    $password = $passwordHelper->hash($password);
+                }
```
sgiehl, line 570: I guess those checks are actually not needed. Why should the password that was provided in the input fields be already hashed??

plugins/Login/tests/Fixtures/PendingUsers.php
```diff
@@ -0,0 +1,54 @@
...
+namespace Piwik\Plugins\Login\tests\Fixtures;
+
+use Piwik\Date;
+use Piwik\Plugins\UsersManager\Model;
+use Piwik\Tests\Framework\Fixture;
+
+/**
+ * Generates tracker testing data for our APITest
+ *
+ * This Simple fixture adds one website and tracks one visit with couple pageviews and an ecommerce conversion
```
sgiehl, line 17: no. it doesn't....

plugins/Login/tests/UI/Invite_spec.js
```diff
@@ -0,0 +1,38 @@
...
+  it('password confirmation error', async function () {
+    await page.type('#password', 'abcd1234');
+    await page.type('#password_confirm', 'abcd123');
+    await page.evaluate(function(){
+      $('#login_form_submit').click();
+    });
+    expect(await page.screenshot({ fullPage: true })).to.matchImage('wrong_password');
+
+  });
+
```
sgiehl, line 37: Wouldn't have a test made sense, where the user really accepts the invitation. The users should be logged in automatically, so we could check if he is afterwards showing the dashboard or similar. That prevent possible regressions in that process in the future.

plugins/UsersManager/API.php
```diff
@@ -1539,12 +1430,37 @@ private function getDeviceDescription()
...
+
+        $this->checkUserIsNotAnonymous($userLogin);
+
+        if (!$this->model->getPendingUser($userLogin)) {
+            throw new Exception(Piwik::translate("UsersManager_ExceptionUserDoesNotExist", $userLogin));
+        }
+
+        $this->model->updateUserFields($userLogin, ['invite_status' => 'pending']);
+        $this->userRepository->sendNewUserEmails($userLogin, $expired, false);
+        Cache::deleteTrackerCache();
```
sgiehl, line 1462: Why exactly should it be required to clear the tracker cache here?

plugins/UsersManager/tests/Integration/APITest.php
```diff
@@ -390,7 +390,7 @@ public function test_updateUser_failsIfPasswordTooLong()
     public function test_update_user_fails_if_email_exists_as_other_user_username()
     {
         $this->expectException(\Exception::class);
-        $this->expectExceptionMessage('UsersManager_ExceptionEmailExistsAsLogin');
+        $this->expectExceptionMessage('UsersManager_CurrentPasswordNotCorrect');
```
sgiehl, line 393: Why does this test result change? There shouldn't have been any changes in the addUser/updateUser methods that should cause this.

plugins/UsersManager/tests/Integration/APITest.php
```diff
@@ -414,7 +414,7 @@ public function test_update_can_update_user_email_to_own_username()
     public function test_cannot_create_user_if_email_exists_as_username()
     {
         $this->expectException(\Exception::class);
-        $this->expectExceptionMessage('UsersManager_ExceptionEmailExistsAsLogin');
+        $this->expectExceptionMessage('UsersManager_ExceptionLoginExists');
```
sgiehl, line 417: This also looks rather unintended.

plugins/UsersManager/tests/Integration/UsersManagerTest.php
```diff
@@ -1084,6 +1090,50 @@ public function testGetAvailableCapabilities()
...
+        $content = <<<END
+<p>General_HelloUser</p>
+<p>CoreAdminHome_UserInviteSubject</p>
+<a target="_blank" href="$host?module=Login&action=acceptInvitation&token=thisisatoken"
+>CoreAdminHome_AcceptInvite</a> |
+<a target="_blank" href="$host?module=Login&action=declineInvitation&token=thisisatoken"
+>CoreAdminHome_DeclineInvite</a>
+<p><b>Notes:</b>CoreAdminHome_UserInviteNotes</p>
+END;
+        $this->assertEquals($content, $view->render());
```
sgiehl, line 1132: What exactly are you trying to test here? Testing if Twig is really able to fill in some placeholders doesn't really look like a useful test case for me. But it would make sense to check if inviting a user sends out an email...

plugins/UsersManager/tests/UI/UsersManager_spec.js
```diff
@@ -36,7 +36,13 @@ describe("UsersManager", function () {
         expect(await page.screenshotSelector('.usersManager')).to.matchImage('load');
     });
 
+    it('should show resend confirm when resend clicked', async function () {
+        await (await page.jQuery('.resend')).click();
+        expect(await page.screenshotSelector('.usersManager')).to.matchImage('resend_popup');
```
sgiehl, line 41: might have been enough to only take a screenshot of the confirmation popover, to avoid updating the screenshot something else on that page changes.

plugins/UsersManager/vue/src/UserEditForm/UserEditForm.vue
```diff
@@ -94,8 +94,9 @@
           </div>
           <div>
             <Field
-              :model-value="theUser.password"
-              :disabled="isSavingUserInfo || (currentUserRole !== 'superuser' && !isAdd)
+               v-if="isPendingUser"
```
sgiehl, line 97: This doesn't seem to work. It is hidden after an user has been invited and the edit screen is displayed. But when I click in edit for a user in pending state in the users list, the password field will be shown.

plugins/UsersManager/vue/src/UserEditForm/UserEditForm.vue
```diff
@@ -128,13 +129,25 @@
...
+                    :disabled="isAdd && (!firstSiteAccess || !firstSiteAccess.id)"
+                    :saving="isSavingUserInfo"
+                    @confirm="saveUserInfo()"
+                />
+              </div>
+              <div class="col s12 m6">
+                <div v-if="isAdd" class="form-help">
+                     <span class="inline-help"
+                      v-html="$sanitize(
+                          translate('UsersManager_InviteSuccessNotification', [7]))"></span>
```
sgiehl, line 147: This actually seems to be shown never. When a new user is invited there is a success message "The changes have been saved", but this message isn't shown.

plugins/UsersManager/vue/src/UserEditForm/UserEditForm.vue
```diff
@@ -564,13 +578,17 @@ export default defineComponent({
...
     },
     saveButtonLabel() {
       return this.isAdd
-        ? translate('UsersManager_CreateUser')
+        ? translate('UsersManager_InviteUser')
         : translate('UsersManager_SaveBasicInfo');
     },
+    isPendingUser() {
+      // eslint-disable-next-line eqeqeq
+      return this.user && (this.theUser.invited_at === '' || !this.theUser.invited_at);
```
sgiehl, line 590: shouldn't this check be based on the invite status?

plugins/UsersManager/vue/src/PagedUsersList/PagedUsersList.vue
```diff
@@ -676,3 +737,8 @@ export default defineComponent({
   },
 });
 </script>
+<style scoped>
+.actions-cell {
+  text-align: left!important;
+}
```
sgiehl, line 743: What benefit has it to introduce a scope style here. The component already has it's own stylesheet, where that property could have been easily changed if needed:
https://github.com/matomo-org/matomo/blob/d036954668dcd4cf8cb4a5235cbed7519503d0f6/plugins/UsersManager/vue/src/PagedUsersList/PagedUsersList.less#L37-L41

plugins/UsersManager/tests/UI/UsersManager_spec.js
```diff
@@ -166,6 +172,12 @@ describe("UsersManager", function () {
     });
 
     it('should delete a single user when the modal is confirmed is clicked', async function () {
+
+        await page.evaluate(function () {
+            $('select[name=access-level-filter]').val('string:view').change();
+            $('#user-text-filter').val('ight').change();
+        });
```
sgiehl, line 179: Why should this be needed now? Can't see any change that should require this test change.
Also one of the ongoing tests now has a unexpected screenshot, where no users are shown at all. 
See https://github.com/matomo-org/matomo/blob/4b38727169e378c3fdc77a4a0ed3ede20cf00ab0/plugins/UsersManager/tests/UI/expected-screenshots/UsersManager_delete_bulk_access.png

plugins/UsersManager/vue/src/User.ts
```diff
@@ -12,6 +12,7 @@ interface User {
   password?: string;
   email: string;
   role?: string;
+  invited_at?: string;
```
sgiehl, line 15: seems the `invitead_at` field, that existed in a previous version of this PR, was meanwhile removed. So this should be removed as well or changed to `invite_status`

plugins/UsersManager/vue/src/UserEditForm/UserEditForm.vue
```diff
@@ -471,6 +484,7 @@ export default defineComponent({
         this.firstSiteAccess = null;
         this.isSavingUserInfo = false;
         this.isUserModified = true;
+        this.theUser.invited_at = 'xx';
```
sgiehl, line 487: What should that be good for 🤔

plugins/UsersManager/vue/src/UserEditForm/UserEditForm.vue
```diff
@@ -319,6 +332,7 @@ const DEFAULT_USER: User = {
   uses_2fa: false,
   password: '',
   email: '',
+  invited_at: '',
```
sgiehl, line 335: This seems no valid user property anymore.

sgiehl (commented):
Here would have really a lot changes been required before I would have considered this PR as mergeable.

The most important ones are:
- Using a token_auth as invite token is kind of a security problem, as it allows using the API without even accepting the invite
- The tests are not covering the whole process. It's nice that all screens are covered by a UI test, but it would also be important to really click the `accept` or `decline` button and see what happens afterwards. Not having tests for that has the risk, that this might regress in the future and we won't see that.
- The user experience of this new feature is quite bad, as in a lot places the success notifications are missing and the users doesn't see if his action actually worked or not.
- ~Some~ Most of the requirements of the original issue aren't fulfilled. To mention some important ones of the list here: https://github.com/matomo-org/matomo/issues/13321#issuecomment-964729030
   - Activity log should recognise "invite user" and "invite user accepted".
     --> see comments about missing / changed events
   - A user can also decline the invite in which case the person that created the invite is being notified by email and we remove the user entry to have no personal data in there. (we might need to store in the user table who invited the user, if that user login no longer exists then we don't notify anyone)
   --> currently the user remains in the database
   - When a user clicks on accept invite, they enter their password. If privacy policy or terms is configured, then we show these links. We say eg if privacy policy is configured: By signing up, I accept the Privacy Policy. If terms is configured we say it similarly for terms. If both are configured then we mention both.
   --> currently there is always Terms & Conditions displayed.
   - In addition the UI currently doesn't allow to see when a invite expire, nor is it possible to show all invited users (e.g. by a filter).
   - Invited users can be seen by super users, and the person that invited that user.
   --> currently everyone who has permission can see the user, as invites are created as a normal users.

bx80:
Thanks for that @sgiehl, you are right -  there are several issues here that should have been caught during review and will need to be addressed with a fix PR.

With the auth token issue, what do you think about adding a boolean flag  `invite_token` to the `user_token_auth` table to indicate that the token is only valid for user invitation, similar to the existing `system_token` option? Any attempt to use an `invite_token` for general API access could then be rejected in addition to removing the token entirely once the invite is accepted / rejected?

sgiehl:
@bx80 using the token auth table for that might not have been the best choice, as it will require to change all the existing code, that by default when checking for tokens it will need to ignore such special tokens. 

Personally I might have added two fields to the user table. `invite_token` and `invite_expire`
I'm not sure if we would need a status field at all.
If someone is invited, the generated token will be stored in `invite_token`, together with the expire datetime in `invite_expire`.
Every user where both fields are filled is in the state `pending`.
If the expire datetime passes, we can show `expired` as status.
If a user declines the request, the requirements actually defined, that the user should be removed, to avoid storing any personal data, so this status isn't needed.
If a user accepts the invite, both invite fields would be cleared and such users will have the status `active`.

peterhashair:
@sgiehl that makes sense, I will rework the table structure on that one.

PR #18903, opened 2022-03-09
Fix PHP8 Trim Error

deybhayden:
### Description:

An error happens if sites if names or skus are arrays in PHP 8 (was just a warning in PHP 7)

> Uncaught exception in core/Tracker/GoalManager.php line 567:
> trim(): Argument #1 ($string) must be of type string, array given

### Review

### DIFF
--- a/core/Tracker/GoalManager.php
+++ b/core/Tracker/GoalManager.php
@@ -561,8 +561,10 @@ class GoalManager
 
         foreach ($cleanedItems as $item) {
             $actionsToLookup = array();
-            list($sku, $name, $category, $price, $quantity) = $item;
+            list($sku_check, $name_check, $category, $price, $quantity) = $item;
+            $sku = is_array($sku_check) ? join(',', $sku_check) : $sku_check;
             $actionsToLookup[] = array(trim($sku), Action::TYPE_ECOMMERCE_ITEM_SKU);
+            $name = is_array($name_check) ? join(',', $name_check) : $name_check;
             $actionsToLookup[] = array(trim($name), Action::TYPE_ECOMMERCE_ITEM_NAME);
 
             // Only one category
--- a/tests/PHPUnit/System/TrackerTest.php
+++ b/tests/PHPUnit/System/TrackerTest.php
@@ -125,6 +125,26 @@ class TrackerTest extends IntegrationTestCase
         $this->assertActionEquals('scary <> movies', $conversionItems[0]['idaction_category']);
     }
 
+    public function test_trackingEcommerceOrder_WithNameAndSKUArrays()
+    {
+        // item sku, item name, item category, item price, item quantity
+        $ecItems = array(array(["sku1", "sku2"], ["name1", "name2"], 'category1', 12.99, 1));
+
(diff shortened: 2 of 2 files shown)


sgiehl (commented):
Hey @deybhayden. Thanks for creating this PR. The code looks fine.
I guess actually name and sku are meant to be passed as a string only, as multiple values for a single product won't make that much sense. But guess it shouldn't hurt to convert arrays to strings to avoid possible errors.

@tsteur what do you think about that? Is it fine to change that, or should an exception be thrown if the values are not provided as strings?

deybhayden:
Awesome! Yeah, I wasn't sure if the fix made sense to the wider community. However, when we upgraded an Matomo install for a client, we started hitting this error as tracking requests tried to process on PHP 8. We didn't have the ability to tell sites to change their tracking code, so this is the fix we went with to get past the exception.

tsteur:
I'm having mixed feelings :) On one side it's not a big harm, on the other side once it's in there we need to keep supporting it and it might cause a regression in the future if we need to change behaviour around that. Considering it's only done for SKU and name it may be fine by me. It be different if it was done for a category as products can be actually assigned to multiple categories. Note that if we merge this, it could regress any time and we wouldn't make it a supported feature. Meaning this could actually cause issues again in the future if we support this now, then stop supporting it for some reason and then we're breaking it. I mean when people don't realise they send an array there. It should all be edge case though and be fine to merge I suppose. @sgiehl do you have any thoughts?

@deybhayden did this tracking actually ever work using an array? I wonder if it wouldn't have recorded like `array` as a product name?

deybhayden:
So, the change I did here isn't exactly what we did for our client, most of the examples were like this:

```php
$name = ['shirt'] 
```

So, we just `$name[0]` if `$name_check` was an array. But, then yea, that did work properly for tracking Ecomm stuff. Totally open to whatever you all think best, but we just got bit here so I thought I'd bring it up.

sgiehl:
@deybhayden actually the tracking on the website of the client seems to be incorrect, as they are sending an array for name, which shouldn't be the case. Our documentation also says it should be given as string.
When using the javascript tracker you actually can't set the sku to an array, as the code would prevent that here (or throw an error):
https://github.com/matomo-org/matomo/blob/05082abdca8909cf27137773f129cf66f495922b/js/piwik.js#L6745-L6755

I guess the name could be set to an array at that point and would then be passed to the tracking. We actually also could change the javascript tracker so it throws an error if someone tries to set an sku or name as an array. So someone implementing the tracking directly sees it won't work.
Nevertheless this wouldn't prevent such tracking request coming e.g. directly from a PHP tracker or another sdk. 

We could merge this to be a bit less error-prone when incorrect data is sent with the request. But as @tsteur this won't be officially supported and could break or change the behavior again any time in the future.

deybhayden:
Nice sleuthing! I didn't know enough to be able to dig into the JS library to follow the steps all the way through. I'm happy to change the code to throw errors or update the JS if that's the right path? I'm definitely a bit out of depth when it comes to the right solve.

tsteur:
@sgiehl sounds good to merge it 👍

PR #18905, opened 2022-03-09
Set the _pk_id in the past when deleting cookie

### LINKED ISSUE #18902: Set the _pk_id in the past when deleting cookie.
When the _pk_id cookie is set to expire it sets the current date. When tools like cookiebot scans for cookies and have a timing anomaly potentially the cookie can still seem active. 

Can we have the cookie set to a time in the past to avoid this issue.

aleksijohansson:
### Description:

Fixes: https://github.com/matomo-org/matomo/issues/18902

Sets the expiration date of the cookies being deleted to 36 hours in the past. There can be 26 hours difference between timezones, so 36 could be a safe bet to make sure the cookies always get deleted immediately. Previously the cookie expiration was only set to ~86 seconds in the past.

The existing OptOutForm UI test should be sufficient for validating this change.

### Review


justinvelluppillai:
86400 is a suspicious number of milliseconds. Sounds like it was meant to be 24 hours.

aleksijohansson:
The build failures (4 out of 12 jobs failing) on specific PHP environments seem to not be related to this change. The only failing test (test_piwikJs_minified_isUpToDate) that seems it could be related only appears on 2 of the failing jobs and does not appear on the other 10 which suggests that it is also not directly related to this change.

aleksijohansson:
@justinvelluppillai indeed it could be that it was meant to be 24 hours originally since 86400 is the number of seconds (not milliseconds) in 24 hours.

sgiehl:
> The only failing test (test_piwikJs_minified_isUpToDate) that seems it could be related

@aleksijohansson Changing the piwik.js requires to rebuild the minified versions of that file. Details how to do that can be found here: https://github.com/matomo-org/matomo/blob/4.x-dev/js/README.md#deployment

aleksijohansson:
@sgiehl I looked into that earlier, but since only two environments were failing on this issue I wasn't sure if this really needs to be done manually or if the test suite would include this minifying. Thanks for the confirmation that it needs to be done manually!

aleksijohansson:
The build failures (3 out of 12 jobs failing) on specific PHP environments seem to **not** be related to this change.

sgiehl:
@aleksijohansson Yes those tests are unrelated. Thanks for pushing the minified JS. And we actually have an action that can update the minified JS, but that one doesn't work for pull requests coming from a fork.

sgiehl (approved):

PR #18913, opened 2022-03-09
Fix possible error on PHP 8.1

### LINKED ISSUE #18910: Deprecated - trim() in PHP 8.1
```
WARNING: /core/CliMulti/Process.php(291): Deprecated - trim(): Passing null to parameter #1 ($string) of type string is deprecated - Matomo 4.8.0 - Please report this message in the Matomo forums: https://forum.matomo.org (please do a search first as it might have been reported already) (Module: Installation, Action: getSystemCheck, In CLI mode: false)
```

1. Install Matomo 4.8.0
2. Have PHP 8.1

## Context
Browse URL:
```
/index.php?date=yesterday&module=CorePluginsAdmin&format=html&action=plugins&idSite=8&period=day&segment=&widget=&showtitle=1&random=9226
```

## Your Environment
* Matomo Version: 4.8.0
* PHP Version: 8.1
* Server Operating System: Debian 10

sgiehl:
### Description:

fixes #18910

### Review

### DIFF
--- a/core/CliMulti/Process.php
+++ b/core/CliMulti/Process.php
@@ -288,6 +288,11 @@ class Process
     {
         $exec = $command . ' > /dev/null 2>&1; echo $?';
         $returnCode = @shell_exec($exec);
+
+        if (false === $returnCode || null === $returnCode) {
+            return false;
+        }
+
         $returnCode = trim($returnCode);
         return 0 == (int) $returnCode;
     }


justinvelluppillai (approved):

justinvelluppillai:
A few test failures here but seem unrelated.

core/CliMulti/Process.php
```diff
@@ -288,6 +288,11 @@ private static function returnsSuccessCode($command)
     {
         $exec = $command . ' > /dev/null 2>&1; echo $?';
         $returnCode = @shell_exec($exec);
+
+        if (false === $returnCode || null === $returnCode) {
+            return false;
```
sgiehl, line 293: It just came to my mind that this will actually change the current behavior. But maybe this might actually fix some possible issues.
`shell_exec` returns `false` or `null` if something didn't work with the command, otherwise the return code is returned.
If an error occurred and e.g. `null` or `false` was returned, the `trim` triggered a warning on PHP 7 and maybe even PHP 8. The error should only occur on PHP 8.1.
After the trim `$returnCode` is an empty string in case of an error and `0 == (int) $returnCode` actually evaluates to `true` and we assume the command succeeded, which it actually didn't.

tsteur: @sgiehl not sure I understand it changes behaviour. Looks like this maybe worked before? https://3v4l.org/7C8Qf

Looks generally good though 👍

justinvelluppillai: previously if `$returnCode` was `false` or `null`, this function would return `true`, now it returns `false`.

justinvelluppillai: I think this is ok to merge. This change in behaviour would almost certainly be better if it affected anyone at all

PR #18917, opened 2022-03-09
[Vue] migrate manage goals directive and twig templates to Vue

diosmosis:
### Description:

Changes:
* Convert _addEditGoal.twig and dependent twig templates + piwik-manage-goals directive to Vue.
* Add exception for target=_blank to DOMPurify configuration (only if rel="noopener noreferrer" is present).
* Allow model value for FieldCheckbox to be a number as well as boolean.
* Remove use of `const` in piwikHelper.js as that should be vanilla JS.
* Modify piwikHelper.compileVueEntryComponents so it can support multiple named slots.

Notes:

* There are several twig events that had to be respected during migration. BC is provided through two separate methods:
  - **vue-entry + named slots**: Using the Vue component directly (ManageGoals), and supplying named slots where the twig `postEvent` call is used. This is the simpler approach and works in cases where the twig template was not parameterized. It doesn't work when, eg, the event is called inside a twig for loop, as that can't be translated to a v-for loop in the frontend.
  - using {% set %} to save the result of a postEvent call in twig, passing this into the template, and referencing this as a dynamic component via `<component :is=...`. This allows us to iterate over data in the twig template, save the result and pass it to Vue for use. The `<component :is` usage is done specifically so the HTML is directly inserted. Using v-html would require a parent element which is not always acceptable (eg, when a postEvent() call is made between `<th>` or `<td>` elements.

Both approaches will be documented.

### Review


peterhashair (approved):

plugins/Goals/vue/src/ManageGoals/ManageGoals.vue
```diff
@@ -0,0 +1,766 @@
...
+
+            {{ translate('Goals_Optional') }} {{ translate('Goals_Ecommerce') }}:
+            <span v-html="youCanEnableEcommerceReports"/>
+          </span>
+        </div>
+
+        <table v-content-table>
+          <thead>
+            <tr>
+              <th class="first">Id</th>
```
sgiehl: We can use `General_Id` here, so it is translatable

sgiehl (changes requested):
@diosmosis This seems to work smoothly on the reporting pages. But when going to goal management within the admin it seems the info about funnels and multi attribution are not shown:

image

It's also not loaded when editing or creating the goal there.

diosmosis:
@sgiehl hopefully its fixed

sgiehl (approved):
works now 👍

PR #18925, opened 2022-03-11
[Vue] remove use of lazyInitSingleton, creating singleton stores outside of global scope cannot work

diosmosis:
### Description:

Seems the approach of creating a plain JS object store (ie, not a vuex store or something similar) outside of the global scope will never work due to how computed properties work in Vue. When a computed property is defined, it creates a ReactiveEffect instance that is attached to the current active "effect scope", if any. When that effect scope is destroyed, the reactive effect is removed, and changes no longer propagate through the computed properties.

When a computed property is created in a global scope, outside of a component, there is no active effect scope. When created within a component, the component's effect scope is used. This means if the component is destroyed, for example because it is used in a `v-if`, the reactive effects to the global singleton store will be removed, and they won't get added again, since the singleton store will not be recreated.

This PR removes the method & its uses and reverts to the more verbose approach of manually initializing stores via AJAX requests in the components that use them (rather than doing that in the store's constructor).

### Review


diosmosis:
Depends on https://github.com/matomo-org/tag-manager/pull/454 being merged first

sgiehl:
@diosmosis is this PR based on another one? It includes the migration of the CustomDimensions stuff. Is that on purpose?

diosmosis:
@sgiehl No not on purpose. Fixed.

peterhashair (approved):

PR #18935, opened 2022-03-14
Don't create cookies when disabled

### LINKED ISSUE #17892: Don't create test cookies when cookies are disabled

justinvelluppillai:
### Description:

Fixes #17892 

Supersedes #18135 (created a few new PRs to get around submodule issues).

This PR moves visitorId cookie creation a bit later, ie only when needed for `trackPageView()` or similar call. VisitorId is now stored internally in the tracker to offset some of the impact of this, but in some cases where multiple trackers are being used there may be some functional differences as implied by the changes I've made to the tests where they only pass when `trackPageView()` is called before other tracker methods.

### Review


justinvelluppillai:
build js

justinvelluppillai:
All relevant tests now passing, so this is ready for review with particular consideration to implications for other plugins or unusual tracking setups.

sgiehl:
I looked through the code changes and imho they are looking fine. The only "issue" I could see is the slightly changed behavior when using multiple trackers. But that actually should only affect using the visitorid in the trackers before a first tracking request was send. To circumvent this, I guess the only possible solution would be to somehow store the initital visitorid in a global scope, that can be accessed in each tracker instance. But not sure if that is needed.
@tsteur do you think we should move that to a global scope?

tsteur:
@sgiehl I don't quite understand the issue. AFAIK, we expect each tracker to have a different visitorId, not the same. But I might remember it wrongly. In what case exactly would it cause an issue? Like how can it be reproduced?

justinvelluppillai:
These tests suggest we want the same visitorId across trackers https://github.com/matomo-org/matomo/blob/e33a147f1c74524ef584ee80fc5c77b8fb6b59aa/tests/javascript/index.php#L2285. There's maybe no issue there but there is a change of behaviour which could be removed by setting a global visitorId on the page.

tsteur:
They are only the same if they point to the same URL and site. 

If there are different trackers used (pointing to a different site, matomo url, or cookie prefix), then they should be different. Which we can also see in  https://github.com/matomo-org/matomo/blob/e33a147f1c74524ef584ee80fc5c77b8fb6b59aa/tests/javascript/index.php#L2303-L2312

Meaning it shouldn't be global if I see this right.


Because in https://github.com/matomo-org/matomo/blob/e33a147f1c74524ef584ee80fc5c77b8fb6b59aa/tests/javascript/index.php#L2291 we are tracking a page view, I would have expected the cookie would be set and we shouldn't see a change in behaviour. 

So the test is definitely expecting the right thing, the question is why isn't the `delayedTracker` reading the same visitorId. Maybe it's not returning the correct tracker instance? My random guess might be calling `tracker.removeAllAsyncTrackersButFirst()` in the beginning might fix the issue in case there's a tracker configured differently from a previous test.

justinvelluppillai:
@tsteur that line 2291 I added. The change in behaviour was that tracking a pageview is now needed where before this would have worked without it

tsteur:
@justinvelluppillai if someone did call `getVisitorId` we could also set the cookie if not set yet? This would fix the issue maybe?

And if someone didn't call `getVisitorId` both trackers would still use the same visitor ID as long as they read from the same cookie I would assume?

sgiehl:
@tsteur what about use cases where `getVisitorId` might be called before the permission to set cookies was granted?

tsteur:
@sgiehl I don't think that be a problem that often (using `getVisitorId` and manually creating the tracker)? Do you maybe have a specific implementation example where this be a problem? Do I see this right if `getVisitorId` is called too early it would return an empty visitor ID?

justinvelluppillai:
If we set the cookie when `getVisitorId` is called then we still have the original problem of potentially having cookies set when disabled. `getVisitorId` wouldn't return an empty visitor ID if called early, it would call `loadVisitorIdCookie` which will generate a UUID and return a new visitor ID if no cookie exists yet.

tsteur:
That be fine in that case @justinvelluppillai 👍  At least then people would have a way to fix this issue by calling `getVisitorId` after configuring cookie consent etc. And we improve the situation for 99.8% of the people

justinvelluppillai:
build js

justinvelluppillai:
build js

justinvelluppillai:
build js

justinvelluppillai:
build js

justinvelluppillai:
Nice one thanks @tsteur. I've implemented that approach here and it passes tests without making those changes 👍 

A couple of unrelated tests fail but this is good for review.

sgiehl:
@justinvelluppillai there is also `getVisitorInfo`, which currently might return varying userids until a cookie was set, guess we should set the cookie in that method as well.
Also I'm not sure if it is expected that the visitorId is currently empty until the cookie could be set. But I guess this actually already was the case before.

justinvelluppillai:
build js

justinvelluppillai:
@sgiehl I've added to now set visitor id cookie in `getVisitorInfo` also. And yes, this doesn't change behaviour from before with the visitorId before the cookie is set, except previously the visitorId cookie was set immediately if possible.

justinvelluppillai:
build js

js/piwik.js
```diff
@@ -7071,6 +7074,15 @@ if (typeof window.Matomo !== 'object') {
             });
 
             Matomo.trigger('TrackerSetup', [this]);
+
+            Matomo.addPlugin('TrackerVisitorIdCookie' + uniqueTrackerId, {
+                // if no tracking request was sent we refresh the visitor id cookie on page unload
+                unload: function () {
+                    if (!hasSentTrackingRequestYet) {
+                        setVisitorIdCookie();
```
sgiehl, line 7082: Wouldn't this mean, that if the user leaves the page (without previous tracking), but did not yet gave consent a cookie would be set nevertheless?

justinvelluppillai: My understanding is that the `setCookie` function takes care of that by respecting `configCookiesDisabled` which is set to `true` if consent is required. I could more explicitly add the check here if needed.

justinvelluppillai: Also remember previously we called `setVisitorIdCookie()` much earlier, so this doesn't change behaviour other than delaying this call.

sgiehl: was just wondering as you do a check for `(!configCookiesDisabled && !getCookie(getCookieName('id')))` within `getVisitorInfo`, but not here.

justinvelluppillai: good point, I'll actually remove the check on `getVisitorInfo()` as it is unnecessary.

justinvelluppillai:
build js

sgiehl (approved):

PR #18940, opened 2022-03-14
Fix added to replace translation strings with %% to %

AltamashShaikh:
### Description:

Fix added to replace translation strings with %% to %

Eg: piwik_translations['HeatmapSessionRecording_HeatmapSampleRateHelp']
`Also known as 'traffic'. When you select 100%%, all visitors that visit the selected target page will be recorded. When you select for example 10%%, only every 10th visitor will be recorded. The lower the percentage you select, the longer it will take to reach the selected sample limit.`

we have similar replace on server side added same in JS

### Review

### DIFF
--- a/plugins/Morpheus/javascripts/piwikHelper.js
+++ b/plugins/Morpheus/javascripts/piwikHelper.js
@@ -13,6 +13,8 @@ function _pk_translate(translationStringId, values) {
         if (typeof values != 'undefined' && values && values.length) {
             values.unshift(translation);
             return sprintf.apply(null, values);
+        } else if (values && values.length === 0) {
+            translation = translation.replaceAll('%%', '%');
         }
 
         return translation;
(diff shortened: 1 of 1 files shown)


plugins/Morpheus/javascripts/piwikHelper.js
```diff
@@ -13,6 +13,8 @@ function _pk_translate(translationStringId, values) {
         if (typeof values != 'undefined' && values && values.length) {
             values.unshift(translation);
             return sprintf.apply(null, values);
+        } else if (values && values.length === 0) {
```
sgiehl: Guess simply using an else would be easier and should also work in all cases

sgiehl (changes requested):

plugins/Morpheus/javascripts/piwikHelper.js
AltamashShaikh: Updated as suggested

sgiehl (approved):

PR #18953, opened 2022-03-16
Add Two New segments

### LINKED ISSUE #15985: New segment for 'Ecommerce Order Revenue' (and ideally also for 'Revenue left in cart')
It would be valuable to offer 2 new segments:

* Order Revenue (Ecommerce)
* Revenue left in cart  (Ecommerce)

It is already possible to segment on Order ID, Product name/sku/category. But revenue segmentation is not yet available.

It would be useful in many contexts for example create a segment for:
* Large orders
* ecommerce orders within a certain price range (using AND combination)
* (for example orders for "free trials" with a revenue of zero)
*  having the new segment means it'll now be possible to delete "Buggy visits with large abandoned carts" from skewing the data using our [GDPR tool delete visits data](https://matomo.org/faq/how-to/how-do-i-delete-rows-of-a-report-and-specific-visits-to-clean-up-some-of-the-reporting-data/) (Very large abandoned cart)

peterhashair:
### Description:

Fixes: #15985 

Add Two New segment Order Revenue (Ecommerce). Revenue left in cart (Ecommerce)

### Review


peterhashair:
Revenue left in the cart still needs a condition on the query, but not sure how to do that. I notice lists of segments maybe needs a update as well https://developer.matomo.org/api-reference/reporting-api-segmentation

sgiehl:
@peterhashair Adding new column classes has a lot bigger impact than you may think. At least the way it is currently done. Having methods to update the column values shouldn't be needed here at all, as that would duplicate code.
I guess the simplest solution would be to overwrite the `configureSegments` method in the class `Piwik\Plugins\Ecommerce\Columns\Revenue` and add both segments there. Btw. Revenue left in cart might be selectable using the idGoal `IDGOAL_CART`

peterhashair:
@sgiehl thanks for the hints, I am still struggling to add an extra condition  `idGoal =-1`, not sure where is the best place to put that, should I make a new Columns and update in `getDbDiscriminator` function?

sgiehl:
@peterhashair Unfortunately this one won't be easy. the `getDbDiscriminator` method is actually only used for the CustomReports plugin. So it won't have any effect on segments. I would need to have a closer look at that as well to provide a detailed solution. But I would start checking if we can use a segment sqlFilter for that.

peterhashair:
@sgiehl thanks for the hints, seems working, with that.
```php
     $segment->setSqlFilter(function ($valueToMatch, $sqlField, $matchType) {
            $table = Common::prefixTable($this->dbTableName);
            $sql = "select {$sqlField} from {$table} where idgoal = -1 and {$sqlField}{$matchType} ? ";
            return [
              'SQL'  => $sql,
              'bind' => $valueToMatch,
            ];
        });

```

peterhashair:
@sgiehl I think somehow it works. but on the bulk tests, `TwoVisitsWithCustomVariablesSegmentMatchNONETest::testApi`. Somewhere bind map to a wrong field, still trying to fight out why.

plugins/Ecommerce/Columns/Revenue.php
```diff
@@ -27,6 +32,44 @@ public function getDbDiscriminator()
...
+    {
+        //new Segment revenue on order
+        $segment = new Segment();
+        $segment->setCategory($this->category);
+        $segment->setName(Piwik::translate('Ecommerce_OrderRevenue'));
+        $segment->setSegment('revenueOrder');
+        $segment->setSqlSegment('log_conversion.idvisit');
+        $segment->setSqlFilter(function ($valueToMatch, $sqlField, $matchType) {
+            $table = Common::prefixTable($this->dbTableName);
+            $sql = " SELECT {$sqlField} from {$table} WHERE (log_conversion.revenue {$matchType} ?) ";
```
sgiehl: doesn't that query need to include `idgoal = 0`? Otherwise it would include the revenue of any goal, even from abandoned carts.

sgiehl (changes requested):
@peterhashair You need to "transform" the `$matchType` parameter. In the segment definition it is `==`, but in SQL it would need to be `=`. You can find something similar in the Segment class of MediaAnalytics plugin for example.

peterhashair:
@sgiehl I think this is working. But not sure where should I document this.

plugins/Ecommerce/Columns/Revenue.php
```diff
@@ -27,6 +33,55 @@ public function getDbDiscriminator()
...
+     * @param false $cart
+     * @return array
+     */
+    private function getRevenueQuery($valueToMatch, $sqlField , $matchType, $idGoal)
+    {
+        if ($matchType === SegmentExpression::MATCH_EQUAL) {
+            $matchType = "=";
+        }
+        $table = Common::prefixTable($this->dbTableName);
+        $sql = " SELECT {$sqlField} from {$table} WHERE (idgoal = {$idGoal} and log_conversion.revenue {$matchType} ?) ";
```
sgiehl: Did you really test that one? When adding such segment queries you should always check the queries that are generated.

This query is actually not doing what you might expect, at least not if matomo is using a table prefix. Assuming someone uses the table prefix `matomo_`, this query would be like `SELECT log_conversion.idvisit FROM matomo_log_conversion WHERE (idgoal = 0 and log_conversion.revenue <= 0)`

Looking e.g. at the query that is generated to fetch the visits log this results in:
```sql
SELECT /* idSegments = [13] */ log_visit.*
FROM matomo_log_visit AS log_visit
LEFT JOIN matomo_log_conversion AS log_conversion ON log_conversion.idvisit = log_visit.idvisit
WHERE 
  (log_visit.idsite in (?) AND log_visit.visit_last_action_time >= ? AND log_visit.visit_last_action_time <= ?) 
AND (
    (log_conversion.idvisit IN (
       SELECT log_conversion.idvisit
         FROM matomo_log_conversion
         WHERE (idgoal = 0 AND log_conversion.revenue >= ?)
)))
GROUP BY log_visit.idvisit
ORDER BY log_visit.idsite DESC, log_visit.visit_last_action_time DESC
LIMIT 0, 27
```
You might notice that the subquery now actually references some fields from the outer query.

plugins/Ecommerce/Columns/Revenue.php
```diff
@@ -27,6 +33,55 @@ public function getDbDiscriminator()
...
+     * @param $sqlField
+     * @param $matchType
+     * @param false $cart
+     * @return array
+     */
+    private function getRevenueQuery($valueToMatch, $sqlField , $matchType, $idGoal)
+    {
+        if ($matchType === SegmentExpression::MATCH_EQUAL) {
+            $matchType = "=";
+        }
```
sgiehl, line 89: You should better handle all supported types here (like in the file I mentioned earlier) and throw an exception for not supported types. Otherwise it would throw an sql error if someone defines a segment like `revenueOrder!@5`

sgiehl (changes requested):
This doesn't work correctly yet.

plugins/Ecommerce/Columns/Revenue.php
```diff
@@ -27,6 +33,70 @@ public function getDbDiscriminator()
...
+        //supported operator type
+        $supportType = [
+          SegmentExpression::MATCH_EQUAL,
+          SegmentExpression::MATCH_GREATER_OR_EQUAL,
+          SegmentExpression::MATCH_LESS_OR_EQUAL,
+          SegmentExpression::MATCH_GREATER,
+          SegmentExpression::MATCH_LESS,
+        ];
+
+        if (!in_array($valueToMatch, $supportType)) {
```
sgiehl: This can't work as you are comparing the wrong variables.

sgiehl (changes requested):
Have you actually checked/tested you changes once? Looking at the failing tests you would have noticed that the ones with the new segments are failing.

peterhashair:
@sgiehl sorry, should be fixed now.

sgiehl:
@peterhashair Seems that is working now. Regarding documentation. The segment listing doesn't need an update, as it is included from demo.matomo.cloud. See https://github.com/matomo-org/developer-documentation/blob/live/docs/4.x/reporting-api-segmentation.md
So once, that was deployed, the list will be updated automatically.

A general documentation around segmentation would also be good to have. Currently it seems there is none, so I have created https://github.com/matomo-org/developer-documentation/issues/630. Feel free to pick that one and start adding a new documentation with the findings you had while developing this issue.

PR #18980, opened 2022-03-22
[Vue] add support for on demand loading of plugin UMD files when a component in a plugin is first used.

diosmosis:
### Description:

This PR provides on demand loading of a plugin's Vue module, so it does not have to be loaded on each page load.

To set a plugin's Vue module to be loaded on demand you'd have to:
- override Plugin::shouldLoadUmdOnDemand() and return true
- if the plugin defines angularjs adapters, move them all to a separate file manually included via the AssetManager event and set the component property to a reference like `{ plugin: 'MyPluginName', component: 'MyComponentName' }`.

If you need to use a plugin that's loaded like this in another plugin, you won't be able to import as you would normally. Instead you'll need to use the `importPluginUmd` method:

```
import { importPluginUmd } from 'CoreHome';

importPluginUmd('TagManager').then((module: typeof import('TagManager')) => {
  // ...
});
```

Note: that adapters for stores or anything that gets added via angular(...).adapter will not be available, so making a plugin load on demand is a potential BC break. For TagManager this is acceptable since the old models weren't meant to be reused.

To see a plugin that uses this, see: https://github.com/matomo-org/tag-manager/pull/461

### Review


diosmosis:
Adding this to the current milestone, but I'm not sure when it is desired to include it.

core/FrontController.php
```diff
@@ -494,14 +493,10 @@ protected function prepareDispatch($module, $action, $parameters)
 
     protected function handleMaintenanceMode()
     {
-        if ((GeneralConfig::getConfigValue('maintenance_mode') != 1) || Common::isPhpCliMode() ) {
```
peterhashair: I think `GeneralConfig::getConfigValue` is the new function get General config. Maybe we should use `GeneralConfig::getConfigValue`

core/AssetManager/UIAssetFetcher/PluginUmdAssetFetcher.php
```diff
@@ -109,6 +113,38 @@ private function getAllPluginUmds()
...
+            // if the plugin dependencies has a plugin that is in $this->plugins, but not in $pluginsToLoadOnInit,
+            // the dependency was set to load on demand, and we report an error since this should only happen
+            // during development.
+            //
+            // if it's not in either $this->plugins and not in $pluginsToLoadOnInit, then it's not activated
+            // and we ignore it.
+            if (!empty(array_diff($pluginDependencies, $pluginsToLoadOnInit))
+                && empty(array_diff($pluginDependencies, $this->plugins))
+            ) {
+                throw new \Exception("Missing plugin dependency: $pluginName requires plugins "
```
peterhashair, line 131: $pluginName  should that be variable instead of string

core/FrontController.php
```diff
@@ -494,14 +493,10 @@ protected function prepareDispatch($module, $action, $parameters)
...
+        if ((Config::getInstance()->General['maintenance_mode'] != 1) || Common::isPhpCliMode()) {
             return;
         }
-
-        // as request matomo behind load balancer should not return 503. https://github.com/matomo-org/matomo/issues/18054
-        if (GeneralConfig::getConfigValue('multi_server_environment') != 1) {
-            Common::sendResponseCode(503);
-        }
+        Common::sendResponseCode(503);
```
peterhashair: I think this was related to an issue, that has a load balancer on Matomo.

core/FrontController.php
diosmosis: Hm, I think maybe this change shouldn't have been committed, it's got nothing to do w/ the PR. I'll revert it.

core/AssetManager/UIAssetFetcher/PluginUmdAssetFetcher.php
diosmosis: I'm not sure what you mean by this, can you provide a code suggestion to clarify?

peterhashair: I mean 

```suggestion
                throw new \Exception("Missing plugin dependency: ".$pluginName." requires plugins "
```

sgiehl:
@diosmosis would you mind rebasing that one to 5.x-dev? We've decided not to put that into the last 4.x release to lower the risk of regressions we might need to fix. Also we will have more time to test that, after it was merge to 5.x-dev...

diosmosis:
@sgiehl should be up to date w/ 5.x-dev

plugins/CoreHome/vue/src/index.ts
```diff
@@ -130,3 +135,14 @@ export { default as ContentIntro } from './ContentIntro/ContentIntro';
...
+// for plugin modules loaded on demand that provide angularjs adapters, the adapters
+// have to be loaded through this event. this is because they must be defined before
+// the angular app is bootstrapped, so they must be in the javascripts/ folder.
+// but they must also be defined after CoreHome is loaded since that is where the
+// createAngularJsAdapter function is defined, and CoreHome is loaded after
+// javascripts/* scripts are. the event allows the adapters to be loaded right after
+// CoreHome is loaded at initial page load.
+//
+// Note: this will be removed in Matomo 5 and is only here for the interim.
+Matomo.postEvent('Matomo.addAngularJsAdapters', createAngularJsAdapter);
```
sgiehl: Can this be removed now, as we directly merge this to Matomo 5?

sgiehl (changes requested):
Tested it with the TagManager PR and seems to work smoothly. Left a comment. Otherwise I guess this should be good to merge.

plugins/CoreHome/vue/src/index.ts
diosmosis: Yes, can definitely be removed :+1:

diosmosis:
fixed

sgiehl:
@diosmosis Could you check the failing UI tests? Seems the language selection is no longer visible in installation and updater. Seems there is a javascript error in that case.

diosmosis:
@sgiehl fixed

sgiehl (approved):
Works now and tests are also passing again

PR #18994, opened 2022-03-24
[Vue] Build plugin module dependencies if type output does not exist locally, otherwise build will fail.

diosmosis:
### Description:

If a developer tries to build a plugin's Vue code without building anything else, it will fail because the dependent types are not present, and the error message won't tell the dev what the exact problem is. This PR changes the build command to build dependencies of the requested plugin if the types are not present. Avoids the failure entirely.

### Review


sgiehl (approved):
Guess I had that issue a couple of times already but wasn't sure what the issue was, as building all files always worked 🙈

PR #18995, opened 2022-03-24
Fix issue where date ranges might include too many subperiods 

### LINKED ISSUE #18992: get_class(): Argument #1 ($object) must be of type object, bool given in MultiSites/API.php
Seen below error a few times. This could be a regression maybe?

> Error: {"message":"get_class(): Argument #1 ($object) must be of type object, bool given","file":"plugins\/MultiSites\/API.php","line":344,"request_id":"e1f08","backtrace":" on plugins\/MultiSites\/API.php(344)\n#0 plugins\/MultiSites\/API.php(344): get_class(false)\n#1 plugins\/MultiSites\/API.php(352): Piwik\\Plugins\\MultiSites\\API->calculateEvolutionPercentages(Object(Piwik\\DataTable\\Simple), false, Array)\n#2 plugins\/MultiSites\/API.php(260): Piwik\\Plugins\\MultiSites\\API->calculateEvolutionPercentages(Object(Piwik\\DataTable\\Map), Object(Piwik\\DataTable\\Map), Array)\n#3 plugins\/MultiSites\/API.php(189): Piwik\\Plugins\\MultiSites\\API->buildDataTable(Array, 'day', '2022-03-01,toda...', false, false, false, false, Array)\n#4 [internal function]: Piwik\\Plugins\\MultiSites\\API->getOne('2', 'day', '2022-03-01,toda...', false, false, false)\n#5 core\/API\/Proxy.php(244): call_user_func_array(Array, Array)\n#6 core\/Context.php(28): Piwik\\API\\Proxy->Piwik\\API\\{closure}()\n#7 core\/API\/Proxy.php(335): Piwik\\Context::executeWithQueryParameters(Array, Object(Closure))\n#8 core\/API\/Request.php(266): Piwik\\API\\Proxy->call('\\\\Piwik\\\\Plugins\\\\...', 'getOne', Array)\n#9 plugins\/API\/Controller.php(46): Piwik\\API\\Request->process()\n#10 [internal function]

URL: https://example.matomo.org/?module=API&method=MultiSites.getOne&format=json&idSite=2&token_auth=XYZANONYMIZED&period=day&date=2022-03-01,today

Referrer: None

GET: {"module":"API","method":"MultiSites.getOne","format":"json","idSite":"2","token_auth":"XYZANONYMIZED","period":"day","date":"2022-03-01,today","filter_limit":100}

sgiehl:
### Description:

When using a magic keyword (like `today`) as end of a period and providing a timezone other than `UTC`, the array of subperiods could actually hold too many periods.

*For debugging purpose:*
The start and end date for the range are initialized here:
https://github.com/matomo-org/matomo/blob/dde206cdfd68a1638aebbe3ae1335e92e6375065/core/Period/Range.php#L256-L267

When using `today` and a timezone like `Europe/Berlin` the `$endDate` would hold a date like `2022-03-24 01:00:00` (without setting the time to `00:00:00`). The start date would be used as provided, but with time `00:00:00`
When generating the the subperiods here:
https://github.com/matomo-org/matomo/blob/dde206cdfd68a1638aebbe3ae1335e92e6375065/core/Period/Range.php#L417-L422
The compare in the while would actually match once more as expected, as the time of the dates differs (`00:00:00` < `01:00:00`) and thus one more date than expected is added.

fixes #18992

### Review

### DIFF
--- a/tests/PHPUnit/Unit/Period/RangeTest.php
+++ b/tests/PHPUnit/Unit/Period/RangeTest.php
@@ -20,6 +20,12 @@ use Piwik\Period\Year;
  */
 class RangeTest extends BasePeriodTest
 {
+    public function setUp(): void
+    {
+        parent::setUp();
+        Date::$now = null;
+    }
+
     /**
      * @dataProvider getDateXPeriodsAgoProvider
      */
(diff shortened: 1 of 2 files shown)


tests/PHPUnit/Unit/Period/RangeTest.php
```diff
@@ -182,6 +182,16 @@ public function testRangeComma3_EndDateIncludesToday()
         $this->assertEquals('2008-01-03', $subPeriods[2]->toString());
     }
 
+    // test range date1,date2
+    public function testRangeComma4_EndDateIncludesTodayWithTimezone()
```
tsteur, line 192: can we also add some tests for other timezones that are say +10 hours and -10 hours?

tsteur: and also check `getEndDate()->toDatetime()` or so?

tsteur (commented):
@sgiehl would this be needed in the `if (preg_match('/^(last|previous)([0-9]*)$/', $this->strDate, $regs)) {` block as well?

Do I see this right there is basically no change when the `timezone` is not set when creating a range? I looked through all the usages of `range` when a timezone is set and then this change should be good. If it does change behaviour for UTC then it may be more complicated

tests/PHPUnit/Unit/Period/RangeTest.php
sgiehl: add some more tests

sgiehl:
@tsteur The issue only happens if the range ends with a magic date word like yesterday or today. If that is the case and timezone differs from UTC the new code makes a difference. Otherwise nothing changes.

tsteur (approved):
I see now. It works for the other one because the startDate is calculated off the endDate and they both have the same "time".

PR #18996, opened 2022-03-24
Add the request hostname to error & exception logs

samjf:
Ref: DEV-2584

### Description:

When viewing the logs of a multi-tenant install it is difficult to determine which Matomo instance was requested when the error (log entry) occurs. 

A simple solution is to print the host from the incoming headers. Whilst this is not safe against tampering, it is only printed in logs and will be effective for normal traffic.

**Notes**

I preferred injecting the error message prefix when providing the prefix in `core/testMinimumPhpVersion.php` due to the fact that the code seems to be early in the bootstrap and must be compatible with early versions of PHP. The `filter_var` validation function of the hostname/domain has a minimum php requirement of 5.2.

### Review


core/Common.php
```diff
@@ -1234,4 +1234,18 @@ protected static function checkValidLanguagesIsSet($validLanguages)
...
+     * @param $hostname
+     * @return false|string
+     */
+    public static function getValidHostname($hostname = null)
+    {
+        if (empty($hostname) && isset($_SERVER['HTTP_HOST'])) {
+            $hostname = $_SERVER['HTTP_HOST'];
+        }
+        return filter_var($hostname, FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME);
+    }
```
sgiehl: This method might be better placed in the `Url` class. There is actually already `Url::getHost(false)`, which could also be used in this method to get the value from server variables. This would automatically use `SERVER_NAME` instead of `HTTP_HOST` if configured (and config is already available).

samjf: Thanks for that suggestion @sgiehl . I made use of `Url::getHost` for the host string and I've moved the method into `Url`. I updated the method name so it is clear it is validating against RFC standards and not trusted hosts.

sgiehl:
@samjf This looks good so far. Is it on purpose that only exceptions / errors that might get logged outside of matomo's logging will be prefixed with the host? Messages logged to the matomo log files won't include the hostname. For that guess it would need to be added somewhere in https://github.com/matomo-org/matomo/blob/115527353a9e75e01aa4d263408956ae45403bea/plugins/Monolog/Formatter/LineMessageFormatter.php

samjf:
@sgiehl I don't think I've really had a problem with the matomo file log. Correct me if i'm wrong but It seems that the hostnames are being logged for me (checking locally). Perhaps archiving/commands already adds them on a multitenant?
image

The main problem was trying to distinguish between a large amount of errors on the PHP/FPM error/exception logging.

sgiehl:
@samjf Guess our cloud plugin has a special log processor, that adds the instanceid. We could consider having that in core and only log it for multitenant maybe.
But as that won't affect many people at all, guess we can save the effort for now.

sgiehl (approved):

samjf:
Many thanks @sgiehl 
That might be a good contribution from cloud in the future. Hard tracking down exactly where it was added haha.

PR #19000, opened 2022-03-25
Clear general cache less often when launching the archiving

tsteur:
### Description:

fix DEV-2707

Since Matomo 4.7 we noticed that we often have thousands of general cache invalidations within a minute. Around 100 general cache invalidations within approx 4 seconds but depends on how many archives there are to be invalidated. It causes one general cache invalidation per site every time we call `prepareArchive`. Aka if you have thousands of sites, and run heaps of archives, and all sites have regular traffic, this can cause a huge massive amount of invalidations.

image

All these cache invalidations are also particularly a problem as we're constantly clearing the cache, and also constantly populate the tracker cache every ms as there are concurrent tracking requests happening. 

The problem is coming from this backtrace which is called once per site when preparing an archive run. So this can be called very often this logic and cause a huge amount of invalidations.

```

#0  Piwik\Tracker\Cache::clearCacheGeneral() called at [core/Archive/ArchiveInvalidator.php:332]
#1  Piwik\Archive\ArchiveInvalidator->markArchivesAsInvalidated() called at [plugins/CoreAdminHome/API.php:166]
#2  Piwik\Plugins\CoreAdminHome\API->invalidateArchivedReports() called at [core/CronArchive.php:928]
#3  Piwik\CronArchive->invalidateWithSegments() called at [core/CronArchive.php:873]
#4  Piwik\CronArchive->invalidateArchivedReportsForSitesThatNeedToBeArchivedAgainImpl() called at [core/CronArchive.php:808]
#5  Piwik\CronArchive->Piwik\{closure}() called at [core/Tracker/Cache.php:302]
#6  Piwik\Tracker\Cache::withDelegatedCacheClears() called at [core/CronArchive.php:809]
#7  Piwik\CronArchive->invalidateArchivedReportsForSitesThatNeedToBeArchivedAgain() called at [core/CronArchive/QueueConsumer.php:182]
#8  Piwik\CronArchive\QueueConsumer->getNextArchivesToProcess() called at [core/CronArchive.php:394]
#9  Piwik\CronArchive->run() called at [core/CronArchive.php:280]
#10 Piwik\CronArchive->Piwik\{closure}() called at [core/Access.php:661]
#11 Piwik\Access::doAsSuperUser() called at [core/CronArchive.php:286]
#12 Piwik\CronArchive->main() called at [plugins/CoreConsole/Commands/CoreArchiver.php:32]
#13 Piwik\Plugins\CoreConsole\Commands\CoreArchiver->execute() called at [vendor/symfony/console/Symfony/Component/Console/Command/Command.php:257]
#14 Symfony\Component\Console\Command\Command->run() called at [vendor/symfony/console/Symfony/Component/Console/Application.php:874]
#15 Symfony\Component\Console\Application->doRunCommand() called at [vendor/symfony/console/Symfony/Component/Console/Application.php:195]
#16 Symfony\Component\Console\Application->doRun()
```

To improve the situation my thought was to invalidate the general cache only every 4 seconds. We could technically only clear the cache after marking all archives as invalid. However, this could take a while depending on the number of archives to invalidate and the number of sites. 

If it takes a few minutes between marking an archive as invalid and calling `clearGeneralCache`, then we risk that the given date won't be re-archived later under circumstances since the tracker wouldn't know we have invalidated the archive and therefore it wouldn't call `rememberToInvalidateArchivedReportsLater` and then we wouldn't archive that date again under some circumstances. It's bit hard to explain but generally we need a general cache clear ideally immediately after marking an archiving as invalid.

My trade-off was that it's unlikely that the only tracking request for the given date happens during those 5 seconds. And for today periods we always archive the date again the next day after midnight so it shouldn't be an issue.

Any thoughts on this? I haven't fully tested it yet but wanted to check for thoughts first.

There could be other solutions like having the information to `rememberToInvalidateArchivedReportsLater` in a tracker site cache but thinking this can make things worse as we can't optimise it as well as this one and causes again a lot of invalidations.

### Review

### DIFF
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -271,6 +271,10 @@ class ArchiveInvalidator
 
         $invalidationInfo = new InvalidationResult();
 
+        // when browser archiving is used then to be safe we always want to invalidate the general cache as otherwise
+        // invalidating of today might not happen.
+        $requiresGeneralCacheInvalidation = Rules::isBrowserTriggerEnabled();
+
         // quick fix for #15086, if we're only invalidating today's date for a site, don't add the site to the list of sites
         // to reprocess.
         $hasMoreThanJustToday = [];
@@ -282,7 +286,13 @@ class ArchiveInvalidator
                 && count($dates) == 1
                 && ((string)$dates[0]) == ((string)Date::factoryInTimezone('today', $tz))
             ) {
+                // date is for today
                 $hasMoreThanJustToday[$idSite] = false;
+            } else {
+                // date is not for today. this means we need to invalidate the general cache so a new tracking request
+                // for the same date can set the flag again that another archive invalidation is needed.
+                // this behaviour is not needed for today as we always force archiving of "yesterday" anyway.
+                $requiresGeneralCacheInvalidation = true;
             }
         }
 
@@ -331,7 +341,7 @@ class ArchiveInvalidator
             }
         }
 
-        if ($clearCache) {
+        if ($clearCache && $requiresGeneralCacheInvalidation) {
             Cache::clearCacheGeneral();
         }
 
--- a/core/Archive/ArchiveInvalidator.php
+++ b/core/Archive/ArchiveInvalidator.php
@@ -271,10 +271,6 @@ class ArchiveInvalidator
 
         $invalidationInfo = new InvalidationResult();
 
-        // when browser archiving is used then to be safe we always want to invalidate the general cache as otherwise
-        // invalidating of today might not happen.
-        $requiresGeneralCacheInvalidation = Rules::isBrowserTriggerEnabled();
-
         // quick fix for #15086, if we're only invalidating today's date for a site, don't add the site to the list of sites
         // to reprocess.
         $hasMoreThanJustToday = [];
@@ -288,11 +284,6 @@ class ArchiveInvalidator
             ) {
                 // date is for today
                 $hasMoreThanJustToday[$idSite] = false;
-            } else {
-                // date is not for today. this means we need to invalidate the general cache so a new tracking request
-                // for the same date can set the flag again that another archive invalidation is needed.
-                // this behaviour is not needed for today as we always force archiving of "yesterday" anyway.
-                $requiresGeneralCacheInvalidation = true;
             }
         }
 
@@ -341,7 +332,7 @@ class ArchiveInvalidator
             }
         }
 
-        if ($clearCache && $requiresGeneralCacheInvalidation) {
+        if ($clearCache) {
             Cache::clearCacheGeneral();
         }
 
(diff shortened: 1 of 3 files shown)


sgiehl:
I actually wasn't aware of that special caching mechanism. But I had a closer look. As far as I have seen those options are actually only stored to the general cache, as the tracker uses the value to check if the archives need to be invalidated or not. If I did not miss anything that is actually only done for tracking requests with a date before today. See https://github.com/matomo-org/matomo/blob/3b00013819863472afe30cc04de49c4365c22d21/core/Tracker/Visit.php#L622-L628
*Note: Maybe we could even change that to before yesterday, as we always rearchive yesterday nevertheless, don't we?*

So for most websites tracking older visits is very unlikely. The cache of those options seems only to be used within the Tracker. So regarding clearing the cache in `markArchivesAsInvalidated`, I was wondering if we can't simply skip it if another period than day should be invalidated or if the invalidated date is today (or yesterday). As in that cases updating the cache won't have any effect at all.

Your solution could be used on top maybe to avoid too many invalidations when older dates are invalidated.
Hope that makes sense...

tsteur:
That's a great point @sgiehl 💯  This should reduce the general cache invalidations already quite a bit possibly (although not sure we set the flag for today or not).

I've adjusted the logic to
 
* Only clear general cache if the date is not for today (using the [same logic as in ArchiveInvalidator](https://github.com/matomo-org/matomo/blob/4.8.0/core/Archive/ArchiveInvalidator.php#L280-L282))
* And only clear cache every 4 seconds only if any of the previous dates actually needed an invalidation

The logic is bit complicated to understand maybe but should tweak the performance well. I haven't fully tested it yet but wanted to see if you have any thoughts on this first @sgiehl ?

sgiehl:
@tsteur I spent some more thoughts on that. Actually I think we did that a lot more complex than actually needed.
Clearing the cache is actually only needed if an option for a remembered invalidation was removed from the database.
That is triggered here:
https://github.com/matomo-org/matomo/blob/5a3ae2d7ff17d1d6bd11ffa65f78101c296bd99e/core/Archive/ArchiveInvalidator.php#L318-L331

So if we would change the `forgetRememberedArchivedReportsToInvalidate` to return true if a record has been removed, we could only clear the cache if that happened at least once within the foreach.
That should limit the number of cache clears to the number of option records in the database.

tsteur:
@sgiehl great point, I didn't notice 🚀  that made it a lot easier. 

For now I also removed the code that would invalidate only every 4 seconds as it be hard to implement with the deleted flag. I would do it this way, and then see if it's still a problem or not. If it then is still a problem, would add some extra code to invalidate less often.

Because in https://github.com/matomo-org/matomo/blob/3b00013819863472afe30cc04de49c4365c22d21/core/Tracker/Visit.php#L622-L628 which you pointed out we don't invalidate for today anyway, I also removed the check for `isToday` which wasn't needed in the first place.

sgiehl (approved):
Looks good to me now.
@tsteur should this one go into the upcoming 4.9 or shall we wait for 4.10?

tsteur:
@sgiehl I'll leave this up to you. Safest be probably to put it into 4.10 and we could always apply a patch for this on production.

sgiehl:
ok. moved it to 4.10..

PR #19012, opened 2022-03-28
[Vue] migrate scheduled reports management from twig/angularjs to vue

diosmosis:
### Description:

Changes:
* Migrated scheduled reports management from twig/angularjs to vue.
* Migrated MobileMessaging extension to ScheduledReports to Vue.
* Add redirectOnSuccess option to AjaxHelper.
* Recognize vue-entry-ignore attribute in compileVueEntryComponents() to avoid mounting vue components even if they have vue-entry.
* Add extraProps to compileVueEntryComponents() to pass extra property values to each vue-entry.

### Review


diosmosis:
Converting to draft for a bit, might be possible to provide full BC.

plugins/ScheduledReports/ScheduledReports.php
```diff
@@ -126,6 +126,37 @@ public function getClientSideTranslationKeys(&$translationKeys)
...
+        $translationKeys[] = 'ScheduledReports_DescriptionOnFirstPage';
+        $translationKeys[] = 'SegmentEditor_ChooseASegment';
+        $translationKeys[] = 'ScheduledReports_WeeklyScheduleHelp';
+        $translationKeys[] = 'ScheduledReports_MonthlyScheduleHelp';
+        $translationKeys[] = 'ScheduledReports_ReportPeriod';
+        $translationKeys[] = 'ScheduledReports_ReportPeriodHelp';
+        $translationKeys[] = 'ScheduledReports_ReportPeriodHelp2';
+        $translationKeys[] = 'ScheduledReports_ReportHour';
+        $translationKeys[] = 'ScheduledReports_ReportType';
+        $translationKeys[] = 'ScheduledReports_ReportFormat';
```
sgiehl: That one seems to be a duplicate
```suggestion
```

plugins/ScheduledReports/vue/src/ManageScheduledReport/ManageScheduledReport.vue
```diff
@@ -0,0 +1,406 @@
...
+      AjaxHelper.post(
+        {
+          method: 'ScheduledReports.sendReport',
+        },
+        {
+          idReport,
+          force: true,
+        },
+      ).then(() => {
+        this.fadeInOutSuccessMessage(
```
sgiehl, line 226: I guess it was already the same before, but `fadeInOutSuccessMessage` currently always does a page reload. At this point that is actually not needed, as no new information could be displayed anyway.

plugins/ScheduledReports/vue/src/ManageScheduledReport/ManageScheduledReport.vue
```diff
@@ -0,0 +1,406 @@
...
+          method: isCreate ? 'ScheduledReports.updateReport' : 'ScheduledReports.addReport',
+          period,
+          hour,
+        },
+        apiParameters,
+        {
+          redirectOnSuccess: true,
+        },
+      ).then(() => {
+        this.fadeInOutSuccessMessage(
```
sgiehl, line 378: That method does a page reload, so guess we could remove the `redirectOnSuccess: true,` above, as the request caused by that gets aborted directly.

plugins/ScheduledReports/templates/index.twig
```diff
@@ -9,19 +9,40 @@
...
+
+{# load the Template.reportParametersScheduledReports event twice, once here outside of vue so any inline scripts
+   will execute. then again in the vue slot so it will be used correctly. This is hack to provide some level of BC. #}
+<div style="display:none" vue-entry-ignore ng-non-bindable>
+    {{ reportParametersScheduledReportsEvent|raw }}
+</div>
+
+<div
+  vue-entry="ScheduledReports.ManageScheduledReport"
+  title="{{ title|default(null)|json_encode|e('html_attr') }}"
```
sgiehl: Using an attribute named `title` causes a tooltip to be shown everywhere within that div. So might be better to use another name here.

sgiehl: In addition the `json_encode` here also causes special charts like umlauts to be displayed encoded in the tooltip.
Does Vue always perform a json_decode on input attributes? Or might that happen for other parameters within vue as well?

plugins/ScheduledReports/vue/src/ListReports/ListReports.vue
```diff
@@ -0,0 +1,246 @@
...
+              :id="`downloadReportForm_${report.idreport}`"
+              :action="linkTo({
+                module: 'API',
+                segment: null,
+                method: 'ScheduledReports.generateReport',
+                idReport: report.idreport,
+                outputType: downloadOutputType,
+                language: language,
+                format: ['html', 'csv', 'tsv'].indexOf(report.format) !== -1
+                  ? report.format : undefined,
```
sgiehl: We maybe could use `original` here as default. Currently downloading a SMS report won't work, as the API request doesn't contain a format and uses `XML` as default.

plugins/ScheduledReports/vue/src/AddReport/AddReport.vue
```diff
@@ -0,0 +1,546 @@
...
+          :model-value="report.idsegment"
+          @update:model-value="$emit('change', { prop: 'idsegment', value: $event })"
+          :options="savedSegmentsById"
+        >
+          <template v-slot:inline-help>
+            <div
+              id="reportSegmentInlineHelp"
+              class="inline-help-node"
+              v-if="segmentEditorActivated"
+              v-html="reportSegmentInlineHelp"
```
sgiehl: this misses a `$sanitize`

plugins/ScheduledReports/vue/src/AddReport/AddReport.vue
```diff
@@ -0,0 +1,546 @@
...
+          </div>
+        </div>
+      </div>
+      <SaveButton
+        :value="saveButtonTitle"
+        @confirm="$emit('submit')"
+      />
+      <div
+        class="entityCancel"
+        v-html="entityCancelText"
```
sgiehl: this misses a `$sanitize`

sgiehl (changes requested):
Left a couple of comments for smaller improvements / fixes. Otherwise looks good.

plugins/ScheduledReports/templates/index.twig
diosmosis: There's a JSON.parse() in the code that handles `vue-entry` in `piwikHelper`. Vue doesn't exist in HTML, all props in Vue are already instantiated, unserialized JavaScript values.

sgiehl:
@diosmosis just found one last regression on this one. If the description of the report contains a special char like `<` or `"` it will be displayed encoded. Guess the "problem" is that the title is already stored encoded in the database. Before it was using `rawSafeDecoded` in twig. Not sure if we already have something similar in vue to use.

diosmosis:
@sgiehl :+1: we'd just decode it client side. Vue just sets textContent/innerText so its safe (unless using v-html which uses innerHTML).

sgiehl (approved):

PR #19021, opened 2022-03-30
Remove invalid hosts tracker settings cache

samjf:
### Description:

Ref: DEV-2583

When a tracking request supplies and invalid host an empty folder structure will be made in the cache directory if the GLOBAL `ENABLE_CONFIG_PHP_CACHE` is `true`. This change removes cleans up the directory once we can determine it is an invalid host.

**Notes**
Ideally we wouldn't even create this structure to begin with, however, the vendor lib we use for the file cache  throws an exception if a directory path isn't supplied. It was easier just to clean up in the case we have an invalid host after the fact. For the time being this will solve arbitrary dirs stacking up in the tmp folder.

### Review


core/Config/Cache.php
```diff
@@ -85,4 +85,23 @@ public function doDelete($id)
...
+            \RecursiveIteratorIterator::CHILD_FIRST
+        );
+        foreach ($iterator as  $name => $file) {
+            if ($file->isDir()) {
+                $path = $file->getPathname();
+                rmdir($path);
+            }
+        }
+        // remove invalid host dir
+        rmdir($targetDir);
```
sgiehl: Guess this actually does the same as `Filesystem::unlinkRecursive()`

samjf: @sgiehl Thanks, I hadn't seen this one 🤦 Though, it doesn't do exactly what we want as it turns out. I wanted this method to be cautious and not delete anything unless it is an empty dir. It seems `unlinkRecursive` would remove files inside the dirs too.

sgiehl (changes requested):
Can't we maybe avoid the path creation at all if we make the valid host check a static method and move the cache creation within that check?

tsteur:
@sgiehl AFAIK we can't because we first need to look into the config to see if it's a valid host before we can know it is a valid host. Although thinking about it now, when using Matomo in a multi tenancy environment where all the config files are in `misc/user/$host` there would be a way by checking if the userSettingsFile exists. If a config file for a specific host exists, then we can assume it's a valid host. This may be good enough maybe. It be only an issue when using a regular `config/config.ini.php`. If I see this right then `ENABLE_CONFIG_PHP_CACHE` is not documented so this could work

```
diff --git a/core/Config/IniFileChain.php b/core/Config/IniFileChain.php
index 2801cbeb2..970466197 100644
--- a/core/Config/IniFileChain.php
+++ b/core/Config/IniFileChain.php
@@ -215,7 +215,7 @@ class IniFileChain
             $this->resetSettingsChain($defaultSettingsFiles, $userSettingsFile);
         }
 
-        $hasAbsoluteConfigFile = !empty($userSettingsFile) && strpos($userSettingsFile, DIRECTORY_SEPARATOR) === 0;
+        $hasAbsoluteConfigFile = !empty($userSettingsFile) && strpos($userSettingsFile, DIRECTORY_SEPARATOR) === 0 && is_file($userSettingsFile);
         $useConfigCache = !empty($GLOBALS['ENABLE_CONFIG_PHP_CACHE']) && $hasAbsoluteConfigFile;
```

samjf:
@tsteur That looks to be a good solution. I'm not sure if we could put it on that line of code because `reload` seems to be the main way to save a missing cache and `$useConfigCache` needs to be `true` for that to happen in a normal situation. 

We could add the same check just below:
```
diff --git a/core/Config/IniFileChain.php b/core/Config/IniFileChain.php
index bc37e84ef3..49efb2f71d 100644
--- a/core/Config/IniFileChain.php
+++ b/core/Config/IniFileChain.php
@@ -218,7 +218,7 @@ class IniFileChain
         $hasAbsoluteConfigFile = !empty($userSettingsFile) && strpos($userSettingsFile, DIRECTORY_SEPARATOR) === 0;
         $useConfigCache = !empty($GLOBALS['ENABLE_CONFIG_PHP_CACHE']) && $hasAbsoluteConfigFile;
 
-        if ($useConfigCache) {
+        if ($useConfigCache && is_file($userSettingsFile)) {
             $cache = new Cache();
             $values = $cache->doFetch(self::CONFIG_CACHE_KEY);
```

tsteur:
I'm not too much into it but would think it pretty much does the same in the end 👍

samjf:
I've stopped the cache from being created in two places in the IniChainSettings code now. The first one was ensuring that the cache being retrieved actually exists as a file. The second one is ensuring that the merged settings has a valid host setting present/set in the config. 
This was close to your suggestion @sgiehl , but it doesn't require actually loading the config (as the code does), just ensuring the config structure is as expected which turns out to be enough.

samjf:
Appears I have a few tests that need fixing: 
`Piwik\Tests\Unit\Config\IniFileChainCacheTest::test_reload_canReadFromCache with data set #0 `
and
`Piwik\Tests\Unit\Config\IniFileChainCacheTest::test_reload_canReadFromCache with data set #1 `

samjf:
I needed to update the test so that it didn't delete the original settings file when testing the cache. That step shouldn't be necessary to test that the cache is indeed being read; a new IniFileChain instance should suffice.

However, this will have an affect, if an original INI settings file is deleted then it will not attempt to read the cache for it (due to `$userSettingsFile` existence check in `reload()`). This will effectively invalidate it I believe. I didn't think this would be a problem, but let me know if it is troublesome.

The only other way to proceed would be the original cleanup method that originally was built.

sgiehl (approved):
Had a look through the changes. For me this looks good now. Can't think of a reason why this may cause issues. Will merge it once the tests ran through

PR #19029, opened 2022-03-30
[Vue] migrate widgetize plugin to vue

diosmosis:
### Description:

Changes:
* Migrate widget-preview directive to Vue component.
* Migrate export-widget controller to Vue component.

### Review


plugins/Widgetize/vue/src/ExportWidget/ExportWidget.vue
```diff
@@ -0,0 +1,169 @@
...
+  computed: {
+    dashboardCode() {
+      return getIframeCode(this.dashboardUrl);
+    },
+    allWebsitesDashboardCode() {
+      return getIframeCode(this.allWebsitesDashboardUrl);
+    },
+  },
+  methods: {
+    linkTo(params: QueryParameters) {
```
sgiehl, line 161: Would it maybe make sense to make that method global in some way. Iirc I have seen similar methods in other components already.

plugins/Widgetize/vue/src/ExportWidget/ExportWidget.vue
```diff
@@ -0,0 +1,169 @@
...
+  @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+-->
+
+<template>
+  <div class="widgetize">
+    <div v-content-intro>
+      <h2>
+        <EnrichedHeadline>{{ title }}</EnrichedHeadline>
+      </h2>
+      <p>With Matomo, you can <a
```
sgiehl, line 13: Would you mind moving all the text to translation files? Otherwise guess we can also create a small issue to do that another time.

plugins/Widgetize/vue/src/WidgetPreview/WidgetPreview.vue
```diff
@@ -0,0 +1,79 @@
...
+        idSite: Matomo.idSite,
+        period: Matomo.period,
+        date: MatomoUrl.urlParsed.value.date,
+        disableLink: 1,
+        widget: 1,
+      };
+
+      delete finalParams.action;
+      delete finalParams.module;
+
```
sgiehl: I guess those lines aren't correct (anymore). The final urls are now lacking `module` and `action` parameter.

sgiehl (changes requested):

plugins/Widgetize/vue/src/ExportWidget/ExportWidget.vue
diosmosis: As far as I know the Vue policy/suggestion is to always use computed properties instead of adding many global functions to Vue apps. This is likely because computed properties are cached, whereas function calls and other code in an component's template will be executed every time a component updates (as far as I know, again). We have two, $sanitize/translate now. translate is used ubiquitously so it would be very inconvenient to make computed properties for everything that calls it. $sanitize isn't and it would probably be faster to use it in computed properties only, really.

I could make these computed properties that just call MatomoUrl given linkTo is usually given hardcoded parameters, but personally I would say its not in the spirit of Vue to add global functions like that, especially when they are only used by a handful of components.

plugins/Widgetize/vue/src/ExportWidget/ExportWidget.vue
diosmosis: That's enough text for me to consider the change out of scope.

plugins/Widgetize/vue/src/ExportWidget/ExportWidget.vue
sgiehl: Thanks for the explanation. Makes sense now.

plugins/Widgetize/vue/src/ExportWidget/ExportWidget.vue
sgiehl: Ok. Will create a follow up issue, once this is merged.

sgiehl (approved):

PR #19030, opened 2022-03-30
[Bug]fix prefilght cors OPTIONS request record in the action visits

### LINKED ISSUE #18812: options request will also be recorded
## Expected Behavior

`options request` should not recorded!

## Current Behavior

`options request` will also be recorded

## Possible Solution

The server should filter out the `options request` and not participate in the record.

image

image

## Your Environment

* Matomo Version: 4.7.1
* MySQL version: 10.6.5-MariaDB-1:10.6.5+maria~focal
* PHP Version: 8.0.15
* Server Operating System: centos7

* Browser: chrome98
* Operating System: macos 12.1

peterhashair:
### Description:

Fixes: #18812 

Extend requests with `headers` and `method` do not record preflight cors in the database. 

To reproduce the issue one needs to trigger prefight cors, eg add a request header more details can be found [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) 
```js
       headers:{'X-PINGOTHER':'pingpong'}
```

### Review


core/Tracker/Request.php
```diff
@@ -60,8 +63,9 @@ class Request
     /**
      * @param $params
      * @param bool|string $tokenAuth
+     * @param null $server
```
justinvelluppillai: Should this be array|null ?

core/Tracker/Request.php
```diff
@@ -921,4 +928,12 @@ private function getVisitorIdAsBinary($idVisitor)
         }
         return false;
     }
+
+    /**
+     * @return bool
+     */
+    public function isPreFightCorsRequest()
+    {
+        return !empty($this->server['HTTP_ACCESS_CONTROL_REQUEST_HEADERS']) || !empty($this->server['HTTP_ACCESS_CONTROL_REQUEST_METHOD']);
```
sgiehl: I guess it should be fine to directly use `$_SERVER` here. There is no need to pass the server variables in constructor.

sgiehl (changes requested):

core/Tracker/Request.php
```diff
@@ -921,4 +924,12 @@ private function getVisitorIdAsBinary($idVisitor)
         }
         return false;
     }
+
+    /**
+     * @return bool
+     */
+    public function isPreFightCorsRequest()
+    {
+        return !empty($_SERVER['HTTP_ACCESS_CONTROL_REQUEST_HEADERS']) || !empty($_SERVER['HTTP_ACCESS_CONTROL_REQUEST_METHOD']);
```
sgiehl: Was it intended that this no longer checks for the request method?

core/Tracker/Action.php
```diff
@@ -381,6 +381,10 @@ public function loadIdsFromLogActionTable()
      */
     public function record(Visitor $visitor, $idReferrerActionUrl, $idReferrerActionName)
     {
+        if ($this->request->isPreFightCorsRequest()) {
+            return;
```
sgiehl: Actually I'm not sure if this is the correct place. It will only prevent a new action from being inserted. A visit would still be created. I would actually expect that the whole request should be dropped in that case.
Maybe we could also have a test, that proves no data was recorded for such a request.

sgiehl:
I'm not really deep into such CORS stuff, but I was wondering if we shouldn't send a proper answer to a CORS request.
According to https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request it seems the correct answer would a `204` with some additional headers containing the allowed methods.
@peterhashair did you test what happens if we send a normal answer to the preflight request? Is the tracking request still sent in that case?

peterhashair:
@sgiehl right that makes sense, will update that.

peterhashair:
@sgiehl I am trying to figure out how to do a response as 204, but I notice there is a set of requests, not just one, if I respond 204 in the foreach, will that affect another request?

peterhashair:
@sgiehl I am trying to figure out how to do a response as 204, but I notice there is a set of requests, not just one, if I respond 204 in the foreach, will that affect another request?

sgiehl:
@peterhashair You may need to handle that at an earlier stage. As it's not relevant if the preflight request might contain one or more tracking requests, you should do that before the request is actually parsed. Guess here it might suite:
https://github.com/matomo-org/matomo/blob/078909ef807910a1f2a4d93d738306d0bf6a9e04/core/Tracker.php#L130-L136

peterhashair:
@sgiehl it seems like the prefight works, just wondering should I use the site url store in the data for the `Access-Control-Allow-Origin`, otherwise there will be security issue I guess.
```php 
             header('Access-Control-Allow-Methods: GET, POST');
               header('Access-Control-Allow-Headers: *');
               header('Access-Control-Allow-Origin: *');
               Common::sendResponseCode(204);
```

core/Tracker.php
```diff
@@ -355,4 +363,13 @@ private static function isDebugEnabled()
 
         return false;
     }
+
+    public function isPreFightCorsRequest()
```
sgiehl: for new methods we should already try to add typehints where possible
```suggestion
    public function isPreFightCorsRequest(): bool
```

sgiehl (changes requested):
> @sgiehl it seems like the prefight works, just wondering should I use the site url store in the data for the Access-Control-Allow-Origin, otherwise there will be security issue I guess.

Actually I guess the `*` might be fine. We can't really know where a tracking request might come from. Using all configured sites might also not be correct, as people could also track unconfigured site urls.

core/Tracker/RequestSet.php
```diff
@@ -45,8 +45,8 @@ public function setRequests($requests)
             if (!$request instanceof Request) {
                 $request = new Request($request, $this->getTokenAuth());
             }
-
             $this->requests[] = $request;
+            Common::sendResponseCode(204);
```
sgiehl:
```suggestion
```

sgiehl:
I've now applied a couple of fixes and added a simple test to get this ready to merge...

